PR #3, opened 2014-07-16
[FIX]Automatically change type when changing sale flow to drop shipping ...

tornoz:
Hi,

When we change sale flow to Drop shipping on SOL it doesn't change type(Procurement method) to make to order which breaks drop shipping flow.

Steps to reproduce bug:
1. Open/Create any product type product and set Procurement Method:Make to Stock
2. Set supplier and do not mark drop shipping
3. Create SO(Sale order)
4. Add product from step one.
5. Change sale flow to drop shipping
6. Confirm Sale order
7. Run Procurement Scheduler

This commit fixes the bug

### DIFF
--- a/sale_dropshipping/sale.py
+++ b/sale_dropshipping/sale.py
@@ -68,7 +68,8 @@ class sale_order_line(orm.Model):
             if sale_flow == 'direct_delivery':
                 vals['type'] = 'make_to_order'
             else:
-                vals['type'] = self.pool.get('product.product').browse(cr,uid,product_id).procure_method
+                product = self.pool['product.product'].browse(cr, uid, product_id, context=context)
+                vals['type'] = product.procure_method
         return {'value': vals}
 
 
--- a/sale_dropshipping/sale.py
+++ b/sale_dropshipping/sale.py
@@ -60,6 +60,18 @@ class sale_order_line(orm.Model):
             result[order_line.id] = po_line_ids and po_line_ids[0] or False
         return result
 
+    def onchange_sale_flow(self, cr, uid, ids, sale_flow, product_id, context=None):
+        """ Change type to make_to_order when sale_flow is direct_delivery """
+
+        vals = {}
+        if product_id:
+            if sale_flow == 'direct_delivery':
+                vals['type'] = 'make_to_order'
+            else:
+                vals['type'] = self.pool.get('product.product').browse(cr,uid,product_id).procure_method
+        return {'value': vals}
+
+
     _columns = {
         'sale_flow': fields.selection([('normal', 'Normal'),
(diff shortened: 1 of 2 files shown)


sale_dropshipping/sale.py
```diff
@@ -60,6 +60,18 @@ def _purchase_order_line_id(self, cr, uid, ids, field_name, arg, context=None):
...
 
+    def onchange_sale_flow(self, cr, uid, ids, sale_flow, product_id, context=None):
+        """ Change type to make_to_order when sale_flow is direct_delivery """
+
+        vals = {}
+        if product_id:
+            if sale_flow == 'direct_delivery':
+                vals['type'] = 'make_to_order'
+            else:
+                vals['type'] = self.pool.get('product.product').browse(cr,uid,product_id).procure_method
```
pedrobaeza: Better put:

```
product = self.pool['product.product'].browse(cr, uid, product_id, context=context)
vals['type'] = product.procure_method
```

pedrobaeza:
Hi, is this problem already present on upstream version on Odoo 8.0?

tornoz:
Thank you for the code review, I will make the modifications in the branch.
I don't think this has been fixed for version 8.0

pedrobaeza:
Hi, thanks for the changes.

:+1: (code review)

Why don't you propose same patch for upstream Odoo 8.0?

Regards.

rvalyi:
:+1:  thanks!

eLBati:
I think this can be merged.

Side note: should patches like this also be proposed for 8.0?
As `sale_dropshipping` is not ported to 8.0 yet, probably they should not.
This is because, when someone migrates a module to 8.0, I think the first thing they should do is to copy the module from 7.0 and then to start migrating. So, these fixes would be included anyway

pedrobaeza:
Hi, @eLBati, I think on contrary, if everyone has to check full history of the module before migrate it, it's going to be a pain, so this kind of patches must be proposed inmediately to both branches, although the module is in `__unported__` directory. Making a cherrypick or applying a diff is very easy.

But this case is different, because this module is a backport of the v8 official drop shipping, so that's why I proposed to make the PR against upstream core.

Regards.

pedrobaeza:
I close it because GitHub have had an error doing the merge, although the merge has been done.

eLBati:
On 08/19/2014 07:23 PM, Pedro M. Baeza wrote:

> Hi, @eLBati https://github.com/eLBati, I think on contrary, if 
> everyone has to check full history of the module before migrate it, 
> it's going to be a pain, so this kind of patches must be proposed 
> inmediately to both branches, although the module is in _unported_ 
> directory. Making a cherrypick or applying a diff is very easy.

ok

PR #12, opened 2014-08-08
New module sale_order_authorized_users

LeartS:
Let Admin (or any user in the `base.group_erp_manager` group) choose, for each sale order,
which users will be able to access and see it. Any other user won't be able to see it.
If no users are set, the sale order has normal permissions.
The field to set allowed users will be visibile only to admin, which makes it
possible to make the users unaware of this feature.

Things that could be improved:
- Tests?
- create security rule also for `sale_order_line` if it's not automatic.

It's not 100% secure at the moment, but probably good enough for "normal" users.

### DIFF
--- a/sale_order_authorized_users/__openerp__.py
+++ b/sale_order_authorized_users/__openerp__.py
@@ -29,9 +29,9 @@
 Sale Order Authorized Users
 ======================================
 
-Let Admin (or any user in the base.group_erp_manager) choose, for each sale order,
-which users will be able to access and see it. Any other user won't be able
-to see it.
+Let Admin (or any user in the base.group_erp_manager) choose,
+for each sale order,w hich users will be able to access and see it.
+Any other user won't be able to see it.
 If no users are set, the sale order has normal permissions.
 The field to set allowed users will be visibile only to admin, which makes it
 possible to make the users unaware of this feature.
--- a/sale_order_authorized_users/__openerp__.py
+++ b/sale_order_authorized_users/__openerp__.py
@@ -30,13 +30,13 @@ Sale Order Authorized Users
 ======================================
 
 Let Admin (or any user in the base.group_erp_manager) choose,
-for each sale order,w hich users will be able to access and see it.
+for each sale order, which users will be able to access and see it.
 Any other user won't be able to see it.
 If no users are set, the sale order has normal permissions.
 The field to set allowed users will be visibile only to admin, which makes it
 possible to make the users unaware of this feature.
 """,
-    'author': 'Leonardo Donelli @ Creativi Quadrati',
+    'author': 'Creativi Quadrati',
     'depends': ['sale'],
     'data': [
         'sale_view.xml',
--- a/sale_order_authorized_users/__openerp__.py
+++ b/sale_order_authorized_users/__openerp__.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+# Author: Leonardo Donelli @ Creativi Quadrati
+# Copyright (C) 2014 Leonardo Donelli
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': 'Sale Order Authorized Users',
+    'version': '1.0',
+    'category': 'Sale',
+    'summary': 'Sale orders, Security, Permissions, Users',
+    'description': """
+Sale Order Authorized Users
+======================================
+
+Let Admin (or any user in the base.group_erp_manager) choose, for each sale order,
+which users will be able to access and see it. Any other user won't be able
+to see it.
+If no users are set, the sale order has normal permissions.
+The field to set allowed users will be visibile only to admin, which makes it
+possible to make the users unaware of this feature.
+""",
+    'author': 'Leonardo Donelli @ Creativi Quadrati',
+    'depends': ['sale'],
+    'data': [
+        'sale_view.xml',
+        'security/hide_sale_orders_security.xml',
+    ],
+    'installable': True,
+    'auto_install': False,
+}
--- a/sale_order_authorized_users/__openerp__.py
+++ b/sale_order_authorized_users/__openerp__.py
@@ -42,8 +42,8 @@ that it's not his, even if he's selected as allowed.
 On the other hand, by setting some users as allowed, all other
 users, even if they are "See all leads", won't be able to see the order.
 
-*When you add allowed users to a sale order you should include yourself, or you
-won't be able to access it anymore!*
+*When you add allowed users to a sale order, unless you're connected as admin,
+you must include yourself, or you won't be able make the change!*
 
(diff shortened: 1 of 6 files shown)


LeartS:
Build fails on main 7.0 branch and not because of my PR, branch status is green (when it should be red) because of https://github.com/OCA/maintainer-quality-tools/issues/60

pedrobaeza:
Can you rebase to see if the problem still remains?

LeartS:
@pedrobaeza the 7.0 branch is currently red (as the mentioned issue has been fixed), if I rebase now won't it fail just because the main branch is failing?

pedrobaeza:
Yeah, I see. I thought it was fixed already. Let's wait then that the upstream problem is fixed to do the rebase.

Thanks.

sale_order_authorized_users/sale.py
```diff
@@ -0,0 +1,35 @@
...
+
+class sale_order(orm.Model):
+    _inherit = 'sale.order'
+
+    _columns = {
+        'allowed_users_ids': fields.many2many(
+            'res.users',
+            'sale_order_res_users_rel',
+            'sale_order_id',
+            'user_id',
```
guewen: Expliciting the 2 relation fields and name of the relation table is useless, they will be generated by the ORM.

sale_order_authorized_users/sale.py
```diff
@@ -0,0 +1,35 @@
...
+            'Allowed Users',
```
guewen: Using the named parameter is cleaner, so it ends up to:

```
    'allowed_users_ids': fields.many2many(
        'res.users',
        string='Allowed Users',
        groups='base.group_erp_manager'
    ),
```

sale_order_authorized_users/security/hide_sale_orders_security.xml
```diff
@@ -0,0 +1,13 @@
...
+
+		<record model="ir.rule" id="rule_hidden_orders">
+			<field name="name">Hidden orders</field>
+			<field name="model_id" ref="model_sale_order" />
+			<field name="global" eval="True" />
+			<field name="domain_force">['|',('allowed_users_ids','in',user.id),('allowed_users_ids','=',False)]</field>
+		</record>
+
+	</data>
+</openerp>
```
guewen, line 27: Could use spaces (preferably 2 for xml files) instead of tabs?

guewen:
Apart the few line comments, seems good to me

guewen:
Thanks

The build is red, but this is because of https://github.com/OCA/OCB/issues/37

So :+1:

sale_order_authorized_users/sale.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class sale_order(orm.Model):
+    _inherit = 'sale.order'
+
+    _columns = {
+        'allowed_users_ids': fields.many2many(
+            'res.users',
+            string='Allowed Users',
+            groups='base.group_erp_manager',
```
lepistone: can you add a help string saying that if left empty, normal rules apply?

sale_order_authorized_users/security/hide_sale_orders_security.xml
```diff
@@ -0,0 +1,13 @@
...
+<openerp>
+  <data noupdate="0">
+
+    <record model="ir.rule" id="rule_hidden_orders">
+      <field name="name">Hidden orders</field>
+      <field name="model_id" ref="model_sale_order" />
+      <field name="global" eval="True" />
+      <field name="domain_force">['|',('allowed_users_ids','in',user.id),('allowed_users_ids','=',False)]</field>
+    </record>
+
```
lepistone, line 18: wouldn't it be straightforward to add a similar rule for lines just in case? like order_id.allowed_users_ids in user.id

sale_order_authorized_users/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+    'name': 'Sale Order Authorized Users',
+    'version': '1.0',
+    'category': 'Sale',
+    'summary': 'Sale orders, Security, Permissions, Users',
+    'description': """
+Sale Order Authorized Users
+======================================
+
+Let Admin (or any user in the base.group_erp_manager) choose,
+for each sale order,w hich users will be able to access and see it.
```
lepistone: typo

lepistone:
Not blocking for that, but tests on user rights can be done, even though there are not many around. I'm happy to approve that once my minor remarks are addressed

LeartS:
Thanks @lepistone, I updated the PR according to your suggestions and also created a simple YAML test. (my first, don't eat me if there are some errors!)

lepistone:
excellent @LeartS!

in fact, now you could add another case that tries to create an order with no authorized users.

LeartS:
@eLBati, why the needs fixing tag? Because I've not implemented the last test case suggested by @lepistone ?

eLBati:
Hello @LeartS. Yes, because of the missing test case

pedrobaeza:
One functional question: don't you think it's better to let any sale manager to decide authorized users instead the admin (usually a tech profile)?

LeartS:
@pedrobaeza I made this module because one of my customers asked me for a way to create "secret orders", only visible to people he specifically chooses, something separate and that takes precedence over the normal sale permissions schema. That's why I used `base.erp_manager` (admin).

If I used sale managers, all of them would see the `authorized_users` field and the module would be useless to hide a sale order from everyone or to some of them, for example. The idea is that nobody should even know there is an additional system to hide sale orders in place.

What I could do though, is create a new group (like `sale.secret_orders`) so that the admin can put in there users he want to be able to see and edit the `the authorized_users` field.

nhomar:
@LeartS I am agreed you should create a new group as you propose, that's the correct way to achieve that, base.erp_manager is a group for technical purpose.

gurneyalex:
The code looks fine. 

I'm not sure there is enough need for the functionality to warrant inclusion in OCA/sale-workflow. This sounds very customer specific to me. 

Edit: :+0:

LeartS:
Thanks for the review. If you guys think it's not a requested enough feature, I will put it up on my GitHub account, no problem.

pedrobaeza:
With the change I proposed, I think it's a nice to have feature, because it's very usual than inside a company you need to restrict sale orders/quotations to certain users. Granularity offered by OpenERP is not enough in these cases: your orders, sales team orders, or all.

LeartS:
@pedrobaeza I alread added the specific group, did you see?

Should I make that users in the secret orders group are automatically users of sale manager group?

lepistone:
that sounds like a good idea @LeartS . I approve the code, no opinion on it being too specific or not. Let's see if @gurneyalex is convinced by @pedrobaeza :smile_cat:

LeartS:
Done. I'm not sure if it's ok to imply only sales manager or I should also include all the other sales groups.

pedrobaeza:
Thanks, for me it's :+1: .

@gurneyalex, features offered by this module is enough to justify it, so please reconsider you position.

Regards.

dcorio:
:+1: 

i would really like to see this applied (this way or another) to all the models, by default as a core feature.

LeartS:
I've changed the description in the manifest file to reflect the usage of the new group instead of `base.erp_manager`. I've also clarified (I hope) that the usual rules still apply, setting an user as allowed **doesn't** necessarily mean he will be able to see the order.

One "inconvenience" of using the custom made group is that if an user in the Secret orders group adds allowed users to a sale order, he must also explicitly add himself or he won't be able to make the change. I've also added this gotcha in the description.

LeartS:
So, any chance of getting out of the limbo? :cloud:

lepistone:
hurm, @gurneyalex is against. I think you can close this if you're still of that idea Alexandre.

gurneyalex:
ok... I'll abstain... please disregard my previous vote and merge.

jgrandguillaume:
Hi,

Thanks for the contrib. I tempted to be inline with @gurneyalex here, may be it's a bit too specific. But as you manage to take into account all remarks and your module has a test, I think you deserve to have this module here. So I'm :+1: 

Regards,

Joël

PR #14, opened 2014-08-11
Port 7.0 add last sale price yvr

yvaucher:


pedrobaeza:
There is an error with this module tests. Can you check it, please?

Regards.

yvaucher:
last_sale_price tests fail because of sale_stock:

IntegrityError: null value in column "picking_policy" violates not-null constraint

This module doesn't depends on sale_stock, thus it should be tested separately

gurneyalex:
@yvaucher I think you can work around this by changing the travis configuration for the repository, in order to make separate tests for the modules which must not be tested with sale_stock installed.

yvaucher:
Separating last_sale_price test works though it seems there is still an error with sale_validity on other test. (this shouldn't have anything to do with last_sale_price)

eLBati:
The test error came from https://github.com/OCA/sale-workflow/pull/22
@yvaucher you can try rebasing

yvaucher:
Rebased

eLBati:
The remaining failure comes from https://github.com/OCA/OCB/issues/37
So, :+1:

gurneyalex:
rebased on top of 7.0

gurneyalex:
:+1:

pedrobaeza:
:+1:

Issue #13, opened 2014-08-08
sale_validity gives error: Can't render view sale_validity.view_order_form_validity for model: sale.order


LeartS:
The module `sale_validity` fails on startup, as you can see from [this Travis report](https://travis-ci.org/OCA/sale-workflow/jobs/32009263).
The status of the branch is green because of issue https://github.com/OCA/maintainer-quality-tools/issues/60, but it makes every pull request fail.

The relevant line of error seems to be:

```
AttributeError: View definition error for inherited view 'sale_stock.view_order_form_inherit' on model 'sale.order': Element '<xpath expr="//button[@name='action_cancel']">' not found in parent view 'sale.view_order_form'
```

Which I really don't understand given sale_validity's [sale_order.xml code](https://github.com/OCA/sale-workflow/blob/7.0/sale_validity/view/sale_order.xml)

```
2014-08-08 12:56:55,416 23157 ERROR openerp_test openerp.addons.base.ir.ir_ui_view: Can't render view sale_validity.view_order_form_validity for model: sale.order
Traceback (most recent call last):
  File "/home/travis/build/OCA/odoo-7.0/openerp/addons/base/ir/ir_ui_view.py", line 126, in _check_render_view
    fvg = self.pool.get(view.model).fields_view_get(cr, uid, view_id=view.id, view_type=view.type, context=context)
  File "/home/travis/build/OCA/odoo-7.0/openerp/osv/orm.py", line 2263, in fields_view_get
    arch=apply_view_inheritance(cr, user, source, sql_res['id']),
  File "/home/travis/build/OCA/odoo-7.0/openerp/osv/orm.py", line 2221, in apply_view_inheritance
    source = apply_inheritance_specs(source, view_arch, view_id)
  File "/home/travis/build/OCA/odoo-7.0/openerp/osv/orm.py", line 2205, in apply_inheritance_specs
    raise_view_error("Element '%s' not found in parent view '%%(parent_xml_id)s'" % tag, inherit_id)
  File "/home/travis/build/OCA/odoo-7.0/openerp/osv/orm.py", line 2098, in raise_view_error
    %  (child_view.xml_id, self._name, error_msg))
AttributeError: View definition error for inherited view 'sale_stock.view_order_form_inherit' on model 'sale.order': Element '<xpath expr="//button[@name='action_cancel']">' not found in parent view 'sale.view_order_form'
[...]

PR #21, opened 2014-08-15
fix installation of sale_cancel_reason to be compatible with sale_stock

yvaucher:
Trying to fix install issue to restore tests

https://github.com/OCA/sale-workflow/issues/13

### DIFF
--- a/sale_cancel_reason/view/sale_view.xml
+++ b/sale_cancel_reason/view/sale_view.xml
@@ -6,11 +6,11 @@
             <field name="model">sale.order</field>
             <field name="inherit_id" ref="sale.view_order_form"/>
             <field name="arch" type="xml">
-                <xpath expr="/form/header/button[@name='action_cancel']" position="attributes">
+                <xpath expr="//button[@name='action_cancel']" position="attributes">
                     <attribute name="type">action</attribute>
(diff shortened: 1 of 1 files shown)


guewen:
My guess:

As soon as `sale_stock` is installed (installed as dependency for another module), it fails to install because it contains:

```
                    <xpath expr="//button[@name='action_cancel']" position="after">
                        <button name="ship_cancel" states="shipping_except" string="Cancel Order"/>
                    </xpath>

```

And the button can't be found with that name.

guewen:
An idea: instead of changing the name of the button, hide it and create a new one for the cancellation reason. (but it may be confusive if other moduls want to modify other attributes on the button such as the `states`!)

yvaucher:
You are right sale_stock fails because this module change the name.

I worked previously because this module was inheriting from sale_stock.

yvaucher:
at least sale_validity now installs

Other errors can be fixed in other PR

guewen:
Thanks!
The build fails but for another module... 

:+1:

rvalyi:
:+1:  LGTM

pedrobaeza:
:+1: As stated, this PR solves the problem with sale_validity.

Let's merge it and make another PR for the following problem.

PR #25, opened 2014-08-22
Merge PO with sale_dropshipping

rdeheele:
Updates on sale_dropshipping to :
- do not merge orders with different destination addresses
- populate the destination address when merging two purchase orders

Original Launchpad MP : https://code.launchpad.net/~camptocamp/sale-wkfl/7.0-dropshipping-merge-po-lep/+merge/216756


sale_dropshipping/__openerp__.py
```diff
@@ -25,6 +25,7 @@
  "website": "http://www.openerp.com",
  "category": "Generic Modules/Purchase",
  "depends": ["purchase",
+             "purchase_group_hooks",
```
guewen, line 28: Where is this module? Could you add a dependency in `.travis.yml`

sale_dropshipping/purchase.py
```diff
@@ -110,3 +110,14 @@ def action_picking_create(self, cr, uid, ids, context=None):
...
+
+    def _key_fields_for_grouping(self):
+        """Do not merge orders that have different destination addresses."""
+        field_list = super(purchase_order, self)._key_fields_for_grouping()
+        return field_list + ('dest_address_id', )
+
+    def _initial_merged_order_data(self, order):
+        """Populate the destination address in the merged order."""
+        res = super(purchase_order, self)._initial_merged_order_data(order)
+        res['dest_address_id'] = order.dest_address_id and order.dest_address_id.id or False
```
guewen, line 130: The ternary form `... and ... or ...` should be avoided, prefer `... if ... else ...`.

pedrobaeza, line 130: I don't see special problems with that form, but in this case is simpler. You can simply put `res['dest_address_id'] = order.dest_address_id.id`, because it will return False in case of browse_null.

guewen, line 130: @pedrobaeza is right, no ternary operator is necessary here

On the form `... and ... or ...` and why it should be avoided: http://tian-yi.me/why-use-ternary-operator-in-python.html

pedrobaeza, line 130: Thanks for the information. Indeed, in other cases, this can be a problem (not for this, because else variable is also False), but it's good to have it as a norm.

guewen, line 130: Yep not a problem here but a good habit to be used to.

rdeheele:
thanks for your remarks, useful explanation about ternary operator, I transfer the PR in #27 (git first steps are a little confused)

PR #28, opened 2014-08-25
8.0 port sale_validity

yvaucher:


gurneyalex:
I'd love to see a small yaml test exercising this addon :imp:

yvaucher:
@gurneyalex Tests are there

lmignon:
it's nice to see tests for a module but IMO yaml is really not the right way to write tests since it's impossible to debug or share logic between tests... just my opinion...

gurneyalex:
@lmignon this is a separate debate. What I'm concerned with is the abysmal number of addons with exactly 0 test. Any test is better than 0.

gurneyalex:
:+1:

lmignon:
@gurneyalex I agree any test is better than 0 and Unittests vs yaml is a separate debat. But if we ask to add tests, it would be safe to have a consensus from OCA on the expected  kind of tests. I'm also concerned by the number of addons without tests since I personally try to always write the tests for my PR.

@yvaucher :+1:

gurneyalex:
@lmignon my personal take on this is that to start providing coverage, yaml tests are a quick win. Especially the yaml import code takes care of calling the `on_change` functions when a new object is created, and this is priceless. I agree that debugging yaml tests is not nice, but then these tests are supposed to be green anyway :smile:. Until we start doing proper TDD, I think that yaml tests are "good enough"  as a first step to get at least some coverage of the module. 

You're right in that I should not have asked explicitely for a _yaml_ test, and let yannick choose. I already feel a bit bad asking people who are doing the grunt work of porting legacy addons to 8.0 to add some tests in the process (big thanks @yvaucher for this by the way), let the porters choose the kind of test they want. 

Side note: we will probably be adding more ways of running tests to the Travis config. Right now only the standard Odoo tests are supported, but in the future we could have normal Python unittests, and OERPScenario tests.

lmignon:
@gurneyalex thank you for your constructive feedback :smile:

PR #31, opened 2014-08-26
[IMP] add multi-company support in module sale_exceptions

nbessi:


sale_exceptions/sale.py
```diff
@@ -39,22 +39,40 @@ class sale_exception(orm.Model):
...
+            readonly=True
+        ),
+
+        'company_id': fields.many2one(
+            'res.company',
+            'Company',
+        )
     }
 
+    def _default_company(self, cr, uid, ids, context=None):
```
pedrobaeza: Signature is not correct. Here ids are not passed.

nbessi:
@pedrobaeza yes good catch

nbessi:
Fix pushed

nbessi:
Hmm There is an exception on OCB only with sale_exception_no_stock

pedrobaeza:
Yeah, this is something that it's pending to check, but apart from this PR. Maybe a regression?

pedrobaeza:
For this, :+1:

gurneyalex:
this really deserves at least an automated test.

lepistone:
:+1:

Answering @gurneyalex: it looks like this PR only adds the `company_id` with a security rule. Sure the `sale_exceptions` module would need a few tests, but this PR probably is fine as it is. What do you think?

gurneyalex:
@nbessi why is company_id computed from the uid and not from the sale.order ? I would have thought that the company_id of a sale exception would be a related field to the company if the sale order. 

Otherwise you could have a sale order validated by a user from a parent company producing an exception not visible by users from children companies. 

Or am I missing something?

gurneyalex:
Ok, I did miss something. The model is called `sale.exception` but it is actually a sale exception rule :-/

So :+1:

PR #32, opened 2014-08-27
8.0 port sale_exceptions

yvaucher:


pedrobaeza:
Please make `git mv` on _sale.py_ and _sale_exception_confirm.py_ to see changes.

Regards.

yvaucher:
@pedrobaeza I did it in rev 2cdfbfb. However, as most of the code changed it cannot recognize the file on this merge

You can still see the diff in 339ff4d

pedrobaeza:
Some flake8:

```
./sale_exceptions/sale.py:65:8: E121 continuation line under-indented for hanging indent
./sale_exceptions/sale.py:77:5: E303 too many blank lines (2)
./sale_exceptions/sale.py:146:17: E126 continuation line over-indented for hanging indent
```

sale_exceptions/sale.py
```diff
@@ -0,0 +1,217 @@
...
+        return True
+
+    @api.one
+    def _popup_exceptions(self):
+        model_data_obj = self.env['ir.model.data']
+        list_obj = self.env['sale.exception.confirm']
+        ctx = self._context.copy()
+        ctx.update({'active_id': self.ids[0],
+                    'active_ids': self.ids})
+        list_id = list_obj.with_context(ctx).create({})
```
yvaucher: To fix with_context extends current context. No need to copy context

pedrobaeza: As context is now a frozen dict, I think it's needed, isn't it?

eLBati:
Here http://pastie.org/9526958 is the diff between 7.0 and 8.0 if it may help the review

eLBati:
:+1:

rvalyi:
:+1:  LGTM thanks!

yvaucher:
ValueError: "'sale.order.line' object has no attribute '_rule_eval'" while evaluating

rvalyi:
I'm not sure how this happened, but I want to mention that I did my review after the commit https://github.com/yvaucher/sale-workflow/commit/5b5d0a0b0045d1133b3f96bcc203800e9e1368b6 unlike this  appears now in the timeline. Pretty sure this is the case of @eLBati  too given the diff he posted. At least I was using your branch on your repo to do so. Anyway waiting for new need review label before testing again.

yvaucher:
@rvalyi It was working fine with sale.order exception but not with sale order lines one.
This happened because I was using self instead of using an object as attribute in some functions.
This should be fixed now.

I did a rebase to clean history.

sale_exceptions/sale.py
```diff
@@ -0,0 +1,213 @@
...
+        string='Sequence',
+        help="Gives the sequence order when applying the test")
+    model = fields.Selection(
+        [('sale.order', 'Sale Order'),
+         ('sale.order.line', 'Sale Order Line')],
+        string='Apply on', required=True)
+    active = fields.Boolean('Active')
+    code = fields.Text(
+        'Python Code',
+        help="Python code executed to check if the exception apply or "
```
gurneyalex, line 48: s/apply/applies/

sale_exceptions/sale.py
```diff
@@ -0,0 +1,213 @@
...
+            [('model', '=', 'sale.order')])
+        line_exceptions = exception_obj.search(
+            [('model', '=', 'sale.order.line')])
+
+        for order in self:
+            if order.ignore_exceptions:
+                continue
+            order._detect_exceptions(order_exceptions,
+                                     line_exceptions)
+        return self.exception_ids
```
gurneyalex, line 154: I don't like this side effect of `_detect_exception`. The previous version had the side effect one level up (in `detect_exceptions`) which was not a very good design, especially with such a method name which tends to indicate "finding something"  rather than "changing state". 

Additionally, returning a column in an `@api.multi` method will cause an error if there are more than 1 record in the recordset, which is not what we want here. Since the return value here AFAIK  is only used as a truth value, a boolean should be returned, computed from a return value of `_detect_exception` (to be added back).  If we wish to keep the previous signature, then the list of exception_ids must be computed by aggregating ids in the `for` loop. 

A docstring mentionning the side effect would be nice too (I understand if we wish to keep the function name as it is a public API of the module).

sale_exceptions/sale.py
```diff
@@ -0,0 +1,213 @@
...
+        for order_line in self.order_line:
+            for rule in line_exceptions:
+                if rule.id in exception_ids:
+                    # we do not matter if the exception as already been
+                    # found for an order line of this order
+                    continue
+                if self._rule_eval(rule, 'line', order_line):
+                    exception_ids.append(rule.id)
+
+        self.exception_ids = [(6, 0, exception_ids)]
```
gurneyalex, line 204: add a return value (see comment above)

gurneyalex:
And now I realize I missed this yesterday when I merged the PR :sweat: I guess I'll make another PR to fix this...

gurneyalex:
cf. #38

PR #35, opened 2014-09-04
[FIX] module 'sale_line_description'

tafaRU:
This PR contains the following changes about the module **sale_line_description**:
- replaced **description** by **description_sale** because the latter is that used in the sale order;
- got the change of the line sale description flag-aware

### DIFF
--- a/sale_line_description/__openerp__.py
+++ b/sale_line_description/__openerp__.py
@@ -21,12 +21,13 @@
 #
 {
     'name': "Sale line description",
-    'version': '0.1',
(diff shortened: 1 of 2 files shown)


tafaRU:
OCB tests are red because of OCA/OCB#37

guewen:
:+1:

lepistone:
As you say @tafaRU, the OCB in red because of something unrelated, and the odoo one is good. Thanks! :+1:

eLBati:
:+1:

PR #36, opened 2014-09-08
sale_validity: backport 8.0 tests to 7.0

gurneyalex:


pedrobaeza:
There is something different in v7, because test is failing:

```
2014-09-08 10:18:16,757 23281 ERROR openerp_test openerp.tools.yaml_import: "Field '_onchange_spec' does not exist in object 'browse_record(sale.order, 10)'"
```

gurneyalex:
Travis is red because of https://github.com/OCA/OCB/issues/37

pedrobaeza:
OK, the previous error has gone (I don't know why - I suppose because of maintainer quality tools change..).

:+1:

gurneyalex:
@pedrobaeza it's gone because I used `git commit --amend` and `git push -f` so that my various failed attempt would not clutter the commit tree :smile_cat:

pedrobaeza:
Ah, that's the trick :wink:
Now, let's wait if we can merge soon the patch for OCB tests.

eLBati:
:+1:

PR #37, opened 2014-09-08
8.0 port sale_exception_nostock

yvaucher:


lepistone:
thanks :+1:

eLBati:
Here is the diff between 7.0 and 8.0 if it may help http://pastie.org/9548060

eLBati:
:+1:

gurneyalex:
caution with `api.one`: it wraps the result in a list. For predicate methods, you're generally better off using `api.multi` + self.ensure_one() in the body to raise if more than 1 id was passed. 

Otherwise the caller may receive `[False]` as return value, which is True-ish.

gurneyalex:
:+1:

PR #38, opened 2014-09-09
fix error when >1 draft SO are checked

gurneyalex:
when the wizard would run, then the @api.multi method `detect_exceptions` would
fail because it returns an attribute but it is called on a recordset of size >1

Code refactoring to:
- move the side effect of writing to so.exception_ids one method up
- document the side effect in the docstring
- fix the computation of the return value for the `detect_exceptions` method

### DIFF
--- a/sale_exceptions/sale.py
+++ b/sale_exceptions/sale.py
@@ -122,8 +122,9 @@ class SaleOrder(models.Model):
         }
         return action
 
-    @api.one
+    @api.multi
     def action_button_confirm(self):
+        self.ensure_one()
         if self.detect_exceptions():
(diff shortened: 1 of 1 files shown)


gurneyalex:
see https://github.com/OCA/sale-workflow/pull/32/files#diff-6d72f9ca7356d5c4701bfe428c03a68eR141 for a rational of the change

eLBati:
Just one PEP8 remark

```
./sale_exceptions/sale.py:214:5: E303 too many blank lines (2)
```

Apart from this, :+1:

guewen:
:+1:

PR #40, opened 2014-09-10
Fix the tests on OCB branch (closes #37)

gurneyalex:
The [test failures](https://github.com/OCA/OCB/issues/37) of sale_exception_nostock on OCB branch are caused by
https://github.com/OCA/OCB/commit/5d795c48fb394505d5a6bbf77961d63da05515c4
which changed the date considered for the stock moves from `date_order` to
`date_confirmed`.

In order to restore the test behavior, we mock `datetime.now`
which is ultimately used in `sale_order.action_wait` to write `date_confirm`.
The current date is restored after each confirmation, in order to avoid
impacting other aspects of the tests (if any).

This does not impact the test on the official branch and gives the correct
results on OCB.

### DIFF
--- a/sale_exception_nostock/test/no_stock_test.yml
+++ b/sale_exception_nostock/test/no_stock_test.yml
@@ -13,6 +13,27 @@
     data['active'] = False
     self.write(cr, uid, ids_to_silent, data)
 -
+  I define a helper function to tweak the system date
+-
+  !python {model: sale.order}: |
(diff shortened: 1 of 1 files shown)


guewen:
good catch :open_mouth:

:+1:

pedrobaeza:
:+1: 

Very good!

michaelkarrer81:
+1

PR #41, opened 2014-09-12
8.0 new addons: sale_quotation_sourcing

gurneyalex:
`sale_quotation_sourcing` allows to invert the purchase sourcing logic in Odoo. In this mode, an RFQ is sent to to suppliers _before_ the SO is confirmed (in order to get the best quote, and to ensure that the price communicated to the customer is computed based on that quote). When the SO is confirmed, a link can be made to the RFQ to use for sourcing the sale, and the procurement order will not create a new PO.


eLBati:
Hello @gurneyalex OCB are red because of

```
2014-09-12 13:48:26,444 23272 INFO openerp_test openerp.modules.loading: loading sale_quotation_sourcing/tests/sourcing.yml
2014-09-12 13:48:27,599 23272 ERROR openerp_test openerp.tools.yaml_import: AssertionError in Python code (line 43): retval: {'view_mode': 'form', 'res_id': 9, 'name': 'Sales Order', 'nodestroy': True, 'view_type': 'form', 'res_model': 'sale.order', 'view_id': (699,), 'type': 'ir.actions.act_window', 'target': 'current'}
```

gurneyalex:
@eLBati yes... I'm pretty sure this is because the OCB 8.0 branch is 10 days behind the official branch.

pedrobaeza:
I'm pushing manually changes to OCB 8.0 for now and restart Travis build. Let's wait for a green! :wink:

jgrandguillaume:
Hi,
Thanks for this work ! I like the wizard and the way to implement the feature. Some remarks I have while testing :
1. ~~In **openerp**.py, the name of the module should be "Sale Quotation Sourcing"~~
2. ~~I think you should overide the "copy()" method of SaleOrderLine in order to empty the "sourced_by" field. We always want to chose from which PO we want to source from. If we copy that info, we do not have the opportunity to chose the PO (it'll take the same one than the copied SO, which is probably not wanted).~~
3. ~~Sourcing a line with another product should not be allowed. We should filter the product of the PO line according to the product of the SO line~~
4. When we cancelled a PO, I think it should cancel the related procurement. This sounds like a generic bug to me (not linked to your module). This worked in v 7.0 and doesn't seems to work in v 8.0 standard.
5. ~~When I manually source a SO Line, I have no delivery generated :~~
   1) ~~Create a PO with ipad mini and confirm~~
   2) ~~Create SO with one line ipad mini, manually sourced by PO line above~~
   3) ~~Confirm the SO, no delivery is generated~~
   - => ~~Here we should generate the delivery for the SO. The PO reception make "Supplier -> Stock" the SO Delivery should make the "Stock -> Customer" move. Otherwise it doesn't work.~~
   - => ~~If the route "drop shipping" was selected, then the behavior is correct => only one PO Reception should be generated (and no SO delivery).~~
6. ~~The sale_quotation_sourcing module should enable the "Choose MTO, drop shipping,... on sales order lines " in Sale configuration as well as enabling in the route called "Make to order" the checkbox: Applicable on "Sale Order Line".~~
7. Sourcing line manually should automatically set the route "Make to Order" on the SO Line (we cannot chose a PO line on stock, it make no sense I think). We should be able to change this to use the "Drop shipping " route instead. 

Well, this is all I figure it out when testing it. A part from that, it's exactly what we needed.

Thanks,

Joël

lepistone:
I'm taking care of @jgrandguillaume's remarks (WIP)

lepistone:
Hi @jgrandguillaume, after your remarks this module got some love. In you were right with your point 5, and in fact the process had to be changed. For a MTO, buy order in standard odoo 8, two procurements are generated, and it is the second one that is attached to the purchase. To reproduce that logic with manual sourcing, the logic had to be changed to work on the second procurement only.

We now have two test cases, where the first one reproduces the standard odoo MTO, buy logic. The purpose is to have two similar tests where the second is manually sourced, and check that we get the same result in the end. We're green!

As for your point 7, it is not as easy as that in v8, because you do not choose the MTO route yourself in the sale line. You enable MTO and buy on the product, and on the sale line either you leave empty or you choose dropshipping. I tried to enable MTO to be chosen in the order line, but that doesn't work due to the logic that odoo used to generate the second procurement (the one that handles the purchase).

Thanks!

gurneyalex:
:+1: thanks @lepistone for this wonderful work.

sale_quotation_sourcing/__openerp__.py
```diff
@@ -0,0 +1,59 @@
...
+#
+{
+    'name': "Sale Quotation Sourcing",
+
+    'summary': "manual sourcing of sale quotations",
+    'description': """
+    This module implements manual sourcing of sale order lines from purchase
+    order lines.
+
+    Instead of having a the confirmation of a SO generate procurements which in
```
nbessi: Something odd in this sentence

lepistone:
Improved the doc as suggested by @nbessi

sale_quotation_sourcing/test/setup_product.yml
```diff
@@ -0,0 +1,10 @@
+-
+  I add the routes "Buy" and "MTO" to my product
+-
+  !python {model: product.template}: |
```
nbessi: Why not use the id idiom ?
http://odoo-new-api-guide-line.readthedocs.org/en/latest/test.html#yaml

sale_quotation_sourcing/__openerp__.py
```diff
@@ -0,0 +1,59 @@
...
+    automatically generated. The end result should be the same.
+
+    To show that, two test cases are provided that show the standard process
+    and the manually sourced one.""",
+
+    'author': "Camptocamp",
+    'website': "http://www.camptocamp.com",
+
+    'category': 'Sales',
+    'version': '0.1',
```
nbessi, line 58: I will add nose in python deps or a comment in the description

nbessi:
Just notice small details but nothing blocking :+1:

lepistone:
I addressed @nbessi's remarks, but there is now a conflict with another module. I will check that tomorrow (for that reason "needs fixing")

yvaucher:
There are 2 issues on the wizard to record a sourcing:
1. on po_line m2o the create is available. Thus will throw an error as you can't create a po line without PO and the view doesn't find parent PO. One option would be to set widget = selection but this prevent to open the object.
   We could use https://github.com/OCA/web/tree/8.0/__unported__/web_m2x_options if still necessary in 8.0
2. nothing saves you from selecting a PO and selecting a PO line from another PO if you select first the PO line and then the PO. Some onchanges must be defined or set read only po line till PO is selected.

Appart from those usability details, works like a charm.

jgrandguillaume:
Hi,

After testing it, I created a pull request here: https://github.com/gurneyalex/sale-workflow/pull/1

That avoid the case 2 raised by Yannick and as well simplify the selection process.

I still face an issue, doing this:
 1) Create a PO for ipadmini
 2) Create a SO with :
      - 1x ipad mini, drop shipping, source manually on first created PO
      - 1x imac, no route, buy with order point created
 3) Confirm the SO
 4) Confirm the PO

Here, you have 2 deliveries, one for the PO and one for the SO, both linked to the SO (when you click on "View delivery order" button). The thing is the delivery generated for the PO is "Supplier -> Stock" ! We want to have "Supplier -> Customer" here as it was drop shipped.

The problem comes from the fact the PO was already generated before, with a destination set as the warehouse. I know the Buyer should have set up properly the destination, but we should at least warn the user about it.

I suggest:
- When selecting the PO line, we should check if the route "Drop shipping" is selected. If yes, then we should check that the PO destination location (in second tab) is of type "customer". If not, we should warn the user that he selected a drop shipping based on a PO that is going to be delivered in the warehouse "XY" instead of a customer location.

Regards,

Joël

lepistone:
Thanks @jgrandguillaume @yvaucher , I'll process that later today

lepistone:
@yvaucher your remarks are addressed.
1. thanks to `options="{'no_create': True}"`
2. thanks to @jgrandguillaume's contribution

yvaucher:
@lepistone thanks! :+1:

lepistone:
@jgrandguillaume for the drop shipping case, I thought that the "drop shipping" route is just configuration data that is created by the stock_dropshipping module. Moreover, the logic for choosing routes and rules is complex, since those could be specified on the SO line, product or category and configured in many different ways, with or without stock_dropshipping.

For that reason I did a variant of what you suggest: when the SO is confirmed, and the procurement is processed, I check that the destination location of the procurement (whatever way it is computed) is the same as the destination location of the chosen PO. If not, I block the validation with a long message.

It would probably be more user-friendly to check immediately (maybe with an onchange or domain) but I think at that point I don't have enough information for that.

I added two new test cases: one for the standard dropshipping case, and one for dropshipping + manual sourcing.

sale_quotation_sourcing/__openerp__.py
lepistone, line 58: @nbessi done

erwintwr2:
Hi guys

trying to see if your module can perform the following ( since i assume this was the original intention of the module - please correct me if i am wrong) :

*****_PROCESS START**_******
- customer request comes in (various items from different suppliers)
- we require to update our pricing on all items, thus a RFQ is started (actually more correct, a Calls  
  for   bid, as this allows me to enter the exact list or products, and let the system create RFQ;s for the correct suppliers matching the products
- Once the call for bid is complete - eg pricing is updated, a Sales Quote is to be generated (i am still unsure how the pricing update is to be handled except by manually going through each approved RFQ, and editing the public price for each product?? - can be very time consuming on large quotes)
- considering that all the products/ Lines is already added on the call for bids, the next step would have been to simply convert the call for bids to a sales order (selecting customer and applicable price sale price list so that products can be quoted using latest price and mark-up)
- Sales quote is sent to customer
- customer accepts Sales quote
- Call for bid is resumed, creating orders for the various suppliers

_after this the usual deliver/ invoice process continues
*_***_PROCESS END**_******

Currently my experience of the module is that it is still very manual-  eg i have to create a bid for all the products, make a list of RFQ's for each product,update product pricing, and then create the sales order lines with manual sourcing and linking to the relevant RFQ.

Considering that multiple quotes can be awaiting customer order at a time, i further see a problem with the huge list of RFQ's that will be sitting in the que waiting for their respective sales orders. this could confuse the stock control/warehouse should they need to procure actual products.

Am i maybe overlooking some functionality?

thank you

Erwin

gurneyalex:
@erwintwr2 this module serves as a basis for the process you describe. It works together with other addons in OCA/purchase-workflow and OCA/vertical-ngo. Note that there will be a presentation of these addons during the community days. 

Could we move the discussion to the community mailing list rather than on a PR?

erwintwr2:
Hi Alex

Thank you for the feedback. I am most definitely interested in receiving updates on this. Can you advise which mailing list i can join?

regards

gurneyalex:
@erwintwr2 community@lists.odoo.com is the best place for this.

dzenkins:
Hey, is this functionality still available in an add-on for Odoo 15?

gurneyalex:
@dzenkins wow you are unearthing old things... To my knowledge this is not supported by any addon or standard in v15 so you would need to "migrate" this module to the version. I put quotes because things have changed a lot in the meantime and probably this will result in a significant rewrite.

PR #46, opened 2014-10-04
7.0 sale tax inc exc

sebastienbeau:
This module give the posibility to sale with ou without taxe, depending of the business type (B2B or B2C).
The tax computation is simple and may not fit with every country but fit perfectly for French case and may fit with all Europe case.
You need to configure on every tax the linked "included taxe version" and then want you create a sale order depending of the option the price will be included or not in the tax.
Now you can make beautiful rounded price for B2B and B2C ;)


sebastienbeau:
I close my PR, indeed after a long discution with Alexis, we found a better datamodel, so a cleaner solution will be propose ;)

alexis-via:
The new solution is the module sale_tax_included_excluded, available here: https://github.com/akretion/sale-workflow/tree/8.0-sale-tax-included-excluded/sale_tax_included_excluded

I plan to do a PR on OCA/sale-workflow soon.

PR #47, opened 2014-10-10
extend sale_exception_nostock to by_line case

lepistone:
In v7, an extra module allowed to select the source location in every
sale order line. In v8 standard, you can select an optional Route in
every line and get the same result.

For that reason, instead of porting the old
`sale_exception_nostock_by_line` to v8, I extended the
`sale_exception_nostock` module so that it takes care of the Route
selected if the order line if there is one.


jgrandguillaume:
Hi Leonardo,

I've tested it and it work perfectly. Code review ok for me, so :+1: 

Good idea to merge the 2 modules taking the new route system.

Thanks !

Joël

jgrandguillaume:
Hi,

Just faced an issue:

1) Create a SO
2) Chose one product and route "Drop Shipping"
3) Confirm SO

You have an error raised "Not enough stock at delivery date". We should not encounter this if drop ship, we'll never have stocks...

Regards,

Joël

lepistone:
@jgrandguillaume, now if the source location is not internal, we don't check. That should cover the dropshipping case.

yvaucher:
checked with last commit about warehouse compatibility :+1:

jgrandguillaume:
Great work ! It's working like a charm now ! :+1:

PR #48, opened 2014-10-13
[ADD] add sale visible tax, that show the subtotal tax and ...

sebastienbeau:
... subtatotal tax inc on sale order line and invoice line

### DIFF
--- a/sale_visible_tax/__openerp__.py
+++ b/sale_visible_tax/__openerp__.py
@@ -38,4 +38,4 @@
  ],
  'installable': True,
  'application': True,
-}
+ }
--- a/sale_visible_tax/__openerp__.py
+++ b/sale_visible_tax/__openerp__.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#   Module for OpenERP
+#   Copyright (C) 2014 Akretion (http://www.akretion.com).
+#   @author Sébastien BEAU <sebastien.beau@akretion.com>
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU Affero General Public License as
+#   published by the Free Software Foundation, either version 3 of the
+#   License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Fields should be defined as readonly. The rest LGTM.

sebastienbeau:
I do not think we need to add the readonly here, indeed the two fields are function field without inverse function so there are in readonly (as we can write on it). You can try it (you will see that the field are in readonly in the interface)

pedrobaeza:
Cool, I didn't know that. I have tried and it works perfectly.

Thanks.

:+1:

sale_visible_tax/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+###############################################################################
+
+{'name': 'Sale Visible Tax',
+ 'version': '0.0.1',
+ 'author': 'Akretion',
+ 'website': 'www.akretion.com',
+ 'license': 'AGPL-3',
+ 'category': 'Generic Modules',
+ 'description': """This module add the field "Sub total tax inc"
```
eLBati: add/adds
field/fields

eLBati:
:+1:

sale_visible_tax/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+ """,
+ 'depends': [
+     'sale',
+ ],
+ 'data': [
+     'sale_view.xml',
+     'invoice_view.xml',
+ ],
+ 'installable': True,
+ 'application': True,
```
joaoalf: I'm not sure that this module qualifies as application.

joaoalf:
It's not a show stopper but have you considered splitting away the invoice part into another module (account_invoice_visible_tax?)?

foutoucour:
Except the orthography in description pointed out by @eLBati, :+1:

sebastienbeau:
Hi I did the change. it should be ok for merging.
Regarding the split of the module, it can be a possibity. But as I never had a customer case that do accounting without sale (I think that the majority of customer use both), I am not sure a such litlle module need to be splitted. But if one someone have the use case and need to show the taxe only on invoice, I will be totaly agree to split it.
Thanks

PR #49, opened 2014-10-13
8.0 port sale sourced by line

yvaucher:
Now allows to set a warehouse on sale order line

Added a test for sale order with multiple warehouses sources


sale_sourced_by_line/view/sale_view.xml
```diff
@@ -17,9 +17,9 @@
             <field name="model">sale.order</field>
             <field name="inherit_id" ref="sale.view_order_form"/>
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='order_line']/form/group/group/field[@name='product_id']"
-                  position="after">
-                    <field name="location_id" domain="[('usage', '!=', 'view')]"/>
+                <xpath expr="//field[@name='order_line']/form/group/group/field[@name='route_id']"
```
jgrandguillaume: This doesn't work because route_id isn't define here on so lines. replacing 'route_id' by 'price_unit' works.

jgrandguillaume:
Hi,

Thanks for porting this one ! My first comments:
- The view commented in-line above
- In openerp.py, at the end, it is MTO and Drop-shipping (not MTS and..)
- Setting up a default WH on SO doesn't fulfill by default the Source WH in the SO Line

Otherwise it's working properly, trying with MTO and dropshipping route is also working. Great !

Regards,

Joël

sale_sourced_by_line/__openerp__.py
```diff
@@ -20,6 +20,7 @@
 #
 
 {'name': 'Sale Sourced by Line',
+ 'summary': 'Multiple warehouse source locations for Sale order',
  'version': '0.1',
```
lepistone: I think we can bump the version, maybe 1.0?

yvaucher: In fact it becomes the version 8.0.0.0.1 as odoo concatenate the odoo version

So I don't usually bump version when porting

yvaucher:
@jgrandguillaume fixed the view and fixed typos in `__openerp__`.py

For the WH, if it is not defined on the line, _prepare_order_line_procurement will use order.warehouse_id. I don't think duplicating the value is necessary. 

I added a better help tooltip on line warehouse_id

sale_sourced_by_line/__openerp__.py
lepistone: Yes, but that is not a straight port: the module has really been changed. I'd bump in that case.

yvaucher:
@lepistone bumped the version as you insist :smile_cat:

yvaucher:
@jgrandguillaume I setted the default warehouse_id on sale_order_line via xml view.

lepistone:
@yvaucher the build fails. Would you like me to take care of that?

lepistone:
@yvaucher we're green!

yvaucher:
@lepistone great! thanks for spliting the tests

lepistone:
:+1:

jgrandguillaume:
:+1:

PR #50, opened 2014-10-20
[8.0] port of module sale_condition_text -> sale_condition_template

yvaucher:
port of sale_condition_template to v8.0
- rename account.condition_text to base.condition_template
- move fields in view to reflect document print
- rename type of field to position
- give better position names
- split module to be able to reuse conditions for purchase condition (used in https://github.com/OCA/purchase-reporting/tree/8.0/purchase_order_webkit_with_notes)


pedrobaeza:
Please provide proper migration scripts for users coming from previous version of the module (7.0).

base_condition_template/condition.py
```diff
@@ -0,0 +1,45 @@
...
+class BaseConditionTemplate(models.Model):
+
+    """add info condition in the invoice"""
+    _name = "base.condition_template"
+    _description = "Base condition template text"
+
+    name = fields.Char('Condition summary', required=True, size=128)
+    position = fields.Selection([('before_lines', 'Before lines'),
+                                 ('after_lines', 'After lines')],
+                                'Postition',
```
guewen: s/Postition/Position/

base_condition_template/condition.py
```diff
@@ -0,0 +1,45 @@
...
+                                required=True,
+                                help="Position on document")
+    text = fields.Html('Condition', translate=True, required=True)
+
+    @api.multi
+    def get_value(self, partner_id=False):
+        self.ensure_one()
+        try:
+            lang = self.env['res.partner'].browse(partner_id).lang
+        except:
```
guewen: Isn't `except` too broad?

base_condition_template/condition.py
```diff
@@ -0,0 +1,45 @@
...
+from openerp import models, fields, api
+
+
+class BaseConditionTemplate(models.Model):
+
+    """add info condition in the invoice"""
+    _name = "base.condition_template"
+    _description = "Base condition template text"
+
+    name = fields.Char('Condition summary', required=True, size=128)
```
guewen: `size` should be removed

yvaucher:
To be improved:

We want to extract new account invoice fields in a new module to have a module that allows to install only those.

This one should depend on this new (account_)invoice_condition_template module to pass values from sale order to invoice.

base_condition_template/condition.py
```diff
@@ -0,0 +1,45 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+from openerp import models, fields, api
+
+
+class BaseConditionTemplate(models.Model):
+
+    """add info condition in the invoice"""
+    _name = "base.condition_template"
```
guewen: It should be `base.condition.template`. Using a underscore in a model name brings confusion IMO.

base_condition_template/condition_view.xml
```diff
@@ -1,50 +1,48 @@
...
     <data>
         <record model="ir.ui.view" id="view_invoice_condition">
-            <field name="name">account.condition_text.form</field>
-            <field name="model">account.condition_text</field>
-            <field name="type">form</field>
+            <field name="name">base.condition_template.form</field>
+            <field name="model">base.condition_template</field>
             <field name="arch" type="xml">
                 <form string="Invoice condition text">
                     <field name="name" select="1"/>
```
guewen: Isn't `select="1"` in views useless now that we have dedicated search views?

base_condition_template/condition.py
yvaucher: Agreed. I was hesitating on this one

base_condition_template/condition_view.xml
```diff
@@ -1,50 +1,48 @@
 <openerp>
     <data>
         <record model="ir.ui.view" id="view_invoice_condition">
```
guewen: Do we really want to keep `invoice` in the ids?

base_condition_template/condition_view.xml
```diff
@@ -1,50 +1,48 @@
...
-            <field name="name">account.condition_text.form</field>
-            <field name="model">account.condition_text</field>
-            <field name="type">form</field>
+            <field name="name">base.condition_template.form</field>
+            <field name="model">base.condition_template</field>
             <field name="arch" type="xml">
                 <form string="Invoice condition text">
```
guewen: It does not only applies to Invoices.

base_condition_template/condition_view.xml
```diff
@@ -1,50 +1,48 @@
...
             <field name="view_type">form</field>
             <field name="view_id" ref="view_invoice_condition"/>
         </record>
-        
+
         <menuitem name="Manage Conditions" id="menu_invoice_condition_base" action="action_invoice_condition" parent="account.menu_finance"/>
         <menuitem name="List Conditions" id="menu_invoice_condition_list" action="action_invoice_condition" parent="menu_invoice_condition_base"/>
-        
+
         <menuitem name="Edit Conditions" id="menu_invoice_condition" action="action_invoice_condition_form" parent="menu_invoice_condition_base"/>
```
guewen: Why is there 2 menus?

sale_condition_template/account_invoice.py
```diff
@@ -0,0 +1,49 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+from openerp.osv import orm, fields
+
+
+class AccountInvoice(orm.Model):
```
guewen: I am going to work on the module using the conditions for the invoice (so I will extract those fields from this module).

sale_condition_template/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+                              date_inv=False, context=None):
+        # function is design to return only one id
+        invoice_obj = self.pool['account.invoice']
+        inv_id = super(SaleOrder, self).action_invoice_create(
+            cr, user, order_id, grouped, states, date_inv, context=context)
+
+        invoice = invoice_obj.browse(cr, user, inv_id, context=context)
+        if isinstance(order_id, list):
+            if len(order_id) > 1:
+                raise orm.except_osv(
```
guewen: if this error is to be read by users, it should give a better message, if it is to be read by developers, it should rather be an assert

guewen:
We should probably move the modules, some thoughts:

| module | project |
| --- | --- |
| base_condition_template | ?? |
| sale_condition_template | sale-reporting |
| invoice_condition_template | account-reporting |
| purchase_condition_template | purchase-reporting |

@yvaucher proposed these dependencies:

```
base_condition_template  ---
|                           |
invoice_condition_template  purchase_condition_template
|
sale_condition_template
```

Not sure were we should put `base_condition_template`? Maybe in account-reporting as it is probably the most low-level dependency?

Another question that arises is: should we add the fields in the QWeb reports directly in these modules or create new modules just for the Qweb modifications (I vote for modification directly in the modules to avoid to have too many modules, anyway the fields are there only for the reports).

base_condition_template/condition_view.xml
guewen: Fields should be inside a group otherwise the labels are not displayed.

guewen:
Working on the invoice_condition_template one: https://github.com/OCA/account-invoice-reporting/pull/14

guewen:
Moved here: https://github.com/OCA/sale-reporting/pull/16

PR #51, opened 2014-10-23
[FIX] sale_dropshipping: set right dest_address_id

rdeheele:
to populate dest_address_id on PO,
select partner_shipping_id on sale instead of partner's delivery contact.
in this way, shipping address will be correct in the two cases :
- standard case
- case when group_sale_delivery_address option is activated

### DIFF
--- a/sale_dropshipping/purchase.py
+++ b/sale_dropshipping/purchase.py
@@ -63,9 +63,8 @@ class purchase_order(orm.Model):
             if sale_flow in ('direct_delivery', 'direct_invoice_and_delivery'):
                 partner = partner_obj.browse(cr, uid, partner_id,
                                              context=context)
-                address = partner.address_get(['delivery'])['delivery']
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This was discussed previously (maybe on Launchpad) and I don't remember if we got to a decission. Do you remember?

rdeheele:
Hi Pedro,
I don't remember about it. I can't find a such discussion.
It's time to (re-)discuss.

guewen:
@pedrobaeza do you think about this discussion? https://github.com/OCA/rma/pull/7 it is quite similar but concerns the rma.

lepistone:
:+1:

pedrobaeza:
Thanks @guewen for pointing out the debate, but in this case, seeing the code, it's not the same, because we have only one data source (the sale order), but we are not choosing the correct field, so :+1: 

Do I proceed to merge due to the two approvals?

tafaRU:
:+1:

PR #54, opened 2014-10-31
[ADD] sale_order_price_recalculation: Module to allow to recalculate on demand sale order line prices

pedrobaeza:
# 
# Recalculation of sale order lines prices on demand

This module adds a button on sale orders below pricelist that recalculates the
prices of the order lines that contain a product in them.

It is launched manually as a button to get the user decide if he/she wants to
recalculate prices when pricelist is changed.
# Usage

Inside a sale order, you can click in any moment a button called
"(Recalculate prices)", that is next to the pricelist selection, to launch
a recalculation of all the prices of the lines, losing previous custom prices.
# Known issues / Roadmap
- In a sale order with lot of lines, the recalculation may slow down, because
  sale general data (amount untaxed, amount taxed...) are recalculated for
  each line.


sbidoul:
Thanks @pedrobaeza!

I see this module calls product_id_change while the other that was mentioned on the mailing list calss price_get. Therefore this one does a bit more than updating the price.

I'm not sure which of the two approaches is better, although I would lend towards a change of the price only, which better correspond to the button name.

What do people think?

pedrobaeza:
Well, in fact, call product_id_change is the same as get the price, but avoiding possible side effects if the way the price is calculated changes. The same method is called when you change quantity, so you are not going to find things in the line that you don't want to get.

I do it this way because you can have other fields like discount modified (with module product_visible_discount), and the other method doesn't change that.

clonedagain:
Code LGTM at first glance. I'll try to test on Monday

rvalyi:
I just want to add that such code will be extremely slow on large orders line with 50+ lines because of the write on every line that will trigger a recomputation of order fields for each line.

This is okay as a 1st approach but would probably be better refactored later to collect the prices and then call a single write at the order level (with nested lines values) while disabling field recomputation and enabling them back after. We did that kind of trick for a customer in 6.1 for purchase and typically went from 15+ minutes to 20 secs on 70+ order lines (disclaimer: our localization adds a few computed fields; but still...)

pedrobaeza:
Hi, @rvalyi. Thanks for the suggestion. I will implement your approach, because indeed it scalates better. I let you know when I have it ready.

vrenaville:
This module is very similar as our module, so I approve this :+1:

pedrobaeza:
@rvalyi Can you point me to which context key defers the computation of fields?, because I can't remember it.

clonedagain:
@vrenaville do you think you could backport this to v7 so that we have a smooth migration path please?

adrienpeiffer:
:+1: 
Code review and tests with more than 200  lines in the same sale order.

sale_order_recalculate_prices/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+            if record.order_line:
+                for line in record.order_line:
+                    res = line.product_id_change(
+                        record.pricelist_id.id, line.product_id.id,
+                        line.product_uom_qty, False, line.product_uos_qty,
+                        False, line.name, record.partner_id.id, False,
+                        True, record.date_order, False,
+                        record.fiscal_position.id, False,
+                        context=self.env.context)
+                    line.write(res['value'])
```
StefanRijnhart: Can you add the names of the keyword arguments in this call to product_id_change, in order to preserve compatibility with overrides that change the order of the keyword arguments?
Doesn't the API compatibility layer take care of the context argument?

StefanRijnhart:
Is this still work in progress, as the tag implies?

sale_order_recalculate_prices/models/sale_order.py
lmignon: For information, we encountered some issues with the same kind of code.. https://github.com/odoo/odoo/issues/4698

pedrobaeza:
@StefanRijnhart, do you have any clue about what @rvalyi mentions?

sale_order_recalculate_prices/README.rst
```diff
@@ -0,0 +1,8 @@
+Recalculation of sale order lines prices on demand
```
yvaucher: Can we prettify this using OCA template ?

sale_order_recalculate_prices/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#                                                                            #
+#  OpenERP, Open Source Management Solution.                                 #
+#                                                                            #
+#  @author Carlos Sánchez Cifuentes <csanchez@grupovermon.com>               #
```
yvaucher: @pedrobaeza It seems we lost the history there, couldn't we get it?

StefanRijnhart:
@pedrobaeza: looks like you could look at setting 'recompute' to False in the context, if you think that is safe.

sale_order_recalculate_prices/models/sale_order.py
pedrobaeza: Done!

pedrobaeza:
This one was one of my old pendings. @StefanRijnhart, I have tried the context argument, but it seems it's not working. Anyway, I prefer not to put it, because it can have side effects on other added computed fields on sale order lines. I have put a note on known issues.

I have also added some tests. Please review.

sale_order_price_recalculation/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# (c) 2014 Carlos Sánchez Cifuentes <csanchez@grupovermon.com>
+# (c) 2015 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+# (c) 2015 Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Price recalculation in sales orders",
+    "version": "8.0.1.0.0",
+    "depends": [
```
lmignon, line 10: the licence miss

sale_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,60 @@
+===================================================
+Recalculation of sales order lines prices on demand
+===================================================
+
```
lmignon, line 4: Can you add the licence badge. IMO, we should move the licence badge under the module title otherwise it's not displayed on the app store since when the file is converted to html the badge is outside a section element. I'll propose a PR on the template in maintainer-tools to move the badge under the module title.

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+    def recalculate_prices(self):
+        for line in self.mapped('order_line'):
+            order = line.order_id
+            res = line.product_id_change(
+                order.pricelist_id.id, line.product_id.id,
+                qty=line.product_uom_qty, uom=False,
+                qty_uos=line.product_uos_qty, uos=False, name=line.name,
+                partner_id=order.partner_id.id, lang=False, update_tax=True,
+                date_order=order.date_order, packaging=False,
+                fiscal_position=order.fiscal_position.id, flag=False)
```
gurneyalex, line 23: I'm concerned about setting uom and uos to False here, because I think product_id_change will reset these to the default product's uom and uos (not sure, and that method is complex enough for the outcome to be uncertain). 

I'd feel much reassured if you could update the automated test to have a non default unit for on the sale.order.line and check that the quantity + unit are not affected by `recalculate_prices`.

lmignon:
+1 with @gurneyalex concern

sale_order_price_recalculation/models/sale_order.py
pedrobaeza, line 23: This is exactly what Odoo does in its onchange: https://github.com/odoo/odoo/blob/8.0/addons/sale/sale_view.xml#L128

That's why I have copied as is. I don't think it's a problem, because that arguments are for conversions, but I'm passing qty and qty_uos

gurneyalex, line 23: I've just tested this on the runbot and I confirm it does not work

Steps to reproduce:
1. activate "manage UoM" in sales settings
2. create a sale order
3. add a sale order line for 2 dozens TONER
4. click "recalculate prices"

Expected result:
- nothing changes

Actual result: 
- the sale order is updated to a sale of 2 Unites TONER

==> needs fixing.

sale_order_price_recalculation/README.rst
pedrobaeza, line 4: I have added it, although it looks weird...

pedrobaeza:
Changes done!

pedrobaeza:
All changes done!

gurneyalex:
:+1:

gurneyalex:
@lmignon if it's ok with you I'll merge.

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+    @api.multi
+    def recalculate_prices(self):
+        for line in self.mapped('order_line'):
+            order = line.order_id
+            res = line.product_id_change(
+                order.pricelist_id.id, line.product_id.id,
+                qty=line.product_uom_qty, uom=line.product_uom.id,
+                qty_uos=line.product_uos_qty, uos=line.product_uos.id,
+                name=line.name, partner_id=order.partner_id.id, lang=False,
+                update_tax=True, date_order=order.date_order, packaging=False,
```
lmignon, line 22: IMO, you should also provide the packaging from the line even if it's not used by default but if the method is redefined by others modules, the params is part of the api..

sale_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,63 @@
...
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
yvaucher, line 6: @pedrobaeza Usually we put the license image on top of the file you just need to swap it with the title

https://github.com/pedrobaeza/sale-workflow/blob/8.0-sale_order_recalculate_prices/sale_order_price_recalculation/README.rst

sale_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,63 @@
...
+`here <https://github.com/OCA/sale-workflow/issues/new?body=module:%20sale_order_price_recalculation%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Firstname Lastname <email.address@example.org>
```
yvaucher: Your name is missing there

sale_order_price_recalculation/README.rst
pedrobaeza, line 6: This is because the previous comment of @lmignon. See https://github.com/OCA/sale-workflow/pull/54/files#r38179222

sale_order_price_recalculation/__openerp__.py
pedrobaeza, line 10: Done

yvaucher, line 10: (fixed below)

sale_order_price_recalculation/models/sale_order.py
pedrobaeza, line 22: But this will add an unneeded dependency to sale_stock module. Are you sure this is needed?

sale_order_price_recalculation/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+        "sale",
+    ],
+    'license': 'AGPL-3',
+    "author": "AvanzOSC,"
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
+              "Grupo Vermon,"
+              "Odoo Community Association (OCA)",
+    "category": "Sales Management",
+    "website": "http://www.odoomrp.com",
```
yvaucher: Is it intended to point to odoomrp?

sale_order_price_recalculation/README.rst
pedrobaeza: Sorry, I forgot it

sale_order_price_recalculation/__openerp__.py
pedrobaeza: Well, this module was under that scope, but I have changed it.

sale_order_price_recalculation/README.rst
yvaucher, line 6: Indeed it looks weird ...

We could try with a 

```
|Licence|

.. |Licence| image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
   :alt: License: AGPL-3
   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
```

yvaucher, line 6: Nope it doesn't work :-/

yvaucher, line 6: Have we already tried to put it after the title? Is there an example we can see on apps?

pedrobaeza, line 6: No, I'm afraid not

yvaucher, line 6: https://github.com/OCA/stock-logistics-transport/blob/8.0/stock_shipment_management/README.rst
https://www.odoo.com/apps/modules/8.0/stock_shipment_management/
I added a logo after title and one before title using http instead of https

And in odoo it is well displayed. Don't they use the same lib in odoo and on apps?

I sent a message to apps@odoo.com the issue is clearly on their side

yvaucher, line 6: We can see that the page is up to date on apps.odoo.com as there are empty <a> balises for the logo
I guess they don't support svg files ...

pedrobaeza, line 6: OK, let's move then with this module in any position you want, and wait for the answer and decide on maintainer-tools about the other topic, OK?

lmignon, line 6: :+1:

lmignon:
:+1:

pedrobaeza:
Please merge this old one that it's already completed.

sale_order_price_recalculation/models/sale_order.py
lmignon, line 22: @pedrobaeza sorry for the noise. You are right. The filed packaging is not defined on the sale.order.line in this module. Strange to see a parameter for a field defined in an optional module...

PR #55, opened 2014-11-04
update travis config

gurneyalex:
simplify environments by using UNIT_TEST=1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,8 @@ python:
   - "2.7"
 
 env:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But does this way execute all tests? Does it make it in an isolated way? Because I think something important is to test possible conflicts between them.

bwrsandman:
Could you also remove .coveragerc as it is supplied by the tests

gurneyalex:
builds with this config take way to much time, and we are finding this difficult to cope with in OCA/vertical-ngo and we are reverting the change. I'm withdrawing this PR for now.

PR #56, opened 2014-11-10
Add automatic relation of SO line and logisitc route based on sourced_by...

nbessi:
... field.

Wee look for the PO related to current SO line by the sourced_by fields.

If the PO has a destination location with usage "customer" we apply the dropshipping route to SO line.

If the PO has a destination location with usage "internal" we apply the make to order route to SO line.

### DIFF
--- a/sale_quotation_sourcing/__openerp__.py
+++ b/sale_quotation_sourcing/__openerp__.py
@@ -40,11 +40,16 @@
     and the manually sourced one.
 
     The drop shipping case is handled as well, with a warning to check if the
-    destination locations of the procurement and the sourced PO are
-    consistent. Since the stock_dropshipping module contains little more than
-    preconfigured Routes, Rules, and Picking Types, we do not depend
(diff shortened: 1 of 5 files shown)


jgrandguillaume:
@nbessi Can you fix travis please ?

nbessi:
@jgrandguillaume not that week. I have asked @lepistone  to have a look

lepistone:
Cool, travis is red because of a real functionality change and not becase or some weird incompatibility. It was checking for a warning in case we sourced a sale order line with dropshipping route with an rfq with destination stock.

This PR automatically changes the route for us, so we have no warning anymore.

lepistone:
Now the new test is green and the new functionality is covered. I updated the documentation to explain why the stock_dropshipping dependency was added again.

It's now :+1: for me, thanks!

jgrandguillaume:
:+1:  Thanks for the work ! I'll merge that.

PR #60, opened 2014-11-14
Fix display exception

lepistone:


lepistone:
#61 is now merged, I'm restarting travis.

lepistone:
green like a :cactus:

max3903:
:+1:

sale_exceptions/sale.py
```diff
@@ -99,16 +99,17 @@ def test_all_draft_orders(self):
...
+    @api.multi
     def _popup_exceptions(self):
-        model_data_obj = self.env['ir.model.data']
-        list_obj = self.env['sale.exception.confirm']
-        list_id = (list_obj
-                   .with_context(active_id=self.ids[0],
-                                 active_ids=self.ids)
-                   .create({}))
-        view_id = model_data_obj.get_object_reference(
+        ModelData = self.env['ir.model.data']
```
pedrobaeza: PEP8 says that you shouldn't use upper letters on local variable names: http://legacy.python.org/dev/peps/pep-0008/#id37

lepistone: Wizard is indeed a local variable, but it is also a class, or at least the closest thing we have to a class.

My argument for `Wizard` is that the deep meaning of that is really that of a class, not an instance (also, it represents a table in the database). For me, fact that odoo uses a registry to recover it instead of imports, does not change what it is.

pedrobaeza: I like your arguments, but up to now, convention is to put with _obj suffix. As other times, the problem here is consistency across all code, but this is not a blocking issue, so as you are the committer, you decide.

pedrobaeza:
:+1:

sale_exceptions/sale.py
lepistone: Pedro, I see your point of consistency.

Still, this is one of those cases where I think the old-time openerp convention is going against a more sane Python convention. I confess, I already slipped in many other `Classes` :see_no_evil: 

Let's leave that one, and I'll drop a small mail to the maining list. (OT: this is a minor topic wrt others where conventions can be improved upon, I'm thinking of testing).

Thanks!

sale_exceptions/sale.py
```diff
@@ -99,16 +99,17 @@ def test_all_draft_orders(self):
...
+        Wizard = self.env['sale.exception.confirm']
```
qtheuret: Here, wizard is not a class definition, it's a variable, so mustn't have upper letter.
Upper letter should only be used for class name definition or class instanciation like Wizard().

lepistone: Openerp has its own ways, like you never create an instance with `Wizard()` but with `Wizard.create()`, and you don't `import Wizard` but `self.env['my.wizard']`. Still what I call `wizard` deeply is in an instance of what I call `Wizard`.

But, ok, thanks for joining in!

sale_exceptions/sale.py
sbidoul: Please no uppercase in local variables. That said I agree _obj is not a nice convention, as the object behind such variables is closer to a class than an instance. _model would be better, but it's a bit long.

sale_exceptions/sale.py
dreispt: IMO there are two "conventions": PEP8 standards, and Odoo codebase _de facto_ practices. Both are legit and should be accepted.

sale_exceptions/sale.py
LeartS: I use `_model` suffix and there are no real problems, it's only 2 letters longer than `_obj` while being a lot more clear and correct.

sale_exceptions/sale.py
LeartS: When global language conventions clash with "local" project conventions, basically all standards say to use the local one.

In the first lines of PEP8: 

> Many projects have their own coding style guidelines. In the event of any conflicts, such project-specific guides take precedence for that project. 

A little later:

>  Some other good reasons to ignore a particular guideline:
> - When applying the guideline would make the code less readable, even for someone who is used to reading code that follows this PEP.
> - To be consistent with surrounding code that also breaks it (maybe for historic reasons) -- although this is also an opportunity to clean up someone else's mess (in true XP style). 

So I'd say to prefer Odoo consistency at the moment, and push for a change in Odoo conventions at the same time, as some of those are really horrible.

pedrobaeza: Hi, @LeartS, thanks for your comments. Can you please put them on the mailing list where @lepistone has started a discussion about this topic?

lepistone:
God rid of the `Model` after the comments. Thanks!

tafaRU:
:+1:

yvaucher:
:+1:

PR #62, opened 2014-11-14
fix future exception for dropshipping

lepistone:
In a previous commit by yours truly, I disabled the stock-based checks
in dropshipping situations, as was suggested in #47.

The idea was correct, but the method future_orders_are_affected needs to
return False to skip the check (it means that future orders are _not_
affected, and we are cool).

### DIFF
--- a/sale_exception_nostock/__openerp__.py
+++ b/sale_exception_nostock/__openerp__.py
@@ -19,7 +19,7 @@
 #
 #
 {'name': 'Sale stock exception',
- 'version': '1.0',
+ 'version': '1.1',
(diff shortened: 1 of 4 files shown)


pedrobaeza:
:+1:

gurneyalex:
@lepistone I think automated test would be useful to document the behavior. Is it feasible?

lepistone:
yep, that was friday afternoon laziness. Let me give a look.

lepistone:
@gurneyalex I added two unit tests

nbessi:
:+1:

PR #63, opened 2014-11-18
8.0: add module sale_tax_included_excluded

alexis-via:
This PR is the remplacement of Sebastian BEAU's PR https://github.com/OCA/sale-workflow/pull/46 on the same topic.

This module aims at solving the following dilemna: in OpenERP by default, you can't properly manage both prices with tax included in price and prices with tax excluded in price. It is an important need if your company does both B2B (tax excluded from price) and B2C (tax included in price, at least in Europe). You also have this need if you do only B2C with tax included in price and you have some foreign B2C customers that buy without VAT and you want to sell them with a lower price without VAT.

This module is the best option we found (after evaluating many different options) in order to handle both prices with tax included in price and with tax excluded in price. One of the goals of this module is minimize the impact on the datamodel of OpenERP and keep it small and simple. As we tried to keep the code as small as possible, we faced some issues with the function product_id_change of the official addons. So, in the short term, we supply a patch to update the code of the sale module and the product_visible_discount module. But, if this PR on odoo/odoo is merged https://github.com/odoo/odoo/pull/3717, I will be able to update the code of this module and remove the need to patch the official addons.


brother-bernard:
Functionnal review: the module works fine ; we will use it soon :+1:

sebastienbeau:
It's ok for me too :+1: Well done Alexis

jgrandguillaume:
Hi,

Thanks for this contrib ! If I understand this PR properly, the purpose is to have the possibility to define various prices on product, some include tax, the other exclude tax. The price list then compute the sale price accordingly. Is that correct ? I'm not sure I really understand the purpose of this feature here.. Can you may be update the description to be more clear ?

Then this is blocking me to vote for this PR:

1) We cannot accept this PR as long as the patch is not in the core, I'm sorry. OCA cannot rely on core patches... Or it is merged, or it'll stay out of OCA. I voted on the official PR

2) Why is travis green if we need the patch ? That is strange to me, it shouldn't. By the way, I see no tests at all here and honestly, talking about price computation deserve at least one please.

So, currently, I :-1: 

Regards,

Joël

alexis-via:
@jgrandguillaume I understand your point about the current need for a patch. Let's hope that my MP on odoo will be accepted in v8, so that we won't need a patch any more. Thanks for your review of my MP on Odoo.

You are also right to say that the module description is not clear enough. I'll try to find some time to write a real documentation for this module, that explains the problem in details and how to configure openerp with this module to handle at the same time prices with taxes and prices without taxes. The other reviewers of this MP already know how the module works in details because I explained them face to face...

alexis-via:
As Joel doesn't want in OCA modules that required a patch on the addons and that Odoo S.A. only merged half of my PR https://github.com/odoo/odoo/pull/3717 , I think this module will never be accepted in OCA for v8. So I decided to maintain it outside of OCA for v8:

https://github.com/akretion/sale-tax-included-excluded

And we'll so for later version...

alexis-via:
By the way, the following statement of my first comment:
"You also have this need if you do only B2C with tax included in price and you have some foreign B2C customers that buy without VAT and you want to sell them with a lower price without VAT."
is not true any more since this commit of August 24 2015 in v8 : https://github.com/odoo/odoo/commit/503820acb6d82a85c0c49ac26f7e7f8cd73851dd

But this module is still needed for those who do both B2B and B2C and want to define prices with tax included and tax excluded.

PR #64, opened 2014-11-19
8.0 : Add module sale_partner_order_policy

alexis-via:
This module is the "twin" of the module purchase_partner_invoice_method that was merged yesterday in OCA/purchase-workflow. I promised that if the module purchase_partner_invoice_method was merged in OCA/purchase-workflow, then I would develop the equivalent module for sale. Here it is !

### DIFF
--- a/sale_partner_order_policy/partner.py
+++ b/sale_partner_order_policy/partner.py
@@ -1,0 +1,42 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Sale Partner Order Policy for Odoo
+#    Copyright (C) 2014 Akretion (http://www.akretion.com)
+#    @author Alexis de Lattre <alexis.delattre@akretion.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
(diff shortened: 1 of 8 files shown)


sale_partner_order_policy/partner.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    customer_order_policy = fields.Selection([
+        ('manual', 'On Demand'),
+        ('picking', 'On Delivery Order'),
+        ('prepaid', 'Before Delivery')
+        ],
+        string='Customer Create Invoice', company_dependent=True,
```
pedrobaeza, line 34: Why don't you call it 'Customer invoice creation policy'?

pedrobaeza:
A minor question, but the rest :+1:

sale_partner_order_policy/partner.py
alexis-via, line 34: I kept the string of the field that is present on the sale.order object. I can agree with you that OpenERP S.A. didn't choose the best possible name on sale.order, but I want to keep that name, so that the user will understand the link between the field on res.partner and the field on sale.order.

pedrobaeza, line 34: OK, let's keep it

bealdav:
:+1:

jgrandguillaume:
:+1:

PR #68, opened 2014-11-20
8.0 sale_sourced_by_line - Fix is_shipped method

yvaucher:
Give positive attitude to method is_shipped which was always False.

### DIFF
--- a/sale_sourced_by_line/model/sale.py
+++ b/sale_sourced_by_line/model/sale.py
@@ -105,15 +105,33 @@ class sale_order(orm.Model):
     # OVERRIDE to use sale.order.line's procurement_group_id from lines
     ###
     def _get_shipped(self, cr, uid, ids, name, args, context=None):
+        """ As procurement is per sale line basis, we check each line
+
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

lepistone:
a "check" method like that, with many corner cases, is a good candidate for isolated unit tests (i.e. not full stories). It can probably be tested without too much effort with the shiny `Model.new()` or mocks.

jgrandguillaume:
I've tested it, it works. I'm in favor of having a test here. This is an important feature for invoicing process. After that, you get my :+1:

yvaucher:
I have some tests ready on https://github.com/yvaucher/sale-workflow/tree/8.0-fix-sale_sourced_by_line-is_delivered-add-tests

But is it stalled because of: https://github.com/odoo/odoo/issues/3922

lepistone:
I see @yvaucher . I am merging this one, you can create a PR for the tests that are stalled. Thanks!

PR #69, opened 2014-11-21
[ADD] Module sale_cancel_picking

achaussende:
OpenERP 7.0 Module to cancel a sale order and by the same time, the related out picking.

PR #71, opened 2014-11-21
[ADD] Modules sale_cancel_purchase and sale_purchase_link

achaussende:
OpenERP 7.0 Module to cancel a sale order and by the same time, related purchase order lines. That simply delete the lines that are linked to the sale order thanks to sale_purchase_link.

sale_cancel_purchase depends on sale_purchase_link
This modules need sale_cancel_picking that you can find on this PR : https://github.com/OCA/sale-workflow/pull/69


sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,58 @@
...
+from osv import orm
+
+class SaleOrder(orm.Model):
+    _inherit = 'sale.order'
+
+    def action_cancel(self, cr, uid, ids, context=None):
+        picking_obj = self.pool.get('stock.picking')
+        po_line_obj = self.pool.get('purchase.order.line')
+        for order_id in ids:
+            order = self.browse(cr, uid, [order_id], context=context)[0]
```
sebastienbeau: never browse the record inside the loop !

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,58 @@
...
+                cr, uid, [('sale_order_id','=',order_id)], context=context
+            )
+            for po_line in po_line_obj.browse(cr, uid, po_line_ids,
+                                                context=context):
+                if not order.cancel_logs:
+                    cancel_logs = ""
+                else:
+                    cancel_logs = order.cancel_logs
+                if po_line.order_id.state in ['draft']:
+                    po_line_obj.unlink(cr, uid, po_line_ids, context=context)
```
sebastienbeau: you should unlink only the current po_line not all po_line_ids here

sale_purchase_link/sale_purchase_link.py
```diff
@@ -0,0 +1,32 @@
...
+##############################################################################
+
+from osv import orm, fields
+
+class ProcurementOrder(orm.Model):
+    _inherit = 'procurement.order'
+
+    def create_procurement_purchase_order(self, cr, uid, procurement, po_vals,
+                                            line_vals, context=None):
+        if procurement.origin:
```
sebastienbeau: Not sure that using the orgin is the best way. Can you test with kit? Maybe we can have a better solution by looking to the outgoing move which is linked to the sale order?

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,58 @@
...
+                            pickings" % len(picking_ids)},
+                        context=context
+                    )
+                else:
+                    self.write(
+                        cr, uid, order_id,
+                        {'cancel_logs': cancel_logs + "<p>Impossible to cancel Sale\
+                        Order. State is %s" % order.state},
+                        context=context
+                    )
```
sebastienbeau: You should add a berfore log message. Something like "can not delete the purchase line for the product XXXX as the purchase order YYYY is in the state ZZZZZ.
Maybe it will be better to add a special method "add_log" that will add a new log?

sale_purchase_link/sale_purchase_link.py
achaussende: It works for kits too. There is no outgoing move linked to the sale order or maybe, I don't see what do you mean. But I agree. The "origin" solution is not good.

guewen:
@sebastienbeau 

> never browse the record inside the loop !

Do never say never :-)
It is true usually (like in this case), but I have seen cases when the ids list is big and there is a refresh() inside the loop where the duration of a loop is growing exponentially.

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,51 @@
...
+        for order in self.browse(cr, uid, ids, context=context):
+            po_line_ids = po_line_obj.search(
+                cr, uid, [('sale_order_id', '=', order.id)], context=context
+            )
+            for po_line in po_line_obj.browse(cr, uid, po_line_ids,
+                                              context=context):
+                if po_line.state in ['draft']:
+                    po_line_obj.unlink(cr, uid, [po_line.id], context=context)
+                    log_line = "<p>Number of deleted lines in Purchase \
+                        Orders: %d</p>" % len(po_line_ids)
```
guewen: Do not use \ and even less inside strings.
BTW the line is not translated. What is `order.add_logs()`?

```
log_line = ("<p>Number of deleted lines in Purchase "
            "Orders: %d</p>" % len(po_line_ids))
```

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,51 @@
...
+                    picking_ids = picking_obj.search(
+                        cr, uid, [('purchase_id', '=', po_line.order_id.id)],
+                        context=context
+                    )
+                    for picking in picking_obj.browse(cr, uid, picking_ids,
+                                                      context=context):
+                        if picking.state in ['assigned', 'confirmed', 'draft']:
+                            picking.action_cancel()
+                            log = "<p>Canceled picking in: %s</p>" % \
```
guewen: Same remark with the \

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,51 @@
...
+                                picking.name
+                        else:
+                            log = "<p>Can't cancel picking in: %s</p>" \
```
guewen: Same remark with the \

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,51 @@
...
+                                % picking.name
+                        order.add_logs(log)
+                else:
+                    log_line = "<p>Impossible to cancel Purchase Order Line \
+                        for product %s because Line's state is in %s" % \
```
guewen: Same remark with the \

sale_purchase_link/sale_purchase_link.py
```diff
@@ -0,0 +1,58 @@
...
+
+class ProcurementOrder(orm.Model):
+    _inherit = 'procurement.order'
+
+    def create_procurement_purchase_order(self, cr, uid, procurement, po_vals,
+                                          line_vals, context=None):
+        if procurement.origin:
+            sale_order_obj = self.pool.get('sale.order')
+            sale_order_ids = sale_order_obj.search(
+                cr, uid, [('name', '=', procurement.origin)]
```
guewen: Is it the only way to find the sale order? It is very weak...

sale_purchase_link/sale_purchase_link.py
```diff
@@ -0,0 +1,58 @@
...
+        id = result and result[1] or False
+        result = act_obj.read(cr, uid, [id], context=context)[0]
+        line_ids = po_line_obj.search(
+            cr, uid, [('sale_order_id', 'in', ids)], context=context
+        )
+        po_ids = []
+        for line in po_line_obj.browse(cr, uid, line_ids, context=context):
+            po_ids.append(line.order_id.id)
+        result['domain'] = "[('id', 'in', [" + \
+            ','.join(map(str, po_ids)) + "])]"
```
guewen: `"[('id', 'in', %s)]" % po_ids` is simpler

sale_purchase_link/sale_purchase_link.py
```diff
@@ -0,0 +1,58 @@
...
+##############################################################################
+#
+#  licence AGPL version 3 or later
+#  see licence in __openerp__.py or http://www.gnu.org/licenses/agpl-3.0.txt
+#  Copyright (C) 2014 Akretion (http://www.akretion.com).
+#  @author Adrien CHAUSSENDE <adrien.chaussende@akretion.com>
+#
+##############################################################################
+
+from osv import orm, fields
```
guewen: `from openerp.osv import orm, fields`

sale_cancel_purchase/sale.py
```diff
@@ -0,0 +1,51 @@
...
+##############################################################################
+#
+#  licence AGPL version 3 or later
+#  see licence in __openerp__.py or http://www.gnu.org/licenses/agpl-3.0.txt
+#  Copyright (C) 2014 Akretion (http://www.akretion.com).
+#  @author Adrien CHAUSSENDE <adrien.chaussende@akretion.com>
+#
+##############################################################################
+
+from osv import orm
```
guewen: `from openerp.osv import orm, fields`

sale_cancel_purchase/sale.py
achaussende: add_logs is a method defined in an other module (which you can see in this PR https://github.com/OCA/sale-workflow/pull/69) that allows to add a message to a log field in sale order.

sale_cancel_purchase/sale.py
achaussende: Is fields necessary as I'm not using it at anytime ?

pedrobaeza: Not in this case, but he is refering to put `from openerp.osv import orm` (that is to say, `openerp.osv` form)

guewen: yes sorry, without the fields

sale_cancel_purchase/sale.py
guewen: Pity that it is not called `add_cancel_log()`. But so it must be translated.

achaussende:
I fixed the link between the purchase order and the sale order with Sebastien Beau's idea, by using the move on the Procurement Order. Tell me if it's ok.

sale_cancel_purchase/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+{
+    'name': 'Sale Cancel Purchase',
+    'version': '1.0',
+    'category': 'Generic Modules',
+    'description': """
+        Allows direct Sale cancellation.
+        Cancels related Purchase Order by the same time.
+        Logs will be stocked in a field of the sale order.
+    """,
+    'author': 'Akretion',
```
gurneyalex, line 32: please add "Odoo Community Association (OCA)" to the list of authors.

StefanRijnhart:
Abandoned as per author's comment on #69

PR #72, opened 2014-11-24
8.0 Set same domain in wizard as on sale order to find po line to fill source_by field

yvaucher:
Sync domains in wizard and on sourced_by in sale_order.py

EDIT: Changed both domain instead of copying one.


pedrobaeza:
:+1:

lepistone:
:+1:

PR #73, opened 2014-11-24
[IMP] the exception should be also visible in the quotation

sebastienbeau:
The main exception field was missing in the tree view. Now it's fixed ;)


pedrobaeza:
:+1:

lepistone:
:+1:

PR #74, opened 2014-11-26
Add sale default discount

lepistone:
migrated from 

https://code.launchpad.net/~camptocamp/sale-wkfl/7.0-add-sale-default-discount-lep/+merge/227214


sale_default_discount/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import model  # noqa
```
pedrobaeza: Why are you disabling quality tools here?

sale_default_discount/model/sale.py
```diff
@@ -0,0 +1,47 @@
...
+
+from openerp.osv import orm, fields
+import openerp.addons.decimal_precision as dp
+
+
+class SaleOrder(orm.Model):
+    _inherit = 'sale.order'
+
+    _columns = {
+        'discount': fields.float(
```
pedrobaeza, line 30: Why don't you put it as an stored related field? It serves the same.
I prefer to name this field also 'default_discount' for not getting confused with a general discount (for example, volume discount).

pedrobaeza, line 30: I'm thinking that maybe your intention is to let the field to be changeable while entering lines, so my comment doesn't apply in that case.

sale_default_discount/__init__.py
lepistone: I guess because I wrote that before we started ignoring import problems in `__init__.py` with our tools. I stopped using that now. I'm getting rid of it.

sale_default_discount/model/sale.py
lepistone, line 30: yep, that's the case!

yvaucher:
Failing on tests

```
Assertion "Order discount should be set" FAILED
test: discount == 32.0
values: ! 0.0 == 32.0
```

gurneyalex:
@lepistone could you rebase and check if travis is still red?

lepistone:
rebased

pedrobaeza:
Any news on this PR?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #77, opened 2014-11-26
[ADD] crm track all lead fields

max3903:


yvaucher:
:+1:

pedrobaeza:
Are "magic columns" ignored with this method?

I think this can be extracted as a general option for any model, don't you think? It could be very interesting.

lepistone:
@pedrobaeza I think it would be cool to have this for any model, but at the moment I think we can merge this one nonetheless. Do you agree?

code review is :+1: for me. Thanks!

pedrobaeza:
OK, we can continue with this and talk about that later. :+1:

PR #78, opened 2014-11-27
add check: route is consistent with selected PO

lepistone:


jgrandguillaume:
:+1:  That's perfect

sale_quotation_sourcing/model/sale_order.py
```diff
@@ -71,6 +76,14 @@ class SaleOrderLine(models.Model):
                " ('order_id.state', 'in', ['draft', 'confirmed'])]")
 
     @api.multi
+    def has_consistent_route(self):
+        self.ensure_one()
+        if self.route_id and self.sourced_by:
+            selected_route = self.route_id
+            self.set_route_form_so()
+            return selected_route == self.route_id
```
gurneyalex, line 84: this is a public API function, it should not be possible for it to return `None`. Could we have an explicit `return False`?

lepistone, line 84: we could

gurneyalex:
:+1:

jgrandguillaume:
Re-tested and :+1:

PR #80, opened 2014-12-01
fix: no route or source is always consistent

lepistone:

PR #81, opened 2014-12-01
split build because of incompatible tests

lepistone:
The modules are not incompatible in production. They are when running
tests. That is of course unfortunate and should be investigated further
(i.e. a module has no access to methods added in other modules, while
constraints created in XML data are, so the constraints crash).


lepistone:
I'm closing this because it is superseded by #80.

PR #82, opened 2014-12-01
Skip exceptions for consumable product

lepistone:


lepistone:
I added the label WIP because I have to update existing tests that were passing because of a bug: the `_is_make_to_stock` method was never called if a product was defined.

lepistone:
Actually the old integration test is fine, the bug was in `_is_make_to_stock` that as I said was almost never called (bug introduced by yours truly).

This needs review then.

gurneyalex:
Do you think this deserves a backport to 7.0?

gurneyalex:
:+1:

jgrandguillaume:
:+1:

PR #83, opened 2014-12-01
8.0 fix sale sourced by line is delivered add tests

yvaucher:
Add missing tests on computed field `shipped`

However this is pending due to issue: odoo/odoo#3922

### DIFF
--- a/sale_sourced_by_line/model/sale.py
+++ b/sale_sourced_by_line/model/sale.py
@@ -19,10 +19,11 @@
 #
 #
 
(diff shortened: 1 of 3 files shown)


yvaucher:
@gurneyalex With new API and with last fix on stored fields on odoo it's now green. We can depreciate #112 as it was only fixing store trigger function

rdeheele:
:+1:

lepistone:
:+1:

PR #86, opened 2014-12-02
[ADD] Module sale_dropshipping_location

JordiBForgeFlow:
Introduces the possibility to receive drop shipment stock into a virtual location within the company, where the reception move is automatically chained to the default customer location.

See: http://www.eficent.com/openerp_en/drop-shipments-improvements-in-odoo/

@rvalyi As author of the "sale_dropshipping" module you might be interested to review.

Note: Incorporated changes to transact a correct drop shipment returns.


JordiBForgeFlow:
I'm closing this PR. Creating a virtual location is not the way to resolve the problem. The root cause of the problem is that Account Anglo Saxon does not properly handle drop shipments.

PR #88, opened 2014-12-16
[7.0] [ADD] sale show addresses

veloutin:
This module changes the Sale Order UI so that the partner fields (partner, invoice, shipping) show
the address details in a separate field. This allows viewing the address details immediately when editing a new record.

### DIFF
--- a/sale_show_addresses/__init__.py
+++ b/sale_show_addresses/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2014 Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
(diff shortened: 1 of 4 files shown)


sebastienbeau:
Hi Vincent

I was facing to the same need this week, Alexis my coworker did some small UI module improvement. I want to extract some module in OCA. One of them solve the same problem (or maybe I miss something) but in a really different way. Instead of adding a new field, it change the context so Odoo will show the full address. So basically is just an inherit on the view and the code is here : https://github.com/akretion/odoo-usability/blob/8.0/sale_stock_show_delivery_address/sale_view.xml

What do you think about this simplier solution, does it fit with your need? Should I extract and adapt Alexis module in order to merge it in OCA, or should we use your propose solution?

Thanks for you feedback

pedrobaeza:
I see clearer the @sebastienbeau's solution.

max3903:
@sebastienbeau Does it work for all the addresses on the sales order (Customer, Invoice Address, Delivery Address) ?

veloutin:
@sebastienbeau The primary goal of the module here is actually to be able to see the address details when creating a sale order, before hitting save. Does "always_reload" accomplish this? I do think a few lines of XML is simpler than the current implementation of this module.

sebastienbeau:
Hi @max3903 @veloutin 
Alexis just inherit the view for showing the full address of the delivery in the sale order and the stock picking. We can do the same inherit  for the invoice address and the customer address.

@veloutin 
The "always reload" will force to reload the address and as in the context there is the 'show_address':  key it will show the full address. I didn't know this tips until Alexis show me his work.

But this solution DO NOT show the full address before hitting the save button. You only have the full address after saving it! So I confirm that this solution do not fit for your need.

I don't think we can solve this issue with only xml, may be an alternative solution (it's just an opinion) will to inherit the widget many2one in order to have a special address_many2one that show the full address when selection an address.

I am ok with your merge :+1:

max3903:
:+1: 

Works like a charm!

Issue #96, opened 2015-01-21
7.0 sale_fiscal_position_update : error in a particular scenario


alexis-via:
Here is the scenario to reproduce the bug on v7:
1) install the module sale_fiscal_position_update
2) create a new quotation:
customer : Agrolait
create a sale order line : 1 x USB adapter
change the fiscal position to a new value
SAVE the quotation (must be done only now)
=> you get an error message and a backtrace:

```
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: state - state]
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo-70/openerp/netsvc.py", line 296, in dispatch_rpc
    result = ExportService.getService(service_name).dispatch(method, params)
  File "/home/alexis/new_boite/dev/odoo-70/openerp/service/web_services.py", line 632, in dispatch
    res = fn(db, uid, *params)
  File "/home/alexis/new_boite/dev/odoo-70/openerp/osv/osv.py", line 193, in execute_kw
    return self.execute(db, uid, obj, method, *args, **kw or {})
  File "/home/alexis/new_boite/dev/odoo-70/openerp/osv/osv.py", line 177, in wrapper
    netsvc.abort_response(1, _('Integrity Error'), 'warning', msg)
  File "/home/alexis/new_boite/dev/odoo-70/openerp/netsvc.py", line 71, in abort_response
    raise openerp.osv.osv.except_osv(description, details)
except_osv: ('Integrity Error', 'The operation cannot be completed, probably due to the following:\n- deletion: you may be trying to delete a record while other records still reference it\n- creation/update: a mandatory field is not correctly set\n\n[object with reference: state - state]')
```

Analysis of the issue:
The "state" field in the sale.order.line is not return by the method self.resolve_2many_commands() (line 40 of sale_fiscal_position_update/sale.py) and therefore it is not present in the "value" key of the dict that is returned by the on_change of the fiscal position. As a consequence, the "state" field doesn't have a value any more in the view, but the "state" field of the sale order line is a required field, so it fails.

gdgellatly:
And that would only be because it is an editable tree view, and that state field probably isn't in it.  Probably easiest fix is to just add 

``` xml
<field state invisible="1" />
```

 to the sale order line view

alexis-via:
[...]

PR #97, opened 2015-01-21
FIX bug #96 in sale_fiscal_position_update

alexis-via:
This also includes a small code clean-up.

In fact, this fix is more a "workaround" than a real fix. The real fix should probably be in resolve_2many_commands()


pedrobaeza:
Annoying bug inded

:+1:

lmignon:
LGTM :+1:

yvaucher:
:+1:

PR #98, opened 2015-01-22
do not discard the owner in chained move

lepistone:
This addresses odoo/odoo#4839, and makes the related fix in the core
work with this module as well.


yvaucher:
:+1:

pedrobaeza:
:+1:

guewen:
:+1:

PR #99, opened 2015-01-27
Improve override of cancel button to ease inheritance of cancel reason view

yvaucher:
Only hide button instead of renaming it


pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #100, opened 2015-02-02
[8.0] product visible discount extension

veloutin:


product_visible_discount_extension/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+    'name': 'Product visible discount extension',
+    'version': '0.1',
+    'author': 'Savoir-faire Linux',
+    'maintainer': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'license': 'AGPL-3',
+    'category': 'Others',
+    'summary': 'Product visible discount extension',
+    'description': """
+Product visible discount extension
```
pedrobaeza, line 33: Please extract it to a README.rst file.

product_visible_discount_extension/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+==================================
+
+This module is based on product_visible_discount that calculate
```
pedrobaeza, line 36: Better to put "This module extends product_visible_discount..."

product_visible_discount_extension/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+the discount rate on unit price calculated from a price list in
+sale order lines. The discount rate can be either calculated from
+the Public Price of the product and the unit price written in
+the sale order line or set by the user.
```
pedrobaeza, line 40: It is not clear what this module extends exactly.

product_visible_discount_extension/models/sale.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    _name = 'sale.order.line'
+    _inherit = 'sale.order.line'
+
+    list_price = fields.Float('Price list',
+                              related="product_id.list_price",
+                              store=True,
+                              readonly=True)
+    visible_discount = fields.Float('Customer  Discount (%)')
```
pedrobaeza: Why are you using a new field instead of the existing one?

llacroix: There is a problem, if I use the real discount, it modify the `price_unit` value to the actual price it would get with the discount calculated.. Which means we can't calculate the discount again. If we were using the discount field, we would have something like this:

```
price_unit = (list_price * (1 - discount))
discount = discount
total price = price_unit * (1 - discount)
```

The result would be to calculate twice the same discount. The problem comes from openerp not showing the price after discount.

One other way to do, would be to use the price_unit as list_price without touching it. Calculating the discount and having an other field to calculate the price after discount. In the end, the discount field is used as usual, the price_unit would be exactly the same as list_price. The price after discount would have no effect anywhere else. 

But we would want to show the price after discount in reports.

pedrobaeza: OK, get it, but I see no relation then with product_visible_discount module.

llacroix: The product visible discount does the changes on the product with the price list. When "visible discount" isn't checked in the price lists, price lists don't seem to have any effect. Yet the module should be working even without that if I'm not mistaken I can check if the dependencies are really necessary.

llacroix:
Tests are passing on this module, there is a module that doesn't seem to work. I believe it's related to a test creating a model without defining a view name. I'm not sure if it's a bug in odoo or There is something specific. I had the same problem with my module. For some reasons it required me to use something like this:

```
!record {model..., id... view=False}:
```

llacroix:
If someone has an idea how to trigger change events from within the tests I'll update it that way. It's not super clean this way. Since some fields are mutual, it should trigger recursive on change events. I'm currently triggering them manually so it should always work that way. But it would be nice to see have odoo handle the event by itself and check that no matter how the events are triggered it will work.

llacroix:
Some of the bugs in travis are related to this issue as far as I can say:
https://github.com/odoo/odoo/pull/5410

A view name seems to be missing  on an inline view and the code that is supposed to log a debug messages raise a KeyError.

product_visible_discount_extension/models/sale.py
```diff
@@ -0,0 +1,206 @@
...
+    Any change to visible_discount will trigger onchanges events in that order:
+        _onchange_visible_discount
+        _onchange_price
+
+    Any change to price_unit will trigger onchange events in that order:
+        _onchange_price
+        _onchange_visible_discount
+    """
+
+    _name = 'sale.order.line'
```
joaoalf, line 57: When you use _inherit is not necessary to add _name unless you want to create a different model.

product_visible_discount_extension/models/sale.py
```diff
@@ -0,0 +1,206 @@
...
+
+from openerp.exceptions import ValidationError
+from openerp import (
+    models,
+    fields,
+    api,
+)
+
+
+def dummy_search(self, *a, **ka):
```
joaoalf, line 30: @llacroix  I'm not sure that this is the way to go. Can you clarify a little the usage of temp_discount and tempo_price?

product_visible_discount_extension/models/sale.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if 'price_unit' in res['value']:
+            new_price = None
+
+            if res['value']['discount'] == 0:
+                # If the pricelist has a fixed price it doesn't set
+                # a discount
+                product_obj = self.pool.get('product.product')
+                product = product_obj.browse(cr, uid, product, context)
+                price_list = product.list_price
```
joaoalf, line 177: @llacroix Are you considering the pricelist in the SO or just the product's base price?

product_visible_discount_extension/models/sale.py
llacroix, line 30: As there is no real way to silence change events and that these onchange events are mutual. A change to field A will trigger a change to field B that will trigger a change to field A again. This isn't probably the best way to handle this.

product_visible_discount_extension/models/sale.py
llacroix, line 177: I'm only considering the base price which is stored in the pricelist. The price_unit is correctly set to the price defined in the price_list and then we can modify the price freely. The price or discount set is the actual price that will be used later.

product_visible_discount_extension/models/sale.py
llacroix, line 57: Right.

product_visible_discount_extension/__openerp__.py
```diff
@@ -0,0 +1,62 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Product visible discount extension',
+    'version': '0.1',
+    'author': 'Savoir-faire Linux',
```
gurneyalex, line 26: please add Odoo Community Association (OCA) to authors list.

gurneyalex:
@llacroix any update on this PR?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #102, opened 2015-02-05
[7.0] sale_dropshipping: improve display on the sale view

guewen:
- The fields are hidden when they should not:
  It should be displayed even if the type if make_to_stock because the
  sale_flow 'direct_invoice' is a make_to_stock flow.
- 'Direct invoice' is a make_to_order delivery: This mode allows to use
  dropshipping for the delivery, but invoiced by us and not by the supplier. The
  onchange must change it to make_to_order to generate a supplier order.

### DIFF
--- a/sale_dropshipping/sale.py
+++ b/sale_dropshipping/sale.py
@@ -76,7 +76,7 @@ class sale_order_line(orm.Model):
 
         vals = {}
         if product_id:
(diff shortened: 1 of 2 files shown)


yvaucher:
:+1:

max3903:
:+1:

PR #103, opened 2015-02-05
[7.0] sale_dropshipping: Use the _prepare method instead of write after create

guewen:
Use the _prepare method introduced in 7.0


pedrobaeza:
Good refactoring!

:+1:

yvaucher:
:+1:

lmignon:
:+1:

PR #104, opened 2015-02-06
[ADD] Product quantity available on SO line

ghost:
Simple module that adds a field on the sale order line to see the product quantity in stock.


sale_order_line_available_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+    standard_price = fields.Float(
+        related='product_id.standard_price',
+        string='Cost Price',
+        digits_compute=dp.get_precision('Product Price'),
+    )
+
+    qty_available = fields.Float(
+        related='product_id.qty_available',
+        string='Available Quantity',
+        digits_compute=dp.get_precision('Product UoS'),
```
guewen, line 58: Is this field correct when the warehouse is changed on a sales order? (by correct I mean, it shows the quantity in the selected warehouse)

ghost, line 58: It is not. Thanks for the comment, I will fix it and add a unit test.

ghost:
@guewen Implemented your suggestion.

joaoalf:
:+1:

sale_order_line_available_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+        self.qty_available = sum(
+            quant.qty for quant in
+            self.env['stock.quant'].search([
+                ('location_id', '=', order.warehouse_id.
+                    lot_stock_id.id),
+                ('product_id', '=', self.product_id.id),
+                '|',
+                ('reservation_id.group_id.name', 'in',
+                    [False, order.name]),
+                ('reservation_id', '=', False),
```
eLBati, line 44: Why not using `qty_available` of `product.product`?

ghost, line 44: This is what I did first. But on the SO, you select a warehouse, so the logic here is to retrieve the available stock from that warehouse.

eLBati, line 44: See https://github.com/odoo/odoo/blob/8.0/addons/stock/product.py#L85
if you set `warehouse` context item and use `qty_available`, the function field should do the job for you

PR #105, opened 2015-02-06
8.0 sale order revision

foutoucour:
# Sale Order Revision

This module allows to revise sale_orders that are not in a state
accepted/ordered/finished/draft.

The revision is done using a Review button added in the header of sale order.
In addition, two new links have been added: _Revised by_ and _Revision of_. The
links ease to follow the historic of the revisions.

The revised sale order is set to _cancel_ state. It has been done this way to
keep standard workflow.

To detect a sale order as a revised sale order, it is adviced to filter:
- sale orders at _cancel_ state
- and _Revised by_ is not empty


pedrobaeza:
As name, I prefer sale_order_review, because revision can be used for saying that you have multiple versions.

Can you also squash all your commits in one, as this is the first iteration of the PR?

rvalyi:
Hello,

I also prefer to rename that one sale_order_review. In fact there is already a sale_order_revision module nearly accepted into the OCA https://code.launchpad.net/~akretion-team/sale-wkfl/70-add-sale_order_revision/+merge/202735
it's a mirror of the purchase_order_revision by Agile BG (already in OCA) so I think it makes the OCA modules consistent...

I'm actually taking some time again to get that module accepted in v7 and port it in v8 and also eventually port purchase_order_revision to v8 so I propose to evaluate your module again and possible refactors once these simpler base modules are in place in v8. What do you think? Hey I was working hard on that one year ago but customer got millions unpaid and project had to downscale big times so sorry for the delay...

foutoucour:
Hey @rvalyi, @pedrobaeza  

It sounds good. But a review and a revision is not necessary the same. I can see the terms are close and I might not have the same perception of them.
I understand review as the process of going over a subject again in study or recitation in order to fix it in the memory or summarize the facts. 
While revision would be the process to alter something already written or printed, in order to make corrections, improve, or update.

The difference would be a revision alter an existing document while review is a comment that might conduct to an alteration.
That is in that sense the module is name as it.

On the other hand, I wasn't aware about the sale_order_revision module. I didn't check bzr merge proposals.

It seems the both modules are doing the same. Would you prefer to close the PR?

ps: I just pushed to fix PEP8 issues.

pedrobaeza:
Well, they seem similar, so it makes no sense to have two modules.

I close this one then.

Sorry for the inconveniences.

eLBati:
For reference: https://github.com/OCA/sale-workflow/pull/137 `sale_order_revision` 8.0

gurneyalex:
@foutoucour sorry I missed this PR, and duplicated part of your work. Can we work together to bring #137 closer to your needs?

bwrsandman:
@gurneyalex I am taking over this issue so we can work together to see what is missing

PR #107, opened 2015-02-10
remove sourcing check on procurement validation

lepistone:
Supersedes #106.

This had two problems, and an attempt to fix did not work:
1. For MTO orders, this error in produced in background in the scheduler,
   and the error is not shown to the user. An exception blocks the
   scheduler completely in that case, and also breaks all subsequent
   runs.
2. With transit locations, the check fails because it compared "Customer"
   and "Transit". This situation is actually OK.

In an attempt to fix 1, I tried to raise the exception only when in
foreground and log an exception to the procurement otherwise. This
works, but in the MTO case the user does not see the problem and the
order is validated (so it is too late to change it).

In an attempt to fix 2, I changed the check to use the "final
destination" of the first move that generated the procurement. That
fixed the transit case, but broke the MTO case, where a BUY procurement
in Stock is chained to a MTO procurement in Customers.

There are already other consistency checks: the route of the sale order
line is changed automatically when a PO is chosen, and a sale exception
does a consistency check as well.

TL;DR: I am removing this check.

### DIFF
--- a/sale_quotation_sourcing/model/procurement.py
+++ b/sale_quotation_sourcing/model/procurement.py
@@ -19,8 +19,6 @@
 #
 from openerp import models, api, _
 
-from openerp import exceptions
-
 
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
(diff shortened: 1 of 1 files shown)


yvaucher:
Shouldn't this check be removed only if we add a key like 'automated' in context?
Isn't this warning useful when manually sourcing?

lepistone:
@yvaucher I tried! but it doesn not make sense. See the write-up above.

jgrandguillaume:
A pity, but I agree we have no choice but to remove it. So :+1:

yvaucher:
Ok the scheduler shouldn't be blocked, so :+1:

PR #108, opened 2015-02-11
lint: move imports to the top (pep8 1.6 new check)

lepistone:

PR #109, opened 2015-02-11
Remove the filter on incoming pickings from the purchase orders

guewen:
The pickings generated by purchase orders are changed to outgoing pickings when
the order's flow is drop shipping.  The 'view_picking' function gets the action
to show the pickings from the XML action. This action has a domain "('type',
'=', 'in')", that prevents to see the drop shipping pickings as they have the
type 'out'.

This commit totally removes the "('type', '=', 'in')" condition from the
action: it is useless because we already filter on the pickings attached to the
purchase order (on the 'picking_ids' field).

### DIFF
--- a/sale_dropshipping/purchase.py
+++ b/sale_dropshipping/purchase.py
@@ -20,6 +20,7 @@
 #
 #
 from openerp.osv import fields, orm
+from openerp.tools.safe_eval import safe_eval
 
 
 class purchase_order_line(orm.Model):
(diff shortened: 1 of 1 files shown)


lepistone:
The broken build is probably fixed by #108.

sale_dropshipping/purchase.py
```diff
@@ -118,6 +119,18 @@ def action_picking_create(self, cr, uid, ids, context=None):
                         context=context)
         return res
 
+    def view_picking(self, cr, uid, ids, context=None):
+        action = super(purchase_order, self).view_picking(cr, uid, ids,
+                                                          context=context)
+        if action.get('domain'):
+            domain = safe_eval(action['domain'])
+            new_domain = []
+            for criterium in domain:
```
pedrobaeza: Better if: `new_domain = [x for x in domain if x != ('type', '=', 'in')]`

lmignon:
:+1:  Travis error is unrelated and fixed by #108

lepistone:
I restarted the build.

lmignon:
@lepistone git rebase?

lepistone:
to make 7.0 green we need something like #110 for 7.0.

guewen:
> to make 7.0 green we need something like #110 for 7.0.

I cherry-picked the change: https://github.com/OCA/sale-workflow/pull/111

sale_dropshipping/purchase.py
guewen: Done

guewen:
Once #111 is merged, I will rebase

pedrobaeza:
Done

guewen:
Rebase done

PR #112, opened 2015-02-17
Recompute shipped value on each procurement state change

yvaucher:


gurneyalex:
module version update is not strictly necessary as the changes do not mandate an update of the module. 

is there a test in the module exercising this new bit of code?

yvaucher:
@gurneyalex The test is there and waiting on a odoo bug :cry: 

https://github.com/OCA/sale-workflow/pull/83

sale_sourced_by_line/model/sale.py
```diff
@@ -136,12 +136,20 @@ def _get_shipped(self, cr, uid, ids, name, args, context=None):
...
         res = set()
         proc_orders = self.pool['procurement.order'].browse(
             cr, uid, ids, context=context)
         for proc in proc_orders:
-            if proc.state == 'done' and proc.sale_line_id:
-                res.add(proc.sale_line_id.order_id.id)
+            current_proc = proc
+            while (not current_proc.sale_line_id and
+                   current_proc.move_dest_id):
+                current_proc = proc.move_dest_id.procurement_id
```
gurneyalex: :bug: infinite :loop: 

s/proc.move/current_proc.move/

gurneyalex:
yvaucher: and the test was :green_apple:  with the infinite loop inside? This is suspicious...

yvaucher:
@gurneyalex There is no test for this, the test is in another PR which should work now and make this PR obsolete: #83

yvaucher:
Thus I close this PR as #83 resolves this

PR #116, opened 2015-02-24
sale_quotation_sourcing fixes

gurneyalex:
Two issues:
- when a WH config with lots of chained locations is used, the procurement
  chain must be unrolled with a proper loop, and not with a hard coded 1 step
  lookahead.
- when stock_route_transit from OCA/stock-logistics-workflow is installed, the usage of the PO locations can be altered, and the code setting the route of the SO line must be updated

### DIFF
--- a/sale_quotation_sourcing/model/procurement.py
+++ b/sale_quotation_sourcing/model/procurement.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 #    Author: Alexandre Fayolle, Leonardo Pistone
-#    Copyright 2014 Camptocamp SA
+#    Copyright 2014-2015 Camptocamp SA
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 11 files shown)


sale_quotation_sourcing/model/procurement.py
```diff
@@ -35,12 +35,13 @@ def make_po(self):
...
-                procurement.move_dest_id.procurement_id.sale_line_id or
-                False
-            )
+            curr_proc = procurement
+            sale_line = False
+            while curr_proc:
+                if curr_proc.sale_line_id:
+                    sale_line = curr_proc.sale_line_id
+                    break
+                curr_proc = curr_proc.move_dest_id.procurement_id
```
nbessi, line 44: I'm not familiar with the version 8.0 procurement code. Is there a possibility to have a loop in procurement and a infinite while ?

sale_quotation_sourcing_stock_route_transit/model/sale_order_line.py
```diff
@@ -0,0 +1,46 @@
...
+#
+from openerp import models, api
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.model
+    def _get_po_location_usage(self, purchase_order_line):
+        """override the usage finding in case the PO delivers to a transit
```
nbessi: Can you reformulate the docstring please. I did not understand it.

sale_quotation_sourcing/model/procurement.py
gurneyalex, line 44: you can probably hand craft a loop by chaining the move_dest of a procurement to the same procurement, but this will cause the scheduler to loop, and no addon should generate such a pathological configuration.

nbessi, line 44: ok then

nbessi:
:+1:

rdeheele:
:+1:

jgrandguillaume:
:+1:

PR #118, opened 2015-03-04
[FIX] forbid to confirm a PO if SO, sourced by it, is not confirmed

yvaucher:
This ensure we won't create twice a move going from incoming transit location to stock

### DIFF
--- a/sale_quotation_sourcing_stock_route_transit/i18n/sale_quotation_sourcing_stock_route_transit.pot
+++ b/sale_quotation_sourcing_stock_route_transit/i18n/sale_quotation_sourcing_stock_route_transit.pot
@@ -1,0 +1,33 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* sale_quotation_sourcing_stock_route_transit
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 8.0\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2015-03-09 10:11+0000\n"
+"PO-Revision-Date: 2015-03-09 10:11+0000\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: sale_quotation_sourcing_stock_route_transit
+#: model:ir.model,name:sale_quotation_sourcing_stock_route_transit.model_purchase_order
+msgid "Purchase Order"
+msgstr ""
+
+#. module: sale_quotation_sourcing_stock_route_transit
+#: model:ir.model,name:sale_quotation_sourcing_stock_route_transit.model_sale_order_line
+msgid "Sales Order Line"
+msgstr ""
+
+#. module: sale_quotation_sourcing_stock_route_transit
+#: code:addons/sale_quotation_sourcing_stock_route_transit/model/purchase.py:46
+#, python-format
+msgid "You cannot confirm the purchase order before confirming the sales order(s) which use it as source. Currently the following sales order(s) still needs confirmation:\n%s"
+msgstr ""
+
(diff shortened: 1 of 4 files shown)


sale_quotation_sourcing_stock_route_transit/model/purchase.py
```diff
@@ -0,0 +1,47 @@
...
+        return bool(unconfirmed_sales_count)
+
+    @api.multi
+    def wkf_confirm_order(self):
+        """ Forbid confirmation of purchase order when there
+        are unconfirmed sale orders sourced by its purchase lines
+        """
+        if self.has_unconfirmed_sales():
+            raise exceptions.Warning(
+                _("You cannot confirm a purchase order. You have to confirm"
```
gurneyalex: I'd rephrase this as

"You cannot confirm the purchase order before confirming the sale orders that is sources. Currently, the following sale order(s) still need confirmation: %s"

This means that has_unconfirmed_sales need rewriting to return a list of SO names.

pedrobaeza: s/sale orders that is/sales orders that are/
s/sale order(s)/sales order(s)/

And better rename field to `unconfirmed_sales` and it can be a many2many, because this can be easily tested also, and you have the complete model for anything you need.

yvaucher: @gurneyalex you mean s/the sale orders that is sources./the sale orders _it_ sources. ?
@pedrobaeza you mean s/the sale orders that are sources./the sale orders that are source_d_. ?

It should mean the sales order is sourced by a purchase order

pedrobaeza: Well, my concerns were only about plural of words, not the meaning, sorry. Unfortunately, I'm not following these set of modules (although very interesting), because I'm overloaded, and I cannot apport too much, but when I see the notification mail, that choked me.

yvaucher: Ok then I'll go for @gurneyalex proposal taking care of plural form of sale

gurneyalex: @yvaucher yes, "that it sources".

gurneyalex: @pedrobaeza sorry for choking you with a typo

yvaucher:
Rephrased and translation added.

yvaucher:
that's it for the "that it" it made it too like that http://en.wikipedia.org/wiki/That_that_is_is_that_that_is_not_is_not_is_that_it_it_is

lepistone:
:+1: thanks

gurneyalex:
:+1:

ghost:
Why don't you put this check in module sale_quotation_sourcing instead of sale_quotation_sourcing_stock_route_transit?

gurneyalex:
@dufresnedavid because when you are not using transit, it works, and it did not sound like a requirement for the basic module. Upon reflection, you have a point though.

@jgrandguillaume @yvaucher: what do you think about this?

lepistone:
@yvaucher there is still an open point about moving this to `sale_quotation_sourcing`. What do you think? I've got no opinion on this, I'm happy to merge in both cases.

yvaucher:
Afterwards, the check might be too restrictive, before changing the PR I was waiting for feedback. I'll put this in WIP

sale_quotation_sourcing_stock_route_transit/i18n/sale_quotation_sourcing_stock_route_transit.pot
```diff
@@ -0,0 +1,33 @@
+# Translation of Odoo Server.
```
pedrobaeza, line 1: Remove POT file.

sale_quotation_sourcing_stock_route_transit/model/purchase.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#
+#    Author: Yannick Vaucher
+#    Copyright 2015 Camptocamp SA
```
pedrobaeza, line 4: Short header

pedrobaeza:
Is this going to be continued?

yvaucher:
I'm closing it, I won't work on this soon.

PR #121, opened 2015-03-05
7.0 add sale credit note

bguillot:
Hello, 

This PR adds the module sale_credit_note.

This module allows the user to set a part of a refund as credit note. This credit note is then available on the sale order to reduce the amount that the customer has to pay. 
This is also a base module for a module that synchronize credit memo on Magento (the PR is not created yet).

It depends on the PR https://github.com/OCA/e-commerce/pull/38 that creates a hook for the field residual on sale order.

The inheritance of the sale order view that adds the credit lines is ugly, but I needed to put the lines on the same group as the sale total group. I am open to suggestions, and if it worth it I can maybe make a PR on the addons to add a new group.

Thanks for you reviews


bguillot:
It seems that Travis fails because of the dependency on the module sale_payment_method. 
I don't know how to make it pass, the module is in e-commerce branch.

gurneyalex:
@bguillot you need to add a line in .travis.yml to pull the new dependency.

bguillot:
Thanks for the review, and the advice.

Travis is green now !

sale_credit_note/sale.py
```diff
@@ -0,0 +1,153 @@
...
+                lines.append((
+                    0, 0, {'refund_id': refund.id, 'amount': refund_amount}))
+                max_amount -= refund_amount
+        return lines
+
+
+class credit_line(orm.Model):
+    _name = "credit.line"
+
+    _columns = {
```
ghost: Please, split each model to a seperate file with the name of the model as file name. It will be easier to maintain. Also, you ommited _description here.

ghost:
@bguillot Thanks for the contribution. Could you rename your models with "sale." at the beginning of the models you created (both Models and Transients)? This way, it will be easier to maintain. credit.line could be something else not related to the sales. For exemple, it could be related to accounting. This is just my opinion, but I think the namespace a the beginning of a table name is important.

ghost:
@bguillot Thanks for the change :+1:

gurneyalex:
@bguillot i've added your github login in the OCA database. This should calm the CLA bot.

bguillot:
Thank you @gurneyalex , I was a bit confuse about that because I signed the CLA.

Thanks @dufresnedavid for the review ;)

About the runbot, should I put the dependency somewhere as for travis ?

gurneyalex:
@bguillot we have no support for this yet, and it is still manual.

gurneyalex:
@bguillot: would it not make sense to move sale_payment_method from e-commerce to sale-workflow? I'm reluctant to have this circular dependency between the two repositories. Users of e-commerce should not be affected as they already have the dependency on sale-workflow.

sebastienbeau:
@bguillot Ithink also that it's will be a better idea to move sale_payment_method in sale_workflow

sebastienbeau:
Hi after some month of using in production this module, and some customer crazy case, we work on a new version that will be quite different than the current one. So I close this PR.

PR #122, opened 2015-03-11
Sale datetime confirm

bguillot:
Hi, 

This module changes the type of the field "date_confirm" from date to datetime. 
It doesn't break anything in the odoo core, and I checked quickly in OCA modules it seems ok too.

The "date_order" field was a date in version 7.0 and before, but it is a datetime now, so having a datetime also for the confirmation date seems interresting to me.
If you calculate the delivery date from the confirmation date instead of the order date you may want to have the hour too.

Regards.


gurneyalex:
I'll abstain on this one, as I'm not sure if it is useful or not, and despite what you say, there is a high potential for breaking stuff.

How did you check it does not break anything in the core? Did you check the various reports, PDF printouts... with this addon installed?

bguillot:
Thank you for the review! 

About the utility, the dates on pickings and moves are datetime, by default the min_date for exemple is computed from the "date_order", but if you want to compute from "date_confirm" like in those modules : https://github.com/akretion/openerp-stock-delivery-times, it's too bad to use a date when you can use a datetime, you lose some informations. 

About the checks, I checked the PDF but forgot the sale.report, thank you for pointing this out. 
I've just fixed it, and it works now (tested to group by confirmation date ...)

Cedric-Pigeon:
Hi,

I agree with @gurneyalex. It is dangerous to do that because it is impossible to be sure that nothing is broken in the core.

For example if Odoo add a sql view using this field, you will encounter some problems.
See issue https://github.com/odoo/odoo/issues/5605

Regards

gurneyalex:
@bguillot @florian-dacosta updated the OCA database. The CLA is ok.

StefanRijnhart:
:-1: for compatibility reasons indeed. It seems safer to add a new field instead of changing the field type of the existing field.

florian-dacosta:
@StefanRijnhart You are right, that is the solution we finally have chosen. 
Indeed, in case of update of the module sale, Odoo would try to put the field back to date type, and then to datetime type.
So, for huge database, it would take a long time to update, and possibly it would lose the time data of the field during the process.

We can close this PR.

gurneyalex:
closing as requested

StefanRijnhart:
Thanks for the feedback!

PR #123, opened 2015-03-16
Add sale_payment_term_interest

guewen:
This module allows to compute interest fees based on payment terms and
to add a line in the sales orders with the computed amount.

PR #125, opened 2015-03-16
Add sale_payment_term_interest

guewen:
This module allows to compute interest fees based on payment terms and
to add a line in the sales orders with the computed amount.


sale_payment_term_interest/model/account_payment_term.py
```diff
@@ -0,0 +1,89 @@
...
+        precision_model = self.env['decimal.precision']
+        prec = precision_model.precision_get('Account')
+        for line in self.line_ids:
+            if line.value == 'fixed':
+                line_amount = round(line.value_amount, precision_digits=prec)
+            elif line.value == 'procent':
+                line_amount = round(value * line.value_amount,
+                                    precision_digits=prec)
+            elif line.value == 'balance':
+                line_amount = round(amount, prec)
```
ghost, line 59: Why do you change the digit precision? It would seem to me more logical to let the precision dp.get_precision('Payment Term').

sale_payment_term_interest/tests/test_sale_order.py
```diff
@@ -0,0 +1,157 @@
...
+            'price_unit': 50,
+            'tax_id': [(6, 0, [tax.id])],
+        }
+        line2_values = {
+            'product_id': product2.id,
+            'product_uom_qty': 2,
+            'product_uom': product1.uom_id.id,
+            'price_unit': 100,
+            'tax_id': [(6, 0, [tax.id])],
+        }
```
ghost: The beggining of each test is almost the same here in the 3 cases. You could move what is repeated to setUp and then just write in your objects what changes from one unit test to the other.

ghost:
@guewen Added you a few comments. Otherwise, looks good to me.

sale_payment_term_interest/model/account_payment_term.py
guewen, line 59: This code is the same than `account_payment_term.compute()` in the core account module. Since I want interests based on the amounts due for each term, I have to use the exact same computation than the `compute()` method.

guewen:
> @guewen Added you a few comments. Otherwise, looks good to me.

Thanks for reviewing. I removed the repetition in the tests and answered you for the decimal precision, also added a comment in the code.

sale_payment_term_interest/README.rst
```diff
@@ -0,0 +1,54 @@
...
+ * Install the `sale_payment_term_interest` module is the only things to do.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+ * Go to *Invoicing > Configuration > Payment Terms*
+ * On the lines of the payment terms, you can configure an interest
+   rate. It is based on the number of days.
```
gurneyalex: I find this strange, but I'm don't know the general practice here. Most of the time I see an annual interest rate, which is divided by 12 to get a monthly rate (and then by 30 to get a daily rate if needed).

Is this meant to be used in a B2B or a B2C context?

sale_payment_term_interest/tests/test_payment_term.py
```diff
@@ -0,0 +1,61 @@
...
+            'value_amount': 0.3333,
+            'days': 10,
+            'interest_rate': 0.0,
+        })
+        self.env['account.payment.term.line'].create({
+            'payment_id': term.id,
+            'value': 'procent',
+            'value_amount': 0.3333,
+            'days': 40,
+            'interest_rate': 0.041666,
```
gurneyalex: for readability purpose, it would be nicer in my opinion to have this written as 15/(12*30) (and to add `from __future__ import division`) at the beginning of the file (and I still feel it would be more user friendly to have 15 for the value here).

sale_payment_term_interest/tests/test_payment_term.py
```diff
@@ -0,0 +1,61 @@
...
+            'payment_id': term.id,
+            'value': 'balance',
+            'days': 70,
+            'interest_rate': 0.041666,
+        })
+        terms = term.compute_interest(469.35, date_ref='2015-03-13')
+        self.assertEquals(len(terms), 3)
+        terms = sorted(terms, key=itemgetter(0))
+        line1, line2, line3 = terms
+        self.assertAlmostEqual(line1[1], 156.43)
```
gurneyalex: same as above: if you put in the mathematical expression instead of the result, you get a test that will be easier to maintain.

sale_payment_term_interest/README.rst
ghost: I agress with @gurnalex here. Interest is in most cases computed on an annual basis. But, dividing by 12 for monthly is not the way to do this, because interest is rarely compounded monthly.

guewen: Done, the rate is now on an annual basis.

sale_payment_term_interest/tests/test_payment_term.py
guewen: Done

sale_payment_term_interest/tests/test_payment_term.py
guewen: Done

ghost:
:+1:

yvaucher:
:+1:  it says in https://github.com/OCA/sale-workflow/pull/123 that this one superseeds it. However #123 is merged. Was it intentional?

guewen:
> it says in #123 that this one superseeds it. However #123 is merged. Was it intentional?

#123 has been reverted

yvaucher:
@guewen Ok thanks I proceed with the merge then

PR #126, opened 2015-03-17
[ADD] Module Reasons to export

ghost:
@nhomar Thanks for your comment.
I moved my PR from https://github.com/OCA/crm/pull/22 to sale-workflow.

I fixed the README.rst. Appologies, it was confusing.
This module adds a new model. Its purpose is to display a field on commercial invoices.


nhomar:
It looks good to me.

Some things in the code are a little out of "form" about how some declarations are commonly used, like the Import statement in several lines [1], and the declaration explicit in field of required arguments [2].
If you make things "like" other modules does, then in the future some migrations of code can be automated, with different approaches (not meaning they are wrong just not commonly accepted in this projects) it become a little more diffiult in the future.

If you change that for me it is a +1

Regards.

[1] https://github.com/OCA/sale-workflow/pull/126/files#diff-231c3fb1fa8613b82ec7476b6fb4e6fdR23
[2] https://github.com/OCA/sale-workflow/pull/126/files#diff-d41528ecaab417c2d729196d55315663R29

ghost:
@nhomar thanks! I fixed your suggestions.

nhomar:
Now ILGTM.

+1

sale_reason_to_export/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+#
+##############################################################################
+
+{
+    'name': 'Sale Reason to Export',
+    'version': '0.1',
+    'author': 'Jean-Philippe Jobin',
+    'maintainer': 'Savoir-faire Linux',
+    'website': ' http://www.savoirfairelinux.com',
+    'license': '',
```
foutoucour: Licence should be AGPL3 please.

sale_reason_to_export/views/sale_order_view.xml
```diff
@@ -0,0 +1,22 @@
...
+            <field name="name">sale.order.form</field>
+            <field name="model">sale.order</field>
+            <field name="inherit_id" ref="sale.view_order_form"/>
+            <field name="arch" type="xml">
+                <group name="sale_total" position="after">
+                    <group col="4">
+                        <group>
+                            <field name="reason_to_export_id"/>
+                        </group>
+                        <group></group>
```
foutoucour, line 16: why do you have an empty group here?
Could you add a comment to explain this if really needed?

foutoucour:
Minor updates would make it perfect for me.

:+1: good to go anyway

PR #128, opened 2015-03-18
use ownership for exceptions, if any

lepistone:
Ownership is introduced in module sale_owner_stock_sourcing. It is not a
dependency, and the field is used if found.

This needs a fix to https://github.com/odoo/odoo/issues/5814 to work
correctly.

### DIFF
--- a/sale_exception_nostock/__openerp__.py
+++ b/sale_exception_nostock/__openerp__.py
@@ -19,7 +19,7 @@
 #
 #
 {'name': 'Sale stock exception',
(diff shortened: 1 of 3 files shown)


yvaucher:
odoo/odoo#5814 has been merged at https://github.com/odoo/odoo/commit/01f1d91006fcf8c1a3cb85e9a8e4754f07151e1b

:+1:

jgrandguillaume:
:+1:

PR #134, opened 2015-03-30
[WIP] Add sale_amendment

guewen:
**Work in Progress**

Depends on
- [x] https://github.com/odoo/odoo/pull/6036
- [x] ~~OCA/OCB#184~~
- [x] https://github.com/OCA/stock-logistics-workflow/pull/94

### DIFF
--- a/sale_amendment/view/sale_order.xml
+++ b/sale_amendment/view/sale_order.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+  <data>
+  </data>
+</openerp>
--- a/sale_amendment/view/sale_order.xml
+++ b/sale_amendment/view/sale_order.xml
@@ -1,5 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
   <data>
+
+    <record id="view_order_form" model="ir.ui.view">
+      <field name="name">sale.order.form.amendment</field>
+      <field name="model">sale.order</field>
+      <field name="inherit_id" ref="sale_stock.view_order_form_inherit"/>
+      <field name="arch" type="xml">
+         <button name="ship_recreate">
+           <button name="button_amend" states="shipping_except"
+             type="object" string="Amend Order"/>
+         </button>
+      </field>
+    </record>
   </data>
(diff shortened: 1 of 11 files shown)


sale_amendment/view/sale_order.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+  <data>
+
+    <record id="view_order_form" model="ir.ui.view">
+      <field name="name">sale.order.form.amendment</field>
+      <field name="model">sale.order</field>
+      <field name="inherit_id" ref="sale_stock.view_order_form_inherit"/>
+      <field name="arch" type="xml">
+         <button name="ship_recreate">
```
gurneyalex: relying on the default value of position... :frowning: explicit is better that explicit IMO (and I never remember that default value...)

yvaucher: Default position is `inside` so there is a :bug: here

guewen: Oops, it was not intentional. I would prefer an error than a default value in such situation.

legalsylvain:
Hi @guewen,
Is this PR still in WIP ?
Regards.

guewen:
@legalsylvain I need to check with @gurneyalex or @lepistone (both on holidays for now)

Cedric-Pigeon:
Hi @gewen, I made some functionnal tests on this PR. It is impossible to amend a sale order which has an invoice. What about an advance invoice which has no link with sale order lines ?
What is your opinion ?

guewen:
> Hi @gewen, I made some functionnal tests on this PR. It is impossible to amend a sale order which has an invoice. What about an advance invoice which has no link with sale order lines ?
> What is your opinion ?

Hi, thanks for your tests! I'm not sure this use case was meant to be supported. @jgrandguillaume may have more thought about that.

Cedric-Pigeon:
Hi @guewen I made a PR on your original branch. Regards.

pedrobaeza:
Is this going to be continued?

pedrobaeza:
It seems no, as 9.0 includes amendment  in core.

PR #138, opened 2015-04-09
7.0 sale stock relation line

chafique-delli:
add module sale_stock_relation_line that creates link between parent lines orders and child in stock move for bundle product

PR #139, opened 2015-04-09
7.0 sale invoice relation line

chafique-delli:
add module sale_invoice_relation_line that creates links between parent lines orders and child in invoice line for bundle product

This Pull Request depends of the pull request 'OCA/sale-workflow#138'.


sale_invoice_relation_line/account_invoice.py
```diff
@@ -0,0 +1,58 @@
...
+    _inherit = 'account.invoice.line'
+    _order = 'sale_parent_line_id, name desc'
+
+    def _get_invoice_line_from_order_line(self, cr, uid, ids, context=None):
+        invoice_line_ids = self.pool['account.invoice.line'].search(
+            cr, uid, [('sale_line_id', 'in', ids)], context=context)
+        return invoice_line_ids
+
+    _columns = {
+        'sale_line_id': fields.many2one(
```
eLBati: Relation between invoice lines and order lines already exists (many2many)
https://github.com/odoo/odoo/blob/7.0/addons/sale/sale.py#L685

If the purpose is to make the related SO lines available on invoice line model, I would suggest to add the inverse many2many, using the existing table `sale_order_line_invoice_rel`

pedrobaeza:
And the name is not very suitable: you are adding the relation in invoice, so it must be `account_invoice_line_sale_link` or something similar.

eLBati:
@pedrobaeza as `sale_order_line_invoice_rel` already exists, if you want to add to `account.invoice.line` the inverse of `invoice_lines` (calling it say `sale_line_ids`) so that it will be set when `invoice_lines` is set, you should use the same relation table used for `invoice_lines` in `sale.order.line` model: `sale_order_line_invoice_rel`

pedrobaeza:
@eLBati, sorry, I was not too clear: I mean the module name, not the column itself or the relation.

pedrobaeza:
And thinking about it, the target should be account-invoicing repo, not this one.

eLBati:
ah ok @pedrobaeza I agree

chafique-delli:
@pedrobaeza , I renamed the module name as you've suggested me but the repository is 'sale-workflow'    because everything starts sale order line.

pedrobaeza:
I don't agree. The link is on the invoice, not on the sale, so name must be the one I have said, not the reverse.

:-1:

chafique-delli:
@pedrobaeza , So, according to your reasoning, the module 'stock_picking_invoice_link' has nothing to do in the repository 'stock-logistics-workflow'  ?

pedrobaeza:
Well, that part can be debateable, because someone can look for the module in the repos following different reasonings, but the name for me is clear.

chafique-delli:
P.R. moved to https://github.com/OCA/account-invoicing/pull/69

PR #141, opened 2015-04-23
[REF] sale_quotation_sourcing and sale_sourced_by_line

yvaucher:
**sale_procurement_group_by_line**

This module extract logic to create multiple procurement group from `sale_sourced_by_line`
for a single sale order for grouped sale order lines.

**sale_quotation_sourcing**

Now depends on `sale_procurement_group_by_line` to create procurement group by purchase order.

Procurment orders are named after their purpose. 
- in case of `sale_sourced_by_line` warehouse name is included in name.
- in case of `sale_quotaiton_sourcing` PO name is included in name.

Compatibility between `sale_quotation_sourcing` and `sale_sourced_by_line` is enhanced by giving priority on the type of procurement group. If both modules are installed it will create first procurement group by PO and if no PO are selected, it create procurement group by warehouse.


yvaucher:
Fixed few PEP8 errors

jgrandguillaume:
Hi,

Thanks for the work ! That's fine for me :+1: 

Regards,

gurneyalex:
:+1:

PR #144, opened 2015-04-30
Migration of sale_line_description

tafaRU:


pedrobaeza:
I'm very strict with the names, hehe: I prefer the complete name _sale_order_line_description_

pedrobaeza:
The rest :+1:

yvaucher:
:+1:

yvaucher:
Travis is failing on 

TypeError: product_id_change() got an unexpected keyword argument 'product'

On module sale_validity

I think both modules changes product_id_change.

Consider using web_context_tunnel

eLBati:
@pedrobaeza @tafaRU in these cases I would avoid renaming modules because this would implies changes to `ir.model.data` table. Thus it would require a migration script. Maybe it isn't worth it. What do you think?

pedrobaeza:
In this case, there is no too much problem:
- There are no associated data, just functionality. A migrated database can uninstall/install the module.
- It's a switch of version, not a new revision.
- We can handle this renaming in OpenUpgrade easily (as I did with several modules).

guewen:
> There are no associated data, just functionality. A migrated database can uninstall/install the module.

You can't uninstall if the old version is not there.

pedrobaeza:
I mean: uninstall in v7, install in v8.

eLBati:
On 4 May 2015 at 12:09, Pedro M. Baeza notifications@github.com wrote:

> In this case, there is no too much problem:
> - There are no associated data, just functionality. A migrated
>   database can uninstall/install the module.
> 
> Actually, 'group_use_product_description_per_so_line' is created by the
> module and used at
> https://github.com/tafaRU/sale-workflow/blob/8.0-sale_line_description-migr/sale_line_description/sale.py#L45

In this case, AFAICS, users should be manually re-added to the group

pedrobaeza:
You're right with the manual option. OpenUpgrade renaming option should handle it without problem.

pedrobaeza:
No, directly including the rename module in one dictionary, OpenUpgrade rename all corresponding stuff (ir_module_module, xml-IDs, and so on).

pedrobaeza:
Here you have the renamed modules for v8 (including some from OCA):

https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/openerp/addons/openupgrade_records/lib/apriori.py#L5

Daniel-CA:
@tafaRU The travis error is because you are using `product_id` as keyword in `product_id_change` function, and it has been renamed to `product` in the sale module.

tafaRU:
@Daniel-CA I should have fixed, thanks for your hint!

And I also have renamed the module so you can proceed with PR containing only changes about new API and close https://github.com/OCA/sale-workflow/pull/151.

tafaRU:
PR looks green and all comments have been addressed.
Reviews are welcomed :wink:

Daniel-CA:
Hello,
Can someone review/merge this module? It looks fine to me.

lmignon:
:+1: Code review only

pedrobaeza:
:+1:

PR #149, opened 2015-05-06
[FIX] Bypass of exceptions through the sent state

mistotebe:
It is possible to skip sale exceptions checking if sending an email first, then confirming.


yvaucher:
:+1:

jgrandguillaume:
:+1:

PR #152, opened 2015-05-07
 [ADD] 'sale_discount_limit'

JordiBForgeFlow:
This module was created to extend the sales process. It allows to
set a maximum permitted sales discount % on each product on a sale quotation.

A sales user cannot approve a sales quotation if any of the items contain a
sales discount % above the maximum allowed.

A user can manually initiate the maximum discount limit check when the quote
is in status draft or sent.

Users belonging to the group 'Sales Quotation Discount Block Releaser' will
have the permission to override this rule and approve the sales quotation.

### DIFF
--- a/sale_discount_limit/README.rst
+++ b/sale_discount_limit/README.rst
@@ -1,0 +1,59 @@
+Sale Discount Limit
+===================
+
+This module was created to extend the sales process. It allows to
+set a maximum permitted sales discount % on each product on a sale quotation.
+
+A sales user cannot approve a sales quotation if any of the items contain a
+sales discount % above the maximum allowed.
+
+A user can manually initiate the maximum discount limit check when the quote
+is in status draft or sent.
+
+Users belonging to the group 'Sales Quotation Discount Block Releaser' will
+have the permission to override this rule and approve the sales quotation.
+
+
+Installation
+============
+
+No specific installation steps are required.
+
+Configuration
+=============
+
+No specific configuration steps are required.
+
+Usage
+=====
+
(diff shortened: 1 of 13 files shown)


jgrandguillaume:
Hi Jordi,

Thanks for the contrib ! Can I ask why you didn't use this module as a base: "https://github.com/OCA/sale-workflow/tree/8.0/sale_exceptions" 

Sale_exception is trying to address the management of all kind of things and checks that need to be performed on a SO and raise it at SO confirmation. This way, you have a summary of all "troubles" at once regarding your order and you can fix them before going further. Exception can be by-passed if needed.

My opinion is that it would be great to have all kind of check on SO confirmation based on this module so we have a unified way to deal with them within the OCA.

What do you think ?

Regards,

Joël

JordiBForgeFlow:
OMG! I didn't check this module when I looked for something already done. I will will then build it on top of sale_exceptions. I'm now closing this PR, and will submit a new one.

Thanks Joël!!! 
Jordi.

jjscarafia:
Jordi,
We also have this module that has been working in many costumer's in 6.1, 7.0 and now in 8.0
It also adds a boolean on products to allow users to change prices on those products (like shipment)
https://github.com/ingadhoc/odoo-addons/tree/8.0/price_security
If you find it useful we can clean the code andake a PR
But perhaps its better what Joel suggests
Thanks!

JordiBForgeFlow:
@jgrandguillaume I have refactored the module, to work on top of 'sale_exceptions'. Thanks for your feedback!!

PR #160, opened 2015-05-21
Automatically merge all POs from a single order

mdietrichc2c:
This PR is to add an automatic grouping functionality for POs generated from the same sale order. Otherwise, as a customer reported, the multiple POs will lead to multiple packages from the supplier to the customer.

The functionality is mostly based on the `do_merge` method in `purchase_group_hooks`; an "origin" criteria is added to only merge POs from the same SO, and the procurements are all updated with the new PO.


mdietrichc2c:
I added a method in order to define which POs/procurements should be merged automatically.

mdietrichc2c:
Added as well: the `sale_flow` field (since otherwise, it will be set as normal)

pedrobaeza:
@gurneyalex this still applies?

pedrobaeza:
Closing as no answer. Feel free to reopen if needed.

PR #146, opened 2015-05-05
Migration of sale_line_quantity_properties_based

tafaRU:
This PR depends on https://github.com/OCA/sale-workflow/pull/147 which also depends on https://github.com/OCA/sale-workflow/pull/167 (both merged)

PR #145, opened 2015-05-05
Migration of sale_line_price_properties_based

tafaRU:
This PR depends on https://github.com/OCA/sale-workflow/pull/147 which also depends on https://github.com/OCA/sale-workflow/pull/167 (both merged)

PR #167, opened 2015-06-19
[8.0] Migration of sale_properties_easy_creation

tafaRU:


yvaucher:
:+1:

sale_properties_easy_creation/mrp_property.py
```diff
@@ -0,0 +1,48 @@
...
+    _columns = {
+        'value': fields.char('Value', size=64),
+    }
+
+    def name_create(self, cr, uid, name, context=None):
+        """
+        This allows the user to digit 'width 0.5' and the system will
+        automatically create a property of group 'width' with value '0.5'
+        """
+        splitted_name = name.split(' ')
```
gurneyalex: I suggest using `name.split()` here, as this will properly handle multiple spaces, tabs, etc.

sale_properties_easy_creation/mrp_property_formula.py
```diff
@@ -0,0 +1,30 @@
...
+##############################################################################
+
+from openerp.osv import orm, fields
+
+
+class MrpPropertyFormula(orm.Model):
+    _name = 'mrp.property.formula'
+    _columns = {
+        'name': fields.char('Name', size=128),
+        'formula_text': fields.text('Formula'),
```
gurneyalex: I think that the code to evaluate the formula should be in a helper method in this class  to avoid duplication. 

Otherwise we are going to have to watch for things like https://github.com/OCA/sale-workflow/pull/146/files#diff-ef08d502d2e08be1274eeab2f6f78114R64 in many places.

tafaRU:
@gurneyalex thanks for your hint. I temporarily set the PR as WIP.

tafaRU:
@gurneyalex, your remarks have been addressed, could you please have a look at it?

Thank you in advance.

gurneyalex:
Tried to test the module on runbot, and IMO the README.rst could be improved a little bit by mentionning the ir.config to activate to get access to the proper settings, and such. 

Also the formula_txt field should provide a help to describe how to write a formula (which variables are available etc. )

Finally there is no call to compute_formula in there. Is this intended?

tafaRU:
> Tried to test the module on runbot, and IMO the README.rst could be improved a little bit by mentionning the ir.config to activate to get access to the proper settings, and such.

https://github.com/tafaRU/sale-workflow/commit/b3503ab0e0a709094760c8da5c3f5d05459f9c75

> Also the formula_txt field should provide a help to describe how to write a formula (which variables are available etc. )

Could be further improved, but:
https://github.com/tafaRU/sale-workflow/commit/9a2cd8fd8ac417f738d9bccf053130c8309d834b

> Finally there is no call to compute_formula in there. Is this intended?

Yes, this is intended.
The calls to compute_formula are in the module sale_line_quantity_properties_based and the module sale_line_price_properties_based (See https://github.com/OCA/sale-workflow/pull/145 and https://github.com/OCA/sale-workflow/pull/146 still set as WIP since they depend on this one).

gurneyalex:
:+1:

PR #169, opened 2015-06-22
[ADD] New module <sale_last_price_info>

oihane:


sale_last_price_info/README.rst
```diff
@@ -0,0 +1,42 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product Last Price Info - Sale
+==================================
```
pedrobaeza: Put the same length

pedrobaeza:
Minor remark. The rest :+1:

jholze:
Great feature. Good work. :+1:

oihane:
Changes done

oihane:
Runbot is failing because other module not related to this.

oihane:
@pedrobaeza there is a coveralls error that I do not understand, it seems that it decreases because the module does not have a test, but it only adds some computed fields so I do not see why will it be required to have tests.

pedrobaeza:
As the code for the computed fields is not being executed, the coverage has decreased. You can make some little test that compare the expected result with the actual value: create a sale order, and test if the fields contains the correct data.

sale_last_price_info/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+            key=lambda l: l.order_id.date_order, reverse=True)
+        self.last_sale_date = lines[:1].order_id.date_order
+        self.last_sale_price = lines[:1].price_unit
+        self.last_customer = lines[:1].order_id.partner_id
+
+    last_sale_price = fields.Float(
+        string='Last Sale Price', compute='_get_last_sale')
+    last_sale_date = fields.Date(
+        string='Last Sale Date', compute='_get_last_sale')
+    last_customer = fields.Many2one(
```
guewen: Should be `last_customer_id`

sale_last_price_info/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.one
+    def _get_last_sale(self):
+        """ Get last sale price, last sale date and last customer """
+        lines = self.env['sale.order.line'].search(
+            [('product_id', '=', self.id),
+             ('state', 'in', ['confirmed', 'done'])]).sorted(
+            key=lambda l: l.order_id.date_order, reverse=True)
```
guewen, line 18: With thousands of sale lines it will start to be very ineffective I think. A direct `cr.execute()` followed by a `self.env['sale.order.line'].search([('id', 'in', ids)]` to apply the security rules could be something to consider.

sale_last_price_info/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+    "name": "Product Last Price Info - Sale",
+    "version": "1.0",
+    "category": "Hidden/Dependency",
+    "license": "AGPL-3",
+    "author": "OdooMRP team, "
+              "AvanzOSC, "
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Odoo Community Association (OCA)",
+    "website": "http://www.odoomrp.com",
+    "contributors": [
```
guewen, line 29: There is no `contributors` key in `__openerp__.py` usually we put this info in the readme (and it is already).

sale_last_price_info/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see http://www.gnu.org/licenses/.
+#
+##############################################################################
+
+{
+    "name": "Product Last Price Info - Sale",
+    "version": "1.0",
+    "category": "Hidden/Dependency",
```
guewen: Why it is `Hidden`? Shouldn't it be in sale?

sale_last_price_info/models/product.py
pedrobaeza: No, we use this naming convention because we prefer it over the `_id` one. Code without these suffixes is clearer, and there's no too much advantage in using them (the type can be easily inspected in lot of places).

sale_last_price_info/models/product.py
pedrobaeza, line 18: We don't need SQL, but have it an stored related field of the date, and use `order=` keyword on search and `limit=1`

sale_last_price_info/models/product.py
guewen: @pedrobaeza 

this is not what is in the OCA guidelines
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field

nor in the odoo guidelines
https://github.com/odoo/odoo/blob/8.0/doc/reference/guidelines.rst#symbols

sale_last_price_info/__openerp__.py
pedrobaeza, line 29: No, that's not true. Contributors field exists here: https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/module/module.py#L272 and is parsed here: https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/module/module.py#L586

sale_last_price_info/models/product.py
pedrobaeza: That OCA guidelines haven't been agreed, because I strongly opposed to that in two times that it has been treated in mailing lists, so I'm surprised to find this already merged.

guewen: Anyway I think we don't have very much choice here, it would be counter-productive to have a different guideline than the odoo one regarding the naming of fields.

pedrobaeza: Odoo already do it in a lot of places: 
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L213
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L241
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L242
and that's only on one model (sale.order), so I don't think it's a big deal.

guewen: You link to very old and recognized as badly named fields ;-)

Since then, they agreed to use the suffxes. You can look at https://github.com/odoo/odoo/pull/5156#issuecomment-73486314 which is the most recent discussion on that.

pedrobaeza: Well, I don't recognise that user, but he is saying:

> Same for the variables name. Especially since the new api. If we have a record or recordset, don't name the variable with _id, when I see a _id in a variable name, I expect a number, not a record

So I read exactly the opposite, and why variable names are going to be different in the behaviour from fields if you are using the same notation for accesing both? I see here clearly an inconsistency

guewen: Pedro, please read again.
You don't recognize this user, so, but the PR has still been accepted.

> It should already be the case but every now and then I happen to see some misleading fields/variable names.
> 
> One2Many and Many2Many fields should always have _ids as suffix (example: sale_order_line_ids)
> Many2One fields should have _id as suffix (partner_id, user_id, ...)

Isn't this clear enough?

> Same for the variables name. Especially since the new api. If we have a record or recordset, don't name the variable with _id, when I see a _id in a variable name, I expect a number, not a record

**variables name**.
And we do the same.
Example:

``` pyhon
partners = record.partner_ids
```

guewen: Eventually if you don't agree, please open a PR to change that on Odoo and on the OCA.

pedrobaeza: That's why I'm saying that it's not consistent between variable names and field names.

I will open a PR, but you can see in every sentence the word `should`, not `must`, so are you going to block this? All my modules use this convention, and I don't want to change it because it's a huge work and as I stated, I don't see an advantage.

gurneyalex: @pedrobaeza so yes it says "should" but this is not an RFC. I see no good reason to not follow the rule in this specific case. The "should" is there to allow some exception such as `order_lines` (with which I'm not especially happy because it requires me to remember them, but they are here and I have to live with it)

I'm :-1: for the PR in its current form. 

If we start not treating the field naming conventions seriously we are going to end up with all sort of unmanageable non standardized names, lost time and the absolute impossibility to develop addons without an IDE with smart context aware name completion, which I consider a huge regression.

And the field naming convention in my opinion is far more important than blank lines and 79 char long code lines.

sale_last_price_info/__openerp__.py
oihane: @guewen, it was at first an autoinstall module, so I made it `Hidden`
What should it be exactly, `Sales Management`?

guewen: ok. Yes `Sales Management`

oihane:
@guewen @gurneyalex could you please review again...

sale_last_price_info/models/product.py
nhomar: 2015-06-26 0:49 GMT-05:00 Pedro M. Baeza notifications@github.com:

> No, we use this naming convention because we prefer it over the _id one.
> Code without these suffixes is clearer, and there's no too much advantage
> in using them (the type can be easily inspected in lot of places).

I think _id have more sense for many2one name of fields (even if yes it is
an object now), introduce new "conventions" becomes thing difficult to
inspect with pure greps +1 for the _id.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

nhomar:
I see all changes asked to be done are applied can somebody give us a +1 to proceed the merge.?

nhomar:
For me it is a +1 BTW thanks for the contribution.

PR #171, opened 2015-06-25
7.0 move sale payment method

bguillot:
This PR moves the module sale_payment_method from e-commerce repo.

See https://github.com/OCA/e-commerce/issues/47 for the reason and discussions about that.


pedrobaeza:
Although we talk about moving this module here, I think we can close this one for not touching in this very old version, and make the move only since 8.0.

PR #173, opened 2015-06-30
[FIX] sale_order_revision: Added German translation

maljac:
Added a German translation for sale_order_revision


pedrobaeza:
Technically :+1:

tafaRU:
From a technical point of view :+1:

PR #140, opened 2015-04-20
[ADD] Add sale order lot selection

hurrinico:
This Module Allows you to select on sale order the production lot that will be delivered

PR #175, opened 2015-07-02
8.0 sale prodlot

Viggor:
[WIP] Add lot number on sale order line and transmitting it in the workflow.


pedrobaeza:
This method is not valid anymore as is, as v9 WMS has been totally reworked. There's also already a PR for the same purpose: https://github.com/OCA/sale-workflow/pull/140. Please review this one.

Viggor:
Thanks I will take a look at this PR.

PR #181, opened 2015-07-07
Move sales related modules (with history) from OCA/department

dreispt:
As described in https://github.com/OCA/department/issues/40.


dreispt:
I'll need to have a better look at the history...

PR #183, opened 2015-07-15
[MIG] migrate 'stock_picking_reorder_lines' to v8

damdam-s:


gurneyalex:
:+1: thanks for porting

pedrobaeza:
This should be in stock-logistics-workflow repo

damdam-s:
@pedrobaeza : this module is in this project in version 7.0. That's why I migrate them in this repo. Now if it doesn't fit the "good" OCA project, I can move it like you suggested.

pedrobaeza:
Yeah, I think so, because it's more attached to that one, don't you think?

damdam-s:
new PR for this module : https://github.com/OCA/stock-logistics-workflow/pull/139

PR #185, opened 2015-07-16
[FIX] cancel button of original view should be hide after installation…

dhp-denero:
… of sale_cancel_reason module

### DIFF
--- a/sale_cancel_reason/view/sale_view.xml
+++ b/sale_cancel_reason/view/sale_view.xml
@@ -13,7 +13,7 @@
                     <button name="%(action_sale_order_cancel)d" states="manual,progress" string="Cancel Order" type="action" groups="base.group_user"/>
                 </xpath>
                 <xpath expr="//button[@name='cancel']" position="attributes">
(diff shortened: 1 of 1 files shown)


dhp-denero:
Hello,

I already signed CLA before. Let me know what do I need to now ?

pedrobaeza:
Why your change? Isn't working the other way?

dhp-denero:
I just correct the xml. 

with line <attribute name="invisible" eval="1"/>, standard Cancel Button is not hiding. So I corrected it with <attribute name="invisible">True</attribute>. It is working well.

pedrobaeza:
:+1: then

dhp-denero:
means ?

I think, the correction should be merged with original branch.

pedrobaeza:
Yeah, but we need at least 3 approvals or 2 and more than 5 working days of delay

yvaucher:
:+1: Thanks

dhp-denero:
Hey thanks Yannick.

Issue #184, opened 2015-07-15
sale_order_revision error on "Confirm Sale"


thomaspaulb:
When you 
- Cancel a quotation
- Create a new copy of Quotation
- Confirm Sale

It says "You cannot confirm a sales order which has no line."

But many lines are there.
I will investigate further, just wish to know if other people have had this problem.

thomaspaulb:
See pull request #186

rafaelbn:
Hi, I thinks this issue can be closed as exists PR. @antiflu could you close issue? Thanks

PR #186, opened 2015-07-17
[FIX] sale_order_revision #184

thomaspaulb:
Fix: set new order lines to draft and old ones to cancel.

Rationale: the way this module creates a new revision is that the old, cancelled revision is renamed to the new revision, then a copy is created and named after the old revision. As a result the lines of the old revision (copy) have state "draft" (the default) and the lines of the new revision have state "cancel". It should be the other way around. So, this fix switches the states around.

### DIFF
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -55,10 +55,6 @@ class sale_order(models.Model):
     def copy_quotation(self):
         self.ensure_one()
 
-        # store existing procurement group id, unset it on new revision
-        procurement_group_id = self.procurement_group_id.id
-        self.write({'procurement_group_id': None})
-
         revision_self = self.with_context(new_sale_revision=True)
         action = super(sale_order, revision_self).copy_quotation()
         old_revision = self.browse(action['res_id'])
@@ -73,9 +69,6 @@ class sale_order(models.Model):
         msg = _('New revision created: %s') % self.name
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
-
-        # set stored procurement group id on old order
-        old_revision.write({'procurement_group_id': procurement_group_id})
 
         # swap order lines of old and new order
         so_line = self.env['sale.order.line']
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -54,6 +54,11 @@ class sale_order(models.Model):
     @api.multi
     def copy_quotation(self):
         self.ensure_one()
+
+        # store existing procurement group id, unset it on new revision
+        procurement_group_id = self.procurement_group_id.id
+        self.write({'procurement_group_id': None})
+
         revision_self = self.with_context(new_sale_revision=True)
         action = super(sale_order, revision_self).copy_quotation()
         old_revision = self.browse(action['res_id'])
@@ -69,15 +74,15 @@ class sale_order(models.Model):
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
 
-        # cancel order lines for old revision
-        line = self.pool.get('sale.order.line')
-        line.write(self._cr, self._uid,
-                   [line.id for line in old_revision.order_line],
-                   {'state': 'cancel'}, context=self._context)
-        # for current revision, set them to draft
-        line.write(self._cr, self._uid,
-                   [line.id for line in self.order_line],
-                   {'state': 'draft'}, context=self._context)
+        # set stored procurement group id on old order
+        old_revision.write({'procurement_group_id': procurement_group_id})
+
+        # swap order lines of old and new order
+        so_line = self.env['sale.order.line']
+        old_lines = so_line.browse(old_revision.order_line.ids)
+        new_lines = so_line.browse(self.order_line.ids)
+        old_lines.write({'order_id': self.id})
+        new_lines.write({'order_id': old_revision.id})
 
         return action
 
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -79,8 +79,8 @@ class sale_order(models.Model):
 
         # swap order lines of old and new order
         so_line = self.env['sale.order.line']
-        old_lines = so_line.browse(old_revision.order_line.ids)
-        new_lines = so_line.browse(self.order_line.ids)
+        old_lines = old_revision.order_line
+        new_lines = self.order_line
         old_lines.write({'order_id': self.id})
         new_lines.write({'order_id': old_revision.id})
 
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -68,6 +68,17 @@ class sale_order(models.Model):
         msg = _('New revision created: %s') % self.name
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
+
+        # cancel order lines for old revision
+        line = self.pool.get('sale.order.line')
+        line.write(self._cr, self._uid,
+                   [line.id for line in old_revision.order_line],
+                   {'state': 'cancel'}, context=self._context)
+        # for current revision, set them to draft
+        line.write(self._cr, self._uid,
+                   [line.id for line in self.order_line],
+                   {'state': 'draft'}, context=self._context)
+
         return action
 
     @api.returns('self', lambda value: value.id)
--- a/sale_order_revision/test/sale_order.yml
+++ b/sale_order_revision/test/sale_order.yml
@@ -3,9 +3,15 @@
 -
   !record {model: sale.order, id: sale_order_1}:
     partner_id: base.res_partner_2
+    partner_invoice_id: base.res_partner_2
+    partner_shipping_id: base.res_partner_2
     order_line:
       - product_id: product.product_product_15
         product_uom_qty: 15.0
+-
+  I confirm the SO
+-
+  !workflow {model: sale.order, action: order_confirm, ref: sale_order_1}
 -
   I cancel the SO
 -
@@ -24,5 +30,6 @@
      assert new_so.name.endswith('-01')
      assert not new_so.order_line[0].procurement_ids
      assert all(line.state == 'draft' for line in new_so.order_line)
+     assert len(new_so.procurement_group_id.ids) == 0
      old_so = new_so.old_revision_ids[0]
      assert all(line.state == 'cancel' for line in old_so.order_line)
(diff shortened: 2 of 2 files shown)


eLBati:
Thanks
Please see travis

```
./sale_order_revision/model/sale_order.py:69:68: W291 trailing whitespace
./sale_order_revision/model/sale_order.py:70:13: E128 continuation line under-indented for visual indent
./sale_order_revision/model/sale_order.py:70:59: W291 trailing whitespace
./sale_order_revision/model/sale_order.py:74:13: E128 continuation line under-indented for visual indent
```

Also, could you extend the automated test adding the failing test case?

thomaspaulb:
Ok I think it is better now, how can I run the unit test to check?

petrus-v:
Hi @antiflu,

To run yaml test locally you can use `--test-enable` option of your odoo server script.
:+1: for the fix

thomaspaulb:
So yea we found another bug with the procurement/shipping, also to do with order lines being swapped... so I changed the fix to really swap the lines between orders, not only toggle the line state. Should be fine now, but we're going to production-test it again. 

@petrus-v I managed to successfully run the tests, so they are OK as well. ("tested" the tests...)

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,20 @@ def copy_quotation(self):
         msg = _('New revision created: %s') % self.name
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
+
+        # put existing procurement group id on old order
+        old_revision.write({'procurement_group_id': procurement_group_id})
+        self.write({'procurement_group_id': None})
```
pedrobaeza: Why do you write this 2 times?

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,20 @@ def copy_quotation(self):
...
+
+        # swap order lines of old and new order
+        old_line_ids = [line.id for line in old_revision.order_line]
```
pedrobaeza: Why don't `old_revision.order_line.ids`?

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,20 @@ def copy_quotation(self):
...
+        new_line_ids = [line.id for line in self.order_line]
```
pedrobaeza: The same: `self.order_line.ids`

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,20 @@ def copy_quotation(self):
...
+        so_line = self.pool.get('sale.order.line')
```
pedrobaeza: Why not new API?

pedrobaeza:
Can you please check also the procurement group in the tests?

sale_order_revision/model/sale_order.py
thomaspaulb: How to write it in one?

pedrobaeza: I mean that you have the same sentence in line 60 and in line 75

thomaspaulb:
Implemented your suggestions in the first commit.

Second commit contains the procurement group test... however:
- procurement group creation requires order_policy = picking
- order_policy = picking requires sale_stock to be listed as dependency
- rest of the module does not necessarily have to depend on sale_stock

This is a logical conundrum that I have run into before, maybe you know the answer - how to make use of a module _when it's there_, but otherwise leave it be? Or in this case, how to extend a test to test cases where there is dependency on a certain module, but not make the whole module dependent? Only answer I have found to that so far is to make a new submodule `sale_order_revision_stock` that contains the stock dependencies...

nhomar:
@antiflu 

Just add to your python class for test:

``` python
@openerp.tests.common.at_install(False)
@openerp.tests.common.post_install(True)
```

On this way your module will be tested when everything finished AFAICU.

But frequently if it is such case is because you have a design issue that need some other workaround more efective (may be a glue module which test the case dependant).

regards.

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,17 @@ def copy_quotation(self):
...
+
+        # set stored procurement group id on old order
+        old_revision.write({'procurement_group_id': procurement_group_id})
+
+        # swap order lines of old and new order
+        so_line = self.env['sale.order.line']
+        old_lines = so_line.browse(old_revision.order_line.ids)
+        new_lines = so_line.browse(self.order_line.ids)
+        old_lines.write({'order_id': self.id})
+        new_lines.write({'order_id': old_revision.id})
```
eLBati, line 78: You don't need `old_lines` and `new_lines` variables. You can call `old_revision.order_line.write` and `self.order_line.write`

eLBati:
@antiflu @pedrobaeza in this case, I would just remove lines
https://github.com/OCA/sale-workflow/pull/186/files#diff-7afbf0b5a2943e16e7a3935595fdf635R8
and
https://github.com/OCA/sale-workflow/pull/186/files#diff-7afbf0b5a2943e16e7a3935595fdf635R35
from the test

I would not make this module depending on sale_stock just for tests purpose, as its functionalities can also be used without sale_stock.
Possibly, we could create an hidden module `sale_order_revision_stock_test`, containing the missing test only

sale_order_revision/model/sale_order.py
thomaspaulb, line 78: I wanted to do that, but I was afraid that after the old_revision.order_line.write action, the new lines will now belong to the old revision and thus self.order_line.ids will contain no items and the second write action will be flawed.

thomaspaulb:
Sounds interesting. Can a hidden module be in a subfolder of the normal module?

sale_order_revision/model/sale_order.py
eLBati, line 78: > I wanted to do that, but I was afraid that after the
> old_revision.order_line.write action, the new lines will now belong to the
> old revision and thus self.order_line.ids will contain no items and the
> second write action will be flawed.
> 
> Did this actually happen?

pablocm-aserti:
Maybe I'm missing something, but sometimes it is neccessary to take 2 steps back to see the whole picture.
What is the reason to do a duplicate, disguise it as the old SO and rename the original one to look like a new revision?
Wouldn't it be easier and better for maintainability to just use the duplicate as the new revision with fresh new lines?

As I see it, the cancel status problem and the procurement management needed to make it work come from that decision.
Furthermore, this could lead to data inconsistencies for third-party modules that add objects pointing to those lines, since they are now in another SO.

thomaspaulb:
I think that before, the module worked like you described, but then someone
changed it around, and the problems that I fixed arose.

The reason I did not change it back is that I thought someone might have
had a reason for doing it this way. A reason I could think of is that when
there is somewhere in some module a reference to an older revision, and
there comes a new revision, in this way the references will automatically
move to the new revision. Instead, if it is done your way, the whole
database needs to be scanned for references to the old revision and updated
to the id of the new revision. Which would not be desirable.

Does that make sense to you?

pablocm-aserti:
> A reason I could think of is that when there is somewhere in some module a reference to an older revision

But third-party modules -including procurement- are not aware of any revision system, just the standard copy feature. They can decide what to do with the the fields/related objects they add -whether carry them over to the new duplicate or not, do a deep copy, etc-. If you do a duplicate and then switch the lines, they know nothing of it and then the responsibility for fixing consistency relies on this module -eg. procurements- or "link modules" to fix things when you use revisions.

> Instead, if it is done your way, the whole database needs to be scanned for references to the old revision and updated to the id of the new revision. Which would not be desirable.

Probably the most desired behaviour is not to change references but let them be, and then let the brand new copy follow its workflow from the beginning, regenerating all those related objects: new procurement group, procurements and so on

pedrobaeza:
One of the reason to make a copy, but preserve the current record as the "new revision" is for not losing communication history - aka chatter -, and also not losing possible attachments, links, and so on.

For me is totally right and it must be this way, so let's fix this one and continue.

thomaspaulb:
Ok so what remains to be done now?

pedrobaeza:
There's still a comment unattended

pablocm-aserti:
Thanks, that's the "something" I was missing.
Maybe it would still be better to carry over the chat history to the new copy because those objects don't have any business logic, but if you say so then it is fine.
If it is useful, since the pending comment is just for the test part, I've reviewed and functionally tested the fix and everything seems OK
Thanks for the fix @antiflu !

eLBati:
@antiflu please see
https://github.com/OCA/sale-workflow/pull/186#discussion_r36955270
and
https://github.com/OCA/sale-workflow/pull/186#issuecomment-130573800

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,17 @@ def copy_quotation(self):
         msg = _('New revision created: %s') % self.name
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
+
+        # set stored procurement group id on old order
+        old_revision.write({'procurement_group_id': procurement_group_id})
```
pedrobaeza: I don't think we should switch procurement group. It's good to have the procurement group set on the newest revision, not in the old one.

sale_order_revision/model/sale_order.py
```diff
@@ -64,6 +69,17 @@ def copy_quotation(self):
...
+
+        # swap order lines of old and new order
+        so_line = self.env['sale.order.line']
+        old_lines = so_line.browse(old_revision.order_line.ids)
```
pedrobaeza: You don't need to rebrowse, just assign the recordset: `old_lines = old_revision.order_line`

pedrobaeza:
Any update on this?

thomaspaulb:
Sorry guys I am extremely busy at the moment.. it's on my todolist tho

pedrobaeza:
Is this still aplying?

pedrobaeza:
It seems this is not going to be finished, so I close.

thomaspaulb:
Pedro, I still would like to finish this. Can we reopen. I have addressed the remaining issues.

sale_order_revision/model/sale_order.py
thomaspaulb: True, have done that now.

sale_order_revision/model/sale_order.py
thomaspaulb: I tested your shorter version; indeed it goes wrong, the old revision will contain both revision's lines and the new revision contains nothing.

sale_order_revision/model/sale_order.py
thomaspaulb: Good point, have fixed that.

pedrobaeza:
The branch still has conflicts.

pedrobaeza:
Alternative patch in #587

PR #187, opened 2015-07-19
8.0 sale_rental :  Fix tax on sale order lines

alexis-via:
Without this fix, when the module sale_rental is installed, the taxes are not set in the sale order lines...


alexis-via:
Could someone with super-power merge this fix rapidly please ? :+1:

sebastienbeau:
+1 tested

PR #190, opened 2015-07-22
[port][8.0]  stock_picking_reorder_lines

legalsylvain:
port stock_picking_reorder_lines for v8.
(no functional changes)
thanks for your review;


__unported__/sale_fiscal_position_update/__openerp__.py
```diff
@@ -42,7 +42,9 @@
 
 * Alexis de Lattre <alexis.delattre@akretion.com>
 """,
-    'author': "Julius Network Solutions, Akretion,Odoo Community Association (OCA)",
+    'author':
+        """Julius Network Solutions,"""
```
pedrobaeza: You don't need triple quote for this

stock_picking_reorder_lines/README.rst
```diff
@@ -0,0 +1,57 @@
+Stock picking lines with sequence number
```
pedrobaeza, line 4: Put AGPL-3 badge

stock_picking_reorder_lines/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+from openerp.osv.orm import Model
+
+
+class sale_order(Model):
```
pedrobaeza: CamelCase for the class name

stock_picking_reorder_lines/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+from openerp import fields
+from openerp.models import Model
+
+
+class StockMove(Model):
+    _inherit = 'stock.move'
+    _order = 'date_expected desc, sequence, id'
+
+    sequence = fields.Integer(
+        default=10,
```
pedrobaeza: The default should be calculated dinamically to be the last one plus 1, because this way, when you add more than 10 lines and reorder, and then add another one, you will get this one in between when saving.

stock_picking_reorder_lines/views/view.xml
```diff
@@ -3,7 +3,6 @@
     <data>
 
         <record id="view_move_picking_tree" model="ir.ui.view">
-            <field name="name">stock.move.tree</field>
```
pedrobaeza, line 6: Don't remove this key. Use with you want another name.

stock_picking_reorder_lines/models/stock_picking.py
```diff
@@ -0,0 +1,35 @@
...
+#
+#
+
+from openerp.osv.orm import Model
+
+
+class stock_picking(Model):
+    _inherit = 'stock.picking'
+
+    def _prepare_invoice_line(
```
pedrobaeza, line 28: You can use new API for this

stock_picking_reorder_lines/__openerp__.py
```diff
@@ -20,20 +20,17 @@
...
-    'demo': [],
-    'test': [],
-    'installable': False,
-    'auto_install': False,
-    'application': False,
-    'license': "AGPL-3",
+    'depends': [
+        'stock',
+        'sale',
+        'sale_stock',
```
pedrobaeza: Only this one is needed to be put, as it implies the other 2

stock_picking_reorder_lines/__openerp__.py
```diff
@@ -20,20 +20,17 @@
 #
 {
     'name': 'Stock picking lines with sequence number',
-    'version': '0.1',
+    'version': '1.0',
```
pedrobaeza: Use 8.0.1.0

pedrobaeza:
I think when you transfer the stock.picking, default report use stock.pack.operation records to be shown on the report. So, if that's true, a sequence is also needed in that object.

stock_picking_reorder_lines/views/view.xml
legalsylvain, line 6: What is the objectif of this field ? I see the name of the view fully useless.

pedrobaeza, line 6: But a required one. It's a name for the view, that it's not correctly used as a view description

stock_picking_reorder_lines/models/stock_move.py
yvaucher: I don't see any example of sequence with a default as a function in other modules. After sequence it is order by id so there no issue having multiple sequence to 10. Using interface to change order will write all sequences fields.

stock_picking_reorder_lines/views/view.xml
yvaucher, line 6: @legalsylvain This name will be displayed in Settings -> ... -> Views you can see it from the interface

yvaucher:
Please fix the remaining comments

stock_picking_reorder_lines/views/view.xml
legalsylvain, line 6: @pedrobaeza @yvaucher : it is a useless field.
And it will be generated automatically when Odoo load views.
if model = 'stock.move' and type = 'tree' then name = 'stock.move tree'.
Again, I don't understand the interest to set manually another value. Please indicate what is the bad consequencies.

(FYI, I shared like 10 / 15 modules in OCA, and none has this field defined...)

yvaucher, line 6: I feel convenient to name it something like stock.move tree.what.it.change

But, ok thank you, I learned something:
https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_ui_view.py#L257-L258

Then I agree with you you can keep it empty

stock_picking_reorder_lines/models/stock_move.py
legalsylvain: @pedrobaeza : 
@yvaucher is right. default value is unncessary on sequence field. I simply remove it.

regards.

stock_picking_reorder_lines/README.rst
legalsylvain, line 4: done.

stock_picking_reorder_lines/README.rst
```diff
@@ -0,0 +1,59 @@
+[![Licence](https://img.shields.io/badge/licence-AGPL--3-blue.svg)](http://www.gnu.org/licenses/agpl-3.0-standalone.html)
```
yvaucher: The proper format is

```
.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
    :alt: License: AGPL-3
```

It doesn't show properly.
https://github.com/grap/sale-workflow/blob/8.0_PORT_stock_picking_reorder_lines/stock_picking_reorder_lines/README.rst

However, I like the idea of the link to agpl license is good. Would you mind doing a PR on https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst ?

legalsylvain:
@pedrobaeza thanks for your review.

> I think when you transfer the stock.picking, default report use stock.pack.operation records to be shown on the report. So, if that's true, a sequence is also needed in that object.

I'm sorry, I don't understand that point. (I don't use stock.pack.operation). The objective of this PR is to port stock_picking_reorder_line from V7 to V8. Maybe I can add a "Limits / roadmap" Section with some text you could provide.

Regards.

stock_picking_reorder_lines/README.rst
yvaucher: In rst it gives: 

```
.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
   :alt: License: AGPL-3
```

legalsylvain: Done in mainteners tools repo.
https://github.com/OCA/maintainer-tools/pull/111/files

yvaucher: @legalsylvain thanks

stock_picking_reorder_lines/models/stock_move.py
pedrobaeza: That's not true. If you don't put default value, then it will be 0, and you'll get the same problem in a tree when editable=bottom (which is the case). Let me show the sequence to reproduce the error:
- Add some order lines. All get 0 as sequence.
- If you save, there's no problem, as the second order is by ID, so the order is respected.
- Then, you reorder any line, and automatically all lines from the move line gets a progressive sequence number (1, 2, 3...).
- If you add a new line, it will start with zero sequence, but in the UI, it will appear at the bottom.
- Then, when you save, this line will be reordered before the lines that has been reordered, provoking an inconsistence, because you are unawared of this change.

StefanRijnhart:
Is this the same as https://github.com/OCA/stock-logistics-workflow/tree/8.0/stock_picking_reorder_lines?

legalsylvain:
;-) @StefanRijnhart : Damn'it ! 
Yes you're right. I didn't know that this module has moved from a repository to another. 
But more generaly, I don't understand the aim of the folder unported...I believed that we had to remove from unported folder the ported module.
I'll change this PR, just to remove this folder.
Thanks for your review.

yvaucher:
Good catch @StefanRijnhart 

@legalsylvain It's clearly an error as it should have been removed from `__unported__` directory here.

Then can you check the version of stock-logistics-workflow and do a PR to remove it from **unported** here with a commit message saying it has moved ?

Thus I close this PR

legalsylvain:
Note : I have to investigate the ported module, because this line in V7, https://github.com/OCA/sale-workflow/pull/190/files#diff-8767e87d7c090b1f16ea65cc83671fbcL58
has not been ported in the V8 module in stock-logistic-workflow, and I don't understand why.

pedrobaeza:
@legalsylvain, the method is there, but in https://github.com/OCA/sale-workflow/pull/190/files#diff-e2e31f7131435446f094eb21f25694a7R28 (maybe because the file has been split)

legalsylvain:
Yes @pedrobaeza , but #190 is my current PR that @yvaucher just closed, because it was duplicated.

But this function doesn't exist in the module pointed by @StefanRijnhart here https://github.com/OCA/stock-logistics-workflow in V8 serie.

In a few words, @damdam-s ported the module but not in the same way as me.
I'd like to understand why.

Kind regards.

pedrobaeza:
Ahm, I don't know then, but some flow has been moved in v8 to procurement orders, so maybe that's the reason.

PR #198, opened 2015-08-21
[ADD] sale_update_pricelist: 

hugosantosred:
Module for change line prices when pricelist changed in sale order.


sale_update_pricelist/model/sale.py
```diff
@@ -0,0 +1,67 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp import models, _
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    def onchange_pricelist_id(self, cr, uid, ids, pricelist_id, order_lines,
```
nhomar: Can not you use new api here?

sale_update_pricelist/model/sale.py
```diff
@@ -0,0 +1,67 @@
...
+                        "Price will remaing unchanged\n") % (
+                        line['name'], line['product_uom_qty'])
+
+        res['value']['order_line'] = line_dict
+        if warning_msg:
+            res['warning'] = {
+                'title': _('Pricelist Warning!'),
+                'message': warning_msg
+            }
+        return res
```
nhomar, line 69: You are not calling the super, the original on_change set the currency:

https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L305-305

Aren't you considering such multicurrency behavior?

sale_update_pricelist/README.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Credits
+=======
+
+Contributors
+------------
+
+* Hugo Santos <hugo.santos@factorlibre.com>
+
+Maintainer
```
nhomar, line 29: Can you add the new format which include the runbot link?

nhomar:
Cool feature, thank @hugosantosred for this job.... I have just some comments about the functionality and algorithm.

Another point if that it should copy and paste the original test case for this method and improve with the new behavior in sale module in order to avoid any regresion due to this installtion..

Thanks.

Conceptually :+1:  until test is added.

sale_update_pricelist/README.rst
hugosantosred, line 29: @nhomar How can I add the runbot link?

Is this flag? [Runbot Status](https://runbot.odoo-community.org/runbot/repo/github-com-oca-sale-workflow-167)

nhomar, line 29: 2015-08-21 6:21 GMT-05:00 Hugo Santos notifications@github.com:

> @nhomar https://github.com/nhomar How can I add the runbot link?
> 
> Is this flag? [image: Runbot Status]
> https://runbot.odoo-community.org/runbot/repo/github-com-oca-sale-workflow-167

here the new template:

The variables can be taken from the runbot url itself (sorry I don't
remember where is exactly the table with all the mapping of IDs.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

hugosantosred:
@nhomar Changes done to README.rst. Thanks for the comments!

I will make the test cases soon.

pedrobaeza:
@hugosantosred, there's already a proposed PR that handles exactly the same use case: https://github.com/OCA/sale-workflow/pull/54. But it's even more versatile, because it includes a button that allows you to launch the calculation whenever you want, even if there's no change on the pricelist (for example, to handle a change in list prices), so for me this one :-1:

pedrobaeza:
I close this one as  #54 is already merged

PR #199, opened 2015-08-27
[FIX] sale_order_type: Fixed bug when picking is created from a purchase order

Daniel-CA:
When the stock.picking is created the 'invoice_state' field cannot be null.

### DIFF
--- a/sale_order_type/models/__init__.py
+++ b/sale_order_type/models/__init__.py
@@ -8,3 +8,4 @@ from . import sale_order_type
 from . import sale_order
 from . import stock_picking
 from . import res_partner
+from . import stock_move
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -11,6 +11,8 @@ class TestSaleOrderType(common.TransactionCase):
         super(TestSaleOrderType, self).setUp()
         self.sale_type_model = self.env['sale.order.type']
         self.sale_order_model = self.env['sale.order']
+        self.stock_picking_model = self.env['stock.picking']
+        self.picking_type_out = self.env.ref('stock.picking_type_out')
         self.sale_line_model = self.env['sale.order.line']
         self.picking_model = self.env['stock.picking']
         self.partner = self.env.ref('base.res_partner_1')
@@ -68,3 +70,10 @@ class TestSaleOrderType(common.TransactionCase):
         for picking in sale_order.picking_ids:
             self.assertEqual(self.sale_type.invoice_state,
                              picking.invoice_state)
+
+    def test_stock_picking_create(self):
+        self.picking_out = self.picking_model.create({
+            'partner_id': self.partner.id,
+            'picking_type_id': self.picking_type_out.id
+        })
+        self.assertTrue(self.picking_out.id)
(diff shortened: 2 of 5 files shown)


pedrobaeza:
Thanks for the patch, Dani, but seeing now that code, I think it's better/safer to look at the chained procurement to get the sale order instead of looking for a raw text. Can you please make this change also?

Daniel-CA:
Yes, I see better in that way.

Daniel-CA:
Code updated. By the way, I have seen that this module has his own test files, they should have warned about the bug?

sale_order_type/models/__init__.py
```diff
@@ -6,5 +6,5 @@
 from . import account_invoice
 from . import sale_order_type
 from . import sale_order
-from . import stock_picking
+from . import stock
```
pedrobaeza: Don't change the file and rename it, but add it a new one (stock_move.py).

sale_order_type/models/stock.py
```diff
@@ -6,16 +6,30 @@
 from openerp import api, models
 
 
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    @api.model
+    def _prepare_picking_assign(self, move):
+        values = super(StockMove, self)._prepare_picking_assign(move)
+        values['picking_group_id'] = move.group_id.id
```
pedrobaeza: Why are you doing this instead of directly change 'invoice_state' value here?

pedrobaeza:
It's not a bug per se, but a possible case: origin is filled automatically by the process, but imagine you have a module that adds to the origin string something more. Then your code will break, but only in that case. Tests covers some things, but they're not covering every possibility you can imagine (although they ideally should). Therefore, when a bug/use case is found, a good technique is to add a test for checking this concrete use case. It can be this case for example, that you can add a picking with an altered origin string, but as we have changed the code, I don't think it would add value to the test suite.

Daniel-CA:
Thanks Pedro, I was talking about the  'invoice_state' field null error, but as you say if the test doesn't covers what happens when a purchase order is created for sure no warning will arise.
I have modified the code as you say, better directly change 'invoice_state' value in the preparation method.

pedrobaeza:
:+1:

yvaucher:
@Daniel-CA Could you add a unit test for this use case?

Daniel-CA:
Test added.

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -68,3 +70,9 @@ def test_sale_order_confirm(self):
         for picking in sale_order.picking_ids:
             self.assertEqual(self.sale_type.invoice_state,
                              picking.invoice_state)
+
+    def test_stock_picking_create(self):
+        self.picking_out = self.picking_model.create({
```
pedrobaeza, line 75: There's no assert here to test any condition, so the tests is useless (except possible code errors). You can just simply check that self.picking_out is really created.

Daniel-CA, line 75: In this way?

Daniel-CA:
Anything more to change here?

pedrobaeza:
:+1:

yvaucher:
Thanks :+1:

yvaucher:
merged at https://github.com/OCA/sale-workflow/commit/96ec074151550fb674f627e8b45c6276a5c5cd07

PR #200, opened 2015-08-28
Recompute standard price when confirming dropshipping order

mdietrichc2c:
This PR is to recompute the `standard_price` field on products where the cost method uses (usually) the inbound stock moves to compute it.

Since the moves created by dropshipping cases are `out`, the cost would never be modified otherwise.


mdietrichc2c:
Actually, I don't think this is the way to go for this; I'll re-do something else.

PR #201, opened 2015-08-28
[FIX] don't rename parameters in product_id_change

gurneyalex:
some addons such as sale_stock call it with named arguments
and this will cause a crash if you rename e.g. the product arg
to product_id.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+language: python
 sudo: false
 cache: pip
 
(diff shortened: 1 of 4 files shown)


mdietrichc2c:
:+1:

guewen:
:+1:

yvaucher:
:+1:

gurneyalex:
runbot is red (warnings in the logs) also on the 7.0 branch

pedrobaeza:
:+1:

PR #204, opened 2015-09-02
[FIX] sale_order_line_description: missing translation for field description_sale

tafaRU:


sale_order_line_description/models/sale.py
```diff
@@ -30,12 +30,12 @@ def product_id_change(
...
             uos=False, name='', partner_id=False, lang=False,
             update_tax=True, date_order=False, packaging=False,
-            fiscal_position=False, flag=False
+            fiscal_position=False, flag=False, context=None
     ):
         res = super(SaleOrderLine, self).product_id_change(
             pricelist, product, qty, uom, qty_uos, uos, name,
             partner_id, lang, update_tax, date_order, packaging,
-            fiscal_position, flag)
+            fiscal_position, flag, context)
```
pedrobaeza: Take the occasion to name each keyword argument in case any inheritance change the argument order

tafaRU: Ok.

This change has no sense with new API, I'm still working on it.

pedrobaeza: Of course, I only see the diff and don't look at the @api.multi...

tafaRU:
@pedrobaeza could you please have a look at it?

eLBati:
LGTM :+1:

pedrobaeza:
:+1:

guewen:
:+1:

PR #209, opened 2015-09-15
New module - Pricelist Nondependent Pricing

richard-willdooit:
This module is designed to allow a pricelist definition where the price is a
fixed figure not calculated from another price.


pedrobaeza:
This one does the same as https://github.com/OCA/product-attribute/pull/50, so you should review that one.

PR #212, opened 2015-09-18
[FIX] <sale_order_type> Invoice state from type not applied in pickings

oihane:
When a picking is assigned to a stock.move if this has invoice_state = 'to2invoiced' but the picking had 'none' it updates the value in picking to 'to2invoiced'


pedrobaeza:
Please add a test about it

oihane:
@pedrobaeza I updated the tests.

pedrobaeza:
:+1:

Daniel-CA:
:+1:

PR #213, opened 2015-09-18
8.0 copy product set

petrus-v:
With this PR, when duplicate product set, product set line will be duplicated as well.

### DIFF
--- a/sale_product_set/i18n/es.po
+++ b/sale_product_set/i18n/es.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: sale-workflow (8.0)\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-09-18 13:31+0000\n"
+"POT-Creation-Date: 2015-09-23 09:20+0000\n"
 "PO-Revision-Date: 2015-09-03 07:10+0000\n"
 "Last-Translator: <>\n"
 "Language-Team: Spanish (http://www.transifex.com/oca/OCA-sale-workflow-8-0/"
@@ -76,24 +76,25 @@ msgid "Last Updated on"
 msgstr "Última actualización el"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set.py:9
 #: field:product.set,name:0
-#, python-format
 msgid "Name"
 msgstr "Nombre"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set_line.py:15
 #: field:product.set.line,product_id:0
-#, python-format
 msgid "Product"
 msgstr "Producto"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/wizard/product_set_add.py:12
 #: model:ir.actions.act_window,name:sale_product_set.act_open_product_set_view
-#: model:ir.model,name:sale_product_set.model_product_set
 #: model:ir.ui.menu,name:sale_product_set.menu_product_set
+#, fuzzy
+msgid "Product Set"
+msgstr "Producto"
+
+#. module: sale_product_set
+#: code:addons/sale_product_set/wizard/product_set_add.py:12
+#: model:ir.model,name:sale_product_set.model_product_set
 #: view:product.set:sale_product_set.view_product_set_form
 #: view:product.set:sale_product_set.view_product_set_search
 #: view:product.set:sale_product_set.view_product_set_tree
@@ -113,21 +114,16 @@ msgid "Product set lines"
 msgstr ""
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set.py:9
 #: help:product.set,name:0
-#, python-format
 msgid "Product set name"
 msgstr ""
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set.py:11
 #: field:product.set,set_line_ids:0
-#, python-format
 msgid "Products"
 msgstr ""
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set_line.py:17
 #: code:addons/sale_product_set/wizard/product_set_add.py:14
 #: field:product.set.add,quantity:0 field:product.set.line,quantity:0
 #, python-format
@@ -135,16 +131,12 @@ msgid "Quantity"
 msgstr "Cantidad"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set_line.py:23
 #: field:product.set.line,sequence:0
-#, python-format
 msgid "Sequence"
 msgstr "Secuencia"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/models/product_set_line.py:21
 #: field:product.set.line,product_set_id:0
-#, python-format
 msgid "Set"
 msgstr ""
 
--- a/sale_product_set/i18n/es.po
+++ b/sale_product_set/i18n/es.po
@@ -1,21 +1,28 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
 # * sale_product_set
-# 
+#
 # Translators:
 msgid ""
 msgstr ""
 "Project-Id-Version: sale-workflow (8.0)\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-09-09 12:27+0000\n"
+"POT-Creation-Date: 2015-09-18 13:31+0000\n"
 "PO-Revision-Date: 2015-09-03 07:10+0000\n"
 "Last-Translator: <>\n"
-"Language-Team: Spanish (http://www.transifex.com/oca/OCA-sale-workflow-8-0/language/es/)\n"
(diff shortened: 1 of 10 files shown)


sale_product_set/models/product_set.py
```diff
@@ -8,4 +8,13 @@ class ProductSet(models.Model):
 
     name = fields.Char(_('Name'), help=_('Product set name'), required=True)
     set_line_ids = fields.One2many(
-        'product.set.line', 'product_set_id', _(u"Products"))
+        'product.set.line', 'product_set_id', _(u"Products"), copy=True)
```
pedrobaeza: Why are you using `_` in field definition? Labels are already handled. This way, you'll get a static translation. Please remove it

petrus-v: All right, I didn't know that!

That was already like this, anyway I think I wrote it, Thanks a lot!

petrus-v: @pedrobaeza 

I'm cleaning all of them in `sale_product_set` and `sale_product_set_layout`, I guess `help` attribute is handle as well?

pedrobaeza: Yeah, the same. Sorry for not noticing before.

petrus-v: @pedrobaeza, 
cc @moylop260 

I don't really know how that's works but I guess this kind of issue could be track by maintainer-quality-tools lint tools or [moylop260/pylint-oca](github.com/moylop260/pylint-oca) project ?

Is there a list of implemented (or expected) lint issues?

pedrobaeza: Yeah, it can be another think to check, but for now all is a WIP, and no list is yet built. Pylint config files are auto-documented on the other side.

moylop260: @petrus-v 
This case is a good lint case.
If you have other ones you can propose the list here:
https://github.com/OCA/maintainer-quality-tools/issues/252
(I added you idea)

moylop260:
:+1:

petrus-v:
@moylop260 

Thanks for reviews and the link to lint check list!

laetitia-gangloff:
Is it also possible to rename the menu "Product set" to "Product **S**et" ?

petrus-v:
@laetitia-gangloff 

Yes, I renamed the menu entry and update translation files.

regards

laetitia-gangloff:
Thank you !

:+1:

sale_product_set/i18n/es.po
```diff
@@ -69,24 +76,25 @@ msgid "Last Updated on"
...
 msgstr "Producto"
 
 #. module: sale_product_set
-#: code:addons/sale_product_set/wizard/product_set_add.py:12
 #: model:ir.actions.act_window,name:sale_product_set.act_open_product_set_view
-#: model:ir.model,name:sale_product_set.model_product_set
 #: model:ir.ui.menu,name:sale_product_set.menu_product_set
+#, fuzzy
+msgid "Product Set"
+msgstr "Producto"
```
pedrobaeza, line 93: Please translate it to "Conjunto de productos"

sale_product_set/models/product_set.py
```diff
@@ -1,11 +1,20 @@
...
-from openerp import fields, models, _
+from openerp import api, fields, models, _
 
 
 class ProductSet(models.Model):
     _name = 'product.set'
     _description = 'Product set'
 
-    name = fields.Char(_('Name'), help=_('Product set name'), required=True)
+    name = fields.Char(u"Name", help=u"Product set name", required=True)
```
pedrobaeza, line 9: unicode is not needed here, because there isn't strange chars, but anyway it doesn't harm

sale_product_set/tests/test_product_set.py
```diff
@@ -68,3 +68,15 @@ def test_add_set_on_empty_so(self):
...
+        pdt_set = self.env.ref('sale_product_set.product_set_i5_computer')
+        pdt_set_copy = pdt_set.copy()
+        self.assertEquals(
+            len(pdt_set.set_line_ids),
+            len(pdt_set_copy.set_line_ids)
+        )
+        self.assertNotEquals(
+            pdt_set.set_line_ids,
+            pdt_set_copy.set_line_ids
+        )
```
pedrobaeza: You can test here also the name of the copy

sale_product_set/models/product_set.py
```diff
@@ -1,11 +1,20 @@
...
     set_line_ids = fields.One2many(
-        'product.set.line', 'product_set_id', _(u"Products"))
+        'product.set.line', 'product_set_id', u"Products", copy=True)
+
+    @api.one
+    def copy(self, default=None):
+        if default is None:
+            default = {}
+        default.update({
```
pedrobaeza: You can also put this shorter one `default['name'] += " %s" % _("(copy)")` and also avoid the space in the translation, because it should be forgotten when translating.

sale_product_set/i18n/es.po
petrus-v, line 93: @pedrobaeza,

Sorry I won't do it, as I'm not a Spanish man and there is Transiflex to do it!

I change this file through msgmerge, it's moving as `fuzzy` after renaming `Product set` to `Product Set` as requiest by reviewers and merging this file according the new pot file.

I hope you understand my point of view.

sale_product_set/models/product_set.py
petrus-v, line 9: Yes your're right! French practices...

sale_product_set/i18n/es.po
pedrobaeza, line 93: OK, but having to make changes and me being spanish, that's why I ask you to do it in one movement, but if you don't want, I'll access Transifex when merged, change it, and then wait for the pull come back.

petrus-v:
@pedrobaeza 

Thanks for reviews, I take care about your 2 last comments: adding unit test on duplicated product set name and removing space on translated string.

regards

sale_product_set/i18n/es.po
petrus-v, line 93: I understand it's a long way to get it, I would like to avoid that Github PR becomes a translation reviews avoiding bikeshedding about things I can't understand!...

sale_product_set/models/product_set.py
ccomb, line 9: Not absolutely needed on this particular example, indeed, but I believe it's a safe practice to always put unicode objects on things that are targeted to be read by humans, and simple strings for things that are read by the machines.

pedrobaeza, line 9: As guidelines indicate that you should put all fields in English, it shouldn't be a real issue :wink:

sale_product_set/models/product_set.py
```diff
@@ -1,11 +1,18 @@
...
     _name = 'product.set'
     _description = 'Product set'
 
-    name = fields.Char(_('Name'), help=_('Product set name'), required=True)
+    name = fields.Char(u"Name", help=u"Product set name", required=True)
     set_line_ids = fields.One2many(
-        'product.set.line', 'product_set_id', _(u"Products"))
+        'product.set.line', 'product_set_id', u"Products", copy=True)
+
+    @api.one
```
ccomb: `api.one` is deprecated on v9. Assuming there is a single incoming record has always been a source of bugs, because you never know if some downstream code will do the same assumption. Maybe it'd a good idea to avoid it and put a loop instead?

pedrobaeza: In this case, a `self.ensure_one()` should be done, as the copy method is only intended with one record each time.

ccomb: I agree with the web interface you cannot copy more than one record, but with the API or xmlrpc or in a module it might be useful to be able to do a mass copy.

pedrobaeza: I'm afraid it's not possible in current ORM implementation: https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L4853, because the argument is only an ID and the return is also only one ID. In v9 the situation is the same (https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L4914), so maybe in a future?

ccomb:
otherwise :+1:

sale_product_set/models/product_set.py
petrus-v: So, I'm not doing anything on that, if I'm right, as the decorator ensure me there is already one record `self.ensure_one()` is not necessary!

pedrobaeza: Well, we are talking to put it as @api.multi and add self.ensure_one().

petrus-v:
Is there something missing on this PR that I should do?

StefanRijnhart:
Just change @api.one to @api.multi and add self.ensure_one() to that method as @pedrobaeza asked and it should be done.

petrus-v:
@pedrobaeza @StefanRijnhart,

As asked, I remove api.one!

pedrobaeza:
:+1: when Travis goes green

petrus-v:
Is it possible to merge this branch now?

pedrobaeza:
Yeah, sure!

petrus-v:
@pedrobaeza 

So quick, thank you.

PR #214, opened 2015-09-18
8.0 sale recalculation optionnaly without name

petrus-v:
This is a purpose to improve `sale_order_price_recalculation` module:
- [x] The existing button is reseting price only (without changing descriptions)
- [x] Adding  a second button to reset sale order lines descriptions only 
- [x] hide those button links when it's a quotation (sale order is draft)
- [x] Move buttons below sale order lines

PR #219, opened 2015-09-29
Display reset button on SO when state == sent

petrus-v:
PR #214 restricted buttons visibilities to `draft` state. As long prices and descriptions are settable at `sent` state I think those buttons should be visible at this stage.


pedrobaeza:
:+1: 

I fast-track as this is a clear enough issue

petrus-v:
@pedrobaeza,

Great thank you!

PR #225, opened 2015-10-15
[ADD] Default sales incoterm per partner

StefanRijnhart:


sale_partner_incoterm/tests/test_sale_partner_incoterm.py
```diff
@@ -0,0 +1,32 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.tests.common import TransactionCase
+
+
+class TestSalePartnerIncoterm(TransactionCase):
+    def test_sale_partner_incoterm(self):
+        """ Check that the customer's default incoterm is retrieved in the \
+sales order's onchange """
```
pedrobaeza, line 26: strange indentation

pedrobaeza:
One minor thing. The rest LGTM.

alexis-via:
@StefanRijnhart I see that you sync the field "sale_incoterm_id" between company and contacts (_commercial_fields). But you display this field in the "Sales & Purchase" tab, not in the "Accounting" tab. If you sync it between company and contacts, I would prefer that you put this field in the "Accounting" tab, because it has the "protection" to not change it on the Contact but only on the company.

Otherwise, users may change this field on the Contact without realizing that it will also impact the company.

pedrobaeza:
I think the change is from company to contacts, not backwards, and there's use cases where you want to put a specific incoterm for a contact (an address outside the country), but if you modify the company one, all children are overwritten.

jgrandguillaume:
Thanks for this one ! :+1:

sale_partner_incoterm/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/
+{sale-workflow}/issues>`_.
```
tafaRU: Please remove the brackets

sale_partner_incoterm/README.rst
```diff
@@ -0,0 +1,58 @@
...
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback `here <https://github.com/OCA/
+{sale-workflow}/issues/new?body=module:%20
+{sale_partner_incoterm}%0Aversion:%20
```
tafaRU: Same as above

sale_partner_incoterm/README.rst
```diff
@@ -0,0 +1,58 @@
...
+{8.0.1.0.0}%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
tafaRU: Please put here only `8.0`

tafaRU:
A little nitpicking, but :+1: 

Thanks!

sale_partner_incoterm/tests/test_sale_partner_incoterm.py
StefanRijnhart, line 26: The test log level shows the first line of the docstring, so I need the line continuation. We can then either have the strange indentation here, or if I fill it out with spaces, these spaces will show up and make the debug log less readable.

StefanRijnhart:
@pedrobaeza @alexis-via I thought synchronization was two-ways but you are right that it is confusing to show it on contacts. How about I keep it as a commercial field, keep it on the sales tab but make it invisible on non-commercial partners?

pedrobaeza:
But I prefer to have it also on contacts, for the use case I have told.

sale_partner_incoterm/README.rst
StefanRijnhart: Thanks ;-) fixed now.

sale_partner_incoterm/README.rst
StefanRijnhart: Are you sure? I thought that this was meant specifically to refer to the actual module version, not to the branch.

StefanRijnhart:
OK @alexis-via do you agree dropping this from the commercial fields?

sale_partner_incoterm/README.rst
tafaRU: I'm quite sure, it refers to odoo version. I remember @pedrobaeza wrote me this in another PR.

pedrobaeza: This is only to have a clue about Odoo version, not module version, because the most of the times if we put the module version, we'll end with an incorrect one because contributors forget to change this data (it happens also in manifest, you know...)

StefanRijnhart: OK, makes sense. Updated.

alexis-via:
Yes, dropping it from the commercial fields is another solution which is fine for me.

StefanRijnhart:
Done. Failing tests due to #228

StefanRijnhart:
Tests restarted and now green.

alexis-via:
:+1:

PR #229, opened 2015-10-19
7.0 fix sale visible tax view

legalsylvain:
Without this fix, the two new fields are not displayed in sale order form view.
This path fixes the bug.

Kind regard.


laetitia-gangloff:
:+1:

bguillot:
:+1:

StefanRijnhart:
I see that the previous expression was ambiguous between the form and the tree views. :+1:

PR #230, opened 2015-10-21
[8.0][FIX] Sale order lines not set to draft

rven:
When creating a new revision of a cancelled sale order, the sale order is set to the state draft, but the order lines are not updated to draft.
The lines are still in state cancelled and cannot be updated.


pedrobaeza:
:+1:

tafaRU:
:+1:

PR #233, opened 2015-10-28
[ADD] Modify sale delivery address

StefanRijnhart:
# 
# Modify sale delivery address

For sales orders that have not been shipped completely, this module allows the
user to change the delivery address on the sales order for the remaining
deliveries.


pedrobaeza:
Why don't simply change the partner on the remaining picking?

StefanRijnhart:
I'm following the way that the address is propagated in standard Odoo, to prevent any unforeseen inconsistencies from occurring.

sale_delivery_address_modify/models/sale_order.py
```diff
@@ -0,0 +1,55 @@
...
+    def write(self, values):
+        """ Propagate a modified delivery address to the procurement group,
+        the pickings and stock moves.
+        """
+        res = super(SaleOrder, self).write(values)
+        if 'partner_shipping_id' in values:
+            for order in self:
+                if not order.procurement_group_id or order.shipped:
+                    continue
+                for procurement in self.env['procurement.order'].search(
```
pedrobaeza, line 22: You don't need to search for procurements. Search only pickings with sale_id field = order.id

guewen:
> Why don't simply change the partner on the remaining picking?

Separation of concerns. I guess that's because the salespeople that need to change the address (for instance after a call of their customer) are not the same people that work on the delivery orders. Also you are sure not to forget a picking.

:+1: 

EDIT: @pedrobaeza I think I misunderstood your question, forget my answer ;-)

pedrobaeza:
OK, understood

sale_delivery_address_modify/models/sale_order.py
StefanRijnhart, line 22: Shouldn't that be equivalent?

StefanRijnhart:
Seems that the INCLUDE/EXCLUDE config messes up coverage. Is there a solution to that?

sale_delivery_address_modify/models/sale_order.py
pedrobaeza, line 22: Well, you save 1 search with my expression

pedrobaeza:
Not for now. See https://github.com/OCA/maintainer-quality-tools/issues/249

Ideal for me is not to have INCLUDE/EXCLUDE.

pedrobaeza:
@guewen, analysing again your answer, I'm still reluctant: a salesman can access the delivery(ies) order(s) via a button in the sales order itself, so no need to access warehouse menu. And he/she can change the date on the non-done picking (it will be always only one).

sale_delivery_address_modify/models/sale_order.py
StefanRijnhart, line 22: Tested it, but it does not work in the following situation: create a sale order and confirm. Update the delivery address. This updates the delivery address on the picking. Cancel the picking. On the sale order, recreate the picking. The picking has the old delivery address because it is taken from the procurement, not the sale order.

pedrobaeza, line 22: OK, that flow is enough to justify this module, although this inline comment refers to the search expression, not the general concern. You can add this rationale in the module description.

StefanRijnhart, line 22: OK, then I did not understand what you meant. Description updated.

pedrobaeza, line 22: I'm talking to use this code instead of the 2 searches:

``` python
            pickings = self.env['stock.picking'].search(
                    [('sale_id', '=', order.id), ('state', '!=', 'done')])
```

StefanRijnhart, line 22: sale_id is a non searchable function field in 8.0. I could use

```
pickings = self.env['stock.picking'].search(
    [('id', 'in', (order.picking_ids.ids), ('state', '!=', 'done')])
```

Where picking_ids is a function field that triggers the same query on pickings that I am curently using, through the procurement group. Would that be an improvement?

pedrobaeza, line 22: Ah, OK, I didn't know about the not searchable issue. What you can do then is:

``` python
pickings = order.picking.ids.filtered(lambda x: x.state != 'done')
```

More pythonic and readable, and if one day this field is stored, we will automatically get the improvement.

StefanRijnhart, line 22: OK, done.

sale_delivery_address_modify/models/sale_order.py
```diff
@@ -0,0 +1,54 @@
...
+                        raise UserError(_(
+                            'The delivery address can only be changed if '
+                            'the related customer location remains the same.'))
+                        procurement.write({'partner_dest_id':
+                                           order.partner_shipping_id})
+                pickings = order.picking_ids.filtered(
+                    lambda picking: picking.state != 'done')
+                pickings.write(
+                    {'partner_id': order.partner_shipping_id.id})
+                self.env['stock.move'].search(
```
pedrobaeza: Another optional way to do this is (not more efficient, but more readable):

``` python
pickings.mapped('move_ids').write(...
```

StefanRijnhart: Done.

pedrobaeza:
Thanks :+1:

guewen:
Travis is red due to 

```
sale_delivery_address_modify/__openerp__.py:20: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
```

StefanRijnhart:
Let me restart as https://github.com/OCA/maintainer-quality-tools/pull/315 has been merged

sale_delivery_address_modify/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+                if not order.procurement_group_id or order.shipped:
+                    continue
+                for procurement in self.env['procurement.order'].search(
+                        [('group_id', '=', order.procurement_group_id.id)]):
+                    if (procurement.location_id != order.partner_shipping_id
+                            .property_stock_customer):
+                        raise UserError(_(
+                            'The delivery address can only be changed if '
+                            'the related customer location remains the same.'))
+                        procurement.write({'partner_dest_id':
```
atchuthan: after error raises with UserError, the next lines won't be executed

atchuthan:
Other than a dead block, code LGTM :+1:

StefanRijnhart:
Thanks @atchuthan, fixed in https://github.com/OCA/sale-workflow/pull/233/commits/7f9c3f4b7da2e5eb546d621385694f5d0fb07f94. Rebased to see what travis comes up with now.

StefanRijnhart:
Travis is green as tests now succeed after working around a bug in Odoo. Runbot is still failing for reasons unrelated to this PR (@gurneyalex, can we have bank-statement-reconcile as a dependency of this repo in Runbot?)

pedrobaeza:
I have added the dependency and rebuild the instance.

StefanRijnhart:
Thanks @pedrobaeza, all green now!

sale_delivery_address_modify/i18n/sale_delivery_address_modify.pot
```diff
@@ -0,0 +1,28 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove the pot file, as this is no longer needed and can confuse people if it's not up to date.

sale_delivery_address_modify/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Can you please use short headers like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

StefanRijnhart:
Sorry for the delay, POT file is now removed and new style license block is applied.

gurneyalex:
:+1:

PR #241, opened 2015-12-01
[ADD] Added wizard for sorting purchase order lines

azoellner:
A wizard for sorting purchase order lines.

Usage: In the `purchase.order` form view, press the newly added `Sort lines...` button to start the wizard. In the wizard, select the key(s) and direction(s) for sorting and press the `Do sort` button to sort the purchase order lines.

### DIFF
--- a/purchase_order_reorder_lines_wizard/models/purchase.py
+++ b/purchase_order_reorder_lines_wizard/models/purchase.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 initOS GmbH (<http://www.initos.com>).
+#    Author Andreas Zoellner <andreas.zoellner at initos.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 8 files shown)


lepistone:
Thanks for your contribution! This seems unrelated to `sale-workflow`: please resubmit somewhere else.

OSguard:
hi @lepistone 

we know the point is we depend on 'purchase_order_reorder_lines' and for version 7.0 it is located here, so we decide to put the PR here as well. Before seperating on version. For the next version it should be somewhere else.

Kind Regards
Markus

lepistone:
OK @OSguard, reopened. Thanks!

OSguard:
thanks. Runbot is not green of 7.0 branch, not related to this pull request.

purchase_order_reorder_lines_wizard/models/purchase.py
```diff
@@ -0,0 +1,118 @@
...
+            'context': context,
+        }
+
+    @staticmethod
+    def sort_order_lines(order_lines, sort_option):
+        key = sort_option['key']
+        dir = sort_option['dir']
+
+        if key == 'name':
+            fun = lambda x: x.product_id.name
```
legalsylvain, line 72: I don't understand the name of the var. fun like 'funny' ?
sorry if my question is not relevant.

azoellner, line 72: No, its `fun` like in `function`.

Maybe `func` or `fcn` would be clearer. Or in this case even `key`, since that is how it's used below in `sorted(...)`.

legalsylvain, line 72: oh sorry. yes maybe func is more usual. (not a blocking point)

azoellner:
As this seems to be obsolete by now due to module "mass_sorting" OCA/server-tools#345, I do close this PR.

PR #244, opened 2015-12-02
sale_automatic_workflow: ship only available goods

lepistone:
This is a new option that will ship what is available, and create a
backorder or the rest, like when we check availability and transfer from
the picking. The default is still as before: all the quantity is shipped
even if it is not available.

This supersedes OCA/e-commerce#80.


lepistone:
@elicoidal @sebastienbeau I addressed your remarks in OCA/e-commerce#80, thanks!

tremlin:
Code review :+1: , looks good.

guewen:
:+1:

PR #245, opened 2015-12-04
[FIX] do not generate payment if amount=0

damdam-s:


gurneyalex:
:+1:

pedrobaeza:
It's a WIP? Anyway, it seems :+1:

lepistone:
:+1: in principle, thanks. Still I am unsure: should we consider a near-zero amount because of float or rounding problems?

lepistone:
closed by mistake, I reopen

mdietrichc2c:
I agree with @lepistone : I think we should consider using `float_is_zero`, especially since the test was added to only call the method when the amount is not 0.

mdietrichc2c:
:+1:

mdietrichc2c:
Actually, :-1: : I didn't notice that the condition was `if float_is_zero`, meaning the call is only made when the amount is equal to 0.
@damdam-s : can you add the `not` in front of `float_is_zero` ?

damdam-s:
@mdietrichc2c ouch ... :( sorry. fix done

sale_payment_method/sale.py
```diff
@@ -103,8 +104,10 @@ def automatic_payment(self, amount=None):
             amounts = [(date, amount)]
 
         # reversed is cosmetic, compute returns terms in the 'wrong' order
+        precision = self.env['decimal.precision'].precision_get('Account')
```
lepistone, line 106: please move this up 1 line, otherwise the comment does not make sense.

lepistone:
A minor remark, otherwise :+1: thanks!

damdam-s:
@lepistone done

mdietrichc2c:
:+1:

StefanRijnhart:
:+1:

PR #246, opened 2015-12-16
[8.0] [IMP] sale_order_type: Make it multi company 

rlizana:


pedrobaeza:
Thanks for this! Maybe you should think about resetting on a post_init_hook all the existing type to company_id=False to not modify the behaviour before this patch of existing installations. You should also bump the module version, according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

pedrobaeza:
@rlizana, can you please check this?

pedrobaeza:
@rlizana, are you going to finish this?

eLBati:
Meanwhile sale_order_type became multi company aware. So I close

PR #247, opened 2015-12-17
[8.0] sale_automatic_workflow : add default journal on workflow process

laetitia-gangloff:
This PR allows user to define which default sale journal use on invoice.


sbidoul:
:+1: (code review, functional test)

Useful to have precise control on the sale journal used for e-commerce.

pedrobaeza:
:+1:

PR #253, opened 2015-12-23
[ADD] [sale_partner_contact]

JordiBForgeFlow:
This module allows to add to a quotation or sales order the contact person.

### DIFF
--- a/sale_partner_contact/README.rst
+++ b/sale_partner_contact/README.rst
@@ -1,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Sale partner contact
+====================
+
+This module allows to add to a quotation or sales order the contact person of
+for that order.
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to to the quotation or sales order and enter the contact. You can enter a contact that
+is related to the company, or enter another one if needed.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/8.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/167/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+167/issues/new?body=module:%20
+sale_partner_contact%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A
+%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Jordi Ballester - Eficent Business and IT Consulting Services S.L. <jordi.ballester@eficent.com>
+
+
+Maintainer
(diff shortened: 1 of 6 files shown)


rafaelbn:
Thanks! Hi @jbeficent could you check travis please?
Cc @antespi @pedrobaeza

pedrobaeza:
This module has no sense for me, as you can put directly a contact in the partner_id field. Then, you only have to customize the report if you want to distinguish between partner_id and partner_id.commercial_partner_id.

JordiBForgeFlow:
If this person leaves company A and goes to company B, you'd have to duplicate the contact to ensure that the sales is attributed to company A. Not nice, because the contractual agreement(aka sales order) is with a company, not with a person.

rafaelbn:
@jbeficent you have to duplicate contact always. This is not only for sales, is for emails, event register, tasks as follower. In Odoo you cannot never reuser contact for diferent companies, you have to use diferent one. Because if not you will have to create partner contact for every model.

In the other hand, @pedrobaeza this is the situation to solve in my opinion:
- Partner of the sale: person who requiest that sale (contact)
- Partner for the invoice: person o place where you have to send invoice (invoice)
- Partner for the shipping: where you will deliver products (shipping)

Other need (which must be linked) is in invoicing. When there are a lot of people in a company (>20 fox exemple) in diferent you and you must separate in the same way of sale:
- Invoice partner (partner for the invoice in sale order)
- Invoice Shipping partner (partner for the shipping in sale order)
- Invoice Contact partner (partner_id field in sale order)

OK, then is more logic to add a new field in invoices invoice_partner contact and take the field from partner_id in sale order.

@jbeficent you have to take care also because all this fields are related with portal modules, who access who not.

JordiBForgeFlow:
@rafaelbn Relating to contacts moving to a separate company, you'd need to then add a restriction to prevent a partner from being moved to a different company, if he's been involved in transactions with the previous company. What you describe is a recommendation based on your experience. Not preventing anyone from doing this simple change means that people could completely screw the sales orders, invoices, etc.. just because they moved the contact person to another company.

It seems to me a better practice that you should be able to remove that contact person as follower from whatever he's been involved in the past, and then assign him to the new company, as a company procedure.  I recall a module that would allow you to view what a partner was being follower of, so that you could quickly remove him the privileges related to his previous company.

A company holds commercials transactions with other companies, or with individuals acting as the direct receiver of that transaction. Not keeping the partner directly associated to the transaction is IMHO incorrect.

As for the proposed partners in the sales order or invoice, I do not really understand. 

In my previous experience in SAP, you had the following partners in a sales order:
- Sales To (the partner with whom you sign the contract and have an obligation. In a sales order the main partner is the "Sales To".)
- Bill To (that's the physical address that will receive the invoice)
- Payer (that is the partner that will be responsible for payment. The journal entry for Accounts Payable is associated to the payer)
- Ship To
- Contact person (optional)

Then, in the invoice:
- Payer (In an invoice the main partner is the "Payer". Is the partner that has an obligation to pay the invoice, and is the one reflected in the accounts payable journal item)
- Bill To
- Contact person (optional)

Generally it's not a good practive to maintain the Ship To in the invoice header. The reason for this is - you can combine multiple deliveries (with different Ship-tp-Party) in a single Invoice. Thus in case like this one, Ship-to Party would be different for some line items in Invoice. 

So Ship-to-Party should be best available only at Line Item Level in Invoice.

rafaelbn:
Well of course you can try it, but it's not my experience only, it's how Odoo implement Odoo and what Odoo says. Contact is not a contact, is a **contact of a company** and if this person move to another company it's a new contact of a company. Check what Odoo says:



This is not enough:

> I recall a module that would allow you to view what a partner was being follower of, so that you could quickly remove him the privileges related to his previous company

What about mailing list, event registry, surveys, opportunities, meetings, calls, ... ?

If you try to change this it doesn't look to be a general approach in my opinion but specific one.

pedrobaeza:
You mustn't change the company of the contact. You create another contact with the proper data (for sure this person will have a new email, and maybe a new phone number), and deactivate the old one.

pedrobaeza:
Closing this one following the discussion.

PR #254, opened 2015-12-28
sale_automatic_workflow : create invoice for sale with only service

laetitia-gangloff:
When sale automatic workflow is configured to create invoice on picking done, in case of a sale order with only service product, there is no picking created and so no invoice created.

I propose, to create invoice if there is only product service when configuration is on picking done.


guewen:
Including a test, neat :)
 :+1:

lmignon:
:+1:

lmignon:
Ping @pedrobaeza This one can be merged.. Thank you in advance :smirk:

pedrobaeza:
:+1:

PR #256, opened 2016-01-21
[8.0] proposal for new addon: rental products used in product sets

rschnapka:
This is to set usable default values to rental products added to sale orders via product-set.


explanar:
You can already rent Product-Sets!

Create a new Product (not Rental!), Add Bill of Material with Type "Set", Add multiple Products to the set at Bill of Material.

Now Create Rental service from the new Product,
The delivery paper should contain the single parts of the set. Doesn't it?

rschnapka:
I agree with the you, the aspect of building a BOM is possible and is reproducable. The module product-set could arguably be replaced by this functionality (BOM) then as well. Adding a product set with both methods results in all the products being displayed on delivery orders, afaik. 
The addon/module product set has the advantage of being easily built by import. Having a product list i. e. from a CAD-system you get a product set very quickly which can then be used in the order directly without first having to define a rental-product from it. That was my way of thinking, thought is was a 'nice to have' refinement anyway. If you believe, that this is a duplication of processes, that's fine with me.

rousseldenis:
Closing this as no activity. @rschnapka Please feel free to reopen it if you still need it. Take care of your PR's

PR #265, opened 2016-02-23
[FIX] When pickings are recreated from shipping excepting, order "View Delivery Orders" button disappears

### LINKED ISSUE #264: [8.0][sale_procurement_group_by_line] Recreate pickings doesn't show View Delivery Orders button
Hi Friends,

Just found bug where group_id always created on recreating deliveries from order
## How to reproduce bug:
- Create and confirm the sale order
- Cancel the pickings
- Recreate picking from order
- Once the pickings are recreated the **View Delivery order** is no longer visible. 

Let me know if you guys need any more information on this. 

Regards,
Kinner

kenvac:
- Fix #264
- Fix : Can't create delivery from shipping exception

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -49,7 +49,12 @@ class SaleOrder(models.Model):
             groups = {}
 
             for line in order.order_line:
+                if order.state == 'shipping_except':
+                    groups[line._get_procurement_group_key()] = \
+                        line.procurement_group_id.id
+
                 group_id = groups.get(line._get_procurement_group_key())
+
                 if not group_id:
                     vals = self._prepare_procurement_group_by_line(
                         cr, uid, line, context=context)
@@ -70,6 +75,8 @@ class SaleOrder(models.Model):
                     # trigger another move
                     proc_ids += [x.id for x in line.procurement_ids
                                  if x.state in ('exception', 'cancel')]
+                    procurement_obj.reset_to_confirmed(cr, uid, proc_ids,
+                                                       context=context)
                 elif sale_line_obj.need_procurement(cr, uid, [line.id],
                                                     context=context):
                     if (line.state == 'done') or not line.product_id:


sale_procurement_group_by_line/model/sale.py
```diff
@@ -49,7 +49,15 @@ def action_ship_create(self, cr, uid, ids, context=None):
             groups = {}
 
             for line in order.order_line:
-                group_id = groups.get(line._get_procurement_group_key())
+                if order.state == 'shipping_except':
+                    try:
```
pedrobaeza: Why a try? Use

```
groups.get(line...
```

kenvac: You are right @pedrobaeza . This can be removed. I'll make the change and also need to check why CI is failing.

kenvac:
@pedrobaeza I've made those changes. Thanks for pointing it out.

pedrobaeza:
:+1:

sale_procurement_group_by_line/model/sale.py
```diff
@@ -70,6 +76,7 @@ def action_ship_create(self, cr, uid, ids, context=None):
                     # trigger another move
                     proc_ids += [x.id for x in line.procurement_ids
                                  if x.state in ('exception', 'cancel')]
+                    procurement_obj.reset_to_confirmed(cr, uid, proc_ids, context=context)
```
pedrobaeza: This line is too long:

```
sale_procurement_group_by_line/model/sale.py:79:80: E501 line too long (90 > 79 characters)
```

kenvac:
@pedrobaeza pep8 convention enforced in code

sale_procurement_group_by_line/model/sale.py
```diff
@@ -49,7 +49,13 @@ def action_ship_create(self, cr, uid, ids, context=None):
             groups = {}
 
             for line in order.order_line:
+                if order.state == 'shipping_except':
+                    group_id = getattr(line.procurement_group_id, 'id')
```
StefanRijnhart: This looks like a void statement to me. _group_id_ is not used in the loop, and overwritten right after.

pedrobaeza:
@kenvac, can you check Stefan's comment?

mistotebe:
I've made the change suggested above.

StefanRijnhart:
Thanks for the change!

PR #270, opened 2016-03-03
[ADD][9.0] partner_default_sale_warehouse

JordiBForgeFlow:
In a company that has delivers orders to customers from multiple warehouses
it is important to try to deliver from the closest warehouse in order
to minimize the transportation costs.

This module makes it possible for users to introduce a default delivery
warehouse in partners, what will then be proposed into the sales order,
based on the delivery address partner entered.


pedrobaeza:
It's better to migrate the existing https://github.com/OCA/sale-workflow/tree/9.0/sale_order_type, that includes the warehouse, than making a new one.

JordiBForgeFlow:
@pedrobaeza Thanks Pedro. This module could be helpful. But I'm faced with an issue that in ERP's such as SAP has been resolved since the last 30 years or so. And that is, the concept of the Distribution Channel.

The approach of sale_order_type is great as a method to seggregate distribution channels. E.g. Internet Sales vs. Store Sales. It is clear that a customer placing an internet sales should trigger the delivery from warehouse A, whereas a Store sale should trigger a delivery from the store warehouse.

However, when it comes to the physical / distance to customer viewpoint, 2 customers in very separate and distant locations could do an internet / store sale, but the delivering warehouse should be different, just because they are served by the closest warehouse.

SAP approaches this issue by means of a concept called "Distribution Channel". You can define, for an order type "Internet Sales" a predefined distribution channel "Central Distribution". Then, at the time of completing a customer master data you can enter, for each Distribution Channel in which he'll be involved, what should be the default delivering warehouse.

That would solve the two scenarios. Distinguish the channel in which sales are made, and be able to choose the best delivering warehouse to use for each channel and customer. 

We at Eficent can implement this Distribution Channel feature, that can be nicely complemented with the sale_order_type module (it would require changes to sale_order_type, to substitute the default warehouse with the default Distribution Channel).

Perhaps Logistics PSC can give their point of view @tafaRU @jgrandguillaume

pedrobaeza:
Well, I don't see that to be splitted from sale_order_type, but let's see what other thinks.

And remember you don't have to make SAP inside Odoo, but adapt Odoo to customer needs with the tools provided :wink:

JordiBForgeFlow:
@pedrobaeza Yes, I know that it's not about replicating, but I've been exposed to these cases, and I try to expose all my previous experience in the hope to build a better product. For instance, I can see that in Oracle the order type can default the warehouse as it is proposed by sale_order_type. So you see, benchmarking a feature with competitors is important...

What I'll do is to close this PR as it's worse than sale_order_type. If distribution channel would be required, we can initiate a separate thread.

Thanks!!

JordiBForgeFlow:
@pedrobaeza Just think a company with 15 warehouses that sells through 2 channels (store, online). You'd need 30 sales order types and 15 warehouses.

With the distribution channel concept, you'd need 2 order types and 15 warehouses to model the scenario.

Another approach is that the sale_order_type should only default the warehouse if it's been provided in the order type. If it's not been defaulted, then other defaulting criteria would take place if the conditions of the new criteria are met.

pedrobaeza:
Indeed, your extreme case can be solved through different approaches. Anyway, you can covert them with the same module (or maybe an extension/patch over it). At the end, if you have 15 different choices, you should have at least all of these data covered.

JordiBForgeFlow:
@pedrobaeza 10's warehouses with multiple channels is everyday business in many countries for thousands of companies. Not extreme at all ;)

pedrobaeza:
But I still don't understand what do you want to split with this distribution channel. You already have it as a field from CRM. This is not something you should set via the sale order type, but predefine according your sales order feed (for example, Magento or Odoo e-commerce), and you can already do that, so I only see 10 order types for 10 warehouses.

JordiBForgeFlow:
Depending on the channel that the company intends to use for distribution for this order, a different warehouse should be used. So an ecommerce sale is usually supplied from Distribution Centers, whereas a store/POS order is supplier from the store warehouse. The same customer may be supplied from different warehouse, depending on the location that the goods should be shipped to, and the distribution strategy that the company chooses to use to supply that order.

JordiBForgeFlow:
@pedrobaeza It's a pitty because this module is now formally correct, and contains all the test suite. Discussing with the customer, their use case is simple, and sale_order_type would not be interesting for them at all. So what to do in this case? Should this module be kept in Eficent's public repo? It's not really interesting to OCA? I need some guidance on what is the criteria by which a module would be accepted.

pedrobaeza:
I'm afraid OCA is not accepting overlapping features because a customer thinks that it's more convenience. The solution is to adapt current module to ease the customer likes, or have another module outside OCA. I would try to convince the customer...

JordiBForgeFlow:
@pedrobaeza Thanks. That's good enough reason. I'll keep it in Eficent's basket then, and next time I'll keep a better look at OCA's repositortiy before starting...

rafaelbn:
We had this problem also @jbeficent "I'll keep a better look at OCA's repositortiy before starting", we always start coding and after discover the same o similar and improvable code in OCA. After we change mind and think: "we and our customers are no so special, OCA has big Odoo providers and big companies, they must have something like this", we first look for it and after use, re-user, re-coding or coding! This save time and money :smile: 
Isn't it @antespi ?
Rule: All is done, look for it. Isn't it? Oh my god: code it!

JordiBForgeFlow:
Agreed. IMHO OCA may need to focus on few but strong business growth
ENGINES, over moduled that solve specific problems. Design from start for
flexible growth.

For example, design a strong pricing engine, mrp engine, or flexible
organizational strucuting of a company.

JordiBForgeFlow:
For instance, sale_order_type adds flexiblity for business to grow. Is a
tool. Can be adjusted flexibly to meet different use cases. It's good for
that. Could be better, but it's good direction.

JordiBForgeFlow:
Mis Builder is another excellent example of good design.

Focus on few tools that constitute the foundations of any business. And
make them flexible, so that a company can adapt more quickly, building on
top of these engines. Make the modular, flexible.

PR #271, opened 2016-03-05
[ADD] sales_team_security

pedrobaeza:
# 
# Security rules for sales teams

This module sets different permissions levels for accessing sales and CRM 
records based on the sales team: customers, sales orders, leads, opportunities,
phone calls and sales teams.

It also handles the propagation of the sales team from commercial partners to
the contacts, which standard doesn't make.
# Installation

At installation time, this module sets int all the contacts that have the sales
team empty the sales team of the parent. If you have a lot of contacts, this
operation can take a while.
# Usage

On the user configuration (Configuration > Users > Users), select in the
_Sales Team_ section the option "See only own team". Then, the documents
mentioned before will be filtered out to have only those belonging to the
teams the user belongs to.

This is complementary to the Sales level access, but sometimes can be
incoherent depending on the combination chosen. If you chose "See Own Leads"
on _Sales_ section, marking on unmarking the sales team check will be
irrelevant, because the most restricting level, which the sales one, will
prevail.
# Known issues/Roadmap
- This module is designed for support only sales part, so someone that has
  access to other Odoo parts (for example, an accountant), shouldn't have
  this new permission, or some access errors will be found when seeing invoices
  and other documents. A _sales_team_security_account_ bridge module can be
  done for fixing this case, but not in the case of for example warehouse.


pedrobaeza:
The Travis red is unrelated and happens on _sale_properties_dynamic_fields_ test.

sales_team_security/README.rst
```diff
@@ -0,0 +1,84 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============================
+Security rules for sales teams
+==============================
+
+This module sets different permissions levels for accessing sales records
+based on the sales team: customers, sales orders and sales teams.
+
```
rafaelbn: Does it means that a user only will see customers, sale orders and sale teams in which he is assigned?

rafaelbn:
Go to....:
- Runbot and open a base database
- Install sales_team_security
- Go to sales - sales - customer and open (for ex.) Agrolait
- You will get:

```
Odoo Client Error

Error: NameError: name 'section_id' is not defined

http://3143095-271-7556a7.runbot2.odoo-community.org/web/static/lib/py.js/lib/py.js:370
```

sales_team_security/README.rst
pedrobaeza: Yeah, that's it

pedrobaeza:
Fixed. It should depend on crm.

sales_team_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+records based on the sales team: customers, sales orders, leads, opportunities,
+phone calls and sales teams.
+
+It also handles the propagation of the sales team from commercial partners to
+the contacts, which standard doesn't make.
+
+Installation
+============
+
+At installation time, this module sets int all the contacts that have the sales
```
bealdav: s/int/in

sales_team_security/security/sales_team_security.xml
```diff
@@ -0,0 +1,82 @@
...
+        <field name="name">Sales Team Order Lines</field>
+        <field ref="sale.model_sale_order_line" name="model_id"/>
+        <field name="domain_force">['|', ('section_id', 'in', user.section_ids.ids), ('section_id', '=', False)]</field>
+        <field name="groups" eval="[(4, ref('sales_team_security.group_see_only_own_team'))]"/>
+    </record>
+
+    <record id="res_partner_team_rule" model="ir.rule">
+        <field name="name">Sales Team Partners</field>
+        <field ref="base.model_res_partner" name="model_id"/>
+        <field name="domain_force">['|', ('section_id', 'in', user.section_ids.ids), ('section_id', '=', False)]</field>
```
OSguard: our first solution was to write `['|',('user_id','=',user.id),('section_id.member_ids','child_of',user.id)]` but as you add section_ids to res.users this maybe can also be fast. (Our messure: 100k res.partners, 25 sales.teams, 50 res.users). Need to try out. And give Feedback.

OSguard:
generell thinks:
- We use the existing groups to modify the configuration of crm.leads to expand it with the sales_team behavior and to add it to the other object. We did this to avoid miss configuration, so you can configure "all leads" in sales, and then "sales team" has noch effect on crm.lead.
- i didn't test it, but i think adding a domain on res.partner makes you a trouble on some other parts. For example that you can not edit your own settings. So you need the following domain to all users:
  `['|','|',('user_ids', 'in', user.id),('parent_id','=',user.company_id.id),('id','=',user.company_id.id)]`

OSguard:
some other points come in my mind
- you should at least document that "no sales_team" means that in this data is shared with everyone. Our customer wants the opposite arround, that contacts are then not seen from restricted users. The best solution is to have it configurable. I don't want to discuss what is better, but it should be documented in the module description.
- `('user_id','=',user.id)` we add to each domain. That you can have a object assign to you and not seen to any team, but not sharing with everyone. This is very helpful on transition, because if you install the module and you have already data, that you need
- `[('message_follower_ids', 'child_of', user.partner_id.id)]` we try this that you can add other user to have access to objects you follow, but this was slow as hell

sales_team_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+team empty the sales team of the parent. If you have a lot of contacts, this
```
elicoidal: of the salesman

pedrobaeza: No, I mean sales team, not salesman.

elicoidal: Yes I saw it later.

victor-ditay:
@pedrobaeza 
I try this module and I found and error. 

The story is:

I create two users with only the permissions "See Own Leads" and "See only own team".
Log in the system with one user. Go to Sales application. Create new Sales Order. Try to save it and I got an error.
Also I test it with the combination "See All Leads" and "See only own team". There is no trouble to create Sales Orders.

For me It seems related with the field `section_id`

Odoo Server Error
......
  File "/opt/odoo/sources/odoo/openerp/osv/expression.py", line 830, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'section_id' in leaf "<osv.ExtendedLeaf: ('section_id', 'in', [4]) on sale_order_line (ctx: )>"

About the functionality I have a question about the reason you can select this two groups at the same time. For me `See only own team` should be between `See All Leads` and `See Own Leads`.

cc @elicoidal

pedrobaeza:
Víctor, I have answered you in the other PR you have made.

pedrobaeza:
@OSguard, I answered you all opened questions:

> - i didn't test it, but i think adding a domain on res.partner makes you a trouble on some other parts. For example that you can not edit your own settings. So you need the following domain to all users:
>   ['|','|',('user_ids', 'in', user.id),('parent_id','=',user.company_id.id),('id','=',user.company_id.id)]

No, it doesn't happen because you should have the same sales team as you are allowed.

> - you should at least document that "no sales_team" means that in this data is shared with everyone. Our customer wants the opposite arround, that contacts are then not seen from restricted users. The best solution is to have it configurable. I don't want to discuss what is better, but it should be documented in the module description.

Default behavior in Odoo is the one put here, for example with company_id field: no company means all companies can see it; a company selected means that only that company sees it. So there's no need to document what is the standard.
- ('user_id','=',user.id) we add to each domain. That you can have a object assign to you and not seen to any team, but not sharing with everyone. This is very helpful on transition, because if you install the module and you have already data, that you need

Well, this is sales_team_security, not salesman security, so that you have to make it with the combination of the groups in your installation.
- [('message_follower_ids', 'child_of', user.partner_id.id)] we try this that you can add other user to have access to objects you follow, but this was slow as hell

You don't need the `child_of` operand, but `=`, because you only need an exact match. Anyway, I consider that as a custom modification, because not everyone that needs this security rules needs it, or uses followers for other tasks.

OSguard:
> No, it doesn't happen because you should have the same sales team as you are allowed.

the res.partner of your company can only belong to one sales team. And this is synct to your contact. So having different sales teams with your coworker. The only solution is that you defined one sales team 'internal' and add all your employess.

pedrobaeza:
No, that's the default sales team. This takes into account the sales team you are member of.

pedrobaeza:
cc @Tecnativa

rafaelbn:
Rebuilding for testing.

sales_team_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/sale-workflow/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+multi-company/issues/new?body=module:%20
+sale-workflow%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
yajo: Drop this.

sales_team_security/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    "name": "Sales teams security",
+    "version": "8.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": [
+        "sales_team",
+        "sale",
+        "crm",
+    ],
+    "author": "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Odoo Community Association (OCA)",
```
yajo: Tecnativa? website?

sales_team_security/models/res_partner.py
```diff
@@ -0,0 +1,40 @@
...
+            submenu=submenu)
+        if view_type == 'form':
+            eview = etree.fromstring(res['arch'])
+            xml_fields = eview.xpath("//field[@name='child_ids']")
+            if xml_fields:
+                context_str = xml_fields[0].get('context', '{}')
+                context_str = (
+                    context_str[:1] + "'default_section_id': section_id, " +
+                    context_str[1:])
+                xml_fields[0].set('context', context_str)
```
yajo: Can't you do all of this in a view? Or at create or write?

sales_team_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+On the user configuration (Configuration > Users > Users), select in the
+*Sales Team* section the option "See only own team". Then, the documents
+mentioned before will be filtered out to have only those belonging to the
+teams the user belongs to.
+
+This is complementary to the Sales level access, but sometimes can be
+incoherent depending on the combination chosen. If you chose "See Own Leads"
+on *Sales* section, marking on unmarking the sales team check will be
+irrelevant, because the most restricting level, which the sales one, will
+prevail.
```
yajo: Wouldn't it make sense to put this module's groups in the same hierarchy as the core sales groups? This way, you can only choose one option...

captura de pantalla de 2016-08-26 17-52-41

sales_team_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Known issues/Roadmap
+====================
+
+* This module is designed for support only sales part, so someone that has
+  access to other Odoo parts (for example, an accountant), shouldn't have
+  this new permission, or some access errors will be found when seeing invoices
+  and other documents. A *sales_team_security_account* bridge module can be
+  done for fixing this case, but not in the case of for example warehouse.
```
yajo: Probably fixing https://github.com/OCA/sale-workflow/pull/271/files#r76442451 would make this behavior more expected, if ever possible.

sales_team_security/models/res_users.py
```diff
@@ -0,0 +1,13 @@
...
+# © 2016 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    section_ids = fields.Many2many(
```
yajo: Forward-looking, I'd call it `team_ids` or `sale_team_ids`.
https://github.com/odoo/odoo/commit/19551aec2ee3bf5e4c14c00058e6b41fea6dd429

sales_team_security/models/res_users.py
```diff
@@ -0,0 +1,13 @@
...
+        comodel_name="crm.case.section", string="Sales teams",
+        relation='sale_member_rel', column1='member_id', column2='section_id')
```
yajo: How about `user_id` and `team_id`?

sales_team_security/security/sales_team_security.xml
```diff
@@ -0,0 +1,82 @@
...
+        <field name="name">Sales Team Phone Calls</field>
+        <field ref="crm.model_crm_phonecall" name="model_id"/>
+        <field name="domain_force">['|', ('section_id', 'in', user.section_ids.ids), ('section_id', '=', False)]</field>
+        <field name="groups" eval="[(4, ref('sales_team_security.group_see_only_own_team'))]"/>
+    </record>
+
+    <record id="crm_phonecall_report_team" model="ir.rule">
+        <field name="name">Sales Team Phone Calls Analysis</field>
+        <field ref="crm.model_crm_phonecall_report" name="model_id"/>
+        <field name="domain_force">['|', ('section_id', 'in', user.section_ids.ids), ('section_id', '=', False)]</field>
```
yajo: I feel this rule and all of the above are just great! They allow you to split your database in a very sensible and powerful way for most salesmen.

However, there's still a problem: What if for any reason I need user from team B see a record from team A? You probably do not want to add him/her to a team just to see a phonecall, partner, etc.

I think all these rules should be expanded with `[('message_follower_ids', 'in', user.partner_id.ids)]` to be able to make any user a follower and s/he gets instant access to that single record.

StefanRijnhart:
Setting to question as per @yajo's review.

sales_team_security/models/res_partner.py
pedrobaeza: No, this should be done this way for assuring maximum compatibility.

sales_team_security/README.rst
pedrobaeza: No, this is about people not having sales group, but accessing the sales order via "Accounting".

sales_team_security/README.rst
pedrobaeza: Well, maybe you can be manager, so that you can have access to Configuration part, but only seeing your team's documents. Do you think it can be possible? The other option is to have 2 groups "See only own team" and "Manager seeing only own team".

sales_team_security/models/res_users.py
pedrobaeza: No, we can't change this, as the table already exists, and have these column names.

sales_team_security/README.rst
pedrobaeza: @rafaelbn, what do you think?

pedrobaeza:
@StefanRijnhart @rafaelbn @Yajo comments attended or answered. Please review.

sales_team_security/models/res_partner.py
```diff
@@ -20,10 +20,9 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
             eview = etree.fromstring(res['arch'])
             xml_fields = eview.xpath("//field[@name='child_ids']")
             if xml_fields:
-                context_str = xml_fields[0].get('context', '{}')
-                context_str = (
-                    context_str[:1] + "'default_section_id': section_id, " +
-                    context_str[1:])
+                context_str = xml_fields[0].get('context', '{}').replace(
+                    '{', "{'default_section_id': section_id, "
```
yajo: Please add `count=1`.

yajo (changes requested):

rafaelbn:
Rebuilding for testing

pedrobaeza:
You can make the final review now.

cc @Yajo @rafaelbn @chienandalu

pedrobaeza:
Please review this for finally merging this @chienandalu @luismontalba

luismontalba:
I go on runbot and create a new user with this permissions:

seleccion_001

I assign a sales team to my new user:

seleccion_002

I log in as the new user (demo2 + demo2)

I go to sales > Leads and I can see other sales team leads.

seleccion_003


Aren't other sales team leads expected to be hidden?

pedrobaeza:
@luismontalba, you have to add it to the sales team members, not to define the sales team in the user preferences. This is because this way, you can be part of several teams, but the sales team field on users only allows one team. This field should be called main sales team. For your reference, in future versions, this field has been even removed, as there's no real utility for it.

luismontalba (approved):
Functionality tested on runbot. It works by this way.

chienandalu:
I've tested on runbot and functionality is ok. Although I get two different errors with sales orders.

1. With a user I created from a duplicate of demo user and the "See only own team" config:
  - When I try to enter in the sales detail o create a new one I get an error.

2. (maybe not related) Logged in as admin: I go to a sale order and try to duplicate it. This error pops:

```
  File "/.repo_requirements/odoo/openerp/http.py", line 544, in _handle_exception
  File "/.repo_requirements/odoo/openerp/http.py", line 581, in dispatch
  File "/.repo_requirements/odoo/openerp/http.py", line 317, in _call_function
  File "/.repo_requirements/odoo/openerp/service/model.py", line 118, in wrapper
  File "/.repo_requirements/odoo/openerp/http.py", line 314, in checked_call
    return self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/openerp/http.py", line 810, in __call__
  File "/.repo_requirements/odoo/openerp/http.py", line 410, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-8.0/addons/web/controllers/main.py", line 944, in call_kw
  File "/home/odoo/odoo-8.0/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 268, in wrapper
    return old_api(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 439, in old_api
    result = new_api(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 443, in new_api
    result = [method(rec, *args, **kwargs) for rec in self]
  File "/home/odoo/build/OCA/sale-workflow/sale_quotation_number/model/sale_order.py", line 33, in copy
    return super(SaleOrder, self).copy(default=default)
  File "/.repo_requirements/odoo/openerp/api.py", line 266, in wrapper
    return new_api(self, *args, **kwargs)
TypeError: copy() got an unexpected keyword argument 'default'
```

pedrobaeza:
@chienandalu check the code and you can see that the problem is in module sale_quotation_number, not for this PR. Take into account that runbot is built with all the current branch modules plus this PR.

pedrobaeza:
Other thing: for quoting multi-line texts, you need to put triple quote ```. I have done it in your comment for you to see it.

chienandalu (approved):
thanks, @pedrobaeza 

then everything else is ok

pedrobaeza:
Everything is attended and this is an old one, so I merge.

PR #274, opened 2016-03-08
[8.0] Migration of sale_dropshipping

mohamedhabibchallouf:
- ```
        <field name="inherit_id" ref="stock.view_picking_internal_search"/>
  ```
- change sale.py /delete import netsv
  add installable=True in **openerp**.py


pedrobaeza:
Dropshipping exists by default in Odoo, so this module should be removed. Why do you need it?

StefanRijnhart:
:-1: Makes sense to replace the contents of this module with only a migration script to the new dropshipment functionality in core Odoo 8.0.

PR #275, opened 2016-03-08
[FIX] fix perf issue with sale_order_lot_selection

sebastienbeau:
Hi, first thank to Agilebg for this module

I have some perf issue with sale_order_selection_lot. Indeed my customer have more than 4000 lot for a product and when he try to select it in a sale order the function spend 3 min to compute the lot available.
With this change it spend less than 0,023s
What do you think?
@hurrinico @eLBati 

Thanks for you're feedback

### DIFF
--- a/sale_order_lot_selection/model/sale.py
+++ b/sale_order_lot_selection/model/sale.py
@@ -44,24 +44,16 @@ class SaleOrderLine(models.Model):
             warehouse_id=warehouse_id)
 
         available_lots = []
-        lot_model = self.env['stock.production.lot']
-        product_model = self.env['product.product']
-        product_obj = product_model.browse(product)
         location = self.env['stock.warehouse'].browse(
             warehouse_id).lot_stock_id
-        lots = lot_model.search(
-            [('product_id', '=', product)])
-        for lot in lots:
-            # for the selected product, search for every associated lot
-            # for every lot, check if it is available (in location.id)
-            # if it is, add it to selectable lots
-            qty_res = product_obj.with_context({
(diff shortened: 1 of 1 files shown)


sale_order_lot_selection/model/sale.py
```diff
@@ -44,24 +44,19 @@ def product_id_change_with_wh(
...
-                'location': location.id,
-                })._product_available()
-            if qty_res[product]['qty_available'] > 0:
-                if lot.id not in available_lots:
-                    available_lots.append(lot.id)
+        locations = self.env['stock.location'].search([
+            ('id', 'child_of', location.id),
+            ])
+        # Search all lot existing lot for the product and location selected
+        self._cr.execute("""SELECT lot_id
```
pedrobaeza: You can use quants for searching, but without SQL. The performance will be for sure still good. You can use read_group function if you want to increase it. This way, you don't overpass multi-company rules, and so on.

sebastienbeau: good idea, let's fix it

sebastienbeau:
@pedrobaeza everything is green and use read_group ;)

pedrobaeza:
Thanks :+1:

hparfr:
:+1: 
Less CPU / Memory consumption is better for Earth :evergreen_tree:  :earth_africa:

bealdav:
:+1:

eLBati:
Thanks :smile:

hurrinico:
Really Thank You @sebastienbeau :+1:

PR #276, opened 2016-03-11
8.0 sales team security fix error on sale order line

victor-ditay:
@pedrobaeza I find an error when a sales person try to create a sales order with the configuration groups See Own Leads and See only own team.

cc @elicoidal


pedrobaeza:
Please make the PR over my branch instead of proposing a new PR. I close this one.

pedrobaeza:
Anyway, seeing the patch, it's not valid. That field is added by module _crm_. Look if you have an old module version (I changed the dependencies).

PR #279, opened 2016-03-16
[MIG][9.0] sale_exception

SodexisTeam:


sale_exception/sale.py
```diff
@@ -1,29 +1,11 @@
...
-#
-#
+# © 2011 Raphaël Valyi, Renato Lima, Guewen Baconnier, Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import time
 
 from openerp import api, models, fields
-from openerp.exceptions import except_orm
+from openerp.exceptions import Warning
```
gurneyalex: in v9, `exceptions.Warning` is deprecated. Please use `exceptions.UserError` instead.

sale_exception/sale.py
```diff
@@ -188,8 +163,7 @@ def _rule_eval(self, rule, obj_name, rec):
                       mode='exec',
                       nocopy=True)  # nocopy allows to return 'result'
         except Exception, e:
-            raise except_orm(
-                _('Error'),
+            raise Warning(
```
gurneyalex: UserError

gurneyalex:
Thanks for porting this module :smile_cat: 

With the removal of sale.order workflow, I think adding an automated test is needed, and I think a number of additional case need to be covered. For instance, in v9 it is possible to edit a confirmed sale.order and I'm not sure this is covered by the current implementation.

atchuthan:
@gurneyalex 
yes, I would agree on checking for sales exception when order is modified after sales confirmation.

As creation of new SO lines creates its own procurement and DO because of _action_procurement_create().  How do I restrict this functionality if the SO line has an exception? or Should I update the list of exceptions only at write() event?

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Sale Exceptions
```
elicoidal: @SodexisTeam   Thanks for your contribution to the OCA. 
Please follow the principles: 
- Contribution guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
- Please follow the OCA README template:
- We recommend to include a screenshot as well to help new comers (nice to have)
- Please use simple headers in all py files following OCA templates
  https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

elicoidal: in particular, document title should have a line before (same length as title)

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+===============
+
+This module allows you attach several customizable exceptions to your
+sale order in a way that you can filter orders by exceptions type and fix them.
+
+This is especially useful in an order importation scenario such as with
```
elicoidal: s/order importation scenario/scenario for mass sales order import

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+the base_sale_multi_channels module, because it's likely a few orders have
```
elicoidal: s/a few/some

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+errors when you import them (like product not found in Odoo, wrong line
+format etc...)
```
elicoidal: period

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Yannick Vaucher <yannick.vaucher@camptocamp.com>
+* SodexisTeam <dev@sodexis.com>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
```
elicoidal: https

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+===============
+
+This module allows you attach several customizable exceptions to your
+sale order in a way that you can filter orders by exceptions type and fix them.
+
+This is especially useful in an order importation scenario such as with
+the base_sale_multi_channels module, because it's likely a few orders have
+errors when you import them (like product not found in Odoo, wrong line
+format etc...)
+
```
elicoidal, line 16: I miss debugger info section (check template)

sale_exception/README.rst
```diff
@@ -0,0 +1,41 @@
...
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

jgrandguillaume:
Hi,

For the record, on the contributor ML:

```
(from @guewen )
```

sale_automatic_workflow:
    At the very least, the module must behave as of today, meaning that
    when you open a sales order it generates/validates the invoice, close
    pickings... My feeling is that all or most of the automatism are still
    valid, as they were triggered by the transition from draft→open and
    this transition still exists. The new thing in 9.0 is the possibility
    to edit the open orders and this is what raise the most questions,
    should we continue to automatically create and validate invoices or
    close pickings on further editions? I tend to say yes but I don't have
    a deep understanding of these open orders editions.

=> +1, changing a confirm order must automate the action like if it was the initial order (e.g. adding a line should invoice it automatically, increasing quantitiy must invoice the difference, etc..)

Regards,

atchuthan:
@gurneyalex @pedrobaeza 
I added a test file to check sale_exception functionality.

In case of modifying confirmed SO, we used @api.constrains to check whether there are any sale exception and raised the list of exception name with ValidationError.  Thus, halting the process of creating its procurement and stock move.

We can process the order only after sales order's ignore_exception is set to True, thus bypassing the check at @api.constrains.

guewen:
Thanks :+1:

gurneyalex:
:+1:

PR #282, opened 2016-03-19
[8.0][IMP][sale_service_project] Apply pricelist when calculating sale order line price

antespi:
This PR improves sale_service_project addon:
- Calculate works and materials price applying pricelist
- Better layout for sale order PDF report, printing prices when `compute_price  = True`


sale_service_project/models/sale.py
```diff
@@ -158,5 +167,21 @@ class SaleOrderLineTaskMaterials(models.Model):
...
+
+    @api.multi
+    def render_html(self, data=None):
+        report_obj = self.env['report']
+        report = report_obj._get_report_from_name('sale.report_saleorder')
+        docargs = {
+            'doc_ids': self._ids,
+            'doc_model': report.model,
+            'docs': self,
+            'formatLang': rml_parse(
```
pedrobaeza: You don't need to inject `formatLang` function. `t-field` already handles the format put on the field definition.

antespi: I know, but  the price is calculated into a variable in Qweb, so I have to use `t-esc` and `formatLang` for formatting.

Another approach is having the price computed in the work and material models. Maybe that way is better.

pedrobaeza: Yeah, I think it can be a better approach. Don't make it stored if it's only used here, but use this option instead of overwriting the parser.

antespi:
rebase done

pedrobaeza:
Superseeded by #288

PR #283, opened 2016-03-21
[8.0] sale_payment_method_automatic_workflow: fix reconciliation when different currencies are used

naglis:
Hello all,

we've had some issues with invoice reconciliation in the `sale_payment_method_automatic_workflow` module: when invoice currency is different from company currency the invoice is not reconciled.

After some debugging we've found that the `balance` on line [sale_payment_method_automatic_workflow/model/account_invoice.py#L141](https://github.com/HBEE/sale-workflow/blob/8.0/sale_payment_method_automatic_workflow/model/account_invoice.py#L141) is always non-zero, because `res_payment['total_amount_currency']` is negative. After fixing that, we've also noticed that the method `_prepare_write_off` on [sale_payment_method_automatic_workflow/model/account_invoice.py#L84](https://github.com/HBEE/sale-workflow/blob/8.0/sale_payment_method_automatic_workflow/model/account_invoice.py#L84) calls a non-existing method `get_write_off_information` on `res.company`. I was unable to locate the method being added in any of the _OCA/sale-workflow_ modules. The only reference of such method is found in the [`account_gain_loss`](https://github.com/akretion/account-gain-loss/blob/7.0/account_gain_loss/res_company.py#L47) module.

We've decided to remove the call to this method and instead we directly take the currency exchange expense/income account from `res.company`.

I would appreciate any comments on how to improve this PR.

### DIFF
--- a/sale_payment_method_automatic_workflow/model/account_invoice.py
+++ b/sale_payment_method_automatic_workflow/model/account_invoice.py
@@ -84,20 +84,18 @@ class AccountInvoice(models.Model):
     def _prepare_write_off(self, res_invoice, res_payment):
         self.ensure_one()
         if res_invoice['total_amount'] - res_payment['total_amount'] > 0:
-            writeoff_type = 'expense'
+            account = self.company_id.expense_currency_exchange_account_id
         else:
-            writeoff_type = 'income'
-        writeoff_info = self.company_id.get_write_off_information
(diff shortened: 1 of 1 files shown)


msteinfeld-intero:
Hello,

are there any news here?

kikopeiro:
:+1:

PR #284, opened 2016-03-22
[FIX] sale_order_type: load refund journal when invoice type is 'out_refund'

agaldona:
When an invoice of type 'out_refund' is created it takes the Sale Journal instead of the Refund Journal defined in the sale order type. So with this change we fix that error.


pedrobaeza:
:+1:

pedrobaeza:
@oihane @anajuaristi, please check this

oihane:
LGTM :+1:

PR #287, opened 2016-04-04
[FIX] pass proper context to ir_sequence.next_by_id()

JosDeGraeve:
A proper sequence object is created with the context 'fiscalyear_id', but
the default context is used in the old-api-style call.

When using the OCA module 'account_auto_fy_sequence', this leads to an error
'The system tried to access a fiscal year sequence without specifying the actual fiscal year.'

### DIFF
--- a/sale_payment_method/sale.py
+++ b/sale_payment_method/sale.py
@@ -140,21 +140,15 @@ class SaleOrder(models.Model):
 
     @api.model
     def _get_payment_move_name(self, journal, period):
-        sequence = journal.sequence_id
+        sequence = journal.sequence_id.with_context(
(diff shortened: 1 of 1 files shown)


sale_payment_method/sale.py
```diff
@@ -153,7 +153,7 @@ def _get_payment_move_name(self, journal, period):
         sequence_model = self.pool['ir.sequence']
         name = sequence_model.next_by_id(self.env.cr, self.env.uid,
```
pedrobaeza: Instead of this, you only need:

```
name = sequence.next_by_id(sequence.id)
```

pedrobaeza:
:+1:

atchuthan:
:+1:

PR #289, opened 2016-04-19
ADD sale_supplier_service

florian-dacosta:
It is a quite simple module that allow to create purchase order automatically for service products when they are sold.
This was possible in version 6.1 and 7, just setting 'make to order' and 'buy' on the product form.
But in version 8, no procurement are created when service are sold. The routes are not even available on product form.

I am not sure the name of the module is apropriate and I am open to change it of course.

One use case : 
A company selling goods which need to be installed at the customer's house but subcontract the installation service.


pedrobaeza:
Look at https://github.com/odoomrp/odoomrp-wip/tree/8.0/procurement_service that makes the same. The name looks to me better. You can also check the code to compare.

florian-dacosta:
Thanks Pedro, I had not seen this module.
But I think I miss something.
I just tested a simple case with one step delivery and another with 3 step delivery and the service product is present in the picking(s).
I guess it is not the normal behavior.

That was partly to avoid this problem that my approch was to configure a pull rule directly on the product, and not routes.

pedrobaeza:
Are you saying with the module I linked? I don't know too much about that module. It was done mostly by AvanzOSC. @anajuaristi, can you answer this?

florian-dacosta:
Yes, I am talking about the module procurement_service.

anajuaristi:
Service products should not create stock_moves in any case so you are right @florian-dacosta

We will check our module and if it is like you say it would be a bug.

florian-dacosta:
@anajuaristi Any news on this?
Thanks!

StefanRijnhart:
ping @anajuaristi about the behaviour of procurement_service

alfredoavanzosc:
@anajuaristi .  The module "procurement_service" does not create a movement of stock.

PR #290, opened 2016-04-20
Add sale_documents_comments

eLBati:
# 
# Comments for sale documents (order, picking and invoice)

With this module you can add specific comments to a customer, for sale order,
delivery order and invoices. Part of the info will be passed from one to other.
Those data will be automatically added to each item.

---

Ported from https://github.com/odoomrp/odoomrp-wip


sale_documents_comments/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================================================
```
elicoidal, line 5: @elbati  Thanks for your contribution to the OCA. 
Let me introduce to you some basic information about contribution in OCA and I will review other points in more comments inside the code..
In general, please follow the principles: 
- Contribution guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
- Please follow the OCA README template:
- We recommend to include a screenshot as well to help new comers (nice to have)
- Please use simple headers in all py files following OCA templates
  https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3
- Please sort all imports in alphabetical order following: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

sale_documents_comments/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+        "sale",
+        "stock_account",
+        "stock",
+        "account",
+    ],
+    "author": "OdooMRP team, "
+              "AvanzOSC, "
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Odoo Community Association (OCA)",
+    "category": "Custom Module",
```
elicoidal: sales

sale_documents_comments/README.rst
```diff
@@ -0,0 +1,58 @@
...
+Comments for sale documents (order, picking and invoice)
+========================================================
+
+With this module you can add specific comments to a customer, for sale order,
+delivery order and invoices. Part of the info will be passed from one to other.
+Those data will be automatically added to each item.
+
+Usage
```
elicoidal, line 13: Please describe a simple workflow for usage

sale_documents_comments/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: simple headers (all python files)

elicoidal:
few details, other than that LGTM

eLBati:
@elicoidal thanks, I updated README and manifest

I think copyright headers can be kept as they are

elicoidal:
@eLBati Please update headers so that it is consistent with the other OCA modules.

eLBati:
@elicoidal 
changed headers.

However, in general, I think this kind of changes should be a nice to have, not a blocking point

elicoidal:
Thanks @eLBati !
This is currently not nice to have but you could propose the discussion in the contributing repo.
I personally see so many different format that I am happy that it is not optional. This way we are sure copyright is properly documented

eLBati:
@elicoidal if you are referring to
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#it-makes-sense-to-be-picky-in-the-following-cases

> Issues with license, copyright, authorship

It seems to me this point refers to **issues** where headers are not clear or infringements are present or so... In the case of this PR, copyright headers were clear enough, what do you think?

elicoidal:
@eLBati This is part of it and part based on the OCA templates that have been followed systematically (eg: https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3).
If you ask me I am fully in favor of standard headers (it is a kind of branding and allows users checking our module to have the information always at the same place, like README, openerp.py and copyright) and would not change it.
As said, you can propose a change in the interpretation and have a public debate about because there might be as well good reasons not to follow it (reduce first contributor roadblock for example).

JordiBForgeFlow:
When this module defines a field customer comments that is propagated across entities (from customer to sales order, to picking, to invoice) it does not take into account what is the intended audience of the note. In our opinion this can lead to significant confusion in the overall process. 

For example a customer has specific requirements on how the products should be delivered to them (specific instructions), and also specific invoicing requirements. With this model a user would use the single field 'Propagated comments' to include both, and thus logsitic and invoicing notes would be mixed. The warehouse may want to print the delivery-specific requirements on the packing list, but then it would not be able because this field could include non-logistic notes.

In our opinion, there should be either model-specific requirement notes that are propagated by default: Picking requirements: customer -> sales order -> Picking 
Invoicing requirements: customer -> sales order -> Invoice
Order requirements: customer -> sales order

We have also been brainstorming on the idea of using a generic object to store notes, that could be 'plugged' into any model. See: https://docs.google.com/document/d/1uHPphDd4BJEvbbYNw5qkvTyzPlcXB0l6dMWv1So3e1g/edit#
cc @aheficent

eLBati:
Thanks @jbeficent , I'm reading the document.
Taking for example

> if the notes are changed in the customer, they must instantly be visible in the pickings that have already been created

Every picking created in the past should be modified? This sounds wrong to me

JordiBForgeFlow:
@eLBati Please comment on the document itself. not here.

sale_documents_comments/models/partner.py
```diff
@@ -0,0 +1,53 @@
...
+    picking_propagated_comment = fields.Text(
+        string='Propagated comments for pickings')
+    invoice_comment = fields.Text(string='Comments for invoices')
+
+    def _get_sale_comments(self):
+        comment_list = []
+        pcomment_list = []
+        if self.sale_comment:
+            comment_list.append(self.sale_comment)
+        if self.parent_id.sale_comment:
```
bealdav, line 26: What if  `self.parent_id == False` ?

sale_documents_comments/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================================================
+Comments for sale documents (order, picking and invoice)
```
bealdav, line 6: - technical module name (folder name) according guidelines must be in singular form: i.e. sale_document_comment.
  Technical name is repeated in other modules for many purpose like inherit views: singular is easier to memorize. Else is it ? sale_document_comments, sale_documents_comments, sale_documents_comment 
- Why friendly module name is so far from technical name (Comments for sale documents => sale_documents_comments)
- maybe consider name as: `sale_workflow_comment` to express in 1 word: sale, picking, invoice

sale_documents_comments/models/partner.py
gurneyalex: @bealdav this is not an issue, this class uses the new API.

sale_documents_comments/models/stock.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @api.one
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        picking_com, picking_pcom = self.partner_id._get_picking_comments()
+        self.sale_comment = picking_com
+        self.sale_propagated_comment = picking_pcom
+
+    @api.model
+    def create(self, values):
```
gurneyalex: I think you also need to  use the @api.return decorator like this:

```
@api.returns('self', lambda value: value.id)
```

gurneyalex (commented):
can you check the inline comment about the create() method decorators?

sale_documents_comments/models/stock.py
eLBati: @gurneyalex do you in which cases avoiding to use it would raise problems?

gurneyalex: you get very weird errors in a mixed old / new API overloading scenario. Say someone creates a new addon with a model with `_inherit` from yours and writes it with the old api. He may call `super(...).create(...)` and not get an integer as return value, which will very likely cause a crash. This can be safely removed in v10.

eLBati:
@gurneyalex thanks, changed and rebased

gurneyalex:
@eLBati you should rename your branch. The runbot expects <target-branch>-<description> to know on which version of Odoo to test -> renaming your branch to "8.0-add_sale_documents_comments" and resubmitting the PR should fix the runbot.

eLBati:
@gurneyalex thanks, moved to https://github.com/OCA/sale-workflow/pull/468

PR #291, opened 2016-04-22
[MIG][9.0] sale automatic workflow

SodexisTeam:
https://odoo-community.org/groups/contributors-15/contributors-39175


sale_automatic_workflow/README.rst
```diff
@@ -0,0 +1,23 @@
...
+  * Set a sales team
+
+- Apply automatic actions:
+  * Validate the order (only if paid, always, never)
+  * Create an invoice
+  * Validate the invoice
+  * Confirm the picking
+
+This module is used by Magentoerpconnect and Prestashoperpconnect.
+It is well suited for other E-Commerce connectors as well.
```
guewen, line 28: Could you apply the OCA's template?  https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

sale_automatic_workflow/automatic_workflow_job.py
```diff
@@ -0,0 +1,122 @@
...
+directly validated and reconciled with the payment, the subworkflow will
+end and send a signal to the sale order workflow.  The problem is that
+the sale order workflow has not yet finished to apply the step 'invoice',
+so the signal of the subworkflow will be lost because the step 'invoice'
+is still not finished. The step invoice should be finished before
+receiving the signal. This means that we can not directly validate every
+steps of the workflow in the same transaction.
+
+If my explanation is not clear, contact me by email and I will improve
+it: sebastien.beau@akretion.com
```
guewen: I think this explanation is irrelevant now that sales workflows are gone and can be removed.

sale_automatic_workflow/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Sale Automatic Workflow',
+    'version': '8.0.0.2.0',
```
lmignon: version must be **9.0.0.0.0**

pedrobaeza: Better 9.0.1.0.0 :wink:

sale_automatic_workflow/sale.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class sale_order(models.Model):
+    _inherit = "sale.order"
+
+    workflow_process_id = fields.Many2one(comodel_name='sale.workflow.process',
+                                          string='Automatic Workflow',
+                                          ondelete='restrict')
+
+    def _prepare_invoice(self, cr, uid, order, lines, context=None):
```
guewen: this method has been changed in 9.0:

```
    @api.multi
    def _prepare_invoice(self):
        ...
```

sale_automatic_workflow/sale.py
```diff
@@ -0,0 +1,88 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+from openerp import models, fields, api, _
+
+
+class sale_order(models.Model):
```
guewen: Should be `SaleOrder` as it uses the new API.

sale_automatic_workflow/sale.py
```diff
@@ -0,0 +1,88 @@
...
+        if not workflow:
+            return invoice_vals
+        invoice_vals['workflow_process_id'] = workflow.id
+        if workflow.invoice_date_is_order_date:
+            invoice_vals['date_invoice'] = order.date_order
+        if workflow.property_journal_id:
+            invoice_vals['journal_id'] = workflow.property_journal_id.id
+        return invoice_vals
+
+    def _prepare_order_picking(self, cr, uid, order, context=None):
```
guewen: `_prepare_order_picking` does no longer exist in 9.0.
You need something like that:

``` python
    _inherit = 'stock.move'

    @api.model
    def _prepare_picking_assign(self, move):
        """ Prepares a new picking for this move as it could not be assigned to
        another picking. This method is designed to be inherited.
        """
        values = super(StockMove, self)._prepare_picking_assign(move)
        if move.procurement_id.sale_line_id:
            sale = move.procurement_id.sale_line_id.order_id
            values['workflow_process_id'] = sale.workflow_process_id.id
        return values

```

lmignon:
It seems that the history of changes is lost... Not sure that we can merge this branch

```
Commits that don’t share a common merge base with the base branch are not supported.
You can view the full commit page here instead.
```

atchuthan:
@lmignon we don't have this module available in version 9.0, so we added the commit histories from 8.0 branch using https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

atchuthan:
FYI, we started working on this PR.  So, could we have "Work in progress" label?

lmignon:
@atchuthan OK thank you for the explanation!

guewen:
> FYI, we started working on this PR. So, could we have "Work in progress" label?

Done!

sale_automatic_workflow/__init__.py
```diff
@@ -0,0 +1,29 @@
+# -*- encoding: utf-8 -*-
```
elicoidal: Please use simplified headers

guewen:
@atchuthan what's your progress? Can I help for the migration?

sale_automatic_workflow/automatic_workflow_job.py
```diff
@@ -0,0 +1,122 @@
...
+                ('workflow_process_id.validate_order', '=', True)]
+
+    @api.model
+    def _validate_sale_orders(self):
+        sale_obj = self.env['sale.order']
+        sales = sale_obj.search(self._get_domain_for_sale_validation())
+        _logger.debug('Sale Orders to validate: %s', sales)
+        for sale in sales:
+            with commit(self.env.cr):
+                sale.action_button_confirm()
```
yvaucher: the method is now: `action_confirm`

sale_automatic_workflow/automatic_workflow_job.py
```diff
@@ -0,0 +1,122 @@
...
+    def _validate_invoices(self):
+        invoice_obj = self.env['account.invoice']
+        invoices = invoice_obj.search(
+            [('state', 'in', ['draft']),
+             ('workflow_process_id.validate_invoice', '=', True)],
+        )
+        _logger.debug('Invoices to validate: %s', invoices)
+        for invoice in invoices:
+            with commit(self.env.cr):
+                invoice.signal_workflow('invoice_open')
```
yvaucher: Workflow have been removed

atchuthan:
@guewen sorry for the late reply as we were on a trip... 
We have made few changes to the code as per the discussions in the community thread.

sale_automatic_workflow/models/account_invoice.py
```diff
@@ -0,0 +1,24 @@
...
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    workflow_process_id = fields.Many2one(
+        comodel_name='sale.workflow.process',
+        string='Sale Workflow Process'
+    )
+    # TODO propose a merge to add this field by default in account module
+    sale_ids = fields.Many2many(
+        comodel_name='sale.order',
```
guewen: The counterpart field on `sale.order` is a computed field so does this `Many2many` work as expected?

`addons/sale/sale.py`:

``` python
     invoice_ids = fields.Many2many("account.invoice", string='Invoices', compute="_get_invoiced", readonly=True, copy=False)
```

guewen: You probably have to go through the invoice lines to get the orders, but in fact I don't think this field is actually used.

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,170 @@
...
+end and send a signal to the sale order workflow.  The problem is that
+the sale order workflow has not yet finished to apply the step 'invoice',
+so the signal of the subworkflow will be lost because the step 'invoice'
+is still not finished. The step invoice should be finished before
+receiving the signal. This means that we can not directly validate every
+steps of the workflow in the same transaction.
+
+If my explanation is not clear, contact me by email and I will improve
+it: sebastien.beau@akretion.com
+"""
```
guewen: This explanation is irrelevant in 9.0, can be removed.

atchuthan:
@guewen I fixed your comments.

stephankeller:
Just a note that we are going to change the way the filters are defined on the automatic workflow window.  The way it is right now, somebody could delete the filter by mistake.  We are going to manage the filters the same way it is done on the "Automatic Action" window.

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,146 @@
...
+        invoices = invoice_obj.search(validate_invoice_filter)
+        _logger.debug('Invoices to validate: %s', invoices.ids)
+        for invoice in invoices:
+            with commit(self.env.cr):
+                invoice.signal_workflow('invoice_open')
+
+    # TODO:
+    # use account_payment_mode to get the payment information
+    @api.model
+    def _register_payments(self, payment_filter):
```
guewen: Should it rather be done in sale_automatic_workflow_payment_mode https://github.com/OCA/sale-workflow/pull/297?

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,146 @@
...
+                self._create_invoices(
+                    safe_eval(sale_workflow.create_invoice_filter_id.domain) +
+                    workflow_domain)
+            if sale_workflow.validate_invoice:
+                self._validate_invoices(
+                    safe_eval(
+                        sale_workflow.validate_invoice_filter_id.domain) +
+                    workflow_domain)
+            # TODO
+#             if sale_workflow.register_payment:
```
guewen: As above, probably to do in `sale_automatic_workflow_payment_mode`?

atchuthan: yes, I agree.
I will create a PR for guewen:add-sale_automatic_workflow_payment_mode

atchuthan: @guewen moved the payment functionality to sale_automatic_workflow_payment_mode and created PR to your fork https://github.com/guewen/sale-workflow/pull/3

atchuthan:
@guewen we can change the label to "Needs review"

guewen:
> @guewen we can change the label to "Needs review"

yeah!

sale_automatic_workflow/views/sale_view.xml
```diff
@@ -0,0 +1,28 @@
...
+      <field name="model">sale.order</field>
+      <field name="inherit_id" ref="sale.view_order_form" />
+      <field name="arch" type="xml">
+        <xpath expr="//page[2]" position="after">        
+          <page name="automation_information" string="Automation Information">
+            <group>
+              <field name="workflow_process_id"/>
+            </group>
+          </page>
+		</xpath>
```
guewen: Tabs spotted

guewen:
There is a small issue with the Configuration menu (same issue in `sale_exception` btw):

selection_032

guewen:
I really like the feature to customize the filters, well done!

Thanks :)

2 small things to fix, but :+1:

guewen:
@atchuthan here's a fix for a corner case during the installation (existing records in `sale.workflow.process`, `default` are called before the filters are created): https://github.com/sodexis/sale-workflow/pull/1

atchuthan:
@guewen thanks for the fix.

gurneyalex:
:+1:

sale_automatic_workflow/README.rst
```diff
@@ -0,0 +1,79 @@
...
+  * Set the invoice's date to the sale order's date
+  * Set a sales team
+
+- Apply automatic actions:
+  * Validate the order (only if paid, always, never)
+  * Create an invoice
+  * Validate the invoice
+  * Confirm the picking
+
+This module is used by Magentoerpconnect and Prestashoperpconnect.
```
rafaelbn, line 27: @pedrobaeza @sergio-teruel take a look for Prestashoperpconnect

rafaelbn:
Nice work thanks. :+1:

rousseldenis:
@atchuthan I try to review the v10 migration, but I need some explanation about the filter (the char field).

Why having removed the char_domain widget ? So, as the char field is not used in code (filter_id is used), why isn't it in readonly?

atchuthan:
@rousseldenis sorry for the delay

first, we had a thought of using the char field directly but later due to time constraint, we worked directly with the filter.
Our idea was to use this char field where the users could setup their custom filter domain instead of making modifications to the filter domain directly.

Let us know if you need any support from our side. Thanks and let us know if you have started PR for v10 migration.

rousseldenis:
@atchuthan The migration to v10 has been started by @lasley 

I want to port this discussion to https://github.com/OCA/sale-workflow/pull/391

PR #294, opened 2016-04-28
[ADD] new module sale_shop_active to add an active field to sale.shop

tremlin:
This PR adds a module which allows to hide `sale.shop`s via an active flag.


StefanRijnhart:
:+1:

sale_shop_active/i18n/sale_shop_active.pot
```diff
@@ -0,0 +1,32 @@
+# Translation of OpenERP Server.
```
pedrobaeza: Remove POT file as it's no longer needed

StefanRijnhart: @pedrobaeza looking good now for you?

pedrobaeza (approved):

PR #299, opened 2016-05-10
[FIX]: Fix bug when adding new picking move. Add list instead of tuple.

lbellier:
Fix bug when adding new picking move.
Add list instead of tuple.

### DIFF
--- a/sale_allotment/__openerp__.py
+++ b/sale_allotment/__openerp__.py
@@ -22,7 +22,7 @@
 
 {
     'name': 'Allotment on sale orders',
-    'version': '8.0.1.1',
+    'version': '8.0.1.1.0',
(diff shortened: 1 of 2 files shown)


clonedagain:
:+1: I naively thought adding a recordset to an existing would not mutate the first recordset (like a list append), but obviously it returns a new recordset instead, so I mustn't use a tuple.

clonedagain:
Can you bump the version to 8.0.1.1.0 please? That should make Travis go green.

lbellier:
@clonedagain : Version bumped to 8.0.1.1.0

leemannd:
:+1:

atchuthan:
:+1:

guewen:
:+1:

PR #301, opened 2016-05-26
8.0 sale payment method auto pay selection

rami-wafaie:


rami-wafaie:
Will create a new PR without conflicts

Issue #267, opened 2016-02-26
Revisions for sale orders Question..need fix


jaronemo:
Dears
1. origin sale order (qty = 50)
   ![image](https://cloud.githubusercontent.com/assets/8476734/13346107/ef1ebb4c-dc9e-11e5-89cc-175f495e5792.png)
2. Delivery Order (qty=50)
   ![image](https://cloud.githubusercontent.com/assets/8476734/13346165/4587dc20-dc9f-11e5-9967-0fda30aa458a.png)

3.cacel sale order and create reversion fix qty to 30
![image](https://cloud.githubusercontent.com/assets/8476734/13346180/6e8c70a4-dc9f-11e5-94ea-4b458cec7cee.png)

4.Delivery Order tree 
![image](https://cloud.githubusercontent.com/assets/8476734/13346191/845cb4a2-dc9f-11e5-9cc8-c344ae80222b.png)

5.Delivery Order Form (qty not 30...is origin qty = 50)

![image](https://cloud.githubusercontent.com/assets/8476734/13346201/9c096370-dc9f-11e5-9ebb-6131d1cbbd5c.png)

pedrobaeza:
Please propose then the same fix

jaronemo:
Dears:

This module issue..problem..is no need fix?

eLBati:
Hello @jaronemo , PR https://github.com/OCA/sale-workflow/pull/306 should fix this

pedrobaeza:
Fixed, so I close.

PR #306, opened 2016-06-04
FIX sale_order_revision: reference to old procurements

eLBati:
FIX https://github.com/OCA/sale-workflow/issues/267

When creating a new revision, remove link to old procurements, otherwise the sale order will use old quantities

### DIFF
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -62,6 +62,9 @@ class sale_order(models.Model):
         self.create_workflow()
         self.write({'state': 'draft'})
         self.order_line.write({'state': 'draft'})
+        # remove old procurements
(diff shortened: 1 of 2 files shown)


sale_order_revision/model/sale_order.py
```diff
@@ -51,6 +51,11 @@ class sale_order(models.Model):
          'Order Reference and revision must be unique per Company.'),
     ]
 
+    def remove_old_procurements(self):
+        for line in self.order_line:
```
pedrobaeza: Better to make:

```
self.mapped('order_line.procurement_ids').write({'sale_line_id': False})
```

eLBati:
@pedrobaeza thanks, done

pedrobaeza:
:+1:

elicoidal:
:+1:

PR #310, opened 2016-06-09
8.0 port sale visible tax

legalsylvain:
Port sale_visible_tax from 7.0 to 8.0
this module simply add total vat incl on each sale order line and account invoice line.
- OCA convention
- new API.

note : 
- set the fields stored.

@ Reviewers : 
- please review only this commit (https://github.com/OCA/sale-workflow/pull/310/commits/ceddd213133ad757e8d775a4a7cc2df64e8d8618). The other is the copy past from V7.0 code.


OSguard:
Theirs is already a module with the same function:
https://github.com/OCA/account-invoicing/tree/8.0/sale_order_line_price_subtotal_gross

legalsylvain:
Hy,
Well in fact sale_visible_tax is older (October 2014) than invoice_line_price_subtotal_gross  (December 2015, https://github.com/OCA/account-invoicing/commits/8.0/account_invoice_line_price_subtotal_gross). What a pity, too much modules and repositories !

But well, the two modules seems better that sale_visible_tax, so I'll use them.

Thanks @OSguard to have pointed the two 8.0 modules !

Could you check this trivial PR to avoid other people to do the same mistake I did.

regards.

https://github.com/OCA/sale-workflow/pull/311

StefanRijnhart:
Closing this as per https://github.com/OCA/sale-workflow/pull/311/files. @legalsylvain thanks for the follow-up!

PR #314, opened 2016-06-22
8.0 sales team security

victor-ditay:


pedrobaeza:
Why are you proposing another PR instead of proposing over my repository (https://github.com/OCA/sale-workflow/pull/271)?

victor-ditay:
@pedrobaeza sorry by that I am trying to fix the history of my commits. I closed it when I realized.

pedrobaeza:
Don't add your copyright over all files, because you haven't changed anything in that files. You should only do it on modified files

pedrobaeza:
Anyway, I see that I have an error, so you can propose the fix to my branch or I can patch myself. What you prefer.

victor-ditay:
@pedrobaeza you can do it if you want.

PR #315, opened 2016-07-01
Disable 'Transfer Picking' on automatic workflow

guewen:
Back in 7.0, this option was not active by default on the automatic
workflow and I think it was better: this option is useful for some use
cases but is not really the _default_ thing we want to do on an
automatic workflow.


atchuthan:
:+1:

elicoidal:
:+1:

florian-dacosta:
:+1:

PR #319, opened 2016-07-29
[9.0] Migration : sale_order_lot_selection

florian-dacosta:
@eLBati


eLBati:
Thank you @florian-dacosta 
looking at [the diff with 8.0](https://gist.github.com/eLBati/7b3eaca4c2796086a6a6aee7fbbe852a) it seems some italian translations are missing.
Did you follow the procedure at https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#if-the-module-doesnt-exist-in-the-90-branch ?

eLBati:
@florian-dacosta tested the following use case:
- make an incoming picking for **iPad Mini**
- process it with lot **123**
- make a sale order selecting **iPad Mini** and lot **123** and confirm it
- make another sale order for **iPad Mini** and lot **123** and confirm it

You can confirm the second sale order

Expected behaviour:
the system should prevent you to confirm a sale order using a reserved lot (like in 8.0)

Side effect: after having confirmed the sale order, in order to process the related picking, you have to manually add quantities to lot **123**

Expected behaviour:
you should be able to validate the related picking just by clicking on **validate**

Thanks

florian-dacosta:
@eLBati 
I did follow the procedure at https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#if-the-module-doesnt-exist-in-the-90-branch
Maybe I made a mistake, or maybe italian translation have been merged after my PR?

About your tests : 
Some things changed from v8 to v9.

For instance, in v9, when the restict_lot_id is filled, Odoo will try to reserve the stock move with a quant of the same lot, and if it does not find any, it will try with a quant without lot.
On the contrary, in v8, it won't reserve a quant without lot.
https://github.com/OCA/OCB/blob/9.0/addons/stock/stock.py#L487

So the module works the way you want if products don't have any stock without lot.
I don't think it is an issue, as, in odoo9, when you choose the option "Tracking : By lots" on the product form, you are forced to put a lot on your incoming picking. So all your product should have a lot.
With demo data, the ipad Mini already has stock without lot, that is why it happens.
If you do a test with a new product, so it does not have stock without lot, it should work as expected.

Regarding the need to validate, just clicking on validate. It is a problem too as Odoo expects the user to indicate how many product he wants to transfer. If product should have lot and the quantity 'done quantity' is 0, it raise an error
https://github.com/OCA/OCB/blob/9.0/addons/stock/stock.py#L1579

For now, the only way I see, to solve this issue, would be to automatically put the qty_done of operations from the reserved quantities when it is created. But it would not impact only the sale orders for which we put a lot, it would impact all stock moves with a restrict_lot_id, and I am not sure it is a good idea.

But it could be a nice feature to have, set the done qty by default, and the user would have to decrease it if needed, instead on increasing it., but I think it would be better in a separated module, as it is not linked to the sale_order_lot_selection main features. And customers could use this feature without using sale_order_lot_selection.
Anyway, if we decide to make this feature in a new module, I don't think sale_order_lot_selection should depend on it. 

What do you think?

eLBati:
@florian-dacosta thanks, ok, then, LGTM :+1:

StefanRijnhart:
:+1:, thank.

PR #329, opened 2016-08-10
[MIG][9.0][sale_product_set] migrate to V9

Olivier-LAURENT:
Migrate to v9, including:
- Some view & menu simplifications or changes
- Some light fixes
- More OCA conventions
- FR tanslations

Rmk: snapshots remain untouched

### DIFF
--- a/sale_product_set/demo/product_set.xml
+++ b/sale_product_set/demo/product_set.xml
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<openerp>
-    <data>
-        <record id="product_set_i5_computer" model="product.set">
-            <field name="name">i5 computer offer</field>
-        </record>
-        <record id="product_set_services"  model="product.set">
-            <field name="name">Services</field>
-        </record>
-    </data>
-</openerp>
+<odoo noupdate="1">
+    <record id="product_set_i5_computer" model="product.set">
+        <field name="name">i5 computer offer</field>
+    </record>
+    <record id="product_set_services"  model="product.set">
+        <field name="name">Services</field>
+    </record>
+</odoo>
--- a/sale_product_set/README.rst
+++ b/sale_product_set/README.rst
@@ -1,6 +1,8 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
+================
 Sale Product set
 ================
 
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -20,6 +20,7 @@ addon | version | summary
 --- | --- | ---
 [sale_automatic_workflow](sale_automatic_workflow/) | 9.0.1.0.0 | Sale Automatic Workflow
 [sale_exception](sale_exception/) | 9.0.1.0.0 | Custom exceptions on sale order
+[sale_product_set](sale_product_set/) | 9.0.1.0.0 | Sale product set
```
pedrobaeza: Don't change this file. It's automatically changed. Please also squash the commits when doing to avoid conflicts.

sale_product_set/demo/product_set.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
```
pedrobaeza: Use `<odoo noupdate="1">` instead of `<openerp>` and `<data>`

sale_product_set/demo/product_set.xml
```diff
@@ -1,11 +1,9 @@
...
-    <data>
-        <record id="product_set_i5_computer" model="product.set">
-            <field name="name">i5 computer offer</field>
-        </record>
-        <record id="product_set_services"  model="product.set">
-            <field name="name">Services</field>
-        </record>
-    </data>
-</openerp>
+<odoo noupdate="1">
```
StefanRijnhart, line 2: Why the noupdate for this demo data?

pedrobaeza:
Are you aware of the more wider approach we are developing in https://github.com/OCA/product-attribute/pull/180?

sale_product_set/demo/product_set.xml
leemannd: Seems strange this noupdate. @Olivier-LAURENT still working on it?

JonathanNEMRY: Isn't a demo data convention to set noupdate="1" for demo data? preventing them to be rested into every -U

JonathanNEMRY (approved):
test+code review

leemannd (approved):

sale_product_set/demo/product_set.xml
StefanRijnhart: @JonathanNEMRY True! Thanks for clarifying, I did not know that.

StefanRijnhart (approved):

StefanRijnhart:
Technically approve. @pedrobaeza is a possible overlap with modules in product-attribute blocking for you?

pedrobaeza:
We can have both as the scope in not exactly the same, so I merge.

PR #273, opened 2016-03-07
8.0 Add sale_rental_lot_selection + several fixes and improvements in sale_rental

alexis-via:

PR #331, opened 2016-08-18
[FIX] populate the 'Return Stock Move' field; module version

jabibi:
When creating the rental record, 'Return Stock Move' field was not
taking the corresponding reference to the incoming stock.move

Also fixed the version number and added Odoo Community Association (OCA)
to author list


pedrobaeza:
@alexis-via, can you please check?

About your inclusion in author list, your patch is not big enough to deserve that credit.

jabibi:
Yeah you are right. Modified

jabibi:
This will continue here #273

PR #332, opened 2016-08-27
8.0 UBL sale stack (import + export) + a module to import CSV orders

alexis-via:
This PR has 6 new modules:
- sale_order_import: base module to import a quote/sale order
- sale_stock_order_import (auto_install=True)
- sale_order_import_ubl: import UBL XML RFQ and PO
- sale_order_import_csv: import CSV orders
- sale_order_ubl: add UBL XML file in PDF quotes and orders
- sale_commercial_partner: small module to add commercial_partner_id on sale.order)

This PR depends on this other PR on account-invoicing: https://github.com/OCA/account-invoicing/pull/122

I still have a few details to finish:
- support some TODO in the code (in particular to handle price-included taxes)
- write some README.rst


StefanRijnhart:
Nice work! Where does base_business_document_import reside?
It seems the phone numbers python module is needed in Travis.

alexis-via:
@StefanRijnhart You point out something which is a mistery for me !
The module base_business_document_import_phone is located in https://github.com/OCA/account-invoicing/pull/122 It depends on 'base_phone' (from OCA/connector-telephony) and 'base_business_document_import' and is auto-installable. As base_phone shouldn't be installed in the test environment, base_business_document_import_phone should not be installed either, so we should not need the phonenumbers lib. And, in the code of base_business_document_import_phone, the import phonenumbers is in a try/except, cf
https://github.com/akretion/account-invoicing/blob/8.0-add-pdf-invoice-import/base_business_document_import_phone/models/business_document_import.py#L8

So I really don't understand why it fails trying to import the phonenumbers lib. Any idea would be greatly appreciated.

alexis-via:
With the help of @sebastienbeau I now understand the problem: the file oca_dependencies.txt of account-invoicing includes connector-telephony and Travis does a recursive dependency resolving, that's why connector-telephony is included in this travis.
So I prepared a PR to protect the import of phonenumbers with try/except:
https://github.com/OCA/connector-telephony/pull/105

alexis-via:
Moved to edi project.

PR #334, opened 2016-09-05
ADD sale_order_type_sale_journal

eLBati:
This module links modules sale_order_type and sale_journal.
That is: you can set an Invoice Type in Sale Order Type that will be used in Sale Order


sale_order_type_sale_journal/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.onchange('type_id')
+    def onchange_type_id(self):
```
pedrobaeza: FYI, instead of calling super, you can also change the name of method to `onchange_type_id_sale_journal` for example and both methods will be called.

eLBati: @pedrobaeza thanks, done

StefanRijnhart:
:+1:
@pedrobaeza good for you as well, I believe?

pedrobaeza (approved):

PR #341, opened 2016-09-28
[ADD] Module to merge draft or confirmed sales orders

StefanRijnhart:


simahawk (approved):

sale_order_merge/README.rst
```diff
@@ -0,0 +1,70 @@
...
+   :alt: License: AGPL-3
+
+=================
+Merge sale orders
+=================
+
+This module allows the sale employee to merge draft or confirmed orders
+from the same customer.
+
+When orders are merged, draft invoices and unprocessed outgoing pickings
```
pedrobaeza, line 12: Why merging pickings? You can leave them unmerged as Odoo standard allows it.

StefanRijnhart: Why else merge the orders? This was the main use case of my customer.

pedrobaeza: Yeah, you're right. You can't merge pickings without merging the "upstream" sales orders, or the links will not be totally correct (although the main control is through sales order lines, not through sales orders, but the main information/reports won't be correct).

sale_order_merge/models/sale_order_merge.py
```diff
@@ -0,0 +1,190 @@
...
+            [('group_id', 'in', other_groups.ids)]).write(
+                {'group_id': group.id})
+        self.env['procurement.order'].search(
+            [('group_id', 'in', other_groups.ids)]).write(
+                {'group_id': group.id})
+        pick_map = {}
+        for picking in self.sale_order.picking_ids:
+            if (picking.state not in ('done', 'cancel') and
+                    picking.location_dest_id.usage == 'customer'):
+                key = (picking.picking_type_id, picking.location_id,
```
JordiBForgeFlow: Looking towards extensibility, please build the key in a separate method. We work, for instance, on another model that various pickings for a sales order, based on the requested date maintained at sales order line level. In this case, when we merge orders we'd want to respect the scheduled date.

sale_order_merge/models/sale_order.py
```diff
@@ -0,0 +1,74 @@
...
+                '|', ('order_policy', '=', sale.order_policy),
+                ('state', 'in', ('draft', 'sent'))]
+
+        domain = [
+            ('id', '!=', arg),
+            ('partner_id', '=', sale.partner_id.id),
+            ('partner_shipping_id', '=', sale.partner_shipping_id.id),
+            ('warehouse_id', '=', sale.warehouse_id.id),
+            ('state', 'in', self.MERGE_STATES),
+        ] + policy_clause
```
JordiBForgeFlow: Looking towards extensibility, please build the domain in a separate method. For example, in sale_operating_unit (https://github.com/OCA/operating-unit/pull/22) you would not want to merge sales orders that belong to separate operating units. Also, why is the company_id not included?

sale_order_merge/models/sale_order_merge.py
```diff
@@ -0,0 +1,190 @@
...
+                       picking.location_dest_id, picking.partner_id)
+                if key not in pick_map:
+                    pick_map[key] = self.env['stock.picking']
+                pick_map[key] += picking
+        for pickings in pick_map.values():
+            if len(pickings) > 1:
+                target = pickings[0]
+                pickings -= target
+                pickings.move_lines.write({'picking_id': target.id})
+                pickings.unlink()
```
JordiBForgeFlow, line 108: In the context of real time inventory valuation this will cause that the journal entries created from the partial transfers of pickings will refer to a deleted picking (https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock_account.py#L278)
Perhaps it would be a good idea to add here a hook (at least) that would allow to update these journal entries.

JordiBForgeFlow:
I create a SO 1, I confirm and creates a DO1.
I create a SO2 and while in draft, I merge SO2 into SO1.
A DO2 is created, but not merged with DO1.

image

JordiBForgeFlow (changes requested):

StefanRijnhart:
@jbeficent cannot seem to reply directly to https://github.com/OCA/sale-workflow/pull/341#issuecomment-250369217, but I cannot reproduce this. What are the order settings of SO1 and SO2?

StefanRijnhart:
@jbeficent found them on the runbot

StefanRijnhart:
@jbeficent You got multiple DO's for the merged order because the runbot instance has the module 'sale procurement group per line' installed. I uninstalled it and it works now.

sale_order_merge/models/sale_order_merge.py
StefanRijnhart: Only draft pickings are merged and deleted, so this is not an issue right?

sale_order_merge/models/sale_order_merge.py
StefanRijnhart: Done

sale_order_merge/models/sale_order.py
StefanRijnhart: Done

sale_order_merge/models/sale_order_merge.py
```diff
@@ -0,0 +1,198 @@
...
+                if key not in pick_map:
+                    pick_map[key] = self.env['stock.picking']
+                pick_map[key] += picking
+        for pickings in pick_map.values():
+            if len(pickings) > 1:
+                target = pickings[0]
+                pickings -= target
+                pickings.mapped('move_lines').write({'picking_id': target.id})
+                pickings.unlink()
+        return True
```
JordiBForgeFlow, line 110: The final picking should have as Source Document the final sales order, not the original.

image

sale_order_merge/README.rst
```diff
@@ -0,0 +1,70 @@
...
+=====
+
+To use this module, you need to go to the main sale order that you want to
+merge with other sale orders. If there are any candidates, you will see a
+*Merge* button in the header of the sale order form. If you click on this
+button, a window will pop-up that has all mergeable sale orders preselected.
+Remove the orders that you do not want to merge, and click on the *Merge*
+button in the footer of the pop-up window. The main window will then refresh
+to the updated main sale order.
+
```
JordiBForgeFlow, line 25: I suggest to add the following:

The mergeability criteria is defined as follows:
- Same customer, shipping address, warehouse and company.
- Orders in status 'Draft Quotation', 'Quotation Sent', 'Waiting Shedule', 'Sale Order', 'Sale to Invoice'.
- Once the order has already been confirmed, only draft orders or confirmed orders with the same invoice policy can be merged.

The criteria can easily be extended using method `_get_merge_domain` of model `sale.order`

StefanRijnhart: Done. Thanks for the review BTW.

sale_order_merge/models/sale_order_merge.py
StefanRijnhart: Done

StefanRijnhart:
Hi @jbeficient, would you like to revise your review based on my changes?

JordiBForgeFlow (approved):

JordiBForgeFlow:
@StefanRijnhart This has all the required approvals. Can I merge now?

StefanRijnhart:
As it has the two require approvals, yes!

pedrobaeza:
Remember to squash on merge as this case there is no significant different commits

StefanRijnhart:
Thanks ;-)

PR #342, opened 2016-09-29
[9.0][sale_automatic_workflow]Add possibility to invoice service automatically on delivery

florian-dacosta:
The case is when I want to automatically invoice my sale orders at delivery.
Product service must be marked as shipped manually, but when we use sale automatic workflow, we do not want to make manual actions.

So I propose to add an option to automatically fill delivered quantity on sale order line if the product is a service. It will be automatically filled during the first invoice creation.
One comon use case is the shipping fee, we want to invoice it together with the other products.

@guewen

### DIFF
--- a/sale_automatic_workflow/models/sale.py
+++ b/sale_automatic_workflow/models/sale.py
@@ -41,3 +41,14 @@ class SaleOrder(models.Model):
             warning = {'title': _('Workflow Warning'),
                        'message': workflow.warning}
             return {'warning': warning}
+
+    @api.multi
+    def action_invoice_create(self, grouped=False, final=False):
+        for order in self:
+            if not order.workflow_process_id.invoice_service_delivery:
+                continue
+            for line in order.order_line:
+                if line.qty_delivered_updateable and not line.qty_delivered:
+                    line.write({'qty_delivered': line.product_uom_qty})
+        return super(SaleOrder, self).action_invoice_create(grouped=grouped,
+                                                            final=final)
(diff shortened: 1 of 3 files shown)


atchuthan:
Runbot tested :+1:

StefanRijnhart (approved):

pedrobaeza (approved):

sale_automatic_workflow/models/sale.py
```diff
@@ -41,3 +41,14 @@ def onchange_workflow_process_id(self):
...
                        'message': workflow.warning}
             return {'warning': warning}
+
+    @api.multi
+    def action_invoice_create(self, grouped=False, final=False):
+        for order in self:
+            if not order.workflow_process_id.invoice_service_delivery:
+                continue
+            for line in order.order_line:
+                if line.qty_delivered_updateable and not line.qty_delivered:
```
guewen, line 51: shouldn't be a check on the type of product being `service`?

StefanRijnhart: This is implicit in the definition of qty_delivered_updateable: it is only true for service products which are invoiced manually (so not through tasks).

florian-dacosta: Indeed the computation of the field check it already : 
https://github.com/OCA/OCB/blob/9.0/addons/sale_stock/sale_stock.py#L123

So I think there is no need to double check it here, but I am not especially against if you think it may be usefull.

guewen: I should have checked what was `qty_delivered_updateable` (name is not self-explanatory). Sorry for the trouble.

StefanRijnhart: No, it's a valid question. I had to look it up as well during the review, so I could have mentioned this observation in my review, or maybe a code comment would have been helpful. Next time!

PR #345, opened 2016-10-04
8.0 porting product_customer_code_sale

espo-tony:


product_customer_code_sale/__openerp__.py
```diff
@@ -1,44 +1,21 @@
...
+    "version": "8.0.1.0.0",
     "author": "Agile Business Group,Odoo Community Association (OCA)",
     "website": "http://www.agilebg.com",
     "category": "Sales Management",
+    'license': 'AGPL-3',
     "depends": [
-        'base',
         'product',
         'sale',
         'product_customer_code'
```
gurneyalex, line 14: this one is in product-attributes which I don't see referenced in the oca-dependencies.txt file. Can you fix this?

gurneyalex (changes requested):

product_customer_code_sale/__openerp__.py
espo-tony: @gurneyalex 
Thanks for your review.

product_customer_code_sale/static/description/index.html
```diff
@@ -0,0 +1,124 @@
+<section class="oe_container">
```
StefanRijnhart, line 1: This seems like an empty placeholder. Maybe it was not included intentionally?

StefanRijnhart (changes requested):
Please move the view file to a subdirectory called 'views'
Can you please add a simple test that works on one of the unconfirmed sale orders from the core demo data to verify that the customer code is propagated correctly?

StefanRijnhart:
The migration of product_customer_code is pending as https://github.com/OCA/product-attribute/pull/122. This PR will not be green before then.

susport (commented):
Order by name the import according to the OCA guide

product_customer_code_sale/models/sale.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Nicola Malcontenti - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openerp import models, api, fields
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.depends('product_id', 'order_id.partner_id')
```
susport, line 10: Suggestion.
Why not use @ api.one and it would look like this?
@api.one
@api.depends('product_id', 'order_id.partner_id')
def _compute_product_customer_code(self):
product_customer_code_obj = self.env['product.customer.code']
partner = self.order_id.partner_id
product = self.product_id
code = ''
if product and partner:
code_ids = product_customer_code_obj.search([
('product_id', '=', product.id),
('partner_id', '=', partner.id),
], limit=1)
if code_ids:
code = code_ids[0].product_code or ''
self.product_customer_code = code

StefanRijnhart: @api.one is deprecated.

eLBati:
See https://github.com/OCA/product-attribute/pull/236#issuecomment-304812958

tafaRU:
@espo-tony, I'd close this as https://github.com/OCA/product-attribute/pull/236#issuecomment-304812958, do you agree?

espo-tony:
@tafaRU Yes, I agree. Thanks to you all!

PR #348, opened 2016-10-13
[ADD] sale_order_calendar_event

hbrunn:
# 
# Calendar events in sale orders

This module was written to allow users to attach calendar events to sale orders.
Usually, this would be some appointments that play a role for working on the
sale order.
# Usage

# . go to some sale order
# . create an event there
# . it will be linked to this sale order automatically and visible with the sale

   order filter in the agenda


andreschenkels:
It does exactly what you describe :+1:

pedrobaeza:
Interesting design decision. Can you expand a bit more the application of this module in a user case?

hbrunn:
I was a bit ashamed to post such a simple module, but it's quite useful for my customer.

They do two things: On the one hand, they have several dates within one `project` (which is encoded as a sale order) that are crucial points, but not from the point of view of the erp system - think of several go/no go moments, or some external audit that takes place, this kind of stuff.
Then, people who are busy with such a project use appointments as reminders to themselves for whatever they don't want to forget.

While all of this is possible with the plain calendar, and more with the project module, it's convenient for them not to have to dive into project management at all, and have a link from appointments to sale orders in order to filter and search.

pedrobaeza:
OK, thanks for the explanation. It would be good to have a similar text in the README

pedrobaeza (approved):

rafaelbn (approved):
This looks like a "shortcut" from SO to create Meetings.
I've tested in runbot functionally

PR #350, opened 2016-10-18
[ADD] migrated sale_sourced_by_line & sale_procurement_group_by_line to v9

SerpentCS:


JordiBForgeFlow:
@aheficent can you review?

JordiBForgeFlow:
cc @aheficent can you review please?

sale_procurement_group_by_line/model/sale.py
```diff
@@ -137,45 +55,73 @@ def _get_shipped(self):
     ###
     # OVERRIDE to find sale.order.line's picking
     ###
-    def _get_picking_ids(self, cr, uid, ids, name, args, context=None):
-        res = {}
-        for sale in self.browse(cr, uid, ids, context=context):
+
+    @api.one
```
AaronHForgeFlow: IMHO this should be api.multi as we are iterating over the records

AaronHForgeFlow: I would do that also in the method "_get_shipped", in this case with a self.ensure_one()

sale_procurement_group_by_line/model/sale.py
```diff
@@ -137,45 +55,73 @@ def _get_shipped(self):
...
+    shipped = fields.Boolean(compute='_get_shipped', string='Delivered',
+                             store=True)
 
 
-class SaleOrderLine(orm.Model):
+class SaleOrderLine(models.Model):
     _inherit = 'sale.order.line'
 
     @api.multi
+    def _action_procurement_create(self):
```
AaronHForgeFlow: I agree with you here, doing it in the line is cleaner

sale_sourced_by_line/model/stock.py
```diff
@@ -1,87 +1,28 @@
...
+                    self.sale_id = lines[0].order_id
 
-            if line_ids:
-                created_lines = sale_line_obj.invoice_line_create(
-                    cr, uid, line_ids, context=context)
-                invoice_line_obj.write(
-                    cr, uid, created_lines, {'invoice_id': invoice_id},
-                    context=context)
-        return invoice_id
+    sale_id = fields.Many2one('sale.order', compute='_compute_get_sale_id',
```
AaronHForgeFlow: The method is correct but I think it would be nice to have another relation to the SO line as in the same SO can coexist several lines related to different procurement groups

AaronHForgeFlow (changes requested):
Succesfully tested on runbot

sale_sourced_by_line/model/stock.py
JordiBForgeFlow: This comment is not applicable, as the sale_id field is removed from the picking in v9

AaronHForgeFlow (approved):

LoisRForgeFlow (approved):
tested on runbot :+1:

JordiBForgeFlow:
@SerpentCS @darshan-serpent Can you please reorganize the PR (or create a new on), so as to respect the commit history of prior versions?

PR #352, opened 2016-10-27
[ADD] New module Sale Order Invoice Link

mileo:


pedrobaeza:
This maybe fits better in OCA/account-invoicing

mileo:
We were also with this doubt.... 

But stock invoice link is here https://github.com/OCA/stock-logistics-workflow/tree/10.0/stock_picking_invoice_link

pedrobaeza:
Yeah, I know, but I consider that also a past mistake. OCA/sale-workflow is enough populated I think, so we can move this one to OCA/account-invoicing, and see if we move the other also in next versions.

mileo:
OK

PR #360, opened 2016-11-22
[sale_exceptions] Count stock from selected Warehouse

TheCloneMaster:
Rule should only count available stock from selected warehouse from sale order.
Without warehouse in context, it currently sums all stock available in all warehouses.


pedrobaeza:
10.0 version is still not installable, so no need to make this patch. Please do it in version 9, that is the other installable version.

PR #365, opened 2016-11-29
[9.0][FIX] sale_order_type: Manager security overwrited with same Id …

carlosdauden:
…in Salesman security

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as this is critical.

PR #373, opened 2016-12-13
update README.rst

DarioLodeiros:
add description to consume materials


pedrobaeza:
The clarification shouldn't be here, but on project_task_materials_stock module.

DarioLodeiros:
Ok, in this module already exists the clarification ;)

PR #374, opened 2016-12-14
Option to set sale order done consider shipping

guewen:
The option to automatically set the sales order as 'done' currently
filters only the sales orders in state 'sale' and invoiced.

I propose to change the default filter to 'all is invoiced AND all is
delivered' which emulates the behaviour of the previous versions of
Odoo.

This is a backward compatible change: as the filter has `noupdate="1"`,
existing installations will keep their domains and new installations
will get the new one.

### DIFF
--- a/sale_automatic_workflow/models/sale.py
+++ b/sale_automatic_workflow/models/sale.py
@@ -5,6 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from openerp import models, fields, api, _
+from openerp.tools import float_compare
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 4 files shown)


sale_automatic_workflow/models/sale.py
```diff
@@ -15,6 +16,24 @@ class SaleOrder(models.Model):
...
+    )
+
+    @api.depends('order_line.qty_delivered', 'order_line.product_uom_qty')
+    def _compute_all_qty_delivered(self):
+        precision = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure'
+        )
+        for order in self:
+            order.all_qty_delivered = all(
+                l.product_id.type != 'product' or
```
pedrobaeza: This should be also for 'consu'

guewen: right! fixed

pedrobaeza (approved):

atchuthan (approved):
:+1:

PR #375, opened 2016-12-14
[FIX][9.0] Fix sale exception menu child of sales settings that make …

jjscarafia:
Now sale exception menu is child of sales settings and makes sale settings menu un clickeable on enterprise interface. We change it to be a child of sales config menu

Now:
seleccion_006

Whit this PR:
seleccion_007

### DIFF
--- a/sale_exception/views/sale_view.xml
+++ b/sale_exception/views/sale_view.xml
@@ -50,7 +50,7 @@
                   <field name="context">{'active_test': False}</field>
               </record>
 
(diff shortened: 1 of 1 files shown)


jjscarafia:
The warning on runbot is because of https://github.com/OCA/sale-workflow/tree/9.0/sale_rental

manuelfcordoba:
:+1: tested on runbot

jjscarafia:
Already done here https://github.com/OCA/sale-workflow/commit/1f7ed31de9de6f0820c01c608caa4461490a435b

PR #376, opened 2016-12-16
[10.0] Port sale_commercial_partner (and move from edi)

alexis-via:
This PR is a port from v8 to v10 of the OCA module sale_commercial_partner, which is currently located by mistake in the OCA/edi project. It will be removed from the OCA/edi project during the port to v10.


AaronHForgeFlow (approved):
:+1:

rafaelbn (approved):

chienandalu (approved):

PR #379, opened 2016-12-23
[MIG][10.0]sale_product_set

leemannd:
Migration of sale_product_set to V10

### DIFF
--- a/sale_product_set/wizard/product_set_add.py
+++ b/sale_product_set/wizard/product_set_add.py
@@ -1,17 +1,17 @@
 # -*- coding: utf-8 -*-
-from openerp import models, fields, api, _
-import openerp.addons.decimal_precision as dp
+from odoo import models, fields, api
+import odoo.addons.decimal_precision as dp
 
 
 class ProductSetAd(models.TransientModel):
     _name = 'product.set.add'
     _rec_name = 'product_set_id'
-    _descritpion = "Wizard model to add product set into a quotation"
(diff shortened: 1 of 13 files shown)


leemannd:
@Yajo @pedrobaeza If you are interested, this one is ready.

pedrobaeza:
We are preferring this other approach, although not exactly the same purpose: https://github.com/OCA/product-attribute/pull/180

leemannd:
@pedrobaeza I like the approach of the bundle. AFAIK our goal is just to quickly add set of products to a quotation/order. So it's fully covered by this one.

pedrobaeza:
OK, no problem, they don't have the exact scope indeed.

simahawk (approved):

gurneyalex (approved):

sale_product_set/wizard/product_set_add.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from openerp import models, fields, api, _
+from odoo import models, fields, api, _
 import openerp.addons.decimal_precision as dp
```
yvaucher, line 5: Can you also fix the two field's strings here by removing _()

Plus the typo on s/_descritpion/_description

leemannd:
@yvaucher Thanks for the comment. Updated.

yvaucher (approved):

PR #386, opened 2017-01-02
[9.0][NEW] product_list_price_tax: Write price in product template with taxes include

sergio-teruel:
With this module you can write list price with taxes include.
You can write list price and price with tax is calculated or you can write list price with taxes and list price is also calculated.

cc@Tecnativa


pedrobaeza:
Check runbot and Travis status

pedrobaeza:
This should be in product-attribute repo.

PR #388, opened 2017-01-03
[FIX] sale_order_type: empty currency

esthermm:
When type_id is changed the pricelist_id changed but the onchange of the pricelist is not executed. So it does not fill currency_id field.

### DIFF
--- a/sale_order_type/models/sale_order.py
+++ b/sale_order_type/models/sale_order.py
@@ -37,7 +37,7 @@ class SaleOrder(models.Model):
             self.pricelist_id = self.type_id.pricelist_id.id
             res = self.onchange_pricelist_id(
                 self.pricelist_id.id, self.order_line.ids)
-            self.currency_id = res['value']['currency_id']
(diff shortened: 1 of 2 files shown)


sale_order_type/models/sale_order.py
```diff
@@ -35,6 +35,9 @@ def onchange_type_id(self):
             self.payment_term = self.type_id.payment_term_id.id
         if self.type_id.pricelist_id:
             self.pricelist_id = self.type_id.pricelist_id.id
+            res = self.onchange_pricelist_id(
+                self.pricelist_id.id, self.order_line.ids)
+            self.currency_id = res['value']['currency_id']
```
pedrobaeza: You should make `self.update(res.get('value', {}))` for updating all possible values returned from onchange.

pedrobaeza (approved):

oihane (approved):
:+1: LGTM

mikelarre:
:+1:

PR #391, opened 2017-01-09
[10.0][MIG] sale_automatic_workflow

lasley:
This PR migrates sale_automatic_workflow to v10


lasley:
Ready for review!

florian-dacosta:
You should include last changes from v9 : 
https://github.com/OCA/sale-workflow/pull/374

Otherwise, it seems good to me

lasley:
Thanks for the review @florian-dacosta, cherry-picked ac188a43020ac87d3c5ec1b48735af483658f93d

sale_automatic_workflow/models/sale.py
```diff
@@ -41,14 +60,3 @@ def _onchange_workflow_process_id(self):
             warning = {'title': _('Workflow Warning'),
                        'message': workflow.warning}
             return {'warning': warning}
-
-    @api.multi
-    def action_invoice_create(self, grouped=False, final=False):
-        for order in self:
-            if not order.workflow_process_id.invoice_service_delivery:
```
florian-dacosta: Seems that this part is gone.

florian-dacosta:
The tests are failing and the part about the option "invoice_service_delivery" is gone.

sale_automatic_workflow/tests/test_flow.py
```diff
@@ -71,7 +54,7 @@ def progress(self):
     def test_full_automatic(self):
         workflow = self._create_full_automatic()
         sale = self._create_sale_order(workflow)
-        sale.onchange_workflow_process_id()
```
angelmoya, line 74: This lines are failing on tests, function is "onchange_workflow_process_id" not "_onchange_workflow_process_id"

mourad-ehm:
Hi @lasley, travis is red, could check it ?

lasley:
Thanks for the ping @mourad-ehm, I forgot about this one. Also, thanks for solving the issue for me @angelmoya! We should be 🍏  now!

florian-dacosta:
@lasley The part about `invoice_service_delivery` is still gone.
I think it has been removed by mistake by the previous commit : 
https://github.com/OCA/sale-workflow/pull/391/files/16e9987e680a6a44acdb14bd7554414dfe261056..c8df2c34b0bb378f8c99ce3d7e542d20b8bc768a#diff-2deedcf3114952d8bdda41f9a1f96f87

lasley:
@florian-dacosta - good catch, looks like it happened in the cherry pick of ac188a43020ac87d3c5ec1b48735af483658f93d to c8df2c34b0bb378f8c99ce3d7e542d20b8bc768a. I'll recreate that commit with a little more magic & hopefully clear that up.

lasley:
Alright I was much more careful this time around & bec653a82536411e06401b860765a92f5d939917 looks solid IMO

mourad-ehm:
:+1:

sale_automatic_workflow/views/sale_workflow_process_view.xml
```diff
@@ -1,116 +1,126 @@
...
+                <group string="Order Configuration">
+                    <field name="picking_policy"/>
+                    <field name="team_id" />
+                    <!-- <field name="team_id" groups="base.group_multi_salesteams"/> -->
+                </group>
+                <group string="Workflow Options" col="4">
+                    <field name="validate_order" />
+                    <label for="order_filter_id" attrs="{'required':[('validate_order','=',True)], 'invisible':[('validate_order','!=',True)]}"/>
+                    <div attrs="{'required':[('validate_order','=',True)], 'invisible':[('validate_order','!=',True)]}">
+                        <field name="order_filter_domain" />
```
rousseldenis: @lasley I know that was changed in v9, but I'm questioning about this field (and other char domain ones).
I understand that it is used to check the domain selected, but it's not readonly (even it's not used anywhere in code)
I like the way it was implemented before as char_domain widget.

Could you explain why (if you know) ?

Thanks

rousseldenis:
@atchuthan @lasley 

As I said above, I'm questioning about char fields that represent domains but are actually unused.
@atchuthan said that these fields should be used (as I understand) to customize filters to user's needs.

Could we make the decision here how this should behave and fix it ?
I like the feature to customize filters. So, in order to use it, we should change the code to take char field value instead of the Many2one.

I like also the char_domain widget, so why not using it.

What do you think about  ?

rousseldenis:
Update : After advanced testing:

- The filter can be updated through the Many2one field, so we can keep the char fields with char_domain widget.

The last question : why not a related field ?

lasley:
@rousseldenis - related field where?

rousseldenis:
@lasley The fields like 'order_filter_domain' which is computed on onchange order_filter_id.

Actually, every onchange-computed fields could be related to 'xxx_filter_id.domain' as they are just used to display the domain to user

lasley:
@rousseldenis - Ah yup I see what you're talking about now, good catch. I've updated those to related fields. 

I also implemented the char_domain widget, which was also a good idea.

rousseldenis (approved):
:+1: Seems good to me. @lasley @atchuthan Do you think it's good to backport to v9 the last changes (related fields + widget char_domain)?

atchuthan (approved):
Runbot :+1:

atchuthan:
@rousseldenis sure we can backport the above functionality to v9

Cedric-Pigeon (approved):

sale_automatic_workflow/models/sale_workflow_process.py
```diff
@@ -35,36 +35,41 @@ def _default_filter(self, xmlid):
         default='direct',
     )
     validate_order = fields.Boolean(string='Validate Order')
-    order_filter_domain = fields.Char(
+    order_filter_domain = fields.Text(
         string='Order Filter Domain',
-        default="[('state', '=', 'draft')]"
+        related='order_filter_id.domain',
```
lmignon, line 40: @lasley `readonly=True` for all your related fields since these are not editable in your model

lmignon (changes requested):

sale_automatic_workflow/models/sale_workflow_process.py
lasley: I'm not following - these are definitely editable as it is, and adding readonly would restrict usage IMO

Currently:
image

With Readonly:
image

rousseldenis: @lasley I think @lmignon would say that generally we set related fields as readonly.
This to avoid modification through this related field.

Moreover, these related fields are 'informative' field => if we want to modify it, we do it through the Many2one.

The behaviour you show is the 'char_domain' one (ok it's weird to have sale.order readonly as we set it on ir.filter.domain...).

So, no restriction in usage with readonly (as we modify through the other field - Many2one)

lasley: If the editing works through a related field, why would we arbitrarily require another click from the user to edit through the M2O instead?

rousseldenis: Indeed, the functionnality introduced by char_domain (show the sale orders), hides the field value.

So, we can decide what is more convenient for users (have the widget or the domain direct edition)

lasley: Looking at it harder, there is some inconsistency when there isn't something selected in the M2O. In light of that, I made the domains invisible when there isn't an associated filter record selected (183f557).

Regarding the readonly - unless we have a technical reason, I do not see the reason to require a more complex workflow just because we feel that a field is reference only. Interestingly enough, editing them through the related fields provided here is actually more convenient than editing them through the M2O because the `ir.filter` does not implement the `char_domain` widget.

lmignon: @lasley I understand you POV. I don't know if it's already the case but I've observed that event if the content of a related field is not changed on edit, odoo call the method write on the related record. In some cases this kink of behavior can lead to security error if the user has not the right to update the target model. That's why I've the habit to declare my related fields readonly by default... But I can approve this PR as it is. :smirk:

lmignon (approved):

lmignon:
@lasley I'm a little bit skeptical with the commit method into the automatic_workflow_job...
https://github.com/OCA/sale-workflow/pull/391/files#diff-e446146691b5650c987b79a9795b52d9R23
IMO it would be safer to enclose the yield into a savepoint....
```python
@contextmanager
def commit(cr):
    """
    Commit the cursor after the ``yield``, or rollback it if an
    exception occurs.
 is Pending
    Warning: using this method, the exceptions are logged then discarded.
    """
    try:
        with cr.savepoint():
            yield
            cr.commit()
    except Exception:
        _logger.exception('Error during an automatic workflow action.')
```

sale_automatic_workflow/models/sale_workflow_process.py
lasley: @lmignon - Oh interesting, that makes complete sense thank you for elaborating. 

I think in the context here we're good, but I am going to do some more investigation on the implications of this.

lasley:
@lmignon Good point with the savepoint vs the commit, change made.

I noticed in your sample that there was still a `cr.commit`. Is that required? It was my understanding that a commit isn't required due to the use of the Postgres savepoints inside of a [context manager handling the creation, rollback, and release appropriately](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/sql_db.py#L396).

lmignon:
> I noticed in your sample that there was still a cr.commit. Is that required? It was my understanding that a commit isn't required due to the use of the Postgres savepoints inside of a context manager handling the creation, rollback, and release appropriately.

@lasley good catch... cr.commit is no more required.

yvaucher:
@lasley @lmignon removing the commit would also makes the tests replayable.

yvaucher:
Currently tests are failing on second run.

```
 test_onchange (odoo.addons.sale_automatic_workflow.tests.test_flow.TestAutomaticWorkflow)
 ======================================================================
 ERROR: test_journal_on_invoice (odoo.addons.sale_automatic_workflow.tests.test_flow.TestAutomaticWorkflow)
 `   File "/opt/odoo/external-src/sale-workflow/sale_automatic_workflow/tests/test_flow.py", line 128, in test_journal_on_invoice
 `     "type": "sale"})
 `   File "/opt/odoo/src/addons/account_check_printing/models/account_journal.py", line 42, in create
 `     rec = super(AccountJournal, self).create(vals)
 `   File "/opt/odoo/src/addons/account/models/account.py", line 481, in create
 `     journal = super(AccountJournal, self).create(vals)
 `   File "/opt/odoo/src/odoo/models.py", line 3830, in create
 `     record = self.browse(self._create(old_vals))
 `   File "/opt/odoo/src/odoo/models.py", line 3925, in _create
 `     cr.execute(query, tuple(u[2] for u in updates if len(u) > 2))
 `   File "/opt/odoo/src/odoo/sql_db.py", line 141, in wrapper
 `     return f(self, *args, **kwargs)
 `   File "/opt/odoo/src/odoo/sql_db.py", line 218, in execute
 `     res = self._obj.execute(query, params)
 ` IntegrityError: duplicate key value violates unique constraint "account_journal_code_company_uniq"
 ` DETAIL:  Key (code, name, company_id)=(TTSA, TTSA, 1) already exists.
 ` 
 Ran 5 tests in 9.314s
```

I was about to do a PR to add this check:

``` python
        if not tools.config['test_enable']:
            cr.commit()
```

lasley:
Excellent note @yvaucher - looks like that is also fixed up with the savepoint 🚀

yvaucher (approved):

yvaucher:
@lasley yup just tested it launching the tests twice locally with savepoints. It works

Let's merge this.

lasley:
Squashed my last few commits

lasley:
Damn not sure whether I got it before merge though 😦

PR #393, opened 2017-01-20
[IMP] add possibility to manually finish sale orders

legalsylvain:
improve module sale_order_mass_action.

add a checkbox 'finish'. if checked, all pending orders will pass in a 'done' state. 

That allows to fix some sale orders in V7.0 that do not pass into 'done' state, even if all is done. (delivery, invoicing, etc...)

### DIFF
--- a/sale_order_mass_action/i18n/fr.po
+++ b/sale_order_mass_action/i18n/fr.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: OpenERP Server 7.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-05-06 12:08+0000\n"
-"PO-Revision-Date: 2015-05-06 12:08+0000\n"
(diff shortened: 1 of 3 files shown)


legalsylvain:
CC : @arthru

arthru:
This is a bit disturbing to have an empty wizard when `finishable_order_qty == 0`

But for the rest, it works fine...

Take care of that CI fails although...

AaronHForgeFlow (commented):
It's quite dangerous to let the user finish many SO at the same time. Maybe only technical user should have access to this. Don't you think so?

LoisRForgeFlow (commented):
I agree with @arthru is quite weird to have an empty wizard with nothing displayed (at least a information message would be nice). Also, I reckon that to show the sale orders that are going to be finished or confirmed instead of just the number would avoid some potential mistakes.

Travis is failing due to other module, in that sense is fine for me.

legalsylvain:
closing due to inactivity.

PR #395, opened 2017-01-22
[9.0] [FIX] sale_order_variant_mgmt: Critical bug related fields on sale.order.line trigger a two product writes.

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):
This ORM behavior is very dangerous and annoying. Fast-tracking it. Runbot has a warning related to sale_rental.

PR #399, opened 2017-01-27
[8.0][FIX] sale order lines of old revision must be in state cancel -…

sergiocorato:
… if not they result in statistics as sale


eLBati:
Shouldn't this be done by the standard 'cancel' button, in case?

sergiocorato:
The sale order is put in status 'open' by python function to copy the lines to the new sale order revision, then the order lines are not more in status 'cancel' and they remain 'draft' (or 'open').

eLBati:
Sorry @sergiocorato could you describe the steps to produce the error, the observed behaviour  and the desired behaviour?

PR #403, opened 2017-02-07
[9.0] Add sale_payment

florian-dacosta:
This module aims to replace modules sale_payment_method and sale_quick_payment from v8.

It allows you to to pay add payment on draft sale order, it allows also to link payment to sale order.
It can be a usefull information for user to see if sale order has associated payment and could also be used for  with sale_exceptions for example, to create validation rule if order is paid, etc...

### DIFF
--- a/sale_payment/README.rst
+++ b/sale_payment/README.rst
@@ -1,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============
+Sale Payment
+============
+
+This module allows to register payments entries on sales orders.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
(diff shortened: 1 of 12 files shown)


mourad-ehm:
When I "Register a payment" from sale order then I create an invoice, This one is not passed to state "Paid" automatically. Why this one remain in "Open" state? Is a normal (wished) behaviour ?

mourad-ehm:
runbot is red, but there no error message. What's happen ?
Some one has a response ?

florian-dacosta:
@mourad-ehm 
I don't think the feature was existing in previous versions and I did not need it for my customer as we use the mass_reconcile module for all reconciliation.
Of course, I guess it would be a nice feature to have and could be added to the module in the future, if anyone need this.
Regarding the runbot, I do not know...

LoisRForgeFlow:
Seem similar problem to what happens here: https://github.com/OCA/sale-workflow/pull/407

In your case is <code>must_have_dates</code> but also <code>type</code> and <code>uom_id</code>

rousseldenis:
@florian-dacosta Is this still relevant? As no activity, it will be closed soon.

florian-dacosta:
@rousseldenis 
Well, actually, I think this module is still relevant, but v9 is old already and this PR probably won't ever move forward, so, agree to close it.

PR #405, opened 2017-02-10
10.0 del sale order back2draft

mourad-ehm:
Delete module sale order back2draft because feature is in odoo v 10.0.


pedrobaeza:
Well, there's no need to import OCA commits then. I'll squash on merge.

pedrobaeza (approved):
Fast-tracking

PR #407, opened 2017-02-10
[8.0][FIX] sale_payment_method_automatic_workflow: fix write off when reconcile payment in foreign currency

Cedric-Pigeon:


elicoidal:
strange for the Runbot: no obvious reasons

Cedric-Pigeon:
indeed I don't understand

LoisRForgeFlow:
The problem seems to be in other modules. <a href="https://github.com/acsone/sale-workflow/blob/8.0-automatic_workflow-cpi/sale_rental/rental.py#L47">This method</a> has as constraint <a href="https://github.com/acsone/sale-workflow/blob/8.0-automatic_workflow-cpi/sale_start_end_dates/sale.py#L83">a non writable field</a>.

to remove it from the constrain in a separate fix PR may solve this and prevent it to happen again. What do you think?

rousseldenis:
Closing this as no activity. @Cedric-Pigeon  Please feel free to reopen it if you still need it. Take care of your PR's

PR #412, opened 2017-02-16
10.0 mig sale order lot selection

mourad-ehm:

PR #415, opened 2017-02-20
[FIX] change product id with qty # of 0

mourad-ehm:
When I migrated the module sale_order_lot_selection to v 10.0, I did change some product ids that were removed in 10.0. This products have broken the test. 
The reason is that odoo don't take into account the lot quantity available but the product quantity available.
So, as the product 'product.product_product_57' has not available quantity, the test was OK. 
With product 'product.product_product_6' test fails. Even the lot is delivered odoo reserve move for the same lot because product is available.
This behaviour come from this method :
https://github.com/OCA/OCB/blob/10.0/addons/stock/models/stock_quant.py#L437 
first odoo check reservation with a lot, if the lot is not available, it check with lot_id = False (by product).

The question is if we should keep the method _check_move_state or not (at least in v10) ? 
Or 
Drop this part of test 
        # I'll try to confirm it to check lot reservation:
        # lot10 was delivered by order1
        with self.assertRaises(Warning):
            self.order3.action_confirm()
I create this PR to reproduce the same error of the PR #412

### DIFF
--- a/sale_order_lot_selection/tests/test_sale_order_lot_selection.py
+++ b/sale_order_lot_selection/tests/test_sale_order_lot_selection.py
@@ -36,8 +36,8 @@ class TestSaleOrderLotSelection(test_common.SingleTransactionCase):
 
         """
         super(TestSaleOrderLotSelection, self).setUp()
-        self.product_57 = self.env.ref('product.product_product_57')
-        self.product_46 = self.env.ref('product.product_product_46')
+        self.product_57 = self.env.ref('product.product_product_6')
+        self.product_46 = self.env.ref('product.product_product_17')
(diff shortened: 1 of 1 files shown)


florian-dacosta:
I don't understand why we shoud change the tests in version 9, because it is broken in version 10?
Can't you just change theses products in the test of the version 10 module?

Actually, if a product is configured to handle lot, I guess it should not have any quants without lots, so the case should not even exist.

About the '_check_move_state' method, I do not think it is really generic, and I think it should be kept in a small sub module, not in the main module sale_order_lot_selection.
In my use case anyway, I have to remove this behavior and I really think it should not be in this module.

mourad-ehm:
This PR is no longer necessary I'll close it.

PR #417, opened 2017-02-23
[MIG] sale_sourced_by_line: Migrated to 9.0

AaronHForgeFlow:
Sale Sourced by Line
====================

This module adds the possibility to source a line of sale order from a specific
warehouse instead of using the warehouse of the sale order.

This will create one procurement group per warehouse set in sale
order lines.

It will only supports routes such as MTO and Drop Shipping.

Supersedes https://github.com/OCA/sale-workflow/pull/350

### DIFF
--- a/sale_sourced_by_line/model/sale.py
+++ b/sale_sourced_by_line/model/sale.py
@@ -1,75 +1,54 @@
 # -*- coding: utf-8 -*-
-#
-#
-#    Author: Guewen Baconnier, Yannick Vaucher
-#    Copyright 2013-2015 Camptocamp SA
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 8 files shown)


sale_sourced_by_line/model/sale.py
```diff
@@ -1,66 +1,33 @@
...
         return vals
 
-    SO_STATES = {
-        'cancel': [('readonly', True)],
-        'progress': [('readonly', True)],
-        'manual': [('readonly', True)],
-        'shipping_except': [('readonly', True)],
-        'invoice_except': [('readonly', True)],
-        'done': [('readonly', True)],
-    }
```
yvaucher, line 53: Any comment or reason about removing this?

Was added in https://github.com/OCA/sale-workflow/commit/40c864e498ad2ebbb6bc4030718c61853cd58bad
which in fact doesn't give more clue why it is here.

To me, it seems reasonable to have that field only editable in draft or open states as once `_prepare_order_line_procurement` has been called there is no reason that I can think of to change this value.

In fact to be consistent with other fields I think it should be:

    warehouse_id = fields.Many2one(
        [...]
        readonly=True,
        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]}

yvaucher (approved):

sale_sourced_by_line/model/sale.py
JordiBForgeFlow: Thanks @yvaucher I changed as per your review comments.

JordiBForgeFlow (approved):

JordiBForgeFlow:
Additional approvals already provided in #350

PR #418, opened 2017-02-23
[MIG] 9.0 sale procurement group by line

AaronHForgeFlow:
Sale Procurement Group by line
==============================

This module was written to extend the functionality of procurement groups
created from a Sale order.

On itself, this module does nothing it is a requirement for modules which
needs to create procurement group per sale order line basis.

Supersedes https://github.com/OCA/sale-workflow/pull/350

PR #419, opened 2017-02-23
[FIX][9.0] sale_rental_fix

AaronHForgeFlow:
must_have_dates, type and uom_id fields are readonly. sale_rental module is using them in a constraint wrongly. There's an issue created: https://github.com/OCA/sale-workflow/issues/420

This is causing this PR to fail: https://github.com/OCA/sale-workflow/pull/418 :confused:

### DIFF
--- a/sale_rental/models/product.py
+++ b/sale_rental/models/product.py
@@ -20,7 +20,7 @@ class ProductProduct(models.Model):
         string='Related Rental Services')
 
     @api.one
-    @api.constrains('rented_product_id', 'must_have_dates', 'type', 'uom_id')
+    @api.constrains('rented_product_id')
     def _check_rental(self):
         if self.rented_product_id and self.type != 'service':
             raise ValidationError(_(


LoisRForgeFlow (approved):

sale_rental/models/product.py
```diff
@@ -20,7 +20,7 @@ class ProductProduct(models.Model):
         string='Related Rental Services')
 
     @api.one
-    @api.constrains('rented_product_id', 'must_have_dates', 'type', 'uom_id')
+    @api.constrains('rented_product_id')
```
andhit-r, line 23: @aheficent runbot was failed for sure. But If you remove type constraint what will restrict user from creating/editing a stockable rent product? Rent product have to be service for sure. Beside field (1) must_have_dates, (2) type, and (3) uom_id is not readonly on product.product. These screenshot taken from runbot:

sale_order_rental

sale_order_rental_1

I don't know what cause runbot failed (i tried to look into the log but i could not find it), but removing constraint will endanger the rental process

AaronHForgeFlow: Runbot is failing because of this:
image

This normally happens when the fields are readonly. It seems that this is not the case. I think it's better to discuss in the issue (https://github.com/OCA/sale-workflow/issues/420) and close this then:

pedrobaeza:
Please don't close this. @andhit-r are not correct. The constraint is never going to be triggered for type field, as you cannot change it as it's not writeable. That's the reason of the warning. You can read its value anyway inside the constraint.

pedrobaeza:
Maybe the change should be to not put on the model as readonly.

AaronHForgeFlow:
@pedrobaeza I changed must_have_dates to be not readonly. However type and uom_id are fields of product.product and declared in other repos. Maybe should we created separated PR for any single field affected?

sale_start_end_dates/models/sale.py
```diff
@@ -54,7 +54,7 @@ class SaleOrderLine(models.Model):
         states={'draft': [('readonly', False)]})
     number_of_days = fields.Integer(string='Number of Days')
     must_have_dates = fields.Boolean(
-        related='product_id.must_have_dates', readonly=True)
```
pedrobaeza: You don't have to remove this readonly. This is correct. The problem is in product.product constraint, not here.

pedrobaeza:
Which repos? Do you mean other modules? You need to find the source of the problem and fix there (although this is not something you have provoked).

sale_start_end_dates/models/sale.py
alexis-via: I agree with Pedro, this code is correct :)

AaronHForgeFlow:
I know that this is not the proper fix, however the only solution I find is to create a related field that would not be readonly. Of course that would be too dirty from my point of view. Any ideas? @pedrobaeza @alexis-via

rafaelbn:
Hi, we need help here @alexis-via @pedrobaeza , is it possible?

pedrobaeza:
What do you want, @rafaelbn ?

AaronHForgeFlow:
This issue was closed because some PR was merged https://github.com/OCA/sale-workflow/issues/420 but the problem is still there, some PR in v9 for the sale-workflow are not passing the checks because of this

alexis-via:
must_have_dates is NOT a readonly field on product.template (it is a regular boolean field).

I think the problem is in the framework of odoo v9: in an inherit of product.product, odoo doesn't accept to have fields in @api.constrains that are attached to product.template.

alexis-via:
Reminder: here is the warning we are trying to fix:
```
2017-05-30 15:03:43,544 17408 WARNING o9_test1 openerp.models: method product.product._check_rental: @constrains parameter 'must_have_dates' is not writeable
2017-05-30 15:03:43,544 17408 WARNING o9_test1 openerp.models: method product.product._check_rental: @constrains parameter 'type' is not writeable
2017-05-30 15:03:43,544 17408 WARNING o9_test1 openerp.models: method product.product._check_rental: @constrains parameter 'uom_id' is not writeable
```

pedrobaeza:
Please workaround it with the explained method to see

AaronHForgeFlow:
I did not close on purpose

pedrobaeza:
Can be closed as it has been fixed upstream.

Issue #420, opened 2017-02-23
[9.0] Runbot warning due to 'must_have_dates', 'type', 'uom_id' constrain


LoisRForgeFlow:
In this pull request https://github.com/OCA/sale-workflow/pull/353 this fields <code>'must_have_dates', 'type', 'uom_id'</code> are added in <a href="https://github.com/akretion/sale-workflow/blob/e544824e533ef3cc67c31f7695758666a36d5f9d/sale_rental/models/product.py#L23">a constrain</a> but they are readonly in their creation (see <a href="https://github.com/OCA/sale-workflow/blob/10.0/sale_start_end_dates/sale.py#L83">one of them</a>).

This is producing a warning in some runbot logs like <a href="https://runbot2.odoo-community.org/runbot/static/build/3280961-418-56844f/logs/job_20_test_all.txt">this one</a> for example and turning the test  :red_circle:.

pedrobaeza:
Can you please make a PR removing then from the constraint decorator?

pedrobaeza:
The PR is done, so I close.

PR #421, opened 2017-02-27
[9.0][sale_order_variant_mgmt] FIX error in view

JordiBForgeFlow:
When the user is in group 'Technical Settings / Properties on lines', installs module 'sale_order_variant_mgmt' and tries to create a quote, at the time of entering a line in the quote and saving she will get this error: 

image

This PR provides the fix for that.

### DIFF
--- a/sale_order_variant_mgmt/views/sale_order_view.xml
+++ b/sale_order_variant_mgmt/views/sale_order_view.xml
@@ -17,6 +17,9 @@
                     />
                 </div>
             </xpath>
+            <xpath expr="//field[@name='order_line']//form" position="inside">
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
cc @tecnativa

rafaelbn (approved):
Functionally tested 👍 in runbot

My only doubt is why runbot is in red?

@sergio-teruel please test this one, thanks

JordiBForgeFlow:
Seems unrelated, due to https://github.com/OCA/sale-workflow/issues/420

pedrobaeza (approved):

PR #422, opened 2017-03-02
sale_quotation_number: Migration to v10.0

DarkoNikolovski:


sale_quotation_number/__manifest__.py
```diff
@@ -24,7 +24,7 @@
 
 {
     'name': 'Sale Quotation Numeration',
-    'version': '8.0.1.1.0',
+    'version': '10.0.1.1.0',
     'category': 'Sales Management',
     'summary': "Different sequence for sale quotations",
     'description': """
```
eLBati, line 30: Please move module description to README file, following the template
https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst
and replacing 'openerp' with 'odoo'

eLBati (changes requested):
Thanks!
Just a remark

antoniocanovas (changes requested):
Tested in #ocacodesprint 
It would be recommended:
- Include the README
- Include a field in sale order to keep the original quotation number.
- Allow searching by default for both codes in "quotations view" once it's confirmed.

rousseldenis:
Supersedes by https://github.com/OCA/sale-workflow/pull/591

PR #424, opened 2017-03-07
Activate 'sale layout' with sale_product_set_layout

guewen:
If you install this addon, it means you want to use sale layouts so it
makes sense to activate the feature.

### DIFF
--- a/sale_product_set_layout/__manifest__.py
+++ b/sale_product_set_layout/__manifest__.py
@@ -12,6 +12,7 @@
     ],
     'data': [
         'views/product_set.xml',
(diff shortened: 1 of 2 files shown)


gurneyalex (approved):

gurneyalex:
I tested this and it solves the useability I had when initially installing the module : I still needed to enable sales layout manually in the settings.

The product set + sales layout combination is great. Thanks to @petrus-v for this.

petrus-v (approved):
LGTM

angelmoya:
:+1:

PR #425, opened 2017-03-08
[10.0][ADD] Add new module 'sale_company_currency'

leemannd:
1. For tree view, when you have SO in multiple currencies, Odoo sums them regardless the different currencies. This modules removes this useless sum and add a column & sum in the company's currency.
 2. The field is also shown in form view after the total.

### DIFF
--- a/sale_company_currency/README.rst
+++ b/sale_company_currency/README.rst
@@ -1,0 +1,55 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+==============================
+Sale Order in company currency
+==============================
+
+This module adds functional fields to show sale order in the company currency:
+amount total.
+
+1. For tree view, when you have SO in multiple currencies, Odoo sums them up
+regardless the different currencies. This modules removes this useless sum and
+add a column & sums in the company's currency.
+
+2. The field is also shown in form view after the total.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/92/10.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/sale-workflow/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Denis Leemann <denis.leemann@camptocamp.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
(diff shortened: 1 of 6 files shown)


sale_company_currency/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Denis Leemann - Camptocamp SA
```
guewen: Copyright is attributed to the company: `Copyright 2017 Camptocamp SA`. The author can be found in the readme and in git.

sale_company_currency/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': "Company currency in sale orders",
+    'version': "10.0.1.0.0",
+    'author': "Camptocamp SA, "
```
guewen: Could you replace `Camptocamp SA` by `Camptocamp`?

guewen (changes requested):

leemannd:
@guewen Updated

gurneyalex (approved):

damdam-s (approved):
thanks

sale_company_currency/README.rst
```diff
@@ -0,0 +1,55 @@
...
+    :alt: License: AGPL-3
+
+==============================
+Sale Order in company currency
+==============================
+
+This module adds functional fields to show sale order in the company currency:
+amount total.
+
+1. For tree view, when you have SO in multiple currencies, Odoo sums them
```
elicoidal: sums them up

sale_company_currency/README.rst
```diff
@@ -0,0 +1,55 @@
...
+regardless the different currencies. This modules removes this useless sum and
+add a column & sum in the company's currency.
```
elicoidal: adds a column and sums

sale_company_currency/views/sale_order_view.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <!-- SO -->
+    <record id="view_order_tree" model="ir.ui.view">
+      <field name="name">sale.order.tree</field>
+      <field name="model">sale.order</field>
+      <field name="inherit_id" ref="sale.view_order_tree"/>
+      <field name="arch" type="xml">
+
+        <!-- Remove multicurrencies sums (makes no sense)-->
```
elicoidal, line 10: Actually it can make sense: if a company is working all sales in USD and company currency is EUR, both totals make sense (actually the one in SO currency more than the company currency).
I would rather have a filter on currency to allow to show only the SO for the given currency (and then this total makes sense when filtering)

elicoidal (changes requested):
Functionally could be a problem in some cases to remove the total for SO. Could we keep it

rafaelbn:
Hi @guewen , please you comments were attended, could you update your review? thanks!

@leemannd please review and and attend comments from @elicoidal thanks!

Both @guewen and @elicoidal must approve this PR in order to merge

sale_company_currency/views/sale_order_view.xml
leemannd: @elicoidal Having this sum on the bottom can be usefull but it is not working properly out of the box. If you have SO in EUR and USD, it will sums them up without taking in account the currency rate.
The idea of the filtering is good, but unless you activate it, it will output a wrong number and that's exactly what we try to avoid here.

elicoidal: @leemannd 
I have been working in trading many years and I do need to see the list in the currency of the PO and I dont care about my accounting currency. Removing the sum makes sense but removes as well a potential useful feature ("How much do I have to pay in USD?")

Will not block for it though.

elicoidal (approved):

guewen (approved):
Thanks

PR #429, opened 2017-03-17
10.0 mig sale automatic workflow exception

mourad-ehm:


rafaelbn:
Hi @mourad-ehm , https://github.com/OCA/sale-workflow/pull/391 is merged. Please could you rebase in order to see it for a review? Thanks

mourad-ehm:
Hi @rafaelbn, I rebased it.

lmignon (approved):

lmignon:
:+1:

setup/sale_automatic_workflow_exception/odoo/__init__.py
```diff
@@ -0,0 +1 @@
+__import__('pkg_resources').declare_namespace(__name__)
```
elicoidal, line 1: I think you dont need this files (setup*)

sale_automatic_workflow_exception/README.rst
```diff
@@ -0,0 +1,72 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================================
+Sale Automatic workflow exception
+=================================
+
+Auto installable Module for the compatibility
```
elicoidal: better use a list for readibility

sale_automatic_workflow_exception/README.rst
```diff
@@ -0,0 +1,72 @@
...
+
+This module will be installed automatically if Sale Automatic Workflow and Sale Exception
+are installed
+
+Configuration
+=============
+
+Activate the cron "Test Draft Orders":
+
+* Go to Setting > Technical > Automation > Scheduled Actions
```
elicoidal: Better use enumerated list ("#. ") for procedures

elicoidal (approved):
some non blocking details

setup/sale_automatic_workflow_exception/odoo/__init__.py
mourad-ehm: ping @lmignon,

lmignon: @elicoidal These files are required to be able to install this addon with pip.

elicoidal: @lmignon I thought they were automatically generated

rafaelbn:
I think coverage should be higher, is it possible to make some more test @mourad-ehm ? not blocking in my side

mourad-ehm:
OK for add some test

mourad-ehm:
@rafaelbn, travis is red beacause test of this module is fail "sale_cancel_reason". I don't know why ?
Hi @rafaelbn, @kittiu  do you have an idea ? Thanks

kittiu:
@mourad-ehm I check the error log, and found the exact location of error here,
```
`   File "/home/travis/odoo-10.0/addons/sale_margin/models/sale_order.py", line 50, in create
`     product_uom_id = self.env['product.uom'].browse(vals['product_uom'])
` KeyError: 'product_uom'
```
Seem the product_uom is not in vals during sale_order_line creation. And the test actually don't have that uom.
https://github.com/OCA/sale-workflow/blob/10.0/sale_cancel_reason/tests/test_sale_cancel_reason.py#L33
I try adding the uom here, let's see if it works.
https://github.com/OCA/sale-workflow/pull/494

kittiu:
@mourad-ehm from #494 above. I now found the same error now in test of sale_order_revision. Both are down to the same line of sale_margin/models/sale_order.py.
I am guessing, before we don't have this error until some addon depends in this repo on sale_margin, and so now we have the error. Possible?
Otherwise, we might need to fix all the test that create sale order line without uom.

mourad-ehm:
Hi @kittiu, OK thanks for response. I checked code and found that product_id and product_uom_id are required in 10.0. So the solution is that you propose.

mourad-ehm:
I add this PR to fix this bug https://github.com/OCA/OCB/pull/625

mourad-ehm:
Hi @rafaelbn could you merge this PR please ?

PR #432, opened 2017-03-20
[10.0][ADD][sale_covenant] v10 new module sale_covenant WIP

bealdav:
Allowing to define settings in Sale fields (default values, allowed values, etc) according settings on a new object Sale Covenant (contract).
Use this module to define specific use case to your market, i.e.:

- specific market: department store,
- deliver products with white mark
- sale to several clients of the same market but invoice a unique partner

depends on https://github.com/OCA/server-tools/pull/784

ping @sebastienbeau 

### This module add this object: sale.covenant

cov

==============

### Define onchange rule here: you can reuse this object to drive pricelist choice

covenant
=============

### Now you can play onchange without to write any code.
sale

-----------------------

### pricelist is set dynamically according to Covenant field and set to readonly via field_view_get()

---------------------

- [ ] implements onchange with selection field
- [ ] add tests


bealdav:
What do you think of the use case ? ping @lasley @rvalyi

sale_covenant/models/sale.py
```diff
@@ -0,0 +1,45 @@
...
+    @api.multi
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        """ Update 'Covenant' fields when the partner is changed:
+        """
+        res = super(SaleOrder, self).onchange_partner_id()
+        if not self.partner_id:
+            self.covenant_id = False,
+        else:
+            self.covenant_id = self.partner_id.covenant_id.id
```
mourad-ehm: Here ,you affect and id to an object (self.covenant_id (object) = self.partner_id.covenant_id.id (id)) ?  that must generate a bug, is not it?

mourad-ehm:
travis is failed because the partner_id is now required in sale_covenant (in covenant_data.xml).

mourad-ehm:
Hi @alexis-via , I would back port this module to v 8.0. Please, could you validate the last changes before I do it ?

rafaelbn:
Hi @bealdav , is this still a WIP? Let us know please. Thanks

bealdav:
Testable as it but we currently refactor the v8 version of its dependency base_onchange_rule (data structure).
We can stay as wip here.

Please Rafael, express your opinion on these modules and the different use case you can imagine, thanks

bealdav:
Replaced by sale_agreement from contract repo

PR #404, opened 2017-02-09
[10.0][Add] sale_manual_delivery

leemannd:
Sale Manual Delivery
====================
The Goal of this module is to allow the manual creation of delivery slips. If activated,
a sale order won't direcly impact directly the stock. It will not make a reservation on the stock.
It allows the delivery and the impact on stock to be done manually when needed.
The goal is to be used on long term projects where not all the material is shipped at once.

PR #434, opened 2017-03-22
[ADD] [9.0] sale_delivery_block

LoisRForgeFlow:
Sale Delivery Block
===================

This module extends the functionality of sales to allow you to block the
creation of deliveries from a sales order and give a reason.

Configuration
=============

To configure this module, you need to:

#. Go to 'Sales > Configuration > Sales > Delivery Block Reason'.
#. Create the different reasons that can lead to block the deliveries of a
   sales order.
#. Add some users to the group 'Release Delivery Block in Sales Orders'.

Additionally, you can set a customer with a 'Default Delivery Block Reason'
policy to add that delivery block to his sales by default:

#. Go to 'Sales > Sales > Customers'.
#. In the 'Sales & Purchases' add a 'Default Delivery Block Reason'.
#. The 'Default Delivery Block Reason' will be added
   automatically when creating a new sales order for the customer.

Usage
=====

#. Create a new sales order and provide a 'Delivery Block Reason'.
#. Confirm Sale (No delivery would be created).
#. Release Delivery Block when it is time to create the deliveries for
   the sales order.

### DIFF
--- a/sale_delivery_block/models/res_partner.py
+++ b/sale_delivery_block/models/res_partner.py
@@ -3,7 +3,7 @@
 #   (http://www.eficent.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from openerp import fields, models
+from openerp import api, fields, models
 
 
 class ResPartner(models.Model):
@@ -14,3 +14,9 @@ class ResPartner(models.Model):
         string='Default Delivery Block Reason',
         help="Set a reason to block by default the deliveries in this "
              "customer sales orders.")
+
+    @api.model
+    def _commercial_fields(self):
+        commercial_fields = super(ResPartner, self)._commercial_fields()
+        commercial_fields.append('default_delivery_block')
+        return commercial_fields
--- a/sale_delivery_block/models/res_partner.py
+++ b/sale_delivery_block/models/res_partner.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    default_delivery_block = fields.Many2one(
+        comodel_name='sale.delivery.block.reason',
+        string='Default Delivery Block Reason',
+        help="Set a reason to block by default the deliveries in this "
+             "customer sales orders.")
--- a/sale_delivery_block/models/sale_order.py
+++ b/sale_delivery_block/models/sale_order.py
@@ -30,8 +30,8 @@ class SaleOrder(models.Model):
         """Add the 'Default Delivery Block Reason' if set in the partner."""
         res = super(SaleOrder, self).onchange_partner_id()
         for so in self:
-            if so.partner_id.default_delivery_block:
-                so.delivery_block_id = so.partner_id.default_delivery_block
+            so.delivery_block_id = so.partner_id.default_delivery_block or \
+                False
         return res
 
     @api.multi
@@ -40,6 +40,15 @@ class SaleOrder(models.Model):
         self.write({'delivery_block_id': False})
         for order in self:
             order.order_line._action_procurement_create()
+
+    @api.multi
+    def copy(self, default=None):
+        new_so = super(SaleOrder, self).copy(default=default)
+        for so in new_so:
+            if (so.partner_id.default_delivery_block and not
+                    so.delivery_block_id):
+                so.delivery_block_id = so.partner_id.default_delivery_block
+        return new_so
 
 
 class SaleOrderLine(models.Model):
--- a/sale_delivery_block/models/sale_order.py
+++ b/sale_delivery_block/models/sale_order.py
@@ -38,6 +38,11 @@ class SaleOrderLine(models.Model):
     @api.multi
     def _action_procurement_create(self):
         new_procs = self.env['procurement.order']
+        # When module 'sale_procurement_group_by_line_is installed' we do not
+        #  want to use this method but call super. See module
+        # 'sale_delivery_block_proc_group_by_line
+        if 'group_by_line' in self.env.context:
+            return super(SaleOrderLine, self)._action_procurement_create()
         for line in self:
             if not line.order_id.delivery_block_id:
                 new_procs += super(SaleOrderLine,
--- a/sale_delivery_block/models/sale_order.py
+++ b/sale_delivery_block/models/sale_order.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models, _
+from openerp.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    @api.constrains('delivery_block_id')
+    def _check_not_auto_done(self):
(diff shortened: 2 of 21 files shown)


LoisRForgeFlow:
Runbot red because of https://github.com/OCA/sale-workflow/pull/419

LoisRForgeFlow:
commit https://github.com/OCA/sale-workflow/pull/434/commits/b18228a6b69d8673f80f5626fd6a1aaedce7c89d adds a similar feature as the one provided by the module <code>partner_prepayment</code> in v8 but adapted to this module. As a consequence, the migration of the referred module might be not needed.

guewen:
Isn't it similar to https://github.com/OCA/sale-workflow/pull/404 ?

JordiBForgeFlow:
@guewen, not exactly the same. Even would be nice that they could work together. The block prevents users from being able to deliver, until the block is released. Typical block reasons are pay before delivery or credit check

sale_delivery_block/models/res_partner.py
```diff
@@ -0,0 +1,16 @@
...
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    default_delivery_block = fields.Many2one(
```
atchuthan, line 12: It is probably a good idea to make it as a **commercial field** such that default delivery block value is applied across all its contact and the company

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,61 @@
...
+        string='Delivery Block Reason', readonly=True,
+        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]})
+
+    @api.multi
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        """Add the 'Default Delivery Block Reason' if set in the partner."""
+        res = super(SaleOrder, self).onchange_partner_id()
+        for so in self:
+            if so.partner_id.default_delivery_block:
```
atchuthan: Remove if block such that the value is reset to False if there is no default block selected in the partner
**so.delivery_block_id = so.partner_id.default_delivery_block or False**

atchuthan (changes requested):
Base functionality works fine(Tested locally) :+1: 
but needs minor changes/tweaks

While duplicating a Sales order, the default delivery block is not set and we need to manually select the default delivery block.  We need this to be automated :)

AaronHForgeFlow (approved):
tested on local + light code review:+1:

LoisRForgeFlow:
@atchuthan Thank you for the review! I have just attended your suggestions. Please have a look.

atchuthan (changes requested):
Sorry for the delay in response @lreficent 
Thanks for the fix and it works fine.

Also, when I started testing I found a bug.  The case scenario is "Cancel the Sales order and reset to draft where the default sale block is not set".

atchuthan:
@lreficent any update?
let me know if you need any help regarding this PR

LoisRForgeFlow:
@atchuthan Sorry, totally forgot to update. Going to do right now!

LoisRForgeFlow:
@atchuthan I'm now checking what you said. That happens in the following scenario:

1. You create a quotation with a devivery block reason.
2. You Confirm Sale.
3. You **Release Delivery Block**. So shipments were created.
4. You cancel
5. You set to quotation.

Don't you think that's a very uncommon scenario and, why to reopen quotation with the delivery block by again if you had released the block?

atchuthan:
@lreficent sure, this is an uncommon scenario.

atchuthan (approved):
:+1:

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models, _
+from openerp.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
```
Garamotte: The API should not allow you to call `self.delivery_block_id` in an `@api.model` method...
A constraint method must be `@api.multi`.

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+        for order in self:
+            order.order_line._action_procurement_create()
+
+    @api.multi
+    def copy(self, default=None):
+        new_so = super(SaleOrder, self).copy(default=default)
+        for so in new_so:
+            if (so.partner_id.default_delivery_block and not
+                    so.delivery_block_id):
+                so.delivery_block_id = so.partner_id.default_delivery_block
```
Garamotte, line 51: I think this could be done by `copy=False` on the `delivery_block_id` field, without overriding the copy method.

atchuthan: when we use copy=False, it will reset the value to False by default, so we used copy to provide a default value

Garamotte: You're right, thanks for the explanation.

sale_delivery_block/models/sale_order.py
LoisRForgeFlow: you're right, fixing it.

Thanks for reviewing!

atchuthan:
@sylvain-garancher corrections are done

sale_delivery_block/tests/test_sale_delivery_block.py
```diff
@@ -0,0 +1,94 @@
...
+        with self.assertRaises(ValidationError):
+            so.write({
+                'delivery_block_id': block_reason.id,
+            })
+
+    def procurement_comp(self, so):
+        """count created procurements"""
+        count = 0
+        for line in so.order_line:
+            count += len(line.procurement_ids)
```
Garamotte, line 67: `count = len(so.order_line.mapped('procurement_ids'))` ?

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]})
+
+    @api.multi
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        """Add the 'Default Delivery Block Reason' if set in the partner."""
+        res = super(SaleOrder, self).onchange_partner_id()
+        for so in self:
+            so.delivery_block_id = so.partner_id.default_delivery_block or \
+                False
```
Garamotte, line 35: I think you can remove `or False` here.

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+        return res
+
+    @api.multi
+    def action_remove_delivery_block(self):
+        """Remove the delivery block and create procurements as usual."""
+        self.write({'delivery_block_id': False})
+        for order in self:
+            order.order_line._action_procurement_create()
```
Garamotte, line 43: `self.mapped('order_line')._action_procurement_create()` ?

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @api.multi
+    def copy(self, default=None):
+        new_so = super(SaleOrder, self).copy(default=default)
+        for so in new_so:
```
Garamotte, line 48: Remove the for loop. `copy` will always create a single record.

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+            if (so.partner_id.default_delivery_block and not
```
Garamotte, line 49: `if so.delivery_block_id` is sufficient in my opinion.
If this value is empty, no matter if the partner has a value or not.

Garamotte (approved):
Thanks.
I added some minor new API-magic comments, but nothing blocking.

LoisRForgeFlow:
@sylvain-garancher Thanks for the tips! I wanted to add that api magic but It is already merged, these tips should be taken into account when migrating to v10.  :memo:

PR #437, opened 2017-03-27
[10.0][ADD] sale_triple_discount module

tafaRU:
This PR is still WIP according to its label.
@ingadhoc FYI.

# TODO List
- [x] migrate from @ingadhoc repo
- [x] adapt code to OCA guidelines
- [x] expand summary
- [x] add some unit tests
- [x] add defaults for discount[123] to partner


pedrobaeza:
Please avoid plurals. What about sale_triple_discount?

Isn't any product_pricelist_triple_discount?

tafaRU:
@pedrobaeza I think it's better `sale_triple_discount` than `product_pricelist_triple_discount` as the module doesn't involve pricelist.

sale_triple_discount/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* ADHOC SA
```
elicoidal: If this is the author, it is already in the manifest and no need to repeat here the name

sale_triple_discount/models/account_invoice_line.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @api.multi
+    @api.depends('discount1', 'discount2', 'discount3')
+    def get_discount(self):
+        for line in self:
+            discount_factor = 1.0
+            for discount in [self.discount1, self.discount2, self.discount3]:
+                discount_factor = (
+                    discount_factor * ((100.0 - discount) / 100.0))
+            line.discount = 100.0 - (discount_factor * 100.0)
```
elicoidal: you allow negative discounts: is it desired?

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    discount1 = fields.Float(
+        'Discount 1 (%)',
+        digits=dp.get_precision('Discount'),
+        readonly=True,
+        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]}
```
elicoidal: why control here and not in invoice lines?

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @api.multi
+    @api.depends('discount1', 'discount2', 'discount3')
+    def get_discount(self):
+        for line in self:
+            discount_factor = 1.0
+            for discount in [self.discount1, self.discount2, self.discount3]:
+                discount_factor = (
+                    discount_factor * ((100.0 - discount) / 100.0))
+            line.discount = 100.0 - (discount_factor * 100.0)
```
elicoidal: same remark as above

sale_triple_discount/views/sale_order_view.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record model="ir.ui.view" id="sale_order_three_discounts_form_view_inh">
+        <field name="name">sale.order.three_discounts.form.inh</field>
```
elicoidal: 'inh' stands for?

elicoidal (changes requested):
I miss UT. some questions and minor changes

sale_triple_discount/views/account_invoice_view.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record model="ir.ui.view" id="account_invoice_three_discounts_form_view_inh">
+        <field name="name">account.invoice.three_discounts.form</field>
```
elicoidal: adapt the name to the newer module name

sale_triple_discount/i18n/es.po
```diff
@@ -0,0 +1,48 @@
...
+#: model:ir.model.fields,field_description:sale_three_discounts.field_sale_order_line_discount3
+msgid "Discount 3 (%)"
+msgstr "Descuento 3 (%)"
+
+#. module: sale_three_discounts
+#: model:ir.model,name:sale_three_discounts.model_account_invoice_line
+msgid "Invoice Line"
+msgstr "Línea de Factura"
+
+#. module: sale_three_discounts
```
elicoidal: you will have to adapt

pedrobaeza:
What I mean is that I would expect also a module called `product_pricelist_triple_discount` with the possibility to set the 3 discounts in pricelists.

sale_triple_discount/models/account_invoice_line.py
tafaRU: @elicoidal, you mean discount[123]?

sale_triple_discount/models/sale_order_line.py
tafaRU: @elicoidal, because invoice fields are already editable only in state draft.

sale_triple_discount/models/account_invoice_line.py
tafaRU: @elicoidal, we prefer to keep it as they are anyway.

tafaRU:
@elicoidal, your comments should have been addressed. Thanks!

sale_triple_discount/models/account_invoice_line.py
elicoidal: ok (this is the standard Odoo behavior as well)

sale_triple_discount/models/sale_order_line.py
elicoidal: ok: thanks for clarifying

elicoidal (approved):

jjscarafia:
@tafaRU Please check this [modification](https://github.com/ingadhoc/sale/pull/71) that fix a compatibility for discounts on purchase invoices.
We are also thinking on a purchase_triple_discount module that extends to purchase orders and purchase invoices.

eLBati (changes requested):
tested on tunbot:
 - open a new sale order
 - add a line
 - try to change discount 1
discount 1 is always set back to 0
If I set the default value on partner, the value is kept

tafaRU:
@eLBati, fixed, thanks!

tafaRU:
@jjscarafia,

thanks for pointing it out, I think the changes related to https://github.com/ingadhoc/sale/pull/71 are more appropriate in a module such as purchase_triple_discount, though.

eLBati:
@tafaRU try the following steps:
 - Add a new sale order line
 - Select '[CARD] Graphics Card' product
 - Set 10 in 'discount 1 (%)'
 - Go to 'unit price' field and set '100'

'Discount 1 (%)' is 10 and 'Discount (%)' is 0

I see 2 ways: if the user is suppsed to re input discounts , the 3 discounts fields must be set to 0.
Otherwise, better, 'discount (%)' should be recomputed according to new values

tafaRU:
@eLBati, fixed again, please check it!

jjscarafia:
@tafaRU I've not address the changes proposed here https://github.com/ingadhoc/sale/pull/71/files in a purchase_... module because I would add dependency for purchase module.I think thas modification is needed in this PR because if not you won't be able to record a supplier invoice with discounts (no matter you use purchase module or not).

And yes, we would like also to work in a module "purchase_triple_discount" that allows to define 3 discounts on purchase order and supplier invoices.

tafaRU:
@jjscarafia, ok, thanks for the clarification!
I also see https://github.com/ingadhoc/sale/commit/b86aaac76f4c465019f184b453ca31640d2a6da8, why did you choose to log instead to raise an error?
Could you please make a PR against my branch containing all changes you would include?

eLBati:
@tafaRU please also fix tests

rafaelbn:
Hi @tafaRU , we are reviewing, please take a look here, travis fails and @eLBati comment is pending! Thanks

tafaRU:
@rafaelbn, @eLBati: I'm working on fixing the tests.

tafaRU:
@eLBati, @rafaelbn: after https://github.com/OCA/sale-workflow/pull/437/commits/b42040e081c5ba6eba92243623be494ba59dcaf3 Travis looks green. Can we merge it?

eLBati (approved):

tafaRU:
@rafaelbn, could you please merge it?

MiquelRForgeFlow (approved):
LGTM :+1:

jjscarafia (changes requested):
To avoid some error we have change the approach on the original module so discount field is not computed, instead we write it with ochange and constraint decorator, please check [here](https://github.com/ingadhoc/sale/blob/9.0/sale_three_discounts/models/account_invoice.py)

To replicate the error:
1) create new invoice
2) add a line with unit price 100, discount 50% and tax 15%
3) Save, the discount is not applied to the tax (tax is 15 and should be 7,5) (check image attached).

The approach on the link solve this issue and some issues we also have before


seleccion_064

jjscarafia:
@tafaRU please check my comment, sorry for the late feedback, we have make this change a few weeks ago and it is on production and working fine.
We haven't done that change on [sale orders](https://github.com/ingadhoc/sale/blob/9.0/sale_three_discounts/models/sale.py#L26) but we would and should

rafaelbn:
@tafaRU could you please review https://github.com/OCA/sale-workflow/pull/437#pullrequestreview-46567065 from @jjscarafia ? Thanks

sale_triple_discount/models/account_invoice_line.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# Copyright 2017 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+import openerp.addons.decimal_precision as dp
```
chienandalu, line 7: Why not import `odoo.addons.decimal_precision as dp`?

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# Copyright 2017 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+import openerp.addons.decimal_precision as dp
```
chienandalu, line 7: Why not import `odoo.addons.decimal_precision as dp`?

chienandalu (approved):
Functional and code review. Just two minor comments.

sale_triple_discount/models/account_invoice_line.py
```diff
@@ -0,0 +1,69 @@
...
+    @api.model
+    def _default_discount3(self):
+        default_discount3 = 0.0
+        partner_id = self.env.context.get('partner_id', False)
+        if partner_id:
+            partner = self.env['res.partner'].browse(partner_id)
+            default_discount3 = partner.default_discount3
+        return default_discount3
+
+    discount1 = fields.Float(
```
pedrobaeza, line 40: Don't redefine this. Use normal discount instead.

sale_triple_discount/models/account_invoice_line.py
```diff
@@ -0,0 +1,69 @@
...
+        'Discount 2 (%)',
+        digits=dp.get_precision('Discount'),
+        default=_default_discount2,
+    )
+    discount3 = fields.Float(
+        'Discount 3 (%)',
+        digits=dp.get_precision('Discount'),
+        default=_default_discount3,
+    )
+    discount = fields.Float(
```
pedrobaeza, line 55: Don't change this to computed

sale_triple_discount/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2017 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    default_discount1 = fields.Float('Default Discount 1 (%)')
```
pedrobaeza, line 11: This is another feature. Don't should be in this module.

pedrobaeza (changes requested):
:-1: in this form. Invoice discounts should go in a module in OCA/account-invoicing called __account_invoice_triple_discount__, as if we want to reuse this for purchase we must depend on sale, which is a nonsense.

tafaRU:
@pedrobaeza, I understand your reasons, expecially in the light of https://github.com/OCA/account-invoicing/tree/10.0/account_invoice_triple_discount, but considering we have already installed this module in some production database of our customers, we prefer to keep it as is to avoid further effort.

I would close this then.

pedrobaeza:
OK, as you wish, but I think we can make compatible both. What are your critical points?

pedrobaeza:
Closing according https://github.com/OCA/sale-workflow/pull/437#issuecomment-325614967 in favor of #525. We have respected the commits there any way.

tafaRU:
> We have respected the commits there any way.

Thank you so much! You are a real gentlemen :bowtie:

pedrobaeza:
:smiley:

PR #440, opened 2017-03-29
[Add][10.0] New module sale_layout_hidden

leemannd:
This module allows to Hide some sections in the printout of sale orders

### DIFF
--- a/sale_layout_hidden/README.rst
+++ b/sale_layout_hidden/README.rst
@@ -6,11 +6,19 @@
 Sale Layout Hidden
 ====================
 
-This module allows to Hide some sections in the printout of sale orders
+This can be used to add service products in a quotation with a price of 0 which
+will create a task in a project without showing them to the customer. This is
+very useful when used in conjunction with product sets (see sale_product_set_layout).
+
+The module also hides the lines from the invoice.
 
 
 Usage
 =====
+
+Installing this module does not enable sale layout. You need to activate it in the sales settings.
+
+Then you can check the hidden field on sale layout categories. These categories will not be displayed.
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
--- a/sale_layout_hidden/README.rst
+++ b/sale_layout_hidden/README.rst
@@ -1,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Sale Layout Hidden
+====================
+
+This module allows to Hide some sections in the printout of sale orders
+
+
+Usage
+=====
+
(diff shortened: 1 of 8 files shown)


elicoidal (approved):

sale_layout_hidden/README.rst
```diff
@@ -0,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Sale Layout Hidden
+====================
+
+This module allows to Hide some sections in the printout of sale orders
+
```
gurneyalex: Can you add a description of why this is useful? Suggestion:

"""
This can be used to add service products in a quotation with a price of 0 which will create a task in a project without showing them to the customer. This is very useful when used in conjunction with product sets (see `sale_product_set_layout`).

The module also hides the lines from the invoice. 
"""

sale_layout_hidden/README.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Usage
+=====
+
```
gurneyalex, line 18: """Installing this module does not enable sale layout. You need to activate it in the sales settings. 

Then you can check the `hidden` field on sale layout categories. These categories will not be displayed. """

gurneyalex (changes requested):

sale_layout_hidden/README.rst
leemannd: @gurneyalex Thanks for the comments. It has been updated.

sale_layout_hidden/models/account_invoice.py
```diff
@@ -0,0 +1,41 @@
...
+                if section['lines'][0].layout_category_id.hidden:
+                    continue
+                new_page.append(section)
+            if new_page:
+                pages.append(new_page)
+        return pages
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'sale.order.line'
```
angelmoya: I think that it's a copy&paste error, and it must be 'account.invoice.line'

sale_layout_hidden/models/sale_layout.py
```diff
@@ -0,0 +1,31 @@
...
+
+class SaleLayoutCategory(models.Model):
+    _inherit = 'sale.layout_category'
+
+    hidden = fields.Boolean(default=False)
+
+    @api.model
+    def create(self, values):
+        if values.get('hidden'):
+            values = values.copy()
```
angelmoya, line 16: Why did you do values.copy()?

gurneyalex: @angelmoya we are changing the values (see line 17 below) => this avoids changing the dictionarly in the caller's context as dictionaries are mutable.

sale_layout_hidden/models/account_invoice.py
gurneyalex: good catch!

gurneyalex (approved):

angelmoya (approved):

PR #464, opened 2017-05-09
[8.0] Add copy=False on field date_validity

alexis-via:
The field "date_order" has copy=False (cf https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L204), so it should be the same for the field "date_validity".

When you duplicate an old sale order, the "Date" field is automatically set to today, so the validity date should also be reset and put to its default value.


LoisRForgeFlow (approved):
It makes sense :+1:

AaronHForgeFlow (approved):
:+1: Thank you for the fix

PR #468, opened 2017-05-12
8.0 add sale_documents_comments

eLBati:
moving from https://github.com/OCA/sale-workflow/pull/290


rafaelbn (approved):
Tested in runbot 👍

AaronHForgeFlow (approved):
:+1: Functional test

chienandalu (approved):

pedrobaeza (changes requested):
Please avoid plurals on the name

eLBati:
@pedrobaeza in this case I think plural is more appropriate, because the purpose of the module is to involve several documents with several comments

pedrobaeza:
Well, semantics shouldn't enforce this, as if not, all Odoo will be full of plurals: `sales.order` instead of `sale.order` and so on. The name itself of the module should be `sales_documents_comments` for being correct. I suggest a more neutral name `sale_comment_propagation`

eLBati:
@pedrobaeza ok, done

pedrobaeza (approved):

PR #280, opened 2016-03-17
[ADD] add module to deliver when creating invoice

laetitia-gangloff:
In sale advance payment, add option to deliver at the same time than invoice.
If picking are not validable, an error is raised.

PR #469, opened 2017-05-12
[10.0][MIG] port sale_invoice_auto_deliver to 10.0

Cedric-Pigeon:
Migration unmerged PR https://github.com/OCA/sale-workflow/pull/280 on 8.0 to 10.0


antoniocanovas:
Testing in #ocacodesprint
Everything works fine but the "done quantity" which is not updated after invoicing:
 
captura de pantalla 2017-05-30 a las 16 02 51

Cedric-Pigeon:
@rafaelbn @antoniocanovas bug fixed and I extended tests.

Cedric-Pigeon:
@rafaelbn Could you please remove the need fixing  as I fixed the bug

lmignon (approved):
@OCA/crm-sales-marketing-maintainers Is it possible to merge this one. It's in production from months...

pedrobaeza:
Isn't this overlapping somehow with `sale_automatic_workflow`?

Cedric-Pigeon:
@pedrobaeza Not really. This addons adds an option to confirm the deliver order in the same time you generate the invoice. It is not automatic, it is on demand through the standard invoicing wizard

pedrobaeza:
But the same effect can be got with the other module, acting on sales order confirmation. The only difference is that you trigger everything together. Has to be in a different moment in time?

Cedric-Pigeon:
Yes but the use case is different. They prepare the goods and we the customer comes to pick it up  they gave it to him. They generate the invoice to get payment and confirm the stock ongoing in a while.
i agree the endpoint is the same but the way to go is different

pedrobaeza:
OK, then let's go ahead with this. Do you want to keep both commits?

Cedric-Pigeon:
@pedrobaeza I squashed last commit. Thanks

sale_invoice_auto_deliver/README.rst
```diff
@@ -0,0 +1,78 @@
...
+To install this module, you need to:
+
+* Click on install button
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to ...
```
pedrobaeza: There are no usage instructions. Remove this section if empty or move part of the main description that are usage instructions.

sale_invoice_auto_deliver/README.rst
```diff
@@ -0,0 +1,78 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/8.0
```
pedrobaeza: Incorrect link. Maybe you should switch to new README by fragments system.

sale_invoice_auto_deliver/README.rst
```diff
@@ -0,0 +1,78 @@
...
+in order to deliver products at the same time than the invoice. This functionality can be useful in the case where products are already take out of the stock.
+
+When invoice sale order line it is possible to choose the option : "Create Invoice + Automatic Deliver" to deliver the product and invoice it.
+
+Consider as pre-requirement that all products must be available.
+
+Installation
+============
+
+To install this module, you need to:
```
pedrobaeza: These instructions are useless, so better to remove them.

pedrobaeza (changes requested):
Sorry for this late review, but when I was going to merge, I see an incorrect README. Can you please fix it for merging?

Cedric-Pigeon:
@pedrobaeza You are absolutely right. I just reworked this part and switched to fragments system

pedrobaeza:
One last thing to make Travis happy:
```
sale_invoice_auto_deliver/readme/USAGE.rst:6: [E7901(rst-syntax-error), ]  Definition list ends without a blank line; unexpected unindent.
 ```

pedrobaeza (approved):

PR #478, opened 2017-05-19
[8.0][IMP] sale_order_type: fix problem when making a supplier refund

oihane:


sale_order_type/models/account_invoice.py
```diff
@@ -17,7 +17,9 @@ def _prepare_refund(self, invoice, date=None, period_id=None,
                         description=None, journal_id=None):
         values = super(AccountInvoice, self)._prepare_refund(
             invoice, date, period_id, description, journal_id)
-        if invoice.origin:
+        if (values.get('type') in ['out_invoice', 'out_invoice'] or
```
pedrobaeza: out_invoice 2 times? And you only need to check invoice, not values.

oihane: ups, misspelled

pedrobaeza (approved):

agaldona (approved):

PR #320, opened 2016-08-01
 [ADD][8.0] Sale procurement group by requested date

AaronHForgeFlow:
# Sale Procurement Group by Requested date

This module creates different procurements groups for different requested dates in a sale order line when the sale order is confirmed. It depends on sale_sourced_by_line so this module will group procurements also by the warehouse in the sale order_line.
### Installation

This module depends on the modules sale_procurement_group_by_line and
sale_sourced_by_line.
### Configuration

No special configuration is required.
### Usage
- Add a requested date for a sale order line.
- Confirmthe sale order

PR #481, opened 2017-05-22
[ADD][9.0] sale_order_line_date

serpentcs-dev1:
## Sale order line date

This module adds requested date to a sales order lines and propagate it to
stock moves and procurement.

When the requested date of the whole sale order is modified the requested date
of the lines change to match.

## Usage

Add a requested date for a sale order line. Confirm the sale. The scheduled date of the procurement will be the same as in the sale order line.


serpentcs-dev1:
@aheficent Could you please review.

AaronHForgeFlow (approved):
Tested on local + code review :+1: LGTM

sale_order_line_date/README.rst
```diff
@@ -14,12 +14,12 @@ of the lines change to match.
 Usage
 =====
 
-Create a Quotation or a Sales Order and fill the requested date in the sale
+Create a Quotation or a Sales Order and it fills the requested date in the sale
```
LoisRForgeFlow: I think this was fine before, because with the current sentence I understand that requested date from the sale order will be passed to lines, which is not happening.

LoisRForgeFlow (commented):
Tested on local, just a comment

sale_order_line_date/README.rst
AaronHForgeFlow: @lreficent I did not notice of that. This Pull Request must be fixed then. The behavior in previous versions is like the current readme describes, the changes in the requested date in the sale order affect all the lines. I'll check it again

serpentcs-dev1: @aheficent I guess the current behaviour is properly working I just need to change the Readme description that's what @lreficent is suggesting

serpentcs-dev1: @lreficent  Actually the `requested_date` on Sale Order line is populated from the requested date onchange of Sale Order.

AaronHForgeFlow: @darshan-serpent We can discuss if we should change the previous behavior but the previous behavior is exactly as the readme describes. The method def _onchange_requested_date_ is supposed to do that. It's not working properly? :confused:

AaronHForgeFlow: Just to be clear @darshan-serpent  I agree that the readme is correct after your changes now. However, the correct behavior is to change the requested dates of all the lines if the user modifies the requested date in the sale order.

AaronHForgeFlow: @lreficent I just checked again and the requested date is passed from the sale order to the lines. I think it's working. I'd not modify anything

serpentcs-dev1: Yes, double checked its working, Thanks @aheficent @lreficent for your review!

rafaelbn:
@lreficent please go ahead with your review

rafaelbn:
@Eficent

sale_order_line_date/README.rst
LoisRForgeFlow: not it is not if it starts empty:

1. Create new quotation with a line with not required date
image

2. Change the required date for the SO:
image

3. Line wasn't updated:
image

serpentcs-dev1: @lreficent Yes you are right, the onchange does not work in the case you mentioned.

I am getting a blank recordset in `self.order_line` which is because the line is never created, so now I am not really sure how to update the requested_date field in the o2m with the help of onchange

AaronHForgeFlow: I would modify the write method, so if the requested date is not in vals it would mean that it did not changed

AaronHForgeFlow:
@darshan-serpent I would use the write method. I tested your solution and it fails when:

- Create a SO with a line and no requested date not in the line nor in the SO
- Save
- Edit and add a requested date for the SO
- The line still has not requested date

:cactus:

AaronHForgeFlow:
Hi @darshan-serpent this last solution fails when:

- Create a SO with required date and a line with other required date
- Save
- Change the quantity in the line
- The requested date in the line is also changed :disappointed: 

What I meant is to override the write method in the SO not in the line. In my opinion that would be enough and the create method in the line would not be necessary.

Regards

AaronHForgeFlow (approved):
Tested on local :+1: Thanks for your contribution

LoisRForgeFlow (approved):
Tested on local, LGTM now :+1:

LoisRForgeFlow:
@rafaelbn This is ready to review/merge :wink:

LoisRForgeFlow:
@darshan-serpent can you attend travis and runbot?

serpentcs-dev1:
Travis is green, though there is an import error in an [outside module](https://travis-ci.org/OCA/sale-workflow/jobs/239888123#L1230) 

Should it need to be resolved in this PR?

While Runbot shows red but I tested the functionality it works fine.

@lreficent

LoisRForgeFlow:
@darshan-serpent 

> Travis is green, though there is an import error in an outside module. Should it need to be resolved in this PR?

No, I think it should be solved in a separeted PR. 

> While Runbot shows red but I tested the functionality it works fine.

Yes, runbot is red due to an odoo issue that is currently being attended. Not related to this PR.

sale_order_line_date/README.rst
```diff
@@ -0,0 +1,63 @@
...
+stock moves and procurements.
+When the requested date of the whole sale order is modified the requested date
+of the lines change to match.
+
+Usage
+=====
+
+Create a Quotation or a Sales Order and it fills the requested date in the sale
+order line
+
```
rafaelbn, line 19: Please could you explain a little bit more the usage? I tested but what must a expect?

AaronHForgeFlow:
Hi @rafaelbn,

This module does not split the picking. The schedule date in the picking is the expected date of the earlier move.
The expected date of the move is the one in the SO line:
image

image

And the scheduled date in the procurement i salso the one in the SO line:
image

AaronHForgeFlow:
@rafaelbn This module it's a base module for the sale_procurement_group_by_requested_date (PR in v8 https://github.com/OCA/sale-workflow/pull/320)

sale_procurement_group_by_requested_date  creates different picking groups depending on the requested date in the SO lines. It depends on this modules already merged in OCA: sale_procurement_group_by_line, which is a base module to create different procurement groups based on different criteria.

Another module that uses the sale_procurement_group_by_line is the sale_source_by_line, also merged in OCA; sale_source_by_line splits the pickings depending on the source warehouse specified in the SO line.

rafaelbn (approved):
I understand https://github.com/OCA/sale-workflow/pull/481#issuecomment-308057529
I tested in runbot and as this modules says it creates a procurement per line with the line date

rafaelbn:
@darshan-serpent please could you squash your commits before merging? thanks!

serpentcs-dev1:
@rafaelbn Squashed, thanks for your review!

PR #396, opened 2017-01-25
[MIG] - v10 migration of sale_order_revision

JayVora-SerpentCS:

PR #482, opened 2017-05-22
[MIG] sale_order_revision: Migration to 10.0

kittiu:
On thing different from 8.0
Since 8.0 already has the button "New Copy of Quotation" but 10.0 do not have. So, I add a button for it but rename it to "New Revision of Quotation".


pedrobaeza:
There was already a PR for this module: #396. Can you check that one instead?

Issue #472, opened 2017-05-12
[10.0] Module sales_rental creating Procurement Exceptions


rogerv95:
I've been trying to set up the rental module for odoo community version 10.0 but I can't get it to work. I've been able to use it in older versions (v7, v8) but not in v10. I follow all the instructions for the new version and I know the workflow from the older versions but in this one the rental never works.

The creation of the product variation for rental works fine but when I create a sales order with the rental product in it instead of creating the delivery and recieve operations to rent out and in the product it creates a procurement exception as it tries to send to the costumer "days of the product" and not the product itself. Also the receive order never gets created.
And, if I go to the Sales page > Rental to se the open rentals I have any record is created after confirming the sale and if I try to create a rental from there I only get a page with all text in it but with 0 functionality.

<img width="1465" alt="captura de pantalla 2017-05-13 a les 1 02 14" src="https://cloud.githubusercontent.com/assets/28661848/26019714/de572508-3777-11e7-9d3d-ff5f581352fd.png">

Anyone has tried it with 10.0 and had the same problems? Or someone has any idea what can be happening? I've hours and hours trying to figure it out by myself but I can't.

dvdhinesh:
@rogerv95 I suggest you to take a look at this [video...](https://youtu.be/7T8nmOYncjw) 
You should enable the **Properties on lines** access rights for the users. Without this access, user has no way to enter the **Rental Quantity.**

@alexis-via Will it be considered as a bug? This is the same for V9 also.

alexis-via:
@dvdhinesh This is well documented in the README of sale_rental so I don't we can consider it as a bug. But, what we should do, is automatically add all users in the "Properties on lines" group when installing the sale_rental module via a post_install script or something like that....

rogerv95:
Thanks @alexis-via  and @dvdhinesh for the video! I can't understand how I missed that at the end of the readme file. 
Now is working completeley fine and I'm able to create orders but something weird happens. In your video I see that if you order 1 macs for 3 days and the delivery order shows 1 mac to deliver. Well, I've tried in my server with my normal install and a clean install and when I add more than 1 day or set the rental quantity to 2 in the delivery order instead of creating an order to deliver 2 macs, creates an order to deliver 6 macs. (2macsx3days) so the delivery order shows the quanity of days of rental and not the units I have to deliver. 
[...]

PR #484, opened 2017-05-24
[10.0] Auto configure sale_rental module via post_install.py script

alexis-via:
Idea taken from the user experience described in bug #472 (before the rename of the bug report)

### DIFF
--- a/sale_rental/README.rst
+++ b/sale_rental/README.rst
@@ -32,6 +32,9 @@ order lines. For that, you must add your user to one of these groups:
 * Manage Product Packaging
 * Properties on lines
 
+Upon module installation, all users are automatically added to the group
+*Properties on lines*.
+
 Usage
 =====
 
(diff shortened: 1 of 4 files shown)


atchuthan:
I agree :+1: but this won't work for newly created users(might be forgotten by users to setup for new odoo users). 
We could add the group(Properties on lines) inherited to employee base group such that the group will be enabled for all users (Existing + New)
my 2 cents

alexis-via:
@atchuthan I will *also* work for newly created users, because it adds the group on the user "default user" (inactive). My tests confirms this.

rafaelbn:
Hi @alexis-via , how is the state of this PR?

alexis-via:
@rafaelbn For me, this PR is ready to be merged.

rogerv95 (approved):

atchuthan (approved):

PR #485, opened 2017-05-30
[10.0][ADD] Add sale_project_fixed_price_task_completed_invoicing

leemannd:
## Main Goal
The main goal of this module is to add the possibility to link a sale.order.line
to a project.task considering the delivery.
Unless sale_timesheet, the quantity shipped won't be linked to a Timesheet nor
to the time spent on the task. The price is fixed on the sale.order.line and it
will be considered as shipped once the task is accomplished.`sale.order.line`. The `sale.order.line` is going to be delivered once the task is accomplished.
## Nota Bene
1. The main links between sale.order, sale.order.line, project and tasks is done in sale_timesheet. In order to ensure compatibility in the future and between them, it has been choosen to inherit from it.
2. There is an auto_install issue with `project_timesheet_analytic_partner` from oca project when I launched tests in local.

### DIFF
--- a/sale_project_fixed_price_task_completed_invoicing/README.md
+++ b/sale_project_fixed_price_task_completed_invoicing/README.md
@@ -1,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Sale Project Fixed Price Task Completed Invoicing
+=================================================
+
+The main goal of this module is to add the possibility to link a sale.order.line
+to a project.task considering the delivery.
+Unless sale_timesheet, the quantity shipped won't be linked to a Timesheet nor
+to the time spent on the task. The price is fixed on the sale.order.line and it
+will be considered as shipped once the task is accomplished.
+
+Usage
+=====
+
+Create a product with product.type 'Service' and track_service 'Completed Task'.
+
+Use it in a Sale Order. Once you validate the Sale Order, it will create a linked
+project and task.
+
+Once the task is finished, on the form view of the corrisponding task click on the
+button 'Invoiceable'. The linked sale.order.line will be considered as shipped.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/sale-workflow/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Denis Leemann <denis.leemann@camptocamp.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/sale_project_fixed_price_task_completed_invoicing/README.md
+++ b/sale_project_fixed_price_task_completed_invoicing/README.md
@@ -7,9 +7,9 @@ Sale Project Fixed Price Task Completed Invoicing
 
 The main goal of this module is to add the possibility to link a sale.order.line
 to a project.task considering the delivery.
-Unless sale_timesheet, the quantity shipped won't be linked to a Timesheet nor
-to the time spent on the task. The price is fixed on the sale.order.line and it
-will be considered as shipped once the task is accomplished.
+The difference with sale_timesheet is that the quantity shipped won't be linked
+to a Timesheet nor to the time spent on the task. The price is fixed on the
+sale.order.line and it will be considered as shipped once the task is accomplished.
 
 Usage
 =====
@@ -19,7 +19,7 @@ Create a product with product.type 'Service' and track_service 'Completed Task'.
 Use it in a Sale Order. Once you validate the Sale Order, it will create a linked
 project and task.
 
-Once the task is finished, on the form view of the corrisponding task click on the
+Once the task is finished, on the form view of the corresponding task click on the
 button 'Invoiceable'. The linked sale.order.line will be considered as shipped.
 
 
--- a/sale_project_fixed_price_task_completed_invoicing/models/product.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/product.py
@@ -20,6 +20,6 @@ class ProductProduct(models.Model):
     def _need_procurement(self):
         for product in self:
             if (product.type == 'service' and
-               product.track_service == 'completed_task'):
+                    product.track_service == 'completed_task'):
                 return True
         return super(ProductProduct, self)._need_procurement()
--- a/sale_project_fixed_price_task_completed_invoicing/models/product.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/product.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    track_service = fields.Selection(selection_add=[
+        ('completed_task', 'Completed Task')]
+    )
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def _need_procurement(self):
+        for product in self:
+            if (product.type == 'service' and
+               product.track_service == 'completed_task'):
+                return True
+        return super(ProductProduct, self)._need_procurement()
--- a/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
@@ -21,7 +21,9 @@ class ProjectTask(models.Model):
         for task in self:
             # We dont' want to modify when the related SOLine is invoiced
             if (not task.sale_line_id or
-                    task.sale_line_id.state in ('done', 'cancel')):
+                    task.sale_line_id.state in ('done', 'cancel') or
+                    task.sale_line_id.invoice_status in ('to invoice',
+                                                         'invoiced')):
                 raise UserError(_("You cannot modify the status if there is "
                                   "no Sale Order Line or if it has been "
                                   "invoiced."))
@@ -38,6 +40,7 @@ class ProjectTask(models.Model):
                                       )
         return super(ProjectTask, self).write(vals)
 
+    @api.model
     def create(self, vals):
         SOLine = self.env['sale.order.line']
         so_line = SOLine.browse(vals.get('sale_line_id'))
--- a/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
@@ -21,7 +21,7 @@ class ProjectTask(models.Model):
         for task in self:
             # We dont' want to modify when the related SOLine is invoiced
             if (not task.sale_line_id or
-               task.sale_line_id.state in ('done', 'cancel')):
+                    task.sale_line_id.state in ('done', 'cancel')):
                 raise UserError(_("You cannot modify the status if there is "
                                   "no Sale Order Line or if it has been "
                                   "invoiced."))
@@ -32,7 +32,7 @@ class ProjectTask(models.Model):
     def write(self, vals):
         for task in self:
             if (vals.get('sale_line_id') and
-               task.sale_line_id.state in ('done', 'cancel')):
+                    task.sale_line_id.state in ('done', 'cancel')):
                 raise ValidationError(_('You cannot modify the Sale Order '
                                         'Line of the task once it is invoiced')
                                       )
--- a/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
@@ -22,8 +22,7 @@ class ProjectTask(models.Model):
             # We dont' want to modify when the related SOLine is invoiced
             if (not task.sale_line_id or
                     task.sale_line_id.state in ('done', 'cancel') or
-                    task.sale_line_id.invoice_status in ('to invoice',
-                                                         'invoiced')):
+                    task.sale_line_id.invoice_status in ('invoiced',)):
                 raise UserError(_("You cannot modify the status if there is "
                                   "no Sale Order Line or if it has been "
                                   "invoiced."))
@@ -49,10 +48,3 @@ class ProjectTask(models.Model):
             raise ValidationError(_('You cannot add a task to and invoiced '
                                     'Sale Order Line'))
         return super(ProjectTask, self).create(vals)
-
-    @api.onchange('invoiceable')
-    def _onchange_invoiceable(self):
-        for task in self:
-            if not task.invoiceable:
-                continue
-            task.sale_line_id._check_delivered_qty()
--- a/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
+++ b/sale_project_fixed_price_task_completed_invoicing/models/project_task.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError, UserError
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    invoiceable = fields.Boolean(
+        string='Invoiceable',
+    )
+    fixed_price = fields.Boolean(
+        string='Fixed Price',
+    )
+
+    @api.multi
+    def toggle_invoiceable(self):
+        for task in self:
+            # We dont' want to modify when the related SOLine is invoiced
+            if (not task.sale_line_id or
+               task.sale_line_id.state in ('done', 'cancel')):
+                raise UserError(_("You cannot modify the status if there is "
+                                  "no Sale Order Line or if it has been "
+                                  "invoiced."))
+            task.invoiceable = not task.invoiceable
+            task.sale_line_id._check_delivered_qty()
+
+    @api.multi
+    def write(self, vals):
+        for task in self:
+            if (vals.get('sale_line_id') and
+               task.sale_line_id.state in ('done', 'cancel')):
+                raise ValidationError(_('You cannot modify the Sale Order '
+                                        'Line of the task once it is invoiced')
+                                      )
+        return super(ProjectTask, self).write(vals)
+
+    def create(self, vals):
+        SOLine = self.env['sale.order.line']
+        so_line = SOLine.browse(vals.get('sale_line_id'))
+        # We don't want to add a project.task to an already invoiced line
+        if so_line and so_line.state in ('done', 'cancel'):
+            raise ValidationError(_('You cannot add a task to and invoiced '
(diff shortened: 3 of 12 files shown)


sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    fixed_price = fields.Boolean(
+        string='Fixed Price',
+    )
+
+    @api.multi
+    def toggle_invoiceable(self):
+        for task in self:
+            # We dont' want to modify when the related SOLine is invoiced
+            if not task.sale_line_id or task.sale_line_id.state in ('done'):
```
gurneyalex: `in ('done', ) `

missing training comma. Probably 'cancel' should be handled in the same way.

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,43 @@
...
+                continue
```
gurneyalex: raise a UserError to give the user some feedback.

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,43 @@
...
+               task.sale_line_id.state in ('done')):
+                raise ValidationError(_('You cannot modify the Sale Order '
+                                        'Line of the task once it is invoiced')
+                                      )
+
+    def create(self, vals):
+        SOLine = self.env['sale.order.line']
+        so_line = SOLine.browse(vals.get('sale_line_id'))
+        # We don't want to add a project.task to an already invoiced line
+        if so_line and so_line.state in ('done'):
```
gurneyalex: ('done', 'cancel')

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,43 @@
...
+            # We dont' want to modify when the related SOLine is invoiced
+            if not task.sale_line_id or task.sale_line_id.state in ('done'):
+                continue
+            task.invoiceable = not task.invoiceable
+
+    @api.multi
+    def write(self, vals):
+        for task in self:
+            if (vals.get('sale_line_id') and
+               task.sale_line_id.state in ('done')):
```
gurneyalex: ('done', 'cancel')

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = 'sale.order.line'
+
+    @api.constrains('product_id')
+    def _onchange_product_id(self):
+        for line in self:
+            if ('track_service' == line.product_id.track_service and
+               line.product_uom_qty != 1.0):
+                raise ValidationError('Error! You cannot have more than one '
+                                      '"track_service" sold in one Sale Order '
+                                      'Line')
```
gurneyalex: error message must be translatable

_('The quantity for 'Complete Task' products must be exactly one')

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo.exceptions import ValidationError
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.constrains('product_id')
+    def _onchange_product_id(self):
+        for line in self:
+            if ('track_service' == line.product_id.track_service and
```
gurneyalex: 'complete_task' ?

gurneyalex (changes requested):

sale_project_fixed_price_task_completed_invoicing/models/product.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def _need_procurement(self):
+        for product in self:
+            if (product.type == 'service' and
+               product.track_service == 'completed_task'):
```
simahawk: fix indentation

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,55 @@
...
+    fixed_price = fields.Boolean(
+        string='Fixed Price',
+    )
+
+    @api.multi
+    def toggle_invoiceable(self):
+        for task in self:
+            # We dont' want to modify when the related SOLine is invoiced
+            if (not task.sale_line_id or
+               task.sale_line_id.state in ('done', 'cancel')):
```
simahawk: fix indentation

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,55 @@
...
+                                  "no Sale Order Line or if it has been "
+                                  "invoiced."))
+            task.invoiceable = not task.invoiceable
+            task.sale_line_id._check_delivered_qty()
+
+    @api.multi
+    def write(self, vals):
+        for task in self:
+            if (vals.get('sale_line_id') and
+               task.sale_line_id.state in ('done', 'cancel')):
```
simahawk: again

simahawk (changes requested):
Overall LGTM, but: some indentation error and above all, any tests?

simahawk:
BTW I'd say that "sale_project_fixed_price" would be enough for module name :)

leemannd:
@pedrobaeza Hello, this is a base new module. The base code is set and the next days I'm building tests. Is is possible to have a first review/opinion about it?

pedrobaeza:
Well, not these days due to the Spanish Odoo days and the big queue I have after the code sprint, and it's not a current interest, but I suggest meanwhile to fix bots.

leemannd:
In order to fix tests in local I had to change a module in project https://github.com/OCA/project/pull/286
error seems related to https://github.com/tinyerp/erppeek/issues/95 and https://github.com/odoo/odoo/pull/17230

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
```diff
@@ -0,0 +1,58 @@
...
+            raise ValidationError(_('You cannot add a task to and invoiced '
+                                    'Sale Order Line'))
+        return super(ProjectTask, self).create(vals)
+
+    @api.onchange('invoiceable')
+    def _onchange_invoiceable(self):
+        for task in self:
+            if not task.invoiceable:
+                continue
+            task.sale_line_id._check_delivered_qty()
```
guewen: Why is this onchange needed? Onchanges are not supposed to have server-side side effects, which you seems to be using here. I think what you want is handle it in create/write.

sale_project_fixed_price_task_completed_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def action_confirm(self):
+        res = super(SaleOrder, self).action_confirm()
+        for order in self:
+            if not order.project_project_id:
+                for line in order.order_line:
```
guewen: Could you extract the content of this loop in a method? It would allow extensions.

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,36 @@
...
+        line = super(SaleOrderLine, self).create(vals)
+        if (line.state == 'sale' and not line.order_id.project_id and
+                line.product_id.track_service in ('completed_task', )):
+            line.order_id._create_analytic_account()
+        return line
+
+    @api.constrains('product_id')
+    def _onchange_product_id(self):
+        for line in self:
+            if ('completed_task' == line.product_id.track_service and
```
guewen: Here was Yoda (https://en.wikipedia.org/wiki/Yoda_conditions)

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,36 @@
...
+                raise ValidationError(_("The quantity for 'Complete Task' "
+                                        "products must be exactly one"))
+
+    @api.multi
+    def _check_delivered_qty(self):
+        for line in self:
+            tasks_count = self.env['project.task'].search_count(
+                [('sale_line_id', '=', line.id)])
+            task_invoiced_count = self.env['project.task'].search_count(
+                [('sale_line_id', '=', line.id), ('invoiceable', '=', True)])
```
guewen: You issue 2 SQL requests for records that intersect. I suggest to only execute one `search()` and do the counting with Python.

guewen (changes requested):

sale_project_fixed_price_task_completed_invoicing/README.md
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Sale Project Fixed Price Task Completed Invoicing
+=================================================
+
+The main goal of this module is to add the possibility to link a sale.order.line
+to a project.task considering the delivery.
+Unless sale_timesheet, the quantity shipped won't be linked to a Timesheet nor
```
gurneyalex: "The difference with sale_timesheet is that the quantity..."

sale_project_fixed_price_task_completed_invoicing/models/project_task.py
leemannd: @guewen Method used in the beginning of the development. Deprecated. I'm removing it.

sale_project_fixed_price_task_completed_invoicing/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Usage
+=====
+
+Create a product with product.type 'Service' and track_service 'Completed Task'.
+
+Use it in a Sale Order. Once you validate the Sale Order, it will create a linked
+project and task.
+
+Once the task is finished, on the form view of the corrisponding task click on the
```
rafaelbn: s/corrisponding/corresponding

sale_project_fixed_price_task_completed_invoicing/README.md
```diff
@@ -0,0 +1,56 @@
...
+button 'Invoiceable'. The linked sale.order.line will be considered as shipped.
+
```
rafaelbn: > Once the task is finished

How do you plan to set the task finished? Or it will be just the button "Invoiceable" ?

sale_project_fixed_price_task_completed_invoicing/README.md
```diff
@@ -0,0 +1,56 @@
...
+The main goal of this module is to add the possibility to link a sale.order.line
+to a project.task considering the delivery.
+Unless sale_timesheet, the quantity shipped won't be linked to a Timesheet nor
+to the time spent on the task. The price is fixed on the sale.order.line and it
+will be considered as shipped once the task is accomplished.
+
+Usage
+=====
+
+Create a product with product.type 'Service' and track_service 'Completed Task'.
```
rafaelbn, line 17: Do you think a new option is needed? Why don't you reuse `Create a task and track hours` ?

rafaelbn (commented):
It looks nice, I will test! Thanks!

rafaelbn:
About https://github.com/OCA/sale-workflow/pull/485#issuecomment-305188321 name @leemannd @simahawk this development has a very nice general approach and is not limited for `fixed price`. This option give a change for the "Project user" to have a button to indicate to sales user that the task is done so you can invoice it, but you can invoice fixed or time spent. I don't think we must limit this in the name 😄 

IMHO name should be `sale_project_task_completed` what do you think?

rafaelbn:
I have tried module in runbot and I cannot feel the improvement with the standard.

In Odoo as is:

- You set a product with invoicing policy: "Ordered quantities" + Track service "create a task and track hours"
- You create a sale order with this product, confirm it -> this creates a task
- After in the task you put times in time-sheet and you have the sale order line with that time "delivered" so you can invoice. Furthermore as you set product with ordered quantities Odoo will invoice just "Fixed price"

Maybe I'm missing something, but I don't see this module very useful. Odoo makes the same.




In any case:

- After you press "invoiceabe" in the task you cannot change it even the sale order line is not invoiced. this is a mistake as you can press the button by error and there is not possibility to fix it.

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
leemannd: Didn't knew the name. Now I understand it can be useless.

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
leemannd: :+1:

sale_project_fixed_price_task_completed_invoicing/README.md
leemannd: In this case I don't want to track hours nor to change the default behaviour that comes from sale_timesheet.

sale_project_fixed_price_task_completed_invoicing/README.md
leemannd: The task is finished when you click on the button "invoiceable".

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,25 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.model
+    def create(self, vals):
```
angelmoya, line 12: What's append if you create sale.order.line and then update product? and updated product has diferent track_service

gurneyalex: @angelmoya it works fine. This bit is only when you create a line in state 'sale' and not draft, although I'm not quite sure when this can happen. @leemannd  can you enlighten us?

gurneyalex:
@rafaelbn Unless I'm mistaken, with the setting you propose, all the lines will be flagged as invoiceable in that setting, which means that when the sales try to create an invoice, all the lines with products invoiced based on sold quantity are there, regardless of the "delivered quantity" (i.e. the hours set on the timesheet). 

Another difference is that with this the quantity sold and the hours on the timesheets are disconnected. Say you are selling a "milestone" with UOM = Unit and qty = 1 for a fixed price. It is annoying to have the qty delivered set to 100 (implicitely 'hours') on the sale order line. 

I agree with you about the unsetting of 'invoiceable' when the related line has not yet been invoiced. @leemannd  can you change this?

gurneyalex:
@rafaelbn @angelmoya thanks for taking the time to review the PR, by the way.

sale_project_fixed_price_task_completed_invoicing/models/sale_order_line.py
leemannd: @angelmoya It takes the logic from `sale_timesheet` and overrides it in order to make it work with the new selection in `line.product_id.track_service` You'll find there the commit from odoo implementing it.  https://github.com/odoo/odoo/commit/90130744e04b4a15973a484a78d461208920e510

```
Before this revision, a small function in sale was extended by
sale_timesheet to know when to (or not to) create an analytic
account for a SO.

This was nice, but since it was extended only in this module,
there's no real need for the added complexity (*reading
complexity*)
```

rafaelbn:
Fist of all, I would like to help because I'm sales manager in Tecnativa and I have same headache as the one who need this solution. "Invoice sales order when the task are flagged to be invoiced independently the quantity". As I'm also project manager, I really would like to get for OCA a solution with a general approach! 🙏 🤞 

> @rafaelbn Unless I'm mistaken, with the setting you propose, all the lines will be flagged as invoiceable in that setting,

Hi @gurneyalex , you are mistaken 😄 . Each sale order line is related with a task so only will be flagged as invoiceable the lines of the task that the "task user" has flagged. If you have 4 sale order lines, then you have 4 tasks and maybe you only set 1 of 4 to invoiceable.

- One objetive set: The possibility of flag in the tasks when they become invoiceable and transfer this flag to the sale order line related. This objective is independent of "Invoicing policy".

> which means that when the sales try to create an invoice, all the lines with products invoiced based on sold quantity are there, regardless of the "delivered quantity" (i.e. the hours set on the timesheet).

- The objetive is that when the sales try to create an invoice, only the sale order lines which are flagged as invoiceable will be in the invoice
- How many? Quantity? Just ordered!! (not the hours set on the timesheets) or just delivered!! here depends on "Invoice policy".

I really wish that you are getting my idea in this point to simplify. Trust me I have all cases 😭 

> Another difference is that with this the quantity sold and the hours on the timesheets are disconnected. Say you are selling a "milestone" with UOM = Unit and qty = 1 for a fixed price. It is annoying to have the qty delivered set to 100 (implicitly 'hours') on the sale order line.

YES! You can send a quotation of 20 hours or just 1 unit of Training (maybe 20, maybe 30 hours) This case in my side is (but of course could be improved in Odoo/OCA):

- A product with ordered qty policy. 1 unit with cost 3000€ for ex. When I confirm sale order Odoo creates a task of 1. Then you work and spend 100 hours but when you go to invoice Odoo just invoice for you 1 unit of 3000€ and not 100 hours. Here will help a lot ! this module because the user which set this task as invoiveable indicate to me (sales man) that I can invoice. And I don't mind how many hours! just what it's in quotation. Is very simple!

> I agree with you about the unsetting of 'invoiceable' when the related line has not yet been invoiced. @leemannd can you change this?

I open to make a hangout or similar with you to explain it if you consider it 😃 

I just would like to cover this need as simple as possible

Thanks!

leemannd:
@rafaelbn @gurneyalex You can now unset 'invoiceable' on tasks when the sale.order.line is not shipped.
@rafaelbn Thanks for the review and the propositions.

guewen (approved):
(only code review)

sergio-teruel:
Can you look at this approach?
#500 
Would there be any use cases that are not covered?

leemannd:
@sergio-teruel I looked at it and seems pretty nice work. As I don't have the full specs from @gurneyalex I must wait for his opinion about it.

gurneyalex (approved):

gurneyalex:
@sergio-teruel I gave your module a look. I see two issues there:

1. the definition of 'finished' : you module seems to use "in a stage with folded = True" which is debatable (could have cancelled tasks in there for instance)
2. sometimes you want to flag as invoiceable something which is not 'finished', which is why I used a different concept when I wrote the specification of the module in #485 : it lets the project manager flag the task as invoiceable, and then the invoicing / sales people will generate an invoice with the correct amount. 

A typical example could be : you have a milestone for a project which was delivered, in a QA instance, and your customer is taking a lot of time to validate the milestone. In this case you could have the task not done (it is not in production), but you still want to invoice the work.

didierdonze (approved):
Thanks for the great work. From a business point of view, I approve the functionality.

sergio-teruel:
@gurneyalex Ok, I understand and I have made changes, can you take a look? Thanks.
This is the commit: https://github.com/OCA/sale-workflow/pull/500/commits/0318a02186e96525e437673bad5138af10fc1f18

simahawk (approved):

leemannd:
@gurneyalex @pedrobaeza Hello, as reviews are done, is it possible to have this PR merged? Thank you.

carlosdauden:
Will we finally have two modules to cover the same need?
https://github.com/OCA/sale-workflow/pull/500

gurneyalex:
I discussed this with @rafaelbn  and @sergio-teruel => #500 suits my needs and covers an additional use case I don't have but they do => I'll close this PR in favor of #500 once I've updated the customers who need it.

gurneyalex:
I discussed this with @rafaelbn  and @sergio-teruel => #500 suits my needs and covers an additional use case I don't have but they do => I'll close this PR in favor of #500 once I've updated the customers who need it.

pedrobaeza:
Can we finally close this?

leemannd:
@pedrobaeza Yes, I'm closing it

PR #488, opened 2017-06-01
FIX sale_order_type in multi company env

eLBati:
In a multi company env, when partners are shared between companies, default order type configured on partner should be variable according to the current company,
just like receivalbe and payable accounts.

### DIFF
--- a/sale_order_type/__openerp__.py
+++ b/sale_order_type/__openerp__.py
@@ -8,7 +8,7 @@
 
 {
     "name": "Sale Order Types",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Good idea, but please check CIs status

eLBati:
@pedrobaeza build restarted and green

pedrobaeza (approved):

AaronHForgeFlow (approved):
Good point! :+1:

PR #493, opened 2017-06-08
[10.0][IMP] sale_automatic_workflow support multicompany

grindtildeath:


rafaelbn:
Please @grindtildeath check travis is red, thansk!

grindtildeath:
@rafaelbn My bad, just fixed it.

jcoux (approved):
LGTM :+1: 
Works on my project.

pedrobaeza:
I have doubts about this method with concurrency, as you can perform several processes at the same time (and more if they last a lot), so changing user company may provoke the other process to fail if it belongs to other company.

We need a better way to fix the multi-company problem. Can this be performed using `company_id` and `force_company` context keys?

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -38,7 +38,10 @@ def _validate_sale_orders(self, order_filter):
         _logger.debug('Sale Orders to validate: %s', sales.ids)
         for sale in sales:
             with savepoint(self.env.cr):
+                user_company = self.env.user.company_id
```
lmignon: @grindtildeath IMO you should define a method to use as context manager...
```python
from contextlib import contextmanager

@contexmanager
def force_company(env, company_id)
    user_company = self.env.user.company_id
    self.env.user.company_id = company_id
    try:  
        yield
    finally:
        self.env.user.company_id = user_company
```
Once you have defined the method it can used as follow...
```python
with savepoint(self.env.cr), force_company(sale.company_id):
    self.action_confirm()
```

With this new method the logic to change the current company is factored into a unique method

lmignon (changes requested):

sale_automatic_workflow/models/automatic_workflow_job.py
pedrobaeza: @lmignon but the problem I pointed out still exists with this? Isn't there a module your company made for having a user to make all these things?

lmignon:
@pedrobaeza each process is transnational and the change is reverted after the call to a target method. Therefore the only problem I can imagine is a concurrent update exception since the company_id on the user could be modified by 2 processes at the same time..

yvaucher:
@pedrobaeza Isn't the change of company only in the transaction?
So you won't write on other company. But we might have an issue if multiple transactions tries to do an update on user.company_id. This means it would be an issue using the same user.

yvaucher:
@lmignon transactional ? (you were faster than me nevertheless :) )

sale_automatic_workflow/models/automatic_workflow_job.py
yvaucher: What we would need is to avoid the 2 updates on user.company_id.

Could we have the user changed only in cache like we do in onchange context manager ?

lmignon:
@yvaucher each process is transactional. (typo)...

pedrobaeza:
OK, you are right. I haven't thought enough about that. Thanks for the clarification. 

About the concurrent update, we have this problem all the days, but well, this a bit bigger as we make a lot of changes. This can be a problem with e-commerce connectors where automatic workflows are launched simultaneous on parallel jobs. Maybe any semaphore strategy on the method you propose, Laurent?

sale_automatic_workflow/models/automatic_workflow_job.py
pedrobaeza: I don't think so, as most of the problems will be with security rules.

yvaucher: Could we try:
```
user = self.env.user
user.update({'company_id': company_id})
```

lmignon:
@grindtildeath Thank you for the fix. The code LGTM. Nevertheless can you add a unittest to avoid regression in the future PLZ?

grindtildeath:
It's quite weird because it seems my unittest would work without the modification on the automatic.workflow.job :/
I suspect we had a property issue on the project from @jcoux but we might have to wait until he is back from holidays to figure it out.

jcoux:
I confirm without this PR, the sale automatic workflow failed on my project.

The sale order is created on company A.
The administrator is connected on company B.

Without this PR, the invoice created have a journal on company B instead of a journal on company A.

@grindtildeath Please increase the unit test to check this case.

grindtildeath:
@lmignon @pedrobaeza 
I finally had some time to dig on this issue after @jcoux could check what happened on his project.
As he stated, the issue is on finding the right account.journal to create the invoice, and it's fixed by the use of the context manager from this PR.

Now I'm not sure it's the good way to fix it because here is what's happening in odoo code : 

First is a default_get on account.journal without taking care of the sale.order company_id :
https://github.com/odoo/odoo/blob/10.0/addons/sale/models/sale.py#L285

Then, as the company_id isn't passed through context, the default is self.env.user.company_id
https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L67

Now, should we use this actual version of my PR, switching company before each method is called in automatic.workflow.job or maybe we should just add a company_id context here : https://github.com/OCA/sale-workflow/blob/10.0/sale_automatic_workflow/models/automatic_workflow_job.py#L52 
?

pedrobaeza:
@grindtildeath thanks for your clarifications. Indeed, it's cleaner to pass the context `company_id` for this specific problem, but it's safer to do the current way for avoiding other possible side effects, so for me let's go this way then.

grindtildeath:
@lmignon Can you update your review please ?

atchuthan (approved):
code review :+1:

guewen:
FTR, in 12.0, the ORM writes on `res_users.company_id`, so it currently restricts to one action at a time.

PR #494, opened 2017-06-13
[FIX] repo error by adding product_uom in tests of sale_cancel_reason

kittiu:
Add product_uom when create order_line to make sure that it pass test


mourad-ehm:
:+1:

bealdav:
:+1:

pedrobaeza:
Travis is still red.

mourad-ehm:
I checked code and found that product_id and product_uom_id are required in 10.0. So I think that the solution is to make the same change for all modules that create sale lien without product_id or product_uom_id in the same PR.

mourad-ehm:
I add this PR to fix this bug https://github.com/OCA/OCB/pull/625
and for odoo https://github.com/odoo/odoo/pull/17613

kittiu:
This issue is just sample for https://github.com/OCA/sale-workflow/pull/429
As that is fixed, this can be closed.

PR #495, opened 2017-06-15
[ADD] sale_procurement_group_by_requested_date v9

serpentcs-dev1:
Sale Procurement Group by Requested date
========================================

- This module creates different procurements groups for different requested
dates in a sale order line when the sale order is confirmed.
- It depends on sale_sourced_by_line so this module will group procurements also
by the warehouse in the sale order line.

### DIFF
--- a/sale_procurement_group_by_requested_date/README.rst
+++ b/sale_procurement_group_by_requested_date/README.rst
@@ -1,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :alt: License: AGPL-3
+
+========================================
+Sale Procurement Group by Requested date
+========================================
+
+This module creates different procurements groups for different requested
+dates in a sale order line when the sale order is confirmed.
+It depends on sale_sourced_by_line so this module will group procurements 
+also by the warehouse in the sale order line.
+
+Installation
+============
+
+This module depends on the modules sale_procurement_group_by_line and
+sale_sourced_by_line.
+
+
+Usage
+=====
+
+#. Add a requested date for a sale order line.
+#. Confirm the sale order
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/9.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/sale-workflow/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+
+Credits
+=======
+
+Contributors
+------------
+
+* Aaron Henriquez <ahenriquez@eficent.com>
+* Jordi Ballester <jordi.ballester@eficent.com>
+* Darshan Patel <darshan.patel.serpentcs@gmail.com>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/sale_procurement_group_by_requested_date/__openerp__.py
+++ b/sale_procurement_group_by_requested_date/__openerp__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Group procurements by requested date",
+    "summary": "Groups pickings based on requested date of order line",
+    "version": "9.0.1.0.0",
+    "category": "Sales Management",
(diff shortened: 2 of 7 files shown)


serpentcs-dev1:
@aheficent Could you please review?

AaronHForgeFlow (commented):
Functional test + code are good to me. Travis is failing because of test_sale_exception in sale_exception. @darshan-serpent Can you check?

serpentcs-dev1:
@aheficent Sure will take a look at the reason of travis annoyance

serpentcs-dev1:
Resolved travis and squashed the commits!

AaronHForgeFlow (approved):
:+1: Functional review

sale_procurement_group_by_requested_date/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+This module depends on the modules sale_procurement_group_by_line and
+sale_sourced_by_line.
+
+
+Usage
+=====
+
+* Add a requested date for a sale order line.
+* Confirm the sale order
```
LoisRForgeFlow: I tend to prefer numbered lists here.

sale_procurement_group_by_requested_date/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Group procurements by requested date",
+    "summary": "Groups pickings based on requested date of order line",
+    "version": "9.0.0.0.1",
+    "category": "Sales Management",
+    "website": "http://www.eficent.com",
+    "author": "Eficent Business and IT Consulting Services S.L., "
```
LoisRForgeFlow: Just `Eficent` would be enough

sale_procurement_group_by_requested_date/models/sale_order.py
```diff
@@ -0,0 +1,54 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    def _prepare_order_line_procurement(self, cr, uid, order, line,
+                                        group_id=False, context=None):
```
LoisRForgeFlow: Adapt to new api please

sale_procurement_group_by_requested_date/tests/test_sale_procurement_group_by_requested_date.py
```diff
@@ -0,0 +1,110 @@
...
+class TestSaleMultiPickingByRequestedDate(TransactionCase):
+    """Check the _get_shipped method of Sale Order. """
+
+    def setUp(self):
+        """Setup a Sale Order with 4 lines.
+        And prepare procurements
+        """
+        super(TestSaleMultiPickingByRequestedDate, self).setUp()
+        SO = self.env['sale.order']
+        SOL = self.env['sale.order.line']
```
LoisRForgeFlow: these two variables should be in lowercase.

LoisRForgeFlow (changes requested):
Code review - some comments inline

sale_procurement_group_by_requested_date/__openerp__.py
serpentcs-dev1: Isn't it better to have the full name?! 

As it is not even going beyond the limited characters for the author key

LoisRForgeFlow: For the headers it's fine but this is shown in the form view of apps in odoo, and due to large length of our company name sometimes OCA is not shown properly and we really want to show OCA as author in the module description.

LoisRForgeFlow: with OCA I mean "Odoo Community Association (OCA)" which is always the last author in the list.

sale_procurement_group_by_requested_date/models/sale_order.py
```diff
@@ -9,12 +9,13 @@
...
-        values = super(SaleOrder, self)._prepare_order_line_procurement(
-            cr, uid, order, line, group_id=group_id, context=context)
-        if line.requested_date:
-            req_datetime = fields.Datetime.from_string(line.requested_date)
+    @api.multi
+    def _prepare_order_line_procurement(self, group_id=False):
+        values = super(SaleOrder, self).\
+            _prepare_order_line_procurement(group_id=group_id)
+        self.ensure_one()
+        if self.requested_date:
```
LoisRForgeFlow: previously it was `line.requested_date:` I think that was refering to `sale.order.line` while `self` here is `sale.order`. Also, are you sure about removing `order` and `line` from the imputs? they might be needed in the super call.

serpentcs-dev1: The method earlier was as per the older version

Now, self is referring to the sale order lines and the parameters used are as per [newer version](https://github.com/odoo/odoo/blob/9.0/addons/sale/sale.py#L565).

AaronHForgeFlow: @darshan-serpent you placed that method in the class SaleOrder, right?

serpentcs-dev1: Yes moved the method to the right place :)

serpentcs-dev1:
All green!! I think now it is good to go, so could you please check it once again??

@aheficent @lreficent

LoisRForgeFlow (approved):
Code review of changes from my last review - LGTM :+1:

AaronHForgeFlow (approved):
It works smoothly :+1:

MiquelRForgeFlow (approved):
LGTM :+1:

PR #496, opened 2017-06-21
[8.0] FIX sale_order_revision, security rule error

kittiu:
If user login as user group "see own lead", whose security is controlled by a RULE.
On copy() with default {'active': False}, will throw error 
`Access Denied (Document type: Sales Order Line, Operation: create)`
I don't know the reason behind, but the fix will work.
Note: Already 10.0 has this similar fix.


pedrobaeza:
Please include a comment about why is done that way in the code itself.

kittiu:
Replace with https://github.com/OCA/sale-workflow/pull/497

PR #497, opened 2017-06-21
[sale_order_revision][8.0][BUG], security rule error

kittiu:
If user login as user group "see own lead", whose security is controlled by a RULE.
On copy() with default {'active': False}, will throw error
Access Denied (Document type: Sales Order Line, Operation: create)
I don't know the reason behind, but the fix will work.
Note: Already 10.0 has this similar fix.

### DIFF
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -85,12 +85,14 @@ class sale_order(models.Model):
             default.update({
                 'name': prev_name,
                 'revision_number': revno,
-                'active': False,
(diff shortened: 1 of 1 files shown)


kittiu:
@pedrobaeza do you think this one is ok to merged, any further suggest?

pedrobaeza:
We need reviews on this one. You can review other PRs and ask in exchange that they review yours.

eLBati (approved):
Changes loooks OK. Even if reasons are not clear, v10 already writes ```active``` field

PR #498, opened 2017-06-22
[10.0][MIG] sale_procurement_group_by_line

MiquelRForgeFlow:
Standard migration.

### DIFF
--- a/sale_procurement_group_by_line/README.rst
+++ b/sale_procurement_group_by_line/README.rst
@@ -1,4 +1,5 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl
    :alt: License: AGPL-3
 
 ==============================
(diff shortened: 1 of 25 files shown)


sale_procurement_group_by_line/README.rst
```diff
@@ -21,44 +22,54 @@ Configuration
 
 No special configuration is required.
 
+
```
AaronHForgeFlow: Why this blank line? Also I'd delete the Installation and Configuration sections, as they are empty

sale_procurement_group_by_line/model/sale.py
```diff
@@ -28,143 +14,33 @@ class SaleOrder(models.Model):
...
-                is_shipped = False
-                break
-            is_shipped &= all([proc.state in ['cancel', 'done']
-                               for proc in group.procurement_ids])
-        self.shipped = is_shipped
+        return self._prepare_procurement_group()
 
     ###
     # OVERRIDE to find sale.order.line's picking
     ###
```
AaronHForgeFlow: I don't understand this. This method is specifically done to get the pickings in the SO, not in the line.

AaronHForgeFlow (changes requested):
Thanks for this migration. As this module does nothing by itself the only functional test it to check no error was introduced in the flow. 

Just a few comments.

sale_procurement_group_by_line/model/sale.py
MiquelRForgeFlow: This comment was added by @yvaucher in v8 and still remains here. He was referring to the `_compute_get_picking_ids` method.

AaronHForgeFlow: I know he refers to that method:
Is it possible to override this method to calculate the pickings related to a SO line and keep the current functionality? I know how to update two different computed fields in a single method for other field types but not for many2many.

If it's possible then keep it but if it's not I think this can be deleted IMHO.

Regards

LoisRForgeFlow (approved):
code review LGTM

sale_procurement_group_by_line/__manifest__.py
```diff
@@ -1,38 +1,23 @@
...
+    'name': 'Sale Procurement Group by Line',
+    'summary': 'Base module for multiple procurement group by Sale order',
+    'version': '10.0.1.0.0',
+    "author": "Camptocamp,"
+              "Eficent,"
+              "Serpent,"
+              "Odoo Community Association (OCA)",
+    'category': 'Warehouse',
+    'license': 'AGPL-3',
+    'complexity': 'expert',
```
berpweb: I am wondering what is the use of this `complexity`?

sale_procurement_group_by_line/__manifest__.py
```diff
@@ -1,38 +1,23 @@
...
+    'website': "https://github.com/OCA/sale-workflow",
+    'depends': ['sale_stock',
+                ],
+    'auto_install': False,
```
berpweb: Do we need this explicitly? It's like redundant to me...

sale_procurement_group_by_line/model/sale.py
```diff
@@ -28,143 +14,33 @@ class SaleOrder(models.Model):
...
-        for sale in self.browse(cr, uid, ids, context=context):
+
+    @api.multi
+    @api.depends('order_line')
+    def _compute_get_picking_ids(self):
+        for sale in self:
             group_ids = set([line.procurement_group_id.id
                              for line in sale.order_line
                              if line.procurement_group_id])
             if not any(group_ids):
```
berpweb: Do we need this `any` here?

sale_procurement_group_by_line/__manifest__.py
elicoidal: This is deprecated

berpweb (approved):
Functional/Code :+1:

berpweb:
@yvaucher Ping... Can you review this?
I guess you have the complete understanding of the regressions it could create...

AaronHForgeFlow (approved):
Code LGTM :+1:

PR #499, opened 2017-06-23
[ADD] sale_order_line_sequence v9 

serpentcs-dev1:
Sale Order Line with Sequence
=============================

Adds sequence to display in Sales Order line.


serpentcs-dev1:
@aheficent Could you please review this?

sale_order_line_sequence/README.rst
```diff
@@ -0,0 +1,57 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=============================
+Sale Order Line with Sequence
+=============================
+
+Propagates Sales Order line sequence to its Invoices and Stock picking.
+
```
AaronHForgeFlow: Can you also explain you can change see and change the sequence in the sale order line? This is a new feature.

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+    # re-defines the field to change the default
+    sequence = fields.Integer(help="Gives the sequence of this line when "
+                                   "displaying the sale order.",
+                              default=9999)
+
+    # displays sequence on the order line
+    sequence2 = fields.Integer(help="Shows the sequence of this line in "
+                               "the sale order.",
+                               related='sequence', readonly=True,
+                               store=True)
```
AaronHForgeFlow, line 61: This field is not showing in the sale order line, in runbot

Edit: It's showing now so consider this solved

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+        return super(SaleOrder, self2).copy(default)
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    # re-defines the field to change the default
+    sequence = fields.Integer(help="Gives the sequence of this line when "
+                                   "displaying the sale order.",
+                              default=9999)
```
AaronHForgeFlow, line 55: I can change the sequence of a confirmed sale order. I think this should not be allowed. And if that were allowed the sequence in the picking and in the invoice should change accordingly.

sale_order_line_sequence/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    "category": "Sales",
+    "website": "https://www.eficent.com/",
+    "license": "AGPL-3",
+    'data': ['views/sale_view.xml'],
+    "depends": [
+        "stock",
+        "sale",
+        "sale_stock",
+        "account_invoice_line_sequence",
+        "stock_picking_line_sequence",
```
AaronHForgeFlow: I don't see any code to transfer the sequence to the invoice and the pickings. This module is not using the dependencies to` account_invoice_line_sequence` and `stock_picking_line_sequence`

AaronHForgeFlow (changes requested):

AaronHForgeFlow:
Also, can you please add the [ADD] to this Pull Request? So everyone know this is a new module in OCA.

Thanks

sale_order_line_sequence/model/sale_order.py
serpentcs-dev1: @aheficent It does in local, I guess that's because the runbot is red.

sale_order_line_sequence/__openerp__.py
serpentcs-dev1: The code to propagate the sequence to invoice and the stock are done in account_invoice_line_sequence and stock_picking_line_sequence respectively.

AaronHForgeFlow: Can you show me?  I just checked and I did not see anything similar. The sequence in those modules does not come from the sales order.

serpentcs-dev1: The sequence to invoice is passed from the base [sale module iteself](https://github.com/odoo/odoo/blob/9.0/addons/sale/sale.py#L723)

And for [stock move here](https://github.com/odoo/odoo/blob/9.0/addons/sale_stock/sale_stock.py#L271).

AaronHForgeFlow: So the dependencies to `account_invoice_line_sequence` and `stock_picking_line_sequence` are not needed, right?

AaronHForgeFlow (approved):
:+1: It works fine now. Please squash two last commits

serpentcs-dev1:
@aheficent I have squashed the commits!

LoisRForgeFlow (approved):
Tested on runbot + code review :+1:

MiquelRForgeFlow (approved):

PR #501, opened 2017-06-28
[ADD] sale_order_line_sequence v10

serpentcs-dev1:
Sale Order Line with Sequence
========================

Displays the sequence of Sale order line and helps to maintain the order.

### DIFF
--- a/sale_order_line_sequence/model/sale_order.py
+++ b/sale_order_line_sequence/model/sale_order.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    @api.depends('order_line')
+    def _compute_max_line_sequence(self):
+        """Allow to know the highest sequence entered in sale order lines.
+        Then we add 1 to this value for the next sequence.
+        This value is given to the context of the o2m field in the view.
+        So when we create new sale order lines, the sequence is automatically
+        added as :  max_sequence + 1
+        """
+        for sale in self:
+            sale.max_line_sequence = (
+                max(sale.mapped('order_line.sequence') or [0]) + 1)
+
+    max_line_sequence = fields.Integer(string='Max sequence in lines',
+                                       compute='_compute_max_line_sequence',
+                                       store=True)
+
+    @api.multi
+    def _reset_sequence(self):
+        for rec in self:
+            current_sequence = 1
+            for line in rec.order_line:
+                line.sequence = current_sequence
+                current_sequence += 1
+
+    @api.multi
+    def write(self, line_values):
+        res = super(SaleOrder, self).write(line_values)
+        self._reset_sequence()
+        return res
+
+    @api.multi
+    def copy(self, default=None):
+        return super(SaleOrder,
+                     self.with_context(keep_line_sequence=True)).copy(default)
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    # re-defines the field to change the default
+    sequence = fields.Integer(help="Gives the sequence of this line when "
+                                   "displaying the sale order.",
+                              default=9999)
+
+    # displays sequence on the order line
+    sequence2 = fields.Integer(help="Shows the sequence of this line in "
+                               "the sale order.",
+                               related='sequence', readonly=True,
+                               store=True)
+
+    @api.model
+    def create(self, values):
+        line = super(SaleOrderLine, self).create(values)
+        # We do not reset the sequence if we are copying a complete sale order
+        if self.env.context.get('keep_line_sequence'):
+            line.order_id._reset_sequence()
+        return line
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,10 @@ addons:
       - python-simplejson
       - python-serial
       - python-yaml
-      - wkhtmltopdf
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0" WKHTMLTOPDF_VERSION="0.12.4"
   - TRANSIFEX_USER='transbot@odoo-community.org'
   - secure: mTXYvzW7/+xP465dNW0p9H+Q8jVgQXmqm8zo+ntZC6ESkTudj6xUkrjJcfzp9LGomxoG6N+6TezjFuhfGKCAc4S2manxj9YQoHvZMHJvt73VsMZU2QLfE+IMAdFS/AeBpYLNE8VvdKIGhiFwvHyRqaF0vAGwmcsOdS2NQb/hoiY=
 
--- a/sale_order_line_sequence/README.rst
+++ b/sale_order_line_sequence/README.rst
@@ -1,0 +1,57 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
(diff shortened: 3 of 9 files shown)


serpentcs-dev1:
@aheficent Could you please review?

AaronHForgeFlow (commented):
It works well, the creation and the modification of the SO and also when duplicating. Only one think. For previous SO the sequence is 10 for all the lines (See picture). I'm thinking about a pre-init hook to force update old sale orders in the database. Do you agree with me?
image

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+        for rec in self:
+            rec._reset_sequence()
+        return res
+
+    @api.multi
+    def copy(self, default=None):
+        if not default:
+            default = {}
+        self2 = self.with_context(keep_line_sequence=True)
+        return super(SaleOrder, self2).copy(default)
```
LoisRForgeFlow: Why not directly `return super(SaleOrder, self).with_context(keep_line_sequence=True).copy(default)`?

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+        if 'keep_line_sequence' not in self.env.context:
+            line.order_id._reset_sequence()
+        return line
+
+    @api.multi
+    def copy(self, default=None):
+        if not default:
+            default = {}
+        if 'keep_line_sequence' not in self.env.context:
+            default['sequence'] = 9999
```
LoisRForgeFlow: is there any case in which this 9999 default would apply?? I cannot figure it out.

sale_order_line_sequence/model/sale_order.py
serpentcs-dev1: better, will change it!

sale_order_line_sequence/model/sale_order.py
serpentcs-dev1: If the whole Sale Order is copied the sequence would be kept as it is, but what would be the sequence if just the order line is duplicated? so its just sets the default sequence.

LoisRForgeFlow: But there is no way to functionally copy just lines, right? Anyway, this does not hurt. Keep it like this!

sale_order_line_sequence/model/sale_order.py
LoisRForgeFlow: To get rid of travis problem you can put it like this:


        return super(SaleOrder, self).with_context(
            keep_line_sequence=True).copy(default)

serpentcs-dev1: Yeah I did it like this, it works! Thanks!

        return super(SaleOrder,
                     self.with_context(keep_line_sequence=True)).copy(default)

LoisRForgeFlow (approved):
:+1:

sale_order_line_sequence/model/sale_order.py
serpentcs-dev1: Hmm.. I think you are right in a way, there is no need for the copy method.

When the line is copied it will trigger the create method which already resets and keeps the sequence.

I think we should remove it, what do you think?

LoisRForgeFlow: Yes, exactly. You can remove it since there are no test to break :wink:

AaronHForgeFlow (approved):
:+1: Thank you @darshan-serpent for attending all the suggestions. Can you please squash the commits before this is merged?

AaronHForgeFlow:
It seems to me that the report is failing on runbot because the proper version of wkhtmltopdf is not installed :disappointed: 
image

serpentcs-dev1:
@aheficent Yes encountered the same thing it is asking for upgrade, what is the solution to this as this could block the test for the report.

AaronHForgeFlow:
Not sure if this can be configured at repo level. @moylop260 can you shed some light here?

moylop260:
You can add a custom wkhtmltopdf version using the environment variable: `WKHTMLTOPDF_VERSION` from [.travis.yml](https://github.com/OCA/sale-workflow/blob/10.0/.travis.yml)

More info about: [travis_install_nightly](https://github.com/OCA/maintainer-quality-tools/blob/f54d6e33c5c48e6bd8051efe018e88ff3db669f8/travis/travis_install_nightly#L85)

AaronHForgeFlow:
I think correct version is 0.12.2.1, isn't it?

serpentcs-dev1:
Is it? I checked other PR but I cannot find any reference where there is WKHTMLTOPDF mentioned in the environment?

Okay I'll try applying 0.12.2.1

serpentcs-dev1:
I don't think this is a proper way!!

I checked other PR, there the other reports are working fine, even though nothing is mentioned in their travis.yml file.

edit : changing to the way it was earlier.

AaronHForgeFlow:
Not sure really how to fix this

moylop260:
@aheficent I just apply the following patch in your branch:

```diff
diff --git a/.travis.yml b/.travis.yml
index 8cdb154aa..a22800348 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,10 @@ addons:
       - python-simplejson
       - python-serial
       - python-yaml
-      - wkhtmltopdf
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0" WKHTMLTOPDF_VERSION="0.12.4"
   - TRANSIFEX_USER='transbot@odoo-community.org'
   - secure: mTXYvzW7/+xP465dNW0p9H+Q8jVgQXmqm8zo+ntZC6ESkTudj6xUkrjJcfzp9LGomxoG6N+6TezjFuhfGKCAc4S2manxj9YQoHvZMHJvt73VsMZU2QLfE+IMAdFS/AeBpYLNE8VvdKIGhiFwvHyRqaF0vAGwmcsOdS2NQb/hoiY=
 
```

If we add `- wkhtmltopdf` in `-addons -packages -apt` from `.travis.yml` file so a `apt-get install wkhtmltopdf` will be executed for ubuntu but default OS ubuntu version is trusty 14.04.
  - And use wkhtmltopdf 0.9.9: https://packages.ubuntu.com/trusty/wkhtmltopdf

If we remove that section of the .travis.yml runbot will work because our custom image have pre-installed a patched version.
But I added a explicit version in order to avoid a difference between travis and runbot for this package.

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+    sequence2 = fields.Integer(help="Shows the sequence of this line in "
+                               "the sale order.",
+                               related='sequence', readonly=True,
+                               store=True)
+
+    @api.model
+    def create(self, values):
+        line = super(SaleOrderLine, self).create(values)
+        # We do not reset the sequence if we are copying a complete sale order
+        if 'keep_line_sequence' not in self.env.context:
```
moylop260: What about use `self.env.context.get('keep_line_sequence')`? (if I use a `False` for this parameter)

sale_order_line_sequence/model/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+    @api.multi
+    def write(self, line_values):
+        res = super(SaleOrder, self).write(line_values)
+        self._reset_sequence()
+        return res
+
+    @api.multi
+    def copy(self, default=None):
+        if not default:
+            default = {}
```
moylop260: You can delete this section since that you are not changing the `default` variable.

moylop260 (approved):
LGTM
(Just small changes)

sale_order_line_sequence/model/sale_order.py
serpentcs-dev1: Right, for further use better to fetch the dict key's value, thanks!

AaronHForgeFlow (changes requested):
The report doesn't look correct. The descriptions are missing in the report but exit in the SO lines

image

serpentcs-dev1:
@aheficent Thanks for the review, this is fixed now!!

screenshot from 2017-07-06 10-56-25

AaronHForgeFlow (approved):
:+1: Report fine now. Please squash the commits before this is merged.

serpentcs-dev1:
@aheficent  Squashed!!

PR #502, opened 2017-07-07
[IMP] Update sale_isolated_quotation

gdgellatly:
- allow to print, email and publish on website_portal
- attempt to at least make website_portal_sale make sense
- override action_confirm behaviour so website_quote cannot confirm quotations
- update backend header to v10 classes as used in view_order_form.

Hi,

I like this module but it faces a lot of limitations when working with website_* modules.  I wouldn't say this update is complete, but it is definitely an improvement.  Still needs to consider things like payments, url redirects on acceptance etc but at least now frontend and backend will behave about the same.

### DIFF
--- a/sale_exception/models/sale.py
+++ b/sale_exception/models/sale.py
@@ -47,10 +47,10 @@ class SaleOrder(models.Model):
 
     @api.multi
     def action_confirm(self):
-        if self.detect_exceptions():
-            return self._popup_exceptions()
-        else:
-            return super(SaleOrder, self).action_confirm()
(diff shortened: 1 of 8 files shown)


gdgellatly:
Might need some help here.  There is a problem in the way the original sets the default for is_order.  I've just reversed its default behaviour for other modules which clearly aren't using a view context to create and confirm orders.  But I wonder if it would be better ditching the default altogether and using default_is_order context key on the view action.

gdgellatly:
I give up

PR #505, opened 2017-07-11
[10.0][FIX] sale_order_margin_percent: computation multi record exception

wtaferner:
Straight forward and easy to review and merge :-)

### DIFF
--- a/sale_order_margin_percent/models/sale_order_margin.py
+++ b/sale_order_margin_percent/models/sale_order_margin.py
@@ -12,5 +12,8 @@ class SaleOrder(models.Model):
 
     @api.depends('margin', 'amount_untaxed')
     def _compute_percent(self):
-        if self.margin and self.amount_untaxed:
-            self.percent = (self.margin / self.amount_untaxed) * 100
+        for order in self:
+            percent = 0
+            if order.margin and order.amount_untaxed:
+                percent = (order.margin / order.amount_untaxed) * 100
+            order.percent = percent
--- a/sale_order_margin_percent/__manifest__.py
+++ b/sale_order_margin_percent/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Sale Order Margin Percent",
     "summary": "Show Percent in sale order",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Sales",
     "website": "http://www.pesol.es",
     "author": "PESOL",


pedrobaeza (approved):

sale_order_margin_percent/models/sale_order_margin.py
```diff
@@ -12,5 +12,6 @@ class SaleOrder(models.Model):
 
     @api.depends('margin', 'amount_untaxed')
     def _compute_percent(self):
-        if self.margin and self.amount_untaxed:
-            self.percent = (self.margin / self.amount_untaxed) * 100
+        for order in self:
+            if order.margin and order.amount_untaxed:
```
espo-tony: IMO, this condition can be rewritten without the first term, as a margin equal to 0 would correctly compute a percent equal to 0.

sale_order_margin_percent/models/sale_order_margin.py
```diff
@@ -12,5 +12,6 @@ class SaleOrder(models.Model):
...
+                order.percent = (order.margin / order.amount_untaxed) * 100
```
espo-tony: I would add an else branch to handle the case with amount_untaxed equal to 0: in this case I think that percent should also be set to 0 (currently, if amount_untaxed was any number and lately drops to 0, the percent would simply not change, thus giving a wrong information).

espo-tony (changes requested):

sale_order_margin_percent/models/sale_order_margin.py
wtaferner: It is not really part of this PR which is an actual bugfix waiting to be merged for quite some time although there is always space to improve. Not sure how it works here.

sale_order_margin_percent/models/sale_order_margin.py
wtaferner: Same as above.

wtaferner:
@tafaRU 
Your review is welcome...as said, it is straight forward and waiting to be fixed and merged ;-)

pedrobaeza:
Indeed it's a bugfix but it's very easy to include the else part. Any way, if not, you should at least change module version number.

wtaferner:
Ok, I rewrote it that it will always provide the correct percentage and increased the version number. I hope this is what is expected and serves the input. Thank you for your reviews.

sale_order_margin_percent/models/sale_order_margin.py
```diff
@@ -12,5 +12,8 @@ class SaleOrder(models.Model):
 
     @api.depends('margin', 'amount_untaxed')
```
wpichler, line 13: adding @api.multi would be no failure.

wpichler:
why not change the percent field to store=True ? I would make it searchable. Eg. Search for all sale orders with margin > 0.1

sale_order_margin_percent/models/sale_order_margin.py
pedrobaeza: On v10, @api.multi is the default

wpichler: correct - but it does not hurt - and makes it clear on a short view

wtaferner: I tend to skip this as it is only needed for public methods to work with RPC in v10 and we should move the same direction as Odoo core and save lines of code for something which is default behavior. We do not add ```store=True``` to every normal field, do we?

wtaferner (commented):
Hmm, I do like the idea to make it searchable by storing the value in the database. Any contradictions @pedrobaeza?

misern2:
👍  to search by margin.

wtaferner:
Later PR and fix was merged (https://github.com/OCA/sale-workflow/pull/596). Therefore I close this PR. Thank you for your contributions and comments.

PR #510, opened 2017-07-22
[FIX] Cancel lines in old revisions

JuaniFreedoo:
[8.0] Cancel lines in old revision orders

### DIFF
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -91,8 +91,7 @@ class sale_order(models.Model):
                 'unrevisioned_name': self.unrevisioned_name,
             })
             prev_order = super(sale_order, self).copy(default=default)
-            for line in prev_order.order_line:
-                line.write({'state': 'cancel'})
+            prev_order.order_line.write({'state': 'cancel'})
         return prev_order
 
     @api.model
--- a/sale_order_revision/model/sale_order.py
+++ b/sale_order_revision/model/sale_order.py
@@ -91,7 +91,8 @@ class sale_order(models.Model):
                 'unrevisioned_name': self.unrevisioned_name,
             })
         prev_order = super(sale_order, self).copy(default=default)
-        prev_order.order_line.write({'state': 'cancel'})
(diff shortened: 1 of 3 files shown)


sale_order_revision/model/sale_order.py
```diff
@@ -90,7 +90,10 @@ def copy(self, default=None):
                 'current_revision_id': self.id,
                 'unrevisioned_name': self.unrevisioned_name,
             })
-        return super(sale_order, self).copy(default=default)
+            prev_order = super(sale_order, self).copy(default=default)
```
pedrobaeza: You always have to call super, not only in the `if` block.

sale_order_revision/model/sale_order.py
```diff
@@ -90,7 +90,10 @@ def copy(self, default=None):
...
+            for line in prev_order.order_line:
+                line.write({'state': 'cancel'})
```
pedrobaeza: Do directly `prev_order.order_line.write({'state': 'cancel'})`

sale_order_revision/model/sale_order.py
pedrobaeza: This is still pending. Please change also module version number.

sale_order_revision/model/sale_order.py
```diff
@@ -90,7 +90,9 @@ def copy(self, default=None):
                 'current_revision_id': self.id,
                 'unrevisioned_name': self.unrevisioned_name,
             })
-        return super(sale_order, self).copy(default=default)
+        prev_order = super(sale_order, self).copy(default=default)
+        prev_order.order_line.write({'state': 'cancel'})
```
pedrobaeza: This should be executed only if self.env.context.get('new_sale_revision'):
```
if self.env.context.get('new_sale_revision'):
    prev_order.order_line.write({'state': 'cancel'})
```

rousseldenis:
@JuaniFreedoo Is this still relevant? This will be closed as no activity

rousseldenis:
Closing as very old. Please feel free to reopen if relevant.

PR #512, opened 2017-07-24
[10.0] [ADD] sale_order_line_date

nikul-serpentcs:
Sale order line date
===============

This module adds requested date to a sales order lines and propagates it to stock moves and procurements.
When the requested date of the whole sale order is modified the requested date of the lines change to match.


nikul-serpentcs:
@aheficent Could you please review.

LoisRForgeFlow:
Please, keep commit history always in your migrations. See: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

nikul-serpentcs:
@lreficent Sure.

AaronHForgeFlow (approved):
It works good. I'll approve this when the commit history is fixed

AaronHForgeFlow (changes requested):
Until the commit history is fixed

sale_order_line_date/models/sale_order.py
```diff
@@ -0,0 +1,59 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    requested_date = fields.Datetime()
+
+    @api.multi
+    def write(self, vals):
+        for line in self:
+            if not line.requested_date and line.order_id.requested_date:
```
espo-tony: I think it should also check if 'requested_date' is not already in 'vals'.

espo-tony (changes requested):
It's not clear to me if this module is meant to force lines to have the same requested_date than their order or if it should be possible to set different dates between order and lines.

sale_order_line_date/models/sale_order.py
nikul-serpentcs: @espo-tony Could you please review.

espo-tony: I rather meant is that if requested_date is in vals it shouldn't change its value. In other words, it should only add 'requested_date' in vals if it's not already present.

Is that consistent with the goal of the module?

sale_order_line_date/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+# © 2016 Eficent Business and IT Consulting Services, S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale Order Line Date",
+    "version": "10.0.1.0.0",
+    "author": "OdooMRP team,"
+              "AvanzOSC,"
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
```
LoisRForgeFlow, line 14: Should this be changed to `Tecnativa`? @pedrobaeza

AaronHForgeFlow (commented):
Hi @nikul-serpentcs can you fix travis? Functional test ok. Thanks.

nikul-serpentcs:
@aheficent Fixed Travis, Could you please review.

AaronHForgeFlow:
Hi @espo-tony, IMHO we should let the user to change the requested date when the SO is in quotation state. Using current approach, if the user makes a mistake there's no way the user can fix it. Furthermore, that is the current behavior in v9.

AaronHForgeFlow:
@espo-tony The requested date in the SO lines is the one that is propagated. The requested date in the SO can be used to set the same requested date to all the SO lines, just that.

AaronHForgeFlow (approved):
:+1: From my side this is ok.

sale_order_line_date/models/sale_order.py
```diff
@@ -0,0 +1,63 @@
...
+        if 'warning' not in result:
+            lines = []
+            for line in self.order_line:
+                lines.append((1, line.id, {'requested_date':
+                                           self.requested_date}))
+            result['value'] = {'order_line': lines}
+        return result
+
+
+class SaleOrderLine(models.Model):
```
AaronHForgeFlow: Hi @nikul-serpentcs I'd put this code in another file called sale_order_line.py to make this easier to read

espo-tony:
Hi @aheficent I totally agree. Indeed it's what I was suggesting: it should only add 'requested_date' in vals if it's not already present - if it's already present in vals, it's because the user manually changed it, and in this case it shouldn't be overwritten.

espo-tony (approved):
:+1:

nikul-serpentcs:
@aheficent  I did put a code in another file, Could you please review.

AaronHForgeFlow (approved):
:+1: Still working

sale_order_line_date/__manifest__.py
AaronHForgeFlow: I think better to change this. @pedrobaeza

AaronHForgeFlow:
@lreficent can you update your review?

LoisRForgeFlow (approved):
LGTM

JordiBForgeFlow:
@nikul-serpentcs can you squash your commits in order to have a single commit for the migration to v10?

nikul-serpentcs:
@jbeficent squashed commits, Could You Please check it ??

atchuthan (approved):

AaronHForgeFlow:
IMHO, this can be merged :innocent:

PR #514, opened 2017-08-01
[9.0][FIX] sale_sourced_by_line: not applying default warehouse when grouping

LoisRForgeFlow:
**Before the fix**:
If you create a sale order with 'Your Company' as *Default Warehouse* and two lines, one with 'Your Company' as *Source Warehouse* and the other without specifying it, you end up with two deliveries associated to this SO.

**After the fix:**
In the same case the lines are correctly grouped and you would only have one delivery for the SO.

selection_125

**Other minor fixes done in this PR:**
Also, I'm taking the occasion to do other two minor fixes:
* Remove an unused file.
* Add a `ensure_one()` constraint in the parent method `_get_procurement_group_key` in the module `sale_procurement_group_by_line`


JordiBForgeFlow (approved):

JordiBForgeFlow:
@aheficent can you review?

AaronHForgeFlow (approved):
:+1: Functional test + code review. It applies the default warehouse when grouping

MiquelRForgeFlow (approved):

PR #517, opened 2017-08-24
[10.0][add] Module product price category

grindtildeath:
Product Price Category
======================

This module adds a field Price Category on Product Template
and allow Pricelist to be applied on this field.


product_price_category/models/product_pricelist.py
```diff
@@ -0,0 +1,249 @@
...
+                price, suitable_rule and suitable_rule.id or False
+            )
+        return results
+
+
+class ProductPricelistItem(models.Model):
+    _inherit = 'product.pricelist.item'
+
+    applied_on = fields.Selection(
+        selection_add=[('2b_product_price_category', 'Price Category')]
```
legalsylvain, line 222: Hi, 
why there is the prefix "2b_" ?

legalsylvain (approved):
Hi @grindtildeath,
Interesting module ! 
LGTM. Code review, no test.

product_price_category/models/product_pricelist.py
grindtildeath: Hi @legalsylvain,
Thanks for your review. This module was developed as a specific module for a customer by @cyrilgdn, who left camptocamp and I only did adapt it to fit OCA standards, so I guess it was to fit specific needs.
We might have to rework it though but I'm just back from 3 weeks holidays and don't know what's the status on it.

yvaucher (approved):

product_price_category/models/product_pricelist.py
yvaucher: @legalsylvain 

as it is an alternative for the 2_product_category option it makes sense to have `2b` instead of 4

``` python
    applied_on = fields.Selection([
        ('3_global', 'Global'),
        ('2_product_category', ' Product Category'),
        ('1_product', 'Product'),
        ('0_product_variant', 'Product Variant')], "Apply On",
        default='3_global', required=True,
        help='Pricelist Item applicable on selected option')

```

PR #525, opened 2017-09-07
[MIG] sale_triple_discount: Migration to 10.0

chienandalu:
Sale Triple Discount
====================

This module allows to have three successive discounts on every sale order
line.

cc @Tecnativa


MiquelRForgeFlow:
What about https://github.com/OCA/sale-workflow/pull/437 and what about have been said in it?

pedrobaeza:
Check https://github.com/OCA/sale-workflow/pull/437#issuecomment-325614967. They said that we can close that PR.

chienandalu:
@mreficent Our bet here (as in the other modules of this family we've made contributions to) is to stick to Odoo's standard as much as possible extending already existing fields and methods which results in a more straightforward approach I think.

MiquelRForgeFlow (approved):
LGTM :+1:

pedrobaeza (approved):

pedrobaeza:
@luismontalba please make an extra review.

luismontalba:
testing it on runbot it works, but it seems to be some problem in the view when creating a sale order line:

saleorder

chienandalu:
@luismontalba Fixed!

luismontalba (approved):
Tested on runbot

PR #527, opened 2017-09-22
Add ability to specify customer delivery address in the wizard.

ghost:
* delivery address filtered by SO customer
* delivery address specified in procurement.order


guewen:
@oleksandrpaziuk there is a lot of commits there, is it normal?

PR #534, opened 2017-09-26
[ADD] sale_order_product_recommendation: SO product wizard

yajo:
This wizard that you can see now in your quotations will let you add products quickly based on the customer's recent purchase history.

~~Currently WIP due to a problem I can't seem to nail down: prices show always as zero:
captura de pantalla de 2017-09-26 11-47-35~~

@Tecnativa


sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,142 @@
...
+        readonly=True,
+    )
+
+    @api.multi
+    @api.depends("partner_id", "product_id", "pricelist_id", "units_included")
+    def _compute_price_unit(self):
+        for one in self:
+            one.price_unit = one.product_id.with_context(
+                partner=one.partner_id.id,
+                pricelist_id=one.pricelist_id,
```
sergio-teruel: Change pricelist_id by pricelist
https://github.com/OCA/OCB/blob/9.0/addons/product/product.py#L789

yajo:
Thanks @sergio-teruel, it's working now! (I also had to add `.id` 😅)

rafaelbn:
Hi @Yajo , travis is red. Could you check it please?

yajo:
Last update includes products from current SO to wizard recommendations. It needed to:

- Autofill its included units
- Delete lines from SO if they were set to zero units
- Merge multiple lines if they share same product
- Update existing lines if qty has changed

Little feature, big job. It should be working in any case.

yajo:
OK so this is 💚 now finally

pedrobaeza (approved):
After I have fixed one thing

carlosdauden (approved):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,177 @@
...
+        start = fields.Datetime.to_string(start)
+        order_lines = self.order_id.order_line
+        existing_product_ids = set(order_lines.mapped("product_id").ids)
+        self.line_ids = False
+        # Search delivered products in previous months
+        found_lines = self.env["sale.order.line"].read_group(
+            [
+                ("order_partner_id", "child_of",
+                 self.order_id.partner_id.commercial_partner_id.id),
+                ("order_id.date_order", ">=", start),
```
carlosdauden: order_id.date_order can be replace with create_date (if little diferences are assumed) or create related field and set index=True to better performance

carlosdauden: Execute Sql sentence with OVER clause can be another option to performance

yajo:
Can we merge then?

pedrobaeza:
Please validate latest change by @carlosdauden

rafaelbn (approved):
Tested functionally in runbot :+1:

rafaelbn:
Shouldn't we migrate this module to v10 or v11 @Yajo ?

yajo:
I guess

pedrobaeza:
v11 as per our roadmap, but not yet.

PR #536, opened 2017-09-28
[MIG][9.0][sale_partner_incoterm] port sale partner incoterm wp

wpichler:


rousseldenis:
@wpichler Is this still applicable ?

wpichler:
I think no one will need it - the module is already in branch 10 and 12 - so no need here for this

PR #540, opened 2017-10-05
[FIX] sale_rental: use v10 version for stock.location.path function

JulieSFL:
`_prepare_push_apply()` doesn't exist anymore in v10 ; use `_prepare_move_copy_values()` instead


alexis-via (approved):
Code review, no test

PR #544, opened 2017-10-10
[10.0][FIX] sale_triple_discount: case taxes globally rounded

luismontalba:
This PR fixes calculation of tax amount in case company tax calculation rounding method is "Round globally".

This is the error:

errorsaletriplediscount

### DIFF
--- a/sale_triple_discount/models/sale_order.py
+++ b/sale_triple_discount/models/sale_order.py
@@ -2,10 +2,43 @@
 # Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
 # Copyright 2017 Alex Comba - Agile Business Group
 # Copyright 2017 Tecnativa - David Vidal
+# Copyright 2017 Tecnativa - Luis M. Ontalba
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
 import odoo.addons.decimal_precision as dp
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends('order_line.price_total')
+    def _amount_all(self):
+        """
+        Compute the total amounts of the SO.
+        """
+        for order in self:
+            if not order.company_id.tax_calculation_rounding_method == (
+                    'round_globally'):
+                return super(SaleOrder, self)._amount_all()
+            amount_untaxed = amount_tax = 0.0
+            for line in order.order_line:
+                amount_untaxed += line.price_subtotal
+                taxes = line.tax_id.compute_all(
+                    line.price_reduce,
+                    line.order_id.currency_id,
+                    line.product_uom_qty,
+                    product=line.product_id,
+                    partner=order.partner_shipping_id,
+                )
+                amount_tax += sum(
+                    t.get('amount', 0.0) for t in taxes.get('taxes', []))
+            order.update({
+                'amount_untaxed': order.pricelist_id.currency_id.round(
+                    amount_untaxed),
+                'amount_tax': order.pricelist_id.currency_id.round(amount_tax),
+                'amount_total': amount_untaxed + amount_tax,
+            })
 
 
 class SaleOrderLine(models.Model):
--- a/sale_triple_discount/models/sale_order.py
+++ b/sale_triple_discount/models/sale_order.py
@@ -14,35 +14,38 @@ class SaleOrder(models.Model):
 
     @api.depends('order_line.price_total')
     def _amount_all(self):
-        """
-        Compute the total amounts of the SO.
-        """
+        extra_context = {'compute_triple_discount': True}
         for order in self:
-            if not order.company_id.tax_calculation_rounding_method == (
+            if (order.company_id.tax_calculation_rounding_method ==
                     'round_globally'):
+                return super(SaleOrder, self.with_context(**extra_context)
+                             )._amount_all()
+            else:
                 return super(SaleOrder, self)._amount_all()
-            amount_untaxed = amount_tax = 0.0
-            for line in order.order_line:
-                amount_untaxed += line.price_subtotal
-                price = line.price_unit * (
-                    1 - (line.discount or 0.0) / 100.0) * (
-                    1 - (line.discount2 or 0.0) / 100.0) * (
-                    1 - (line.discount3 or 0.0) / 100.0)
-                taxes = line.tax_id.compute_all(
-                    price,
-                    line.order_id.currency_id,
-                    line.product_uom_qty,
-                    product=line.product_id,
-                    partner=order.partner_shipping_id,
-                )
-                amount_tax += sum(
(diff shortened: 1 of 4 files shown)


pedrobaeza:
I don't get why are you using a weird field like `price_reduce`. Make the same technique like the rest: store previous `price_unit` value, change it to the new one, call super, and restore the value.

sale_triple_discount/models/sale_order.py
```diff
@@ -18,27 +18,22 @@ def _amount_all(self):
...
-                'amount_total': amount_untaxed + amount_tax,
-            })
+                prev_price_unit['line.id'] = line.price_unit
+                prev_price_subtotal = line.price_subtotal
+                price_unit = line.price_unit * (
+                    1 - (line.discount2 or 0.0) / 100.0)
+                price_unit *= (1 - (line.discount3 or 0.0) / 100.0)
+                line.update({
+                    'price_unit': price_unit,
+                    'price_subtotal': prev_price_subtotal
```
pedrobaeza: Why do you need to set price_subtotal?

sale_triple_discount/tests/test_sale_triple_discount.py
```diff
@@ -112,13 +111,3 @@ def test_04_sale_order_triple_discount_invoicing(self):
         self.assertEqual(self.so_line2.discount3,
                          invoice.invoice_line_ids[1].discount3)
         self.assertEqual(self.order.amount_total, invoice.amount_total)
-
-    def test_05_sale_order_triple_discount_globally_rounding(self):
```
pedrobaeza: Why are you removing the test? It's a valid use case that should be checked.

sale_triple_discount/models/sale_order.py
```diff
@@ -18,27 +18,22 @@ def _amount_all(self):
...
-                amount_tax += sum(
-                    t.get('amount', 0.0) for t in taxes.get('taxes', []))
-            order.update({
-                'amount_untaxed': order.pricelist_id.currency_id.round(
-                    amount_untaxed),
-                'amount_tax': order.pricelist_id.currency_id.round(amount_tax),
-                'amount_total': amount_untaxed + amount_tax,
-            })
+                prev_price_unit['line.id'] = line.price_unit
+                prev_price_subtotal = line.price_subtotal
```
pedrobaeza: You're still not following the same scheme than in the `_compute_amount` of `sale.order.line` model.

luismontalba: I'm tried to follow same scheme than in the _compute_amount of sale.order.line mode, but it doesn't work, I think because updating sale order line values trigers again _amount_all function in sale.order model ...

For example, this code is wrong:

    @api.depends('order_line.price_total')
    def _amount_all(self):
        for order in self:
            prev_price_unit = {}
            prev_discount = prev_discount2 = prev_discount3 ={}
            for line in order.order_line:
                prev_discount['line.id'] = line.discount
                prev_discount2['line.id'] = line.discount2
                prev_discount3['line.id'] = line.discount3
                prev_price_unit['line.id'] = line.price_unit
                price_unit = line.price_unit * (
                    1 - (line.discount or 0.0) / 100.0) * (
                                 1 - (line.discount2 or 0.0) / 100.0) * (
                                 1 - (line.discount3 or 0.0) / 100.0)
                line.update({
                    'price_unit': price_unit,
                    'discount': 0.0,
                    'discount2': 0.0,
                    'discount3': 0.0,
                })
            super(SaleOrder, self)._amount_all()
            for line in order.order_line:
                line.update({
                    'price_unit': prev_price_unit['line.id'],
                    'discount': prev_discount['line.id'],
                    'discount2': prev_discount2['line.id'],
                    'discount3': prev_discount3['line.id'],
                })

The only way I find is rewriting the function in order to calculate the new price inside.

sale_triple_discount/tests/test_sale_triple_discount.py
luismontalba: Test added

damendieta:
I, I'm back porting this module to v9 and I'd like to merge this PR on this, could you please tell me the OCA's standard way to do it.

The code is on #562 

Thanks a lot.

luismontalba:
Hi @damendieta ,
You can find in OCA repository a guide:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

pedrobaeza:
@luismontalba this is not an Odoo module (but OCA module), so the second link doesn't apply.

luismontalba:
Right! I deleted the second link in my previous comment

sale_triple_discount/models/sale_order.py
```diff
@@ -2,12 +2,49 @@
...
+
+    @api.depends('order_line.price_total')
+    def _amount_all(self):
+        """
+        Compute the total amounts of the SO.
+        """
+        for order in self:
+            if not order.company_id.tax_calculation_rounding_method == (
+                    'round_globally'):
+                return super(SaleOrder, self)._amount_all()
```
pedrobaeza: This shouldn't return or you will cut at half the computation, and it should be `order`, and not `self`

pedrobaeza (changes requested):
@chienandalu can you check this and see if there's a better method for not rewriting all that stuff?

chienandalu:
@pedrobaeza I'm trying it this way: https://github.com/OCA/sale-workflow/pull/544/commits/54bafd25887791f63df02e032c61598b2a4e4fac but it doesn't work well yet. The compute is done correctly but somehow gets lost after de `update()`

My first go was to try on the tax `compute_all()` method, but we don't have any relation to wich line is doing the request, so it can't be determined wich is the price to pass instead of the `super()` one

rafaelbn:
Hi @pedrobaeza @chienandalu ,  @mpanarin make this PR about discount computation https://github.com/OCA/sale-workflow/pull/671 maybe we should finish this PR?

rafaelbn:
I cannot test this because this issue: https://github.com/OCA/sale-workflow/issues/630

rafaelbn:
@pedrobaeza @chienandalu could we close this PR? We are not using this module anymore.

pedrobaeza:
Well, the problem is still there, but let's close it as no longer needed

SimoRubi:
@pedrobaeza I just tried this in runbot v10 and looks like it is solved:
image

pedrobaeza:
Do you mean with global taxes? Maybe another patch landed there and this is the reason why the conflict appeared.

SimoRubi:
> Do you mean with global taxes? Maybe another patch landed there and this is the reason why the conflict appeared.

yes I had global taxes enabled, you can still check the order and the configuration in http://3360394-10-0-14af82.runbot3.odoo-community.org/web as long as it is online

pedrobaeza:
OK, then glad to hear that it's solved. Then definitively this PR is not needed.

PR #549, opened 2017-10-23
[9.0][IMP][sale_automatic_workflow_payment_mode] use date invoice as date of payment

vrenaville:
In some case automatic payment is activated after a few month.
If you have invoice in currency you can have a different rate, between the invoice date and the payment date
so my proposition is to take the date_invoice as payment date instead of current date

### DIFF
--- a/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
 import logging
-from openerp import models, api, fields
+from openerp import models, api
 from openerp.tools.safe_eval import safe_eval
 from openerp.addons.sale_automatic_workflow.models.automatic_workflow_job \
     import commit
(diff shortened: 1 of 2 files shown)


simahawk (approved):

pedrobaeza:
@sergio-teruel can you check this?

leemannd:
@sergio-teruel @pedrobaeza Do you still want to check it?

leemannd (approved):

pedrobaeza:
I would say that you don't want to register the payment in the past...

rafaelbn:
Hi @simahawk @leemannd I think @guewen could be the one who make the las review o just merge it

rafaelbn:
Hi @vrenaville , Why are you  modifying sale_rental in this PR? This PR applies yo  sale_automatic_workflow_payment_mode only, isn't it?

rafaelbn:
Hi @vrenaville do you want this PR or can we close it?

leemannd:
@rafaelbn We are using it in production with success since the creation of this PR. As we integrate it and are able to retrieve it. I think that it is not an issue for us and this particular client.

Sidenote: After searching a bit, the first sentence is telling the base use `In some case automatic payment is activated after a few month.` We had issues with this module and had to deactivate it. This development aims to be more robust in such cases.

So in the end that's your call. If you want to merge it or not.

cc @simahawk

simahawk:
> Hi @vrenaville , Why are you modifying sale_rental in this PR? This PR applies yo sale_automatic_workflow_payment_mode only, isn't it?

this change could be discarded indeed but it won't hurt (explicit catch of generic exception).

PR #556, opened 2017-10-31
[IMP] sale_order_invoicing_finished_task

gurneyalex:
When the product has an invoicing policy Ordered Quantity, setting the task
as Invoiceable will set the delivered qty on the sale order line to the ordered quantity.

### DIFF
--- a/sale_order_invoicing_finished_task/models/sale_order.py
+++ b/sale_order_invoicing_finished_task/models/sale_order.py
@@ -48,3 +48,20 @@ class SaleOrderLine(models.Model):
             else:
                 line.qty_to_invoice = 0.0
         super(SaleOrderLine, self - lines)._get_to_invoice_qty()
+
+    @api.multi
+    def _compute_analytic(self, domain=None):
+        if not domain and self.ids:
+            domain = [
+                ('so_line', 'in', self.ids),
+                # don't update the qty on sale order lines which are not
+                # with a product invoiced on ordered qty +
+                # invoice_finished task = True
+                '|',
+                ('so_line.product_id.invoice_policy', '!=', 'order'),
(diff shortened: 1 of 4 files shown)


gurneyalex:
@rafaelbn @sergio-teruel this PR adds the bit I'm missing in the new code from #485 : for fixed price sales lines, I really need the delivered qty to be == ordered qty when the task is set as invoiceable. (I'm using these values to convert these to amounts and cannot afford to have 2 different and incompatible units in theses columns, such as Units and Hours).

sale_order_invoicing_finished_task/models/sale_order.py
```diff
@@ -48,3 +48,19 @@ def _get_to_invoice_qty(self):
...
+                      # don't update the qty on sale order lines which are not
+                      # with a product invoiced on ordered qty +
+                      # invoice_finished task = True
+                      '|', ('so_line.product_id.invoice_policy', '!=', 'order'),
+                      ('so_line.product_id.invoicing_finished_task', '=', False),
+                      '|',
+                      ('amount', '<=', 0.0),
+                      ('project_id', '!=', False),
+                      ]
+            print domain
```
pedrobaeza: Remove the print

pedrobaeza (approved):
I suppose that you want to track the line via timesheets but still have a fixed price. This is a bit outside the scope of this module, but well, we can include it.

carlosdauden (approved):

rafaelbn:
Tested in runbot 👍 
But travis is red @gurneyalex

rafaelbn (approved):
Please merge after travis is green! Thanks!

jcoux:
I added a commit to fix flake8 errors.

jcoux:
I added a commit to fix flake8 errors.

PR #563, opened 2017-11-11
10.0 mig sale packaging price

grauschnabel:
should be fine, works good here


grauschnabel:
stop here, not all checks have passed .... I come back later.

PR #535, opened 2017-09-28
[ADD] Module Sale Order Action Invoice Create Hook

rgarnau:
This module adds hook points in sale_order.action_invoice_create() in order to add more flexibility in the grouping parameters for the creation of invoices and to give the possibility to take into account draft invoices.

@jbeficent @etobella

PR #567, opened 2017-11-14
[11.0][MIG] sale_order_action_invoice_create_hook

rgarnau:
no major changes

Sale Order Action Invoice Create Hook
=====================================

This module adds hook points in sale_order.action_invoice_create() in order
to add more flexibility in the grouping parameters for the creation of
invoices and to give the possibility to take into account draft invoices.

Although the original code has been modified a bit, the logic has been
respected and the method still does exactly the same.

Usage
=====

#. Add dependency of this module
#. Inherit from 'sale.order'
#. Change the _get_invoice_group_key function in order to return more
   grouping fields to take into account when creating a single invoice from
   various Sales Orders.
#. Change the _get_draft_invoices function in order to take into account
   existing draft invoices when creating new ones. This feature is already
   implemented in the module 11.0-sale_merge_draft_invoice.


JordiBForgeFlow (approved):
:+1:

sale_order_action_invoice_create_hook/README.rst
```diff
@@ -22,11 +23,11 @@ Usage
    various Sales Orders.
 #. Change the _get_draft_invoices function in order to take into account
    existing draft invoices when creating new ones. This feature is already
-   implemented in the module 10.0-sale_merge_draft_invoice.
+   implemented in the module 11.0-sale_merge_draft_invoice.
```
StefanRijnhart: is `11.0-` part of the module name?

StefanRijnhart (approved):
I checked that the upstream method was not modified between Odoo 10 and Odoo 11

sale_order_action_invoice_create_hook/model/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+    @api.model
+    def _get_invoice_group_key(self, order):
+        return (order.partner_invoice_id.id, order.currency_id.id)
+
+    @api.model
+    def _get_draft_invoices(self, invoices, references):
+        return invoices, references
+
+    @api.multi
+    def action_invoice_create(self, grouped=False, final=False):
```
StefanRijnhart: Oh, I now see that you potentially break the inheritance chain. That was already the case in the previous version of course, but there was no discussion or rationale about that on the original PR https://github.com/OCA/sale-workflow/pull/535. As I understand it, hooks need to be monkeypatched in the post_load hook to overwrite the original method from odoo.addons.sale.models.sale.SaleOrder. That way, any inheritance on the method is kept intact.

rgarnau:
@StefanRijnhart Thanks! Just added the post_load_hook.

sale_order_action_invoice_create_hook/hooks.py
```diff
@@ -0,0 +1,125 @@
...
+
+from odoo import _
+from odoo.exceptions import UserError
+from odoo.tools import float_is_zero
+from odoo.addons.sale.models.sale import SaleOrder
+
+
+def post_load_hook():
+
+    def new_action_invoice_create(self, grouped=False, final=False):
```
StefanRijnhart, line 12: Thanks for making this a post_load monkeypatch! To prevent problems with multiple databases on the same instance, please store the original method on the SaleOrder object as well and call that if `self` does not have _get_invoice_group_key (which will be the case on databases that don't have this module installed):

In post_load_hook:
```
if not hasattr(SaleOrder, 'action_invoice_create_original'):
    SaleOrder.action_invoice_create_original = SaleOrder.action_invoice_create
```
In new_action_invoice_create:
```
if not hasattr(self, '_get_invoice_group_key'):
    return self.action_invoice_create_original(grouped=grouped, final=final)
```

rgarnau:
@StefanRijnhart good point! thank you! :+1:

JordiBForgeFlow:
@StefanRijnhart Thanks for this valuable insight!!

StefanRijnhart:
Thanks for the follow up! Looking good now. One thing that this module is still missing is a test. Would you mind copying a simple test from the sale module to test if invoice generation still works? You will need to call the hook method manually in your test as it will not be called by Odoo itself until after the test.

rgarnau:
@StefanRijnhart Test added!

StefanRijnhart:
Excellent, thanks!

StefanRijnhart (approved):

etobella (approved):
LGTM :+1:

PR #570, opened 2017-11-18
10.0 FIX automatic validation of pickings if the domain returns more than one.

jaredkipe:
If you have your automatic workflow setup to validate pickings (e.g. import some retail/POS orders), the current method will give the following error.

```
ERROR test10113 odoo.addons.sale_automatic_workflow.models.automatic_workflow_job: Error during an automatic workflow action.
   File "/opt/odoo/addons/sale_automatic_workflow/models/automatic_workflow_job.py", line 23, in savepoint
     yield
   File "/opt/odoo/addons/sale_automatic_workflow/models/automatic_workflow_job.py", line 83, in _validate_pickings
     pickings.validate_picking()
   File "/opt/odoo/addons/sale_automatic_workflow/models/stock_picking.py", line 21, in validate_picking
     self.do_transfer()
   File "/opt/odoo/addons/connector_ecommerce/models/stock.py", line 34, in do_transfer
     result = super(StockPicking, self_context).do_transfer()
   File "/opt/odoo/odoo/addons/delivery/models/stock_picking.py", line 113, in do_transfer
     self.ensure_one()
   File "/usr/local/lib/python2.7/site-packages/odoo/models.py", line 4823, in ensure_one
 ValueError: Expected singleton: stock.picking(17626, 17627,...)
```

### DIFF
--- a/sale_automatic_workflow/__manifest__.py
+++ b/sale_automatic_workflow/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sale Automatic Workflow',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Sales Management',
     'license': 'AGPL-3',
     'author': "Akretion,Camptocamp,Sodexis,Odoo Community Association (OCA)",
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -78,9 +78,9 @@ class AutomaticWorkflowJob(models.Model):
         picking_obj = self.env['stock.picking']
         pickings = picking_obj.search(picking_filter)
         _logger.debug('Pickings to validate: %s', pickings.ids)
-        if pickings:
+        for picking in pickings:
             with savepoint(self.env.cr):
-                pickings.validate_picking()
+                picking.validate_picking()
 
     @api.model
     def _sale_done(self, sale_done_filter):
--- a/sale_automatic_workflow/models/stock_picking.py
+++ b/sale_automatic_workflow/models/stock_picking.py
@@ -17,6 +17,7 @@ class StockPicking(models.Model):
 
     @api.multi
     def validate_picking(self):
-        self.force_assign()
-        self.do_transfer()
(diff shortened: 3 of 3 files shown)


pedrobaeza:
Please change module version number.

jaredkipe:
Done.  May I ask why?  This change doesn't modify any fields or require an 'upgrade'.

pedrobaeza:
Check https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers for the guidelines. 

The goal is to distinguish through numbering the version that is working from the version is not.

jaredkipe:
Did I get it right? (usually just a server restart is required)

pedrobaeza:
Yes, thanks!

pedrobaeza (approved):

jaredkipe:
@pedrobaeza Anything more required to get his merged?

pedrobaeza:
Yes, you need at least 2 approvals and 5 working days or 3 approvals: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#please-respect-a-few-basic-rules

You can review other PRs and ask in exchange to review yours.

jaredkipe:
@guewen FWIW this does affect the Magento Connector as a dependency ;)

LoisRForgeFlow (approved):

atchuthan (changes requested):
Suggestion
On line https://github.com/OCA/sale-workflow/blob/10.0/sale_automatic_workflow/models/automatic_workflow_job.py#L81

Instead of 
if pickings:
  with savepoint(self.env.cr):
  pickings.validate_picking()

we could replace with 
for picking in pickings:
  with savepoint(self.env.cr):
  picking.validate_picking()

As this change will automatically save the changes done for each picking instead of committing changes completely in validate_picking()

jaredkipe:
@atchuthan 
That is fair, and makes it more similar to other automatic workflow jobs.

I've left the other commit as the method is decorated `@api.multi` and will now be suitable for using as a server action on its own (e.g. on a list view).

atchuthan (approved):
sure :+1:

PR #571, opened 2017-11-20
[FIX][10.0]Sale procurement group by line. Get pickings

AaronHForgeFlow:
Current behavior: Overwriting the standard completely without calling super
Expected behavior: Call the standard method before introducing changes

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -22,7 +22,8 @@ class SaleOrder(models.Model):
 
     @api.multi
     @api.depends('order_line')
-    def _compute_get_picking_ids(self):
+    def _compute_picking_ids(self):
+        super(SaleOrder, self)._compute_picking_ids()
         for sale in self:
             group_ids = set([line.procurement_group_id.id
                              for line in sale.order_line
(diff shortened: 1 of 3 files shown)


sale_procurement_group_by_line/model/sale.py
```diff
@@ -34,10 +35,7 @@ def _compute_get_picking_ids(self):
                 [('group_id', 'in', list(group_ids))])
             sale.delivery_count = len(sale.picking_ids)
 
-    picking_ids = fields.Many2many('stock.picking',
-                                   compute='_compute_get_picking_ids',
-                                   method=True,
-                                   string='Picking associated to this sale')
+    picking_ids = fields.Many2many(compute='_compute_picking_ids')
```
LoisRForgeFlow: I would say that this isn't even needed as it is correctly declared in standard odoo :thinking:

AaronHForgeFlow:
@lreficent you are totally right :)

LoisRForgeFlow (approved):
:+1:

LoisRForgeFlow (commented):
Ups! hold on, seems that some test is failing, could you check thah?

AaronHForgeFlow:
@lreficent yeap it was failing because the test was calling the previous method. It should be fine now

LoisRForgeFlow:
good for me now! :+1:

MiquelRForgeFlow (approved):

PR #572, opened 2017-11-23
[10.0][MIG] sale_delivery_block

AaronHForgeFlow:
- Bump code to new version
- No Extra features added/removed

If https://github.com/OCA/sale-workflow/tree/10.0/sale_procurement_group_by_line is installed you should also use this one: https://github.com/OCA/sale-workflow/pull/573


AaronHForgeFlow:
@lreficent you may like this :hugs:

LoisRForgeFlow:
I'm having some issues validating throught runbot:
* sale_order_revision module is always blocking me with the constraint: "Order Reference and revision must be unique per Company."
* some times a singleton error pops when entering SO.

However, none of them are related to this module I think.
EDIT: 1st point is clearly not related, just to clarify that 2nd either does I attach a snapshot:
selection_207

LoisRForgeFlow (approved):
Code review LGTM and tests are green so :+1: 

@aheficent could you please squash transifex commits?

sale_delivery_block/README.rst
```diff
@@ -0,0 +1,78 @@
...
+To use this module, you need to:
+
+#. Create a new sales order and provide a 'Delivery Block Reason'.
+#. Confirm Sale (No delivery would be created).
+#. Release Delivery Block when it is time to create the deliveries for
+   the sales order.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/9.0
```
LoisRForgeFlow: 10.0

AaronHForgeFlow:
It seems to me that the singleton issue is related to `sale_order_margin_percent` module

StefanRijnhart (approved):

sale_delivery_block/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from . import sale_order
+from . import sale_delivery_block_reason
```
JayVora-SerpentCS: sale_delivery_block_reason file should be loaded earlier than sale_order

LoisRForgeFlow: why?

JayVora-SerpentCS: There is a manye2one reference in sale_order file and probably thats why it should come after sale_delivery_block_reason. So the order of import stays justified. I just had an experience of a trouble in one another such case

StefanRijnhart: That would be odd. In my experience, Odoo is able to instantiate the models in a module regardless of loading order even if there are cross references in relational fields (which is very common). And tests are green so that happens here as well.

sale_delivery_block/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+{
+    "name": "Sale Delivery Block",
+    "summary": "Allow you to block the creation of deliveries "
+               "from a sale order.",
+    "version": "10.0.1.0.0",
+    "author": "Eficent, "
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "category": "Sales",
+    "depends": ["sale_stock"],
```
rousseldenis: IMHO it's better to define explicit dependencies (at first sight I saw base and sale). It facilitates understanding module aim and migration in case of changes in dependant modules.

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.multi
+    @api.constrains('delivery_block_id')
+    def _check_not_auto_done(self):
+        for so in self:
+            if so.delivery_block_id and self.env['ir.values'].get_default(
```
rousseldenis: get_default can be done before the loop to avoid calling it several times.

Maybe, you can change the loop by an 'any' function.

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+        res = super(SaleOrder, self).onchange_partner_id()
+        for so in self:
+            so.delivery_block_id = so.partner_id.default_delivery_block or \
+                False
+        return res
+
+    @api.multi
+    def action_remove_delivery_block(self):
+        """Remove the delivery block and create procurements as usual."""
+        self.write({'delivery_block_id': False})
```
rousseldenis: Maybe you can add before a filter (the same as in the view) as you cannot rely logic just on form view side (remember that this can be called through xml-rpc).

sale_delivery_block/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+    _inherit = "sale.order.line"
+
+    @api.multi
+    def _action_procurement_create(self):
+        new_procs = self.env['procurement.order']
+        # When module 'sale_procurement_group_by_line' is installed we do not
+        # want to use this method but call super. See module
+        # 'sale_delivery_block_proc_group_by_line'. This solves a inheritance
+        # order issue.
+        if 'group_by_line' in self.env.context:
```
rousseldenis: This is not correct.

As sale_delivery_block_proc_group_by_line depends on this module, the _action_procurement_create will be first called in that module (to make it simple).

Move this code there.

rousseldenis (changes requested):
Code review. Some comments

sale_delivery_block/__manifest__.py
StefanRijnhart: I recently had a case with exactly the same dependency. It depended on 'sale' and 'stock' instead of 'sale_stock' and it broke the tests as they were executed before the glue module was loaded. So I think keeping this should be fine.

sale_delivery_block/models/sale_order.py
AaronHForgeFlow: So I understand that the method at sale_delivery_block_proc_group_by_line will be always called before?

rousseldenis: I just say that this code has nothing to do with the purpose of this module and should not be here.

You just have to move it to sale_delivery_block_proc_group_by_line and it should be sufficient (as you depends on this module in that one).

AaronHForgeFlow:
@rousseldenis I just checked the behavior with both modules installed and I think it's still fine after removing the relevant code for sale_procurement_group_by_line. The resulting picking is still split by line according to different criteria.

sale_delivery_block/__manifest__.py
rousseldenis: I think that let dependencies holes. Moreover, I think building dependencies on tests failures (or not) is a bad idea. But, just my opinion

StefanRijnhart: Sorry, I don't understand that. Can you clarify the exact problem? I think the test failure simply indicates that sale_stock is the proper dependency because Odoo is not designed to have sale and stock installed but not sale_stock.

rousseldenis: I see. I just say that explicit depends is better than implicit.

If your module depends on [logic, data, ..] defined on another, take it (to summarize, if you have to depend on 'sale_stock', take it, but if you depend on sale and stock, take them both too)

AaronHForgeFlow: According to this almost every module should depend on base. I have no idea if putting all the implicit dependencies to every module would lead to other problems.

pedrobaeza:
I would't use the term `delivery`, but `stock`, as delivery is used for delivery methods, expenses, and so on. Something like `sale_stock_picking_blocking`

rousseldenis (approved):
Code review

AaronHForgeFlow:
@pedrobaeza  Maybe the term `delivery` can be a little confusing. However, I think it's better to deal with this than renaming the module. Renaming the data model will increase a lot the difficulty of the migration scripts for this module. Models, fields and security groups would be affected.

pedrobaeza:
No, everything is handled easily with only one openupgradelib function: `update_module_names`

AaronHForgeFlow:
What if we just change the module name and the file names but not the data model and the security group?

pedrobaeza:
OK, deal.

AaronHForgeFlow:
Do not merge this yet until the module works as it was before it was renamed
Update: Ready for review again, it works as before.

StefanRijnhart (approved):
Approved again!

PR #577, opened 2017-12-01
Mig/10.0/sale revert done

StefanRijnhart:


pedrobaeza (changes requested):
You should raise an exception if `action_revert_done` and the sale is not in done state.

StefanRijnhart:
@pedrobaeza where does this change request come from? This is just a migration of a trivial module. When I look in the core sale module of Odoo, there is no such raising at all based on state flow integrity in any of the state change methods so there is no precedent to add it here. State integrity is up to the caller (in this case the user interface), and the check on the correct state is in the button attrs in this module so we have it covered.

pedrobaeza:
It's for avoiding incorrect calls from code, the same as in old module sale_order_back2draft (https://github.com/OCA/sale-workflow/blob/abfc5d42a361ba22f612e9e607c1a8f16fd74ec2/sale_order_back2draft/models/sale_order.py#L16). Odoo upstream habits don't mean that they do always the proper way.

sale_revert_done/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, models
```
rgarnau: odoo instead of openerp

rgarnau (approved):

sale_revert_done/models/sale_order.py
StefanRijnhart: Thanks, fixed!

StefanRijnhart:
@pedrobaeza added a validation of the original order state.

sale_revert_done/tests/test_sale_revert_done.py
```diff
@@ -0,0 +1,18 @@
...
+class TestSaleRevertDone(TransactionCase):
+    def test_sale_revert_done(self):
+        order = self.env.ref('sale.sale_order_4').copy()
+        order.action_confirm()
+        self.assertEqual(order.state, 'sale')
+        order.action_done()
+        self.assertEqual(order.state, 'done')
+        order.action_revert_done()
+        self.assertEqual(order.state, 'sale')
+        with self.assertRaisesRegexp(ValidationError, 'not in state'):
```
pedrobaeza, line 15: As a side note, I prefer to not check texts, only exception types, because they can change on the life cycle of the module or because the user is not in 'English' language. It's very weird to have an exception in these small pieces of code that is not the one we are looking for.

pedrobaeza (approved):

sale_revert_done/tests/test_sale_revert_done.py
StefanRijnhart: Will the nitpicking ever stop! I go at lengths to implement this validation check at your preference even if I don't see the point. I then check the contents of the ValidationError because *I* prefer it that way and then you come along with *your* preferences. You have to give some air to the contributor, @pedrobaeza, please. And it's not just this particular review of course.

pedrobaeza: This is not blocking, just commenting between colleagues. You have already my approval.

yajo (approved):

yajo:
It looks quite good. Would you mind squashing the migration-related commits to do the merge please? 😊

StefanRijnhart:
@yajo there is one migration related commit and a commit containing some new validation.

JordiBForgeFlow (approved):

PR #579, opened 2017-12-07
[ADD] sale_credit_point

simahawk:
Excerpt from README:

> Sale based on partners' credit point.
> 
> Use case
> -------------
> 
> You have employees that have some credit (up to you where this come from)
> and you allow employees to buy products based on this credit.
> 
> Products are sold by points rather than money.

### DIFF
--- a/sale_credit_point/models/sale_order.py
+++ b/sale_credit_point/models/sale_order.py
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright 2017 Camptocamp SA
+# Copyright 2018 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, models, _
@@ -34,4 +33,4 @@ class SaleOrder(models.Model):
             sale.credit_point_check()
             sale.partner_id.credit_point_decrease(
                 sale.amount_total, comment=self.credit_point_decrease_msg)
-        return super(SaleOrder, self).action_confirm()
+        return super().action_confirm()
--- a/sale_credit_point/models/sale_order.py
+++ b/sale_credit_point/models/sale_order.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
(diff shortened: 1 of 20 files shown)


pedrobaeza:
Why not using `credit_limit` + `partner_sale_risk` (or an extension over it)?

sale_credit_point/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016-2017 Camptocamp SA
```
mpanarin: only 2017? ;)

mpanarin (approved):

simahawk:
@pedrobaeza are you talking about these?

1. https://github.com/OCA/partner-contact/tree/9.0/partner_financial_risk
2. https://www.odoo.com/apps/modules/11.0/partner_credit_limit/

If yes:
a) I didn't knew about them but anyway sound a bit over-complicated for the goal here
b) credit limit is not OCA as far as I see
c) this is a module that we already had and that I cleaned up for sharing.

If ppl won't like this solution I can move it to our repo ;)

pedrobaeza:
I'm talking about 1. that is also migrated to 10.0 in https://github.com/OCA/partner-contact/pull/442 and extended for sale in https://github.com/OCA/partner-contact/tree/9.0/partner_sale_risk.

They are more complete/complicated than yours, but the use case fits exactly for them (changing the `credit_limit` field for `credit_point`), so I think they duplicate the functionality. Why don't you try the other and tell me back?

simahawk:
@pedrobaeza I'll do (next week probably) thanks. Can you set WIP / needs review / whatever label you think fits better here? tnx :)

pedrobaeza:
OK, I set "Work in progress", as we don't have a label "Pedro is again complaining about something" :stuck_out_tongue_winking_eye:

rafaelbn:
Hi @simahawk ! are you going to fix this PR? Soon @OCA-git-bot will close it :smile:

rousseldenis:
@simahawk Is this still applicable ?

simahawk:
@rousseldenis well, I don't know. It's on an old project... I'll check. I'll rebase meanwhile to see what happens w/ tests.

rousseldenis:
@simahawk So, closing as it won't be fixed

PR #580, opened 2017-12-11
[11.0][ADD] sale_disable_inventory_check

phucngta:
This module disables show warning "Not enough inventory" when there isn't enough product in stock


sales_disable_inventory_check/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.onchange('product_uom_qty', 'product_uom', 'route_id')
+    def _onchange_product_id_check_availability(self):
+        res = super(SaleOrderLine,
```
gdgellatly: shame we have to call super, that is an expensive function call just to discard the result.

sales_disable_inventory_check/tests/test_disable_inventory_check.py
```diff
@@ -0,0 +1,34 @@
...
+        # Sale product_1 with quantity is 15
+        order_line = self.env['sale.order.line'].create({
+            'order_id': order.id,
+            'product_id': self.product_1.id,
+            'product_uom': self.product_1.uom_id.id,
+            'product_uom_qty': 15,
+        })
+        res = order_line._onchange_product_id_check_availability()
+        self.assertEqual(res.get('warning', {}), {},
+                         "There's should not be a warning!!!")
```
gdgellatly: spelling. There should not be a warning.  

The tests are probably OK as is, however, as we are testing an onchange new is probably better than create.

gdgellatly (commented):
Hi,

1 minor typo, 2 opinions, but LGTM.

Also, complete the summary field in __manifest__.py and substitute {project_repo} in the README and delete the Usage block

sales_disable_inventory_check/README.rst
```diff
@@ -0,0 +1,55 @@
...
+------------
+
+* Nguyen Tan Phuc <phuc.nt@komit.consulting.com>
+
+Do not contact contributors directly about support or help with technical issues.
+
+Funders
+-------
+
+The development of this module has been financially supported by:
```
jcdrubay: We  should remove Komit from this

sales_disable_inventory_check/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2017 Komit <http://komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Sales Disable Inventory Check",
+    "summary": "Disable warning 'Not enough inventory' when there isn't enough"
+               " product stock",
+    "version": "11.0.1.0.0",
+    "category": "Sale",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Phuc.nt, "
```
jcdrubay: For author, can you put your name and company on the same line "Nguyen Tan Phuc (komit-consulting.com)"

jcdrubay (changes requested):
After taking into account last tiny changes you should request maintainers to take a look ;)

"@OCA/core-maintainers : could you please help to review?"

jcdrubay (approved):

jcdrubay:
Dear @OCA/core-maintainers : could you please help to review? Thanks

pedrobaeza (changes requested):
Module name should be singular.

phucngta:
@pedrobaeza i have changed module name to singular.

StefanRijnhart:
I think this module could be a lot more useful if you could make this a company dependent property setting on the product template model, so that you can disable the warning either globally or per product and/or per company.

jcdrubay:
@StefanRijnhart How about we make it company dependent on the product category with a default value, then overidable on the product template.

Something very similar to what is done for "Income Account": 
`'income': self.property_account_income_id or self.categ_id.property_account_income_categ_id,`

We would need:

 - A new property field "property_is_check_stock" on Product Template and on Product Category
 - A default property like the property with xml record `property_account_income_categ_id`

GSLabIt:
Any eta for this?

StefanRijnhart:
Well, looks like this has stalled completely and hopefully not because I never confirmed @jcdrubay's suggestion because that's pretty much the idea.

jcdrubay:
We will finalize with latest suggestion from my comment: on Jun 22, 2018

NL66278:
I miss any explanation why somebody would want this, and in what circumstances. Checking for availability of a product when you try to sell it seems like a good idea, not?

jcdrubay:
A car dealer would rarely sell cars that are in stock because there are so many models and variants ... they know they don't have the stock because they have only few cars in stock: the 6 cars in display.

In this kind of situation, and other less extreme, the warning message is annoying and useless.

JordiBForgeFlow:
Cc @hveficent

NL66278:
@jcdrubay would the solution than not be to not define the car as a stockable product?

pedrobaeza:
@NL66278 In B2B businesses that works with delayed deliveries, it's also very annoying to have each time the out of stock warning, but you need them to be stockable, not consumable, so +1 for this module.

NL66278 (approved):
LGTM

NL66278:
@pedrobaeza Thanks for the explanation.

HviorForgeFlow (approved):
Also this warning doesn't show quantities correctly in specific locations. Great feature. :+1:

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
@pedrobaeza Can you update your review?

pedrobaeza (approved):

StefanRijnhart:
Did you guys miss that this would be refactored into making this a property on products and categories?

Or did I miss that you all find this unnecessary?

pedrobaeza:
Well, it's a good to have, but on our cases it's not needed. We can at least add it to the roadmap, or if you want to directly make the PR adding the feature, it's welcome.

StefanRijnhart:
OK!

AaronHForgeFlow (commented):
:tada:

jcdrubay:
Thanks for the merge :) the improved version is on the way !

hieulucky111:
New Implement was done in PR https://github.com/OCA/sale-workflow/pull/805

PR #582, opened 2017-12-12
[10.0][FIX] sale_order_invoicing_finished_task: Error creating task without stage

sergio-teruel:
cc @Tecnativa


pedrobaeza (commented):
Please change module version number

pedrobaeza (approved):
As I have just merged a change in this module with version number changed, I merge also this one.

PR #590, opened 2018-01-15
[WIP] [ADD] PoC for index prices via price lists

hbrunn:
Indexed prices on pricelists
============================

This module was written to extend the functionality of price lists to support indexed prices. This means that starting from some base price (the original product price), you can define date ranges where some factor is applied on the price. This is possible with standard Odoo, but a little bit inconvenient. This module saves you the manual work of creating all the price list items neccessary yourself.

Configuration
=============

To configure this module, you need to:

- go to Accounting / Configuration / Price indexes
- create an index, which is just a mapping from periods to price levels, by either creating ranges manually or using the wizard to generate periods
- in a price list item, select computation method `Index` and select your price index
- now if your price list item is picked according to the usual rules, the price will be adjusted from the index, that is the product's price multiplied by the current indexing factor


pricelist_indexed_price/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Therp BV <http://therp.nl>
```
NL66278, line 2: https on all our url's?

hbrunn: sure, will update our template too

NL66278 (changes requested):
When adding a pricelist item I immediately get an exception:


  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 640, in _handle_exception
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 677, in dispatch
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 333, in _call_function
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/service/model.py", line 101, in wrapper
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 935, in __call__
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/addons/web/controllers/main.py", line 890, in call_kw
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/addons/web/controllers/main.py", line 882, in _call_kw
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/api.py", line 696, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/api.py", line 673, in call_kw_model
  File "/home/openeyedev/var/projects/oe_woonbond_10/odoo/parts/odoo/odoo/models.py", line 1374, in fields_view_get
    if '.' in view_ref:
TypeError: argument of type 'int' is not iterable

Apart from this I have real difficulty understanding the logic in the module. While we should have prices based on index, I nowhere see a definition of an index, or how the index should be maintained.

An index is not something that can be computed automatically. It is a figure that is coming from the outside world. Like the dutch statistics bureau (CBS) that each year publishes how prices for household goods have changed (you will have something similar in any country).  Where do I find this in this module?

Indexes are especially used in contract agreements: we agree on a price, with a provision that each year it might increase according to some more or less objective index. So I doubt the usefullnes of basing the price itself on an index. At the start of a contract you add a line for a product. The price is taken from the product, from a pricelist if present, and might be changed due to discounts, or negotiation. Only from that moment on the prices will start to be indexed, so that each time an invoice is created, you will have a freshly computed price, taking into account the original price, the index at the start of the contract (line) and the index at the invoice date: current_price = original price * (current_index / start_index).

rousseldenis:
@hbrunn What's the status of this ?

hbrunn:
abandoned

PR #595, opened 2018-01-23
[10.0][ADD] sale_blanket_order

### LINKED ISSUE #592: [10.0] New feature: blanket sales order management
One of my customers needs to manage blanket sales orders. I cannot find any thing fitting my needs neither in core neither in OCA. So I am planning to implement a new simple module.

My idea is to create a new object "Blanket Order" with these attributes:

- Customer
- Pricelist
- Payment terms
- Validity date
- State (Draft, Opened, Expired)
- List of products with proposed quantity and price

The workflow is:

1. Blanket Order Creation 
The user encodes the blanket order and once he finds an agreement with the customer, he confirms it

2. Consumption of quantities
Once it is confirmed, a wizard is available to generate sale orders. The user can select the product in the list and set the ordered quantity, which can not exceed the blanket order quantity.

3. End of validity

If all proposed quantities has been consumed, no new sale order can be created.
If the validity date is past, the blanket sale orders jump to "Expired" state whatever all quantities are consummed or not

At all moment, the user can know for all products:

- the original quantity
- the ordered quantity
- the delivered quantity
- the invoiced quantity

and can access to linked sales orders.

Does anyone has on opinion on my suggestion ?

andreparames:
This module adds blanket orders, as described in the corresponding issue.

This PR depends on another for the migration of `web_action_conditionable`: https://github.com/OCA/web/pull/845

Fixes #592

### DIFF
--- a/sale_blanket_order/README.rst
+++ b/sale_blanket_order/README.rst
@@ -1,0 +1,69 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+==============
+Blanket Orders
+==============
+
+A blanket order is a pre-agreement to sell a certain number of quantities of
+products at a specific price. From a confirmed blanket order, the users can
+create new sale orders at such price, until the blanket order expires, either
+due to reaching the validity date or exhausting all the quantities of products.
+
+Usage
+=====
+
+A new menu in the Sale area is created, allowing users to create new blanket
+orders.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/11.0
+
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/sale-workflow/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* André Pereira <github@andreparames.com> (https://www.acsone.eu/)
+
+Do not contact contributors directly about support or help with technical issues.
+
+Funders
+-------
+
+The development of this module has been financially supported by:
+
+* Acsone SA/NV
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/sale_blanket_order/__manifest__.py
+++ b/sale_blanket_order/__manifest__.py
@@ -1,0 +1,28 @@
+# coding: utf-8
+# Copyright 2018 Acsone
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Blanket Orders',
+    'category': 'Sale',
+    'license': 'AGPL-3',
+    'author': 'Acsone SA/NV,Odoo Community Association (OCA)',
+    'version': '10.0.1.0.0',
+    'website': 'https://github.com/OCA/sale-workflow',
+    'summary': "Blanket Orders",
+    'depends': [
+        'sale',
+        'web_action_conditionable',
+    ],
+    'data': [
+        'data/sequence.xml',
+        'data/ir_cron.xml',
+        'security/ir.model.access.csv',
+        'security/security.xml',
+        'wizard/create_sale_orders.xml',
+        'views/blanket_orders.xml',
+        'views/sale_orders.xml',
+        'report/templates.xml',
+        'report/report.xml',
+    ],
+    'installable': True,
+}
--- a/sale_blanket_order/__manifest__.py
+++ b/sale_blanket_order/__manifest__.py
@@ -14,6 +14,7 @@
         'web_action_conditionable',
     ],
     'data': [
+        'views/sale_config_settings.xml',
         'data/sequence.xml',
         'data/ir_cron.xml',
         'security/ir.model.access.csv',
--- a/sale_blanket_order/__manifest__.py
+++ b/sale_blanket_order/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     'name': 'Blanket Orders',
-    'category': 'Sale',
+    'category': 'Sales',
     'license': 'AGPL-3',
     'author': 'Acsone SA/NV,Odoo Community Association (OCA)',
     'version': '10.0.1.0.0',
--- a/sale_blanket_order/data/ir_cron.xml
+++ b/sale_blanket_order/data/ir_cron.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo noupdate="1">
+
+    <record forcecreate="True" id="expired_blanket_orders_cron"
+            model="ir.cron">
+        <field name="name">Expire Blanket Orders</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="nextcall" eval="(DateTime.now() + relativedelta(hour=00, minute=01, second=0)).strftime('%Y-%m-%d %H:%M:%S')" />
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False"/>
+        <field name="model">sale.blanket.order</field>
+        <field name="function">expire_orders</field>
+        <field name="args">()</field>
+    </record>
+
+</odoo>
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -294,7 +294,7 @@ class BlanketOrderLine(models.Model):
         return max(base_price, final_price)
 
     @api.multi
-    @api.onchange('product_id')
+    @api.onchange('product_id', 'original_qty')
     def onchange_product(self):
         if self.product_id:
             self.product_uom = self.product_id.uom_id.id
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -31,7 +31,8 @@ class BlanketOrder(models.Model):
         'res.partner', string='Partner', readonly=True,
         states={'draft': [('readonly', False)]})
     lines_ids = fields.One2many(
-        'sale.blanket.order.line', 'order_id', string='Order lines')
+        'sale.blanket.order.line', 'order_id', string='Order lines',
+        copy=True)
     pricelist_id = fields.Many2one(
         'product.pricelist', string='Pricelist', required=True, readonly=True,
         states={'draft': [('readonly', False)]})
@@ -40,7 +41,7 @@ class BlanketOrder(models.Model):
     payment_term_id = fields.Many2one(
         'account.payment.term', string='Payment Terms', readonly=True,
         states={'draft': [('readonly', False)]})
-    confirmed = fields.Boolean()
+    confirmed = fields.Boolean(default=False)
     state = fields.Selection(selection=[
         ('draft', 'Draft'),
         ('opened', 'Opened'),
@@ -129,6 +130,13 @@ class BlanketOrder(models.Model):
         if self.partner_id.team_id:
             values['team_id'] = self.partner_id.team_id.id
         self.update(values)
+
+    @api.multi
+    def copy_data(self, default=None):
+        if default is None:
+            default = {}
+        default.update(self.default_get(['name', 'confirmed']))
+        return super(BlanketOrder, self).copy_data(default)
 
     @api.multi
     def _validate(self):
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -1,0 +1,321 @@
+# coding: utf-8
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models, api, _
+from odoo.exceptions import Warning
+from odoo.tools import float_is_zero
+
+
+class BlanketOrder(models.Model):
+    _name = 'sale.blanket.order'
+    _inherit = ['mail.thread', 'ir.needaction_mixin']
+    _description = 'Blanket Order'
+
+    @api.model
+    def _get_default_team(self):
+        return self.env['crm.team']._get_default_team_id()
+
+    @api.model
+    def _default_currency(self):
+        return self.env.user.company_id.currency_id
+
+    @api.model
+    def _default_company(self):
+        return self.env.user.company_id
+
+    name = fields.Char(
+        default='Draft',
+        readonly=True
+    )
+    partner_id = fields.Many2one(
+        'res.partner', string='Partner', readonly=True,
+        states={'draft': [('readonly', False)]})
+    lines_ids = fields.One2many(
+        'sale.blanket.order.line', 'order_id', string='Order lines')
+    pricelist_id = fields.Many2one(
+        'product.pricelist', string='Pricelist', required=True, readonly=True,
+        states={'draft': [('readonly', False)]})
+    currency_id = fields.Many2one(
+        'res.currency', related='pricelist_id.currency_id', readonly=True)
+    payment_term_id = fields.Many2one(
+        'account.payment.term', string='Payment Terms', readonly=True,
+        states={'draft': [('readonly', False)]})
+    confirmed = fields.Boolean()
+    state = fields.Selection(selection=[
+        ('draft', 'Draft'),
+        ('opened', 'Opened'),
+        ('expired', 'Expired'),
+    ], compute='_compute_state', store=True)
+    validity_date = fields.Date(
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    client_order_ref = fields.Char(
+        string='Customer Reference', copy=False, readonly=True,
+        states={'draft': [('readonly', False)]})
+    note = fields.Text(
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    user_id = fields.Many2one(
+        'res.users', string='Salesperson', readonly=True,
+        states={'draft': [('readonly', False)]})
+    team_id = fields.Many2one(
+        'crm.team', string='Sales Team', change_default=True,
+        default=_get_default_team, readonly=True,
+        states={'draft': [('readonly', False)]})
+    company_id = fields.Many2one(
+        'res.company', string='Company', default=_default_company,
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    sale_count = fields.Integer(compute='_compute_sale_count')
+
+    @api.multi
+    def _get_sale_orders(self):
+        lines = self.mapped('lines_ids')
+        sale_lines = lines.mapped('sale_order_lines_ids')
+        sale_orders = sale_lines.mapped('order_id')
+        sale_orders_ids = list(set(sale_orders.ids))
+        return self.env['sale.order'].browse(sale_orders_ids)
+
+    @api.multi
+    @api.depends('lines_ids.remaining_qty')
+    def _compute_sale_count(self):
+        for blanket_order in self:
+            blanket_order.sale_count = len(blanket_order._get_sale_orders())
+
+    @api.multi
+    @api.depends(
+        'lines_ids.remaining_qty',
+        'validity_date',
+        'confirmed'
+    )
+    def _compute_state(self):
+        today = fields.Date.today()
+        precision = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure')
+        for order in self:
+            if not order.confirmed:
+                order.state = 'draft'
+            elif order.validity_date <= today:
+                order.state = 'expired'
+            elif float_is_zero(sum(order.lines_ids.mapped('remaining_qty')),
+                               precision_digits=precision):
+                order.state = 'expired'
+            else:
+                order.state = 'opened'
+
+    @api.multi
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        """
+        Update the following fields when the partner is changed:
+        - Pricelist
+        - Payment term
+        """
+        if not self.partner_id:
+            self.payment_term_id = False
+            return
+
+        values = {
+            'pricelist_id': (self.partner_id.property_product_pricelist and
+                             self.partner_id.property_product_pricelist.id or
+                             False),
+            'payment_term_id': (self.partner_id.property_payment_term_id and
+                                self.partner_id.property_payment_term_id.id or
+                                False),
+        }
+
+        if self.partner_id.user_id:
+            values['user_id'] = self.partner_id.user_id.id
+        if self.partner_id.team_id:
+            values['team_id'] = self.partner_id.team_id.id
+        self.update(values)
+
+    @api.multi
+    def _validate(self):
+        try:
+            today = fields.Date.today()
+            for order in self:
+                assert order.validity_date, _("Validity date is mandatory")
+                assert order.validity_date > today, \
+                    _("Validity date must be in the future")
+                assert order.partner_id, _("Partner is mandatory")
+                assert len(order.lines_ids) > 0, _("Must have some lines")
+                order.lines_ids._validate()
+        except AssertionError as e:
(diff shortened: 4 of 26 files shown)


pedrobaeza:
Please avoid plurals on module name.

sale_blanket_orders/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018 Acsone
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Blanket Orders',
+    'category': 'Sale',
+    'license': 'AGPL-3',
+    'author': 'Acsone,Odoo Community Association (OCA)',
+    'version': '11.0.1.0.0',
```
Cedric-Pigeon: Version should be 10

sale_blanket_orders/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018 Acsone
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Blanket Orders',
+    'category': 'Sale',
+    'license': 'AGPL-3',
+    'author': 'Acsone,Odoo Community Association (OCA)',
```
Cedric-Pigeon: Acsone SA/NV

sale_blanket_orders/models/blanket_orders.py
```diff
@@ -0,0 +1,232 @@
...
+        'product.product', string='Product', required=True)
+    product_uom = fields.Many2one(
+        'product.uom', string='Unit of Measure', required=True)
+    price_unit = fields.Float(string='Price', required=True)
+    original_qty = fields.Float(string='Original quantity', required=True)
+    ordered_qty = fields.Float(
+        string='Ordered quantity', compute='_compute_quantities', store=True)
+    invoiced_qty = fields.Float(
+        string='Invoiced quantity', compute='_compute_quantities', store=True)
+    remaining_qty = fields.Float(
```
Cedric-Pigeon: Perhaps you should add the delivered quantities also

sale_blanket_orders/views/blanket_orders.xml
```diff
@@ -0,0 +1,121 @@
...
+                        <h1>
+                            <field name="name" readonly="1"/>
+                        </h1>
+                    </div>
+                    <group>
+                        <group>
+                            <field name="partner_id" domain="[('customer','=',True)]" context="{'search_default_customer':1, 'show_address': 1}" options='{"always_reload": True}'/>
+                        </group>
+                        <group>
+                            <field name="validity_date"/>
```
Cedric-Pigeon: should be set required in draft state through an attrs attribute

sale_blanket_orders/views/blanket_orders.xml
```diff
@@ -0,0 +1,121 @@
...
+                                    <field name="remaining_qty"/>
+                                </tree>
+                            </field>
+                            <field name="note" class="oe_inline" placeholder="Setup default terms and conditions in your company settings."/>
+                            <div class="oe_clear"/>
+                        </page>
+                        <page string="Other Information">
+                            <group>
+                                <group string="Sales Information" name="sales_person">
+                                    <field name="pricelist_id"/>
```
Cedric-Pigeon: should be set required in draft state

Cedric-Pigeon (changes requested):

sale_blanket_orders/views/blanket_orders.xml
andreparames: No problem. I had added it initially, but then thought it could be useful to create the order before the validity date is agreed, so instead I added a check when the user confirms it.
I'll add them back.

sale_blanket_order/data/ir_cron.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo noupdate="0">
```
atchuthan: should be noupdate="1"

atchuthan (approved):
Code review :+1:

andreparames:
Thanks @pedrobaeza @Cedric-Pigeon @atchuthan for the review. The checks are failing because the module depends on `web_action_conditionable`, which I've migrated and is also waiting to be merged: https://github.com/OCA/web/pull/845

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,235 @@
...
+    @api.depends(
+        'sale_order_lines_ids.blanket_line_id',
+        'sale_order_lines_ids.product_uom_qty',
+        'sale_order_lines_ids.qty_delivered',
+        'sale_order_lines_ids.qty_invoiced',
+        'original_qty'
+    )
+    def _compute_quantities(self):
+        for line in self:
+            sale_lines = self.sale_order_lines_ids
```
Cedric-Pigeon: you should use `line.` instead of `self`

Cedric-Pigeon (approved):
Code review and functionnal tests

misern2 (changes requested):
Thanks for the contribution.
Just a minor change:
 if you duplicate the blanket order it keeps the same sequence and state.

andreparames:
Thanks @misern2, I've pushed a commit that fixes that problem, and also lets the lines be copied when duplicating an order.

misern2 (approved):
:+1: LGTM . Thanks!.

andhit-r:
@andreparames thanks for contribution. Few things to consider:

Sequence
-------------

Since sequence will be assigned when blanket order confirmed, it will leave draft BO with **draft** identifier. IMO, it will give user a problem to identify draft BO.

**Propose Solution**

1. If you still want to assign sequence when BO confirmed, make display_name *{ID} like in account.move. Or,
2. Assign sequence when BO saved. With default name value == "/".

State
-------

It always useful to have approval step after confirm

**Propose Solution**

Draft -> Waiting for Approval -> Opened

If you agree my points above, i will create PR for those propose solutions.

Many thanks.

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+        copy=True)
+    pricelist_id = fields.Many2one(
+        'product.pricelist', string='Pricelist', required=True, readonly=True,
+        states={'draft': [('readonly', False)]})
+    currency_id = fields.Many2one(
+        'res.currency', related='pricelist_id.currency_id', readonly=True)
+    payment_term_id = fields.Many2one(
+        'account.payment.term', string='Payment Terms', readonly=True,
+        states={'draft': [('readonly', False)]})
+    confirmed = fields.Boolean(default=False)
```
rousseldenis: That's default

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+        'res.company', string='Company', default=_default_company,
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    sale_count = fields.Integer(compute='_compute_sale_count')
+
+    @api.multi
+    def _get_sale_orders(self):
+        lines = self.mapped('lines_ids')
+        sale_lines = lines.mapped('sale_order_lines_ids')
+        sale_orders = sale_lines.mapped('order_id')
```
rousseldenis: You can fetch them through one mapped('line_ids.sale_order_lines_ids.order_id')

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+        sale_orders_ids = list(set(sale_orders.ids))
+        return self.env['sale.order'].browse(sale_orders_ids)
```
rousseldenis: Return just the result of mapped.

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+
+    @api.multi
+    @api.depends('lines_ids.remaining_qty')
```
rousseldenis, line 81: Is that correct ?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+            today = fields.Date.today()
+            for order in self:
+                assert order.validity_date, _("Validity date is mandatory")
+                assert order.validity_date > today, \
+                    _("Validity date must be in the future")
+                assert order.partner_id, _("Partner is mandatory")
+                assert len(order.lines_ids) > 0, _("Must have some lines")
+                order.lines_ids._validate()
+        except AssertionError as e:
+            raise Warning(e.message)
```
rousseldenis: Deprecated, use UserError

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+        states={'draft': [('readonly', False)]})
+    currency_id = fields.Many2one(
+        'res.currency', related='pricelist_id.currency_id', readonly=True)
+    payment_term_id = fields.Many2one(
+        'account.payment.term', string='Payment Terms', readonly=True,
+        states={'draft': [('readonly', False)]})
+    confirmed = fields.Boolean(default=False)
+    state = fields.Selection(selection=[
+        ('draft', 'Draft'),
+        ('opened', 'Opened'),
```
rousseldenis: Maybe 'open' is more correct

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+    invoiced_qty = fields.Float(
+        string='Invoiced quantity', compute='_compute_quantities', store=True)
+    remaining_qty = fields.Float(
+        string='Remaining quantity', compute='_compute_quantities', store=True)
+    delivered_qty = fields.Float(
+        string='Delivered quantity', compute='_compute_quantities', store=True)
+    sale_order_lines_ids = fields.One2many(
+        'sale.order.line', 'blanket_line_id', string='Sale order lines')
+    company_id = fields.Many2one(
+        'res.company', related='order_id.company_id', store=True)
```
rousseldenis: Readonly=True is recommended for related fields

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,329 @@
...
+    @api.multi
+    def _validate(self):
+        try:
+            for line in self:
+                assert line.price_unit > 0.0, \
+                    _("Price must be greater than zero")
+                assert line.original_qty > 0.0, \
+                    _("Quantity must be greater than zero")
+        except AssertionError as e:
+            raise Warning(e.message)
```
rousseldenis: Deprecated

sale_blanket_order/models/sale_orders.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    blanket_order_id = fields.Many2one(
+        'sale.blanket.order', string='Origin blanket order',
+        related='order_line.blanket_line_id.order_id')
```
rousseldenis: readonly=True

sale_blanket_order/wizard/create_sale_orders.py
```diff
@@ -0,0 +1,111 @@
...
+        self.ensure_one()
+
+        line_obj = self.env['sale.order.line']
+        order_lines = []
+        for line in self.lines_ids:
+            if line.qty == 0.0:
+                continue
+
+            if line.qty > line.remaining_qty:
+                raise Warning(
```
rousseldenis: Deprecated

rousseldenis (changes requested):
Some comments.

One remark : why don't you use decimal precision on quantities ?

sale_blanket_order/models/blanket_orders.py
andreparames: "Explicit is better than implicit" :smile:

andreparames:
Thanks for the review @rousseldenis!

andreparames:
@andhit-r they seem like good suggestions. I'd just add an option to enable the approval step.

rousseldenis (approved):

heathfx:
I stumbled across this module while trolling the web and hope I'm not intruding by commenting here. This module could be very useful at my company and when I was testing it, noticed that the price compute method is likely missing an onchange decorator for when qty is updated.

This is important when pricelists are in use and different price tiers are based on qty.

andreparames:
@heathfx good point, I've made the onchange recalculate the price when the initial quantity changes.

andreparames:
I've pushed an update that adds a configuration option to allow or disallow adding new lines to SOs created from BOs (by default it allows).

sale_blanket_order/views/blanket_orders.xml
```diff
@@ -0,0 +1,122 @@
...
+                        </button>
+                    </div>
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name" readonly="1"/>
+                        </h1>
+                    </div>
+                    <group>
+                        <group>
+                            <field name="partner_id" domain="[('customer','=',True)]" context="{'search_default_customer':1, 'show_address': 1}" options='{"always_reload": True}'/>
```
AdriaGForgeFlow, line 44: I think it should be set to required in draft state to make it clear for the user. In the _validate (blanket_orders.py) function you specify that Partner field is mandatory but that field does not appear in purple color in the web interface and that's a little bit confusing IMO.

bjeficent:
Just a question:
If you Create a blanket order > Create sale order and instead of confirm it you cancel it and then remove it, the ordered quantity of that sale order and the "updated" remaining quantity is shown. Is this how is it supposed to work?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @api.multi
+    @api.depends(
+        'sale_order_lines_ids.blanket_line_id',
+        'sale_order_lines_ids.product_uom_qty',
+        'sale_order_lines_ids.qty_delivered',
+        'sale_order_lines_ids.qty_invoiced',
+        'original_qty'
+    )
+    def _compute_quantities(self):
```
JordiBForgeFlow, line 304: I think that you need to add api.depends('sale_order_lines_ids', 'sale_order_lines_ids.ordered_qty',  'sale_order_lines_ids.invoiced_qty', 'sale_order_lines_ids.delivered_qty'. Otherwise there will always be discrepancies.

JordiBForgeFlow:
@bjeficent You are right. See my comment. Perhaps you can propose the fix.

sale_blanket_order/models/blanket_orders.py
bjeficent: You are right, those lines need to be added but without the sale_order_lines_ids in ordered_qty, invoiced_qty... I mean:
api.depends('sale_order_lines_ids', 'ordered_qty', 'invoiced_qty', 'delivered_qty', 'remaining_qty')

sale_blanket_order/README.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Usage
+=====
+
+A new menu in the Sale area is created, allowing users to create new blanket
+orders.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/11.0
```
nikul-serpentcs, line 22: @andreparames `target` 10.0 :smiley:

sale_blanket_order/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'license': 'AGPL-3',
+    'author': 'Acsone SA/NV,Odoo Community Association (OCA)',
+    'version': '10.0.1.0.0',
+    'website': 'https://github.com/OCA/sale-workflow',
+    'summary': "Blanket Orders",
+    'depends': [
+        'sale',
+        'web_action_conditionable',
+    ],
+    'data': [
```
nikul-serpentcs, line 16: @andreparames Sequence should be [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L29)

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        string='Original quantity', required=True, default=1,
+        digits=dp.get_precision('Product Unit of Measure'))
+    ordered_qty = fields.Float(
+        string='Ordered quantity', compute='_compute_quantities', store=True)
+    invoiced_qty = fields.Float(
+        string='Invoiced quantity', compute='_compute_quantities', store=True)
+    remaining_qty = fields.Float(
+        string='Remaining quantity', compute='_compute_quantities', store=True)
+    delivered_qty = fields.Float(
+        string='Delivered quantity', compute='_compute_quantities', store=True)
```
bjeficent, line 199: I think the store=True parameter should be removed from ordered_qty, invoiced_qty, remaining_qty and delivered_qty to fix that issue.

nikul-serpentcs (commented):
Improve Code

rgl1234:
At the moment we see the following missing features for this functionality:

1.) possibility to have an analytic-account on BO if needed

2.) In our case, the products and quantities can change during the lifetime of a blanket order. So it would be good if the blanket-order could be modified after approving it and after having started the deliveries on a BO (add/remove products, change quantities on products, change validity date...)

does this maybe make sence too for someone else?

René

rgl1234:
At the moment we see the following missing features for this functionality:

1.) possibility to have an analytic-account on BO if needed

2.) In our case, the products and quantities can change during the lifetime of a blanket order. So it would be good if the blanket-order could be modified after approving it and after having started the deliveries on a BO (add/remove products, change quantities on products, change validity date...)

does this maybe make sence too for someone else?

René

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        'confirmed'
+    )
+    def _compute_state(self):
+        today = fields.Date.today()
+        precision = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure')
+        for order in self:
+            if not order.confirmed:
+                order.state = 'draft'
+            elif order.validity_date <= today:
```
daramousk, line 99: If the order's validity date is today, then the order is considered expired? Is the equals needed?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        Update the following fields when the partner is changed:
+        - Pricelist
+        - Payment term
+        """
+        if not self.partner_id:
+            self.payment_term_id = False
+            return
+
+        values = {
+            'pricelist_id': (self.partner_id.property_product_pricelist and
```
daramousk: Is it perhaps possible to just do `self.partner_id.property_product_pricelist.id` since this (despite it's name) is a `Many2one`? Same for the `property_payment_term_id`

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        if self.partner_id.user_id:
+            values['user_id'] = self.partner_id.user_id.id
+        if self.partner_id.team_id:
+            values['team_id'] = self.partner_id.team_id.id
+        self.update(values)
+
+    @api.multi
+    def copy_data(self, default=None):
```
daramousk: From what I understand, here you are getting the default values of the fields `name` and `confirmed` and injecting them on the newly created record.

Is this override necessary though? Because `create()` calls the `_add_missing_default_values()` which in turn calls `default_get` and adds the values later. So removing this would make sense. Or am I missing something here?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        if default is None:
+            default = {}
+        default.update(self.default_get(['name', 'confirmed']))
+        return super(BlanketOrder, self).copy_data(default)
+
+    @api.multi
+    def _validate(self):
```
daramousk, line 131: Maybe this, with it's assertions, could be replaced by `@api.constrains('confirmed')` and conditionals?

So it does not explicitly requires us to call it and it would be called when necessary.
Also, there are certain flags passed to the interpreter which can remove the assertions

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+            name = sequence_obj.next_by_code('sale.blanket.order')
+            order.write({'confirmed': True, 'name': name})
+        return True
+
+    @api.multi
+    def action_view_sale_orders(self):
+        sale_orders = self._get_sale_orders()
+        action = self.env.ref('sale.action_orders').read()[0]
+        if len(sale_orders) > 0:
+            action['domain'] = [('id', 'in', sale_orders.ids)]
```
daramousk, line 161: The `action['domain']` already has a domain which is removed when this assignment takes place. Perhaps it would be better to append it?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+                    force_company=order.company_id.id)
+            name = sequence_obj.next_by_code('sale.blanket.order')
+            order.write({'confirmed': True, 'name': name})
+        return True
+
+    @api.multi
+    def action_view_sale_orders(self):
+        sale_orders = self._get_sale_orders()
+        action = self.env.ref('sale.action_orders').read()[0]
+        if len(sale_orders) > 0:
```
daramousk: If you want to see whether your recordset has any records in it, you can avoid the `len()` altogether since `BaseModel.__nonzero__` is implemented. Therefore `if sale_orders` will suffice.

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+    def _compute_quantities(self):
+        for line in self:
+            sale_lines = line.sale_order_lines_ids
+            line.ordered_qty = sum(l.product_uom_qty for l in sale_lines)
+            line.invoiced_qty = sum(l.qty_invoiced for l in sale_lines)
+            line.delivered_qty = sum(l.qty_delivered for l in sale_lines)
+            line.remaining_qty = line.original_qty - line.ordered_qty
+
+    @api.multi
+    def _validate(self):
```
daramousk, line 313: Same here, could this be perhaps replaced by `api.constrains()` and conditions instead of assert?

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+    delivered_qty = fields.Float(
+        string='Delivered quantity', compute='_compute_quantities', store=True)
+    sale_order_lines_ids = fields.One2many(
+        'sale.order.line', 'blanket_line_id', string='Sale order lines')
+    company_id = fields.Many2one(
+        'res.company', related='order_id.company_id', store=True,
+        readonly=True)
+
+    def _get_real_price_currency(self, product, rule_id, qty, uom,
+                                 pricelist_id):
```
daramousk, line 207: `pricelist_id` seems to not be used anywhere here.

sale_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,331 @@
...
+        """Retrieve the price before applying the pricelist
+            :param obj product: object of current product record
+            :parem float qty: total quentity of product
+            :param tuple price_and_rule: tuple(price, suitable_rule) coming
```
daramousk: Perhaps `price_and_rule` should be renamed to `rule_id`?

daramousk (commented):
Hello and thank you for this, looks great.
I have left some comments mostly about the assertions.

Let me know what you think.

rafaelbn:
Hello @andreparames , please could you check bots are green :green_heart: ? I see travis red :red_circle:

rousseldenis:
Hi @rafaelbn, @andreparames does not wor k anymore on Odoo. I will try to finish this one.

rafaelbn:
Hi @rousseldenis , if you cannot work on this we can close this PR

sale_blanket_order/models/blanket_orders.py
rousseldenis: I'll modify asserts but this cannot be constrains because it's called on order validation (action)

rousseldenis: And finally I think I will let asserts because I like the way it was done (catching AssertionError and raise an Odoo Exception)

sale_blanket_order/models/blanket_orders.py
rousseldenis: See here under message

sale_blanket_order/models/blanket_orders.py
rousseldenis: A simple copy=False will be sufficent

sale_blanket_order/models/blanket_orders.py
rousseldenis: @bjeficent Why removing store=True ?

sale_blanket_order/models/blanket_orders.py
rousseldenis: Yes but can be overriden

sale_blanket_order/views/blanket_orders.xml
rousseldenis: @Cedric-Pigeon What do you think of this ?

rousseldenis:
@rafaelbn @daramousk Remarks that seems relevant to me has been attended. Waiting travis but ok for me.

sale_blanket_order/views/blanket_orders.xml
Cedric-Pigeon: it makes sense IMO

Cedric-Pigeon:
@rafaelbn This one is ready to be merged.

PR #596, opened 2018-01-23
[FIX] depends is called on recordsets

SimoRubi:
Method decorated with depends is called on recordsets


LoisRForgeFlow (approved):
LGTM :+1: 

would be great if you take the occasion to fix some of the pylint errors from this module. Of course not blocking.

SimoRubi:
@lreficent minor fixes done

yajo (approved):

pedrobaeza (approved):

PR #600, opened 2018-01-30
11.0 mig sale order line date

OpenCode:


sale_order_line_date/__manifest__.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
eLBati: @OpenCode why removing everywhere the ```coding``` line?

OpenCode: https://github.com/OCA/account-invoicing/pull/341#discussion_r160036766

eLBati: ok

eLBati (approved):

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    requested_date = fields.Datetime()
```
JayVora-SerpentCS, line 14: Good to add string to it

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+                })
+        return super(SaleOrderLine, self).write(vals)
+
+    @api.model
+    def create(self, vals):
+        res = super(SaleOrderLine, self).create(vals)
+        if res.order_id.requested_date and not res.requested_date:
+            res.write({'requested_date': res.order_id.requested_date})
+        return res
+
```
JayVora-SerpentCS, line 32: Rather than overriding create() and write(), cannot we make an onchange / depends(order_id.requested_date), and do the rest?

sale_order_line_date/models/sale_order_line.py
OpenCode: "requested_date" is converted automatically in "Requested Date" by Odoo. Why add a string?

JayVora-SerpentCS: True that, I realised later.

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.multi
+    def _prepare_order_line_procurement(self, group_id=False):
```
LoisRForgeFlow: This method's name is different in v11. change to:
`def _prepare_procurement_values(self, group_id=False):`

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+        self.ensure_one()
```
LoisRForgeFlow: This is not needed as it already present on the parent method.

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+        vals = super(SaleOrderLine, self).\
+            _prepare_order_line_procurement(group_id)
```
LoisRForgeFlow: change accordingly to above comment

LoisRForgeFlow (changes requested):
Thanks for the migration. See my comments inline.

LoisRForgeFlow:
@OpenCode  I really needed this fixed so I've just proposed you a PR.

mreficent:
@lreficent I think you may superseed this PR.

OpenCode:
@lreficent Done

JayVora-SerpentCS (approved):

MiquelRForgeFlow (approved):

PR #604, opened 2018-02-07
[8.0] Fix missing move_in in sale_rental when creating sale extension

sergiocorato:
Using sale rental extension, move_in is not get from compute function.
This mp resolve this bug.


rafaelbn:
Please @sergiocorato review travis and if possible add some test thanks!

sergiocorato:
thanks @rafaelbn , I think there is a problem in the repo or travis build

PR #609, opened 2018-02-19
[10.0] Remove module sale_partner_order_policy

alexis-via:
Remove module sale_partner_order_policy because this module doesn't make in v9+ versions.


pedrobaeza (approved):
Fast-tracking this

PR #610, opened 2018-02-20
[10.0][FIX][sale_order_lot_selection] If res contain action we should return and show it

ghost:


atchuthan (approved):

sale_order_lot_selection/model/sale.py
```diff
@@ -72,6 +72,7 @@ def _check_move_state(self, line):
     @api.multi
     def action_confirm(self):
         res = super(SaleOrder, self).action_confirm()
-        for line in self.order_line:
-            self._check_move_state(line)
+        if res is True:
+            for line in self.order_line:
```
atchuthan: self may contain multiple records.
So, always loop order the records i.e. for order in self:

atchuthan (changes requested):

sale_order_lot_selection/model/sale.py
```diff
@@ -72,6 +72,7 @@ def _check_move_state(self, line):
     @api.multi
     def action_confirm(self):
         res = super(SaleOrder, self).action_confirm()
```
guewen, line 74: Would have been a shorter diff to write:

```python
if not res:
    return res
```

And avoided an indentation level as well.

guewen (dismissed):

sale_order_lot_selection/model/sale.py
guewen: Actually what is the reason to check `res`? It seems safer to me to actually check the `state` of the sales orders and call `_check_move_state` only if it's the correct state.

guewen (changes requested):

sale_order_lot_selection/model/sale.py
ghost: in `sale-workflow/sale_exception/models/sale.py`
```
    @api.multi
    def action_confirm(self):
        if self.detect_exceptions():
        else:
            return super(SaleOrder, self).action_confirm()
```
in some case return action `dict` and don't confirm SO.
if SO not in correct state it can cause othe error that overwrite `res` results by `raise` some exception

ghost: about
```
if not res:
    return res
```
we need to check if `res` contain `True` (is the instance of `bool` and  value is `True`)
```
z = {'a': 1}
if z: print(bool(z))
True
if not z: print(bool(z))

```
not empty `dict` evaluate to `True`
so it not what we want.
we can do something like this
```
if instanceof(res, dict):
    return res
```
but it also not good idea

rousseldenis:
Closing as very old. Feel free to reopen it if needed.

PR #611, opened 2018-02-20
11.0 imp sale product set variant

mpanarin:
Made prodcut_variants use many2many_tags widget in product set variant as it is more usable than tree view.
If there is only one variant for the product, it will be auto-filled in wizard

### DIFF
--- a/sale_product_set_variant/wizard/product_set_add.py
+++ b/sale_product_set_variant/wizard/product_set_add.py
@@ -17,7 +17,12 @@ class ProductSetAddLine(models.TransientModel):
         'product.set.line',
     )
     product_variant_ids = fields.Many2many(
-        'product.product', domain=[('sale_ok', '=', True)],
+        'product.product',
+        domain="""[
+            '&',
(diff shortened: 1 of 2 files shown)


sale_product_set_variant/wizard/product_set_add.py
```diff
@@ -96,12 +110,27 @@ def add_set(self):
...
-                raise UserError(_("Please select the "
-                                  "appropriate product variants"))
+                variants = set_line.product_template_id.product_variant_ids
+                if len(variants) == 1:
+                    set_line.product_variant_ids = [(6, 0, variants.ids)]
+                else:
+                    raise UserError(
+                        _("Please select the "
+                          "appropriate product variants "
+                          "for product {}".format(
```
guewen: The translations `_()` should close right after the closing `"`, here you are translating the text *after* the interpolation.

guewen (changes requested):

yvaucher (approved):

guewen (approved):

PR #613, opened 2018-02-27
[10.0][FIX] sale_order_invoicing_finished_task

jcoux:
If we have one line "to invoice", we also want that the sale order is "to invoice".

And we don't need to keep the depends added on `order_line.task_ids.invoiceable`,
because the core depends on `order_line.invoice_status` is enought to trigger correctly the computation.

### DIFF
--- a/sale_order_invoicing_finished_task/models/sale_order.py
+++ b/sale_order_invoicing_finished_task/models/sale_order.py
@@ -6,20 +6,6 @@
 from odoo import api, fields, models
 
 
-class SaleOrder(models.Model):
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM

pedrobaeza (approved):
@sergio-teruel can you review this?

carlosdauden:
Fixed in https://github.com/OCA/sale-workflow/pull/618

pedrobaeza:
OK, it seems that the other patch is more sensible with the features of the module, so I close this one and fast-track the other for fixing this.

PR #615, opened 2018-03-02
[11.0][ADD] Module which add a redirection button to the project from the SO

benwillig:
Simply adds a button on the sale order to be redirected on the linked projects.
Linked projects are:

- Projects linked to order's analytic account
- Projects linked to order line's tasks
- Projects linked to the order line's product


sale_timesheet_project_link/models/sale_order.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @api.multi
+    @api.depends(
+        'order_line.task_id.project_id',
+        'project_ids',
+    )
+    def _compute_all_projects(self):
+        for rec in self:
+            rec.all_project_ids = (
+                rec.order_line.mapped('task_id.project_id') | rec.project_ids)
```
sbidoul: Is a computed field the best way to do this? A simple method would be less stressful for the orm cache.

sale_timesheet_project_link/README.rst
```diff
@@ -0,0 +1,67 @@
...
+===========================
+Sale Timesheet Project Link
+===========================
+
+Simply adds a button on the sale order to be redirected on the linked projects.
+Linked projects are:
+
+#. Projects linked to order's analytic account
+#. Projects linked to order line's tasks
+#. Projects linked to the order line's product
```
sbidoul, line 14: I don't see this last item in _compute_all_projects, is it obtained indirectly?

benwillig: The last item is given by the ```project_id``` field available in the ```sale_timesheet``` module.

sale_timesheet_project_link/models/sale_order.py
benwillig: Yes I don't really need this field, it is only used when counting the number of projects to show the button in the form view. I will change it for a simple method.

rafaelbn:
Thanks @benwillig ! Please could you review travis?

benwillig:
The travis error is due to the ```sale_automatic_workflow``` module and this is not caused by this PR. Unless I missed something.

pedrobaeza:
Main branch is green. Please rebase to see if it helps.

benwillig:
Indeed, everything is green now.

rousseldenis (approved):
Code review

PR #616, opened 2018-03-02
[11.0][ADD] Module which block update and deletion of an invoiced analytic line

benwillig:


pedrobaeza:
I think you can simply call it `sale_timesheet_lock`

benwillig:
I think it's more explicit with ```_invoiced```, because you can lock lot of things which were implemented in ```sale_timesheet``` module

pedrobaeza:
OK about being explicit, but I still don't see the `_invoiced` part related to what you describe, but I admit I don't have a better one, so go on!

benwillig:
The purpose of this module is to lock analytic lines linked to a SO which were invoiced. So think it would be better with something like ```sale_timesheet_lock_invoiced_analytic_line``` then. But I'm not sure that having modules with long names is a good habit. What do you think?

pedrobaeza:
Yeah, we can have problems with that large name in some identifiers. OK, let's keep the same name as the best option, and thinking twice, it's defining enough. Thanks for the explanations and the brainstorm about this!

sale_timesheet_lock_invoiced/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Sale Timesheet Lock Invoiced
+============================
+
+Forbid to update/delete a timesheet line if it has been invoiced.
+*sale_timesheet* module only forbid to edit some fields. This module will block
```
sbidoul: s/forbid/forbids/

sale_timesheet_lock_invoiced/models/account_analytic_line.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @api.multi
+    def unlink(self):
+        self._check_locked_lines()
+        return super(AccountAnalyticLine, self).unlink()
+
+    @api.multi
+    def _check_locked_lines(self):
+        for rec in self:
+            if rec.task_id and rec.timesheet_invoice_id:
```
sbidoul: 2 questions:
- why the test on rec.task_id?
- what happens when one deletes a draft invoice linked to a timesheet line?

benwillig: 1) About ```rec.task_id```, it's because the combination of ```task_id``` and ```timesheet_invoice_id``` tells that this line is linked to a sale order line. And we want to lock thoses lines, but indeed, it makes more sense to check ```so_line``` field. I will change it. But then I'm wondering if I should only check ```timesheet_invoice_id``` field.

2) The ```timesheet_invoice_id``` field is set when the invoice is being opened. So we can't delete this invoice.

rousseldenis (approved):
Code review

PR #617, opened 2018-03-05
[FIX] Fixed tests for sale_automatic_workflow

hugosantosred:
With this commit in odoo-server https://github.com/OCA/OCB/commit/7ed86b53cdfbab9a7867de328604850138692caf the moves that where force-assigned are not transferred in the picking, so in order to fix test is necessary to add stock to the stockable products in test sale order.


pedrobaeza (approved):
If green

hugosantosred:
@pedrobaeza Test are green now.

Any other changes requested?

pedrobaeza:
OK, let's fast track it for greenifying the branch.

PR #618, opened 2018-03-10
[FIX] sale_order_invoicing_finished_task: missed picking_policy condition

carlosdauden:
Before this PR:
sale order invoice_status allways 'no'

After this PR:
Evaluate condition to set invoice_status = 'no'
Invoiceable button only visible if invoicing_finished_task product

@Tecnativa


pedrobaeza (approved):
Fast-tracking as mentioned in the other PR

Issue #622, opened 2018-04-02
sale_rental 'must_have_dates' field trigger warning 


bealdav:
Impacted versions: v10

### Steps to reproduce:
Trigger test on sale_rental: `-u sale_rental --test-enable`

### Current behavior:
` WARNING db odoo.models: method sale.order.line._check_sale_line_rental: @constrains parameter 'must_have_dates' is not writeable`

see https://runbot2.odoo-community.org/runbot/static/build/3321760-516-828b35/logs/job_20_test_all.txt

https://github.com/OCA/sale-workflow/pull/516

It breaks other PR

must_have_dates is a related field on a sale.order.line from product.product:
defined in sale_start_end_dates module

In sale_rental the method:
```    
    @api.one
    @api.constrains(
        'rental_type', 'extension_rental_id', 'start_date', 'end_date',
        'rental_qty', 'product_uom_qty', 'product_id', 'must_have_dates')
    def _check_sale_line_rental(self):
```
must_have_dates field is defined in constraints but there is no control on this field because it's a related one and the setting on the product can change over the time.

### Expected behavior:
No warning

### Paths to fix it

- only remove must_have_dates from argument of the decorator constrains
- use sale_exception modules to prevent to confirm sale instead of check inside a constraint if must_have_dates is checked.
- other ideas ....

Ping experts @alexis-via @dvdhinesh. Thanks to share your opinion

PR #624, opened 2018-04-04
[10.0] sale_rental: fix warning @constrains parameter 'must_have_dates' is not writeable

alexis-via:


bealdav (approved):
fix https://github.com/OCA/sale-workflow/issues/622 Thanks

hparfr (approved):

PR #542, opened 2017-10-09
[10.0] [MIG] product_customer_code_sale

nikul-serpentcs:
Product Customer Code For sale
==============================
Based on product_supplierinfo_for_customer, this module loads in every sale order the
customer code defined in the product.

PR #625, opened 2018-04-04
[10.0] [MIG] product_supplierinfo_for_customer_sale

AaronHForgeFlow:
supersedes #542


rafaelbn:
Thanks ! @aheficent ! Please could you review runbot and travis? :smile:

AaronHForgeFlow:
@rafaelbn sure. There's a missing dependency here: https://github.com/OCA/product-attribute/pull/292

tafaRU:
@aheficent, is intended to also add the module `product_customer_code_sale`?

AaronHForgeFlow:
@tafaRU No, because product_customer_code was deprecated in favour of product_supplierinfo_for_customer. Discussed here: https://github.com/OCA/product-attribute/pull/236

rafaelbn:
@aheficent I cannot test it! I need bots :green_heart: :smiley:

MiquelRForgeFlow:
@rafaelbn you can test now in runbot 😃

rafaelbn:
Hi @mreficent have you tried? Try to use this module in runbot. It's impossible because there are surely other modules that breaks sale order... :cry: 

I just trust you :+1:

AaronHForgeFlow:
Now the error on runbot shows as soon as log in. It's hard to remove :face_with_head_bandage:

AaronHForgeFlow:
Also, some tests are needed here...

AaronHForgeFlow:
I think we have to remove the priority, that will affect the sale_generator module but there's no other way and this is a blocking issue.

MiquelRForgeFlow:
@rafaelbn you can try now. After some iterations, finally I amended `sale_generator` as @aheficent said.

mreficent:
- [ ] Depends of https://github.com/OCA/product-attribute/pull/363.

MiquelRForgeFlow:
:green_apple:

HaraldPanten:
Hi, I've been trying the module but I couldn't replicate what I need to.

I need to show the product name/code for different partners in the sale order. To do this, I:

1.1- Created a product --> PRO1

1.2- Created a Partner --> PART1

1.3- I put in sales tab for PRO1 a customer name for PART1. The name was PROPART1.

1.4- I put in sales tab for PRO1 a customer code for PART1. The name was PROCODE1.

1.5. I put a price and a minimal quantity, for example, 20€ and 5 units.

1.6. I created a pricelist based on "Partner Prices" for PRO1, called PARTNER1.

After this steps, I created a basic quotation with:

2.1- Customer : PART1

2.2- Pricelist: PARTNER1

2.3- Line 1: PRO1

I detected some issues:

3.1- Product Customer Code is not displayed. It's empty.

3.2- Description: The description that I put in PRO1 for this customer is not being shown.

3.3- Ordered Qty: The minimal Qty that I selected for this product/customer is not being shown, it appears just "1".

I think that points 3.1, 3.2 and 3.3 should be solved. Am I missing any step to make it work in the correct way? 

Thanks!

AaronHForgeFlow:
Thank you for reviewing @HaraldPanten I'll check

AaronHForgeFlow:
still issues sale-workflow/10.0 cannot create So line on runbot:
`Uncaught Error: QWeb2 - template['ListView.rows']: Runtime Error: Error: QWeb2 - template['ListView.row']: Runtime Error: Error: Unknown field is_template in domain [["is_template","=",true]]`
:scream:

JordiBForgeFlow (changes requested):
Only users in group 'sale.group_pricelist_item' should be able to view prices in the customer supplier info.

AaronHForgeFlow:
@jbeficent this is supposed to be fixed here: https://github.com/OCA/product-attribute/pull/368

JordiBForgeFlow (approved):

MiquelRForgeFlow:
@HaraldPanten I think your points have been attended.

I think this PR is ready. More reviews are welcome

AaronHForgeFlow (commented):
Not working on runbot. The minimum qty from supplier info is not passed to the sale order line. The product customer code is also empty. @mreficent can you check?

HaraldPanten:
Hi @mreficent , 

@aheficent is right. It's still not moving the minimum quantity and the customer code to the sales order.

TY!

MiquelRForgeFlow:
I think I will add a test to show that it works. You maybe are doing something wrong. For example, you don't have to put a min qty in the pricelist. If you put, odoo checks `qty_in_product_uom < rule.min_quantity` and if true, it will compute prices as 'Public Price' instead.

AaronHForgeFlow:
@mreficent Yes, I think you're right this is standard Odoo behavior. Putting a minimum qty means the system will always use the list price. It seems really hard to overwrite this behavior. At least, can you put this in the readme as a known issue?

MiquelRForgeFlow:
full :green_apple: 

@HaraldPanten @aheficent now the tests show the correct use of the module :)

HaraldPanten:
Hi @mreficent ,

All the functionalities are working fine except for one:

Even if I define a "customer name" for a product, the sales order is still showing the original name. The "customer name" should be shown in the "description" field. Souldn't it?

Sorry for my persistence!

MiquelRForgeFlow:
@HaraldPanten well, that was not an expected functionality of the module but I added now because it makes sense.

AaronHForgeFlow (commented):
:+1: Thank for fixing this PR. It works like a charm now! Even the minimal qty is passed.

MiquelRForgeFlow (approved):

HaraldPanten (approved):
@mreficent , it's working perfectly fine for me. Thank you so much! :+1:

rafaelbn:
Please, take a look in : https://github.com/OCA/sale-workflow/issues/737

PR #633, opened 2018-04-13
[11.0][ADD] sale_order_multiple_discount

espo-tony:
This module allows to express discounts on SO lines and invoice lines as mathematical expressions


pedrobaeza:
Do you know that we already have `account_invoice_triple_discount` and `sale_order_triple_discount`? It's for the same purpose and well tested across 3 versions.

espo-tony:
Hi @pedrobaeza,

Yes, I know that those modules exist and I used them myself in the past.
Nevertheless, I think that the module proposed in this PR is more generic and could maybe cover some cases not covered by the other modules. That's because it allows you to set on a single line as many discounts as you wish, overcoming the limit of 3 discounts.

I reckon that the limit of 3 discounts could be considered negligible for most practical cases, but I have experience of scenarios where that's not the case.

pedrobaeza:
Well, then this module should be adapted to the same approach:

* Rename the module to `sale_order_discount_formula`.
* Extract the invoice part in a `account_invoice_discount_formula` for being able to build the corresponding purchase part without needing to duplicate invoice part.

espo-tony:
@pedrobaeza ,

I renamed the module and separated the Sale Order part from the Account Invoice part (which is moved to a different PR: oca/account-invoicing#373 )

espo-tony:
Runbot, obviously, doesn't work because module `account_invoice_discount_formula`, from which this module depends, is not merged yet.

pedrobaeza:
You can temporarily put in oca_dependencies.txt your fork and branch for getting the dependency.

sale_order_discount_formula/models/sale_order.py
```diff
@@ -0,0 +1,104 @@
...
+    @api.onchange('multiple_discount')
+    def onchange_multiple_discount(self):
+        def _normalize_discount(discount):
+            discount = discount.replace(" ", "")
+            discount = discount.replace(",", ".")
+            if discount and discount[0] == '+':
+                discount = discount[1:]
+            return discount
+
+        for line in self:
```
hieulucky111, line 51: Do we need to put the loop in onchange function?

sale_order_discount_formula/models/sale_order.py
```diff
@@ -0,0 +1,104 @@
...
+                    line.multiple_discount):
+                raise ValidationError(
+                    _('Warning! The discount format is not recognized.'))
+
+    @api.multi
+    def write(self, vals):
+        res = super(SaleOrderLine, self).write(vals)
+        if 'multiple_discount' in vals:
+            for line in self:
+                line.onchange_multiple_discount()
```
hieulucky111, line 95: Do we need to call onchange again?

sale_order_discount_formula/models/sale_order.py
```diff
@@ -0,0 +1,104 @@
...
+        return res
+
+    @api.multi
+    def _prepare_invoice_line(self, qty):
+        self.ensure_one()
+        res = super(SaleOrderLine, self)._prepare_invoice_line(qty)
+        res['multiple_discount'] = self.multiple_discount
```
hieulucky111, line 102: why do you put `multiple_discount` in here?

sale_order_discount_formula/models/sale_order.py
espo-tony: It's maybe an overkill, but even if an onchange is usually invoked on a recordset containing a single record, it's still formally possible to have it invoked on a recordset made of multiple records. For safety sake I prefer to iterate over the recordset.

sale_order_discount_formula/models/sale_order.py
espo-tony: As an onchange method is only triggered when a change is made from the View, it wouldn't be invoked on direct write, therefore I think it better to enforce it here.

sale_order_discount_formula/models/sale_order.py
espo-tony: Because _multiple_discount_ is also an attribute of _account.invoice.line_ that needs to be propagated from _sale.order.line_ when this method is invoked.

astirpe:
@espo-tony even if the dependency (account_invoice_discount_formula) is not yet merged, in this PR you can temporary add a line in the `oca_dependencies.txt` file, in order to make travis and runbot working.

Add the following line:
```
account-invoicing https://github.com/espo-tony/account-invoicing 11_add_account_invoice_discount_formula
```

sale_order_discount_formula/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2018 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': "Sale Order Discount Formula",
+    'summary': "Express discounts on SO lines as mathematical formula",
+    'author': 'Onestein, Odoo Community Association (OCA)',
+    'website': 'http://www.onestein.eu',
```
MeetKD: Can you please change website key to `https://github.com/OCA/sale-workflow`

sale_order_discount_formula/models/sale_order.py
```diff
@@ -0,0 +1,104 @@
...
+                        numeric_tokens.append(float(token)*last_sign)
+                marginal_discount = 1
+                for token in numeric_tokens:
+                    marginal_discount = marginal_discount * (1-(token/100))
+                total_discount = 1 - marginal_discount
+                line.discount = total_discount * 100
+
+                if normalized_discount != line.multiple_discount:
+                    line.multiple_discount = normalized_discount
+
```
MeetKD: remove space.

angelmoya:
Hi, I functionally tested module, and there is a problem.

If you use pricelist and set 'Show public price & discount to the customer' on pricelist, then you can not see discount related to pricelist.

We fixed (I will push) creating an onchange on discount field.

But... would be nice to do:
- Public pricelist:
- Pricelist 20% - that add discount 20%
- Pricelist 20% + 10% - that add 10% discount to Pricelist 20%

Now, with our patch, on sale order with last pricelist you get 28% discount, we are working to get 20+10 discount.

espo-tony:
@angelmoya Thanks, you're perfectly right indeed. I completely forgot about the 'Show public price & discount to the customer' option on pricelists.
In my last commit I addressed the issue in a way that I suppose is very similar to what you suggested (an onchange on discount, in addition to an extension of the onchange on product).
Still, the pricelist discount is shown as a whole, not automatically splitted in its components. I'm not sure if this should be within the scope of this PR, as I rather believe it should be the object of a new PR for a new module depending on this one. What do you think?

astirpe:
@espo-tony could you resolve the oca_dependencies.txt conflicts ?

sale_order_discount_formula/models/sale_order.py
```diff
@@ -0,0 +1,103 @@
...
+            return discount
+
+        for line in self:
+            if line.multiple_discount:
+                if self._validate_discount(line.multiple_discount):
+                    normalized_discount = _normalize_discount(
+                        line.multiple_discount)
+                else:
+                    line.discount = 0
+                    raise Warning(
```
nikul-serpentcs, line 58: @espo-tony Here odoo provided  `UserError` so use this, because `Warning` is a Python keyword.

sale_order_discount_formula/views/sale_order_view.xml
```diff
@@ -0,0 +1,15 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='order_line']/tree//field[@name='discount']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='order_line']/tree//field[@name='discount']" position="after">
+                <field name="multiple_discount" groups="sale.group_discount_per_so_line"/>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
nikul-serpentcs, line 15: add new blank line

sale_order_discount_formula/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2018 Onestein (<http://www.onestein.eu>)
```
nikul-serpentcs, line 1: Remove Copyright from all `__init__` file

nikul-serpentcs (commented):
Improve Code

robotiko:
> Hi, I functionally tested module, and there is a problem.
> 
> If you use pricelist and set 'Show public price & discount to the customer' on pricelist, then you can not see discount related to pricelist.
> 
> We fixed (I will push) creating an onchange on discount field.
> 
> But... would be nice to do:
> 
> * Public pricelist:
> * Pricelist 20% - that add discount 20%
> * Pricelist 20% + 10% - that add 10% discount to Pricelist 20%
> 
> Now, with our patch, on sale order with last pricelist you get 28% discount, we are working to get 20+10 discount.

Hi @angelmoya,  is the proposed patch available in a  public repo?
thanks

rousseldenis:
@espo-tony Do you plan to finish this ?

rousseldenis:
Closing this as very old and no answer from author. Please feel free to reopen it if needed.

PR #639, opened 2018-04-19
[10.0][ADD]sale_returned_quantity

AaronHForgeFlow:
Control the quantity returned in the sale order lines. The purpose is not to use the tool as an RMA but correct mistakes after validating a shipment:

- Show the returning quantity in the sale order lines so the user knows what have been really shipped.
- Allows to manual recreate a procurement in the order. The purpose of this feature is to allow users to correct wrong validated shipments. When that occurs they would be able to return the items and manually procure the products they need.

Pending to include tests


AaronHForgeFlow:
Closing this module as stock_cancel does the thing. Correct errors when validating. I'll keep the branch just in case though: https://github.com/OCA/stock-logistics-workflow/pull/418

PR #628, opened 2018-04-06
11.0 mig sale order lot selection

OpenCode:

PR #643, opened 2018-05-04
[PORT] sale_order_lot_selection to v11

andriisem:
PORT Sale Order Lot Selection to v11


pedrobaeza:
There's already other migration in #628. Please review that one.

PR #645, opened 2018-05-10
[10.0][sale_cancel_reason]Added description field

paolomonniLP:
Added a text field for the description of the cancellation of the order.

### DIFF
--- a/sale_cancel_reason/model/sale.py
+++ b/sale_cancel_reason/model/sale.py
@@ -7,12 +7,10 @@ from odoo import models, fields
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
 
-    cancel_reason_id = fields.Many2one(
-        'sale.order.cancel.reason',
-        string="Reason for cancellation",
-        readonly=True,
+    cancel_reason_id = fields.Many2one('sale.order.cancel.reason', string="Reason for cancellation", readonly=True,
         ondelete="restrict")
-
+    description = fields.Text(string="Description", readonly=True,
+        ondelete="restrict")
 
 class SaleOrderCancelReason(models.Model):
     _name = 'sale.order.cancel.reason'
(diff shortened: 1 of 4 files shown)


sale_cancel_reason/wizard/cancel_reason.py
```diff
@@ -14,6 +14,7 @@ class SaleOrderCancel(models.TransientModel):
         'sale.order.cancel.reason',
         string='Reason',
         required=True)
+    description = fields.Text(string='Description', size=12)
```
pedrobaeza: You put a text field but then limited to 12 characters. Isn't this a nonsense?

pedrobaeza (changes requested):
What is the use case for 2 strings for the reason?

paolomonniLP:
This field is used to describe in detail the reason for canceling the order.

pedrobaeza:
Uhm, I think you don't need too much explanations about cancel reasons (they are internal), but I let other reviewers to decide.

JayVora-SerpentCS:
Agree with @pedrobaeza .Simple M2O is enough. Always minimise fields. If the reason is different, it should be mentioned in a comment may be. Or choose OTHER as a reason on sale.order.cancel.reason

sale_cancel_reason/model/sale.py
```diff
@@ -7,12 +7,10 @@
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
 
-    cancel_reason_id = fields.Many2one(
-        'sale.order.cancel.reason',
-        string="Reason for cancellation",
-        readonly=True,
+    cancel_reason_id = fields.Many2one('sale.order.cancel.reason', string="Reason for cancellation", readonly=True,
```
JayVora-SerpentCS, line 10: You made travis go red forcefully:) Was right earlier.

JayVora-SerpentCS:
ping @paolomonniLP

rousseldenis:
Closing this as very old and comments not attended. Please feel free to reopen if needed.

PR #647, opened 2018-05-11
[FIX][10.0] sale_order_type: avoid error on invoices if user is only …

cubells:
…acountant.

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking when CI finishes because this is important.

rafaelbn:
:+1:

PR #649, opened 2018-05-16
11.0 mig sale force invoiced

andreasdsp:


pedrobaeza (approved):

sale_force_invoiced/README.rst
```diff
@@ -0,0 +1,77 @@
...
+#. Create a sales order and confirm it.
+#. Deliver the products/services.
+#. Create an invoice and reduce the invoiced quantity. The sales order
+   invoicing status is 'To Invoice'.
+#. Lock the Sale Order, to change the status of it to 'Done'.
+#. Check the field 'Force Invoiced'
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/10.0
```
RoelAdriaans: Links to old runbot, update to 11.0

sale_force_invoiced/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
```
RoelAdriaans: Remove al coding: utf-8 lines, no longer needed with python 3?

RoelAdriaans (changes requested):
Functional test :+1:

andreasdsp:
Thank you for the review. i had remove utf-8 code and change runbot url.

sale_force_invoiced/README.rst
andreasdsp: done

sale_force_invoiced/__init__.py
andreasdsp: done

RoelAdriaans (approved):

huyly0909 (approved):

hieulucky111 (approved):

novawish (approved):

DaoTheKien (approved):

etobella (approved):
LGTM :+1:

AaronHForgeFlow:
ready to merge

RoelAdriaans:
@pedrobaeza I think this can be merged

sale_force_invoiced/view/sale_view.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
```
MiquelRForgeFlow: I think this change should be reverted

MiquelRForgeFlow (changes requested):

sale_force_invoiced/view/sale_view.xml
RoelAdriaans: @mreficent This header is no longer required, assumed is UTF-8. 
See odoo documentation, there is the header is missing: 
https://www.odoo.com/documentation/11.0/reference/data.html
In the newer xml files in default Odoo you can also see that this line is absent.

MiquelRForgeFlow: But your example is not valid. In https://www.odoo.com/documentation/10.0/reference/data.html the header is also missing.

pedrobaeza: For being a correct xml with UTF-8 coded words, you should include that attribute and the header. If not, there can be problems depending on the lxml library implementation and version. What is not needed in python 3 is for python files.

RoelAdriaans: @mreficent I did not say it was new with Odoo 11.. 
In that case Odoo is also doing things wrong, for example in this file: https://github.com/odoo/odoo/blob/11.0/addons/hr_recruitment/views/hr_department_views.xml

(For extra confusion, odoo guidelines specify the utf-8 header for python files, but not for the xml files.. :) )

JayVora-SerpentCS:
Question : When the SO is done, the field becomes editable ! But when the SO is done, in invoice and delivery both are said done, isnt it ?
Sorry if its a confusion.

emagdalenaC2i:
@pedrobaeza I think this PR should be closed because this module is merged in #688

PR #650, opened 2018-05-16
[10.0][IMP] sale_product_set add image

TDu:
Allow to add an image to a product set and move the product set lines
to a notebook as to simplify adding custom fields to the view.

### DIFF
--- a/sale_product_set/models/product_set.py
+++ b/sale_product_set/models/product_set.py
@@ -19,3 +19,7 @@ class ProductSet(models.Model):
         'product_set_id',
         string="Products"
     )
+    image = fields.Binary(
+        string='Image',
+        help='Select an image to display for the product set'
+    )
--- a/sale_product_set/models/product_set.py
+++ b/sale_product_set/models/product_set.py
@@ -21,5 +21,6 @@ class ProductSet(models.Model):
     )
     image = fields.Binary(
         string='Image',
+        attachment=True,
         help='Select an image to display for the product set'
     )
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
Looks like a good improvement to me :+1:

sale_product_set/views/product_set.xml
```diff
@@ -21,15 +21,24 @@
         <form string="Product set" version="7.0">
```
JayVora-SerpentCS: As this is now for v10, we may remove the version param

JayVora-SerpentCS (changes requested):

JayVora-SerpentCS:
@TDu

TDu:
@JayVora-SerpentCS I removed that version parameter.

JayVora-SerpentCS (approved):

bodedra (approved):
LGTM :+1:

sale_product_set/models/product_set.py
```diff
@@ -19,3 +19,7 @@ class ProductSet(models.Model):
         'product_set_id',
         string="Products"
     )
+    image = fields.Binary(
+        string='Image',
+        help='Select an image to display for the product set'
```
pedrobaeza, line 25: Please add `attachment=True` for having the advantages it brings

TDu: @pedrobaeza Thanks for the hint, I applied the change.

pedrobaeza:
Please forward-port it to v11 (and if you want to migrate it to v12!)

Issue #630, opened 2018-04-11
[10.0][sale_generator] Unable to create sale order lines


AaronHForgeFlow:
Cannot create sale order lines on any sale-workflow runbot v10. There's some problem with the sale view. The error message:

![image](https://user-images.githubusercontent.com/19620251/38604334-4e3d7080-3d70-11e8-8038-01ac04436347.png)

The module that introduces that domain to the sale views is sale_generator. After first installed the module no issues were found. However, after uninstalling the module, some domains remain.

bealdav:
Ok, I have to fix it

LoisRForgeFlow:
@bealdav This keeps happening and is very annoying, could you please have a look?

bealdav:
Yes I searched, but I can't to reproduce the problem, in my workstation for now

bealdav:
OK, strange weird unpredictible behavior. But now convinced that unpredectiblable behavior not caused by other module in branch.

I'll focus now on sale_generator.

bealdav:
@lreficent please could you rebuild the branch because impossible to reproduce this bug elsewhere.

Thanks

rafaelbn:
@aheficent it looks this is solved as I have tried and it works. Take a look in: http://3325894-10-0-0674d6.runbot2.odoo-community.org/web#id=12&view_type=form&model=sale.order&menu_id=210&action=282

rafaelbn:
Hi @bealdav ! Here you are: http://3326177-10-0-0674d6.runbot1.odoo-community.org/web?debug

`Error: Unknown field is_template in domain [["is_template","=",true]]`

```
http://3326177-10-0-0674d6.runbot1.odoo-community.org/web/content/549-8252a30/web.assets_backend.js:1767
Traceback:
compute_domain@http://3326177-10-0-0674d6.runbot1.odoo-community.org/web/content/549-8252a30/web.assets_backend.js:1767:42
compute_domain@http://3326177-10-0-0674d6.runbot1.odoo-community.org/web/content/549-8252a30/web.assets_backend.js:2188:339
init/<@http://3326177-10-0-0674d6.runbot1.odoo-community.org/web/content/549-8252a30/web.assets_backend.js:2222:341
trigger@http://3326177-10-0-0674d6.runbot1.odoo-community.org/web/content/543-300e47e/web.assets_common.js:3104:166
[...]

PR #657, opened 2018-06-06
[ADD] sale_order_priority

SimoRubi:
Module to manage the priority of sale orders and link it with the generated picking.

### DIFF
--- a/sale_order_priority/i18n/it.po
+++ b/sale_order_priority/i18n/it.po
@@ -1,0 +1,61 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* sale_order_priority
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 10.0+e\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2018-06-06 10:42+0000\n"
+"PO-Revision-Date: 2018-06-06 10:42+0000\n"
+"Last-Translator: Simone Rubino <simone.rubino@agilebg.com>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: sale_order_priority
+#: selection:sale.order,priority:0
+#: selection:sale.order.line,priority:0
+msgid "Normal"
+msgstr "Normale"
+
+#. module: sale_order_priority
+#: selection:sale.order,priority:0
+#: selection:sale.order.line,priority:0
+msgid "Not urgent"
+msgstr "Non urgente"
+
+#. module: sale_order_priority
+#: model:ir.model.fields,field_description:sale_order_priority.field_sale_order_line_priority
+#: model:ir.model.fields,field_description:sale_order_priority.field_sale_order_priority
+msgid "Priority"
+msgstr "Priorità"
+
+#. module: sale_order_priority
+#: model:ir.model.fields,help:sale_order_priority.field_sale_order_priority
+msgid "Priority for this sale order. Setting manually a value here would set it as priority for all the order lines"
+msgstr "Priorità di questo ordine di vendita. Impostando qui un valore, verrà impostato anche in tutte le righe"
+
+#. module: sale_order_priority
+#: model:ir.model,name:sale_order_priority.model_sale_order
+msgid "Sales Order"
+msgstr "Ordine di vendita"
+
+#. module: sale_order_priority
+#: model:ir.model,name:sale_order_priority.model_sale_order_line
+msgid "Sales Order Line"
+msgstr "Riga d'Ordine di Vendita"
+
+#. module: sale_order_priority
+#: selection:sale.order,priority:0
+#: selection:sale.order.line,priority:0
+msgid "Urgent"
+msgstr "Urgente"
+
+#. module: sale_order_priority
+#: selection:sale.order,priority:0
+#: selection:sale.order.line,priority:0
+msgid "Very Urgent"
+msgstr "Molto urgente"
--- a/sale_order_priority/models/sale.py
+++ b/sale_order_priority/models/sale.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+from odoo.addons.procurement.models import procurement
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    priority = fields.Selection(
+        procurement.PROCUREMENT_PRIORITIES, string='Priority', default='1')
+
+    @api.multi
+    def _prepare_order_line_procurement(self, group_id=False):
+        self.ensure_one()
+        res = super(SaleOrderLine, self) \
+            ._prepare_order_line_procurement(group_id=group_id)
+        res['priority'] = self.priority \
+            or self.default_get(['priority'])['priority']
+        return res
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.depends('order_line.priority')
+    def _compute_priority(self):
+        for order in self:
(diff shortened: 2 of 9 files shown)


sale_order_priority/i18n/it.po
```diff
@@ -0,0 +1,61 @@
...
+#. module: sale_order_priority
+#: model:ir.model.fields,field_description:sale_order_priority.field_sale_order_line_priority
+#: model:ir.model.fields,field_description:sale_order_priority.field_sale_order_priority
+msgid "Priority"
+msgstr "Priorità"
+
+#. module: sale_order_priority
+#: model:ir.model.fields,help:sale_order_priority.field_sale_order_priority
+msgid "Priority for this sale order. Setting manually a value here would set it as priority for all the order lines"
+msgstr "Prioritdi questo ordine di vendita. Impostando qui un valore, verrà impostato anche in tutte le righe"
```
tafaRU: typo

tafaRU (changes requested):
Only minor nitpicking otherwise :+1:

sale_order_priority/i18n/it.po
SimoRubi: thanks for the review, done!

AaronHForgeFlow:
The diff is showing changes in other modules

SimoRubi:
@aheficent I kept them in a separate commit, they fix the errors that are blocking from https://github.com/OCA/maintainer-quality-tools/pull/554

AaronHForgeFlow:
I think it's better to say in the readme that the priority is inherited by the procurement not the picking

SimoRubi:
@aheficent all in all, it's cleaner to move the fix of README errors to another PR -> https://github.com/OCA/sale-workflow/pull/658

SimoRubi:
@aheficent fixed the README too, in order to explain the procurements' role

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+from odoo.addons.procurement.models import procurement
```
rousseldenis: @SimoRubi As you depends on procurement addon, it is better to reflect it in the manifest.

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
...
+        return res
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.depends('order_line.priority')
+    def _compute_priority(self):
+        for order in self:
+            order.priority = order.mapped('order_line') and \
```
rousseldenis: @SimoRubi Better to store order.mapped(...) in a variable before:
```python
priority = order.mapped('order_line.priority')
order.priority = priority and max(priority) or '1'
```

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
...
+        res['priority'] = self.priority \
+            or self.default_get(['priority'])['priority']
+        return res
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.depends('order_line.priority')
+    def _compute_priority(self):
```
rousseldenis: @SimoRubi Please decorate @api.multi see https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

Here under too

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
...
+        for order in self:
+            order.priority = order.mapped('order_line') and \
+                max(order.mapped('order_line.priority')) or '1'
+
+    def _inverse_priority(self):
+        for order in self:
+            order.order_line.write({'priority': order.priority})
+
+    priority = fields.Selection(
```
rousseldenis: @SimoRubi Please put fields definition on top

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
```
rousseldenis: api, fields, models

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @api.depends('order_line.priority')
+    def _compute_priority(self):
+        for order in self:
+            order.priority = order.mapped('order_line') and \
+                max(order.mapped('order_line.priority')) or '1'
+
+    def _inverse_priority(self):
+        for order in self:
+            order.order_line.write({'priority': order.priority})
```
rousseldenis: @SimoRubi You cannot call write in an inverse. Affect recordset values directly

rousseldenis (changes requested):
Code review

AaronHForgeFlow:
Runbot still affected by https://github.com/OCA/sale-workflow/issues/630 cannot create a sale order line. I'll test in local later

sale_order_priority/models/sale.py
SimoRubi: @rousseldenis why can't I call `write` in an inverse method? I took it from https://github.com/odoo/odoo/blob/0780f10c5474ae28e768b1bbba0245b0fb46e990/addons/stock/models/stock_picking.py#L341 and tested it, it works.

Moreover I can't `order.order_line = order.priority` because `order.order_line` is `One2Many`

SimoRubi:
@rousseldenis thanks for the accurate review, everything done but I have [this](https://github.com/OCA/sale-workflow/pull/657#discussion_r193680015) little question

sale_order_priority/models/sale.py
rousseldenis: @SimoRubi You cannot trust everything in code but go to https://github.com/odoo/odoo/blob/0780f10c5474ae28e768b1bbba0245b0fb46e990/odoo/fields.py#L206

The problem is that inverse methods are called by write, see https://github.com/odoo/odoo/blob/0780f10c5474ae28e768b1bbba0245b0fb46e990/odoo/models.py#L3586

If you write in inverse, that can trigger some unwanted recomputations.

You have to loop into order_line and affect values.

SimoRubi: @rousseldenis now I understand what you meant, done, thanks.

Anyway, I checked that assigning directly to the model (https://github.com/odoo/odoo/blob/0780f10c5474ae28e768b1bbba0245b0fb46e990/odoo/models.py#L5238) actually seems to call the `write` method too (https://github.com/odoo/odoo/blob/0780f10c5474ae28e768b1bbba0245b0fb46e990/odoo/fields.py#L954), is it correct or I am mixing up already?

tafaRU (approved):

SimoRubi:
@aheficent actually runbot is green

AaronHForgeFlow:
@SimoRubi It is green. I meant that it is not possible to create a sale order line in runbot. Got error. But it is unrelated to this PR. See issue #630

rousseldenis (approved):

AaronHForgeFlow (commented):
Tested successfully on runbot. It works. 

However, can anybody explain what is the priority in the picking used for? 

If I have two pickings for the same item 1 unit. Only 1 unit in stock. One picking is priority normal and the other is priority urgent. However, the first picking is already reserved, so the second picking remains waiting availability. Is there a way to automatically unreserve the first picking when confirming the second one? Thank you!

PS: I know my question refers to standard Odoo, just asking.

rousseldenis:
@aheficent I think it's just used in the _order property. So, procurement and picking are 'just' ordered (in tree views too). But in case of automatic assignation (scheduler or...), the picking with higher priority will reserve products before other with smaller one. In normal cases, we never see a difference because of equal values.

AaronHForgeFlow:
@rousseldenis all right. So confirming a SO is manual assignation. We have to manually unreserve the SO with lower priority if we need another more urgent SO to take over.

AaronHForgeFlow (approved):
:+1:

SimoRubi:
@OCA/crm-sales-marketing-maintainers is this ok to be merged?

PR #663, opened 2018-07-03
[9.0][IMP] sale_exception: constraint validation tested only one time

zaoral:
There is no need to check state and ignore exception, only lines change. With this improvement exceptions are only tested once upon order validation instead of twice


rousseldenis:
@zaoral Is this still applicable?

rousseldenis:
Closing this as very old and no answer from author. Please feel free to reopen it if needed.

PR #671, opened 2018-07-13
[IMP] 10.0 sale_triple_discount: rework discount computation

mpanarin:
update the calculation of the discounted prices

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -5,14 +5,49 @@
 # Copyright 2018 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 import odoo.addons.decimal_precision as dp
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    @api.depends('discount2', 'discount3')
+    def _get_final_discount(self):
+        self.ensure_one()
+        if self.discounting_type == "additive":
+            return self._additive_discount()
+        elif self.discounting_type == "multiplicative":
+            return self._multiplicative_discount()
+        else:
+            raise ValidationError(_(
+                "Sale order line %s has unknown discounting type %s"
+            ) % (self.name, self.discounting_type))
+
+    def _additive_discount(self):
+        self.ensure_one()
+        discount = sum(
+            [getattr(self, x) or 0.0 for x in self._discount_fields()]
+        )
+        if discount <= 0:
+            return 0
+        elif discount >= 100:
+            return 100
+        return discount
+
+    def _multiplicative_discount(self):
+        self.ensure_one()
+        discounts = [1 - (self[x] or 0.0) / 100
+                     for x in self._discount_fields()]
+        final_discount = 1
+        for discount in discounts:
+            final_discount *= discount
+        return 100 - final_discount * 100
+
+    def _discount_fields(self):
+        return ['discount', 'discount2', 'discount3']
+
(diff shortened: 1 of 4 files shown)


sale_triple_discount/models/sale_order_line.py
```diff
@@ -39,6 +67,16 @@ def _compute_amount(self):
...
         default=0.0,
     )
+    discounting_type = fields.Selection(
+        string="Discounting type",
+        selection=[
+            ('additive', 'Additive'),
+            ('multiplicative', 'Multiplicative'),
+        ],
+        default="multiplicative",
+        help="""Specifies whether discounts should be added additively
```
pedrobaeza: Don't put triple quota for this, or the help will be ugly, and also makes the help to be useful. What is `additively` and what `multiplicatively`

pedrobaeza (changes requested):
Change also the README for explaining new options.

rafaelbn:
Also related with sale_triple_discount https://github.com/OCA/sale-workflow/pull/544

rafaelbn:
@pedrobaeza your comments have been attended

rafaelbn:
I cannot test this because https://github.com/OCA/sale-workflow/issues/630

sale_triple_discount/models/sale_order_line.py
```diff
@@ -11,22 +11,53 @@
...
+        self.ensure_one()
+        discounts = [getattr(self, x) or 0.0 for x in self._discount_fields()]
+        final_discount = 1 - sum(discounts) / 100
+        if final_discount <= 0:
+            return 0
+        return final_discount
+
+    def _multiplicative_discount(self):
+        self.ensure_one()
+        discounts = [1 - (getattr(self, x) or 0.0) / 100
```
chienandalu: It's safer to use `self[x]` to invoke the fields since `getattr` could give access to the object methods.

chienandalu (commented):
It'd also be nice to test the new cases globally or per line

sale_triple_discount/models/sale_order_line.py
```diff
@@ -11,22 +11,53 @@
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    @api.depends('discount2', 'discount3')
+    def _get_final_discount(self):
+        self.ensure_one()
+        if self.discounting_type == "additive":
+            return self._additive_discount()
+        elif self.discounting_type == "multiplicative":
+            return self._multiplicative_discount()
```
TDu, line 21: It should return 0 (zero) if no discount_type configured, I think.
I got None when running some tests and then it breaks the later computation

mpanarin: I made the field `discount_type` required. As it has a default it will pick the correct value if not set in data

TDu: @mpanarin  But should we not be able to have a sale_order_line with no discount applied to it ?

mpanarin: @TDu we can set the discount to 0

mpanarin:
@chienandalu updated ;)

rafaelbn:
Hi @mpanarin thank for this contribution

It could be great to OCA if you could review https://github.com/OCA/sale-workflow/pull/740 as is the same module in the same version and you could benefit of that improvement

We try to fix in 2017 another issue but nobody review us: https://github.com/OCA/sale-workflow/pull/544 (now I think we could close this PR)

About this rework I think we should try to merge it @chienandalu @pedrobaeza could you give a re-review please?

Thanks!

SimoRubi (changes requested):
Thanks for your PR!
This is a good idea indeed, could you please rebase it on 10.0?
Due to the new structure of the module, it should be enough to edit the `_get_triple_discount` method.

grindtildeath:
@pedrobaeza @rafaelbn Can we close this in favour of #889 ?

PR #672, opened 2018-07-16
[10.0][ADD] New module: sale_product_country_filter

andreparames:
This is a new proposed module which restricts the sale of certain products to chosen countries, to comply with legal restrictions and such.


sale_product_country_filter/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+10.0.0.0.1 (2018-07-16)
```
pedrobaeza: 10.0.0.0

pedrobaeza (changes requested):
Isn't better to add a domain over product many2one for preventing the forbidden products to be listed?

andreparames:
A domain might be a good addition, but my fear is that it might compromise the performance, since it involves a rather complex query with a lot of data to scan (since it filters no just on `product_product` and `product_template`, but also on the full hierarchy of categories, since any can block the sale), which will re-run on every key press when searching for the product.

pedrobaeza:
No, the domain is only evaluated once, not each time you type one letter on the many2one AFAIK. Any way, you can use as an alternative a static computed many2many for loading the allowed products, and then use that one as source of the domain, so the performance doesn't impact.

sale_product_country_filter/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+                res["warning"] += u"\n" + warning
+            else:
+                res["warning"] = warning
+        return res
+
+    @api.multi
+    @api.onchange("partner_shipping_id", "partner_id")
+    def onchange_partner_shipping_id(self):
+        res = super(SaleOrder, self).onchange_partner_shipping_id()
+        if (
```
Cedric-Pigeon: The code could perhaps be added in a common method called by both onchange

Cedric-Pigeon:
The problem with the domain on many2one is that in case of partner_id change, the check on products won't be triggered again.
The problem with the actual behavious is that the user is warned too late. He should be aware immediately of the problem.

Cedric-Pigeon (changes requested):

pedrobaeza:
You can change the domain on the onchange, or make the compute dependent of that field.

andreparames:
> Any way, you can use as an alternative a static computed many2many for loading the allowed products, and then use that one as source of the domain, so the performance doesn't impact.

But wouldn't that require loading into that field every product ID except the ones which are embargoed to the specific country of the partner? That may be thousands, or more.

OTOH, I tried figuring out how to apply a domain on the field (without using a helper m2m), but I'm struggling to adapt the [categories query](https://github.com/OCA/sale-workflow/pull/672/files#diff-5d6d974f6bcc8ed2fa23dfbdd8b51b0eR102) into it, since the domain is a query for products.

rafaelbn:
Hi @Cedric-Pigeon @rousseldenis you pushed 2 commits to this PR, could please review? Thanks

PR #668, opened 2018-07-09
[10.0] [ADD] sale_procurement_group_by_requested_date

AaronHForgeFlow:
Supersedes #541

PR #673, opened 2018-07-18
[11.0] [MIG] sale_procurement_group_by_line

murtuzasaleh:
Sale Procurement Group by line
==============================

- This module was written to extend the functionality of procurement groups
created from a Sale order.

- On itself, this module does nothing it is a requirement for modules which
needs to create procurement group per sale order line basis.

Migration of module Sale Procurement Group by line v10 to v11.


pedrobaeza:
@jbeficent is this intended for the same as https://github.com/OCA/purchase-workflow/pull/563?

sale_procurement_group_by_line/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# © 2016 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Sale Procurement Group by Line',
+    'summary': 'Base module for multiple procurement group by Sale order',
+    'version': '11.0.1.1.0',
+    "author": "Camptocamp,"
+              "Eficent,"
+              "Serpent,"
```
JayVora-SerpentCS: Full name plz

sale_procurement_group_by_line/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+              "Odoo Community Association (OCA)",
+    'category': 'Warehouse',
+    'license': 'AGPL-3',
+    'website': "https://github.com/OCA/sale-workflow",
+    'depends': ['sale_stock',
+                ],
```
JayVora-SerpentCS: pep-8

JayVora-SerpentCS (changes requested):

murtuzasaleh:
@pedrobaeza @mreficent Could you please check it.

nikul-serpentcs (commented):
Code Review LGTM

nikul-serpentcs (approved):
Code Review LGTM

nikul-serpentcs:
@pedrobaeza @JayVora-SerpentCS Could you please review and move forward.

pedrobaeza:
I'm afraid I don't know this module, so others should review it.

sale_procurement_group_by_line/README.rst
```diff
@@ -0,0 +1,64 @@
...
+==============================
+
+This module was written to extend the functionality of procurement groups
+created from a Sale order.
+
+On itself, this module does nothing it is a requirement for modules which
+needs to create procurement group per sale order line basis.
+
+
+Usage
```
HviorForgeFlow, line 16: Improve usage section (add an usage example, where you canbenefit from this module avoid using technical language), add some pictures, would be nice to move readme to [new format](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme)

sale_procurement_group_by_line/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2013-2014 Camptocamp SA - Guewen Baconnier
```
HviorForgeFlow: remove copyright from __init__ files.

sale_procurement_group_by_line/model/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2013-2014 Camptocamp SA - Guewen Baconnier
```
HviorForgeFlow: Same here

sale_automatic_workflow_payment_mode/tests/test_automatic_workflow_payment_mode.py
```diff
@@ -52,5 +52,8 @@ def test_full_automatic(self):
         invoice = sale.invoice_ids
```
HviorForgeFlow, line 52: That file is not related with this module, please remove

sale_procurement_group_by_line/tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2013-2014 Camptocamp SA - Guewen Baconnier
```
HviorForgeFlow: Same here

sale_procurement_group_by_line/model/sale.py
```diff
@@ -0,0 +1,115 @@
...
+                                      rounding_method='HALF-UP')
+                procurement_uom = quant_uom
+
+            try:
+                self.env['procurement.group'].\
+                    run(line.product_id, product_qty, procurement_uom,
+                        line.order_id.partner_shipping_id.
+                        property_stock_customer, line.name, line.order_id.name,
+                        values)
+            except UserError as error:
```
HviorForgeFlow, line 108: As an UserError is caught, maybe it's interesting to append  which products have raised it.

HviorForgeFlow (changes requested):

JayVora-SerpentCS:
@murtuzasaleh check.

sale_procurement_group_by_line/model/sale.py
```diff
@@ -0,0 +1,115 @@
...
+        return 8, self.order_id.id
+
+    @api.multi
+    def _action_launch_procurement_rule(self):
+        """
+        Create procurements based on quantity ordered.
+        If the quantity is increased, new procurements are created.
+        If the quantity is decreased, no automated action is taken.
+        """
+        precision = self.env['decimal.precision'].\
```
mreficent: add
```python
res = super(SaleOrderLine, self)._action_launch_procurement_rule()
```
and then end the method with a `return res` instead of `return True`

sale_procurement_group_by_line/model/sale.py
```diff
@@ -77,23 +62,54 @@ def _action_procurement_create(self):
...
-                group_id=line.procurement_group_id.id)
-            vals['product_qty'] = line.product_uom_qty - qty
-            new_proc = self.env["procurement.order"].create(vals)
-            new_proc.message_post_with_view(
-                'mail.message_origin_link',
-                values={'self': new_proc, 'origin': line.order_id},
-                subtype_id=self.env.ref('mail.mt_note').id)
-            new_procs += new_proc
-        new_procs.run()
-        super(SaleOrderLine, self)._action_procurement_create()
```
MiquelRForgeFlow: Why did you delete this super?

MiquelRForgeFlow (commented):
@murtuzasaleh will you finish this PR?

sale_procurement_group_by_line/model/sale.py
murtuzasaleh: @mreficent Because In odoo v11 _action_procurement_create method is removed and added _action_launch_procurement_rule method [https://github.com/odoo/odoo/commit/c914df5a369eb49e4e72a698c0811200b5825f22#diff-bf6a9af0e4d7537746d73deb353f3f8aL210](https://github.com/odoo/odoo/commit/c914df5a369eb49e4e72a698c0811200b5825f22#diff-bf6a9af0e4d7537746d73deb353f3f8aL210) so I have overwrite this method and added code.

sale_procurement_group_by_line/model/sale.py
AaronHForgeFlow: If super cannot be called then you need to add a comment telling clearly that you are overwriting the original method. Also include it in the known issues section.

AaronHForgeFlow (commented):
Code seems ok to me, it is not calling super but conserving the original behavior, plus adding the grouping by key feature.

sale_procurement_group_by_line/model/sale.py
MiquelRForgeFlow: Yeah, but why you remove the super? You can do super with the _action_launch_procurement_rule method.

nikul-serpentcs: @mreficent [Here](https://github.com/odoo/odoo/commit/c914df5a369eb49e4e72a698c0811200b5825f22#diff-bf6a9af0e4d7537746d73deb353f3f8aR235) super return `True` , So How can modify the code with super?.

That's way @murtuzasaleh overwrite code and modify that.

Here Super is not need. @aheficent already https://github.com/OCA/sale-workflow/pull/673#discussion_r233780204

MiquelRForgeFlow:
```python
return super(self, SaleOrderLine)._action_launch_procurement_rule()
```

AaronHForgeFlow: Hi @murtuzasaleh @nikul-serpentcs Did you try to call super at the end? what does the system do in that case? Are other procurement groups created or the ones you create updated? Maybe they are not. I am not sure looking at the code.

nikul-serpentcs: >  Did you try to call super at the end?

@aheficent Yes, But it's method Call again.

AaronHForgeFlow: But does that affect the result?

LoisRForgeFlow:
@murtuzasaleh Can you rebase? Error in CI was due to another module, but it's fixed now.

AaronHForgeFlow:
BTW sale_source_by_line will be deprecated. This, this module is only a dependency for this one #668 Do you agree @murtuzasaleh?

MiquelRForgeFlow:
Superseeded by https://github.com/OCA/sale-workflow/pull/758.

PR #674, opened 2018-07-19
[11.0] [IMP] sale_order_type: commercial propagation + new README style

pedrobaeza:
* If a contact doesn't have sale type, we use the commercial partner one.
* New README style

@Tecnativa


carlosdauden (approved):

eLBati (approved):

sale_order_type/models/sale_order.py
```diff
@@ -16,8 +16,10 @@ def _get_order_type(self):
     @api.onchange('partner_id')
     def onchange_partner_id(self):
         super(SaleOrder, self).onchange_partner_id()
-        if self.partner_id.sale_type:
-            self.type_id = self.partner_id.sale_type
+        sale_type = (self.partner_id.sale_type or
+                     self.partner_id.commercial_partner_id.sale_type)
+        if sale_type:
+            self.type_id = sale_type
```
chienandalu, line 22: Shouldn't it fallback to the default value if the partner changes and there's no sale type in it?

Example:

#. New SO -> sale type: Normal (default)
#. Change to Partner 1 -> sale type: Special (partner's one)
#. Change to Partner with no sale type: sale type: Special

The last one should go back to the default, isnt't it?

pedrobaeza: That's already do on the default of the field, but it shouldn't be here. Imagine this flow:

* You create the sale order.
* System puts the default order type.
* You change it to another.
* Then you select the partner, but that partner doesn't have any order type in the child nor the parent, then putting the default here on the onchange, will change your preference again to the default one.

pedrobaeza: For the case you mention, the last one should prevail

chienandalu: Ok, I get your point

chienandalu (approved):

pedrobaeza:
Thanks all for the quick review

PR #627, opened 2018-04-05
[11.0][MIG] sale_triple_discount

xavierjimenez:
Migration of sale_triple_discount module to version 11.0

Depends on:

- [x] https://github.com/OCA/account-invoicing/pull/370 account_invoicing_triple_discount

PR #676, opened 2018-07-31
[11.0][NEW] sale_general_discount

sergio-teruel:
New module for set a discount in sale order for all lines additional to standard discount line.
In customer you cat set a default discount.

cc @Tecnativa


sale_global_discount/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    discount = fields.Float(
```
pedrobaeza: `sale_discount`

sale_global_discount/models/sale.py
```diff
@@ -0,0 +1,84 @@
...
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    discount = fields.Float(
+        digits=dp.get_precision('Discount'),
+        default=0.0,
```
bodedra: Remove it. By default it will take 0.0

sale_global_discount/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+        digits=dp.get_precision('Discount'),
+        default=0.0,
```
bodedra: Same. Remove it.

chienandalu (commented):
Isn't it a bit a misleading in some cases? A example  I made:

- Make a pricelist with a discount on products of 50% and set *show public price & discount*
- Assign it to a customer (i.e. Agrolait)
- Assign a default discount to that customer of 50%
- Make a new SO to that customer and add any product (i.e. unit_price = 100)
- The discount shown on the line will be is 50% and the global 50%. The subtotal in the line is price_unit/4 (25).
- The amount discount in totals is price_unit/4 (25) as well.

So the discount is being computed succesively: line -> global

Is this the expected behaviour?

pedrobaeza:
This is going to be discarded as very dangerous, so we can close it.

rafaelbn:
This module work great! It's the best I've seen. In fact is the real solution for global discount in all lines.

@chienandalu , yes it is expected behavior and it's IMHO more clear that for ex. another module like triple discount which is absolutely a headache

I don't see the dangerous and it must be discussed so re-opening

rafaelbn (commented):
@sergio-teruel tested in all use cases and work grear! :+1: 

Just should be fixed the PDF report

rafaelbn (approved):

rafaelbn:
Please @chienandalu test this PR: https://github.com/OCA/sale-workflow/pull/627 try same cases as https://github.com/OCA/sale-workflow/pull/676#pullrequestreview-142737256 Thanks :smile:

rafaelbn:
@sergio-teruel i tested it 01-08-2018 a week ago and we should improve PDF report. This is the screenshot I did:

test sale general discount

pedrobaeza:
I insist, this one is dangerous changing computed fields. @sergio-teruel is already aware and will push other version.

rafaelbn:
OK, cleared up. :+1:

PR #150, opened 2015-05-06
[ADD] new module to realize mass actions (confirm, ...) on sale orders;

legalsylvain:
Add new modules that simply add multi action to realize mass action on sale orders;
For the time being :
- possibility to confirm many quotations at a time ;

(possible improvements)

Thanks for your review.

PR #677, opened 2018-07-31
[10.0][NEW] module to confirm several sale orders at once

Olivier-LAURENT:
New module to confirm several quotations in one click.

cc @astirpe


Olivier-LAURENT:
Hi @astirpe , many thx in advance for reviewing this PR.

astirpe (approved):
@Olivier-LAURENT thanks to you!

astirpe:
@Olivier-LAURENT could you please fix this?
```
2018-07-31 19:09:42,282 200 WARNING openerp_test odoo.addons.base.module.module: module sale_order_mass_confirm: description is empty !
```

legalsylvain (changes requested):
Hehe. This module exists seems 3 years in the OCA ! https://github.com/OCA/sale-workflow/pull/150 (v7)
Could you rename your module into  sale_order_mass_action ?

- it will avoid to create duplicated module with the same features. (in apps, etc..)
- name is more generic, and allow more feature later. (mass invoice, etc...)
- to be consistent with stock.picking module. (https://github.com/OCA/stock-logistics-workflow/pull/390#issuecomment-403444205)

Otherwise, LGTM !

regards.

Olivier-LAURENT:
Hi @astirpe , Are you OK with the @legalsylvain 's request ?

astirpe:
I'm only concern about the name of the module. Right now is clear that the module provides a mass confirmation functionality and only that. Changing the name to something more generic but providing the same functionality is something that I would avoid. This is also the reason why I made a module that was already existing: I didn't notice it because I was looking for something specific and I didn't find it. 

Anyway is a personal preference. I'm not going to block for this. Do what you think is better. My approval stays valid.

sbidoul:
@legalsylvain I share @astirpe 's concern. Perhaps we could leave this one named mass_confirm and later make it depend on a more generic mass_action? What do you think?

legalsylvain:
> I didn't notice it because I was looking for something specific and I didn't find it.

You can find the current OCA 7.0 module with the following tools : 

* odoo-code-search

http://odoo-code-search.com/#sale%20order%20mass%20confirm~~%7B%7D

* odoo-community github tools ([link](https://odoo-community.org/web#view_type=kanban&model=odoo.module&menu_id=567&action=793))

image

> Perhaps we could leave this one named mass_confirm and later make it depend on a more generic mass_action

Well, As said before, I'm more in favour for a generic module, that introduce a tool to do mass actions for each models. (adding action + ir.value + wizard) (like stock_picking_mass_action does, and by the way, the module was named before stock_picking_mass_assign [ref](https://github.com/OCA/stock-logistics-workflow/pull/136)). 

Then we can very easily introduce extra features (set in the [roadmap section](https://github.com/OCA/sale-workflow/blob/7.0/sale_order_mass_action/__openerp__.py#L39) of the current OCA module) like : 
- mass confirm
- mass confirm and invoice
- mass invoice
- (and later mass cancel, etc...)

just with some checkboxes, and a check on the sale order status.

I see changing the name of the current OCA module from mass_action to mass_confirm and creating a second module later as something more complex, and generating more source code to maintain.

My two cents.

astirpe:
@legalsylvain you are totally right and I think everyone here agrees with you regarding the potentials of a generic module. As an experienced OCA contributor, your explanation is even obvious I would say.

But please notice that there are also so many companies that do not have that kind of sensibility. Let's say you are a sales manager of a successful company and you are tired of confirming sales order one by one. What  do you do? When you are back from the coffee machine and you sit at your desk, you go to the apps store of Odoo (or more probably Google) and you search exactly the way you did in your screenshot. (You can try yourself what the result is.)
You will be happy to find out that there is a module which description says "_.. confirm several quotations in one click..._". No extra checkboxes, no need of imagine about extra future functionalities.

Module _sale_order_mass_confirm_ will continue its life in the apps store, even when it exists a powerful OCA alternative named _sale_order_mass_action_. So for many end users it does not really matter if the mass_confirm module will be in OCA or not. Does it matter for OCA?

pedrobaeza:
One note: if there's already a module with that name outside OCA, apps won't let OCA to "own" the module when found on OCA repositories, so you should change the name any way. Module description can be kept as is.

sbidoul:
In any case we don't have budget at the moment to create a generic mass action framework.

So what do we do?

astirpe:
Mass invoicing functionality is already present in the core. I think the discussion is more regarding the technical name to give to the module.

For the _mass_confirm_ module hosted in our repository, I could add "OCA" in the list of authors to accomplish the https://github.com/OCA/sale-workflow/pull/677#issuecomment-410153022. Probably sending a request to Odoo SA would be fixing the update of the github hosting path in the app store.

pedrobaeza:
Yes, you should contact Odoo apps manager for allowing them to change the source repository, as that's the control they make

legalsylvain:
Well. In a few words.

**Option 1** :  keep the current OCA name (sale_order_mass_action)
- doesn't require huge refactoring. (I don't understand your budget constraint @sbidoul. It's just a ```git mv``` + ```sed```  commands on this current PR, don't you think ?).
- the module will be available in odoo apps store without manipulation. (will be displayed [here](https://www.odoo.com/apps/modules/7.0/sale_order_mass_action/), with the mention "available in version 10.0")
- allow to have more features in the future because the name is more generic
- allow users using the current OCA module, to use the new one in V10, when they will migrate, without any operation

**Option 2** : Use the new name proposed in that PR (sale_order_mass_confirm)
- the reverse ;-)

More general point : 
About your story about your sales manager @astirpe, I think that it is not so clear. Creating more little modules for such little features (mass_confirm, mass_cancel, etc....) make more modules to port. The reality is that your sales manager will not found many modules, because there are not ported in the last version. [See](https://twitter.com/legalsylvain/status/1025354474112647175). 

That being said, it's not a big deal and not very strategic. And my PoV for OCA, for the time being is "_Better an imperfect merged module than a perfect unmerged module_" (*) due to the lake of reviewers and developers porting module each year. The rhythm is very hard to follow.

(*) : perfect is subjective, of course ;-) 

So I abstain on that point and I let other contributors saying what is the best. 

Kind regards, and thanks for this work.

astirpe:
I'm still receiving interest for module `sale_order_mass_confirm`. I just published the V8 version (it was requested to me by email) and I already had a request for the V12. As said, people are contacting us because they can easy find the module in the Odoo app store. The reason is that the name given to the module (`sale_order_mass_confirm`) is self explanatory. Instead the name of the OCA version (`sale_order_mass_action`) is not.

I don't want to reopen the discussion again. Was just to inform you in a friendly way :smile:

PR #678, opened 2018-08-08
[MIG]sale_order_add_variants: Migration to 11.0

VRodrigo:
Add variants in sale orders
============================

This module adds a wizard to select a template product showing all of this variants and allowing to change each variant quantity.
Adding this variants to sale order with the quantities selected

Migration of module Sale Order Add Variants from v10 to v11


pedrobaeza:
This module shouldn't be migrated. Instead, you should migrate `sale_order_variant_mgmt`. It has been replaced since 9.0.

pedrobaeza:
And it should migrated in the repository OCA/product-variant. Here you have 10.0 version: https://github.com/OCA/product-variant/tree/10.0/sale_order_variant_mgmt

PR #679, opened 2018-08-10
[MIG] sale-order-revision: Migration to 11.0

devang-dreambits:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0 and sign the CLA (https://odoo-community.org/page/cla)

devang-dreambits:
@pedrobaeza  Sure Sir. Can you cancel this PR? It would be easy to have another proper PR.

pedrobaeza:
If you use the same branch name and make `git push -f`, the PR will be auto-updated without the need of opening a new PR. You can try and if not successful, then we close it.

karan-dreambits:
@OCA/crm-sales-marketing-maintainers Can we ignore that codecov fails for the minor percentage of coverage difference or it would be better to add tests to cover the left out onchange method ?

pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

It would be good to have perfect coverage, but if you are not able to do it, it's not a hard requirement.

karan-dreambits (approved):

sale_order_revision/models/sale_order.py
```diff
@@ -0,0 +1,119 @@
...
+         'unique(unrevisioned_name, revision_number, company_id)',
+         'Order Reference and revision must be unique per Company.'),
+    ]
+
+    @api.returns('self', lambda value: value.id)
+    @api.multi
+    def copy(self, default=None):
+        if default is None:
+            default = {}
+        if self.env.context.get('new_sale_revision'):
```
rousseldenis: @devang-dreambits I think using context could be avoided as it can be done outside (calling another function).

sale_order_revision/models/sale_order.py
```diff
@@ -0,0 +1,119 @@
...
+        return super(SaleOrder, self).create(values)
+
+    @api.multi
+    def create_revision(self):
+
+        revision_ids = []
+        # Looping over sale order records
+        for sale_order_rec in self:
+            # Calling  Copy method
+            copied_sale_rec = sale_order_rec.with_context(
```
rousseldenis: As I said here upper, on each iteration you create a new environment at each call of 'with_context'. Prefer calling a new function like copy_with_new_revision or ...

sale_order_revision/models/sale_order.py
```diff
@@ -0,0 +1,119 @@
...
+                                     copy=False,
+                                     default=0)
+    unrevisioned_name = fields.Char('Order Reference',
+                                    copy=True,
+                                    readonly=True)
+    active = fields.Boolean('Active',
+                            default=True)
+
+    has_old_revisions = fields.Boolean('Has old revisions',
+                                       compute=_has_old_revisions)
```
rousseldenis: Prefer using quoted argument. Facilitate inheritance.

sale_order_revision/models/sale_order.py
```diff
@@ -0,0 +1,119 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends('current_revision_id', 'old_revision_ids')
+    def _has_old_revisions(self):
+        for sale_order in self:
+            if sale_order.old_revision_ids:
+                sale_order.has_old_revisions = True
+            else:
+                sale_order.has_old_revisions = False
```
rousseldenis: You can avoid this as default value is False

rousseldenis (changes requested):
Some comments

sale_order_revision/models/sale_order.py
devang-dreambits: From that statement; we are creating revision by using copy method, and making some changes in a few fields..... and if the user duplicates the records from the action menu, it should also create a revision with different values in some other fields. 

I tried to use a separate method, but it is difficult to identify that revision creation is done by a method or via 'Duplicate' action.

rousseldenis: Ok, but call it on self not in the loop :

`for sale_order_rec in self.with_context(...):`

devang-dreambits: Ok. Thank you for the reply. Meanwhile, I tried on your idea and somehow managed to do that without context, Hope that works well. Look in new commit

sale_order_revision/models/sale_order.py
devang-dreambits: fixed in new commit

gurneyalex (approved):

PR #684, opened 2018-08-24
[8.0][FIX] sale_order_weight: Fix template error

geluboros:
Hello,

There is a small bug in the template that causes the report printing to fail.

The problem is that _t-field_ directive can only be used for field access, so in this case, where we have a calculated value, _t-esc_ is needed.

Regards


pedrobaeza (approved):
Fast-tracking this.

pedrobaeza:
Travis errors are unrelated due to RST syntax that I'll fix later.

PR #685, opened 2018-08-26
[11.0][NEW] sale_general_discount: New module for apply a discount for all sale order lines

sergio-teruel:
cc @Tecnativa


sale_general_discount/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Sale General Discount',
+    'summary': 'Discount per sale order',
+    'version': '11.0.1.0.0',
+    'development_status': 'Alpha',
```
pedrobaeza: Alpha?

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    discount = fields.Float(
```
pedrobaeza: Better `general_discount`?

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,31 @@
...
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        for order in self:
+            super(SaleOrder, self).onchange_partner_id()
+            order.discount = order.partner_id.sale_discount
+
+    @api.multi
+    @api.onchange('discount')
+    def onchange_discount(self):
+        with self.env.norecompute():
```
pedrobaeza: I think you can do everything in one shot. By the way, never `write` on an onchange:

`self.mapped('order_line').update({'discount': order.discount})`

sale_general_discount/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows set a discount for all lines in a sale order, this
```
pedrobaeza: Change it to:
```
This module allows to set a general discount in a sales order. This general
discount is set to each line order in the standard `discount` field.

You can also set a default general discount on customers.
```

sale_general_discount/views/sale_order_view.xml
```diff
@@ -0,0 +1,21 @@
...
+    <record id="sale_order_general_discount_form_view" model="ir.ui.view">
+        <field name="name">sale.order.general.discount.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='amount_tax']" position="after">
+                <field name="discount"/>
+            </xpath>
+            <xpath expr="//field[@name='order_line']" position="attributes">
+                <attribute name="context">{'default_discount': discount}</attribute>
```
pedrobaeza: I prefer the method I showed you through `fields_view_get` to be resilient to other modules modifying the same context. I'm not sure if we can use the new XML attribute `add` for adding the context here. See an example in:

https://github.com/odoo/odoo/blob/7526b64e8a1cf62c2a22e75c0c051a4abad67ec6/addons/website_blog/views/website_blog_templates.xml#L207

pedrobaeza (changes requested):

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+    general_discount = fields.Float(
+        digits=dp.get_precision('Discount'),
+        string='Discount (%)',
+    )
+
+    @api.multi
+    @api.onchange('partner_id')
+    def onchange_partner_id(self):
+        for order in self:
+            super(SaleOrder, self).onchange_partner_id()
```
yajo: You can use `super()` directly, it's Python 3!

OTOH, I don't know if you realized you're calling `super()` inside a loop! :speak_no_evil:  Put the call before entering the loop or we could get a bottleneck here.

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+            order.general_discount = order.partner_id.sale_discount
+
+    @api.multi
+    @api.onchange('general_discount')
+    def onchange_general_discount(self):
+        with self.env.norecompute():
```
yajo: I'd appreciate a little comment explaining why this.

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+            for order in self:
+                order.mapped('order_line').update({
+                    'discount': order.general_discount,
+                })
+        self.recompute()
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        res = super(SaleOrder, self).fields_view_get(
```
yajo: just `super()` too

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+            view_id, view_type, toolbar, submenu)
+        if view_type == 'form':
+            order_xml = etree.XML(res['arch'])
+            order_line_path = "//field[@name='order_line']"
```
yajo: Can't we set the context [in the field directly](https://www.odoo.com/documentation/11.0/reference/orm.html#relational-fields)? Or can't we set a default method instead (in model `sale.order.line` and field `discount`; this default method could take into account `active_model` and `active_id` in the context)? :thinking:

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+        for order in self:
+            super(SaleOrder, self).onchange_partner_id()
+            order.general_discount = order.partner_id.sale_discount
+
+    @api.multi
+    @api.onchange('general_discount')
+    def onchange_general_discount(self):
+        with self.env.norecompute():
+            for order in self:
+                order.mapped('order_line').update({
```
yajo: Doing `self.mapped('order_line').update({` wouldn't be the same? :thinking:

sale_general_discount/models/sale.py
```diff
@@ -0,0 +1,58 @@
...
+                order_line_field = order_line_fields[0]
+                if 'context' not in order_line_field.attrib:
+                    order_line_field.set(
+                        "context", "{'default_discount': general_discount}")
+                else:
+                    ctx_orig = order_line_field.attrib['context']
+                    if ctx_orig[-2] == ",":
+                        char_replace = ',}'
+                    else:
+                        char_replace = '}'
```
yajo: If the answer to https://github.com/OCA/sale-workflow/pull/685#discussion_r212888122 is _no_, let me recommend you a shorter and less error-prone, approach:

```python
context = order_line_field.attrib.get("context", "{}")
context = context.replace("{", "{'default_discount': general_discount,", 1)
order_line_field.attrib['context'] = context
```

This should work, no matter the whitespaces and commas you can find, and it's easier to read IMHO.

yajo (changes requested):

sale_general_discount/models/sale.py
pedrobaeza: That's what I suggested too: https://github.com/OCA/sale-workflow/pull/685#discussion_r212823526

sale_general_discount/models/sale.py
pedrobaeza: See my comment about that: https://github.com/OCA/sale-workflow/pull/685#discussion_r212824134

pedrobaeza: Although your second proposal sounds good

sergio-teruel: I discard the default value option, but i think that if active_... is in context so i will have the parent record to set de default value but I'm not sure.... I try it because the method fiels_view_get is wierd

pedrobaeza: Have you tried the `add` argument in the XML?

sale_general_discount/views/sale_order_view.xml
sergio-teruel: I tried it, but only join as string, you can not update a dictionary

sale_general_discount/models/sale.py
sergio-teruel: ups!!

sale_general_discount/views/sale_order_view.xml
pedrobaeza: Thanks for telling. I will check if it's possible to use it with little changes and propose them to Odoo.

sale_general_discount/models/sale.py
sergio-teruel: @Yajo I tested it, but in context has not active_model.. 
The new line that I am creating has not order_id record...

sergio-teruel:
@Yajo @pedrobaeza xhanges done

sale_order_general_discount/models/sale.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.onchange('general_discount')
+    def onchange_general_discount(self):
+        self.mapped('order_line').update({
+            'discount': self.general_discount,
+        })
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        res = super(SaleOrder, self).fields_view_get(
```
pedrobaeza: Put docstring about the purpose of this overwrite

pedrobaeza (changes requested):
Remove the SVG file.

sergio-teruel:
@pedrobaeza Changes done..

pedrobaeza (approved):

yajo (approved):

sale_order_general_discount/models/sale.py
```diff
@@ -0,0 +1,53 @@
...
+            view_id, view_type, toolbar, submenu)
+        if view_type == 'form':
+            order_xml = etree.XML(res['arch'])
+            order_line_path = "//field[@name='order_line']"
+            order_line_fields = order_xml.xpath(order_line_path)
+            if order_line_fields:
+                order_line_field = order_line_fields[0]
+                if 'context' not in order_line_field.attrib:
+                    order_line_field.set(
```
yajo: I think this `if` is not needed, since the `else` should work in any case.

pedrobaeza: Indeed. Changed.

PR #686, opened 2018-08-28
[FIX] sale_order_invoicing_finished_task: Add missing dependency

yajo:
Currently, this addon uses `sale.order`'s `picking_policy` field, but that field is defined in `sale_stock`, which isn't included in the dependency tree.

Testing the addon, thus, can produce this error:

    ERROR: test_check_qty_to_invoice (odoo.addons.sale_order_invoicing_finished_task.tests.test_sale_order_invoicing_finished_task.TestInvoicefinishedTask)
    `   File "/opt/odoo/auto/addons/sale_order_invoicing_finished_task/tests/test_sale_order_invoicing_finished_task.py", line 140, in test_check_qty_to_invoice
    `     self.sale_order.action_confirm()
    `   File "/opt/odoo/auto/addons/sale_timesheet/models/sale_order.py", line 60, in action_confirm
    `     result = super(SaleOrder, self).action_confirm()
    `   File "/opt/odoo/auto/addons/portal_sale/models/sale_order.py", line 16, in action_confirm
    `     return super(SaleOrder, self).action_confirm()
    `   File "/opt/odoo/auto/addons/sale/models/sale.py", line 462, in action_confirm
    `     order.state = 'sale'
    `   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 954, in __set__
    `     record.write({self.name: write_value})
    `   File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 274, in write
    `     result = super(MailThread, self).write(values)
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3567, in write
    `     self._write(old_vals)
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3793, in _write
    `     self.recompute()
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5340, in recompute
    `     vals = rec._convert_to_write({n: rec[n] for n in ns})
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5340, in <dictcomp>
    `     vals = rec._convert_to_write({n: rec[n] for n in ns})
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5236, in __getitem__
    `     return self._fields[key].__get__(self, type(self))
    `   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 915, in __get__
    `     self.determine_value(record)
    `   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1001, in determine_value
    `     self.compute_value(recs)
    `   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 981, in compute_value
    `     self._compute_value(records)
    `   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 972, in _compute_value
    `     getattr(records, self.compute)()
    `   File "/opt/odoo/auto/addons/sale_order_invoicing_finished_task/models/sale_order.py", line 16, in _get_invoiced
    `     for order in self.filtered(lambda o: (
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5019, in filtered
    `     return self.browse([rec.id for rec in self if func(rec)])
    `   File "/opt/odoo/auto/addons/sale_order_invoicing_finished_task/models/sale_order.py", line 17, in <lambda>
    `     o.invoice_status != 'no' and o.picking_policy == 'one')):
    ` AttributeError: 'sale.order' object has no attribute 'picking_policy'

I simply added the dependency.

### DIFF
--- a/sale_order_invoicing_finished_task/models/sale_order.py
+++ b/sale_order_invoicing_finished_task/models/sale_order.py
@@ -13,8 +13,7 @@ class SaleOrder(models.Model):
                  'order_line.task_ids.invoiceable')
     def _get_invoiced(self):
         super(SaleOrder, self)._get_invoiced()
-        for order in self.filtered(lambda o: (
-                o.invoice_status != 'no' and o.picking_policy == 'one')):
+        for order in self.filtered(lambda o: o.invoice_status != 'no'):
             if not all(t.invoiceable for t in order.mapped(
                     'order_line.task_ids') if t.invoicing_finished_task):
                 order.update({


pedrobaeza:
That field is really not needed and I remove it on v11. I prefer to not add the dependency and remove that part of code.

@gurneyalex @carlosdauden @sergio-teruel can you confirm?

carlosdauden:
This field adds the possibility of appearing, or not, as invoiceable orders of several lines that still do not have all their lines as invoiceable.
Another option is to define that field also in this module, because the option to create a new field for the same does not like me too much.

pedrobaeza:
AFAIK, you are using a field that is intended for other use to reinterpret it here, so it should be added here. I'm still not fully understanding its use though.

yajo:
Please let me add that I have no idea what this addon is for :sweat_smile: , I just wanted to fix the tests. If you know better how the fix should be, could any of you take over the PR please?

carlosdauden:
It is true that its use is being reinterpreted a little.
So maybe we should define a new field for this.
Its use would be:
  - we have an order with, for example, 3 lines that generate tasks
  - we mark a task as invoiceable
  - be able to decide if that order should already appear as invoiceable or have all the tasks performed to invoice the order

carlosdauden:
@Yajo At the moment cherry-pick this:
https://github.com/OCA/sale-workflow/commit/fa996036236b5d521447680311f9c5a205d25635
If after, this feature is needed, we will add the new field.
Ok @gurneyalex ?

yajo:
Done here. Fast-track?

rafaelbn:
Travis is :red_circle: 

Please add full functionality for the module. If we have a SO with 4 sale order lines (each has a task) and 2 of that 4 task are done are marked as invoiceable the SO should be To invoice! Add the field then here in v10 and in v11 please.

rafaelbn:
Check correct functionality before this PR



Check incorrect functionality after this PR

rafaelbn (changes requested):
We cannot change functionality of the module.

Don't make the change or make it but maintaining the same functionality

Thanks

rousseldenis:
@Yajo Do you plan to finish this ?

yajo:
No, sorry.

PR #688, opened 2018-08-29
[11.0][MIG] sale_force_invoiced

MiquelRForgeFlow:
Standard migration.


sale_force_invoiced/model/sale_order.py
```diff
@@ -8,20 +7,19 @@
...
 
     @api.depends('force_invoiced')
     def _get_invoiced(self):
         super(SaleOrder, self)._get_invoiced()
-        for order in self:
-            if order.force_invoiced and order.invoice_status == 'to invoice':
-                order.invoice_status = 'invoiced'
+        self.filtered(lambda so: so.force_invoiced
+                      and so.invoice_status == 'to invoice').write(
+            {'invoice_status': 'invoiced'})
```
LoisRForgeFlow: Revert this change, you cannot use write inside an api.depends.

LoisRForgeFlow (changes requested):

sale_force_invoiced/model/sale_order.py
MiquelRForgeFlow: done

hieulucky111 (approved):

LoisRForgeFlow (approved):
:+1:

sale_force_invoiced/model/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    force_invoiced = fields.Boolean(
+        string='Force invoiced',
+        help='When you set this field, the sales order will be considered as '
+             'fully invoiced, even when there may be ordered or delivered '
+             'quantities pending to invoice.',
+        readonly=True,
+        states={'done': [('readonly', False)]},
```
espo-tony, line 16: I'm wondering if it shouldn't be editable on _sale_ state rather than on _done_ state (or maybe on both).

espo-tony (commented):
Code review LGTM. I just have a question about how this module should be used.

sale_force_invoiced/model/sale_order.py
rafaelbn: Please @mreficent could you answer this doubt? thanks

PR #693, opened 2018-09-05
[11.0][NEW] sale_order_secondary_unit: New module for sale in a secondary product unit

sergio-teruel:
cc @Tecnativa
This module depends of https://github.com/OCA/product-attribute/pull/377


hieulucky111 (approved):

tschanzt (approved):
Looks good to me.

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
```diff
@@ -0,0 +1,69 @@
...
+                'name': cls.product.name,
+                'product_id': cls.product.id,
+                'product_uom_qty': 1,
+                'product_uom': cls.product.uom_id.id,
+                'price_unit': 1000.00,
+            })],
+            'pricelist_id': cls.env.ref('product.list0').id,
+        })
+        so.onchange_partner_id()
+        cls.order = cls.env['sale.order'].create(
```
nikul-serpentcs, line 46: @sergio-teruel here use a central variable because it uses much time.

nikul-serpentcs (commented):
Improve code

rousseldenis (approved):
Tested. It has no more interface conflicts with stock-logistics-warehouse/sale_packaging module. They can coexist.

The error is due to missing https://github.com/OCA/product-attribute/pull/377

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
pedrobaeza: This is not time consuming, but a convenience for not writing long expression. Not blocking any way.

PR #710, opened 2018-10-08
[FIX] sorp: position of button is not good with note field

fmdl:
Before :
image

After :
image

@pedrobaeza


bodedra (approved):

nikul-serpentcs:
@fmdl  If you fixed any issue then add `reference number`(Issues Number)

fmdl:
@nikul-serpentcs ok, but there are no linked issues :-)

fmdl:
Hi @novawish can you check, and approuve if it is ok. Thanks

sla-subteno-it (approved):

PR #711, opened 2018-10-09
[IMP] sale_order_line_input: Use existing view to include fields added by

carlosdauden:
other modules.

For example:
https://github.com/odoo/odoo/blob/704bdba6b44ee13cddd9c25c2ee0d9314f3f4de5/addons/sale_stock/views/sale_order_views.xml#L44
https://github.com/OCA/sale-workflow/blob/d648633462bdd17e74932cb637df98c35f458dd4/sale_order_line_date/views/sale_order_view.xml#L27

@Tecnativa

### DIFF
--- a/sale_order_line_input/views/sale_order_line_view.xml
+++ b/sale_order_line_input/views/sale_order_line_view.xml
@@ -5,8 +5,10 @@
     <record id="view_sales_order_line_input_tree" model="ir.ui.view">
         <field name="name">sale.order.line.input.tree.</field>
         <field name="model">sale.order.line</field>
-        <field name="priority">20</field>
+        <field name="inherit_id" ref="sale.view_order_line_tree"/>
+        <field name="priority">1</field>
         <field name="arch" type="xml">
+          <tree position="replace">
             <tree string="Sales Orders Lines"
                   decoration-info="invoice_status=='to invoice'" decoration-warning="state=='draft'" decoration-muted="state=='cancel'" decoration-success="state=='done'"
                   editable="top" create="true">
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

sergio-teruel:
Tested with other modules... great!!

chienandalu (approved):
:+1:

sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -5,8 +5,10 @@
     <record id="view_sales_order_line_input_tree" model="ir.ui.view">
         <field name="name">sale.order.line.input.tree.</field>
         <field name="model">sale.order.line</field>
-        <field name="priority">20</field>
+        <field name="inherit_id" ref="sale.view_order_line_tree"/>
+        <field name="priority">1</field>
         <field name="arch" type="xml">
+          <tree position="replace">
```
pedrobaeza, line 11: A replace of the whole tree? This is not very inheritable.

carlosdauden: It is done on purpose, respecting that all the fields of the original view exist

What would be the best alternative for this?
A long list of xpath to get to have the same?
A PR to Odoo that will never approve?

carlosdauden: Precisely this change is to improve the inheritance

pedrobaeza: No, simply put a primary view and add what you need.

carlosdauden:
Closed because for full functionality a PR to Odoo will be tried.

As the state field does not exist in the original view, the modules that add fields can not evaluate if they are readonly.
If each module that wants to add a field to that view has to add also the state field is a problem.
The state field is one, but there are more fields that condition the readonly.

PR #712, opened 2018-10-10
[10.0][ADD] sale_discount_display_amount: add FR and ES translations

Cedric-Pigeon:
@sebastienbeau @lmignon


pedrobaeza:
Why not doing this through Weblate?

Cedric-Pigeon:
@pedrobaeza Sorry, you are totally right.I just did it.

PR #713, opened 2018-10-10
 [FIX] sale_order_product_recommendation: wizard didn't properly get on memory created lines

chienandalu:
I couldn't get to the origin point of the issue (lines info gets lost in the wizard action) but this patch fixes it.

cc @Tecnativa

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -110,13 +110,6 @@ class SaleOrderRecommendation(models.TransientModel):
             if i == self.line_amount:
                 break
 
-    @api.model
-    def create(self, vals):
-        if 'line_ids' in vals:
-            vals['line_ids'] = [
-                line for line in vals['line_ids'] if line[2]["is_modified"]]
-        return super(SaleOrderRecommendation, self).create(vals)
-
     @api.multi
     def action_accept(self):
         """Propagate recommendations to sale order."""
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
@@ -22,7 +22,7 @@
                                 <field name="currency_id" invisible="1"/>
                                 <field name="sale_line_id" invisible="1"/>
                                 <field name="is_modified" invisible="1"/>
-                                <field name="product_id" readonly="1"/>
+                                <field name="product_id" readonly="1" force_save="1"/>
                                 <field name="price_unit"/>
                                 <field name="times_delivered"/>
                                 <field name="units_delivered"/>


pedrobaeza:
@Yajo you were the one that creates this. Can you please check why this is needed?

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -80,42 +80,37 @@ def _generate_recommendations(self):
         )
         found_dict = {l["product_id"][0]: l for l in found_lines}
         RecomendationLine = self.env["sale.order.recommendation.line"]
+        wiz_line_ids = self.env["sale.order.recommendation.line"]
         existing_product_ids = []
         # Add products from sale order lines
         for line in self.order_id.order_line:
             found_line = found_dict[line.product_id.id]
-            new_line = RecomendationLine.new({
+            new_line = RecomendationLine.create({
```
yajo: In general terms, creating records in an onchange is not a very good idea... This is a transient model, so records get removed every now and then, so maybe we can make an exception?

:thinking: Have you tried converting each `.new()` to `(0, 0, {...})`? I think that behind the scenes both get casted to the same operation, but @rco-odoo told me recently that this wasn't 100% the case. At least it's worth trying?

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -80,42 +80,37 @@ def _generate_recommendations(self):
...
             if i == self.line_amount:
                 break
-
-    @api.model
-    def create(self, vals):
-        if 'line_ids' in vals:
-            vals['line_ids'] = [
-                line for line in vals['line_ids'] if line[2]["is_modified"]]
-        return super(SaleOrderRecommendation, self).create(vals)
+        self.line_ids = wiz_line_ids
```
yajo: Or is it possible that all those `+=` mess things up? Have you tried just storing all new records in a separate variable and assigning it just once at the end like here?

BTW you can remove line 56 with this approach.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -182,7 +178,6 @@ class SaleOrderRecommendationLine(models.TransientModel):
         "sale.order.recommendation",
         "Wizard",
         ondelete="cascade",
-        required=True,
```
yajo: This should be required...

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: Of course, you shouldn't `create` on an onchange! @chienandalu this is not correct, so you have to find another way to solve the problem.

pedrobaeza:
@Yajo, this is the correct solution

@chienandalu please review it.

chienandalu (commented):
:+1: `force_save` was the key

rafaelbn (approved):

yajo (approved):

PR #715, opened 2018-10-15
[11.0][ADD] sale_elaboration: Assign an elaboration for a sale order line

sergio-teruel:
cc @Tecnativa


sale_elaboration/models/elaboration.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class Elaboration(models.Model):
+    _name = 'elaboration'
```
pedrobaeza: I think a better name would be `product.elaboration`

sale_elaboration/models/elaboration.py
```diff
@@ -0,0 +1,35 @@
...
+    name = fields.Char(
+        required=True,
+    )
+    code = fields.Char(
+        string='Short Code',
+    )
+    product_id = fields.Many2one(
+        comodel_name='product.product',
+        string='Product',
+        ondelete='restrict',
```
pedrobaeza: `index=True`

sale_elaboration/models/elaboration.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    _sql_constraints = [
+        ('name_uniq', 'unique(name)', 'Name must be unique!'),
+        ('code_uniq', 'unique(code)', 'Code must be unique!'),
+    ]
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
```
pedrobaeza: Why not computing the stored field `display_name` and put `_rec_name` to this field for having this automatic? Check https://github.com/OCA/partner-contact/pull/638/files#diff-8304d4924d370b6222c6d916bcf81f1fR22 for an example

sale_elaboration/models/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+            lambda x: x.product_id == product)
+        if sol_for_product:
+            sol_for_product.product_uom_qty += qty
+            return sol_for_product
+        sol = SaleOrderLine.new({
+            'order_id': self.id,
+            'product_id': product.id,
+            'is_elaboration': True,
+        })
+        sol._origin = SaleOrderLine
```
pedrobaeza, line 37: Why this?

sale_elaboration/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    @api.multi
+    def action_done(self):
+        res = super(StockPicking, self).action_done()
+        elaboration_product = defaultdict(lambda: 0.0)
+        for pick in self.filtered(lambda x: x.picking_type_code == 'outgoing'):
+            elaboration_lines = pick.move_lines.filtered(
```
pedrobaeza: It's better to take here directly sale.order.line records:
```python
pick.mapped('move_lines.sale_line_id').filtered(
    ....
```

sale_elaboration/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module extends the functionality of sale orders to allow add an
```
pedrobaeza: s/to allow add/to allow to add

sale_elaboration/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
...
+elaboration in a sale order line and add an extra line with an elaboration
+product linked to the elaboration.
```
pedrobaeza: Define what is an elaboration: `An elaboration is a process that needs to be done over the product, usually on picking/handling phase. It doesn't modify too much the product for needing an specific product, but it adds a surcharge on the final price`.

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+To use this module you need to:
+
+#. Go to a *Sale > Configuration > Products > Elaboration Sale*.
```
pedrobaeza: Without `a`

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+   Create elaborations and a product linked to this.
```
pedrobaeza: Up to this should be in CONFIGURE, not in USAGE

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Go to a *Sale > Quotations > Create*.
```
pedrobaeza: Remove `a`

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+   Create a sale order and select a elaboration in the line.
```
pedrobaeza: s/sale order/sales order

pedrobaeza: s/select a/select an

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Go to picking created by this sale order and validate it.
```
pedrobaeza: s/Go to/Go to the

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Go to *Sales > Quotation*. A new line is created with the product linked to
+   a elaboration.
```
pedrobaeza: s/a/the

sale_elaboration/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_elaboration_user,access_elaboration_user,model_elaboration,base.group_user,1,1,1,1
```
pedrobaeza: I don't think elaboration should be created by all employees. They should be created by sales managers, and readable by all employees. Check also what happens with portal/public users.

pedrobaeza (changes requested):

sale_elaboration/models/elaboration.py
sergio-teruel: Usually, I do not search by product here and this table has not a lot of records.

sale_elaboration/models/elaboration.py
sergio-teruel: By this way, I can search by shot code with '=' and not ilike??

sale_elaboration/models/sale_order.py
sergio-teruel: When i did the test and I call to onchange method the decorator is not executed, so I do not hav the attribute _origin.
Perhaps this would be solved by other way...

sale_elaboration/models/stock_picking.py
sergio-teruel: I do not understand... i want stock move records not sale order lines records...

sale_elaboration/models/elaboration.py
pedrobaeza: OK

sale_elaboration/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+#. Go to *Sale > Quotations > Create*.
+   Create a sales order and select an elaboration in the line.
+#. Go to the picking created by this sale order and validate it.
```
chienandalu: I was testing this in runbot. In this step, how do the warehouse guy know what elaboration is needed. No info is supplied in the picking.

chienandalu: And once I confirm the picking, another picking is created to compensate the stock of the elaboration product. Anyway, this shouldn't be the case if the product is a consumable, isn't it?

rousseldenis:
@sergio-teruel @pedrobaeza In which case this cannot be modelized by a manufacture step ?

sergio-teruel:
@rousseldenis This is a particular use case, there is not final a transformation product, the user only update in sale line an aditional product to add depending of the elaboration selected. Manufactur process is very slow for this mecanism.

pedrobaeza:
@rousseldenis there are several conditions for this being ideal for several industries (ours is the fish commerce):

* There's no final product nor raw product. It's basically the same, but with a little process done before delivering. For example, cut in pieces.
* This elaboration is optional. Some customers want the original piece as is.
* The elaboration should be invoiced.
* The elaboration is conditional to the final delivered quantity.
* This is done streamlined while preparing the delivery.

This makes very unmanageable (and slower) to adapt a manufacturing flow here (and more when this has to be done with not ideal working conditions for using a compute/tablet device - imagine the environment in a fish shop).

rousseldenis:
Ok. I see.

> * This is done streamlined while preparing the delivery.

As manufacturing orders appear in Inventory Dashboard (Operation types), the entry point is the same as deliveries (and you could add an automatic push rule at the end from PRODUCTION to OUT)

The other points could be added to the mrp.bom type configuration maybe. The manufacturing process is slow if you use standard flow.

It's instinctively what I felt at the moment :smile:

pedrobaeza:
Yeah, but you have to:

1. Confirm sales order.
2. Go to manufacturing order or inventory.
3. Validate the pickings.
4. Go to the other delivery order.
5. Validate.
6. Change quantities if different.

They only want to validate one picking and go. Seriously this module is key in their business. And at the end, the logic is very simple if you see it.

rafaelbn (approved):
Tested functionally :+1: 

Running in production already

sale_elaboration/readme/USAGE.rst
pedrobaeza: Elaboration products are services.

sale_elaboration/models/elaboration.py
pedrobaeza: The code should be searched by the same operator. I don't see correct to change it for forcing for whole code matching.

sale_elaboration/models/stock_picking.py
pedrobaeza: OK, got it.

sale_elaboration/models/sale_order.py
pedrobaeza: The only reference I find for this is https://github.com/odoo/odoo/blob/f8b38f64d3cad59228b9ab4bbc32507d915c6b7d/addons/sale_stock/models/sale_order.py#L193, and in this case it doesn't add anything valuable.

LoisRForgeFlow:
@sergio-teruel @pedrobaeza could you check travis? test are failling.

pedrobaeza:
Error is on main branch also

pedrobaeza:
@sergio-teruel didn't you pull the branch? You have pushed useless commits and a merge with conflicts branch. Please fix it.

pedrobaeza (approved):

PR #725, opened 2018-10-31
[ADD] product_pricelist_supplierinfo: priceslists based in supplier info

cubells:
New module when you can create sales pricelists based on supplierinfo prices.

cc @Tecnativa


pedrobaeza:
Superseeded by OCA/product-attribute#400, as sale dependency has been removed.

PR #726, opened 2018-10-31
[FIX] sale_force_invoiced: typo

cubells:
Field `track_service` only is created if `stock` module is installed. And the field in odoo 11.0 is `tracking`.
Remove field as It's not necessary for tests and this changes avoid warnings.

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this as obvious

PR #731, opened 2018-11-05
[8.0][FIX] sale_delivery_split_date: sale_id not working

oihane:
When a picking is invoiced it will invoice using the delivery address (partner_id from stock.picking)
This fix avoids that error.

cc @tafaRU

### DIFF
--- a/sale_delivery_split_date/models/stock_picking.py
+++ b/sale_delivery_split_date/models/stock_picking.py
@@ -3,15 +3,37 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from openerp import api, fields, models
+from openerp.osv import fields as old_fields
 
 
 class StockPicking(models.Model):
 
     _inherit = 'stock.picking'
 
+    def _get_sale_id(self, cr, uid, ids, name, args, context=None):
+        res = {}
+        line_model = self.pool['sale.order.line']
+        for picking in self.browse(cr, uid, ids, context=context):
+            res[picking.id] = super(StockPicking, self)._get_sale_id(
+                cr, uid, picking.ids, name, args, context=context)[picking.id]
+            if not res[picking.id] and picking.group_id:
+                line_id = line_model.search(
+                    cr, uid,
+                    [('procurement_group_id', '=', picking.group_id.id)],
+                    limit=1, context=context)
+                line = line_model.browse(cr, uid, line_id, context=context)
+                res[picking.id] = line.order_id
+        return res
+
     min_dt = fields.Date(
         string='Scheduled Date (for filter purpose only)',
         compute='_compute_min_dt', store=True)
+
+    _columns = {
+        'sale_id': old_fields.function(
+            _get_sale_id, type="many2one", relation="sale.order",
+            string="Sale Order"),
+    }
 
     @api.multi
     @api.depends('min_date')
--- a/sale_delivery_split_date/tests/test_sale_delivery.py
+++ b/sale_delivery_split_date/tests/test_sale_delivery.py
@@ -45,6 +45,11 @@ class TestSaleDelivery(TransactionCase):
         self.assertEqual(
             len(self.so.picking_ids), 1,
             "There must be 1 picking for the SO when confirmed")
+        for picking in self.so.picking_ids:
+            self.assertEqual(
+                picking.sale_id, self.so,
+                "The sale order related to the picking must be the "
+                "previously confirmed SO")
         self.assertEqual(
             self.so.picking_ids[0].min_date[:10],
             self.so.picking_ids[0].min_dt)
@@ -68,6 +73,11 @@ class TestSaleDelivery(TransactionCase):
         self.assertEqual(
             len(self.so.picking_ids), 2,
             "There must be 2 pickings for the SO when confirmed")
+        for picking in self.so.picking_ids:
+            self.assertEqual(
+                picking.sale_id, self.so,
+                "The sale order related to the picking must be the "
+                "previously confirmed SO")
         sorted_pickings = self.so.picking_ids.sorted(lambda x: x.min_date)
         self.assertEqual(
             self.so.picking_ids[0].min_date[:10],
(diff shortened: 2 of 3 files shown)


oihane:
@sbidoul @max3903 @pedrobaeza @mreficent @jbeficent 
Can anybody check this out? Thanks in advance.

sale_delivery_split_date/models/stock_picking.py
```diff
@@ -3,16 +3,38 @@
...
+                line = line_obj.browse(cr, uid, line_id, context=context)
+                res[picking.id] = line.order_id
+        return res
+
     min_dt = fields.Date(
         string='Scheduled Date (for filter purpose only)',
         compute='_compute_min_dt', store=True)
 
+    _columns = {
+        'sale_id': old_fields.function(
```
MiquelRForgeFlow, line 33: Why do you need to add this field? This field already exists.

oihane: If I'm not mistaken when the field is a function (old API) you need to define again in order to extend the function itself. Not sure, and I haven't been able to prove.

MiquelRForgeFlow: Ah, ok, I am not proficient with old api

MiquelRForgeFlow (approved):
LGTM

anajuaristi (approved):

tafaRU:
@oihane, thanks for the PR. I'm hoping to review it in a day.

tafaRU:
@oihane, could you please describe a bit better the use case this PR is intended to cover?
Thank you in advance.

oihane:
I'll try

Usually, without this module, if you are using multiple addresses in sales orders you can define a delivery address and a different invoicing address
image

In that case picking will have Charlie Bernard and in invoice the address is going to be Jessica Dupont (when invoicing from picking or sale order)

Without this fix, if you invoice a picking the address of the invoice is Charlie Bernard and this PR fixes that.

Hope this helps a bit.

tafaRU:
Thanks @oihane but I cannot understand what's the relation with this module :thinking: I mean, how is this relevant to splitting the delivery according to the requested date?
It would be better if you write the steps to reproduce it. Thanks!

oihane:
I was wrong with the example.
First of all, with a demo order, that already has a picking, I've changed the payment term to "15 days" that is different from the one defined in the partner, by default it has "30 Net Days".
image

The invoice created from the transferred picking is:
image

Creating one with this module installed:
image
image

Then 2 delivery orders are created without sale.order
image

So when one of them is invoiced, the payment term is taken from the partner, not from the sale order
image

If you use the [sale_order_type](https://github.com/OCA/sale-workflow/tree/8.0/sale_order_type) module it will not pass order type from sale.order to stock.picking, and maybe there will be other modules also failing.

Maybe it is more clear now

sale_delivery_split_date/models/stock_picking.py
```diff
@@ -3,16 +3,38 @@
...
+from openerp.osv import fields as old_fields
 
 
 class StockPicking(models.Model):
 
     _inherit = 'stock.picking'
 
+    def _get_sale_id(self, cr, uid, ids, name, args, context=None):
+        res = {}
+        line_obj = self.pool.get('sale.order.line')
```
tafaRU: I'd prefer: `line_model = self.pool['sale.order.line']`

sale_delivery_split_date/tests/test_sale_delivery.py
```diff
@@ -45,6 +45,10 @@ def test_check_single_date(self):
         self.assertEqual(
             len(self.so.picking_ids), 1,
             "There must be 1 picking for the SO when confirmed")
+        for picking in self.so.picking_ids:
+            self.assertEqual(
+                picking.sale_id, self.so,
+                "The sale order form picking must be the confirmed SO")
```
tafaRU: could you rephrase it a bit? what's `form` stand for?

sale_delivery_split_date/tests/test_sale_delivery.py
```diff
@@ -68,6 +72,10 @@ def test_check_multiple_dates(self):
         self.assertEqual(
             len(self.so.picking_ids), 2,
             "There must be 2 pickings for the SO when confirmed")
+        for picking in self.so.picking_ids:
+            self.assertEqual(
+                picking.sale_id, self.so,
+                "The sale order form picking must be the confirmed SO")
```
tafaRU: same as above

sale_delivery_split_date/tests/test_sale_delivery.py
```diff
@@ -101,6 +109,10 @@ def test_check_same_dates(self):
         self.assertEqual(
             len(self.so.picking_ids), 1,
             "There must be only one picking for the SO when confirmed")
+        for picking in self.so.picking_ids:
+            self.assertEqual(
+                picking.sale_id, self.so,
+                "The sale order form picking must be the confirmed SO")
```
tafaRU: same as above

sale_delivery_split_date/models/stock_picking.py
```diff
@@ -3,16 +3,38 @@
...
+        for picking in self.browse(cr, uid, ids, context=context):
+            res[picking.id] = super(StockPicking, self)._get_sale_id(
+                cr, uid, picking.ids, name, args, context=context)[picking.id]
+            if not res[picking.id] and picking.group_id:
+                line_id = line_obj.search(
```
tafaRU: s/line_obj/line_model (see above)

tafaRU (changes requested):
Thanks @oihane I finally got your point! 
Some comments, otherwise LGTM.

sale_delivery_split_date/tests/test_sale_delivery.py
oihane: misspelled, `from`

oihane:
@tafaRU changes done

tafaRU (approved):

PR #732, opened 2018-11-05
[11.0][MIG] sales_team_security: Adaptation to v11 + enhancement

pedrobaeza:
Sales documents permissions by channels (teams)
===============================================

This module adds a new group called "Channel manager", that includes
the proper permissions for showing only the information related to that
channel:

* Leads/Opportunities
* Customers
* Quotations/Sales Orders

It also handles the propagation of the sales team from commercial partners to
the contacts, which standard doesn't do.

And finally, partners are restricted to the own ones for the group
"User: Own Documents Only" for being coherent with the permission scheme.

Installation
---

At installation time, this module sets in all the contacts that have the sales
team empty the sales team of the parent. If you have a lot of contacts, this
operation can take a while.

Configuration
---

* Go to *Configuration > Users & Companies > Users*.
* Open or create a user.
* On the section "Application Accesses", select "Channel Manager" option.

Known issues / Roadmap
---

* This module modifies sales security groups hierarchy, so any other module
  doing something similar might conflict with this one.
* This module is designed for supporting only sales part, so someone that has
  access to other Odoo parts (for example, an accountant), shouldn't have this
  new permission, or some access errors will be found when seeing invoices and
  other documents. A `sales_team_security_account` bridge module can be done
  for fixing this case, but not in the case of other parts like warehouse.
* Split the module in 2 as now `crm` is independent.


pedrobaeza:
@ernestotejeda please review

sales_team_security/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+#. Go to *Configuration > Users & Companies > Users*.
+#. Open or create a user.
+#. On the section "Application Accesses", select "Channel Manager" option.
```
cristinamartinrod: Minor change: 
... on "Sales" select "Channel Manager" option.

cristinamartinrod (approved):
Works perfect on Leads, Pipeline and Quotations. 👍🏻

ernestotejeda (commented):
Error when accessing a client's form, even if I access as administrator

ernestotejeda (commented):
In the readme it says that the group "User: Own Documents Only" shows those clients who have the user in 'Salesperson'' field, or do not have an assigned Salesperson, however, when I try to reproduce this, also shows clients who have others users in 'Salesperson' field.

ernestotejeda (commented):
For Leads, Pipeline and Quotations I think it works well

oihane (approved):
LGTM

pedrobaeza:
@ernestotejeda thanks for the remarks. I have modified some things according your comments and there are something that should be done in the DB. An alternative solution would be to overwrite some `_check_*` method in the ORM for doing that, but it's not worth being possible to do it simply modifying the record rule in the DB.

ernestotejeda (approved):

PR #733, opened 2018-11-06
[11.0][MIG] sale_stock_picking_blocking

MiquelRForgeFlow:
Standard migration.

PR #739, opened 2018-11-20
[11.0][FIX] sale_stock_picking_blocking: view error

MiquelRForgeFlow:
Fixes https://github.com/OCA/sale-workflow/pull/733#issuecomment-440274502.

### DIFF
--- a/sale_stock_picking_blocking/views/res_partner_view.xml
+++ b/sale_stock_picking_blocking/views/res_partner_view.xml
@@ -6,10 +6,10 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form"/>
         <field name="arch" type="xml">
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@pedrobaeza

pedrobaeza:
This is very fragile to inheritance. Can you please make the xpath a bit less specific?

MiquelRForgeFlow:
Done!

pedrobaeza:
This is because there are several `user_id` fields, corresponding to different models? Can you point to the conflicting XML lines?

MiquelRForgeFlow:
I think what you mention is not the problem. Really, I don't know. But happy to see that with xpath the problem is solved. My work in this is finished.

LoisRForgeFlow (approved):
:green_apple:

PR #743, opened 2018-11-24
[12.0][MIG] sale_commercial_partner

Swapnesh-SerpentCS:
**Sale Commercial Partner**
===================

- This module adds a related stored field *Commercial Entity* on sale orders.


sale_commercial_partner/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
```
nikul-serpentcs, line 1: @Swapnesh-SerpentCS Remove Copyright from `__init__` file

nikul-serpentcs (approved):
Minor Change, Otherwise LGTM :+1:

Swapnesh-SerpentCS:
@nikul-serpentcs 
Do I need to add a new(replace) readme File?  (Based on oca-gen-addon-readme)

nikul-serpentcs:
> @nikul-serpentcs
> Do I need to add a new(replace) readme File? (Based on oca-gen-addon-readme)

@Swapnesh-SerpentCS Yes

Swapnesh-SerpentCS:
@pedrobaeza  This can be merged now.

sale_commercial_partner/README.rst
```diff
@@ -1,56 +1,87 @@
...
-* in *Sales > Sales > Quotations*,
-* in *Sales > Sales > Sales Orders*,
-* in *Sales > Reports > Sales* (it is a native feature in this menu)
+Usage
+=====
 
-.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
-   :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/167/11.0
+#. Go to **Inventory** dashboard and open any picking.
```
pedrobaeza: This text is not the proper for the module.

sale_commercial_partner/models/sale.py
```diff
@@ -8,5 +8,10 @@ class SaleOrder(models.Model):
...
 
     commercial_partner_id = fields.Many2one(
-        'res.partner', related='partner_id.commercial_partner_id', store=True,
-        string='Commercial Entity', index=True)
+        'res.partner',
+        related='partner_id.commercial_partner_id',
+        store=True,
+        string='Commercial Entity',
+        index=True,
+        readonly=False
```
pedrobaeza: Why do you put `readonly=False`? This can be considered a bug in previous versions.

sale_commercial_partner/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+#. Go to **Inventory** dashboard and open any picking.
```
pedrobaeza: Wrong usage

pedrobaeza (changes requested):

Swapnesh-SerpentCS:
Opps, Bad copy paste from module.

sale_commercial_partner/models/sale.py
Swapnesh-SerpentCS: Related fields were editable earlier (till v11) so a user can always change that on Record, But from v12 related fields are readonly by default, So I added `readonly=False` to keep the same behaviour.

pedrobaeza: As said, that's clearly a bug in previous versions. This mustn't be editable!

Swapnesh-SerpentCS: Removed and Forced Pushed, Can be reviewed again.
(Do we need to do anything for the older versions ?)

pedrobaeza: Yes, adding a PR for putting that as `readonly=True` will help all the <=11 installations with this module to avoid bottlenecks and possible DB locks.

Swapnesh-SerpentCS: Will create PRs to fix this for v9 to 11. 
Do I need to create Separate PR for each or Are we doing same as odoo  (As they forward port Fixes to other versions )?

pedrobaeza: In OCA we don't have a forward port process, as each version can require a lot of changes on each version. Sometimes, PSC cherry-picks the commit in upper versions in we find that it's equal. Making both PRs in this case can be the best option, and I can fast-track them.

Swapnesh-SerpentCS:
@pedrobaeza 
This can be Reviewed again.

pedrobaeza (approved):

PR #745, opened 2018-11-24
[12.0][MIG] sale_last_price_info

Swapnesh-SerpentCS:
**Product Last Price Info - Sale**
=======================

This module adds the last sale info of the product.

- Last Sale Price
- Last Sale Date
- Last Customer


sale_last_price_info/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
```
nikul-serpentcs: @Swapnesh-SerpentCS Add new blank line

sale_last_price_info/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import test_sale_last_price_info
```
nikul-serpentcs: same here

nikul-serpentcs (commented):
Minor Change

Swapnesh-SerpentCS:
Commit Squashed and  Travis Done.
CC: @pedrobaeza

nikul-serpentcs (approved):
Code Review LGTM

sale_last_price_info/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import product
```
sudhir-erpharbor: Please add a blank line at the EOF

sudhir-erpharbor (commented):
Minor change

sale_last_price_info/models/product.py
```diff
@@ -8,13 +8,16 @@ class ProductProduct(models.Model):
...
-            key=lambda l: l.order_id.date_order, reverse=True)
-        self.last_sale_date = lines[:1].order_id.date_order
-        self.last_sale_price = lines[:1].price_unit
-        self.last_customer_id = lines[:1].order_id.partner_id
+        so_line_obj = self.env['sale.order.line']
+        for product in self:
+            lines = so_line_obj.search(
+                [('product_id', '=', product.id),
+                 ('state', 'in', ['sale', 'done'])]).sorted(
+                key=lambda l: l.order_id.date_order, reverse=True)
```
sebalix: It would be worth having a stored related field to get the date order at the line level, and use an order by directly with `search` + use the parameter `limit=1` to get the first. This field is going to be called quiet often, and the number of sale order lines for a given product can be huge as the time pass, PostgreSQL will handle that much better than Python.

sale_last_price_info/models/product.py
```diff
@@ -8,13 +8,16 @@ class ProductProduct(models.Model):
...
+            product.last_sale_date = \
+                fields.Datetime.to_string(lines[:1].order_id.date_order)
+            product.last_sale_price = lines[:1].price_unit
+            product.last_customer_id = lines[:1].order_id.partner_id
```
sebalix: No need for the slice `:1` once the `search` call is improved.

sebalix (changes requested):

sale_last_price_info/models/__init__.py
Swapnesh-SerpentCS: Done @sudhir-erpharbor

sale_last_price_info/models/product.py
Swapnesh-SerpentCS: That's what I thought but I decided to not change (Minimal diff policy) the behaviour (I know this doesn't change any functional behaviour)

However, I will add that field and Improve `_compute_last_sale` accordingly if we are getting same reviews from other @oca member too.

bizzappdev: @Swapnesh-SerpentCS IMO best will be to search on sale.order with order='date_order desc' and limit=1 with product_id in the domain.

rven (changes requested):
I'm missing the readme (https://github.com/OCA/maintainer-tools/tree/master/template/module/readme) section used by oca-gen-addon-readme

sale_last_price_info/models/product.py
rven: I fully agree with the remark of @sebalix, in the beginning you won't notice the performance issue.
But when using the system for a couple of time, it will get slower and slower.
Maybe a new module should be added, a module that adds the date_order as related stored field on the sale order line. Then this can be used to solve this suggested change?

Swapnesh-SerpentCS: Hello @rven 
Hope you are doing well.

> Maybe a new module should be added, a module that adds the date_order as related stored field on the sale order line.

I think it's not good to add a new module just to add a new field (Without having additional usage) for one Purpose.
We can add a new module and add that in dependency if the field is going to be used in other modules too.

pedrobaeza: No need for the stored field if you limit search to 1 and put the order on the `search` method directly (not after the results).

Swapnesh-SerpentCS: @pedrobaeza  Do you mean like this?
`search( [('product_id', '=', product.id), ('state', 'in', ['sale', 'done'])], limit=1).sorted( key=lambda l: l.order_id.date_order, reverse=True)`

rven: @pedrobaeza this is not possible, because the field is only available on sale.order.
The search method does not work with order_id.date_order
`    def _check_qorder(self, word):
        if not regex_order.match(word):
            raise UserError(_('Invalid "order" specified. A valid "order" specification is a comma-separated list of valid field names (optionally followed by asc/desc for the direction)'))
        return True
`
To be able to use it in the search, it should be a database field defined on the sale.order.line itself.

Swapnesh-SerpentCS: I think we should go ahead with adding a related field in this module itself.

pedrobaeza: You have 2 options then:

1. Add the date as related stored field in `sale.order.line`. This is convenient that the SQL column is created and populated in the pre-init-hook through an SQL for low installation time. And then do what I have proposed before. It's recommended also to put a new that can't conflict with other fields, like `date_order_sale_last_price_info` (I know, a bit cumbersome but it's the only way to assure uniqueness`, similar to what is done on onchanges).
2. Use the SQL view model `sale.report`, doing the search there, which is going to be more performant any way than current solution.

Swapnesh-SerpentCS: > @Swapnesh-SerpentCS IMO best will be to search on sale.order with order='date_order desc' and limit=1 with product_id in the domain.

@bizzappdev 
This would not be possible as we also need to get `price_unit` from related order line. (We have to use additional search for `sale.order.line where id=%s and product_id=%s` to get Order line if we use this way)

Swapnesh-SerpentCS:
@bizzappdev @pedrobaeza @sebalix  @sudhir-erpharbor @rven 
This can be reviewed again.

sale_last_price_info/models/product.py
```diff
@@ -8,13 +8,17 @@ class ProductProduct(models.Model):
...
+        for product in self:
+            line = so_line_obj.search(
+                [('product_id', '=', product.id),
+                 ('state', 'in', ['sale', 'done'])], limit=1,
+                order="date_order_sale_last_price_info desc")
+            product.last_sale_date = \
+                fields.Datetime.to_string(
+                    line[:1].date_order_sale_last_price_info)
+            product.last_sale_price = line[:1].price_unit
+            product.last_customer_id = line[:1].order_id.partner_id
```
sebalix: No need to use slice here, `line.` is enough.

sale_last_price_info/tests/test_sale_last_price_info.py
```diff
@@ -17,16 +16,13 @@ def setUp(self):
...
-            datetime.strptime(
-                sale_lines[:1].order_id.date_order,
-                "%Y-%m-%d %H:%M:%S").date(),
-            datetime.strptime(self.product.last_sale_date, "%Y-%m-%d").date())
+            sale_line[:1].price_unit, self.product.last_sale_price)
         self.assertEqual(
-            sale_lines[:1].price_unit, self.product.last_sale_price)
-        self.assertEqual(
-            sale_lines[:1].order_id.partner_id, self.product.last_customer_id)
+            sale_line[:1].order_id.partner_id, self.product.last_customer_id)
```
sebalix: Same

sebalix (changes requested):

sale_last_price_info/models/product.py
Swapnesh-SerpentCS: :+1:  and also this will work even if there is no line.

sale_last_price_info/tests/test_sale_last_price_info.py
Swapnesh-SerpentCS: Done

bizzappdev (approved):
:+1:

sebalix (approved):

Swapnesh-SerpentCS:
Thanks, @pedrobaeza.

However, I think still it will be Issue (Due to timezone) on converting datetime to date.
Eg, 

- Create a sale order with Datetime 22 Jan 2018 02:30 AM ((Indian Time  + 05:30))
- It will be stored as 21 Jan 2018 09:00 PM in UTC 
- So here it will be 21 Jan while 22 Jan is correct

Ps: I have not tested this in local yet.

PR #749, opened 2018-11-28
[12.0][ADD] sale_order_rename: quotation / sale order renaming

alexey-pelykh:
Allows setting custom name of sale order / quotation during creation and renaming after creation (similar to `account_invoice_force_number` but for sale orders)


alexey-pelykh:
Calling for good people in request for review with a hope they don't mind 👍 @guewen @mreficent @aheficent @pedrobaeza

pedrobaeza:
Reinventing the wheel :wink: 

https://github.com/OCA/sale-workflow/tree/11.0/sale_quotation_number
https://github.com/OCA/sale-workflow/tree/10.0/sale_order_force_number

alexey-pelykh:
I should've guessed! How I managed to miss _that_!

alexey-pelykh:
@pedrobaeza

Okies, I should've _first_ check and then close the PR 💯 

https://github.com/OCA/sale-workflow/tree/11.0/sale_quotation_number
This one totally different

https://github.com/OCA/sale-workflow/tree/10.0/sale_order_force_number
This is almost the same, yet I see no value of porting that, it would be a complete rewrite for 12.0

pedrobaeza:
Then keep the module name. Any way, I don't recommend nor install none of them. You are going against the paradigm of Odoo design. This is not to complement, but to try to change some designed behaviors, and this will lead for inconsistencies.

alexey-pelykh:
Regarding keeping the name - it's not technically a forced name, I've checked the OCA and Odoo codebase for possible collisions, yet those fields are just UI readonly and that's all. Indeed, not all need such modules, though I see no actual behavioral change. Same result (of renaming) can be achieved by editing the sequence, then creating the SO and then reverting settings of sequence back - which is legit. So, technically, I'm just streamlining that process :)

pedrobaeza:
Yes, in the case of just touching the name, that's it, but this is not allowed in UI for avoiding errors if you use the next number that is going to be assigned by sequence, having an unique constraint error.

alexey-pelykh:
@pedrobaeza you're making great points! Though there was no SQL constraint originally, this module defines one and it has to deal with issues like that. What's your opinion, removing constraint or altering the create to handle the issue?

pedrobaeza:
Well, the problem is that once you have forced one "forbidden" number, you are not going to be able to change it, but repeating the call to sequence for getting a "valid" value can provoke hops on the numbering, because sometimes the problem is only a lock on the sequence, so if you can fine grain the error, then better to bypass it, but if not, then remove the constraint.

alexey-pelykh:
There's a
```
    @api.model
    def create(self, vals):
        if vals.get('name', _('New')) == _('New'):
            if 'company_id' in vals:
                vals['name'] = self.env['ir.sequence'].with_context(force_company=vals['company_id']).next_by_code('sale.order') or _('New')
            else:
                vals['name'] = self.env['ir.sequence'].next_by_code('sale.order') or _('New')
```
that would have to be copy-pasted to this module and wrapped a bit or in `create` call super again if I'll be able to catch specific SQL exception. I don't even know what's worse...

pedrobaeza:
Yes, that's the only way right now, although as said, it's ugly...

alexey-pelykh:
I think copying code would be worse than catching `IntegrityError`, your thoughts @pedrobaeza ?

pedrobaeza:
Yes, better to call to super several times doing rollback.

alexey-pelykh:
@pedrobaeza done

alexey-pelykh:
@rafaelbn I think you'd find this useful too?

rafaelbn:
@alexey-pelykh I don't understand, what do you rename? 

Sale Orders in draft can be rename by default, isn't it?

alexey-pelykh:
@rafaelbn setting custom "Order Reference" during creation, and editing it afterwards.

bealdav:
FYI look at https://github.com/OCA/sale-workflow/issues/1033

rousseldenis:
@alexey-pelykh @rafaelbn @pedrobaeza What's the status of this ?

alexey-pelykh:
@rousseldenis still needs reviews, and still valid PR

rousseldenis:
@alexey-pelykh Sorry, but I don't get what this module do ... Could you explain a little bit more ?

Apart of that, I don't like the code as @pedrobaeza already said

alexey-pelykh:
@rousseldenis as per description
> This module provides feature to rename Quotation / Sale Order in Draft state.

Yet probably you've seen it already, maybe there's some specific question about it?

rousseldenis:
> @rousseldenis as per description
> 
> > This module provides feature to rename Quotation / Sale Order in Draft state.
> 
> Yet probably you've seen it already, maybe there's some specific question about it?

Ok, I've tested it. I was influenced by @rafaelbn comment.

Is there any other mean to do that without the ugly code ?

alexey-pelykh:
@rousseldenis unfortunately, no :( Design of Odoo code in that part is forcing this ugly approach

rousseldenis (approved):

PR #754, opened 2018-12-03
[12.0][MIG] sale_order_archive

astirpe:


sale_order_archive/README.rst
```diff
@@ -0,0 +1,98 @@
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fsale--workflow-lightgray.png?logo=github
+    :target: https://github.com/OCA/sale-workflow/tree/11.0/sale_order_archive
```
bodedra, line 17: Please update readme with `12.0`

bodedra (approved):

sale_order_archive/README.rst
astirpe: It will be updated automatically once the module is merged.

yung-wang (approved):
Ok, seems fine

tafaRU (commented):
I'd add `sale_management` as dependency.

astirpe:
`sale_management` is not required for the installation and the usage of this module. You may want to keep `sale_management` uninstalled and continue to use the functionality provided by `sale_order_archive` for some reason.

tafaRU:
> `sale_management` is not required for the installation

Then the usage section is not compatible! How can I get, for istance, `Open the tree view of sale orders` whithout that module is installed?
The only alternative way I know is by using the `Open view` menu. I therefore suggest to improve the README file adding this info. What do you think?

pedrobaeza:
I wouldn't put the dependency, but on `INSTALL.rst`, it's usual to say: "you need to install `sale_management` module for accessing the needed menus".

tafaRU:
Thanks a lot @astirpe! I merge then.

tafaRU (approved):

pedrobaeza:
Well, last commit would be better to be squashed with migration commit. Now there's no need to make additional commits instead of amending previous one, as GitHub shows the same the diff from previous commit.

tafaRU:
Ok @pedrobaeza good to know it! Next time I'll use the `Squash and merge` option providev by GitHub.
Are there any guidelines the PSC members have to follow when they are merging a PR?

pedrobaeza:
@tafaRU no, I'm not saying to use that option, as you will lose commit history. This is a comment about:

1. When adding last commit, @astirpe can use amend instead of adding a new commit, as any reviewer can view now the diff between previous commit state and current (with the amend) one.
2. In this case, the only solution is to squash locally both commits and push again.

tafaRU:
Ah ok! Now I get your point! Thanks @pedrobaeza .

PR #758, opened 2018-12-04
[11.0][MIG] sale_procurement_group_by_line

MiquelRForgeFlow:
Superseed of https://github.com/OCA/sale-workflow/pull/673 to put the CI in green.


JordiBForgeFlow (approved):

sale_procurement_group_by_line/model/sale.py
```diff
@@ -0,0 +1,115 @@
...
+    @api.multi
+    def _get_procurement_group_key(self):
+        """ Return a key with priority to be used to regroup lines in multiple
+        procurement groups
+
+        """
+        return 8, self.order_id.id
+
+    @api.multi
+    def _action_launch_procurement_rule(self):
```
bjeficent, line 32: I think that you should explain that this function overwrite the standard without calling ```super```

sale_procurement_group_by_line/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+On itself, this module does nothing it is a requirement for modules which
+needs to create procurement group per sale order line basis.
+
+
+Usage
+=====
+
+To use this module, you need to install e.g.
+**Sale Sourced By Line** (`sale_sourced_by_line`)
```
bjeficent: I think this module is actually deprecated.

AaronHForgeFlow (commented):
Agree with @bjeficent  and also put the overwrite thing in the readme please.

sale_procurement_group_by_line/model/sale.py
MiquelRForgeFlow: I updated the PR with calling `super`. It's not an overwrite anymore. 
Please, update review everybody!

bjeficent (approved):
LGTM :+1:

AaronHForgeFlow (approved):
:+1: LGTM

PR #723, opened 2018-10-26
[11.0][FIX] sale_order_secondary_unit: define sale secondary uom in this module instead of product_secondary_unit

sergio-teruel:
cc @Tecnativa
Ping @lmignon

PR #760, opened 2018-12-10
[FIX] sale_order_secondary_unit: field does not exits in product

cubells:
Tests are failing due to a field that not exists.

cc @Tecnativa


yajo (approved):

yajo:
Indeed: https://travis-ci.org/OCA/sale-workflow/jobs/465679283#L2184

carlosdauden (approved):

pedrobaeza:
This is not the solution, but finishing #723. I'm on it.

PR #764, opened 2018-12-12
[10.0] new module sale_delivery_state

yvaucher:
Portage of module `sale_picking_state` renamed to `sale_delivery_state` from: https://github.com/akretion/ak-odoo-incubator/pull/119

This module is there to address the issue when a sale order goes direcly to 'done' state while the delivery is not complete yet. Which in case of synchronisation with other system is lacking accuracy.

Will work well with `delivery` module but doesn't depends on it.

A button to force delivery state appears only in Locked (done) state of the sale order.

### DIFF
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -8,9 +8,10 @@
     "summary": "Show the delivery state on the sale order",
     "version": "10.0.1.0.0",
     "category": "Product",
-    "website": "www.akretion.com",
-    "author": " Akretion",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Akretion, Odoo Community Association (OCA)",
     "license": "AGPL-3",
+    "installable": True,
     "depends": ["sale"],
     "data": [
         "views/sale_order_views.xml",
@@ -18,5 +19,4 @@
     "demo": [
         "demo/sale_demo.xml",
     ],
-    "installable": True,
 }
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion (http://www.akretion.com).
+# @author Pierrick BRUN <pierrick.brun@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale delivery State",
+    "summary": "Show the delivery state on the sale order",
+    "version": "10.0.1.0.0",
+    "category": "Product",
+    "website": "www.akretion.com",
+    "author": " Akretion",
+    "license": "AGPL-3",
+    "depends": ["sale_stock"],
+    "data": [
+        "views/sale_order_views.xml",
+    ],
+    "installable": True,
+}
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -11,9 +11,12 @@
     "website": "www.akretion.com",
     "author": " Akretion",
     "license": "AGPL-3",
-    "depends": ["sale_stock"],
+    "depends": ["sale"],
     "data": [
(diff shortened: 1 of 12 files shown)


sale_delivery_state/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright 2018 Akretion (http://www.akretion.com).
+# @author Pierrick BRUN <pierrick.brun@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale delivery State",
+    "summary": "Show the delivery state on the sale order",
+    "version": "10.0.1.0.0",
+    "category": "Product",
+    "website": "www.akretion.com",
```
nikul-serpentcs: @yvaucher Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

sale_delivery_state/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "author": " Akretion",
```
nikul-serpentcs: @yvaucher same as [author](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L12)

sale_delivery_state/views/sale_order_views.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <data>
```
nikul-serpentcs: @yvaucher remove `data`

sale_delivery_state/views/sale_order_views.xml
```diff
@@ -0,0 +1,34 @@
...
+                </xpath>
+           </field>
+        </record>
+
+        <record id="view_order_tree_inherit_delivery_state" model="ir.ui.view">
+            <field name="name">sale.order.tree</field>
+            <field name="inherit_id" ref="sale.view_order_tree"/>
+            <field name="model">sale.order</field>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='date_order']" position="after">
```
nikul-serpentcs: @yvaucher Here Direct use `<field name=date_order`

nikul-serpentcs (changes requested):
Improve Code

sebastienbeau:
@yvaucher I am ok this the migration, can you fix the little point done by @nikul-serpentcs  and look at travis (test are failing)

sale_delivery_state/models/sale_order.py
```diff
@@ -0,0 +1,114 @@
...
+
+        :returns: boolean
+        """
+        self.ensure_one()
+        # Skip delivery costs lines
+        sale_lines = self.order_line
+        Carrier = self.env.get('delivery.carrier')
+        if Carrier:
+            sale_lines = sale_lines.filtered(
+                lambda rec: not rec.is_delivery_cost())
```
lmignon, line 44: @yvaucher If delivery is installed, a new field is added on the line to specify is a line is a delivery product or not: `is_delivery`
https://github.com/odoo/odoo/blob/12.0/addons/delivery/models/sale_order.py#L130

IMO we should remove the `is_delivery_cost`method and rely on this attribute. That should improve the performances by avoiding an sql search on each call to is_delivery_cost.

lmignon (changes requested):

sale_delivery_state/models/sale_order.py
yvaucher: @lmignon good catch, I missed that field. Will do the change.

grindtildeath: Ping @yvaucher

rafaelbn:
Please @yvaucher review comments and let go forward thanks

rousseldenis:
@yvaucher Is this still applicable ?

rousseldenis:
Closing as this is old. Please feel free to reopen it if needed.

PR #766, opened 2018-12-19
[MIG] sale_order_revision: Migration to 12.0

rven:


rven:
@pedrobaeza do I need to fix this warning?
WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (unrevisioned_name, name) of sale.order() have the same label: Order Reference.

pedrobaeza:
Well, that would be great. It's a new warning from Odoo to avoid problems in the advanced search when locating the proper field.

sale_order_revision/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "name": "Sale order revisions",
+    "version": "12.0.1.0.0",
+    "category": "Sale Management",
+    "author": "Agile Business Group,"
+              "Dreambits,"
+              "Camptocamp,"
+              "Akretion,"
+              "Odoo Community Association (OCA), "
+              "Serpent Consulting Services Pvt. Ltd.",
+    "website": "http://www.agilebg.com",
```
nikul-serpentcs: @rven URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11) 

Add `summary` key

sale_order_revision/tests/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2013 Agile Business Group sagl (<http://www.agilebg.com>)
```
nikul-serpentcs: @rven Remove `Copyright` from all `__init__` file

nikul-serpentcs (commented):
Improve Code

sale_order_revision/view/sale_order.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="sale_order_form" model="ir.ui.view">
```
tbaden: aesthetical change (optional):
use original id

> The naming rules should be followed even when a view is inherited, the module name prevents xid conflicts. In the case where an inherited view has a name which does not follow the guidelines set above, prefer naming the inherited view after the original over using a name which follows the guidelines. This eases looking up the original view and other inheritance if they all have the same name.

tbaden (approved):
thanks for the work 👍 
i did a code review: lgtm
maybe a aesthetic change (see below)

sale_order_revision/view/sale_order.xml
rven: Done!

PR #767, opened 2018-12-19
[9.0] sale_validity: add migration script

chienandalu:
cc @Tecnativa


pedrobaeza (approved):
Merging after removing an unused variable.

PR #768, opened 2018-12-19
[10.0][FIX] sale_triple_discount: Correctly compute taxes with round_globally configuration

SimoRubi:
Steps to reproduce the issue:

0. In debug mode, configure:
    * Tax calculation rounding method* = Round Globally (in Invoicing / Configuration / Settings)
    * Discount = Allow discounts in sales order lines (in Sales / Configuration / Settings)
1. Create a sale order with a product and set:
    * price_unit = 100
    * discount = 50
    * discount2 = 50
    * discount3 = 50
    * any percent tax
2. Save the sale order

Behavior before PR:
The tax is computed on a price of 50 (only discount is applied).
For more proof, try to write something in discount2 and discount3 and check that they are not taken into account at all for taxes computation.

Behavior after PR:
The tax is computed on a price of 12.5 (all discounts are applied).

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -2,6 +2,7 @@
 # Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
 # Copyright 2017 Alex Comba - Agile Business Group
 # Copyright 2017 Tecnativa - David Vidal
+# Copyright 2018 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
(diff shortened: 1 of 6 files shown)


SimoRubi:
The approach is a bit different than before: 
* **before**: Update `price_unit` and `discount`;
* **now**: Update the 3 `discount*` fields.

Other than that note:
* Use of `cache` when updating the line, otherwise recomputations caused the new values to be overwritten by db values;
* The logic has been grouped in `preprocess` and `postprocess` methods, so that it can be easily replicated in `sale.order`.

Any suggestion for improvements is welcome!

stevech091 (approved):
functional test ok

gurneyalex (approved):

sale_triple_discount/models/sale_order_line.py
```diff
@@ -57,9 +54,40 @@ def _prepare_invoice_line(self, qty):
...
-            if line.discount3:
-                line.price_reduce *= (1 - line.discount3 / 100)
+            prev_values[line] = dict(
+                discount=line.discount,
+                discount2=line.discount2,
+                discount3=line.discount3,
+            )
+            line._cache.update({
+                'discount': line._get_triple_discount(),
+                'discount2': 0.0,
```
pedrobaeza, line 80: Why do you need to empty these fields?

SimoRubi: Thanks for your review!
Suppose we have:

- price: 100
- discount: 50
- discount2: 50
- discount3: 50

I am updating the line so that
- price: 100
- discount: 88.5
- discount2: 0
- discount3: 0

If the method is called multiple times (that is the case if rounding method is round_globally), it doesn't change.

On the other hand, if I updated the line so that
- price: 100
- discount: 88.5
- discount2: 50
- discount3: 50

Calling the method another time would give
- price: 100
- discount: 96.875
- discount2: 50
- discount3: 50
And so on..

pedrobaeza: OK, got it!

pedrobaeza:
Is this applicable to v11?

SimoRubi:
@pedrobaeza probably, anyway we should first check if there is the same issue in v11.
"If it ain't broke, don't fix it"

pedrobaeza:
Exactly :wink:

PR #716, opened 2018-10-16
[10.0][IMP] sale_order_type: find type by rule

SimoRubi:
Add feature to automatically select the sale order type based on products (or their categories) of the sale order.

PR #770, opened 2018-12-21
[11.0][IMP] sale_order_type: Find sale order type based on the products of the sale order

SimoRubi:
Forward port of https://github.com/OCA/sale-workflow/pull/716


SimoRubi:
@OCA/crm-sales-marketing-maintainers please check, I don't know if this is the correct way to forward port a change, any suggestion for improvement is appreciated

sale_order_type/README.rst
```diff
@@ -35,6 +35,17 @@ You can see sale types as lines of business.
 You are able to select a sales order type by partner so that when you add a
 partner to a sales order it will get the related info to it.
 
+Rules can also be associated with sale order types.
+
+Inside each rule, you can select any number of products and/or product categories.
+
```
rafaelbn, line 41: Case 1 

- 1 sale order with 2 product
- Each of that 2 products are in 2 different rules of 2 different sale_order_type

Product 1 -> Sale order type 1
Product 2 -> Sale order type 2

Sale Order TEST has product 1 +  product 2

Which is the result?

rafaelbn (approved):
Functionally tested :+1: 

Well I  have some doubts but It works

sale_order_type/README.rst
SimoRubi: Hi, thanks for your review!
As stated few lines below, the order types are inspected following the order of their `sequence` field, so:
0. Suppose Type 1 has sequence 1 and Type 2 has sequence 2;
1. Type 1 is inspected: does this match the current order?
   1. Does any rule of _type 1_ match any product in the order? 
    *. Yes -> choose _type 1_ and exit
    *. No -> go on
   2. Does any rule of _type 1_ match any product category in the order?
    *. (same as above)
2. Repeat step 1 for the following order types

At least this is the behavior I tried to describe in the README, if that is not clear enough I can add an example.

SimoRubi:
@OCA/crm-sales-marketing-maintainers can this be merged? Consider that this functionality is just forward ported from https://github.com/OCA/sale-workflow/pull/716, thanks

sale_order_type/models/sale_order.py
```diff
@@ -36,6 +36,23 @@ def onchange_type_id(self):
             if order.type_id.incoterm_id:
                 order.incoterm = order.type_id.incoterm_id.id
 
+    @api.multi
+    @api.onchange('order_line')
+    def _onchange_order_line(self):
```
pedrobaeza: This onchange shouldn't be triggered automatically.

pedrobaeza (changes requested):

sale_order_type/models/sale_order.py
SimoRubi: Hi, thanks for the review!
It is triggered automatically but it only does something if the `type_id` of the order is empty.
Are you suggesting it should depend on something else or should I remove it completely?

pedrobaeza: It should be removed IMO due to performance reasons.

SimoRubi: Ok then I'll remove it, double-thinking about it, it might be confusing when:

1. add first line -> type is set
2. add second line, type should change but it doesn't

In the end, everyone would be clicking on the `recompute` button to be _sure_ the type is correct..

SimoRubi: Done in https://github.com/OCA/sale-workflow/pull/770/commits/65a7dc240c43f0426b00431a48e3c75e8361c0e1

pedrobaeza (approved):

pedrobaeza:
@SimoRubi please fix this as you have broken several CIs due to hard-coded dependency in test of `l10n_generic_coa`. Please create accounting stuff in the test itself.

SimoRubi:
Hi, I see that Travis of the [merged commit](https://github.com/OCA/sale-workflow/commit/aadd9f72dfbb94d104bb3a558a1cd9d553922b10) is :heavy_check_mark:, which CIs are broken?

pedrobaeza:
Those that don't install `l10n_generic_coa`, which belongs to our customer ones with `l10n_es`. Here is "magically" working due to the auto-install of the module, but that doesn't happen always.

PR #772, opened 2018-12-24
[MIG] sale_order_general_discount: Migration to 12.0

rven:


rven:
@pedrobaeza A lot of the sale-workflow modules depend on module 'sale', but to be able to have the menu entries available in the Odoo application the module 'sale_management' needs to be installed.
Shouldn't all the modules depend on that one instead?

pedrobaeza:
Well, I prefer to keep dependencies to minimum if not technically needed, but I can't impose my opinion here. You can:

1. Leave it as is and say in the README that for being able to access functionality, you have to install `sale_management`.
2. Depend on `sale_management` directly.

rven:
I looked into the already migrated modules and saw it was done in INSTALL.rst file of the module https://github.com/OCA/sale-workflow/blob/12.0/sale_order_archive/readme/INSTALL.rst

tbaden (approved):
code review: lgtm
still open: https://github.com/odoo/odoo/pull/26607 so fields_view_get method remains

did some tests in runbot as well.

maybe a little note in the readme to set "discount" in the general setting of "sale", if you install sale_management, for visible "discount" column in invoice_line tree inside invoice form.

PR #773, opened 2018-12-24
[MIG] sale_order_price_recalculation: Migration to 12.0

rven:


pedrobaeza:
There are other files touched in this PR. I think you haven't performed correctly the procedure for extracting commits.

sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="sorp_view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="arch" type="xml">
+            <field name="order_line" position="after">
```
pedrobaeza: Please use here:
```suggestion
            <field name="note" position="after">
```
the same as in:

https://github.com/OCA/commission/blob/64dc262a10a408c10e54a1cf3e36b74289dc7bc4/sale_commission/views/sale_order_view.xml#L33

as I detected this in that module that moves the total and taxes summary very down out of sight.

rven:
> There are other files touched in this PR. I think you haven't performed correctly the procedure for extracting commits.

I squashed some translation commits... maybe they caused the files to be modified

rven:
@pedrobaeza I don't see anything strange in my commit:
https://github.com/OCA/sale-workflow/pull/773/commits/845970773e12ed921bc049065d23386976c9dd0f
Here only the files I touched are shown.

pedrobaeza:
It's not on your migration commit, but on previous one.

seleccion_020

Maybe an outdated rebase?

rven:
@pedrobaeza does this look better now?

pedrobaeza:
It seems so, thanks

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -1,8 +1,3 @@
-# Copyright 2014 Carlos Sánchez Cifuentes <csanchez@grupovermon.com>
```
pedrobaeza: Why removing here copyright? You should do it only on `__init__.py` files.

sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -6,17 +6,17 @@
...
         <field name="arch" type="xml">
-            <field name="order_line" position="after">
+            <field name="note" position="after">
                 <group name="sorp">
                     <div class="btn-group">
                         <button name="recalculate_prices"
                                 string="Recalculate prices"
                                 attrs="{'invisible': [('state', 'not in', ['draft', 'sent'])]}"
-                                type="object" colspan="4" class="btn-default"/>
+                                type="object" colspan="4" class="ml-2 btn btn-primary"/>
```
pedrobaeza: Don't put it as primary button.

sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -6,17 +6,17 @@
...
                         <button name="recalculate_names"
                                 string="Reset descriptions"
                                 attrs="{'invisible': [('state', 'not in', ['draft', 'sent'])]}"
-                                type="object" colspan="4" class="btn-default"/>
+                                type="object" colspan="4" class="ml-2 btn btn-primary"/>
```
pedrobaeza: Not primary

pedrobaeza (changes requested):

sale_order_price_recalculation/models/sale_order.py
rven: Okay, I misunderstood another PR where I needed to remove these Copyrights, but it was probably only the __init__.py file

pedrobaeza:
Please include https://github.com/OCA/sale-workflow/commit/a3f7b91a3a744c909d04cad19394a7180ff6d603 that I have found that can be a problem in integration tests.

pedrobaeza (approved):

sale_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,113 @@
...
+
+.. image:: /sale_order_price_recalculation/static/description/sale_order_price_recalculation.png
+    :alt: Sale order price recalculation
+
+Known issues / Roadmap
+======================
+
+* In a sale order with lot of lines, the recalculation may slow down, because
+  sale general data (amount untaxed, amount taxed...) are recalculated for
+  each line.
```
NL66278, line 64: You might have a look at the two methods here to boost performance:
- disable prefetch of fields
- delay recompute until all changes that affect the computation have been done, that is in this case, until after all lines have been updated: https://github.com/OCA/contract/blob/f9c3e98a274610bb21be1b81f81c940b0ed131d1/contract/models/account_analytic_account.py#L325

pedrobaeza: Good ideas!

rven: `    
    @api.multi
    def recalculate_prices(self):
        _self = self.with_context(prefetch_fields=False)
        with _self.env.norecompute():
            for line in _self.mapped('order_line'):
                dict = line._convert_to_write(line.read()[0])
                if 'product_tmpl_id' in line._fields:
                    dict['product_tmpl_id'] = line.product_tmpl_id
                line2 = _self.env['sale.order.line'].new(dict)
                # we make this to isolate changed values:
                line2.product_uom_change()
                line.price_unit = line2.price_unit
        _self.recompute()
        return True

    @api.multi
    def recalculate_names(self):
        _self = self.with_context(prefetch_fields=False)
        with _self.env.norecompute():
            for line in _self.mapped('order_line').filtered('product_id'):
                # we make this to isolate changed values:
                line2 = _self.env['sale.order.line'].new({
                    'product_id': line.product_id,
                })
                line2.product_id_change()
                line.name = line2.name
        _self.recompute()
        return True`

Like this?
Do I have to recompute the order lines also?

NL66278: I think this should be enough, as you already update the move lines with the information you want. But to be totally sure you could test this, or even better add a unit test.

ivantodorovich:
Unit price is automatically recalculated whenever you change the quantity. It's very annoying standard behaviour.

Since we are adding a button to recalculate prices, shouldn't we try to avoid this standard behaviour?

wRCfSpTLo3

ivantodorovich (approved):
Tested on runbot

LGTM 👍

pedrobaeza:
> Unit price is automatically recalculated whenever you change the quantity. It's very annoying standard behaviour.

It makes sense when you have escalated prices. Any way, this should be a separated module IMO, but you can combine both.

sale_order_price_recalculation/README.rst
pedrobaeza: @rven are you going to apply these improvement?

ivantodorovich:
> > Unit price is automatically recalculated whenever you change the quantity. It's very annoying standard behaviour.
> 
> It makes sense when you have escalated prices. Any way, this should be a separated module IMO, but you can combine both.

What about a setting in the Sale's Configuration view, under this very same module?
To be honest, I want to avoid thinking of a name for this new module  😅

pedrobaeza:
OK, that can be an option, but it can be added later as improvement with other PR.

ivantodorovich:
> OK, that can be an option, but it can be added later as improvement with other PR.

Yes of course

pedrobaeza:
Please take into account #833 here

rven:
> Please take into account #833 here

Done!

mohamedhagag:
Hi all,
can we merge this ?

bealdav (approved):

PR #779, opened 2019-01-07
[11.0][IMP/FIX] sale_secondary_unit limit secondary uom in order line

GSLabIt:
Added domain to limit `secondary_uom_id` to the one defined in product template.


sergio-teruel (approved):
@GSLabIt Thanks for your contribution

pedrobaeza (approved):

PR #780, opened 2019-01-08
[10][IMP] sale_exception add field

hparfr:
Add a reverse field to leverage optimizations of base_exception.

Since base_exception@10.0.2.0.0, if there is a relation from exception.rule to a model, it will be used to write results once per rule instead of iterate through all the sale order records.


PierrickBrun (approved):

chafique-delli (approved):
Thank you @hparfr  for this improvement

pedrobaeza:
I think this is not going to be very useful when a bit of time has passed and you have tons of past orders in exception.

pedrobaeza:
And this will also slow down a lot the load of the exception form. This should go as an smart-button that navigates to list.

hparfr:
>I think this is not going to be very useful when a bit of time has passed and you have tons of past orders in exception.

The bigger len(sale.order)  / len(exception.rule) is, the bigger this commit will improve perfs.

>And this will also slow down a lot the load of the exception form. This should go as an smart-button that navigates to list.

I don't get it. Which form ? This one  https://github.com/OCA/server-tools/blob/10.0/base_exception/views/base_exception_view.xml#L18 ?

pedrobaeza (changes requested):
See me previous comment.

hparfr:
Added test that can pass only after the following PR is merged in server-tools https://github.com/OCA/server-tools/pull/1480

It's not necessary anymore to scan all the sales to get accurate results.
It's useful when lot of sales can't be excluded by domain (like state=draft).

sale_exception/__manifest__.py
```diff
@@ -4,14 +4,15 @@
 
 {'name': 'Sale Exception',
  'summary': 'Custom exceptions on sale order',
- 'version': '10.0.2.0.0',
+ 'version': '10.0.2.0.1',
  'category': 'Generic Modules/Sale',
  'author': "Akretion, Sodexis, Camptocamp, Odoo Community Association (OCA)",
  'website': 'http://www.akretion.com',
```
nikul-serpentcs: URL Should be `https://github.com/OCA/sale-workflow`

sale_exception/tests/__init__.py
```diff
@@ -3,3 +3,4 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from . import test_sale_exception
+from . import test_multi_records
```
nikul-serpentcs: @hparfr add a new blank line

florian-dacosta:
This PR depends on https://github.com/OCA/server-tools/pull/1601
It fixes the module after the last refactore.

@pedrobaeza Could you please recheck, I believe your request for change is outdated.

@bealdav @mourad-ehm @sebastienbeau

florian-dacosta (approved):

bealdav (approved):
LGTM

mourad-ehm (approved):

florian-dacosta:
I reverted oca_dependencies since https://github.com/OCA/server-tools/pull/1601 has been merged.

This PR is now ready to be merged.

sebastienbeau (approved):
LGTM code review

sebastienbeau:
@pedrobaeza it's ok for you?

sebastienbeau:
The same PR in 12 have been merged so let's merge it https://github.com/OCA/sale-workflow/pull/873

PR #784, opened 2019-01-18
[11.0][IMP] sale_secondary_unit: Set secondary uom quantity as 1.0 by default

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_order_secondary_unit/models/sale_order.py
+++ b/sale_order_secondary_unit/models/sale_order.py
@@ -61,5 +61,16 @@ class SaleOrderLine(models.Model):
             self.secondary_uom_qty = qty
 
     @api.onchange('product_id')
-    def onchange_secondary_unit_product_id(self):
+    def product_id_change(self):
+        """
+        If default sales secondary unit set on product, put on secondary
+        quantity 1 for being the default quantity. We override this method,
+        that is the one that sets by default 1 on the other quantity with that
+        purpose.
+        """
+        res = super(SaleOrderLine, self).product_id_change()
         self.secondary_uom_id = self.product_id.sale_secondary_uom_id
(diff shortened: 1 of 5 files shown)


sale_order_secondary_unit/models/sale_order.py
```diff
@@ -63,3 +63,6 @@ def onchange_product_uom_for_secondary(self):
     @api.onchange('product_id')
     def onchange_secondary_unit_product_id(self):
```
pedrobaeza: This should be named `_onchange_product_id_sale_order_secondary_unit`

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -63,3 +63,6 @@ def onchange_product_uom_for_secondary(self):
...
         self.secondary_uom_id = self.product_id.sale_secondary_uom_id
+        if self.secondary_uom_id:
+            self.secondary_uom_qty = 1.0
```
pedrobaeza, line 74: Shouldn't this be run after quantity set and put the proper amount after conversion of the quantity to this secondary one?

pedrobaeza (changes requested):
I'm not sure about the patch...

Can you put the use case and why this prevents the problem?

sale_order_secondary_unit/models/sale_order.py
sergio-teruel: This is only when the product is changed as a default value

sergio-teruel:
@pedrobaeza changes done!!

rafaelbn (approved):
:+1: tested

@pedrobaeza please review

pedrobaeza (approved):

PR #735, opened 2018-11-14
12.0 mig sale automatic workflow [WIP]

gdgellatly:
PR #707 is dead, so I cherry-picked and started fixing it.  Just WIP for now. Lots and lots of change and my demo system is a bit full of installed stuff atm breaking tests so want to see if fixed here.

\cc @pedrobaeza you probably want to close #707

PR #786, opened 2019-01-18
[12.0][MIG] sale_automatic_workflow_payment_mode

florian-dacosta:
This PR should be ready, but it is blocked until https://github.com/OCA/sale-workflow/pull/735 is merged
 and https://github.com/OCA/bank-payment/issues/536 is fixed.


pedrobaeza:
@florian-dacosta maybe you can help finishing #735

sale_automatic_workflow_payment_mode/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'name': 'Sale Automatic Workflow - Payment Mode',
+    'version': '12.0.1.0.0',
+    'author': 'Camptocamp,Sodexis,Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'category': 'Sales Management',
+    'depends': [
+        'sale_automatic_workflow',
+        'account_payment_sale',  # oca/bank-payment
+    ],
+    'website': 'http://www.camptocamp.com',
```
nikul-serpentcs: https://github.com/OCA/sale-workflow

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    sale_automatic_workflow_payment_mode for Odoo
+    Copyright (C) 2016 Sodexis
+    The licence is in the file __openerp__.py
+-->
+<odoo>
```
nikul-serpentcs: `<odoo noupdate="1">`

nikul-serpentcs (changes requested):
Improve Code

ljimenezsidoo (approved):
Tested With #735 PR :+1:

sale_automatic_workflow_payment_mode/tests/test_automatic_workflow_payment_mode.py
```diff
@@ -52,8 +52,19 @@ def test_full_automatic(self):
...
+        default_vals = {
+            'product_uom_id': move.product_uom.id,
+            'picking_id': move.picking_id.id,
+            'move_id': move.id,
+            'product_id': move.product_id.id,
+            'location_id': move.location_id.id,
+            'location_dest_id': move.location_dest_id.id,
+        }
+        move_line = self.env['stock.move.line'].create(default_vals)
+        move_line.qty_done = 1
```
sebalix: @florian-dacosta I don't think we should force the picking this way. From the UI, Odoo still allows to force a picking by updating the `quantity_done` field on the `stock.move` (like before). See https://github.com/OCA/sale-workflow/blob/7b74129b55aa2f3ffd064143b08f6555182cf04f/sale_automatic_workflow/models/stock_picking.py#L31
Or do you encountered issues with this?

In fact you could simply call the `validate_picking` method from the `sale_automatic_workflow` module here. If we need a fix I can include it in https://github.com/OCA/sale-workflow/pull/830

florian-dacosta: @sebalix Thanks for your comment.
Actually, I do not even remember why I did that.
I removed all this part, as running the automatic workflow already transfer the picking.

pedrobaeza:
Please rebase

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
```diff
@@ -0,0 +1,24 @@
...
+-->
+<odoo noupdate="1">
+
+        <record id="automatic_workflow_payment_filter" model="ir.filters">
+            <field name="name">Automatic Workflow Payment Filter</field>
+            <field name="model_id">account.invoice</field>
+            <field name="domain">[('state', '=', 'open')]</field>
+            <field name="user_id" ref="base.user_root"/>
+        </record>
+        <record id="sale_automatic_workflow.automatic_validation" model="sale.workflow.process">
```
guewen, line 15: This xmlid is in a `noupdate="1"` section in `sale_automatic_workflow` so I think it is never updated here.

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
```diff
@@ -0,0 +1,24 @@
...
+            <field name="register_payment" eval="0" />
+            <field name="payment_filter_id" ref="automatic_workflow_payment_filter" />
+        </record>
+        <record id="sale_automatic_workflow.manual_validation" model="sale.workflow.process">
```
guewen, line 19: This xmlid is in a `noupdate="1"` section in `sale_automatic_workflow` so I think it is never updated here.

florian-dacosta:
@pedrobaeza Rebase done, but It won't pass the tests yet because the issue on account_payment_sale is still there I guess.

@guewen Good catch, but strangely, it seems to work anyway... I just did some test and when installing these modules, the automatic workflow manual and automatics are well updated...(If I do an update once installed, they are not though) .
So all seems good to me. ( In worse case scenario, it is no big deal, since the `register_payment` is not active by default, a configuration has to be made to active it, so if it is empty, the user will fill it)

hzwzw:
@florian-dacosta 
Thank you for your contribution. 
I checked these two recently CI log. They all show the same error:
`raise UserError(_("You try to install module '%s' that depends on module '%s'.\nBut the latter module is not available in your system.") % (module.name, dep.name,))
odoo.exceptions.UserError: ("You try to install module 'sale_automatic_workflow_payment_mode' that depends on module 'account_payment_sale'.\nBut the latter module is not available in your system.", '')`.
I guess `account_payment_sale ` had not been migrated to V12 at that time. 
Luckily [`account_payment_sale`](https://github.com/OCA/bank-payment/tree/12.0/account_payment_sale) is already available on V12.  
Would you mind run the tests again?
Thank you!

florian-dacosta:
@hzwzw you are right, I did forgot to add bank-payment repository to the oca_dependencies file.
I just did it. 
But still, the tests won't pass until https://github.com/OCA/bank-payment/issues/536 is solved

rvalyi:
(I just rebased on 12.0 to avoid the conflict on oca_dependencies.txt)

lmignon:
@florian-dacosta This PR is fixed by https://github.com/OCA/bank-payment/pull/630

lmignon (approved):

lmignon:
@pedrobaeza This one is now :green_apple:

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
pedrobaeza: When installing, the record is updated although noupdate="1", but on consecutive updates this won't be touched again, as well as when migrating with OpenUpgrade.

pedrobaeza: I have checked and data is the same in 11.0, so nothing to do on migration scripts.

pedrobaeza:
Checked quickly the comments, replied about a doubt from @guewen and impact in OpenUpgrade migration and seems good to go, so /ocabot merge !

PR #787, opened 2019-01-21
partner_contact_sale_info_propagation: New module

ernestotejeda:
Cc @Tecnativa


partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -0,0 +1,69 @@
...
+                childs = record.mapped('child_ids').filtered(
+                    lambda r: not r.team_id or r.team_id == record.team_id)
+                if childs:
+                    childs.write({'team_id': vals['team_id']})
+        return super(ResPartner, self).write(vals)
+
+    @api.model
+    def create(self, vals):
+        if vals.get('parent_id'):
+            if not vals.get('user_id'):
```
Nikul-Chaudhary: @ernestotejeda IMO is  better
```
 if not vals.get('user_id') and not vals.get('team_id'):
    vals.update({
        'user_id': self.browse(vals['parent_id']).user_id.id,
        'team_id': self.browse(vals['parent_id']).team_id.id,
    })
```

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -0,0 +1,69 @@
...
+        return super(ResPartner, self).create(vals)
+
+    @api.onchange('parent_id')
+    def onchange_parent_id(self):
+        """Change Salesperson or Sales Channel if the parent company changes
+        and there's no Salesperson or Sales Channel defined yet"""
+        res = super(ResPartner, self).onchange_parent_id()
+        if self.parent_id and self.parent_id != self:
+            parent = self.parent_id
+            if not self.user_id:
```
Nikul-Chaudhary, line 44: same here

Nikul-Chaudhary (changes requested):
Improve Code

partner_contact_sale_info_propagation/models/res_partner.py
pedrobaeza: This should be:
```python
if 'user_id' not in vals and 'team_id' not in vals:
```
or empty fields won't be propagated.

pedrobaeza: But any way, it's correct to split the `if`, as there are 2 conditions inside.

Nikul-Chaudhary: Ok, No Problem @pedrobaeza ,Sound Good

cubells (approved):
:+1: 

LGTM

Code review and test

pedrobaeza (approved):

PR #662, opened 2018-06-20
[10.0] sale_commercial_partner: add compute_sudo=True and readonly=True

alexis-via:

PR #790, opened 2019-01-28
[FIX]sale_commercial_partner: Make commercial_partner_id readonly

Swapnesh-SerpentCS:
Make commercial_partner_id readonly
Based on https://github.com/OCA/sale-workflow/pull/743#discussion_r251329310

@pedrobaeza  For v10 (https://github.com/OCA/sale-workflow/pull/662) It is already there with `compute_sudo` too  (IMO, we dont need `compute_sudo` if it is already `readonly`)

### DIFF
--- a/sale_commercial_partner/__manifest__.py
+++ b/sale_commercial_partner/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Sale Commercial Partner',
     'summary': "Add stored related field 'Commercial Entity' on sale orders",
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'author': 'Akretion,Odoo Community Association (OCA)',
     'website': 'http://www.akretion.com',
     'category': 'Sales',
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
related fields are `compute_sudo=True` by default, but it has nothing to do with `readonly=True`.

pedrobaeza:
Fast-tracking this

Swapnesh-SerpentCS:
No, `compute_sudo=True` is not by default. (Pretty sure upto v10 atleast).

PS: We should use related_sudo instead of compute_sudo. (I am not sure aboute `related_sudo` is True or not be default)
https://github.com/odoo/odoo/blob/11.0/odoo/fields.py#L297

Swapnesh-SerpentCS:
Related fields are `related_sudo=True` always. https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L340

It mentions about reading fields as Admin  (Not about related write) (Maybe compute_sudo is needed to write in that field)

pedrobaeza:
> Related fields are related_sudo=True always. https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L340

That's what I meant to say. `compute_sudo` is always for computing, not for writing.

PR #796, opened 2019-02-07
[11.0][IMP] sale_order_recommendation: Call to all onchange methods

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -6,6 +6,12 @@ from datetime import datetime, timedelta
 from odoo import api, fields, models
 
 
+def _onchange_method(record, field_list):
+    for field in field_list:
+        for onchange_method in record._onchange_methods[field]:
+            onchange_method(record)
+
+
 class SaleOrderRecommendation(models.TransientModel):
     _name = "sale.order.recommendation"
     _description = "Recommended products for current sale order"
@@ -133,9 +139,9 @@ class SaleOrderRecommendation(models.TransientModel):
                 "product_id": wiz_line.product_id.id,
                 "sequence": sequence,
             })
-            so_line.product_id_change()
+            _onchange_method(so_line, ['product_id'])
             so_line.product_uom_qty = wiz_line.units_included
-            so_line.product_uom_change()
+            _onchange_method(so_line, ['product_uom'])
             so_lines |= so_line
         self.order_id.order_line |= so_lines
 


sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -6,6 +6,12 @@
 from odoo import api, fields, models
 
 
+def _onchange_method(record, field_list):
+    for field in field_list:
+        for onchange_method in record._onchange_methods[field]:
```
pedrobaeza, line 11: @Yajo can you link here the piece of code which Odoo uses (you have done it in previous occasions) for using that one instead?

yajo: It's not very easy. There's an addon for that, you can use it if you want: https://github.com/OCA/server-tools/blob/11.0/onchange_helper/models/base.py

pedrobaeza: No, that's not the one I'm referring, and I prefer to avoid that module, as it targets to get the dictionary vals, not what we want. I mean the one in Odoo core.

AaronHForgeFlow: Does this depend on on_change_helper?

pedrobaeza: No, it's a "manual way" of doing something similar

tbaden: do you mean the core function ?https://github.com/odoo/odoo/blob/d6ad22227cc7b3f36ac9914cce4f8a1f0a0f36a2/odoo/models.py#L577-L604

or this part?
```
        if onchange in ("1", "true"):
            for method in self._onchange_methods.get(field_name, ()):
                method_res = method(self)
                process(method_res)
            return
```

pedrobaeza: @tbaden yes, it seems to be that. I saw it in the past, but didn't remember where. Hopefully with odoo/odoo#32784, we will avoid this need.

yajo: The method that the ORM calls is [`onchange()`](https://github.com/odoo/odoo/blob/d6ad22227cc7b3f36ac9914cce4f8a1f0a0f36a2/odoo/models.py#L5052-L5092).

In v12 we can use [the `Form` class from the testing suite](https://www.odoo.com/documentation/12.0/reference/testing.html#odoo.tests.common.Form), although using it for non-tests could be considered abusing the feature.

AFAICS the sanest and best supported way to do this is with the `onchange_helper` module, which is a community-supported helper done precisely for this kind of usages.

tbaden: for future reference this part is also in the models.py
https://github.com/odoo/odoo/blob/d6ad22227cc7b3f36ac9914cce4f8a1f0a0f36a2/odoo/models.py#L5019-L5023

yajo: The more that I look into this issue, the more I get convinced you need to use the `onchange_helper` addon or the `onchange()` method. Let me explain:

- [As you can see](https://github.com/odoo/odoo/blob/d6ad22227cc7b3f36ac9914cce4f8a1f0a0f36a2/odoo/models.py#L4997-L5001), it's legal for an onchange method to return a dictionary like `{'value': {...}}` instead of modifying `self`. This is legal and required sometimes because it's also legal to return other keys like `domain` or `warning` in that dict. You're not supporting this method here.
- [Onchanges can expect having a `__onchange: True` key in their context, or having a `self._origin` attribute pointing to the original record. Also they are expected to be executed in a `newId` record. This is also done in a different context, with different cache. Not supporting this can lead to unexpected results](https://github.com/odoo/odoo/blob/d6ad22227cc7b3f36ac9914cce4f8a1f0a0f36a2/odoo/models.py#L5158-L5161).
- The `newId` record is created from the values that the `onchange` method gets, not from the values that the current record has (in upstream odoo; [`onchange_helper` handles this](https://github.com/OCA/server-tools/blob/e0d943466a4e545bed01b48a126094a703ae0628/onchange_helper/models/base.py#L56)).

Summarizing, it has many corner cases because it's supposed to be used from the web client only, not from the backend; and it's designed as such. To bridge that gap, the safest and sanest way to work seems to be `onchange_helper`, which is not perfect but handles all you need.

yajo (changes requested):

pedrobaeza:
@Yajo there are 2 problems right now:

* We can't change the dependencies of the module in a module in stable.
* The module doesn't handle onchanges with an associated context, which is one of the problems we have with `product_variant_configurator`.

yajo:
In such case, the best that can be done is:

1. Prepare values as expected by `onchange()` mehtod.
1. Call `onchange()` with those values. It will return a dict which will contain a `value` key.
1. Process that key (possibly just `self.write(result["value"])` would work fine.

Other methods don't seem to be resilient enough, as explained above. :roll_eyes: 

I wonder if such a helper would be acceptable upstream for master. :thinking:

pedrobaeza:
There's already SSF in test suite for something similar. The best step is to go for the "computed only world".

yajo:
WDYM with SSF? :stuck_out_tongue:

pedrobaeza:
Clarified by chat

rafaelbn:
@Yajo @sergio-teruel @pedrobaeza then what is needed for this improvement?

rafaelbn (approved):
Functionally tested :+1: 

@yajo please take a look

pedrobaeza:
@Yajo current PR is better than what we have right now, so better to merge it and change on v12 the dependency to onchange_helper, so please approve it.

yajo (approved):
OK, it's better to be better :blush:

PR #443, opened 2017-04-05
[10.0][ADD] sale_double_validation

leemannd:
Porting https://github.com/OCA/sale-workflow/pull/423 from @yvaucher in V10.0

**Double validation for sale**

This module adds an option to enable two steps validation.

When a minimal amount is reached on a Sale order, if a salesman confirm the sale.
The sale order will be set to "To approve" state and will require a Sales manager
to Approve it.

This module is base on the purchase option for two levels validation in module purchase.

PR #423, opened 2017-03-07
[9.0] [NEW] sale_double_validation

yvaucher:
Double validation for sale
==========================

This module adds an option to enable two steps validation.

When a minimal amount is reached on a Sale order, if a salesman confirm the sale.
The sale order will be set to "To approve" state and will require a Sales manager
to Approve it.

This module is base on the purchase option for two levels validation in module `purchase`.

PR #799, opened 2019-02-18
[12.0][MIG] sale double validation

tschanzt:
based on #423 and #443
Moves the validation point to creation as we thought confirm is too late.


sale_double_validation/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+{
+    'name': 'Double validation for Sales',
+    'summary': "",
+    'version': '12.0.1.0.0',
+    'author': 'Camptocamp,Odoo Community Association (OCA)',
+    'maintainer': 'Camptocamp',
+    'license': 'AGPL-3',
+    'category': 'sale',
+    'depends': [
+        'sale',
```
nikul-serpentcs, line 13: @tschanzt no need to `sale`,  `sale_management` depends on `sale`

damdam-s: @nikul-serpentcs you're right. IMHO, I think explicit is always better than implicit :smile:

rousseldenis (approved):
Code review

sale_double_validation/__manifest__.py
leemannd: Hello @nikul-serpentcs , with the answer from @damdam-s , is it ok for you?
If yes can you update your review?

GSLabIt:
An info ..why approve_action set order to draft state when approved? I'd expect state = sale.

tschanzt:
@GSLabIt The thought behind it was that if you create an offer for a client you want it to be checked before it is sent to the client so you can spot errors. To validate after the client has received the offer and accepted makes less sense for me.

GSLabIt:
@tschanzt i think would make more sense to approve a sales order instead, that said, i'd change it on my local, or perhaps add a configuration to choose if approve quotation or sale order.

leemannd (approved):
Thanks

PR #801, opened 2019-02-20
Wrong journal assignment for 'out_refund' invoices

disna-sistemas:
onchange('sale_type_id') - Wrong journal assignment when invoice type is 'out_refund'

### DIFF
--- a/sale_order_type/models/account_invoice.py
+++ b/sale_order_type/models/account_invoice.py
@@ -58,6 +58,6 @@ class AccountInvoice(models.Model):
         if (self.type == 'out_refund'):
             if self.sale_type_id.refund_journal_id:
                 self.journal_id = self.sale_type_id.refund_journal_id.id
-            else:
-                if self.sale_type_id.journal_id:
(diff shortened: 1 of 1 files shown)


sale_order_type/models/account_invoice.py
```diff
@@ -55,5 +55,9 @@ def onchange_partner_id(
     def onchange_sale_type_id(self):
         if self.sale_type_id.payment_term_id:
             self.payment_term = self.sale_type_id.payment_term_id.id
-        if self.sale_type_id.journal_id:
-            self.journal_id = self.sale_type_id.journal_id.id
+        if (self.type == 'out_refund'):
+            if self.sale_type_id.refund_journal_id:
+                self.journal_id = self.sale_type_id.refund_journal_id.id
+            else:
```
rousseldenis: @disna-sistemas I think wrong indent. Could you add tests for that ?

disna-sistemas: True, I think now it's fixed.

PR #804, opened 2019-03-07
[11.0][IMP] sale_order_secondary_unit: add secondary unit to sale report

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_order_secondary_unit/report/sale_report_templates.xml
+++ b/sale_order_secondary_unit/report/sale_report_templates.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="report_saleorder_document" inherit_id="sale.report_saleorder_document">
+    <xpath expr="//t[@t-as='page']//th[2]" position="before">
+        <th class="text-right" groups="product_secondary_unit.group_secondary_unit">Second. Qty</th>
+    </xpath>
+    <xpath expr="//span[@t-field='l.product_uom_qty']/.." position="before">
+        <td class="text-right" groups="product_secondary_unit.group_secondary_unit">
+            <t t-if="l.secondary_uom_id">
+                <span t-field="l.secondary_uom_qty"/>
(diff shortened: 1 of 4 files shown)


yajo (approved):

yajo:
Check Travis.

sergio-teruel:
The error is related to sale_automatic_workflow_payment_mode module

sergio-teruel:
Travis checked

pedrobaeza (changes requested):
With the recent addition of security groups, this should have the same security group for not being showed in that case.

sergio-teruel:
@pedrobaeza Changes done!!

pedrobaeza (approved):

sale_order_secondary_unit/report/sale_report_templates.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="report_saleorder_document" inherit_id="sale.report_saleorder_document">
+    <xpath expr="//t[@t-as='page']//th[2]" position="before">
+        <th class="text-right" groups="product_secondary_unit.group_secondary_unit">Second. Qty</th>
```
pedrobaeza, line 6: I'm seeing this can have access problems. You have to put the group at template level (which is at the end a view), and not at field level).

pedrobaeza (changes requested):

sale_order_secondary_unit/report/sale_report_templates.xml
sergio-teruel: https://github.com/odoo/odoo/blob/22c6084688b7c67d93a1ba5962c2506d7a5b9262/odoo/addons/base/ir/ir_ui_view.py#L357

pedrobaeza: Thanks, I didn't know! Then I suppose there's no access problem in the fetching of that fields.

pedrobaeza (approved):

PR #805, opened 2019-03-08
[11.0][ADD] New Implement on module sale_disable_inventory_check

hieulucky111:
New module sale_disable_inventory_check in PR: https://github.com/OCA/sale-workflow/pull/580 and implement @jcdrubay suggestion 

Something very similar to what is done for "Income Account":
'income': self.property_account_income_id or self.categ_id.property_account_income_categ_id,

We would need:

    A new property field "property_is_check_stock" on Product Template and on Product Category
    A default property like the property with xml record property_account_income_categ_id

### DIFF
--- a/sale_disable_inventory_check/data/product_check_stock_data.xml
+++ b/sale_disable_inventory_check/data/product_check_stock_data.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+
+    <record forcecreate="True" id="default_product_template_check_stock_message" model="ir.property">
+        <field name="name">Check Stock Message Product Template</field>
+        <field name="fields_id" ref="field_product_template_check_stock_on_sale"/>
+        <field name="value">defer</field>
+        <field name="type">selection</field>
+    </record>
+    <record forcecreate="True" id="default_product_category_check_stock_message" model="ir.property">
+        <field name="name">Check Stock Message Product Category</field>
+        <field name="fields_id" ref="field_product_category_check_stock_on_sale"/>
+        <field name="value_integer">1</field>
+        <field name="type">boolean</field>
+    </record>
+
+</odoo>
--- a/sale_disable_inventory_check/models/product_category.py
+++ b/sale_disable_inventory_check/models/product_category.py
@@ -1,0 +1,16 @@
+# Copyright Komit <http://komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    check_stock_on_sale = fields.Boolean(
+        company_dependent=True,
+        help="Uncheck if you want to disable warning 'Not enough inventory'"
+             " when there isn't enough product in stock",
+        string="Check Stock on Sale",
+        default=True
+    )
--- a/sale_disable_inventory_check/models/product_template.py
+++ b/sale_disable_inventory_check/models/product_template.py
@@ -1,0 +1,33 @@
+# Copyright Komit <http://komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    check_stock_on_sale = fields.Selection(
+        selection=[('skip', 'Skip the check'),
+                   ('not_skip', 'Do not skip the check'),
+                   ('defer', 'Defer to the setting on the category')],
+        company_dependent=True,
+        help="'Skip the check' will disable warning 'Not enough inventory' "
+             "when there isn't enough product in stock!\n"
+             "'Do not skip the check' won't disable warning 'Not enough "
+             "inventory' when there isn't enough product in stock!\n"
+             "'Defer to the setting on the category' will use the "
+             "setting from the Internal Category",
+        string="Check Stock on Sale",
+        default="defer",
+    )
+
+    @api.multi
+    def _check_stock_on_sale(self):
+        self.ensure_one()
+        if self.check_stock_on_sale == 'skip':
+            return False
+        elif self.check_stock_on_sale == 'defer':
+            return self.categ_id.check_stock_on_sale
+        else:
+            return True
--- a/sale_disable_inventory_check/models/sale_order_line.py
+++ b/sale_disable_inventory_check/models/sale_order_line.py
@@ -1,7 +1,7 @@
 # Copyright Komit <http://komit-consulting.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import api, models, _
 
 
 class SaleOrderLine(models.Model):
@@ -9,4 +9,12 @@ class SaleOrderLine(models.Model):
 
     @api.onchange('product_uom_qty', 'product_uom', 'route_id')
     def _onchange_product_id_check_availability(self):
-        return {}
+        res = \
+            super(SaleOrderLine,
+                  self)._onchange_product_id_check_availability()
+        if self.product_id \
+                and not self.product_id.product_tmpl_id._check_stock_on_sale():
+            if res.get('warning', {}).get('title') == _(
+                    'Not enough inventory!'):
+                res.pop('warning')
+        return res
--- a/sale_disable_inventory_check/tests/test_disable_inventory_check.py
+++ b/sale_disable_inventory_check/tests/test_disable_inventory_check.py
@@ -8,11 +8,29 @@ class TestDisableInventoryCheck(common.TransactionCase):
     def setUp(self):
         super(TestDisableInventoryCheck, self).setUp()
 
+        self.category_1 = self.env['product.category'].create({
+            'name': 'Category 1',
+            'check_stock_on_sale': True,
+        })
+        self.category_2 = self.env['product.category'].create({
+            'name': 'Category 2',
+            'check_stock_on_sale': True,
+        })
+
         # Create stockable product
         self.product_1 = self.env['product.product'].create({
             'name': 'Product 1',
             'type': 'product',
-            'uom_id': self.env.ref('product.product_uom_unit').id,
+            'uom_id': self.ref('product.product_uom_unit'),
+            'check_stock_on_sale': 'skip',
+            'categ_id': self.category_1.id,
+        })
+        self.product_2 = self.env['product.product'].create({
+            'name': 'Product 2',
+            'type': 'product',
+            'uom_id': self.ref('product.product_uom_unit'),
+            'check_stock_on_sale': 'defer',
+            'categ_id': self.category_2.id,
         })
 
     def test_disable_inventory_check(self):
@@ -23,12 +41,26 @@ class TestDisableInventoryCheck(common.TransactionCase):
             'state': 'draft',
         })
         # Sale product_1 with quantity is 15
-        order_line = self.env['sale.order.line'].new({
+        order_line1 = self.env['sale.order.line'].new({
             'order_id': order.id,
             'product_id': self.product_1.id,
             'product_uom': self.product_1.uom_id.id,
             'product_uom_qty': 15,
         })
-        res = order_line._onchange_product_id_check_availability()
-        self.assertEqual(res.get('warning', {}), {},
-                         "There should not be a warning!!!")
+        res1 = order_line1._onchange_product_id_check_availability()
+        self.assertNotIn('warning', res1, "There should not be a warning!!!")
+        # Sale product_2 with quantity is 15
+        order_line2 = self.env['sale.order.line'].new({
+            'order_id': order.id,
+            'product_id': self.product_2.id,
+            'product_uom': self.product_2.uom_id.id,
+            'product_uom_qty': 15,
+        })
+        res2 = order_line2._onchange_product_id_check_availability()
+        self.assertIn('warning', res2, "Should have returned a warning")
+        self.assertIsInstance(
+            res2['warning'], dict, "Warning should be a dictionary")
+        self.assertIn('title', res2['warning'], "Warning should have a title")
+        self.assertEqual(
+            res2['warning']['title'], 'Not enough inventory!',
+            "There should be a warning!!!")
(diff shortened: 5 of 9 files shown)


novawish:
Can you fix travis error?
image

hieulucky111:
> Can you fix travis error?
> image

I fixed, but travis still faces some problems with the other module.

sale_disable_inventory_check/tests/test_disable_inventory_check.py
```diff
@@ -23,12 +41,23 @@ def test_disable_inventory_check(self):
...
+        # Sale product_2 with quantity is 15
+        order_line2 = self.env['sale.order.line'].new({
+            'order_id': order.id,
+            'product_id': self.product_2.id,
+            'product_uom': self.product_2.uom_id.id,
+            'product_uom_qty': 15,
+        })
+        res2 = order_line2._onchange_product_id_check_availability()
+        self.assertEqual(
+            res2.get('warning', '').get('title', ''),
```
aisopuro: You are defaulting to a string (`''`), but then you are calling `.get()` on it. Since this is a test, you can be more explicit about how exactly it fails:
```python
self.assertIn('warning', res2, "Should have returned a warning")
self.assertIsInstance(res2['warning'], dict, "Warning should be a dictionary")
self.assertIn('title', res2['warning'], "Warning should have a title")
self.assertEqual(res2['warning']['title'], 'Not enough inventory!', "There should be a warning!!!")
```

sale_disable_inventory_check/data/product_check_stock_data.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
nikul-serpentcs: @hieulucky111 `<odoo noupdate="1">`

sale_disable_inventory_check/tests/test_disable_inventory_check.py
```diff
@@ -8,11 +8,29 @@ class TestDisableInventoryCheck(common.TransactionCase):
...
             'name': 'Product 1',
             'type': 'product',
             'uom_id': self.env.ref('product.product_uom_unit').id,
+            'property_is_check_stock': True,
+            'categ_id': self.category_1.id
+        })
+        self.product_2 = self.env['product.product'].create({
+            'name': 'Product 2',
+            'type': 'product',
+            'uom_id': self.env.ref('product.product_uom_unit').id,
```
nikul-serpentcs: @hieulucky111 if you need to direct id then use `self.ref('product.product_uom_unit')`

nikul-serpentcs (changes requested):
netpicking

hieulucky111:
@nikul-serpentcs @aisopuro Could you please to help me to finalize this PR?

sale_disable_inventory_check/models/sale_order_line.py
```diff
@@ -9,4 +9,8 @@ class SaleOrderLine(models.Model):
 
     @api.onchange('product_uom_qty', 'product_uom', 'route_id')
     def _onchange_product_id_check_availability(self):
-        return {}
+        if self.product_id.product_tmpl_id._get_disable_check_stock():
+            return {}
```
aisopuro: The original method doesn't just return a warning, it also sets packaging: https://github.com/odoo/odoo/blob/11.0/addons/sale_stock/models/sale_order.py#L163

You should always allow the super to run and only remove the warning from the result if it's there instead of breaking the call-chain.

Also, you shouldn't assume that some other module might not return a different warning related to a different matter.

```python
result = super(...)
if self.product_id.product_tmpl_id._get_disable_check_stock():
    if result.get('warning', {}).get('title') == _('Not enough inventory!'):
        result.pop('warning')
return result
```

sale_disable_inventory_check/tests/test_disable_inventory_check.py
```diff
@@ -23,12 +41,27 @@ def test_disable_inventory_check(self):
...
+        order_line1 = self.env['sale.order.line'].new({
             'order_id': order.id,
             'product_id': self.product_1.id,
             'product_uom': self.product_1.uom_id.id,
             'product_uom_qty': 15,
         })
-        res = order_line._onchange_product_id_check_availability()
-        self.assertEqual(res.get('warning', {}), {},
+        res1 = order_line1._onchange_product_id_check_availability()
+        self.assertEqual(res1.get('warning', {}), {},
```
aisopuro: A clearer way to express this would be to use `assertNotIn`:
```python
self.assertNotIn('warning', res)
```

aisopuro (changes requested):
Here's a review @hieulucky111 . However I don't believe I have merge access to this repo so I'm not sure if this counts toward mergeability.

jcdrubay (approved):

jcdrubay:
@OCA/core-maintainers @StefanRijnhart @lreficent : could you please help to review? Thanks in advance for your help :)

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,21 @@
...
+    property_is_check_stock = fields.Boolean(
+        company_dependent=True,
+        help="This checkbox disable warning 'Not enough inventory'"
+             " when there isn't enough product in stock",
+        string="Disable Check Stock Message",
+    )
+
+    @api.multi
+    def _get_disable_check_stock(self):
+        return \
```
StefanRijnhart: PEP8 prefers parentheses over line continuations
```
return (
    self.property_is_check_stock or
    self.categ_id.property_is_check_stock or False)

File: sale_disable_inventory_check/models/product_category.py [hunk 7]
```diff
@@ -0,0 +1,15 @@
+# Copyright Komit <http://komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    property_is_check_stock = fields.Boolean(
```
StefanRijnhart: The semantics of this field is not immediately clear to me 'is_check_stock' is not immediately. In any case, I would advise against introducing a double negation: Make it so that a negative value corresponds to Odoo not doing something.

You could rename to 'check_stock_on_sale' to make the name a little clearer (and of course change the default property value).

hieulucky111:
@StefanRijnhart Could you please to help me to finalize PR?

sale_disable_inventory_check/models/product_category.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    check_stock_on_sale = fields.Boolean(
+        company_dependent=True,
+        help="This checkbox disable warning 'Not enough inventory'"
+             " when there isn't enough product in stock",
```
StefanRijnhart, line 13: I have to come back to the field semantics. If the field is called 'Check stock on sale', and this option is active, I would expect stock to be checked on the sale order. I don't think that is what is currently the case.

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+             " when there isn't enough product in stock",
+        string="Disable Check Stock Message",
+        default=True,
+    )
+
+    @api.multi
+    def _check_stock_on_sale(self):
+        return all((
+            self.check_stock_on_sale,
+            self.categ_id.check_stock_on_sale, True))
```
StefanRijnhart: Having a fallback on the category value works less well for boolean fields. It is different for Many2one fields because in that case, the option on the product is either entered in which case it is used, or it is not entered in which case there is a fallback on the category setting. In the case of boolean fields, how is the user to know when the fallback on the category value kicks in?

sale_disable_inventory_check/models/product_category.py
aisopuro: Might be as simple as adding `skip_` to the start of the field name.

sale_disable_inventory_check/models/product_template.py
aisopuro: Yeah, this is a good point. If you want to have the field defer to the category it should maybe be a `Selection` field that enables you to explicitly choose between:
- Skip the check
- Do not skip the check
- Defer to the setting on the category

The field might be allowed to be left empty as well, but then you need to decide what that means. If it means "do nothing different", then that's the same as "Do not skip the check".

sale_disable_inventory_check/models/product_category.py
StefanRijnhart: That is a double negation: activate an option so that something is not happening. What do the usability guys think about that?

sale_disable_inventory_check/models/product_template.py
hieulucky111: i will take into account this comment https://github.com/OCA/sale-workflow/pull/805#discussion_r273815740

hieulucky111:
@StefanRijnhart Could you please to help me to finalize the PR?

jcdrubay:
@aisopuro @StefanRijnhart @novawish Your help would be welcome.

nikul-serpentcs (approved):
Only Code Review LGTM :+1:

sale_disable_inventory_check/models/product_category.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    check_stock_on_sale = fields.Boolean(
+        company_dependent=True,
+        help="Uncheck if you want to disable warning 'Not enough inventory'"
+             " when there isn't enough product in stock",
+        string="Disable Check Stock Message",
```
aisopuro: The `help` and `string` contradict each other: `help` says you should "Uncheck to disable the warning" -> If true, do check. But `string` says "Disable Check Stock message" -> If true, do not check.

I think the `string` should be changed to be "Check Stock on Sale".

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+                   ('defer', 'Defer to the setting on the category')],
+        company_dependent=True,
+        help="'Skip the check' will disable warning 'Not enough inventory' "
+             "when there isn't enough product in stock!\n"
+             "'Do not skip the check' won't disable warning 'Not enough "
+             "inventory' when there isn't enough product in stock!\n"
+             "'Defer to the setting on the category' will base on "
+             "'Disable Check Stock Message' on Product Category",
+        string="Disable Check Stock Message",
+        default=True,
```
aisopuro: This default value seems wrong, shouldn't it be one of the possible selections?

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        selection=[('skip', 'Skip the check'),
+                   ('not_skip', 'Do not skip the check'),
+                   ('defer', 'Defer to the setting on the category')],
+        company_dependent=True,
+        help="'Skip the check' will disable warning 'Not enough inventory' "
+             "when there isn't enough product in stock!\n"
+             "'Do not skip the check' won't disable warning 'Not enough "
+             "inventory' when there isn't enough product in stock!\n"
+             "'Defer to the setting on the category' will base on "
+             "'Disable Check Stock Message' on Product Category",
```
aisopuro: I think this line could be reworded for clarity:
`'Defer to the setting on the category' will base on 'Disable Check Stock Message' on Product Category`
->
`'Defer to the setting on the category' will use the setting from the Internal Category`

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        string="Disable Check Stock Message",
```
aisopuro: This `string` is causing confusing double-negatives with the selections. I think "Check Stock on Sale" would be a clearer field name.

sale_disable_inventory_check/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        default=True,
+    )
+
+    @api.multi
+    def _check_stock_on_sale(self):
+        if any((self.check_stock_on_sale == 'skip',
```
aisopuro: This assumes `self` is a single recordset: you should call `self.ensure_one()` at the start of the method to make that expectation explicit.

Since `self` is a single record, you don't really need to use `any`; in this case I think it would be clearer to just have separate `if`-statements:
```python
if self.check_stock_on_sale == 'skip':
    return False
elif self.check_stock_on_sale == 'defer':
    return self.categ_id.check_stock_on_sale
else:
    return True
```

aisopuro (changes requested):
Here is the latest review.

Again however, I am not sure that my review counts towards mergeability, so I suggest @jcdrubay that you check with the maintainers whether my review actually counts towards mergeability.

hieulucky111:
@aisopuro I have updated the code.

aisopuro:
OK @hieulucky111 , almost there. I would argue that a more reasonable default for `check_stock_on_sale` would be `defer` rather than `not_skip`. The reason is that I think in most cases we want to manage these settings on a per-category basis, not per-product. Sometimes we will have exceptions, but I think the more sensible default is to have the default be to go with the category setting, and then you change that on the product when you want the product to act differently.

hieulucky111:
@StefanRijnhart Please help us to finalize PR, thanks :)

aisopuro:
One weirdness I just noticed working on a different feature in another project: it seems to me that `company_dependent` does not play well with defaults for `Selection` fields. It looks like when you install a module with that kind of field the initial setting of values into the field does not respect the default value, because `company_dependent` overrides the usual default calculation logic.

Can you check whether creating new products actually sets the expected default value? I found that I needed to override the `default_get` method to be able to provide a default value for creation: setting the `default` argument for the field did not work.

hieulucky111:
> One weirdness I just noticed working on a different feature in another project: it seems to me that `company_dependent` does not play well with defaults for `Selection` fields. It looks like when you install a module with that kind of field the initial setting of values into the field does not respect the default value, because `company_dependent` overrides the usual default calculation logic.
> 
> Can you check whether creating new products actually sets the expected default value? I found that I needed to override the `default_get` method to be able to provide a default value for creation: setting the `default` argument for the field did not work.

hieulucky111:
@aisopuro could you please help to check again?, i fixed following your comment.

aisopuro:
Looks good.

jcdrubay:
@StefanRijnhart Could you please approve this PR if you are now ok with it? Thanks :)

StefanRijnhart (approved):

PR #807, opened 2019-03-08
[ADD][12.0] sale_invoice_plan

kittiu:
By standard feature, user can gradually create partial invoices, one by one.
This module add ability to create invoices based on the predefined invoice plan,
either all at once, or one by one.
The plan support both advance invoice and installment invoices.

image


pedrobaeza:
This should be checked by reviewers before merging.

kittiu:
Any comments for this module?

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,200 @@
...
+        ondelete='cascade',
+    )
+    installment = fields.Integer(
+        string='Installment',
+    )
+    plan_date = fields.Date(
+        string='Plan Date',
+        required=True,
+    )
+    type = fields.Selection(
```
nikul-serpentcs: @kittiu IMO I think here change field name it looks like a function

nikul-serpentcs (changes requested):
Improve Code

sale_invoice_plan/models/sale.py
kittiu: @nikul-serpentcs thanks for your comment.

kittiu:
This module has been here a while. Any more comments?

jgrandguillaume (approved):
Functional test: Ok, thanks for the contribution.
:+1:

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+            rec.ip_invoice_plan = rec.use_invoice_plan and \
+                rec.invoice_plan_ids and \
+                len(rec.invoice_plan_ids.filtered(lambda l: not l.invoiced))
+
+    @api.constrains('state')
+    def _check_invoice_plan(self):
+        for rec in self:
+            if rec.state != 'draft':
+                if rec.invoice_plan_ids.filtered(lambda l: not l.percent):
+                    raise UserError(
```
rousseldenis: You should use ValidationError in constrains

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+                        _('Please fill percentage for all invoice plan lines'))
+
+    @api.multi
+    def action_confirm(self):
+        for rec in self:
```
rousseldenis: You could use this :

```suggestion
        if self.filtered(lambda r: r.use_invoice_plan and not r.invoice_plan_ids):
            raise...
```

just to avoid indent

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+    @api.multi
+    def create_invoice_plan(self, num_installment, installment_date,
+                            interval, interval_type, advance):
+        self.ensure_one()
+        self.invoice_plan_ids.unlink()
+        invoice_plans = []
+        Decimal = self.env['decimal.precision']
+        prec = Decimal.precision_get('Product Unit of Measure')
+        percent = round(1.0 / num_installment * 100, prec)
+        percent_last = 100 - (percent * (num_installment-1))
```
rousseldenis, line 66: This could be set in an helper computed field no ?

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+    @api.multi
+    def _compute_new_invoice_quantity(self, invoice):
+        self.ensure_one()
+        if self.last:  # For last install, let the system do the calc.
+            return
+        percent = self.percent
+        for line in invoice.invoice_line_ids:
+            assert len(line.sale_line_ids) >= 0, \
+                'No matched order line for invoice line'
+            order_line = line.sale_line_ids[0]
```
rousseldenis: It's better to use:

```suggestion
            from odoo.fields import first
            order_line = first(line.sale_line_ids)
```

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+            if order_line.is_downpayment:
+                line.quantity = -percent/100  # Always based on 1 unit
+            else:
+                plan_qty = order_line.product_uom_qty * (percent/100)
```
rousseldenis, line 216: Is the plan_qty could be set in an helper field with digit precision as the following comparison should be done with float_compare?

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @api.multi
+    def action_invoice_create(self, grouped=False, final=False):
+        inv_ids = super().action_invoice_create(grouped=grouped, final=final)
+        invoice_plan_id = self._context.get('invoice_plan_id')
+        if invoice_plan_id:
+            plan = self.env['sale.invoice.plan'].browse(invoice_plan_id)
+            invoices = self.env['account.invoice'].browse(inv_ids)
+            invoices.ensure_one()  # Expect 1 invoice for 1 invoice plan
+            plan._compute_new_invoice_quantity(invoices[0])
```
rousseldenis, line 115: IMHO in '_compute_new_invoice_quantity', you should do a write on invoice values.

rousseldenis (changes requested):
Code review. Some comments.

Looks great!

sale_invoice_plan/models/sale.py
kittiu: Sorry, I don't quite understand. You mean moving 
`plan.invoice_ids += invoice`
inside _compute_new_invoice_quantity()?

rousseldenis: No, maybe I'm at the wrong place to do the comment.

But here you are in an action (launched by a button I suppose).

that's why you have to use write() in '_compute_new_invoice_quantity' instead of doing 'line.quantity = plan_qty'

sale_invoice_plan/models/sale.py
kittiu: @rousseldenis sorry, I don't understand what you mean about -> **plan_qty could be set in an helper field with digit precision**

sale_invoice_plan/models/sale.py
kittiu: Done

sale_invoice_plan/models/sale.py
kittiu: @rousseldenis sorry, I don't understand what you mean.

sale_invoice_plan/models/sale.py
kittiu: Done

sale_invoice_plan/models/sale.py
kittiu: Done.

AaronHForgeFlow:
It works fine. It would be great to include another option that allows to invoice every X days (X is entered by the user).

It also would be awesome to do partial deliveries according to a plan. But that would go to another module. Thank you @kittiu

sale_invoice_plan/models/sale.py
kittiu: Done, I changed to write()
By the way, I also curious. In what situation we should not use record.field = value?
Thank you

kittiu:
> It works fine. It would be great to include another option that allows to invoice every X days (X is entered by the user).

@aheficent thanks for your comment. Currently, we can set X days (month, year) as interval. You are looking for something different?
image

kittiu:
> It also would be awesome to do partial deliveries according to a plan. But that would go to another module. Thank you @kittiu

Yes, that might be sale_delivery_plan ? :)

AaronHForgeFlow:
> sale_delivery_plan

Yes :)

AaronHForgeFlow:
> > It works fine. It would be great to include another option that allows to invoice every X days (X is entered by the user).
> 
> @aheficent thanks for your comment. Currently, we can set X days (month, year) as interval. You are looking for something different?
> image

That's good, did not notice of that on runbot. Thanks.

AaronHForgeFlow (approved):
:+1:

sale_invoice_plan/models/sale.py
rousseldenis: Usually, such affectation should only be used in compute and inverse functions as the write is done after (with recomputes and so on). When it is launched through an action or,... you are not sure that a write will be done(and good recomputes on fields that depends on the one you've modified).

sale_invoice_plan/models/sale.py
rousseldenis: It was a minor remark but generally it is better to store some intermediate computed values in a field to lower the function code and to make it more readable.

sale_invoice_plan/models/sale.py
rousseldenis: The field 'line.quantity' is a field with decimal precision. You must at least compare with 'float_compare'.

The remark about 'helper' field on 'plan_qty' is quite the same I did here above.

It's sometimes useful to make a computed field with a part of a function to make it more readable and sometimes reuse the field elsewhere.

sale_invoice_plan/models/sale.py
kittiu: Thank you @rousseldenis , I think I understand your point now :)
My design decision so far is to minimize the new field in sale.order (I only have 1 new column which is `use_invoice_plan`).
From what I see from this function, the useful compute field would be `percent` and `percent_last`. But to compute it, I need to store `num_installment` permanently in sale.order. I think it might not worth it yet? In the future, if there is the need of `num_installment` then it will be sound. WDYT?

sale_invoice_plan/models/sale.py
kittiu: Done use float_compare(), thank you.
For the plan_qty as helper, as I try to look at, may be not very easy to do at this point.

rousseldenis (approved):
Code review

AaronHForgeFlow:
Hi @nikul-serpentcs  IMHO your comments have been attended here, can you update your review please?

nikul-serpentcs (approved):
Code Review LGTM :+1:

mohamedhagag:
just did a functional test and working fine, should we merge ?

kittiu:
+1 :)

rousseldenis:
@OCA/core-maintainers

kittiu:
Hi @OCA/core-maintainers, this addon has been used and been pending here quite sometime. Would appreciate the merge :)

mohamedhagag:
SWM ? :)

PR #756, opened 2018-12-04
[12.0][MIG] Migrated partner_prospect to v12

sudhir-erpharbor:
Please review and provide feedback.

PR #808, opened 2019-03-08
[12.0][MIG] partner_prospect: Migration to 12.0

randall-vx:


pedrobaeza:
I'm seeing that there's other migration in progress in #756. Why didn't you check?

sale_partner_prospect/models/res_partner.py
```diff
@@ -0,0 +1,37 @@
...
+                    'child_ids.invoice_ids'))
+            partner.prospect = (
+                not sale_ids.filtered(
+                    lambda r: r.state not in
+                    ('draft', 'sent', 'cancel')) and
+                not invoice_ids.filtered(
+                    lambda r: r.type in ('out_invoice', 'out_refund')))
+
+    prospect = fields.Boolean(
+        string='Prospect', compute='_compute_prospect', default=False,
```
Nikul-Chaudhary, line 36: @rcastro-tyc  no need to `default=False`

randall-vx:
Hi @pedrobaeza, Yes, I reviewed it before making the PR, I decided to do it again only because of the date the PR was made and because I needed it for a database.

Regards,

pedrobaeza:
But the PR is still active and the contributor is attending reviews, so no need to superseed the other. Please indicate in the other PR what are the remaining points and if the author doesn't answer, then we can superseed that one with other, but not this way. Closing this one.

PR #811, opened 2019-03-13
[11.0][ADD] product_template_tag_sale: Set menu in sales to manage product tags.

alvaro-domatix:
This addon places a menu in sales configuration menu which allows you to manage product tags.


pedrobaeza:
Please avoid plurals in module name

PR #814, opened 2019-03-19
[12.0][FIX] To approve quotation display & Confirm on creation

tschanzt:
- Add the state to_apporve to quotations instead of sales
- dont allow managers to confirm directly when creating
- allow cancelation in to_approve state

### DIFF
--- a/sale_double_validation/models/sale.py
+++ b/sale_double_validation/models/sale.py
@@ -45,8 +45,8 @@ class SaleOrder(models.Model):
     @api.model
     def create(self, vals):
         obj = super().create(vals)
-        if obj.is_to_approve():
-            obj.state = 'to_approve'
+        if not obj.is_to_approve():
+            obj.state = 'draft'
         return obj
 
     @api.multi
--- a/sale_double_validation/models/sale.py
+++ b/sale_double_validation/models/sale.py
@@ -19,10 +19,11 @@ class SaleOrder(models.Model):
                 exists = True
         if not exists:
             selection.insert(0, ('to_approve', _('To Approve')))
-        for field in self._fields:
-            act_field = self._fields.get(field)
+        for act_field in self._fields.values():
 
-            if act_field.states and all(x in act_field.states.keys() for x in ['draft', 'sent']):
+            if act_field.states and all(
+                x in act_field.states.keys() for x in ['draft', 'sent']
+            ):
                 act_field.states = {'to_approve': [('readonly', False)]}
 
     @api.multi
@@ -57,4 +58,9 @@ class SaleOrder(models.Model):
     @api.depends('state')
     def _compute_type_name(self):
         for record in self:
-            record.type_name = _('Quotation') if record.state in ('to_approve', 'draft', 'sent', 'cancel') else _('Sales Order')
+            record.type_name = _('Quotation') if record.state in (
+                'to_approve',
+                'draft',
+                'sent',
+                'cancel'
+            ) else _('Sales Order')
--- a/sale_double_validation/models/sale.py
+++ b/sale_double_validation/models/sale.py
@@ -37,11 +37,18 @@ class SaleOrder(models.Model):
             precision_rounding=self.currency_id.rounding) <= 0
 
     @api.multi
+    def can_skip(self):
+        self.ensure_one()
+        if self.company_id.so_double_validation_manager_skip:
+            if self.user_has_groups('sales_team.group_sale_manager'):
+                return True
+        return False
+
+    @api.multi
     def is_to_approve(self):
         self.ensure_one()
         return (self.company_id.so_double_validation == 'two_step' and
-                self.is_amount_to_approve() and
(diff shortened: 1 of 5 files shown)


leemannd (commented):
Seems legit.
Could you at least add a comment into your PR to tell people with a few details the points you are trying to solve?

p-tombez (commented):
Seems legit, but flake8 is not happy

sale_double_validation/models/sale.py
```diff
@@ -17,6 +19,11 @@ def _setup_fields(self):
                 exists = True
         if not exists:
             selection.insert(0, ('to_approve', _('To Approve')))
+        for field in self._fields:
+            act_field = self._fields.get(field)
```
mpanarin: why not use `_fields.items()` or `_fields.values()`?

mpanarin (changes requested):

tschanzt:
@mpanarin addressed. also hopefully pep8 is happy now :pray:

mpanarin (approved):

tschanzt:
@simahawk im looking for reviewers. im sure you're excited to :smile:

simahawk:
@tschanzt sorry, I have no clue :man_shrugging: 
@p-tombez @leemannd ping :)

rousseldenis:
@tschanzt Do you plan to add a few tests to highlight changes ?

tschanzt:
@rousseldenis i didn't till now :smile: , but i can add some.

marylla:
Hi @tschanzt 
I tested your PR in the runbot and everything worked fine so far, though I have some thoughts about it.

I guess, what you want is a real "draft" state which is unfortunately occupied by the quotation state.
But I think the naming is kind of misleading, especially the name of the button "approve order". 
It actually isn't an order yet and it is also not really an approvement...
Since the "save"-button also does the state conversion from "to approve" to "quotation", you actually don't need the "approve order"-button, do you?

sale_double_validation/models/sale.py
```diff
@@ -17,6 +19,12 @@ def _setup_fields(self):
                 exists = True
         if not exists:
             selection.insert(0, ('to_approve', _('To Approve')))
+        for act_field in self._fields.values():
```
daramousk, line 22: Use `self.fields_get(attributes=['readonly'])`

sale_double_validation/models/sale.py
```diff
@@ -38,10 +46,21 @@ def is_to_approve(self):
     @api.model
     def create(self, vals):
         obj = super().create(vals)
-        if obj.is_to_approve():
-            obj.state = 'to_approve'
+        if not obj.is_to_approve():
```
daramousk, line 56: Why was this reversed?

sale_double_validation/models/sale.py
```diff
@@ -1,12 +1,14 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
-from odoo import api, models, _
+from odoo import api, models, fields, _
 from odoo.tools import float_compare
 
 
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
 
+    state = fields.Selection(default='to_approve')
```
daramousk, line 10: Wouldn't this set the default state to `to_approve` for all the quotations/sale orders? Even for those that `is_to_approve() == False`?

daramousk (commented):
I am not really sure about this, this does not seem to be a fix, rather a change of functionality. And things are already complicated when we add a new state to a document.

Perhaps it would be best to have the states not changed and have a different way of approving them.

tschanzt:
@marylla it should only be set to draft if double valdiation is not needed otherwise it will stay in 'to_approve'

sale_double_validation/models/sale.py
tschanzt: yes but if is_to_approve() == False it will be moved to draft on the create

rousseldenis:
@tschanzt Do you plan to fix this ?

sale_double_validation/models/sale.py
tschanzt: we want to_approve to be the default state and move to draft if it doesn't need to be approved. Reason is that in draft state there are possibilities to move it to a different state directly from creation form if we have draft as the default

tschanzt:
@rousseldenis im checking on it.

tschanzt:
@rousseldenis i checked on runbot and it was working as i expected it.  But i think the issue might have been that sale manager skip the process, so if you test with admin it will skip it. I added a flag now if sale managers should be allowed to skip the process or if they still should go through it.

sale_double_validation/models/sale.py
rousseldenis: IMHO, changing this default behaviour could lead to errors (e.g.: you can see in tests). This change should be set by option.

rousseldenis:
Closing this as too old. Please feel free to reopen it if needed.

PR #815, opened 2019-03-19
sale_order_line_price_history: add price history in sale order line

ernestotejeda:
App price history of a product from a sale order line and set one of
this old prices in the sale order line.

Cc @Tecnativa


victor-ditay:
Travis is failing but not on the module you are doing now.
The migration of `sale_automatic_workflow_payment_mode` has all test passed but was done before this PR https://github.com/odoo/odoo/pull/31076 so we might need fix UT

cc @pedrobaeza

sale_order_line_price_history/README.rst
```diff
@@ -0,0 +1,101 @@
...
+    :target: https://translation.odoo-community.org/projects/sale-workflow-11-0/sale-workflow-11-0-sale_order_line_price_history
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/167/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of Sales to allow you to see the price
+history of a product from a sale order line and set one of this old prices in
```
pedrobaeza: s/this/these

sale_order_line_price_history/README.rst
```diff
@@ -0,0 +1,101 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales -> Quotations* and select a Quotation.
+#. Click on the new clock button in one of the sale order lines.
+#. A pop-up will open and you will see the *price history* for the product of
+   the sale order line and for the partner of the sale order.
+#. You can select other partner or left empty to see the price history for all
```
pedrobaeza: s/left/leave it

sale_order_line_price_history/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales -> Quotations* and select a Quotation.
+#. Click on the new clock button in one of the sale order lines.
+#. A pop-up will open and you will see the *price history* for the product of
+   the sale order line and for the partner of the sale order.
```
pedrobaeza: Call it customer instead of partner

sale_order_line_price_history/README.rst
```diff
@@ -0,0 +1,101 @@
...
+   partners.
+#. You can also set the price of one of the price history lines in the sale
+   order line. To do that, you have to open the price history line desired and
+   click the smart button named *Set*.
+
+Known issues / Roadmap
+======================
+
+* The *Customer* column of the *price history lines* list is not hidden when
+  there is a specific client selected because odoo doesn't allow
```
pedrobaeza, line 56: s/a/an

pedrobaeza (changes requested):
Unrelated error should be fixed now. Code review only for now.

pedrobaeza (changes requested):
Comments after functional review:

- On the wizard, you have to include the product in readonly as reference.
- Limit the number of entries to a reasonable amount (20?) and put in the roadmap to make this number configurable.
- Add to the roadmap that the button is not clickable until you save the record.
- Remove from the history the own line from the current sales order.
- The button for setting the price is not visible right now, and it shouldn't be if the sales order is confirmed.
- The date you must put is the sales order date, not the creation date.

sale_order_line_price_history/README.rst
ernestotejeda: done

sale_order_line_price_history/README.rst
ernestotejeda: done

sale_order_line_price_history/README.rst
ernestotejeda: done

sale_order_line_price_history/README.rst
ernestotejeda: In that case it must be *a specific*, because the article goes before a consonant.

ernestotejeda:
> Comments after functional review

@pedrobaeza it's done

pedrobaeza (approved):

sale_order_line_price_history/wizards/sale_orderLine_price_history.py
```diff
@@ -0,0 +1,96 @@
...
+    def _onchange_partner_id(self):
+        self.line_ids = False
+        domain = [
+            ("product_id", "=", self.product_id.id),
+            ("state", "in", ["sale", "done"]),
+        ]
+        if self.partner_id:
+            domain += [
+                "|",
+                ("order_id.partner_id", "=", self.partner_id.id),
```
pedrobaeza: I have just seen that you have to put directly (and only this that includes all):
```python
("order_id.partner_id", "child_of", self.partner_id.commercial_partner_id.ids),
```

pedrobaeza (changes requested):

pedrobaeza (approved):

victor-ditay (approved):

PR #816, opened 2019-03-22
[10.0][FIX] sale_triple_discount: Cache management during onchange

SimoRubi:
Behavior before this commit:
1. Create an invoice with one line: qty 1, unit price 100, discount 0, tax 22% -> tax amount is 22 OK
2. Edit discount to 50 -> tax amount is still 22 KO
3. Edit the discount back to 0 -> tax amount becomes 11 KO

Behavior after this commit:
1. Create an invoice with one line: qty 1, unit price 100, discount 0, tax 22% -> tax amount is 22 OK
2. Edit discount to 50 -> tax amount is 11 OK
3. Edit the discount back to 0 -> tax amount becomes 22 OK

### DIFF
--- a/sale_triple_discount/__manifest__.py
+++ b/sale_triple_discount/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sale Triple Discount',
-    'version': '10.0.1.0.1',
+    'version': '10.0.1.0.2',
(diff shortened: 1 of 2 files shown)


SimoRubi:
My bad, this should be in account_invoice_triple_discount. Closing

PR #812, opened 2019-03-13
[12.0][WIP][MIG] sale_exception

ljimenezsidoo:
Depends from this PR https://github.com/OCA/server-tools/pull/1395

PR #823, opened 2019-04-09
[12.0][ADD] sale_exception_credit_limit

ivantodorovich:
It depends on https://github.com/OCA/sale-workflow/pull/812
This module is moved from https://github.com/ingadhoc/sale and migrated to `12.0`
Cleaned the history a little bit.
@jjscarafia 

It adds a sale_exception rule that checks the partner credit limit and makes sure it doesn't exceed it.


pedrobaeza:
This feature is already developed through other method in https://github.com/OCA/credit-control/. Both can coexist of course, but I mention it for alternatives. I consider the other more powerful.

ivantodorovich:
Didn't know about that one, thanks!
Personally, I like that this one depends on `sale_exception`/`base_exception`

pedrobaeza:
If you use that flow for other things like e-commerce, it makes sense to use the same logic for credit limit. If not, the good of other one (without demeriting this one), is the flexibility for defining which things count for the limit, and where to block.

ivantodorovich:
Will definitely check that one out, too.

kittiu:
@ivantodorovich I would like to do a full test of this PR. But not sure on the status after the comment above. Is this ready for review? Also it seem to depend of module that is not merged yet?

sale_exception_credit_limit/views/res_partner_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                <field name="credit_limit" readonly="1"/>
+            </group>
+        </field>
+    </record>
+
+    <!-- Make editable for user_edit_credit_limit  -->
+    <record id="view_credit_editable_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="view_credit_readonly_partner_form"/>
+        <field name="groups_id" eval="[(6, 0, [ref('credit_config')])]"/>
```
nikul-serpentcs, line 19: @ivantodorovich  should be `ref('module_name.credit_config', False)`

ivantodorovich: Hi @nikul-serpentcs that's actually not necessary. That group is defined in this same module.

Also if you use ref(.., False) you would be silently ignoring the exception if the xmlid is not found. Why would you want to do that?

ivantodorovich:
> @ivantodorovich I would like to do a full test of this PR. But not sure on the status after the comment above. Is this ready for review? Also it seem to depend of module that is not merged yet?

Hi @kittiu, this PR depends on sale_exception. At the time that module was not migrated yet, I've just rebased this branch, so it's ok to review now

ivantodorovich:
BTW travis is failing but it looks like it's an issue with the `sale_order_action_invoice_create_hook` module

sale_exception_credit_limit/views/res_partner_views.xml
nikul-serpentcs: > Also if you use ref(.., False) you would be silently ignoring the exception if the xmlid is not 
> found. Why would you want to do that?


If `xmlid is not found` then silently ignoring the exception That Good, No need to False/ `ref('module_name.credit_config', False)`.

> that's actually not necessary. That group is defined in this same module.

Yes, I know but if you put `module_name` Is Good For us.

 Otherwise, Code Look Good.

rousseldenis:
@ivantodorovich Is this still applicable?

PR #826, opened 2019-04-15
[11.0][IMP] sale_order_secondary_unit: Add security group

sergio-teruel:
cc @Tecnativa
This PR depends of https://github.com/OCA/product-attribute/pull/445

### DIFF
--- a/sale_order_secondary_unit/migrations/11.0.1.0.2/post-migration.py
+++ b/sale_order_secondary_unit/migrations/11.0.1.0.2/post-migration.py
@@ -1,0 +1,23 @@
+#  Copyright 2019 Tecnativa - Sergio Teruel
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    # Remove old security group_uom from product template view because now has
+    # a new group for secondary units
+    group_uom = env.ref('product.group_uom')
+    view_product_template = env.ref(
+        'sale_order_secondary_unit.product_template_form_view')
+    view_product_template.write({
+        'groups_id': [(3, group_uom.id)]
+    })
+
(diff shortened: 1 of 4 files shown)


sale_order_secondary_unit/migrations/11.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    # Remove old security group_uom from product template view because now has
+    # a new group for secondary units
+    group_uom = env.ref('product.group_uom')
+    view_product_template = env.ref(
+        'sale_order_secondary_unit.product_template_form_view')
+    view_product_template.write({
+        'groups_id': [(3, group_uom.id)]
```
pedrobaeza, line 14: Isn't enough with putting in the XML `groups_id` empty?

sergio-teruel: I had two options, one is puting in the xml groups_id = [(5,0), (4, ref('new_group'))] and the other in a post_migration remove the old security group. I have chosen the second.

pedrobaeza: But this technique on views seems a bit weird, and most being noupdate=0. Even more, `groups_id` in views doesn't use `(4,` terminology if I don't remember bad, but directly use the ref of the group.

sergio-teruel:
How can I restart the CI?

pedrobaeza:
I have forced a rebuild on runbot

sale_order_secondary_unit/migrations/11.0.1.0.2/post-migration.py
pedrobaeza: Let's continue this way as I'm not sure.

pedrobaeza (approved):

PR #829, opened 2019-04-23
[12.0][MIGR] sale order revision

renatonlima:


pedrobaeza:
There's already #766. Please check that one instead.

PR #762, opened 2018-12-12
[12.0][ADD] sale_timesheet_hook: adds hooks in sale_timesheets

alexey-pelykh:
This module adds hook points in `sale_timesheets` module in order to add more
flexibility in the billing types for the creation of *Sale Orders*.

Although the original code has been modified a bit, the logic has been
respected and the method still does exactly the same.

Technical Enhancements:
 * `account.analytic.line` extendability improvements:
   * `_get_valid_so_line_ids` overridable
 * `project.create.sale.order` extendability improvements:
   * `action_create_sale_order` refactor, reported in odoo/odoo#31047.

PR #832, opened 2019-04-27
[12.0][ADD] sale_timesheet_task_exclude: exclude specific task from Sale Order

alexey-pelykh:
Allows to exclude specific _Task_ and all _Timesheets_ tracked towards it from
Sale Order.

This functionality is not available in Odoo, reported in odoo/odoo#31042.


alexey-pelykh:
@pedrobaeza @rafaelbn @victormmtorres as discussed, extracted Task exclusion feature to separate PR from #762

tbaden (approved):
code review: lgtm

rafaelbn (commented):
Hi @alexey-pelykh ,

THANKS and great! Just one thing about hiding standard Odoo fields.

- Please don't delete Sale Order's Smart button in the task. It is not necessary.
- Also the field sale order item is not necessary to delete.
- It works fine and even you don't hide, it works, timesheets are excluded. That is the purpose.

dont delete smartbutton


- And in fact they are necessary to have. Why?

Because is information that a Project manager need to have to see from the COST point of view. This module and this functionality makes that the timesheets won't be taken in the "Delivered Quantity" in the Sale Order, so won't be invoiced. And deleting that don't change any functionality but if you have a task "excluded" from the sale order and the timesheet on that task grow and grow you want to check quickly how the SO goes event that task is exclude.

Because is better maintainability talking not to hide any standard Odoo field.

One thing to be taken in care and maybe we should comment in readme is that the Sale Order Analysis will be "wrong" or will not give all the information. Because, is it possible to show delivered qty vs to invoice qty, see:

delivered vs to invoice

Let me know

Thanks again

rafaelbn:
This PR comes from https://github.com/OCA/sale-workflow/pull/762

rafaelbn:
sale_timesheet_exclude_task --> sale_timesheet_task_exclude ( there are not written rules for names but @pedrobaeza usually have some rules to give coherence in all module so with the name you know something like Mnemonic code )

alexey-pelykh:
Totally agree, maybe we should finally write those rules down? :)

alexey-pelykh:
@rafaelbn this module does not hide those smart button or column, it only adds the checkbox field.

Regarding the "Qty Delivered" - yes, unfortunately it's affected since it counts what has been delivered by Sale Order Lines, but "Excluded from Sale Order" have no Sale Order Lines. Technically speaking, "Qty Delivered" is a property of Sale Order Line, and should've been read as "Qty Delivered according to this Sale Order Line". What should be the best wording to explain this?

alexey-pelykh:
("Sale Order" smart button is hidden as side-effect of `billable_type` being `no`)

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        self.SudoAccountAnalyticLine = self.AccountAnalyticLine.sudo()
+        self.ProductProduct = self.env['product.product']
+        self.SudoProductProduct = self.ProductProduct.sudo()
+        self.SaleOrder = self.env['sale.order']
+        self.SudoSaleOrder = self.SaleOrder.sudo()
+        self.SaleOrderLine = self.env['sale.order.line']
+        self.SudoSaleOrderLine = self.SaleOrderLine.sudo()
+        self.ProjectCreateSaleOrder = self.env['project.create.sale.order']
+
+    def test_1(self):
```
victor-ditay: This is not a good name, please think one that reflects what you are trying to prove

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        account = self.SudoAccountAccount.create({
```
victor-ditay, line 43: Why don't you include on the setup all the preparation? I mean to create data is not what you want to test, right?

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        timesheet = self.SudoAccountAnalyticLine.create({
+            'project_id': task.project_id.id,
+            'task_id': task.id,
+            'name': 'Entry #1',
+            'unit_amount': 1,
+            'employee_id': employee.id,
+        })
+        self.assertTrue(timesheet.so_line)
+        self.assertEqual(task.billable_type, 'task_rate')
+
```
victor-ditay, line 117: here it looks to me should end a test

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        task.exclude_from_sale_order = True
```
victor-ditay, line 118: here is a new test as you are changing something you want to assert later

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        self.assertEqual(task.billable_type, 'no')
+        self.assertFalse(timesheet.so_line)
+
+        task.exclude_from_sale_order = False
```
victor-ditay, line 122: New test?

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
```diff
@@ -0,0 +1,129 @@
...
+        self.assertEqual(task.billable_type, 'task_rate')
+        self.assertTrue(timesheet.so_line)
+
+        sale_order.action_invoice_create()
```
victor-ditay, line 126: New test

victor-ditay (commented):
In my opinion the test should be divided and choose a name that reflects what is the purpose of test

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: Yes, but it's a data local to test-case, not test-suite.

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: I'm checking behaviour of setting and unsetting the flag, there's no reason to duplicate scenario

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: Why? I'm always open for suggestions (in code form)

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: Nope, behaviour testing

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: Also no, behaviour testing

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
alexey-pelykh: Offer a name :)

victor-ditay: `test_exclude_from_sale_order` ?

alexey-pelykh: Sounds good to me, changed

sale_timesheet_task_exclude/tests/test_sale_timesheet_exclude_task.py
victor-ditay: Sorry if not clear what is the purpose of my comment but I will try. Here you use Transaction test kind and just one test inside so that means you will not experience any problem but imagine you have three or more test and all of them needs this local data.

Each test will only share the setup and runs on it's own transaction so any change will not affect the others so normally you could add all the common data or setup on this constructor.

If you read my other comments I trying to convince you to split on more than one test with proper name so any other will easily understand how to maintain this module :)

cc @pedrobaeza @Yajo

alexey-pelykh: @victormmtorres when/if this would be needed for more than one test case (as it was before), it data will be either copied to complement local adjustments or will be shared if can be shared. At this point, I see no justification of which data to share or not, so such re-arrangement of code is premature

rafaelbn:
@alexey-pelykh 

> ("Sale Order" smart button is hidden as side-effect of `billable_type` being `no`)

Then why you add a new field `exclude_from_sale_order`? Would be better to show as editable billable_type? (As you want to do also in projects in the "enhance")

I do not want to despair, really, is that I have questions and I prefer to ask them :smile: :thinking:

alexey-pelykh:
@rafaelbn because `billable_type` is a computed field :) That depends on `sale_line_id` and project `billable_type`

rafaelbn (approved):
Functional review :+1: 

@alexey-pelykh  please add the know issues section with:

"Important: The use of this module makes that the hours of timesheets in the tasks excluded, will not appear or taken in consideration as "Quantity Delivered" in the Sales Order Lines and Sale Order Reports"

pedrobaeza:
For simplicity, I think this module should depend (or directly integrated) on `sale_timesheet_line_exclude`, that marks the timesheets of a line as excluded when you mark the task (or viceversa).

alexey-pelykh:
@pedrobaeza you mean just to set flags on timesheets? If yes - it won't work like that. And 100% it would not simplify things since will mix up different levels of hierarchy.

alexey-pelykh:
@rafaelbn done

alexey-pelykh:
@pedrobaeza removed spaces, optimized code

pedrobaeza:
I don't really get why you need this module, as if you want to exclude it, simply remove the link with the sales order line, isn't it?

alexey-pelykh:
It's not the case for "Per Employee" billing type, I've shown this to Rafael - in that case that so_line of task is useless (not used at all!)

sale_timesheet_task_exclude/README.rst
```diff
@@ -0,0 +1,94 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+This feature proves itself useful for *By Employee* billing approach, when
```
pedrobaeza: Move this to the main description.

sale_timesheet_task_exclude/README.rst
```diff
@@ -0,0 +1,94 @@
...
+a any timesheet entry tracked towards a specific task need to be excluded
```
pedrobaeza: remove initial `a`

sale_timesheet_task_exclude/README.rst
```diff
@@ -0,0 +1,94 @@
...
+from the Sale Order.
+
+# Go to *Project > All Tasks*
+# Open specific task for editing
+# Check *Exclude From Billing*
+
+Known issues / Roadmap
+======================
+
+Important: The use of this module makes that the hours of timesheets in the
```
pedrobaeza: This should be put in a bullet list (although only one point now).

pedrobaeza (changes requested):

alexey-pelykh:
@pedrobaeza https://github.com/odoo/odoo/blob/2efcefb50410e234faf777bdf88d879368907915/addons/sale_timesheet/models/project.py#L218 FYI

sale_timesheet_task_exclude/README.rst
alexey-pelykh: Done

sale_timesheet_task_exclude/README.rst
alexey-pelykh: Fixed

sale_timesheet_task_exclude/README.rst
alexey-pelykh: Fixed

pedrobaeza (approved):

pedrobaeza:
My dilemma now is why putting this in OCA/sale-workflow instead on OCA/hr-timesheet. The question is that repository is very-crowded and so transversal that everything can go here. The other one instead is smaller and it's tightly related to these modules. We can rename it to `OCA/timesheet` if the `hr` part scares you.

alexey-pelykh:
@pedrobaeza name of modules would still be `sale_timesheet_*`, but I agree that we should then rename `hr-timesheet` to `timesheets` and move it there - as long as it's not going to take ages :)

pedrobaeza:
No, the renaming has been already done: https://github.com/OCA/timesheet, and I'll respect the reviews here.

pedrobaeza:
I have issued also a mail to contributors list about this renaming.

alexey-pelykh:
Superseded by https://github.com/OCA/timesheet/pull/243

PR #833, opened 2019-04-29
[11.0][FIX] sale_order_price_recalculation: set discount from pricelist with discount_policy without_discount

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_fixed_discount/models/sale_order.py
+++ b/sale_fixed_discount/models/sale_order.py
@@ -39,8 +39,10 @@ class SaleOrderLine(models.Model):
 
     @api.onchange('discount')
     def _onchange_discount(self):
+        res = super(SaleOrderLine, self)._onchange_discount()
         if self.discount:
             self.discount_fixed = 0.0
+        return res
 
     @api.onchange('discount_fixed')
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Yes, very needed with that pricelist policy

rafaelbn:
@sergio-teruel there is nor runbot and travis is :red_circle: could you please check to test? Thanks

sale_fixed_discount/models/sale_order.py
```diff
@@ -39,6 +39,7 @@ class SaleOrderLine(models.Model):
 
     @api.onchange('discount')
     def _onchange_discount(self):
+        super(SaleOrderLine, self)._onchange_discount()
```
pedrobaeza: `res = super(...)`

and return that value at the end (it can contain a domain)

sergio-teruel: Are you sure?

sergio-teruel: done!!..

pedrobaeza:
Let's merge this for fixing the annoying bug

PR #834, opened 2019-04-29
[12.0][ADD] sale_timesheet_line_exclude: exclude Timesheet Line from Sale Order

alexey-pelykh:
Allows to exclude specific _Timesheet_ line from Sale Order.

This functionality is not available in Odoo, reported in odoo/odoo#31043.


alexey-pelykh:
@pedrobaeza @victormmtorres @rafaelbn extracted the line exclude feature from #762

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,107 @@
...
+    _inherit = 'account.analytic.line'
+
+    exclude_from_sale_order = fields.Boolean(
+        string='Exclude from Sale Order',
+        help=(
+            'Checking this would exclude this timesheet entry from Sale Order'
+        ),
+    )
+
+    @api.onchange('task_id', 'employee_id')
```
victor-ditay, line 18: Why not add here the other condition `exclude_from_sale_order` as you are doing the same code inside?

https://github.com/OCA/sale-workflow/pull/834/files#diff-6b07d3f063d79d29a8834694cdbbf559R23

alexey-pelykh: Good point, I'm not sure how it's better to proceed here

cc: @pedrobaeza

sale_timesheet_line_exclude/README.rst
```diff
@@ -0,0 +1,97 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+Excluding Timesheet line from Sale Order
```
pedrobaeza: Remove this useless subtitle

sale_timesheet_line_exclude/models/account_analytic_line.py
pedrobaeza: I don't see why you need this onchange

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,107 @@
...
+    def _onchange_task_id_employee_id(self):
+        if self.project_id:  # timesheet only
+            self.so_line = self._timesheet_get_sale_line()
+
+    @api.onchange('exclude_from_sale_order')
+    def _onchange_exclude_from_sale_order(self):
+        if self.project_id:  # timesheet only
```
pedrobaeza, line 29: If you change `project_id` to `False`, standard code remove the sales order line?

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,107 @@
...
+                self.env['hr.employee'].sudo().browse(values['employee_id'])
+            ) if values and 'employee_id' in values else self.employee_id,
+        }
+
+    @api.multi
+    @api.depends('exclude_from_sale_order')
+    def _compute_timesheet_invoice_type(self):
+        result = super()._compute_timesheet_invoice_type()
+
+        for timesheet in self:
```
pedrobaeza: `self.filtered('project_id')` and you avoid several line codes

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,107 @@
...
+            # Skip non-timesheet entries
+            if not timesheet.project_id:
+                continue
+
```
pedrobaeza: You know that I don't like too much these useless empty lines around the code.

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,107 @@
...
+                so_line = timesheet._timesheet_get_sale_line()
+                result[timesheet.id].update({
+                    'so_line': so_line.id if so_line else False,
+                })
+
+        return result
+
+    @api.multi
+    def write(self, values):
+        if 'exclude_from_sale_order' in values and self.filtered(
```
pedrobaeza: Why not doing this in an `@api.constrains('exclude_from_sale_order')`?

pedrobaeza (changes requested):

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Nope, see https://github.com/odoo/odoo/blob/2efcefb50410e234faf777bdf88d879368907915/addons/sale_timesheet/models/account.py#L41

sale_timesheet_line_exclude/README.rst
alexey-pelykh: Sure

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Thanks!

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Removed

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: To override parent class behaviour?

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: To be consistent with how it's done in base module, but can redo that to constraint if you think it's worth

pedrobaeza: Yes, it's clearer

sale_timesheet_line_exclude/models/account_analytic_line.py
pedrobaeza: All of them?

sale_timesheet_line_exclude/models/account_analytic_line.py
pedrobaeza: I think that should be fixed upstream then

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Most of them

pedrobaeza: OK, I'm content with that

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Yet it won't be accepted in v12 anyhow, so this code should be consistent with base?

pedrobaeza: Not sure... manual analytic account lines without project_id are filled with sales order line?

alexey-pelykh: It will be erased on `write()` anyhow, since there can not be task_id without project. I can rewrite it to explicitly set so_line to False if there's no project while in UI (in onchange)

pedrobaeza: OK, understood. I think there's no need for that and it's not related to this module.

alexey-pelykh: Agreed

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: @pedrobaeza any feedback on if to join the methods? I still think it's better to keep overriding an overriding

pedrobaeza: But still don't understand why you need to override this behavior. Please always that you add controversial code, you must document it with docstring or comment in code.

alexey-pelykh:
```
    @api.onchange('task_id', 'employee_id')
    def _onchange_task_id_employee_id(self):
        """Override implementation in sale_timesheet to call
        _timesheet_get_sale_line() instead of resolving so_line in-place"""
        if self.project_id:  # timesheet only
            self.so_line = self._timesheet_get_sale_line()
```

alexey-pelykh: added description

pedrobaeza: OK, understood, and my final call is this code:
```python
if self.project_id:
    return self._onchange_exclude_from_sale_order()
return super()._onchange_task_id_employee_id()
```

alexey-pelykh: Did that a bit differently - calling one onchange from other isn't good style :P

pedrobaeza: That's not true. It's the most verbose way of doing it. You repeat an `if`, but not very important. Doing your way, I even prefer to save the extra method and simply add the other line.

alexey-pelykh: And potentially trigger unwanted behaviour is that method is ever overriden - why to make extra links between code where they are not justified?

pedrobaeza: Inheritance is for being respected, so if you override in other module that onchange (which is specific for this task), then you surely want the same to be executed in this case.

alexey-pelykh: Adjusted the code

sale_timesheet_line_exclude/README.rst
```diff
@@ -0,0 +1,94 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+This feature proves itself useful for *By Task* billing approach, when
```
pedrobaeza: This should go in description

alexey-pelykh: Done

pedrobaeza (approved):

alexey-pelykh:
Superseded by https://github.com/OCA/timesheet/pull/244

PR #837, opened 2019-05-03
ADD module sale_recompute_price

bealdav:


pedrobaeza:
AFAIK, this already exists since long and it's being migrated in #773

bealdav:
ok, thanks

PR #838, opened 2019-05-05
[FIX]sale_quotation_number: Sale Order number not genersted if tick L…

hugho-ad:
…ock Confirmed Orders


rousseldenis (changes requested):
@hugho-ad Is this still applicable ? If yes, could you add some tests that highlight your changes ?

PR #822, opened 2019-04-08
[10.0][MIG] sale_stock_picking_note

LoisRForgeFlow:
Backported from v11.

@Eficent

cc @aheficent @bjeficent

PR #843, opened 2019-05-10
[10.0][ADD] Add module sale_delivery_note

acsonefho:
Add a `delivery_note` field on the `sale.order`.

This MR is red due to an issue into account_invoice_triple_discount. Fixed into this MR: https://github.com/OCA/account-invoicing/pull/525


sale_delivery_note/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from . import sale_order
+from . import stock_move
```
rousseldenis: @acsonefho This makes tests failing

rousseldenis:
@acsonefho Isn't this a duplicate of https://github.com/OCA/sale-workflow/pull/822 ?

acsonefho:
> @acsonefho Isn't this a duplicate of #822 ?

@rousseldenis It seems a duplicate but in our case, the delivery_note is a new field (into sale and into picking (from another module)).
And I think picking_note is more like internal note (between employees) and delivery_note is more like an instruction from the end-customer.

lmignon (approved):
LGTM :+1:

lmignon:
PR rebased to see if the issue with sale_triple_discount is solved...

rafaelbn:
Is this module like sale_stock_picking_note ?

https://github.com/OCA/sale-workflow/tree/11.0/sale_stock_picking_note

rafaelbn:
Hi @acsonefho 

>  And I think picking_note is more like internal note (between employees) and delivery_note is more like an instruction from the end-customer.

https://github.com/OCA/sale-workflow/tree/11.0/sale_stock_picking_note is already in use to set notes in sale order and this notes are printed in the picking for end-customers

It looks like the same. Up to you not blocking

rafaelbn (changes requested):
I have done more deep review (my best intention).

You can see the video **sale_stock_picking_note VS sale_delivery_note**

https://drive.google.com/file/d/1uughmWYqZD4GVAz8L0bgqRYoO6N45RrA/view?usp=sharing

The OCA goal here is: we have a module from v9 in production which is like the same of this one, but in the case this both modules are different objetive we should care of making both "compatible" at UX experience:

- [ ] Both fields together with sense
- [ ] Very well explained the differences so the user understand them (improve README and USAGE)
- [ ] Each field is printed correctly in the PDF reports

We could ( I have to check with @sergio-teruel and @carlosdauden in this moment in holidays) agree that:

- sale_stock_picking_note is for notes for making the picking, not for the customer, so this notes are printed in "Picking Operations" and not in "Delivery Slip"

- sale_delivery_note: is for notes for the "Delivery man" (distributor)


In my test (check video) in runbot:

- [ ] We talk in this module about delivery so Is it related to : https://github.com/OCA/delivery-carrier ?

|                                                                                                                                                                    |         sale_stock_picking_note        |                   sale_delivery_note                   |
|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|:--------------------------------------:|:------------------------------------------------------:|
| Usage section                                                                                                                                                      |                   YES                  |                           NO                           |
| position field in sale order                                                                                                                                       |  TAB "other information" (see image_1) | botton of sale order  (see image_2) - [ ] Fix position |
| printed in sale order                                                                                                                                              |                   NO                   |                           NO                           |
| position field in stock picking                                                                                                                                    |             doesn't appears            |        TAB "additional info" Notes (see image_3)       |
| printed in PDF "Picking operations"                                                                                                                                |                   YES                  |                           NO                           |
| printed in PDF "Delivery Slip"                                                                                                                                     |                   NO                   |                           NO                           |
| Can define the note in res.parner the note? (a) Take the info from partner_shipping_id in sale.order (b) The note goes from partner to sale order to stock.picking |                   NO                   |                           NO                           |

Image 1:


Image 2:


Image 3:


My best

rafaelbn:
IMHO the best is one single module, both are the same. The only difference is is the note is printed o not in the "Delivery slip", that's all.

rousseldenis:
@acsonefho @Cedric-Pigeon

PR #844, opened 2019-05-15
[12.0][ADD] sale_milestone_profile_invoicing

TDu:
This module compares for milestone products the budget spent (timesheeted by employees)
with the amount invoiced to the customer.


sale_milestone_profile_invoicing/views/sale_milestone_profile_invoicing.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+
+<odoo>
+
+  <data>
```
kittiu: This data tag is not required.
```suggestion
```

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,129 @@
...
+        string='Invoiced from task cc',
+        compute='_compute_amount_invoiced_from_task',
+    )
+
+    @api.multi
+    def _get_timesheet_for_amount_calculation(self, only_invoiced=False):
+        """Return all timesheet line related to the sale order line."""
+        self.ensure_one()
+        if not self.task_id:
+            self.amount_delivered_from_task = 0
```
kittiu, line 54: It is more pythonic to assign number as following (as amount_delivered_from_task is a float)
```suggestion
            self.amount_delivered_from_task = 0.0
```

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,129 @@
...
+            return []
+        tasks_linked_to_line = self.env['project.task'].search(
+            [('sale_line_id', '=', self.id)]
+        )
+        tasks = self.env['project.task'].search(
+            [('id', 'child_of', tasks_linked_to_line.ids)]
+        )
+        ts = tasks.mapped('timesheet_ids').filtered(lambda r: r.employee_id)
```
kittiu: This can be shorten a bit.
```suggestion
        ts = tasks.mapped('timesheet_ids').filtered('employee_id')
```

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,129 @@
...
+        if only_invoiced:
+            ts = ts.filtered(lambda r: r.timesheet_invoice_id)
```
kittiu:
```suggestion
            ts = ts.filtered('timesheet_invoice_id')
```

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @api.multi
+    @api.depends(
+        'task_id',
+        'task_id.timesheet_ids.timesheet_invoice_id',
+        'task_id.timesheet_ids.unit_amount',
+    )
+    def _compute_amount_delivered_from_task(self):
+        for line in self:
+            total = 0
```
kittiu, line 78:
```suggestion
            total = 0.0
```

sale_milestone_profile_invoicing/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
+This module compares budget sold for milestones product with the time spent by various employee profiles.
```
kittiu: Non-blocking. But for person who don't have much background about timesheet, I don't quite get.
If you could explain a bit more on why and how for this module, would be perfect.

kittiu:
@TDu thank you for contribution. Most of the coding part looks ok to me. Just some final check.

- README file?
- Possible to add demo data about the milestone products? This will help anyone to try your module with ease.

kittiu:
@TDu just adding some thought. For me, I think the order line columns can get crowded very easily. And while the 2 new fields only make sense for some products, will it be better if we can conditionally display it on such case.
I..e, add new compute field that will be True only on such products, and display column conditionally using attrs column_invisible tag.
Non-blocking, just my thought.

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,129 @@
...
+    def _compute_invoice_status(self):
+        """Change invoice status for milestone line"""
+        super()._compute_invoice_status()
+        for line in self:
+            if line._is_linked_to_milestone_product():
+                mapping = line.project_id.sale_line_employee_ids.filtered(
+                    lambda r: r.sale_line_id == line.id
+                )
+                if not mapping:
+                    if line.qty_delivered <= line.product_uom_qty:
```
nikul-serpentcs, line 149: @TDu  No need to `Else` part
```
line.invoice_status = 'upselling'
if line.qty_delivered <= line.product_uom_qty:
    line.invoice_status = 'no'
```

sale_milestone_profile_invoicing/views/sale_milestone_profile_invoicing.xml
TDu: Indeed thanks removed.

sale_milestone_profile_invoicing/models/sale_order_line.py
TDu: Thanks changed

sale_milestone_profile_invoicing/models/sale_order_line.py
TDu: Thanks

sale_milestone_profile_invoicing/models/sale_order_line.py
TDu: Is it more pythonic ?
I would think the first version is simpler

sale_milestone_profile_invoicing/models/sale_order_line.py
TDu: I tried but the test cases fail if I do so

TDu:
@kittiu Thank you for your review. There is no README file as such because it should be auto generated with the content of the readme directory (it's a new thing for version 12.0 I think)
But you are correct that the description of what the module does is not that great, will try to improve it a little.

kittiu:
@TDu for README file, I always though we have to run the command locally to create it. This is new knowledge to me too, thank you!

sebalix (approved):

sale_milestone_profile_invoicing/tests/test_sale_milestone_profile_invoicing.py
```diff
@@ -0,0 +1,483 @@
...
+                'name': 'Test account',
+                'code': 'TEST',
+                'user_type_id': cls.account_type.id,
+                'reconcile': True,
+            }
+        )
+
+    def test_step_1(self):
+        """
+        """
```
i-vyshnevska, line 172: comment expected here

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,146 @@
...
+            return []
+        tasks_linked_to_line = self.env['project.task'].search(
+            [('sale_line_id', '=', self.id)]
+        )
+        tasks = self.env['project.task'].search(
+            [('id', 'child_of', tasks_linked_to_line.ids)]
+        )
+        ts = tasks.mapped('timesheet_ids').filtered('employee_id')
+        if only_invoiced:
+            ts = ts.filtered('timesheet_invoice_id')
```
i-vyshnevska: do we need filter them on state like 'canceled'

sale_milestone_profile_invoicing/tests/test_sale_milestone_profile_invoicing.py
TDu: Yes, I added some comments on each test methods.

sale_milestone_profile_invoicing/models/sale_order_line.py
TDu: I think we do, good catch, I updated that filter.

i-vyshnevska (approved):

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -0,0 +1,149 @@
...
+        compute='_compute_amount_delivered_from_task',
+        currency_field='company_currency_id',
+    )
+    amount_invoiced_from_task_company_currency = fields.Monetary(
+        string='Invoiced from task (company currency)',
+        compute='_compute_amount_invoiced_from_task',
+        currency_field='company_currency_id',
+    )
+    company_currency_id = fields.Many2one(
+        related='company_id.currency_id', store=True
```
gurneyalex: make this readonly, and change the label to avoid a warning on runbot
```suggestion
        related='company_id.currency_id', store=True, readonly=True, string="Company currency"
```

TDu: Done. A rebased, squash the all thing.

gurneyalex (approved):

rafaelbn:
@alexey-pelykh did you saw this PR? I was last 3 weeks overwhelmed

PR #852, opened 2019-05-27
[FIX] product_supplierinfo_for_customer_sale: TypeError: unorderable types: int() >= NoneType()

hugho-ad:
image

### DIFF
--- a/product_supplierinfo_for_customer_sale/models/product_product.py
+++ b/product_supplierinfo_for_customer_sale/models/product_product.py
@@ -13,7 +13,9 @@ class ProductProduct(models.Model):
     def name_search(self, name='', args=None, operator='ilike', limit=80):
         res = super(ProductProduct, self).name_search(
             name, args=args, operator=operator, limit=limit)
-        if (not name or not self._context.get('partner_id') or
+        if not limit or len(res) >= limit:
+            limit = (limit - len(res)) if limit else False
+        if (not name and limit or not self._context.get('partner_id') or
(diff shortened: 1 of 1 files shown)


product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,6 +13,8 @@ class ProductProduct(models.Model):
     def name_search(self, name='', args=None, operator='ilike', limit=80):
         res = super(ProductProduct, self).name_search(
             name, args=args, operator=operator, limit=limit)
+        if not limit or len(res) >= limit:
+            limit = (limit - len(res)) if limit else False
```
pedrobaeza, line 17:
```suggestion
            limit = (limit - len(res)) if limit else sys.maxsize
```
importing `sys` previously

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,6 +13,8 @@ class ProductProduct(models.Model):
...
         if (not name or not self._context.get('partner_id') or
```
pedrobaeza: or alternatively
```suggestion
        if (not name and limit or not self._context.get('partner_id') or
```

hugho-ad: Thanks!, Done!

ValentinVinagre (approved):
LGTM

pedrobaeza (approved):

PR #859, opened 2019-05-29
12.0 mig sales team security

Montxu:
cc @guadaltech


anajuaristi (approved):
LGTM :+1:

oihane (approved):
LGTM

sales_team_security/models/res_partner.py
```diff
@@ -11,7 +11,8 @@ class ResPartner(models.Model):
     @api.model
     def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
                         submenu=False):
-        """Patch view to inject the default value for the team_id and user_id.
+        """
+        Patch view to inject the default value for the team_id and user_id.
```
pedrobaeza, line 15: Why changing this? docstring can start with the triple quote and the first doc line in the same line

sales_team_security/models/res_partner.py
```diff
@@ -11,7 +11,8 @@ class ResPartner(models.Model):
...
         """
         # FIXME: Use base_view_inheritance_extension when available
```
pedrobaeza, line 17: Please consider this for simplifying a lot the code.

pedrobaeza (dismissed):
For avoiding to be merged until questions are answered.

sales_team_security/models/res_partner.py
Montxu: hi, this module (base_view_inheritance_extension) not in version 12.0, as soon this module is migrated i fix this.

pedrobaeza: We should push for having it migrated as if we merge this one without the dependency, then we won't be able to change the dependency in this branch due to stability policy.

pedrobaeza: As the other module is not merged and this solution only implies a bit more of code, I'm dismissing it and we go on.

PR #861, opened 2019-05-29
12.0 mig sale disable inventory check

Montxu:
cc @guadaltech


rousseldenis (approved):
Code review.

I know this is a migration, but usually it's better to have an option to enable this as it breaks the standard flow (it's wanted I know) and could break other tests. Maybe put in ROADMAP.

sale_disable_inventory_check/tests/test_disable_inventory_check.py
```diff
@@ -0,0 +1,34 @@
...
+class TestDisableInventoryCheck(common.TransactionCase):
+
+    def setUp(self):
+        super(TestDisableInventoryCheck, self).setUp()
+
+        # Create stockable product
+        self.product_1 = self.env['product.product'].create({
+            'name': 'Product 1',
+            'type': 'product',
+            'uom_id': self.env.ref('product.product_uom_unit').id,
```
nikul-serpentcs: @Montxu `self.env.ref('product.product_uom_unit')` --> `self.ref('uom.product_uom_unit')`

nikul-serpentcs (changes requested):
Improved Code

HaraldPanten (approved):
Functional review 👍

pedrobaeza:
> I know this is a migration, but usually it's better to have an option to enable this as it breaks the standard flow (it's wanted I know) and could break other tests. Maybe put in ROADMAP.

The way to do it is via context and config check:
```
from odoo.tools import config

    @api.onchange('product_uom_qty', 'product_uom', 'route_id')
    def _onchange_product_id_check_availability(self):
        if (config['test_enable'] and 
                not self.env.context.get('test_sale_disable_inventory_check')):
            return super()._onchange_product_id_check_availability()
        return {}
```

and put that context on current test.

Montxu:
Fixed, @nikul-serpentcs. Are the changes OK?

nikul-serpentcs (approved):
only Code review LGTM :+1:

rousseldenis:
> The way to do it is via context and config check:

I don't like this way. You cannot put code that depends on tests...

pedrobaeza:
Denis, that's done several times in Odoo and in OCA. It's not ideal but it allows you a good way to control this kind of incompatibilities

rousseldenis:
> that's done several times in Odoo and in OCA

Why not another way of thinking ? :smiley: 

> It's not ideal but it allows you a good way to control this kind of incompatibilities

That's why an option is more convenient.

I'm not blocking on this but remind it's not a good way of coding.

pedrobaeza:
Your alternative seems to me more convoluted and requires more configuration

legalsylvain (approved):
thanks for porting this module ! 
LGTM. code review / not test.

PR #868, opened 2019-06-07
add module sale_order_line_date

open-net-sarl:
Hello,

Instead of doing sale_order_line_date from scratch, we tought of helping you with migration, given that this module does what we wanted.

If the validation is not passed we won't do more. You can retake what we did if you want.
NB: i did not checked the tests. Python and XML are okay, and apparently the traductions too.

Have a nice day.

PR #703, opened 2018-10-01
12.0 mig sale order line date

mpanarin:
Migration of the sale_order_line_date as well as adding the setup directory

PR #875, opened 2019-06-13
12.0 mig sale order line date

open-net-sarl:
I took time to make the things as asked.
Hope this time it is okay.

Nb: I first wanted to add changes, that's why i did it again "properly". Finally, i did not do the changes wanted: it is still the module from 11.0 corrected, nothing was added or removed. The tests are not mine, i just took it from 11.0

Have a nice day


open-net-sarl:
last pull request:
https://github.com/OCA/sale-workflow/pull/868

sale_order_line_date/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# © 2016 OdooMRP team
+# © 2016 AvanzOSC
+# © 2016 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# © 2016 Eficent Business and IT Consulting Services, S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale Order Line Date",
+    "version": "12.0.0.0.0",
```
rousseldenis:
```suggestion
    "version": "12.0.1.0.0",
```

sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    commitment_date = fields.Datetime()
```
rousseldenis: Shouldn't have migration script as you changed the field ?

sale_order_line_date/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    @api.onchange('commitment_date')
+    def _onchange_commitment_date(self):
+        """Warn if the commitment dates is sooner than the commitment date"""
+        result = super(SaleOrder, self)._onchange_commitment_date()
```
rousseldenis, line 18: All super could be simplified (not blocking) :

```suggestion
        result = super()._onchange_commitment_date()
```

rousseldenis (changes requested):
LGTM. Code review. Some comments

sale_order_line_date/models/sale_order_line.py
open-net-sarl: As we never used this module, we just migrated what we needed.
Maybe add the attributes "old_name='requested_date'" would be enough, but we won't do the retro-compatibility fix. Sorry

sale_order_line_date/models/sale_order.py
open-net-sarl: I don't think it is a good way to do that.
I personally prefer being explicit, considering the result is the same.
Do as you want

sale_order_line_date/models/sale_order_line.py
rousseldenis: Yes, at least the old_name should be sufficent. Could you add it ?

open-net-sarl:
I did a new commit adding the commitment_date after the description in the sale.order document.
That's the change i initially wanted to do.

sale_order_line_date/README.rst
```diff
@@ -0,0 +1,64 @@
...
+------------
+* Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>
+* Esther Martín <esthermartin@avanzosc.es>
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
+* Ana Juaristi <anajuaristi@avanzosc.es>
+* Jordi Ballester <jordi.ballester@eficent.com>
+* Aaron Henriquez <ahenriquez@eficent.com>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
+* Francesco Apruzzese <f.apruzzese@apuliasoftware.it>
+
```
rousseldenis, line 51: You could add yourself as contributor of course !

sale_order_line_date/models/sale_order_line.py
open-net-sarl: It is done.
As told before, we tried to help the OCA, especially for the switzerland modules.
The process change <--> review does not apply only once but can last very long.
For that we are an enterprise, we can't spend a lot of time with reviews.

rousseldenis (approved):
Code review.

sale_order_line_date/models/sale_order_line.py
rousseldenis: Thank you.

> The process change <--> review does not apply only once but can last very long.
> For that we are an enterprise, we can't spend a lot of time with reviews.

That's a point of view/ Most of people in OCA are tied to enterprise too. IMHO code review should be part of development as you can confont to other people, get ideas, stay in touch with changes, ... many advantages. But that depends on enterprise philosophy, that's true.

rousseldenis:
Work as expected on runbot.

IMHO you should add a group on 'commitment_date' in views as for the field on 'sale.order'. Otherwise, field appear in lines but not on order level.

Group : 'sale.group_sale_order_dates'.

AaronHForgeFlow (changes requested):
The commitment_date is being not propagated to the stock moves.

rousseldenis:
> The commitment_date is being not propagated to the stock moves.

Oops, just tested the onchange...

open-net-sarl:
> > The commitment_date is being not propagated to the stock moves.
> 
> Oops, just tested the onchange...

_prepare_procurement_values methods should work exactly as in V11.

open-net-sarl:
I just did some changes in the code.
In the write method of sale.order.line, if one line add no commitment_date, all lines would have taken its sale.order's commitment_date.
In onchange, using comprehension list instead of a loop + append

I will try to help as much as i can in my free time, i just hope that it won't become an infinite review loop..
I just fixed a bug, but the rest works like in V11

AaronHForgeFlow:
@open-net-sarl Thank you for your improvements. There are still a couple of things missing:

* Schedule date is still not passed to the stock moves when confirming the sales order

* Test are calling to unexisting method:
`AttributeError: 'sale.order' object has no attribute 'onchange_commitment_date`

Regards.

AaronHForgeFlow:
sorry @open-net-sarl there was an open PR for this module and it is much older than this one. SO better keep the other one: https://github.com/OCA/sale-workflow/pull/703

rafaelbn:
Dear @open-net-sarl , we are very happy to have a new contributor. And we all thank you for this contribution.

But It was a PR before yours, so the correct way to work is continuing the PR before: https://github.com/OCA/sale-workflow/pull/703

As you can see here you can make PR to the original author https://github.com/OCA/sale-workflow/pull/703#issuecomment-470964019

I just want you to understand that you could be in the same situation of @mpanarin , that nobody review your PR and after some time a new contributor like you, make the same PR despite of checking before if it was done.

You question could be: and how could I know that? Well:

- The easy way is the migration issue: https://github.com/OCA/sale-workflow/issues/700
- The other way is searching before in the existing PR

cc @alexey-pelykh FYI

rafaelbn:
IMHO the interested people in having this module in v12 should contribute in https://github.com/OCA/sale-workflow/pull/703 Thanks and feel free to comment.

The most delicate thing here ( cc @jgrandguillaume ) is that I don't want to discourage @open-net-sarl but also I don't want to discourage @mpanarin .

rafaelbn:
I'm going to test this one as https://github.com/OCA/sale-workflow/pull/703#pullrequestreview-291517422 is not working and author @mpanarin  is not going to work on it anymore.

rafaelbn (changes requested):
Tested in runbot functionally: doesn't work

I used commitment date 1st October and this date doesn't propagate any where

alexeirivera87:
I know why the SO line's commitment dates are not being propagated to stock.moves.
Must I get some permission to contribute?

Regards

CasVissers-360ERP:
@alexeirivera87 
> I know why the SO line's commitment dates are not being propagated to stock.moves.
> Must I get some permission to contribute?
> 
> Regards

What fix do you want to propose?

sale_order_line_date/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+        if not result:
+            result = {}
+        if not self:
+            return result
+        self.ensure_one()
+        if 'warning' not in result:
+            lines = []
+            for line in self.order_line:
+                lines.append((1, line.id, {'commitment_date':
```
RLeeOSI:
```suggestion
                if not line.commitment_date:
                    lines.append((1, line.id, {'commitment_date':
```
Is there a reason why the order commitment date should overwrite an existing line commitment date? 

If not, there is a solution to the issue with the date not propagating to picking. The reason the picking scheduled date doesn't inherit properly is because it's looking for an order commitment date. If you add an onchange to the line commitment date that overwrites order commitment date if it doesn't exist, then the picking will have a date to inherit.

sale_order_line_date/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    @api.onchange('commitment_date')
+    def _onchange_commitment_date(self):
+        """Warn if the commitment dates is sooner than the commitment date"""
```
RLeeOSI, line 17:
```suggestion
        """Warn if the commitment date is sooner than the expected date"""
```
Comment from base sale module

RLeeOSI (commented):
The picking is not inheriting a scheduled date because it expects there to be a commitment date on the sale order. A possible solution would be to add an onchange to the line commitment date that sets the order commitment date if it doesn't exist. That way, there will be a date for the picking to inherit.

I would also suggest a warning message if the commitment date of a line is later than the commitment date of the order. In the case that the user wants one product to be promised later than the order itself, it will be possible, but since this is abnormal, there should be a message. In a normal situation, the order commitment date should be the same as the latest line commitment date, and the picking should be scheduled for the latest date.

The moves are not inheriting the commitment date because the module does not depend on sale_stock. Right now, sale_stock is calling super on sale.order.line._prepare_procurement_values and overwriting the changes from this module. Add the dependency and the moves will reflect the line commitment date.

NuriaXifre:
@open-net-sarl I have done a functional review and it does not work. The commitment date does not propagate. Do you have in mind to continue working on this module?

MiquelRForgeFlow:
Superseded by https://github.com/OCA/sale-workflow/pull/984.

PR #876, opened 2019-06-13
[12.0][MIG] sale_validity

MiquelRForgeFlow:
Standard migration.


HaraldPanten (changes requested):
Runbot is crashing it seems to have some problems. I tried to install the module but and error occurs.

sale_validity/views/res_config_settings.xml
```diff
@@ -0,0 +1,34 @@
...
+                    <div class="o_setting_left_pane"/>
+                    <div class="o_setting_right_pane">
+                        <label for="default_sale_order_validity_days"/>
+                        <div class="text-muted mt4 mb4">
+                            By default, the validity date of sale orders will be the date of the sale order plus the number of days defined in this field.
+            If the value of this field is 0, the sale orders will not have a validity date by default.
+                        </div>
+                        <div class="text-muted mt4 mb4">
+                            <field name="default_sale_order_validity_days"/>
+                            <label string="Days" style="margin-left: 5px;"/>
```
MiquelRForgeFlow: Invalid XML:  ``label`` nodes must have a ``@for``

MiquelRForgeFlow (commented):
anyone knows how to solve this? what do I put in the `for` of this label?

sale_validity/views/res_config_settings.xml
AdriaGForgeFlow: I believe you can put whatever, even leave it empty:
`<label for="" string="Days" style="margin-left: 5px;"/>`

Worked for me.

sale_validity/models/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
...
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    default_sale_order_validity_days = fields.Integer(
```
AdriaGForgeFlow: Using default at the beginning on the name causes: 
image

It also needs the parameter **readonly=False** to be able to be modified

```
sale_order_validity_days = fields.Integer(
        related='company_id.default_sale_order_validity_days',
        readonly=False,
    )
```

AdriaGForgeFlow (changes requested):
Some comments

HaraldPanten (approved):
Functional review 👍  seems to be OK

MiquelRForgeFlow:
@rafaelbn it's green :green_apple:

rafaelbn (approved):

PR #880, opened 2019-06-20
[MIG] portal_sale_personal_data_only: Migration to 12.0

chienandalu:
By default, portal users are allowed to see all the orders and invoices in
which a member of their organization are followers. That could cause a leaking
of  documents between members and departments and of the organization that
should stay private.

This module restricts that behaviour so the portal users only see their own
documents.

cc @Tecnativa


HaraldPanten (approved):
Functional review on runbot 👍

tbaden (approved):
thanks for the work 👍 
I did a code review: LGTM

PR #884, opened 2019-06-25
[12.0][FIX][sale_stock_picking_blocking] the _commercial_fields method shoud return something

Tonow-c2c:
the _commercial_fields method should return something
and 
```python
>>> mylist = ['a', 'b']
>>> bar = mylist.append('foo')
>>> bar == None
True
```


rousseldenis (approved):
Of Course !

PR #890, opened 2019-07-04
[12.0][ADD] sale_timesheet_limit_date

i-vyshnevska:
This addon allows to constrain sale order and invoices with date, so order will gather analytic lines  only till some date, delivered qty will be computed  accordingly


hr_timesheet_limit_sale_date/models/account_invoice_line.py
```diff
@@ -0,0 +1,39 @@
...
+                    old_timesheets = self.env[
+                        'account.analytic.line'].search(domain)
+                    domain = expression.AND([domain, [
+                        ('date', '<=',
+                            invoice_line.invoice_id.timesheet_limit_date)
+                    ]])
+                    constrained_timesheets = self.env[
+                        'account.analytic.line'].search(domain)
+                    to_clean = old_timesheets - constrained_timesheets
+                    to_clean.write({'timesheet_invoice_id': ''})
```
sebalix:
```suggestion
                    to_clean.write({'timesheet_invoice_id': False})
```

hr_timesheet_limit_sale_date/models/sale_order.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+    _name = 'sale.order'
```
sebalix: no need for _name

hr_timesheet_limit_sale_date/models/sale_order_line.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, models
+from odoo.osv import expression
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+    _name = 'sale.order.line'
```
sebalix: idem

hr_timesheet_limit_sale_date/views/invoice_order_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
```
sebalix:
```suggestion
    <record id="invoice_form" model="ir.ui.view">
```

sebalix (commented):
Travis is failing on `sale_milestone_profile_invoicing`, don't know if it's related to this module.
Code review only

hr_timesheet_limit_sale_date/i18n/fr.po
```diff
@@ -0,0 +1,52 @@
...
+#. module: hr_timesheet_limit_sale_date
+#: model:ir.model.fields,field_description:hr_timesheet_limit_sale_date.field_sale_order_timesheet_limit_date
+msgid "Timesheet Limit Date"
+msgstr "Date limite Timesheet"
+
+#. module: hr_timesheet_limit_sale_date
+#: code:addons/hr_timesheet_limit_sale_date/models/sale_order_line.py:21
+#, python-format
+msgid "You cannot apply the TS limit date on multiple orders."
+msgstr "Vous ne pouvez pas appliquer la date limite des timesheet sur plusieurs bons de commande."
```
TDu:
```suggestion
msgstr "Vous ne pouvez pas appliquer la date limite des timesheets sur plusieurs bons de commande."
```

hr_timesheet_limit_sale_date/models/account_invoice_line.py
```diff
@@ -0,0 +1,39 @@
...
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    @api.model
+    def create(self, values):
+        # aal linked to timesheet on model level
+        # we need to personalize it for each invoice even it belong to
+        # same SO, to rich it we iterate on records and correct aal
+        # base goal here is to recompute timesheet_ids to drop ones
+        # under limit date
```
grindtildeath: Can you please rephrase? I did read it 3 or 4 times but still cannot understand what it means...

hr_timesheet_limit_sale_date/models/sale_order.py
```diff
@@ -0,0 +1,48 @@
...
+        string='Timesheet Limit Date',
+        help='Invoice will be created with timesheet prior and including'
+        ' the date set',
+    )
+
+    @api.multi
+    @api.depends('timesheet_limit_date')
+    def _compute_timesheet_ids(self):
+        # this method copy of base method, it injects date in domain
+        if not self.timesheet_limit_date:
```
TDu: Could this fail, when called with multiple sale order some with a timesheet_limit_date some without ?

hr_timesheet_limit_sale_date/models/account_invoice_line.py
grindtildeath: If the write is always the same, please collect all the `to_clean` records in a specific recordset, and call the write after the loop. Because with the actual version, you going to send as many UPDATE commands as you have a.a.l to update, whereas you could send only one, calling write a single time on all the records that need it.

hr_timesheet_limit_sale_date/models/sale_order.py
grindtildeath: This will raise an error when the method is called on multiple records, you must move this check inside the loop.

hr_timesheet_limit_sale_date/models/sale_order_line.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def _timesheet_compute_delivered_quantity_domain(self):
+        domain = super()._timesheet_compute_delivered_quantity_domain()
+        order = self.mapped('order_id')
+        if len(order) == 1:
```
grindtildeath: This seems broken to me. 
Consider we have two different sale orders, and we update the timesheet_limit_date on both orders, then `_compute_qty_delivered` is triggered and here, the extra domain won't be appended here, and we'd have a different result by updating two sale orders at the same time than updating the same sale orders one by one...

grindtildeath (changes requested):
Firsts and foremost, please rename this module sale_timesheet_limit_date.
Then please consider the case of updating two orders at the same time that is described and add a test.

hr_timesheet_limit_sale_date/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Simone Orsi <simone.orsi@camptocamp.com>
```
TDu: Is it a new module ? If so did they all contribute ?

hr_timesheet_limit_sale_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
+This module adds a date field to sale order and invoice. This date is limitation
```
TDu:
```suggestion
This module adds a date field to sale order and invoice. This date limits the
```

hr_timesheet_limit_sale_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+selection of timesheet lines for including them in sale order and invoice.
+When create an invoice limit date will be inherited from sale order, then date
```
TDu:
```suggestion
When creating an invoice the limit date will be inherited from the sale order, then date
```

sale_milestone_profile_invoicing/models/sale_order_line.py
```diff
@@ -149,7 +149,7 @@ def _compute_invoice_status(self):
         for line in self:
             if line._is_linked_to_milestone_product():
                 mapping = line.project_id.sale_line_employee_ids.filtered(
-                    lambda r: r.sale_line_id == line.id
```
TDu: This should have been in a separated PR, I think

hr_timesheet_limit_sale_date/models/sale_order_line.py
i-vyshnevska: it's not broken, if domain triggered on many lines which belongs to different SO we skip injecting of date
it's tested in test_multiple_line_pass
this is the hard part which I tried to explain, domain is class method and  we trying to works on object

hr_timesheet_limit_sale_date/models/sale_order.py
i-vyshnevska: updated

hr_timesheet_limit_sale_date/models/sale_order_line.py
grindtildeath: The test `test_multiple_line_pass` doesn't bring anything actually as there are no assertions in it.
Please add timesheets on both sale orders, with entries after and before the limit date that you're going to set afterwards, and you'll see that results are different if you write on both sale order or only on one....

sale_milestone_profile_invoicing/models/sale_order_line.py
i-vyshnevska: I will keep them here to make travis green, for time economy, later it can be moved to separate PR

hr_timesheet_limit_sale_date/readme/CONTRIBUTORS.rst
i-vyshnevska: they contribute in previous dev

hr_timesheet_limit_sale_date/models/sale_order_line.py
TDu: I tend to agree with @grindtildeath this is broken if the sale.order.lines passed to this method belongs to different sale.order

i-vyshnevska: if different lines passed here it's not executed, but I am looking for better way

hr_timesheet_limit_sale_date/readme/CONTRIBUTORS.rst
TDu: ok

pedrobaeza:
Shouldn't this be in OCA/timesheet?

hr_timesheet_limit_sale_date/models/account_invoice_line.py
i-vyshnevska: updated

sale_milestone_profile_invoicing/models/sale_order_line.py
i-vyshnevska: moved to https://github.com/OCA/sale-workflow/pull/891

hr_timesheet_limit_sale_date/models/sale_order_line.py
i-vyshnevska: updated @grindtildeath @TDu

sale_timesheet_limit_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+    @api.depends('order_id.timesheet_limit_date')
+    @api.multi
+    def _compute_qty_delivered(self):
+        # call recomputation on group of lines to be able apply limit date
+        # in domain
+        orders = self.mapped('order_id')
+        lines_groups = [
+            self.filtered(lambda sol: sol.order_id.id == o.id) for o in orders
+        ]
+        lines_without_order = self.filtered(lambda sol: sol.order_id == False)
```
grindtildeath: `order_id` is required on `sale.order.line`, so this case should never happen, and there's no reason to keep it here.

sale_timesheet_limit_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+        elif len(order) > 1:
+            raise UserError(
+                'You cannot apply the TS limit date on multiple orders.'
+            )
+        return domain
+
+    @api.depends('order_id.timesheet_limit_date')
+    @api.multi
+    def _compute_qty_delivered(self):
+        # call recomputation on group of lines to be able apply limit date
```
grindtildeath:
```suggestion
        """Group lines by sale order to allow application of 'per order' domain above"""
```

sale_timesheet_limit_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+    def _timesheet_compute_delivered_quantity_domain(self):
+        domain = super()._timesheet_compute_delivered_quantity_domain()
+        order = self.mapped('order_id')
+        if len(order) == 1:
+            if order.timesheet_limit_date:
+                domain = expression.AND([
+                    domain,
+                    [('date', '<=', self.order_id.timesheet_limit_date)]]
+                )
+        elif len(order) > 1:
```
grindtildeath: This case should never happen, please remove.

sale_timesheet_limit_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo.osv import expression
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    def _timesheet_compute_delivered_quantity_domain(self):
+        domain = super()._timesheet_compute_delivered_quantity_domain()
+        order = self.mapped('order_id')
+        if len(order) == 1:
```
grindtildeath: This check is unneeded.

sale_timesheet_limit_date/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+            if order.timesheet_limit_date:
+                domain = expression.AND([
+                    domain,
+                    [('date', '<=', self.order_id.timesheet_limit_date)]]
```
grindtildeath: As `self` can include multiple lines, this would raise a Singleton exception error.
```suggestion
                    [('date', '<=', order.timesheet_limit_date)]]
```

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+This module adds a date field to sale order and invoice. This date limits the
+selection of timesheet lines for including them in sale order and invoice.
+When creating an invoice the limit date will be inherited from the sale order, then date
+in sale order is cleaned, when invoice validated lines will be constraint by
+limitation date.
+Known issues if sale order has two invoices in draft state, and the validated
```
grindtildeath: Please move this to `ROADMAP.rst`

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+This module adds a date field to sale order and invoice. This date limits the
+selection of timesheet lines for including them in sale order and invoice.
```
grindtildeath:
```suggestion
selection of timesheet lines to include them in sale order and invoice.
```

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+When creating an invoice the limit date will be inherited from the sale order, then date
```
grindtildeath:
```suggestion
When an invoice is created from a sale order, the limit date will be propagated from the sale order to the invoice,  and the date
```

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+in sale order is cleaned, when invoice validated lines will be constraint by
```
grindtildeath:
```suggestion
in sale order is removed.
```

I don't understand what you mean with "when invoice validated lines will be constraint by limitation date". There is no constraint on the invoice models as far as I can see...

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+limitation date.
+Known issues if sale order has two invoices in draft state, and the validated 
+one with the newest date or without date that it pick all lines till this date, 
+and invoice with older date will have no any lines.
```
grindtildeath: Please rephrase this.

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+
+Compatibility with sale_milestone_profile_invoicing are not tested.
```
grindtildeath: Please move to `ROADMAP.rst` as well

sale_timesheet_limit_date/models/sale_order.py
```diff
@@ -0,0 +1,51 @@
...
+    _inherit = 'sale.order'
+
+    timesheet_limit_date = fields.Date(
+        string='Timesheet Limit Date',
+        help='Invoice will be created with timesheet prior and including'
+        ' the date set',
+    )
+
+    @api.multi
+    @api.depends('timesheet_limit_date')
```
grindtildeath, line 17: Shouldn't you add `analytic_account_id.line_ids` here ? :thinking:

sale_timesheet_limit_date/models/account_invoice_line.py
```diff
@@ -0,0 +1,36 @@
...
+                    )]
+                    old_timesheets = self.env[
+                        'account.analytic.line'].search(domain)
+                    domain = expression.AND([domain, [
+                        ('date', '<=',
+                            invoice_line.invoice_id.timesheet_limit_date)
+                    ]])
+                    constrained_timesheets = self.env[
+                        'account.analytic.line'].search(domain)
+                    to_clean += old_timesheets - constrained_timesheets
```
grindtildeath: Just wondering, but wouldn't this do the same ?
```
to_clean += self.env['account.analytic.line'].search([('timesheet_invoice_id', '=', invoice_line.invoice_id.id), '|', ('date', '>', invoice_line.invoice_id.timesheet_limit_date), ('date', '=', False)])
```
Then I feel like this could be improved heavily without looping on all the created lines and doing the searches on each iteration as long as you can collect all the related invoices. Maybe add a comment and let's not spend more time on it at the moment...

grindtildeath (commented):
I didn't review the tests yet, will do it after having lunch.

sale_timesheet_limit_date/tests/test_sale_lines.py
```diff
@@ -0,0 +1,80 @@
...
+            'name': 'New Acc.'
+        })
+        order_2.action_confirm()
+        orders._compute_timesheet_ids()
+        so_lines = self.sale_order.mapped('order_line')
+        so_lines += order_2.mapped('order_line')
+
+        with self.assertRaises(exceptions.UserError):
+            so_lines._timesheet_compute_delivered_quantity_domain()
+        so_lines._compute_qty_delivered()
```
grindtildeath: You shouldn't need to call this explicitly (as it must be triggered by `api.depends`)

sale_timesheet_limit_date/tests/test_sale_lines.py
```diff
@@ -0,0 +1,80 @@
...
+            unit_amount=1,
+            date='2019-05-10',
+            task_id=new_task.id,
+        )
+        new_lines.write({'so_line': order_2.order_line[0].id})
+        order_2.analytic_account_id = self.aaa_model.create({
+            'name': 'New Acc.'
+        })
+        order_2.action_confirm()
+        orders._compute_timesheet_ids()
```
grindtildeath: You shouldn't need to call this explicitly (as it must be triggered by api.depends)

sale_timesheet_limit_date/tests/test_sale_lines.py
```diff
@@ -0,0 +1,80 @@
...
+        # orders get quantity and lines independently
+        self.assertEqual(len(self.sale_order.timesheet_ids), 2)
+        self.assertEqual(len(order_2.timesheet_ids), 1)
+        self.assertEqual(self.so_line.qty_delivered, 2)
+        self.assertEqual(order_2.order_line.qty_delivered, 1)
+        # change of limit date triggers recomputation
+        orders.write({'timesheet_limit_date': '2019-05-10'})
+        self.assertEqual(self.so_line.qty_delivered, 3)
+        self.assertEqual(order_2.order_line.qty_delivered, 3)
+        self.assertEqual(len(order_2.timesheet_ids), 3)
```
grindtildeath: If both sale order have the same quantity of lines, you cannot ensure the right lines are computed. 
Suggestion: Do not create a line on `2019-05-09` for `order_2` and assert you have 3 on the last two lines here.

sale_timesheet_limit_date/models/sale_order.py
i-vyshnevska: I think this dependencies are enough https://github.com/odoo/odoo/blob/fb1839fa17a3197d781b45dbfe755277b4816904/addons/sale/models/sale.py#L1242

sale_timesheet_limit_date/readme/DESCRIPTION.rst
i-vyshnevska: This is not valid for v12 so I drop that at all

sale_timesheet_limit_date/models/sale_order.py
grindtildeath: I guess you meant those deps: https://github.com/odoo/odoo/blob/fb1839fa17a3197d781b45dbfe755277b4816904/addons/sale_timesheet/models/sale_order.py#L23
Anyway, the api must take care of it...

sale_timesheet_limit_date/models/account_invoice_line.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    @api.model
+    def create(self, values):
+        # unlink lines from invoice if they under limit_date
+        res = super().create(values)
+        for invoice_line in res:
+            to_clean = self.env['account.analytic.line']
```
grindtildeath: Please move this one line above.

sale_timesheet_limit_date/models/account_invoice_line.py
```diff
@@ -0,0 +1,30 @@
...
+                    lambda sol: sol.product_id.invoice_policy == 'delivery'
+                    and sol.product_id.service_type == 'timesheet')
+                if sale_line_delivery:
+                    to_clean += self.env['account.analytic.line'].search([
+                        ('timesheet_invoice_id', '=',
+                            invoice_line.invoice_id.id),
+                        ('date', '>',
+                            invoice_line.invoice_id.timesheet_limit_date),
+                    ])
+            to_clean.write({'timesheet_invoice_id': False})
```
grindtildeath: Please unindent this line of 4 spaces to call `write` outside the loop.

grindtildeath (commented):
We're almost there :grinning: 

Still, I agree with @pedrobaeza and think this PR should be moved to OCA/timesheet

i-vyshnevska:
closed in favor of https://github.com/OCA/timesheet/pull/266/files

PR #895, opened 2019-07-10
[12.0][MIG] migrate sale order line description

mohamedhagag:
migrate sale order line description to v 12.0


mohamedhagag:
@aheficent OK I got the reason for the error in your test:
You are trying to add a line without selecting a partner, I will fix it .

mohamedhagag:
@aheficent I fixed the issue in my code
But I found that the below module does not handle such issue:
sale_milestone_profile_invoicing
I uninstalled the mentioned module and tried the workflow and it's working fine.

AaronHForgeFlow:
Ok, so another fix has to be done to that module sale_milestone_profile_invoicing @TDu 

Thanks for the clarification @mohamedhagag

mohamedhagag:
@TDu @aheficent 

I just made a PR for sale_milestone_profile_invoicing #896 

waiting for the build to test

AaronHForgeFlow (approved):
Functional tests good to me.

sale_order_line_description/models/sale_order_line.py
```diff
@@ -16,6 +16,8 @@ def product_id_change(self):
         if (self.user_has_groups(
                 'sale_order_line_description.'
                 'group_use_product_description_per_so_line') and
-                self.product_id.description_sale):
-            self.name = self.product_id.description_sale
+                self.product_id.description_sale) and self.order_id.partner_id:
```
pedrobaeza: You shouldn't put the partner as condition here, but only for filling the lang context retrieving the field contents. If no partner, no context, but field must be retrieved.

pedrobaeza (changes requested):
And please squash your migration commits together

sale_order_line_description/models/sale_order_line.py
mohamedhagag: I added the partner to avoid the error mentioned by @aheficent - will remove it .

pedrobaeza: No, you have to do a middle solution.

mohamedhagag: sorry but I tried to squash my commits but I failed to do.
How to squash them ? sorry for that .

mohamedhagag:
I made the commit squash, I'm not sure if this is what you mean .

pedrobaeza:
Yes the squash is correct, but the code remains the same. I will suggest proper code.

sale_order_line_description/models/sale_order_line.py
```diff
@@ -17,5 +17,7 @@ def product_id_change(self):
                 'sale_order_line_description.'
                 'group_use_product_description_per_so_line') and
                 self.product_id.description_sale):
-            self.name = self.product_id.description_sale
+            self.name = self.product_id.with_context(
```
pedrobaeza: Proper code for this:
```python
product = self.product_id
if self.order_id.partner_id:
    product = product.with_context(
        lang=self.order_id.partner_id.lang,
    )
self.name = product.description_sale

mohamedhagag: done, thanks for guidance

pedrobaeza (approved):
Although the 2 commits must be squashed

PR #896, opened 2019-07-10
[12.0] sale_milestone_profile_invoicing Fix error if no partner

mohamedhagag:
Fix error if no partner or task_id set on sale order line .

### DIFF
--- a/sale_milestone_profile_invoicing/models/sale_order_line.py
+++ b/sale_milestone_profile_invoicing/models/sale_order_line.py
@@ -74,7 +74,7 @@ class SaleOrderLine(models.Model):
         'task_id.timesheet_ids.unit_amount',
     )
     def _compute_amount_delivered_from_task(self):
-        for line in self:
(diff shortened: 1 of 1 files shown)


mohamedhagag:
just tested it on runbot and working fine

TDu:
I am not sure why this is needed as the same test on an empty task_id is done here
https://github.com/OCA/sale-workflow/blob/d161b170fc794f45b3bdfaa0d3a8a02c44c243d7/sale_milestone_profile_invoicing/models/sale_order_line.py#L53

Maybe adding a unit test would help

flotho (commented):
Hi,

Thanks for the proposal
I'm not sure to understand the name of your PR because you don't filter on partner.
Task could be defined with no partners so filtering on task won't probably be enough to have a defensive for the "no partner" issue.
Am I right ?

rousseldenis:
@mohamedhagag What's the status of this ? Could you modify your commit and PR messages as there is no reference to partner in your changes?

PR #899, opened 2019-07-14
[12.0] Migrate sale_rental and sale_start_end_dates

alexis-via:


sale_rental/models/stock.py
```diff
@@ -69,142 +68,138 @@ def _get_rental_push_pull_rules(self):
...
-                    for obj, rules_list in\
-                            self._get_rental_push_pull_rules().items():
-                        for rule in rules_list:
-                            self.env[obj].create(rule)
+                for rule_vals in self._get_rental_push_pull_rules():
+                    self.env['stock.rule'].create(rule_vals)
             else:
-                for warehouse in self:
-                    pull_rules_to_delete = self.env['procurement.rule'].search(
+                for wh in self:
```
nikul-serpentcs, line 170: @alexis-via why changed `warehouse -> wh`?

sale_start_end_dates/models/sale_order.py
```diff
@@ -104,45 +111,24 @@ def end_date_change(self):
...
-            elif self.end_date:
-                self.start_date = fields.Date.from_string(self.end_date) -\
-                    relativedelta(days=self.number_of_days - 1)
 
     @api.onchange('product_id')
     def start_end_dates_product_id_change(self):
-        if self.order_id.default_start_date:
-            self.start_date = self.order_id.default_start_date
+        if self.product_id.must_have_dates:
+            if self.order_id.default_start_date:
```
nikul-serpentcs, line 124: @alexis-via No Needed `else` 
```
self.start_date = self.end_date = False
if self.order_id.default_start_date:
    self.start_date = self.order_id.default_start_date
if self.order_id.default_end_date:
    self.end_date = self.order_id.default_end_date
```

pedrobaeza:
@alexis-via I have told you several times: don't put several modules in the same PR. Please migrate the dependency first, and then the other module. And Travis and Runbot are red. Closing meanwhile.

sale_rental/models/stock.py
alexis-via: To have shorter variable names, therefore make it easier to respect the 79-chars per line rule

sale_start_end_dates/models/sale_order.py
alexis-via: With the code you propose, it first writes False then writes the default date. With my code, it only writes once. That's why I think it's better to keep the else.

alexis-via:
@pedrobaeza These 2 modules go together, that's why I ported them together in the same PR. You can't test sale_rental without sale_start_end_dates. So, in real life, it makes things more complex to have 2 PR than have 1 PR both for the developer and the reviewer (easier to test on runbot).
I added account-closing to oca_dep ; Travis should be better on the next run.

pedrobaeza:
No, they don't go together. You can migrate `sale_start_end_dates` first and then, the other, as the rest of the OCA contributors do.

alexis-via:
@pedrobaeza Travis is green now.
Yes, it is possible to migrate one by one. But it takes more time and it makes things more complex for developers and testers. sale_start_end_dates is just a small technical module that is only used by sale_rental for the moment. That's why I decided to migrate them both in the same PR.

pedrobaeza:
Sorry, but not this time (and not in the future). You always bypass OCA rules in this and has been tolerated, but it's not more time (it's only to be organized), and your PRs become unreviewable due to the high number of changes/commits/files. If you don't want to do it, other will do.

alexis-via:
@pedrobaeza I think this kind of rule should be applied intelligently. In this case, it makes more sense to port these 2 modules together. Otherwise, you cannot test sale_rental in the runbot until sale_start_end_dates is merged, so it makes things more difficult for testers and reviewers.
Please re-open this PR.

pedrobaeza:
@alexis-via that's always your argument, and thanks to that, we are still suffering the whole bank-payment mega-mix of commits each time we migrate them to newer versions (and we do it a module each time). You can migrate independently each of the modules, and it makes sense per se, so please do it.

rafaelbn:
@alexis-via could you please make two different PRs? I will review fast and they will be merged quickly 🙏

alexis-via:
@rafaelbn I won't do it myself. Feel free to do it.

pedrobaeza:
You are wasting more time replying than the time it takes to split into 2 PRs. That's very logical... but as you prefer.

PR #903, opened 2019-07-19
[IMP] sale_elaboration: Reset "is_elaboration" in order line when user changes product

sergio-teruel:
Prevent lines setted as elaboration after copying a sale order.
cc @Tecnativa

### DIFF
--- a/sale_elaboration/models/sale_order.py
+++ b/sale_elaboration/models/sale_order.py
@@ -70,3 +70,9 @@ class SaleOrderLine(models.Model):
         if self.is_elaboration:
             vals['name'] = '{} - {}'.format(self.order_id.name, self.name)
         return vals
+
+    @api.onchange('product_id')
+    def product_id_change(self):
+        result = super().product_id_change()
+        self.is_elaboration = self.product_id.is_elaboration
+        return result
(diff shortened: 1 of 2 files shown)


sale_elaboration/models/sale_order.py
```diff
@@ -70,3 +70,7 @@ def _prepare_invoice_line(self, qty):
         if self.is_elaboration:
             vals['name'] = '{} - {}'.format(self.order_id.name, self.name)
         return vals
+
+    @api.onchange('product_id')
+    def onchange_sale_elaboration_product_id(self):
```
pedrobaeza: Please override standard onchange instead of creating a new one, calling super. Any way, why not putting `is_elaboration` as a related field?

sergio-teruel: Because this line always will be an elaboration although you change it in product for all the lines that were created.
In this case is not needed call to super product_id change

rafaelbn:
@sergio-teruel travis is red, is related to this PR? :nerd_face:

rafaelbn (approved):
Functionally only

pedrobaeza (approved):

pedrobaeza (changes requested):
@sergio-teruel tests are affected by this change

PR #904, opened 2019-07-19
[12.0][MIG] sale_product_multi_add

HviorForgeFlow:
Standard migration

CC @Eficent


sale_product_multi_add/wizards/sale_import_products_view.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    
+    <record id="view_import_product_to_sale" model="ir.ui.view">
+        <field name="name">import.products.to.sale.form</field>
+        <field name="model">sale.import.products</field>
+        <field name="arch" type="xml">
+        <form string="Products">
+        <field name="products" />
+        <footer>
```
TimLai125: It will be nice if the indentation can be fixed in this xml file.

TimLai125 (changes requested):

sale_product_multi_add/wizards/sale_import_products_view.xml
HviorForgeFlow: Thanks for code review :smiley:

lfreeke:
Is there a reason why the line number will be 9,999 when I add the products via the add multiple products wizard?

HaraldPanten (approved):
Functional review OK, with the exception of the sale order line sequence.

sale_product_multi_add/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2016 Cédric Pigeon, ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Sale Product Multi Add",
+    'summary': """
+        Sale Product Multi Add """,
+    'author': 'ACSONE SA/NV, Odoo Community Association (OCA)',
+    'website': "http://acsone.eu",
```
nikul-serpentcs: @hveficent  URL should be `https://github.com/OCA/sale-workflow`

HviorForgeFlow:
> Is there a reason why the line number will be 9,999 when I add the products via the add multiple products wizard?

It is related with definition of sequence in sale.order.line on sale_order_line_sequence module.
The default is set to 9999 and as this module create sale order lines from a wizard it uses the default sequence in all products. What I noticed also is that if you confirm the order that sequence is rebuild properly.

sale_product_multi_add/__manifest__.py
HviorForgeFlow: Done

lfreeke:
> > Is there a reason why the line number will be 9,999 when I add the products via the add multiple products wizard?
> 
> It is related with definition of sequence in sale.order.line on sale_order_line_sequence module.
> The default is set to 9999 and as this module create sale order lines from a wizard it uses the default sequence in all products. What I noticed also is that if you confirm the order that sequence is rebuild properly.

It looks a little bit weird but if it rebuilds after confirm i'm fine with it :+1:

lfreeke (approved):
Approve (functional test)

HviorForgeFlow:
@TimLai125 could you update your review? thanks.

rafaelbn (approved):

PR #866, opened 2019-06-07
[12.0][MIG] sale_fixed_discount

MiquelRForgeFlow:
Standard migration.

PR #910, opened 2019-08-02
[MIG] Migration of sale_fixed_discount from v11 to v12

BT-dherreros:
Migration of the sale_fixed_discount module that was available on v11 to v12


LoisRForgeFlow:
There is already a open proposal, https://github.com/OCA/sale-workflow/pull/866, please review ti instead.

Please, remember to check for open PR before starting to work right away, it could save you time :)

BT-dherreros:
Ok I will close this PR and contribute on #866

PR #916, opened 2019-08-13
[10.0] sale_exception: Remove side effect from api.constrains

guewen:
The method called by 'sale_check_exception' has a side effect, it writes
on 'exception.rule' + on the Many2many relation between it and
sale.order(.line). When decorated by @api.constrains, any error during
the method will be caught and re-raised as "ValidationError".
This part of code is very prone to concurrent updates as 2 sales having
the same exception will both write on the same 'exception.rule'.
A concurrent update (OperationalError) is re-raised as ValidationError,
and then is not retried properly.

Calling the same method in create/write has the same effect than
@api.constrains without shadowing the exception type.

Full explanation:
https://github.com/OCA/server-tools/issues/1642

PR #917, opened 2019-08-14
[12.0] sale_exception: Remove side effect from api.constrains

guewen:
Port of #916

The method called by 'sale_check_exception' has a side effect, it writes
on 'exception.rule' + on the Many2many relation between it and
sale.order(.line). When decorated by @api.constrains, any error during
the method will be caught and re-raised as "ValidationError".
This part of code is very prone to concurrent updates as 2 sales having
the same exception will both write on the same 'exception.rule'.
A concurrent update (OperationalError) is re-raised as ValidationError,
and then is not retried properly.

Calling the same method in create/write has the same effect than
@api.constrains without shadowing the exception type.

Full explanation:
https://github.com/OCA/server-tools/issues/1642

### DIFF
--- a/sale_exception/models/sale.py
+++ b/sale_exception/models/sale.py
@@ -48,7 +48,31 @@ class SaleOrder(models.Model):
         order_set.detect_exceptions()
         return True
 
-    @api.constrains('ignore_exception', 'order_line', 'state')
+    def _fields_trigger_check_exception(self):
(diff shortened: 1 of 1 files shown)


yvaucher (approved):

rousseldenis (approved):
Code review

MiquelRForgeFlow (approved):

PR #919, opened 2019-08-15
[9.0] sale_automatic_workflow: avoid domain confusion

sbidoul:
Domain fields are for display only.
Make them related readonly to avoid confusion.
Partial backport of 921bdc9e3d962c4066be22529dbf6caf00c6eafd

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "2.7"
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/sale_automatic_workflow/models/sale_workflow_process.py
+++ b/sale_automatic_workflow/models/sale_workflow_process.py
@@ -35,25 +35,28 @@ class SaleWorkflowProcess(models.Model):
         default='direct',
     )
     validate_order = fields.Boolean(string='Validate Order')
-    order_filter_domain = fields.Char(
+    order_filter_domain = fields.Text(
         string='Order Filter Domain',
-        default="[('state', '=', 'draft')]"
+        related='order_filter_id.domain',
+        readonly=True,
     )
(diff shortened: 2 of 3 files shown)


guewen (approved):

lmignon (approved):

sbidoul:
postgres is not running anymore by default on travis?

lmignon:
@sbidoul Indeed, you must specify postgresql into travis.yml
```yaml
addons:
  postgresql: "9.6"
  apt:
    packages:
      - expect-dev  # provides unbuffer utility
...
```

pedrobaeza (approved):

pedrobaeza (changes requested):
There's a problem... ir.filters.domain is a `Text` field, and related fields are `Char` ones, so you either change the type of the fields, or do a computed method for transferring data.

And I see another potential problem: now fields are not stored, so this can impact on performance.

guewen:
> And I see another potential problem: now fields are not stored, so this can impact on performance.

For the processing of the automatic workflow, these fields were not used:

https://github.com/OCA/sale-workflow/blob/d55a30b29eb8170961c399f6f0f2ce35fe594e50/sale_automatic_workflow/models/automatic_workflow_job.py#L88-L109

Perfs will be the same

pedrobaeza:
OK, then it only rest the other thing, switching field types should be enough then, but maybe UI should adapted to new widget.

lmignon:
@pedrobaeza The web_domain_field widget is not available into 9.0. Requested changes done.
Regards,
lmi

pedrobaeza:
Well, I mean text widget (multi-line) against char widget (single-line). I think you can put `widget="char"` in the view for mimicking old layout, or see if new widgets fit well on the view.

lmignon:
@pedrobaeza The view is updated with `widget="char"` :smirk:

PR #923, opened 2019-08-22
[12.0][ADD] sale_product_pack

ernestotejeda:
Depends on: 

- [ ] product_pack - OCA/product-attribute#498

Cc @Tecnativa


ernestotejeda:
Changes done according to changes done in product_pack https://github.com/OCA/product-attribute/pull/498

chienandalu (commented):
I find strange the behavior of  "None - detailed" option: if I confirm an order with a pack composed of stockable products I'd expect to have the picking to deliver them, but it doesn't happen.

So either the option should go away or implement the explode of the pack contents (similar to the way MRP BoM kits do)

pedrobaeza:
@chienandalu can it be that's part of still missing `stock_product_pack` module?

ernestotejeda:
Superseded by https://github.com/OCA/product-pack/pull/2

PR #926, opened 2019-08-27
[IMP] sale_order_general_discount: Apply general discount to order lines created without UI form view

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_order_general_discount/models/sale.py
+++ b/sale_order_general_discount/models/sale.py
@@ -48,3 +48,18 @@ class SaleOrder(models.Model):
                 order_line_field.attrib['context'] = context
                 res['arch'] = etree.tostring(order_xml)
         return res
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.model
+    def create(self, vals):
+        """Apply general discount for sale order lines which are not created
+        from sale order form view.
+        """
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Bravo for including now always tests for bugfixes!

sale_order_general_discount/models/sale.py
```diff
@@ -48,3 +48,18 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
...
+
+    @api.model
+    def create(self, vals):
+        """Apply general discount for sale order lines which are not created
+        from sale order form view.
+        """
+        if 'discount' not in vals and 'order_id' in vals:
+            sale_order = self.env['sale.order'].browse(vals['order_id'])
+            if sale_order.general_discount:
+                vals['discount'] = sale_order.general_discount
```
victor-ditay, line 64: LGTM but could be done also with an default value on `discount` from `order_id.general_discount`?

victor-ditay (approved):

sale_order_general_discount/models/sale.py
pedrobaeza: In default, you don't have the `order_id` to determine the applied discount.

rafaelbn:
Please take care of the improvements in purchase: https://github.com/OCA/purchase-workflow/pull/741

PR #930, opened 2019-09-04
[IMP] sale_order_product_recommendation: Display only active or sale ok products in recommendation wizard

sergio-teruel:
cc @tecnativa TT19289 TT19978

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -46,6 +46,22 @@ class SaleOrderRecommendation(models.TransientModel):
     def _default_order_id(self):
         return self.env.context.get("active_id", False)
 
+    def _recomendable_sale_order_lines_domain(self):
+        """Domain to find recent SO lines."""
+        start = datetime.now() - timedelta(days=self.months * 30)
+        start = fields.Datetime.to_string(start)
+        other_sales = self.env["sale.order"].search([
+            ("partner_id", "child_of",
+             self.order_id.partner_id.commercial_partner_id.id),
+            ("date_order", ">=", start),
+        ])
+        return [
+            ("order_id", "in", (other_sales - self.order_id).ids),
+            ("product_id.active", "=", True),
+            ("product_id.sale_ok", "=", True),
+            ("qty_delivered", "!=", 0.0),
+        ]
+
     @api.multi
     @api.onchange("order_id", "months", "line_amount")
     def _generate_recommendations(self):
@@ -57,21 +73,9 @@ class SaleOrderRecommendation(models.TransientModel):
         if self.last_compute == last_compute:
             return
         self.last_compute = last_compute
-        start = datetime.now() - timedelta(days=self.months * 30)
-        start = fields.Datetime.to_string(start)
-        self.line_ids = False
         # Search delivered products in previous months
-        sales = self.env["sale.order"].search([
-            ("partner_id", "child_of",
-             self.order_id.partner_id.commercial_partner_id.id),
-            ("date_order", ">=", start),
-        ])
         found_lines = self.env["sale.order.line"].read_group(
-            [
-                ("order_id", "in", sales.ids),
-                '|', ("qty_delivered", "!=", 0.0),
-                     ("order_id", "=", self.order_id.id),
-            ],
+            self._recomendable_sale_order_lines_domain(),
             ["product_id", "qty_delivered"],
             ["product_id"],
         )
@@ -85,36 +89,37 @@ class SaleOrderRecommendation(models.TransientModel):
             reverse=True,
         )
         found_dict = {l["product_id"][0]: l for l in found_lines}
-        RecomendationLine = self.env["sale.order.recommendation.line"]
-        existing_product_ids = []
-        # Add products from sale order lines
+        recommendation_lines = self.env["sale.order.recommendation.line"]
+        existing_product_ids = set()
+        # Always recommend all products already present in the linked SO
         for line in self.order_id.order_line:
-            found_line = found_dict[line.product_id.id]
-            new_line = RecomendationLine.new({
+            found_line = found_dict.get(line.product_id.id, {})
+            new_line = recommendation_lines.new({
                 "product_id": line.product_id.id,
-                "times_delivered": found_line["product_id_count"],
-                "units_delivered": found_line["qty_delivered"],
+                "times_delivered": found_line.get("product_id_count", 0),
+                "units_delivered": found_line.get("qty_delivered", 0),
                 "units_included": line.product_uom_qty,
                 "sale_line_id": line.id,
             })
-            self.line_ids += new_line
-            existing_product_ids.append(line.product_id.id)
-        # Add those recommendations too
+            recommendation_lines += new_line
+            existing_product_ids.add(line.product_id.id)
+        # Add recent SO recommendations too
         i = 0
         for line in found_lines:
             if line["product_id"][0] in existing_product_ids:
                 continue
-            new_line = RecomendationLine.new({
+            new_line = recommendation_lines.new({
                 "product_id": line["product_id"][0],
                 "times_delivered": line["product_id_count"],
                 "units_delivered": line["qty_delivered"],
             })
-            self.line_ids += new_line
+            recommendation_lines += new_line
             # limit number of results. It has to be done here, as we need to
             # populate all results first, for being able to select best matches
             i += 1
-            if i == self.line_amount:
+            if i >= self.line_amount:
                 break
+        self.line_ids = recommendation_lines
 
     @api.multi
     def action_accept(self):
--- a/sale_order_product_recommendation/tests/test_recommendation.py
+++ b/sale_order_product_recommendation/tests/test_recommendation.py
@@ -71,7 +71,7 @@ class RecommendationCase(SavepointCase):
         self.assertEqual(wizard.order_id, self.new_so)
         # Product 12 is first recommendation because it's in the SO already
         self.assertEqual(wizard.line_ids[0].product_id, self.product_12)
-        self.assertEqual(wizard.line_ids[0].times_delivered, 2)
+        self.assertEqual(wizard.line_ids[0].times_delivered, 1)
         self.assertEqual(wizard.line_ids[0].units_delivered, 25)
         self.assertEqual(wizard.line_ids[0].units_included, 3)
         # Product 46 appears second
@@ -88,6 +88,19 @@ class RecommendationCase(SavepointCase):
         wizard.line_amount = 1
         wizard._generate_recommendations()
         self.assertEqual(len(wizard.line_ids), 2)
+
+    def test_recommendations_archived_product(self):
+        self.env["sale.order"].create({
+            "partner_id": self.camptocamp.id,
+        })
+        self.product_12.active = False
+        self.product_10.sale_ok = False
+        wizard = self.wizard()
+        wizard._generate_recommendations()
+        self.assertFalse(
+            self.product_12 in wizard.line_ids.mapped('product_id'))
+        self.assertFalse(
+            self.product_10 in wizard.line_ids.mapped('product_id'))
 
     def test_transfer(self):
         """Products get transferred to SO."""


carlosdauden (approved):

rafaelbn (commented):
- Make sales to a partner to have products for the recomendation. Example SO123 with 3 products.
- Duplicate SO123, and the new is SO124, clic in Recommended product and you get:

Error:
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 653, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 312, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 695, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 344, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 97, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 337, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 939, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 517, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 934, in call_kw
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 926, in _call_kw
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
  File "/.repo_requirements/odoo/odoo/models.py", line 5222, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/.repo_requirements/odoo/odoo/models.py", line 5039, in _onchange_eval
    method_res = method(self)
  File "/home/odoo/build/OCA/sale-workflow/sale_order_product_recommendation/wizards/sale_order_recommendation.py", line 98, in _generate_recommendations
    found_line = found_dict[line.product_id.id]
KeyError: 31

yajo (approved):
I was wondering why this happened. TIL [`active_test` is ignored by `read()`](https://github.com/odoo/odoo/blob/cd1b56b41fb6e5b28d22e33867bea31343475fe8/odoo/models.py#L4270), and I guess it also happens with `read_group()`.

pedrobaeza:
Before merging, I would like a test in production of times before and after this patch.

yajo:
What possible bottlenecks do you see introduced here?

pedrobaeza:
An extra domain on submodel, that introduces by design of ORM a query with `IN` ids, which is slow with a lot of records.

sergio-teruel:
Sorry!!!! but I do not add any "IN" clause...

pedrobaeza:
I know, @sergio-teruel, it's not you, it's the ORM. Let me explain:

- You a have an ORM domain that is `[('product_id.active', '=', True)]`.
- Odoo queries `product_product` table for getting ids that are active. Imagine this returns `[1, 2, 4]`.
- Then Odoo performs this domain in the main table: `[('product_id', 'in', [1, 2, 4])]`. 
- That query is very slow with a high number of elements in the `in` part.

sergio-teruel:
Ohh!! Yeah... I understood!!

sergio-teruel:
@rafaelbn Fixed error done!!

pedrobaeza:
And why doing `filtered` afterwards? You should do one or the other, nor both.

sergio-teruel:
No, because if you have a sale order with products... after that you disable any product of this sale order... and click on recommendations the methos generated_recommendations try fill the wizard lines with the orderline products and others, so if any has been disabed you obtain key error...

pedrobaeza:
Well, I vote for removing then the domain part.

sergio-teruel:
No, but what do you think if I add this part of code in line https://github.com/OCA/sale-workflow/pull/930/files#diff-bf4bc078958d60f65a62f59551d2e6e1R114

`product = self.env['product.product'].browse(line["product_id"][0])`
`if not product.active or not product.sale_ok:`
`continue`

sergio-teruel:
I don't see disable products in recommended lines

pedrobaeza:
OK, as I'm on holidays, I let the rest of the reviewers to decide the final patch. I just wanted to warn about the performance impact.

yajo:
I'm not sure about the underlying query result, but it would be strange for me that it does that `IN` stuff. That query is quite easy for SQL and can be solved with a `JOIN` and/or a `WHERE` clause. Maybe you can enable sql query logging and see the result. If it were to use a `IN`, theoretically that wouldn't have performance impact in a postgres server that has `work_mem = 32MB`, so again probably not harming.

*Just thinking out loud, I have not analyzed this in deep.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -87,8 +91,9 @@ def _generate_recommendations(self):
         found_dict = {l["product_id"][0]: l for l in found_lines}
         RecomendationLine = self.env["sale.order.recommendation.line"]
         existing_product_ids = []
-        # Add products from sale order lines
-        for line in self.order_id.order_line:
+        # Add products from sale order lines that are active and can be sold
+        for line in self.order_id.order_line.filtered(
+                lambda x: x.product_id.active and x.product_id.sale_ok):
```
yajo:
```suggestion
                lambda x: x.product_id.active and x.product_id.sale_ok):
            if not (line.product_id.active and line.product_id.sale_ok):
                continue
```

This should avoid double-looping.

pedrobaeza:
`JOIN` is only done if the relation field has `auto_join=True`. The explanation is that if not, record rules are not applied (as well as active field ORM filter).

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -100,9 +105,13 @@ def _generate_recommendations(self):
             self.line_ids += new_line
             existing_product_ids.append(line.product_id.id)
         # Add those recommendations too
+        products_allowed_ids = self.env['product.product'].browse(
+            [l["product_id"][0] for l in found_lines]).filtered(
+            lambda x: x.active and x.sale_ok).ids
```
yajo: I'd say this makes more sense here:

```python
products_allowed_ids = set(self.env["product.product"].search([
    ("id", "in", [l["product_id"][0] for l in found_lines]),
    ("id", "not in", existing_product_ids),
    ("active", "=", True),
    ("sale_ok", "=", True),
).ids)
```

You can even remove the `active=true` test because it's automatic when you're filtering on the main model. I left it just to let you understand better the change.

If you're gonna make a query, then at least you save the latter python-side filtering.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -100,9 +105,13 @@ def _generate_recommendations(self):
...
         i = 0
         for line in found_lines:
-            if line["product_id"][0] in existing_product_ids:
+            if (line["product_id"][0] in existing_product_ids or
```
Yajo: If you do the above, this `if` should change accordingly.

yajo (changes requested):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
sergio-teruel: In this loop I have actives and sale_or products, so I do nor need this condition

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,14 @@ class SaleOrderRecommendation(models.TransientModel):
     def _default_order_id(self):
         return self.env.context.get("active_id", False)
 
+    def _prepare_sale_order_line_domain(self, sales):
+        return [
+            ("order_id", "in", sales.ids),
+            '|',
+            ("qty_delivered", "!=", 0.0),
+            ("order_id", "=", self.order_id.id),
+            ]
```
yajo:
```suggestion
        ]
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
yajo: Yes, I mean that you can remove the `.filtered()` and use a normal condition instead, to avoid double-looping.

Other wise Odoo loops once for the filtered and a 2nd time for the rest of execution.

sergio-teruel:
Changes done

sale_order_product_recommendation/wizards/sale_order_recommendation.py
Yajo:
```suggestion
            if line["product_id"][0] not in products_allowed_ids:
```

You don't need now the existing_product_ids check because it's applied already in the domain above.

rafaelbn (approved):
:+1:

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,14 @@ class SaleOrderRecommendation(models.TransientModel):
     def _default_order_id(self):
         return self.env.context.get("active_id", False)
 
+    def _prepare_sale_order_line_domain(self, sales):
```
yajo: I think this title is a bit misleading.

```suggestion
    def _prepare_sale_order_line_recommendations_domain(self, sales):
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -100,9 +106,15 @@ def _generate_recommendations(self):
             self.line_ids += new_line
             existing_product_ids.append(line.product_id.id)
         # Add those recommendations too
+        products_allowed_ids = set(self.env["product.product"].search([
+            ("id", "in", [l["product_id"][0] for l in found_lines]),
+            ("id", "not in", existing_product_ids),
+            ("sale_ok", "=", True),
```
yajo: I think you're repeating this part, which is done at line 96.

```suggestion
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -100,9 +106,15 @@ def _generate_recommendations(self):
...
+        ]).ids)
         i = 0
         for line in found_lines:
-            if line["product_id"][0] in existing_product_ids:
+            if (line["product_id"][0] in existing_product_ids or
+                    line["product_id"][0] not in products_allowed_ids):
```
yajo: This should be:

```suggestion
            if line["product_id"][0] not in products_allowed_ids:
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -67,11 +75,7 @@ def _generate_recommendations(self):
             ("date_order", ">=", start),
         ])
         found_lines = self.env["sale.order.line"].read_group(
-            [
-                ("order_id", "in", sales.ids),
-                '|', ("qty_delivered", "!=", 0.0),
-                     ("order_id", "=", self.order_id.id),
-            ],
+            self._prepare_sale_order_line_domain(sales),
```
yajo:
```suggestion
            self._prepare_sale_order_line_recommendations_domain(sales),
```

sergio-teruel:
@Yajo please, Can you takes up this job?? thanks

sergio-teruel:
TT19289

yajo (approved):

pedrobaeza:
Wait a moment! I have deleted the merging branch.

This is installed in several customers and we are modifying the behavior, which can be not desirable for the rest. We should introduce a flag for allowing to preserve previous behavior.

yajo:
It seems that all our customers agree that this is the expected behavior, so we can safely merge.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,15 @@ class SaleOrderRecommendation(models.TransientModel):
     def _default_order_id(self):
         return self.env.context.get("active_id", False)
 
+    def _recomendable_sale_order_lines_domain(self, so_ids):
```
pedrobaeza: Please pass always that you are able to recordsets, not ids

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,15 @@ class SaleOrderRecommendation(models.TransientModel):
...
+        return [
+            ("order_id", "!=", self.order_id.id),
+            ("order_id", "in", so_ids),
+            ("product_id.active", "=", True),
```
pedrobaeza: This is redundant, as Odoo already do that

pedrobaeza:
Now Travis is red

yajo:
:man_facepalming: I removed one thing with dates thinking I was in v12, sorry, it's back.

I have also removed these lines now: https://github.com/OCA/sale-workflow/blob/cc0a9f2e06480bf8c564b748ee37512dba958866/sale_order_product_recommendation/wizards/sale_order_recommendation.py#L96-L97

I think it's a mistake. I didn't understand what was this doing, but now that I do, it shouldn't be there.

This process was doing:

1. Find all products already in the current SO.
1. Filter out those that are not `active` or `sale_ok`.
1. Put them all on top of the recommendation list.
1. Find recent products in recent SO.
1. Add them as recommendations.

Step 2 seems wrong IMHO because, if for any chance the product that shouldn't be saleable is found in the present SO, you still want it there to be able to use the add/remove buttons easily.

What do you think @sergio-teruel?

carlosdauden (approved):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,23 @@ class SaleOrderRecommendation(models.TransientModel):
...
+        """Domain to find recent SO lines."""
+        start = datetime.now() - timedelta(days=self.months * 30)
+        start = fields.Datetime.to_string(start)
+        other_sales = self.env["sale.order"].search([
+            ("partner_id", "child_of",
+             self.order_id.partner_id.commercial_partner_id.id),
+            ("date_order", ">=", start),
+        ])
+        return [
+            ("order_id", "!=", self.order_id.id),
```
pedrobaeza: Don't do this for performance, but be sure `self.order_id` is not in `other_sales`

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -46,6 +46,23 @@ class SaleOrderRecommendation(models.TransientModel):
...
+            ("order_id", "in", other_sales.ids),
+            ("product_id.active", "=", True),
```
pedrobaeza, line 60: As most of the products will be active and probably `sale_ok`, I would bet for doing this in a later filter, but I'm not sure if you have already tried this and measure performance

pedrobaeza (changes requested):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
yajo: I don't understand... What's the performance problem of excluding an ID? That's an indexed column and should be blazing fast... :confused:

sale_order_product_recommendation/wizards/sale_order_recommendation.py
yajo: Given we have to ask the DB for a recordset, it's better to filter it directly in the DB. It's always faster than downloading extra recordsets just to ignore them later in Python code.

If the recordsets were already cached, that could be a different story, but it's not the case here.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: You are using 2 times `order_id`. Just do it once.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: The problem is that this is not auto-joined, so it actually requires several queries + the problem with `in` operator in SQL a list with thousand of elements.

yajo: I have been profiling this.

I didn't understand what you were saying, but now I see that indeed these filters are made in separate SQL queries. That's quite strange, since it could just be added to the same query. I guess that's to apply permissions...

Anyways, removing these leafs from the domain we save 2 queries and ~7ms in total, but later we'd have to browse all product records and manually prefetch their `active` and `sale_ok` fields to actually end up saving 1 query, at the cost of code readability.

IMHO it's not worth it.

pedrobaeza: We would save that queries if `product_id` has `auto_join=True`, but that disables ACL check. In this case, maybe it makes sense to add it. In current state, the problem is not the extra 2 queries, but having thousand of product, creating the third query: `('product_id', 'in', [<thousand of ids>])`, which the slowest one. Have you tried?

yajo: I haven't tried that, but I doubt that is a performance problem.

I think you are in a confusion with the performance problems we have faced in the past with subqueries, not for lists of ids. I mean, it's more things like `where id in (select id from lalala)`, because if the server has not enough worker memory, it executes the subquery record by record and the execution time grows exponentially. But that's a completely different case.

Even in Python, if you use the `in` operator with a `set()`, you get instant results no matter the set size. I imagine Postgres uses a similar optimization under the hood to avoid that problem.

I'd say the best we can do is merge and then patch it if we really see some real bottleneck in production. Otherwise we might be wasting time in fixing something that is not broken.

pedrobaeza: Seen directly in a meeting and it doesn't seem to affect the performance, so we keep it this way.

pedrobaeza (approved):

pedrobaeza:
Cherry-picked and adapted for v12 in 40f8e9e39e000292b4b44e43d7d01ae4c115c2a1 and v13 in 0614cb6f7975c14563d431e983e91c1e984b27af

PR #936, opened 2019-09-23
[10.0][ADD] add module partner_picking_policy

acsonefho:
Add a new module `partner_picking_policy`

**Current behavior:**
You can specify a global picking policy into sale settings.

**New behavior:**
You can define a picking policy by partners/customers.
You can let the field empty (on partner) to use the global rule (into sale settings)


pedrobaeza:
Module sale_order_type already includes this setting:

https://github.com/OCA/sale-workflow/blob/10.0/sale_order_type/models/sale_order_type.py#L36

and it's more flexible, having also the option to define a sale type per partner. I think it's better to use it.

acsonefho:
@pedrobaeza Oh ok thank you! I didn't see this module!

PR #939, opened 2019-09-24
[IMP][12.0][sale_order_price_recalculation] Adding information about the order to recalculate_names line2 add 'order_id': self

masteriusz:
Adding information about the order to recalculate_names line2 dummy order line will allow for using the correct language. Without this fix, the recalculate_names function was useless in multi-language environments.

### DIFF
--- a/sale_order_price_recalculation/models/sale_order.py
+++ b/sale_order_price_recalculation/models/sale_order.py
@@ -32,6 +32,7 @@ class SaleOrder(models.Model):
             # we make this to isolate changed values:
             line2 = self.env['sale.order.line'].new({
                 'product_id': line.product_id,
+                'order_id': self
             })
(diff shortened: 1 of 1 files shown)


rousseldenis (changes requested):
Code Review. @masteriusz Could you add a test for that ?

masteriusz:
It’s been some time since I did some programming. I made the change I proposed in my own working instance and it’s been doing its job flawlessly for the last three years.
After all it’s a single line of code, including order information at the time of creating order lines. 
Writing a correct test after three years - I won’t be able to do it anymore, I would have to relearn everything from scratch.
Cheers!

rousseldenis (approved):
Ok, nevertheless, it seems legit.

Let's merge.

/ocabot merge patch

PR #940, opened 2019-09-24
[12.0] Add multi-country multi-language Sale General Terms

### LINKED ISSUE #938: [RFC] Sale General Terms
### Specification

This Module should allow to create sales terms ("General Selling Consitions") for different countries and languages.

- It should allow to work on drafts version of the existing terms.
- Suggest matching sales terms based on Customer's country and language
- The matching sales terms should be selected and shown as the last page(s) of the Quotation and Sale Order.

dreispt:
Closes #938


daramousk:
As I understand you want the contents of a field to be translated into a language specified by another field. Odoo by default on the sale order provides a field named `note` which pretty much is used to hold the terms and conditions.

From what I see the `note` field is a `Text` field which makes it a subclass of `_String`. I see that over here https://github.com/OCA/OCB/blob/12.0/odoo/fields.py#L1371 you can set a `callable` on the `translate` attribute of a field.

What if you did that and fetched the lang from the `partner_id.lang` in order to get the translated text? Would that help here?

dreispt:
@daramousk It would be simpler that that - the standard Sales Terms is already a translatable field - you just need to make it visible in the Company form and you will have a multi-language text widget. And the SO report is partner language aware.
So it should work almost out of the box.

But that does not cover my requirement: multi-country and be able to be working on draft versions, while still having an enabled version.

mgosai (approved):

sale_general_term/demo/sale_general_term.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903:
```suggestion
```

sale_general_term/report/sale_report_templates.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903:
```suggestion
```

sale_general_term/security/sale_general_term.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903:
```suggestion
```

sale_general_term/views/sale_general_term.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903:
```suggestion
```

sale_general_term/views/sale_order.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903:
```suggestion
```

sale_general_term/views/sale_order.xml
```diff
@@ -0,0 +1,19 @@
...
+<!-- Copyright 2019 Open Source Integrators
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record model="ir.ui.view" id="sale_order_form_view">
+        <field name="name">sale.order.form (in sale_general_term)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="arch" type="xml">
+
```
max3903:
```suggestion
```

sale_general_term/views/sale_order.xml
```diff
@@ -0,0 +1,19 @@
...
+            <field name="payment_term_id" position="before">
+                <field name="general_term_id" required="1" />
+            </field>
+
```
max3903:
```suggestion
```

sale_general_term/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Daniel Reis <dreis@opensourceintegrators.com> (https://opensourceintegrators.com)
```
max3903:
```suggestion
* Open Source Integrators <https://opensourceintegrators.com>

  * Daniel Reis <dreis@opensourceintegrators.com>
```

max3903 (approved):

sale_general_term/readme/CONTRIBUTORS.rst
dreispt: Max, OSI is already stated as the Author.
IMO Contributors should be actual people, whenever possible.

bodedra: It seems he wants manage contributors by company group of peoples.

dreispt:
From Rufus tests, I ensured that General Terms are printed after Payment Terms, and that we have a page break:

image

dreispt:
@rufus01 The page break was fixed. Can you confirm please?

rufus01:
> 
> 
> @rufus01 The page break was fixed. Can you confirm please?

Hi Daniel, confirmed, is working

rufus01:
@dreispt, for our use case General Terms should be mandatory on SO. 
Also I find the Enabled/ Active fields not clear, my Suggestion: 
- Rename active to archive and lets make it an action (similar to delete, duplicate)
- Enabled could be a radio button: Enabled or Draft

dreispt:
This module should be good to merge.
@rufus01 can you have a last review please?

dreispt:
I believe this is ready to merge. Thanks!

sale_general_term/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,10 @@
+To configure this module, go to 
+Sales > Configuration > Sale Orders > General Terms
+and create the General Term to be used for each countries and languages.
+
+The Countries and Languages field are optional.
+If no terms are found for a particular Country and Language combination,
+a lookup will be done for Terms with only that COuntry, and then for only
```
newtratip:
```suggestion
a lookup will be done for Terms with only that Country, and then for only
```

sale_general_term/models/sale_general_term.py
```diff
@@ -0,0 +1,72 @@
...
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class SaleGeneralTerm(models.Model):
+
+    _name = 'sale.general.term'
+    _description = 'Sale General Term'
+
+    @api.constrains('is_enabled', 'country_ids')
```
newtratip:
```suggestion
    @api.constrains('is_enabled', 'country_ids', 'lang')
```

sale_general_term/models/sale_general_term.py
```diff
@@ -0,0 +1,72 @@
...
+
+class SaleGeneralTerm(models.Model):
+
+    _name = 'sale.general.term'
+    _description = 'Sale General Term'
+
+    @api.constrains('is_enabled', 'country_ids', 'lang')
+    def _check_is_enabled_country_ids(self):
+        for term in self:
+            enabled_terms = self.search([
```
rousseldenis, line 16: Maybe do the search outside of the loop and filter inside to avoid too much queries.

sale_general_term/models/sale_general_term.py
```diff
@@ -0,0 +1,72 @@
...
+                raise ValidationError(_(
+                    'There are already enabled terms for the Language '
+                    'and at least one of the Countries selected. '
+                    'Each Country/Language is allowed only one '
+                    'enabled General Terms.'
+                ))
+
+    @api.constrains('is_enabled', 'active')
+    def _check_is_enabled_active(self):
+        for term in self:
```
rousseldenis, line 31: I'm not sure this will raise as active is filtering recordsets.

sale_general_term/models/sale_general_term.py
```diff
@@ -0,0 +1,72 @@
...
+
+        :param partner:id: a singleton Partner record
+        :returns: the General Terms record if found, False otherwise
+        """
+        if partner_id:
+            partner_id.ensure_one()
+            c1 = [('country_ids', 'in', partner_id.country_id.id)]
+            c2 = [('lang', '=', partner_id.lang or self.env.user_id.lang)]
+            for domain in [c1 + c2, c1, c2]:
+                term = self.search([('is_enabled', '=', True)] + domain)
```
rousseldenis, line 68: I would have put the domain retrieval in a function to allow override.

sale_general_term/models/sale_general_term.py
```diff
@@ -0,0 +1,72 @@
...
+        Looks for the best match based on the Partner's Country and Language.
+
+        :param partner:id: a singleton Partner record
+        :returns: the General Terms record if found, False otherwise
+        """
+        if partner_id:
+            partner_id.ensure_one()
+            c1 = [('country_ids', 'in', partner_id.country_id.id)]
+            c2 = [('lang', '=', partner_id.lang or self.env.user_id.lang)]
+            for domain in [c1 + c2, c1, c2]:
```
rousseldenis, line 67: It's a bad usage for domain building. Use osv.expression instead

rousseldenis (changes requested):
Code review. Essentially performances or usage

sale_general_term/models/sale_general_term.py
dreispt: osv.expression is useful if I need to combine unknown domains, and the join operation might not be linear.
That is not the case here, and using osv.expression does not bring any additional benefit.

sale_general_term/models/sale_general_term.py
dreispt: The self recordset can have active=False records. I expect to be the case when writing an a record setting the on the active field .
If api.constrains does not work for active=False records, that would be an ORM bug.

sale_general_term/models/sale_general_term.py
dreispt: Isn't this too early optimization?

sale_general_term/models/sale_general_term.py
rousseldenis: @dreispt

rousseldenis (commented):
@dreispt Could you add some tests ?

PR #943, opened 2019-09-30
[MIG] 12.0: sale_start_end_dates

gurneyalex:
split the migration of the module from #899


gurneyalex:
needed for #944

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @api.onchange('end_date')
+    def end_date_change(self):
+        if self.end_date:
+            if self.start_date and self.start_date > self.end_date:
+                self.start_date = self.end_date
+
+    @api.onchange('start_date')
+    def start_date_change(self):
+        if self.start_date:
```
Swapnesh-SerpentCS, line 117:
```suggestion
        if self.start_date and self.end_date and self.start_date > self.end_date:
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,134 @@
...
+                line.number_of_days = 1
+            if line.start_date:
+                line.end_date = line.start_date +\
+                    relativedelta(days=line.number_of_days - 1)
+            elif line.end_date:
+                line.start_date = line.end_date -\
+                    relativedelta(days=line.number_of_days - 1)
+        return res
+
+    @api.constrains('start_date', 'end_date')
```
Swapnesh-SerpentCS, line 82: Not sure but as we are checking for `must_have_dates` here, IMO it should be added  here as well

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,134 @@
...
+        string='Start Date', readonly=True,
+        states={'draft': [('readonly', False)]})
+    end_date = fields.Date(
+        string='End Date', readonly=True,
+        states={'draft': [('readonly', False)]})
+    number_of_days = fields.Integer(
+        compute='_compute_number_of_days',
+        inverse='_inverse_number_of_days',
+        string='Number of Days', readonly=False, store=True)
+    must_have_dates = fields.Boolean(related='product_id.must_have_dates')
```
Swapnesh-SerpentCS, line 52: As we are checking `must_have_dates` as multiple places (It will call the related function each time), Can't we make it `store=True`

sale_start_end_dates/models/sale_order.py
gurneyalex: I disagree: we don't want to block setting "must_have_dates" for a product which was sold in the past.

sale_start_end_dates/models/sale_order.py
gurneyalex: no benefit in the context of the migration of the module, in my opinion.

rafaelbn (approved):
:+1: Thank you!

@gurneyalex if possible could you increase coverage? No blocking

cc @alexis-via

rafaelbn:
PLease @alexis-via could you take a look?

pedrobaeza (approved):

sbidoul:
@sbejaoui would it make sense to have the contract module depend on this?

sbejaoui:
> @sbejaoui would it make sense to have the contract module depend on this?

It make sens.

For consistency between OCA modules, that product_contract module should depends on this one. For now both implement start_date and end_date in sale lines.

rafaelbn:
@sbidoul @sbejaoui contract module was simple and easy to use. Now is complex, you don't see this, we don't see because we spents a lot of hours on it, but contact module is not anymore a simple and easy module to use for users as it was in v11 and before.

rafaelbn:
> @sbejaoui would it make sense to have the contract module depend on this?

Not, no sense in my opinion. Why I should have start and end dates in sale order because I use contract? Users don't want any start or end dats in sale order, there are already a lot of columns @sbidoul @sbejaoui

sbejaoui:
> @sbidoul @sbejaoui contract module was simple and easy to use. Now is complex, you don't see this, we don't see because we spents a lot of hours on it, but contact module is not anymore a simple and easy module to use for users as it was in v11 and before.

Hi @rafaelbn ,

As you said, we don't see this complexity neither our users. So can you please open a issue in contract repos and let discuss what concern you.

sbejaoui:
> > @sbejaoui would it make sense to have the contract module depend on this?
> 
> Not, no sense in my opinion. Why I should have start and end dates in sale order because I use contract? Users don't want any start or end dats in sale order, there are already a lot of columns @sbidoul @sbejaoui

@rafaelbn ,
@sbidoul meant `product_contract` module. The module that allow contract creation from sale order. Currently, `product_contract` module adds `date_start` and `date_end` to sale order lines.

sbidoul:
@rafaelbn it's sure the module has become more sophisticated (and more robust too, we think). That has some UI impact. But we can surely improve UI for simple cases while preserving advanced functionality.

PR #946, opened 2019-10-01
[MIG] sale_order_archive: Migration to 13.0

espo-tony:


sale_order_archive/readme/INSTALL.rst
```diff
@@ -0,0 +1 @@
+You need to install *sale_management* module for accessing the needed menus.
```
sudhir-erpharbor, line 1: Why don't we add sale_management module in dependency as user will have to install that module?

sale_order_archive/readme/USAGE.rst
```diff
@@ -0,0 +1,14 @@
+To archive sale orders, you need to:
+
+#. Open the tree view of sale orders.
+#. Select a sale order (in status Locked or Cancelled) you want to archive.
```
sudhir-erpharbor, line 4: Just an idea, what if user archive the Sales Order which is done or in process? It should have some mechanism to prevent the user to do that.
@pedrobaeza What do you think?

espo-tony:
Replaced by #980

PR #949, opened 2019-10-03
[10.0][FIX] oca_dependencies.txt: Remove unnecessary entry

rousseldenis:
This has to be removed as partner_address_version module is merged

@bguillot @pedrobaeza


pedrobaeza:
Indeed

PR #952, opened 2019-10-04
[MIG] Migrated partner_prospect to v13

sudhir-erpharbor:


partner_prospect/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Partner Prospect",
+    "version": "13.0.1.0.0",
+    "category": "Sales Management",
```
rousseldenis: @sudhir-erpharbor Could you change this to 'Sales' as it won't appear in OCA shop in the good category?

rousseldenis (changes requested):
Code review. LGTM. Minor comment

pedrobaeza:
Does this makes sense now with new system that detect sales the same way (with the removal of sale_ok / purchase_ok fields)?

partner_prospect/__manifest__.py
sudhir-erpharbor: Done.

simahawk:
@sudhir-erpharbor can you check @pedrobaeza comment?

rousseldenis (approved):

sudhir-erpharbor:
@pedrobaeza @simahawk sale_ok & purchase_ok are the fields of product, not the partner and it's still available in v13. This module is all about Prospect Partners. 

A prospect partner will be those to which you still haven’t sold anything, considering sales as not canceled and nor draft sale orders to that partner.

It would be good if you can give more details on this regarding any change, please.

partner_prospect/views/res_partner_view.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
newtratip, line 1: I think xml'id should same original id.

newtratip (commented):
Code Review.

NCS-75:
> https://github.com/OCA/sale-workflow/pull/952#issuecomment-591970500

Hi
In reaction to @pedrobaeza question about the legitimity or need of the module, related to the changes in v13 (v12: bollean fields for customer-supplier vs v13: customer-supplier rank value), I think this module still has its place and is usefull for "crm" context.

-> However, currently, and while the "prospect" field is "read-only" on the views, this module considers that any contact created will necessarily be a prospect, including social/fiscal administrations, employees, banks, etc...

Are we going to target our bank or our fiscal office, to become  future customer ? No !

As a result, despite the very good fonctional perimeter _(eg: any virgin contact added on SO, becomes "prospect")_ ,  the module doesn't match the goal, and i even no longer see the point of this module for the crm environnement because the goal, is : 

> **Goal** : With this module you can easily distinguish if a partner is currently customer
or prospect. A prospect partner will be those to which you still haven't sold
anything, considering sales as non cancelled nor draft sale orders to that partner.

-To overcome this, and while waiting for a modification, I can on my side, quicly :
- add 2 "server-actions" restricted to "Sales\administrator" which updates the records. eg: "records['prospect'] = false/true".
- Also, I can eventually customize my views with an option: and add multi_edit="1" to provide "multi-edition feature" on what it needs to be restore.

But, IMHO, to avoid regression, misinterpretation or a reject use of what this module is supposed to bring; i suggest to authors, and by the way @sudhir-erpharbor and reviewer @rousseldenis a modification before publish on 13.0 branch. 
(At leat, modify read-only value, to allow an admin\sales user to act upon the bollean field )

Best regards.

newtratip (approved):

dreispt:
@sudhir-erpharbor Please apply precommit reformatting at it should be good to go.

hailangvn:
@sudhir-erpharbor , below code can reproduce the merge error. Please help to fix and rebase changes to this branch
```
git checkout 13.0-mig_partner_prospect
git fetch origin; git merge origin/13.0
pre-commit run -a
```

dreispt:
Superseded by https://github.com/OCA/sale-workflow/pull/1644

PR #953, opened 2019-10-04
[ADD] Add sequence field to sale.order.line form

wpichler:
when using the sale.order.line form instead of the editable tree - then the generated default sequence will get lost without that


AaronHForgeFlow (approved):
:+1: LGTM. Could you also forward port this to v12?

wpichler:
i will forward port it to v12

wpichler:
forward port in https://github.com/OCA/sale-workflow/pull/957

HviorForgeFlow (approved):

MiquelRForgeFlow (approved):

AdriaGForgeFlow (approved):

PR #963, opened 2019-10-12
remove obsolete base_sale_multi_channel reference

zeroheure:
The base_sale_multi_channel module doesn't exist for a long time but is still indicated in README. Remove it.


rvalyi (approved):

pedrobaeza (changes requested):
You have to make the change in https://github.com/OCA/sale-workflow/blob/12.0/sale_exception/readme/DESCRIPTION.rst, as the README.rst is auto-generated from there.

zeroheure:
Oh yes, I forgot that point, sorry.

pedrobaeza (approved):

PR #888, opened 2019-07-04
[12.0][MIG] sale_blanket_order: Migrations to 12.0

ruter-lyu:

PR #966, opened 2019-10-15
[12.0][MIG] sale_blanket_order module

tafaRU:


tafaRU:
@acsone @eficent any of you could please review it? Thank you in advance.

LoisRForgeFlow:
@ageficent Could you have a look?

sale_blanket_order/wizard/create_sale_orders.py
```diff
@@ -142,63 +137,53 @@ def create_sale_order(self):
...
+                'pricelist_id': pricelist_id,
+                'payment_term_id': payment_term_id,
                 'order_line': order_lines_by_customer[customer],
-            })
+            }
             sale_order = self.env['sale.order'].create(order_vals)
             res.append(sale_order.id)
         return {
             'domain': [('id', 'in', res)],
-            'name': _('Sales Orders'),
```
AdriaGForgeFlow: Why do you remove this? Shouldn't it be translatable?

AdriaGForgeFlow (commented):
Functional review: Overall works fine, I found one problem when duplicating Sale Blanket Orders the state is kept. If you have a confirmed Blanket Order and you duplicate it the newly created one will have state confirmed and name "Draft" which does not make sense.

Code review: just one comment below:

sale_blanket_order/wizard/create_sale_orders.py
tafaRU: It already is: https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_actions.py#L173

AdriaGForgeFlow: This is not an 'ir.actions.act_window' record, it is created on the go from python.
AFAIK, if you don't set the name to be translatable here, it will not appear in the pot file for its translation.

tafaRU: fixed

tafaRU:
@ageficent about your functional remark, for consinstency, I could add `default='draft` to the field `state` of the `blanket.order`. What do you think about it?

sale_blanket_order/wizard/create_sale_orders.py
tafaRU: thanks, my fault, I trusted https://github.com/OCA/sale-workflow/pull/888#discussion_r329150525 without investigating deeper :wink:

AdriaGForgeFlow:
> @ageficent about your functional remark, for consinstency, I could add `default='draft` to the field `state` of the `blanket.order`. What do you think about it?

That would not prevent the state from being copied. You need to add, in addition, the `copy=False` parameter to the field definition. 
This is an example: https://github.com/OCA/sale-workflow/pull/966/files#diff-b175af03110d728bd289598dd24e4106R49

tafaRU:
> That would not prevent the state from being copied. You need to add, in addition, the copy=False parameter to the field definition.

It already is: https://github.com/tafaRU/sale-workflow/blob/12.0-mig-sale_blanket_order/sale_blanket_order/models/blanket_orders.py#L76

What I suggested it is also add `default='draft'`.

AdriaGForgeFlow:
> > That would not prevent the state from being copied. You need to add, in addition, the copy=False parameter to the field definition.
> 
> It already is: https://github.com/tafaRU/sale-workflow/blob/12.0-mig-sale_blanket_order/sale_blanket_order/models/blanket_orders.py#L76
> 
> What I suggested it is also add `default='draft'`.

You're right. My bad, I didn't check. Then I guess with the default should do the trick.

tafaRU:
@ageficent, just force pushed new changes. Please try again when runbot build will be ready.
Thanks!

AdriaGForgeFlow (approved):
Looks good now! Thanks for the work :+1:

tafaRU:
@ageficent, just pushed a fix for a bug already existing in the previous versions. Could you please update your review?

Thank you in advance.

AdriaGForgeFlow (approved):
Code changes look good :+1: 
Thanks!

tafaRU:
@ageficent, please also review https://github.com/OCA/sale-workflow/pull/982

stevech091 (approved):
Functional OK

HviorForgeFlow (approved):
Functional review for fixed bug :+1:

tafaRU:
@OCA/crm-sales-marketing-maintainers could you please merging this?
I'd do it by myself but I don't want to violate the rule about self-merging. BTW, is it always valid?

tafaRU:
ping @OCA/crm-sales-marketing-maintainers!

eLBati:
Scope question: what's the difference from native blanket orders?
https://www.odoo.com/documentation/user/12.0/purchase/purchases/tender/blanket_orders.html

eLBati:
Never mind, this is about sale

sale_blanket_order/readme/USAGE.rst
```diff
@@ -0,0 +1,53 @@
+A new menu in the Sales area is created, allowing users to create new blanket orders.
+
+To create a new Sale Blanket Order go to the sale menu in the Sales section:
+
+.. image:: /sale_blanket_order/static/description/BO_menu.png
```
eLBati: Please use suggested syntax
https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/readme/CONFIGURE.rst
to correctly display images in both README and index

eLBati (changes requested):

sale_blanket_order/readme/USAGE.rst
tafaRU: Thanks @eLBati, I applied changes as you suggested but it only works in https://github.com/tafaRU/sale-workflow/blob/12.0-mig-sale_blanket_order/sale_blanket_order/readme/USAGE.rst
Trying to regenerate the README file I get https://raw.githubusercontent.com/tafaRU/sale-workflow/12.0-mig-sale_blanket_order/sale_blanket_order/README.rst that doesn't work.
Any idea why that might be?

HviorForgeFlow: Will work when merged

tafaRU: @hveficent, ah cool! I wasn't aware of it. Do you have any reference?

eLBati (approved):
Thanks

PR #970, opened 2019-10-20
12.0 mig sale order secondary unit

digitalsatori:


pedrobaeza:
You have to change some XML-IDs in tests that point to `product` module instead of the new `uom` module.

digitalsatori:
@pedrobaeza Thank you very much for the pointer. I've updated the test accordingly and passed the unit test locally. 

It seems the install of `sale_stock` module caused the test failure. Can you please advice how can I fix this?

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2018 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase
+
+
+class TestSaleOrderSecondaryUnit(SavepointCase):
+    at_install = False
```
pedrobaeza: @digitalsatori you have to put this as decorators instead of variables, as in v12 this is not supported this way:
```python
from odoo.tests import tagged

@tagged('post_install', '-at_install')
class TestSaleOrderSecondaryUnit(SavepointCase):
    ...
```

digitalsatori: 👍 awesome, will do as instructed .

digitalsatori:
changed and squashed.

digitalsatori:
@pedrobaeza Could you please explain what's wrong with the test? Thank you

pedrobaeza:
The failure is in other module, so maybe this module is bad interacting with the other, or this is modifying something incorrectly.

digitalsatori:
> The failure is in other module, so maybe this module is bad interacting with the other, or this is modifying something incorrectly.

Thank you @pedrobaeza for the explanation. I think this should be the related issue: #972

pedrobaeza:
Yes, it seems so

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2018 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase, tagged
+
+
+@tagged('post_install', '-at_install')
```
sergio-teruel, line 6: You can remove tags

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
sergio-teruel: @pedrobaeza Is mandatory? I think that to pass tests in OCA is not needed...

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
pedrobaeza: No, it won't work without them, as `sale_stock` is installed before this module and it raises the error `picking_warn` not set when creating products in the test.

sergio-teruel: Ok.ok

pedrobaeza:
Please rebase now that I have disabled in main branch failing test

pedrobaeza (approved):

pedrobaeza:
@sergio-teruel this will be green when merging, so you can approve it if it's OK.

sergio-teruel (approved):

PR #979, opened 2019-10-29
[FIX] 10.0 sale_triple_discount

gurneyalex:
there are cases where the cache trick done in triple_discount_preprocess
is lost when sale.order.line _compute_amount is called and processes the
1st line of the sale order => all remaining lines are impacted, and don't
have the discounts 2 and 3 taken into account.

Not updating the sale.order _amount_all method because it is unlikely that
fields on the lines depend on the total amount of the sale.

### DIFF
--- a/sale_triple_discount/README.rst
+++ b/sale_triple_discount/README.rst
@@ -80,6 +80,7 @@ Contributors
 * David Vidal <david.vidal@tecnativa.com>
 * Simone Rubino <simone.rubino@agilebg.com>
 * Jacques-Etienne Baudoux (BCIM sprl) <je@bcim.be>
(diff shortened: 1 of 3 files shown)


guewen (approved):

grindtildeath (approved):
Seems legit :+1:

AaronHForgeFlow (approved):
:+1: LGTM

rousseldenis:
@lmignon This would be interesting for you

lmignon (approved):
Makes sens. Fix after fix, these triple discount addons seems to be a nightmare. Shouldn't the overall design of these addons be redesigned? (But that may be impossible)

PR #983, opened 2019-11-04
[11.0][MIG] sale_manual_delivery

grindtildeath:
Based on: #404 
Supersedes: #922 

New feature added:
 - [x] IMP: Use a procurement group according to date planned (160d82b + 4e55bba)

TODO:
 - [ ] Converge with @aheficent 's work on #935
 - [ ] Squash commits

PR #935, opened 2019-09-17
[ADD][10.0] sale_manual_delivery

AaronHForgeFlow:
This is a proposal to supersede #404 

Basically this continuous with the work in that PR. Apart of minor bug fixing I introduced some features:

- Possibility yo deliver from sale order lines

The reason behind that is the following:

- Many companies have many long term sales. Thus, makes sense to deliver them at sale order line so we have one main screen to process all today's sales.
- In many implementations is the warehouse user the one that creates the delivery. It is very useful in those cases to have a direct access from inventory to the sale order lines that needs to create delivery for.
- We need to make this module compatible with others in this repo, that is why I included the module sale_stock_picking_blocking_sale_manual_delivery. Otherwise those two cannot be installed together. With the version at #404 that was not possible.

TODO:

- [x] Keep planning date in the wizard that creates the deliveries
- [x] Ensure different pickings are created where different planned dates are selected in the wizard

PR #987, opened 2019-11-08
[12.0][MIG] sale_manual_delivery

AaronHForgeFlow:
Based on #983 + https://github.com/grindtildeath/sale-workflow/pull/2 at 8th November

Standard migration. Functional tests done from my side and working fine IMHO. Pending todo:

- [x] squash commits
- [x] Check if glue module `sale_stock_picking_blocking_sale_manual_delivery` between this and `sale_stock_picking_blocking`  (proposed in https://github.com/OCA/sale-workflow/pull/935) is really needed to be ported to v12. --> not needed


AaronHForgeFlow:
cc @grindtildeath

stevech091 (approved):
functional ok

AaronHForgeFlow:
Added he possibility of changing the route when creating the delivery. This is useful when at the time of creating the order, we don't really know from warehouse and what location we are shipping.

In case no route is provided then the one in the sale order line will be taken.

rousseldenis (approved):
Code review

rousseldenis:
@AaronHForgeFlow If ready, could you squash ?

AaronHForgeFlow:
@rousseldenis Tried to squash by author and main topics. If the authors of the commits prefer different squashing just let me know @jgrandguillaume @Tonow-c2c @grindtildeath @vrenaville

leemannd (changes requested):
There is a regregression from the first version we did for V10 and V11.

There was the possibility to choose the address form the sale manual delivery wizard.

AaronHForgeFlow:
Hi @leemannd, in the wizard I've seen so far the possibility of changing the carrier and the planned date, but not the delivery address. It is ok to me to add that feature. Could you point me to your repo/the repo where this feature is implemented? Thanks

Also, if I include the delivery_address in the wizard, Would be ok for you to continue with this Pull requests? (same in v10 and v11)

leemannd:
@AaronHForgeFlow Thanks for asking. You have been removing it while helping on the migration of the V11 module -> https://github.com/OCA/sale-workflow/pull/983/commits/1f394ebe123faf580cad03e359093d1f42b31866#diff-ff81e585e998789b0b224c322c8cc742L19

This is present also in the original PR in V10.0 -> https://github.com/OCA/sale-workflow/pull/404/files?file-filters%5B%5D=.xml#diff-ff81e585e998789b0b224c322c8cc742R19

This is for us a blocking point as we have clients using it and needing this possibility.

AaronHForgeFlow:
@leemannd I see. I tried to revert that removal but I missed that part. 

I will add it back here, and in v10/v11

leemannd:
@AaronHForgeFlow Just ping me on them, I will update my reviews right away

AaronHForgeFlow:
@leemannd  Thanks to @mmequignon we have the delivery address back in the wizard :tada:  It would be nice if you check that everything is as expected.

Thanks.

AaronHForgeFlow:
Applied latest fixes in v11

sale_manual_delivery/models/stock_move.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2017 Denis Leemann, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields, api
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    carrier_id = fields.Many2one("delivery.carrier", string="Delivery Method")
```
phuctranfxvn, line 10: @AaronHForgeFlow, why do we need to define the carrier at the stock.move level ?
cc @i-vyshnevska

AaronHForgeFlow: It is a requirement of the original authors of the module. https://github.com/OCA/sale-workflow/pull/404#discussion_r100574843

phuctranfxvn: @AaronHForgeFlow natively Odoo defines the carrier at stock picking level (in delivery module). It confuses us if we have multiple carriers in the same picking (stock.move level), but only one deliver provider defined in picking.

In actual usage, if a picking is delivered by different providers, the picking should be splitted (partial transfer + create back order).

AaronHForgeFlow: Splits are not performed anymore with this module. In this case two separate manual deliveries will be created.

phuctranfxvn: @i-vyshnevska, do you have any idea?

leemannd (approved):

AaronHForgeFlow:
@rousseldenis is this fine to be merged?

PR #994, opened 2019-11-14
[12.0] Add sale_wishlist

simahawk:
`sale_wishlist` is an extension of `sale_product_set` and allows to define customers' wishlists.

In this PR:

* full test cov for sale_product_set
* full test cov for sale_product_set_variant
* add optional partner to product.set
* add new module sale_wishlist to deal w/ wishlist in a generic way


lmignon:
@Cedric-Pigeon This one might interest you.

rousseldenis:
@simahawk Could you check travis ?

simahawk:
@rousseldenis should be fixed now. We should probably add security tests because the rules where broken due to duplicated xid.

Implementation detail: I don't think that `sale_wishlist` needs to inherit from _variant module. 
We only need to link a single specific variant -> I've replaced the dependency.

CC @lmignon

simahawk:
FTR sale_product_set_variant implementation is too complex for its goal: add N variants from a specific template. IMO we should have a wizard to ease the process: you select a template and the variants -> one line per variant is added, then remove `product_variant_ids` from the line.
This way core logic stays the same and if you rely on `line.product_id` for other features (eg: sale_wishlist) your implementation won't be broken (or useless because product_id is not populated).

sale_product_set/wizard/product_set_add.py
```diff
@@ -10,37 +10,68 @@ class ProductSetAdd(models.TransientModel):
...
+        'sale.order', 'Sale Order', required=True,
+        default=lambda self: self.env.context.get('active_id')
+    )
     product_set_id = fields.Many2one(
         'product.set', 'Product set', required=True)
     quantity = fields.Float(
         digits=dp.get_precision('Product Unit of Measure'), required=True,
         default=1)
 
+    def _check_partner(self):
```
lmignon, line 36: @simahawk Why not defining a constrains on order_id? With a constrains you don't need to explicitly call this check into the `add_set` method. In the same time, the `onchange` on the product_set_id should not be required is we set the right domain on the product_set_id. (To help we could add a read only related field for `order_id.partner_id`, The domain into the view on product_set_id will be `[ '|', ('partner_id', '=', False), ('partner_id', '=', partner_id)]` )

lmignon (commented):
MO, the access to this functionality should be moved from the sale-> condiguration  -> sale order to sale -> products 
image

sale_wishlist/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
...
+
+    wishlists_count = fields.Integer(
+        compute='_compute_wishlists_count',
+        string="# Wishlists"
+    )
+
+    def _compute_wishlists_count(self):
+        # do just one query for all records
+        data = self.env['product.set'].read_group([
+            ('partner_id', 'in', self.ids)
```
lmignon:
```suggestion
            ('partner_id', 'in', self.ids), ('typology', '=', 'wishlist')
```
@simahawk The domain to compute the count is not the same as the one given to action IOW, if a product set exists for a given partner the smart button will display '1 wishlist but once we'll click on the action, the list will be empty since the action's domain include a filter on the typology.

lmignon (changes requested):
@simahawk Great new feature. A few comments...

No related to this PR but IMO, an post_init_hook should be defined into sale_product_set_variant, to fill the `product_tmpl_id` and `product_variant_ids`from product_id` when the addon is installed and some product sets already exists. This will avoid this kind of screen 
image

sale_product_set/wizard/product_set_add.py
simahawk: good idea!
I removed the onchange but I left it for the wizard action because you could call it w/out noticing you are passing wrong data.

simahawk:
@lmignon shall be good now. Regarding trashing _variant module I opened this https://github.com/OCA/sale-workflow/issues/996 since it's out of scope here.

simahawk:
@lmignon I've just realized I missed an important bit: set lines work w/ product.product but we deal w/ invader variants. I'd say:

* add to product.set a m2o to invader backend
* add to set line a m2o to invader variant
* compute the product from the invader variant and viceversa

simahawk:
@lmignon can you update your review?

sale_product_set/wizard/product_set_add.py
lmignon: @simahawk No `@constrains("partner_id")`

sale_product_set/wizard/product_set_add.py
```diff
@@ -10,37 +10,77 @@ class ProductSetAdd(models.TransientModel):
     _rec_name = 'product_set_id'
     _description = "Wizard model to add product set into a quotation"
 
+    order_id = fields.Many2one(
+        'sale.order', 'Sale Order', required=True,
+        default=lambda self: self.env.context.get('active_id')
```
rousseldenis: You can set ondelete cascade too here.

rousseldenis (approved):
Code review. Minor comment

lmignon (approved):
Thank you for this useful improvement!
(Code + functional)

simahawk:
@sbidoul question regarding merging: if I use `ocabot merge minor` is going to increase the version of the new module as well. Hence, I think is better to update versions manually and use `ocabot merge` only. Correct?

rousseldenis:
> @sbidoul question regarding merging: if I use `ocabot merge minor` is going to increase the version of the new module as well. Hence, I think is better to update versions manually and use `ocabot merge` only. Correct?

Yes. You need to update sale product set version in this before using bot merge.

That's why too it's more correct to make separate PR's by module (even in real life it is not always possible)

PR #997, opened 2019-11-20
[MIG] sale_order_product_recommendation: Migration to 12.0

chienandalu:
Standard migrations

cc @Tecnativa TT20130


sale_order_product_recommendation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Sale Order Product Recommendation",
     "summary": "Recommend products to sell to customer based on history",
-    "version": "11.0.1.0.1",
+    "version": "12.0.1.0.1",
```
pedrobaeza: 12.0.1.0.0

pedrobaeza (changes requested):
Tests are failing

yajo (approved):

pedrobaeza:
Travis still red

chienandalu:
Looks like there's some interaction with another module of the report, my bet goes for sale_automatic_workflow changing delivered_qty behavior.

pedrobaeza:
> Looks like there's some interaction with another module of the report, my bet goes for sale_automatic_workflow changing delivered_qty behavior.

I don't think so. I would say the test is not well programmed, trying to use `[2]` index while there aren't that product lines in the test data. Why don't you filter that lines by product instead for doing the test more resilient?

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,198 @@
...
+        required=True,
+        help="The less, the faster they will be found.",
+    )
+    last_compute = fields.Char()
+
+    @api.model
+    def _default_order_id(self):
+        return self.env.context.get("active_id", False)
+
+    @api.multi
```
sergio-teruel: Remove multi

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,198 @@
...
+        """Generate lines according to context sale order."""
+        last_compute = '{}-{}-{}'.format(
+            self.id, self.months, self.line_amount)
+        # Avoid execute onchange as times as fields in api.onchange
+        # ORM must control this?
+        if self.last_compute == last_compute:
+            return
+        self.last_compute = last_compute
+        start = datetime.now() - timedelta(days=self.months * 30)
+        start = fields.Datetime.to_string(start)
```
sergio-teruel: In V12.0 do you need convert to string??

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,198 @@
...
+            # Use a new in-memory line otherwise
+            so_line = so_lines.new({
+                "order_id": self.order_id.id,
+                "product_id": wiz_line.product_id.id,
+                "sequence": sequence,
+            })
+            so_line.product_id_change()
+            so_line.product_uom_qty = wiz_line.units_included
+            so_line.product_uom_change()
+            so_lines |= so_line
```
sergio-teruel: For performance is better store createed line id and finally do a browse with all ids...

sergio-teruel (changes requested):

sergio-teruel:
Fix travis...

sale_order_product_recommendation/wizards/sale_order_recommendation.py
chienandalu: Even with on memory records? Changed anyway

chienandalu:
> I would say the test is not well programmed, trying to use `[2]` index while there aren't that product lines in the test data. Why don't you filter that lines by product instead for doing the test more resilient?

I'd say the problem comes from computing the delivered qtys. Anyway, I changed the tests as to avoid sorting problems like that

chienandalu:
Green at last. It was indeed a problem with the computation of delivered qtys in the test. Setting it wasn't enough and finally I had to force the product to be services so the stock methods didn't interfere...

pedrobaeza (approved):

PR #1003, opened 2019-11-26
[12.0] sale_automatic_workflow_split

i-vyshnevska:
this addon allows avoiding performance issues by splitting workflow cron on smaller, it works in tune with  cron from sale_automatic_workflow, no need to disable it

### DIFF
--- a/sale_automatic_workflow_split/__init__.py
+++ b/sale_automatic_workflow_split/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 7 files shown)


sebalix (approved):
LG, error on Travis is:
```
************* Module sale_automatic_workflow_split
sale_automatic_workflow_split/__init__.py:1: [C7902(missing-readme), ] Missing ./README.rst file. Template here: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
```
Not blocking IMO.

guewen (changes requested):
I don't think this should land in the OCA, it was a quick hack that I did years ago but IMO the proper solution that should land in the OCA for this purpose is to use `queue_job` to run each individual action.

PR #1005, opened 2019-11-28
[MIG] sale_order_product_recommendation_secondary_unit: Migration to 12.0

chienandalu:
cc @Tecnativa TT20511

PR #1006, opened 2019-11-29
[IMP] sale_order_product_recommendation: slipt tests

chienandalu:
- Improve test declaration inherability

For cases like https://github.com/OCA/sale-workflow/pull/1005

cc @Tecnativa TT20511

### DIFF
--- a/sale_order_product_recommendation/tests/test_recommendation.py
+++ b/sale_order_product_recommendation/tests/test_recommendation.py
@@ -1,82 +1,10 @@
-# Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# Copyright 2017 Tecnativa - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from odoo.tests.common import SavepointCase
+from .test_recommendation_common import RecommendationCase
 from odoo.exceptions import UserError
 
 
-class RecommendationCase(SavepointCase):
-    @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
-        cls.partner = cls.env['res.partner'].create({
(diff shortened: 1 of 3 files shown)


sale_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -1,75 +1,13 @@
...
-
-from odoo.tests.common import SavepointCase
+from .test_recommendation_common import RecommendationCase
 from odoo.exceptions import UserError
 
 
-class RecommendationCase(SavepointCase):
+class RecommendationCaseTests(RecommendationCase):
     @classmethod
     def setUpClass(cls):
```
pedrobaeza: No need of keeping this method

sale_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -1,75 +1,13 @@
...
-                    "qty_delivered": 50,
-                }),
-            ],
-        })
-        # Create a new sale order for the same customer
-        cls.new_so = cls.env["sale.order"].create({
-            "partner_id": cls.partner.id,
-        })
 
     def wizard(self):
```
pedrobaeza: Isn't interesting to move this also to base?

sale_order_product_recommendation/tests/test_recommendation.py
chienandalu: ok, thanks :)

sale_order_product_recommendation/tests/test_recommendation.py
chienandalu: Yes, I noticed it the moment I started writing the extension :)

pedrobaeza (approved):
Fast-tracking this as it doesn't involve any change of logic.

PR #1007, opened 2019-12-06
[10.0][FIX] sale_promotion_rule: Apply coupon to lines if percentage

rousseldenis:
If the coupon rule apply a discount in fixed price, the rule should
not be applied on each sale order line

### DIFF
--- a/sale_promotion_rule/tests/test_promotion.py
+++ b/sale_promotion_rule/tests/test_promotion.py
@@ -407,3 +407,30 @@ class PromotionCase(TransactionCase, AbstractCommonPromotionCase):
         for line in self.sale.order_line:
             self.check_discount_rule_set(line, promo_copy)
         return
+
+    def test_promotion_fixed_plus_discount(self):
+        """
+        Apply a manual discount on a line and apply a coupon code with
+        fixed amount and combine strategy. Both should apply.
+        """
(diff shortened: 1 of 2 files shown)


rousseldenis:
@Cedric-Pigeon @lmignon

sale_promotion_rule/tests/test_promotion.py
```diff
@@ -407,3 +407,29 @@ def test_multi_promotion_rules(self):
...
+        self.add_coupon_code(FIXED_AMOUNT_CODE)
+        self.sale.apply_promotions()
+        self.assertEquals(
+            20.0,
+            so_line.discount,
+        )
+        self.assertEquals(
+            690.0,
+            self.sale.amount_total
+        )
```
Cedric-Pigeon, line 435: i would add an assert to check coupon_promotion_rule_id is null

Cedric-Pigeon (changes requested):

lmignon (approved):
LGTM. Thank you  for the fix @rousseldenis

PierrickBrun (approved):

sale_promotion_rule/tests/test_promotion.py
rousseldenis: @Cedric-Pigeon done

Cedric-Pigeon (approved):

PR #1009, opened 2019-12-06
[11.0] [FIX] sale_order_type: Missing warnings in onchange method

nicomacr:
In the original onchage method by odoo, this return an dict with values(warnings and domains), but when inherit this method in this module they not return this dict. Because of this the alert setting in the partner was not working.


jcadhoc (approved):

rousseldenis (approved):
Code review. Of course !

rousseldenis:
Fast-tracking this one.

/ocabot merge patch

pedrobaeza (approved):

pedrobaeza:
Sorry, we have both launched the command. I have let yours.

PR #1011, opened 2019-12-09
[12.0] [FIX] sale_order_type: Missing warnings on inv onchange method

nicomacr:
In the original onchange method by odoo, this return an dict with values(warnings and domians), but when inherit this method in this module they not return this dict. Because of this the alert setting in the partner was no working.


rousseldenis (approved):
Code review.

rousseldenis:
Fast-tracking this

/ocabot merge patch

PR #1013, opened 2019-12-10
[13.0][MIG] sale automatic workflow

tfossoul:
Ready ;)


rousseldenis:
@tfossoul Thanks for your PR. Nice to see you back!

sale_automatic_workflow/models/sale_workflow_process.py
```diff
@@ -83,7 +83,6 @@ def _default_filter(self, xmlid):
     )
     order_filter_id = fields.Many2one(
         "ir.filters",
-        string="Order Filter",
```
rousseldenis: Oops, this one is needed (The string will be 'Order Filter Id')

sale_automatic_workflow/models/sale_order.py
```diff
@@ -72,3 +71,9 @@ def action_invoice_create(self, grouped=False, final=False):
                 if line.qty_delivered_method == "manual" and not line.qty_delivered:
                     line.write({"qty_delivered": line.product_uom_qty})
         return super().action_invoice_create(grouped=grouped, final=final)
+
+    def write(self, vals):
+        workflow = self.workflow_process_id
```
rousseldenis: You need to make it multi compatible as if recordset is > 1 it will crash

rousseldenis (changes requested):
Code review. Two changes needed

sale_automatic_workflow/models/sale_workflow_process.py
tfossoul: It is managed by odoo: fields.py line 473
self.string = (
                name[:-4] if name.endswith('_ids') else
                name[:-3] if name.endswith('_id') else name
            ).replace('_', ' ').title()

sale_automatic_workflow/README.rst
```diff
@@ -85,6 +85,7 @@ Contributors
 * Sodexis
 * Dave Lasley <dave@laslabs.com>
 * Akim Juillerat <akim.juillerat@camptocamp.com>
+* Thomas Fossoul <thomas@niboo.com>
```
rousseldenis, line 88: @tfossoul As there are readme fragments (in readme directory) add it there as it will be overwritten after merge.

rousseldenis:
After solving my last comment, could you squash your commits into only one ?

tfossoul:
@rousseldenis Done

rousseldenis (approved):
Code review

ThomasBinsfeld (approved):

sale_automatic_workflow/models/sale_order.py
```diff
@@ -0,0 +1,86 @@
...
+        if workflow.picking_policy:
+            self.picking_policy = workflow.picking_policy
+        if workflow.team_id:
+            self.team_id = workflow.team_id.id
+        if workflow.warning:
+            warning = {"title": _("Workflow Warning"), "message": workflow.warning}
+            return {"warning": warning}
+
+    def action_invoice_create(self, grouped=False, final=False):
```
simahawk, line 66: @tfossoul @rousseldenis we have an issue here: this method has been renamed https://github.com/odoo/odoo/commit/f808f36f3843875f0b87ab191277970564201429
I guess there's no test coverage for it...

rousseldenis: Weird, normally there is one :

https://github.com/OCA/sale-workflow/pull/1013/files#diff-04e90d09c8aa55dcca21d82befb795eeR90

rousseldenis: #1042

kittiu:
@tfossoul @simahawk @rousseldenis @ThomasBinsfeld 

Understand this module is merged quite sometime. I just have some question. As I saw in v12, the so code is with this code,
```
    def _do_validate_sale_order(self, sale):
        """Validate a sales order"""
        with savepoint(self.env.cr), force_company(self.env, sale.company_id):
            sale.action_confirm()

    @api.model
    def _validate_sale_orders(self, order_filter):
        sale_obj = self.env['sale.order']
        sales = sale_obj.search(order_filter)
        _logger.debug('Sale Orders to validate: %s', sales.ids)
        for sale in sales:
            self._do_validate_sale_order(sale)
```
But in v13, it is changed to this,
```
    @api.model
    def _validate_sale_orders(self, order_filter):
        sale_obj = self.env["sale.order"]
        sales = sale_obj.search(order_filter)
        _logger.debug("Sale Orders to validate: %s", sales.ids)
        for sale in sales:
            with savepoint(self.env.cr), force_company(self.env, sale.company_id):
                sale.action_confirm()
```

Please note that the method `_do_validate_sale_order` is missing. And this is important to be called in `sale_automatic_workflow_job`

Currently, as it can only create job at `_validate_sale_orders(order_filter)`. So, 1 job will potentially calling big number of of sale orders at the same time instead of one sale order and cause system halt. Which I think not the purpose of having job queue.

May be there are reasons that you make this changes?

Thanks you!

PS: should we bring `_do_validate_sale_order` back? 
https://github.com/OCA/sale-workflow/pull/1252

PR #1014, opened 2019-12-12
[12.0][ADD] sale_contact_type

Tardo:
This module adds a new contact address type called "Ordering address" and modify the "Customer" domain in Sale Orders to show contacts that are "Ordering address" and "master contacts" (contacts without parent) without childs or without childs with "ordering address" type.

Cc @tecnativa TT18576


sale_contact_type/views/sale_order.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr='//field[@name="partner_id"]' position='attributes'>
+                <attribute name="domain">[('customer', '=', True), '|', ('type', '=', 'ordering'), '&amp;', ('parent_id', '=', False), '|', ('child_ids', '=', False), ('has_ordering_contact_child', '!=', True)]</attribute>
```
ernestotejeda: Maybe it can be reduced to:
```suggestion
                <attribute name="domain">[('customer', '=', True), '|', ('type', '=', 'ordering'), '&amp;', ('parent_id', '=', False), ('has_ordering_contact_child', '!=', True)]</attribute>
```

ernestotejeda (commented):
Functional and code review OK, but check out my comment bellow.

sale_contact_type/README.rst
```diff
@@ -0,0 +1,80 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/sale-workflow-12-0/sale-workflow-12-0-sale_contact_type
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/167/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Adds a new type for contacts called 'Ordering address'.
```
pedrobaeza: It must 'Ordering contact' (not address)

sale_contact_type/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    type = fields.Selection(selection_add=[('ordering', 'Ordering address')])
```
pedrobaeza:
```suggestion
    type = fields.Selection(selection_add=[('ordering', 'Ordering contact')])
```

sale_contact_type/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+Create/modfy a contact and set the address type to 'Ordering address'
```
pedrobaeza:
```suggestion
Create/modify a children contact and set the address type to 'Ordering contact'
```

sale_contact_type/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
...
+Create a SO and select it in 'Customer' field.
```
pedrobaeza: Don't compact `SO`. This is user documentation. And specify full path to go for creating that sales order.

sale_contact_type/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+    has_ordering_contact_child = fields.Boolean(
+        compute='_has_ordering_contact_child',
```
pedrobaeza:
```suggestion
        compute='_compute_has_ordering_contact_child',
```

sale_contact_type/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+        help='Technical field for use in views', store=True)
+
+    @api.depends('child_ids.type')
+    def _has_ordering_contact_child(self):
+        for record in self:
+            record.has_ordering_contact_child = any(record.child_ids.filtered(
```
pedrobaeza: Can you optimize this with `read_group`? I think you can group by `parent_id` and by `type`, and assume all yes unless there's a group for `ordering` (and has parent).

sale_contact_type/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Alexandre Díaz <alexandre.diaz@tecnativa.com>
```
pedrobaeza: Put me as designer and extend reviewer ;)

And put also with Tecnativa URL and 2 levels.

sale_contact_type/views/sale_order.xml
pedrobaeza: Yes, I think so

pedrobaeza (changes requested):

sale_contact_type/models/res_partner.py
rousseldenis: Moreover, >= v13, you need to initialize all records with initial value as they are with current value before the compute function.

sale_contact_type/models/res_partner.py
```diff
@@ -0,0 +1,26 @@
...
+        help='Technical field for use in views', store=True)
+
+    @api.depends('child_ids.type')
+    def _compute_has_ordering_contact_child(self):
+        datas = self.read_group([
+            ('parent_id', 'in', self.ids),
+            ('type', '=', 'ordering')
+        ], ['parent_id'], groupby=['parent_id'])
+        parent_ids = {data['parent_id'][0] for data in datas}
+        with_ordering_c_recs = self & self.browse(parent_ids,
```
pedrobaeza: Why doing `self & ...` and not directly `self.browse(...`?

sale_contact_type/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Alexadre Díaz <alexandre.diaz@tecnativa.com>
+  * Pedro M. Baeza <pedro.baeza@tecnativa.com>
```
pedrobaeza: I prefer to not put emails at all

sale_contact_type/models/res_partner.py
Tardo: orf! yeah.. isn't really necessary.

Tardo:
Changes done

carlosdauden (approved):

pedrobaeza (approved):

PR #1016, opened 2019-12-13
[13.0][ADD] sale_automatic_workflow_payment

rousseldenis:
Depends on :

* [x] #1013 

This is a change from v12 sale_automatic_workflow_payment_mode module as :

- It does not depend anymore from account_payment_mode module and the payment mode itself. It uses the standard 'acquirer' from Odoo standard payment module.
- It does not create account.payment entries anymore as it relies on Odoo standard flow with transactions.


rousseldenis:
@lmignon @sbidoul @guewen @Cedric-Pigeon

Cedric-Pigeon (approved):
Looks great and much more simple than previously
Code review only.

sale_automatic_workflow_payment/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This modules allows to define an automatic workflow to a Sale Order when a
+payment transaction is created for it (to manage automatic invoice creation or
+...).
+
+Then, when the transaction is done, let the Odoo standard process flow
+to create corresponding payments and reconciliation (through a cron).
```
lmignon: @rousseldenis Is this cron active by default? If not, we should mention that the cron must be activeted

rousseldenis: Yes! See: https://github.com/odoo/odoo/blob/13.0/addons/payment/data/payment_cron.xml#L8

lmignon (approved):
LGTM (Code review only)

sale_automatic_workflow_payment/readme/DESCRIPTION.rst
rousseldenis: I've added more info

ghost:
Hi we dont need anymore sale_automatic_workflow_payment_mode right?

rousseldenis:
> Hi we dont need anymore sale_automatic_workflow_payment_mode right?

Yes, that's the idea. The workflow will be stored on acquirer. Then, let the normal Odoo flow occur.

rousseldenis:
Rebased

rousseldenis:
@guewen

sopharoth (approved):

PR #1021, opened 2019-12-18
13.0 mig sale order revision

jeroenev:


jeroenev:
Tests fail because for some reason both the `domain="["|", ("active", "=", False), ("active", "=", True)]"` and the `context={"active_test": False},` don't seem to work
if I update the tests and manually include `.with_context(active_test=False).old_revision_ids` the tests would pass, but I don't think that's intended. The field also shows up empty in the views.
It seems to me like this is a bug in Odoo itself: https://github.com/odoo/odoo/issues/41939

EDIT: Fixed

jeroenev:
Issue with active_test should be fixed in https://github.com/odoo/odoo/pull/42784. So tests should pass now. :pray:

pedrobaeza:
Take into account that OCB doesn't propagate the changes until tonight.

jeroenev:
Ah oke, didn't think about that, thanks.

marylla (approved):
Hi there,
I made a functional review.
I could create a new revision from a cancelled sale order which got the same name with a number as suffix, and after cancelling this order, I could create a new revision which got the next suffix. 
Works as expected.
Thank you for your contribution. :)

angelmoya (approved):
LGTM

AaronHForgeFlow (approved):
:+1: Functional test LGTM

AdriaGForgeFlow (approved):
Code review :+1:

sale_order_revision/tests/test_sale_order_revision.py
```diff
@@ -43,35 +41,34 @@ def test_order_revision(self):
 
         # Check the previous revision of the sale order
         revision_1 = sale_order_1.current_revision_id
-        self.assertEqual(sale_order_1.state, 'cancel')
-        self.assertFalse(sale_order_1.active)
+        self.assertEqual(sale_order_1.state, "cancel")
+        # self.assertFalse(sale_order_1.active)
```
rousseldenis, line 47: Please remove this if unused

sale_order_revision/tests/test_sale_order_revision.py
```diff
@@ -43,35 +41,34 @@ def test_order_revision(self):
...
 
         # Create a new revision of the Sale Order
         self._revision_sale_order(revision_1)
         revision_2 = revision_1.current_revision_id
 
         # Check the previous revision of the sale order
-        self.assertEqual(revision_1.state, 'cancel')
-        self.assertFalse(revision_1.active)
+        self.assertEqual(revision_1.state, "cancel")
+        # self.assertFalse(revision_1.active)
```
rousseldenis, line 64: Same as above

rousseldenis (changes requested):
Please check the two tests lines. Could you reactivate them? If test fail, why?

danimv5 (approved):
Functional review! LGTM

marylla:
@jeroen7s Are you still working on this?

sale_order_revision/tests/test_sale_order_revision.py
tbaden: I tested this on my local machine. It's working with the commented line as well. No change in code coverage.
As there is no change for sale.order active I think this can be dropped.

sale_order_revision/tests/test_sale_order_revision.py
tbaden: I tested this on my local machine. It's working with the commented line as well. No change in code coverage.
In this case there is actually a change from line 52 revision_1.active = True to this line revision_1.active = False. So I think we should keep this.

tbaden:
@rousseldenis is it ok for me to take this PR over, make the two small adjustments and create a new PR for this one?
will keep the author in history

jeroenev:
> @rousseldenis is it ok for me to take this PR over, make the two small adjustments and create a new PR for this one?
> will keep the author in history

Sorry for the late reply, haven't worked on Odoo13 stuff for a while (stuck in Odoo11 land)
Feel free to create an updated pr to make some final improvements, as I currently don't have the time for it.

tbaden:
superseed [1354](https://github.com/OCA/sale-workflow/pull/1354) including the changes requested by @rousseldenis

PR #1022, opened 2019-12-18
[FIX] sale_invoice_group_method: Respect your parents

yajo:
Before this patch, unless a context like `sale_invoice_group_method=False` was passed, `super()` would never be called.

Now, it's called always, and only when `sale_invoice_group_method=True` context is passed, it overrides what it needs to.

@Tecnativa TT20969


pedrobaeza:
I have cherry-picked 0b6b2b42911ca982e9c7d378add579b77bc97ec4 from 12.0, that was the original patch, but not backported here.

PR #1032, opened 2019-12-31
[12.0][MIG] sale_invoice_policy

AaronHForgeFlow:
Standard Migration


rousseldenis (approved):
Code review

rousseldenis:
@AaronHForgeFlow I think there is an incompatibility with sale_milestone_profile_invoicing module

AaronHForgeFlow:
@rousseldenis I will check. Looking at the code it seems both can work together.

Edited: This module makes field invoice_policy computed so users will change a different field, 'default_invoice_policy'. There is no incompatibility from the user perspective. However, from the test perspective, all tests that changes the field invoice_policy to 'delivery' will fail, they need to change the field default_invoice_policy instead. But of course, we cannot do that because those modules does not depend on this one.

I think the best here is just to configure CI to not install both modules simultaneously.

rousseldenis:
@AaronHForgeFlow Do you plan to add few lines of test for missing coverage ?

AaronHForgeFlow:
@rousseldenis Yes.

AaronHForgeFlow:
still planning to improve tests, just a moment :grin:

AaronHForgeFlow:
I think now it is fine. Regards.

Cedric-Pigeon (approved):
Code review

PR #1039, opened 2020-01-07
FIX sale order revision sequence on date order

sergiocorato:


sale_order_revision/models/sale_order.py
```diff
@@ -87,7 +88,13 @@ def create(self, values):
         if 'unrevisioned_name' not in values:
             if values.get('name', '/') == '/':
                 seq = self.env['ir.sequence']
-                values['name'] = seq.next_by_code('sale.order') or '/'
+                date_order = values.get('date_order', False)
```
rousseldenis, line 91: @sergiocorato Is this feature has to take place in this module ?

sergiocorato: Hi @rousseldenis , this is not an improvement, but a bugfix: the sequence needs a date to work correctly, otherwise an order with date in 2019 created in 2020 will have incorrectly a sequence of 2020.

rousseldenis (approved):

sergiocorato:
@OCA/crm-sales-marketing-maintainers merge?

rousseldenis:
Maybe another review.

sale_order_revision/models/sale_order.py
```diff
@@ -87,7 +88,13 @@ def create(self, values):
...
+                if date_order:
+                    date_order = fields.Date.from_string(date_order).strftime(
```
simahawk, line 93: why do you need to convert back and forth?

PR #1040, opened 2020-01-08
[13.0][MIG] sale_order_lot_selection

Khalid-SerpentCS:
This Module Allows you to select on sale order the production lot that will be delivered
You can select a lot for every sale order line, and it will be proposed on the delivery.


ThomasBinsfeld (approved):
Is it still a work in progress?

sale_order_lot_selection/models/stock.py
```diff
@@ -0,0 +1,35 @@
...
+            package_id=package_id,
+            owner_id=owner_id,
+            strict=strict,
+        )
+
+    def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
+        vals = super(StockMove, self)._prepare_move_line_vals(
+            quantity=quantity, reserved_quant=reserved_quant
+        )
+        if reserved_quant:
```
ThomasBinsfeld: It should only apply for sales:

```suggestion
        if reserved_quant and self.sale_line_id:
```

ThomasBinsfeld (changes requested):

sale_order_lot_selection/models/sale.py
```diff
@@ -0,0 +1,79 @@
...
+                move.refresh()
+            if move.state != "assigned":
+                raise UserError(
+                    _("Can't reserve products for lot %s") % line.lot_id.name
+                )
+        return True
+
+    def action_confirm(self):
+        res = super(SaleOrder, self.with_context(sol_lot_id=True)).action_confirm()
+        for line in self.order_line:
```
acsonefho, line 71: It should be great to add a context to bypass this check (same for `_check_move_state(...)`)
In some case (I have this case), I don't want to check the availability because my lot is in transit location.
I create a dedicated branch (based on your migration) to add these changes.
If you want I can create an PR to include that into your branch.
Can you open your MR please?
https://github.com/acsone/sale-workflow/tree/13.0-mig-imp-sale_order_lot_selection_fho

acsonefho (changes requested):
Just minor improvements but great!

bodedra:
@Khalid-SerpentCS are you available for address above comments?

sale_order_lot_selection/models/sale.py
dreispt: @acsonefho  Are you proposing your changes to this feature branch, or plan to propose to the OCA branch to merge once this is approved?
if you ask me, I prefer a separate OCA PR for each change.

sale_order_lot_selection/models/sale.py
```diff
@@ -42,7 +42,8 @@ def get_move_from_line(self, line):
         lot_count = 0
         for p in line.order_id.picking_ids:
             for m in p.move_lines:
-                move_line_id = m.move_line_ids.filtered(lambda line: line.lot_id)
+                move_line_id = \
+                    m.move_line_ids.filtered(lambda line: line.lot_id)
```
dreispt, line 46: These formatting changes are not needed.
FYI we now tolerate up to 88 character per line, the default for black.

dreispt (commented):
@Khalid-SerpentCS Can we have the reformatting in a separate commit? Otherwise it is very hard to review the real changes.
I suggest to:

1. Copy away your current module files
2. Reset the branch to the last commit before your migration changes
3. Apply reformatting and commit
4. Copy back you saved files and commit. This will have only your actual code changes.

Thank you!

dreispt:
Also, you have a lint error: `sale_order_lot_selection/tests/test_sale_order_lot_selection.py:4: [W0622(redefined-builtin), ] Redefining built-in 'Warning'`

Change the `Warning` exception to a `UserError`

sale_order_lot_selection/models/sale.py
acsonefho: @dreispt As you prefer. But my change is a little improvement, not a feature.

bodedra:
Ping @Chandresh-SerpentCS

rousseldenis:
@Khalid-SerpentCS Can we close this in favor of that one https://github.com/OCA/sale-workflow/pull/1115 ?

Chandresh-SerpentCS:
> @Khalid-SerpentCS Can we close this in favor of that one #1115 ?

Ok, We can close.

Khalid-SerpentCS:
> @Khalid-SerpentCS Can we close this in favor of that one #1115 ?

OK, We can close

sale_order_lot_selection/models/sale.py
ns-landoo: Hi,
I would like to share with us the incidence that occurs to me when creating an order with the 'lot_id' was empty ... and when odoo creating the 'delivery' it does not assign 'lot_id' but the quantity is correct, I think that by default odoo should choose the oldest. Is it right?
On the other hand, if the 'lot_id' is defined from the order, the behavior is correct. 
Could you help me?

Thank you in advance,

PR #1044, opened 2020-01-09
[FIX] 12.0 sale_invoice_group_method: Get correct invoice_group_method_id

Roodin:
Get the field from commercial partner if it is not set in partner


pedrobaeza:
Take a look to https://github.com/OCA/account-invoicing/tree/12.0/sale_order_invoicing_grouping_criteria that is less invasive.

sale_invoice_group_method/model/partner.py
```diff
@@ -11,3 +11,9 @@ class ResPartner(models.Model):
         string='Default Invoice Group Method',
         comodel_name='sale.invoice.group.method'
     )
+
+    def _get_invoice_group_method_id(self):
+        if self.invoice_group_method_id:
+            return self.invoice_group_method_id
+        else:
```
nikul-serpentcs, line 18:
```suggestion
        return self.commercial_partner_id.invoice_group_method_id
```

sale_invoice_group_method/model/partner.py
```diff
@@ -11,3 +11,9 @@ class ResPartner(models.Model):
...
+            return self.commercial_partner_id.invoice_group_method_id
```
nikul-serpentcs, line 19: @Roodin no need to else part

pedrobaeza: Both are needed, but a compact way to put it is:
```python
return self.invoice_group_method_id or self.commercial_partner_id.invoice_group_method_id
```

nikul-serpentcs: @pedrobaeza  I mean method should be
```
def _get_invoice_group_method_id(self):
    if self.invoice_group_method_id:
        return self.invoice_group_method_id
    return self.commercial_partner_id.invoice_group_method_id
```

pedrobaeza: And it's less good than my proposal as you can see.

marcelsavegnago:
Is there something strange about this PR or am I wrong?

image

PR #1065, opened 2020-02-10
[13.0][MIG] sale_order_product_recommendation: Migration to 13.0

Tardo:
cc @tecnativa TT21669


Tardo:
Idk why runbot is red :/

Tardo:
Ok, all green! ready to review

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,194 @@
...
+            so_line_ids.append(so_line.id)
+        self.order_id.order_line += so_line_obj.browse(so_line_ids)
+
+
+class SaleOrderRecommendationLine(models.TransientModel):
+    _name = "sale.order.recommendation.line"
+    _description = "Recommended product for current sale order"
+    _order = "id"
+
+    currency_id = fields.Many2one(related="product_id.currency_id", readonly=True)
```
sergio-teruel: Remove readonly=True, by default is True

sergio-teruel (approved):
Only a little comment

Tardo:
Done!

Tardo:
This PR need reviews

rousseldenis (approved):
Code review

rousseldenis:
@pedrobaeza Following the conversation on contributors mailing list, don't use nobump as it won't generate wheels to pypi. Use patch instead (even if it will generate a 13.0.1.0.1 version)

pedrobaeza:
Sorry, but I don't like that. And any way you have to deploy with --pre for pip for being sure, so it doesn't add anything.

rousseldenis:
> Sorry, but I don't like that. And any way you have to deploy with --pre for pip for being sure, so it doesn't add anything.

What don't you like ?

So when using nombump, the module on pypi won't be marked as release one :

image

And the --pre option is usually used for development (even if for releases you should use the version directly).

pedrobaeza:
I don't like to have a recently migrated module with version 13.0.1.0.1 and people wondering where it's the previous version. And you know that I don't think pypi is the ideal way of distributing this, but let's not starting again the same debate.

PR #1068, opened 2020-02-17
[11.0][ADD] sale_stock_delivery_address

LoisRForgeFlow:
This module allows to specify a different delivery address per sales order line, splitting deliveries if needed. Also you can specify different stock locations associated to those addresses, then the sales order procurements will be run from this location.

@ForgeFlow


sale_stock_delivery_address/models/sale_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    dest_address_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Destination Address",
+        help="If set this address will be the delivery address instead of the "
+             "one specified in the Sales Order header.",
+        domain="['|', ('id', '=', order_partner_id), "
```
rousseldenis, line 15: Isn't this too restrictive as it does not allow drop shipping?

rousseldenis (approved):
Code review. Just one question

StefanRijnhart (approved):

sale_stock_delivery_address/models/sale_order_line.py
LoisRForgeFlow: mmm it is a fair point, but I guess I prefer not to cover that remote case and have a more usable and less-error-prone UX.

PR #1071, opened 2020-02-21
12.0-FIX-installation-product_form_sale_link

jado95:
Before This PR, if i try to install the module, i have this error:
Element <button name="action_view_sales"> cannot be located in parent view.
I have the error because the Odoo addon sale_stock replace the button in view product_template_view_form_inherit_sale, that i replace the inherited view sale.product_template_form_view_sale_order_button with stock.product_template_form_view_procurement_button

### DIFF
--- a/product_form_sale_link/views/product_template.xml
+++ b/product_form_sale_link/views/product_template.xml
@@ -8,7 +8,7 @@
         <field name="name">product.template.form (in product_form_sale_link)</field>
         <field name="model">product.template</field>
         <field name="groups_id" eval="[(4, ref('sales_team.group_sale_salesman'))]"/>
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
Hi @jado95. thanks for sharing this patch. 

unfortunately, this module does'nt depends on stock module. I think that you should change replacing by xpath.

jado95:
> Hi @jado95. thanks for sharing this patch.
> 
> unfortunately, this module does'nt depends on stock module. I think that you should change replacing by xpath.

Hi @legalsylvain with the xpath the problem is the same. Have you another idea?

legalsylvain:
something like ``<xpath expr="//header" position="inside">``

PR #1073, opened 2020-02-24
[12.0][IMP] sale_mrp_link: One2many field on sale.order

DonatasV:
As mrp.production already has a Many2one field pointing to sale.order it makes sense to have a One2many on the "other side".

This adds multiple benefits:
- Easier to open linked production with the smart button action
- Simpler production count computing (also now can store the result)
- Dependent modules can use the One2many field for other purposes


gaikaz (approved):
LGTM
@NachoAlesLopez Could you take a quick look as well?

ps-tubtim (approved):
Code LGTM

PR #1076, opened 2020-02-25
[IMP] sale_order_product_recommendation: Add 'numeric_step' widget

Tardo:
cc @tecnativa


pedrobaeza:
There are information fields that are not needed to be put with the widget. Please add also in the readme a reference to the module as recommended for mobile layouts in the installation sections. Take also into account the secondary unit extension for adding the widget

Tardo:
Done

legalsylvain (approved):

pedrobaeza (approved):

PR #1079, opened 2020-02-27
[MIG][12.0] sale_generator: Migration to 12.0

kevinkhao:


sale_generator/models/sale_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        comodel_name="res.partner", string="Partner"
+    )
+    sale_ids = fields.One2many(
+        comodel_name="sale.order", inverse_name="generator_id", string="Sales"
+    )
+    tmpl_sale_id = fields.Many2one(
+        comodel_name="sale.order", string="Sale Template", required=True,
+    )
+    date_order = fields.Datetime(
+        string="Date", oldname="date", default=fields.Datetime.now()
```
bealdav: oldname no more used
```suggestion
        string="Date", default=fields.Datetime.now()
```

sale_generator/views/generator_view.xml
```diff
@@ -45,51 +45,53 @@
...
+                    <button class="oe_stat_button"
+                            name="%(act_sale_generator_2_sale_order)d"
+                            type="action"
+                            icon="fa-strikethrough"
+                            string="Sales"/>
+                </div>
                 <group name="header" colspan="2">
                     <field name="date_order"/>
                     <field name="tmpl_sale_id"
                            context="{'default_is_template': True, 'default_name': '', 'default_partner_id': %(sale_generator.res_partner_for_sale_template)d}"/>
```
bealdav: is_template field is removed
```suggestion
                           context="{'default_name': '', 'default_partner_id': %(sale_generator.res_partner_for_sale_template)d}"/>
```

sale_generator/models/partner.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def close_from_customer_wizard(self):
+        generator_id = self.env.context.get("from_generator_id")
+        action = self.env.ref("sale_generator.act_sale_generator_2_sale_order")
+        action["res_id"] = generator_id
+        return action
+
+    @api.model
+    def create(self, vals):
+        res = super(ResPartner, self).create(vals)
```
bealdav:
```suggestion
        res = super().create(vals)
```

sale_generator/models/sale_generator.py
```diff
@@ -0,0 +1,116 @@
...
+    def action_confirm(self):
+        for rec in self:
+            rec.write({"state": "done"})
+            # need single record action_confirm or else it crashes with
+            # sale_timesheet: sale_order.py:54 see self.company_id
+            for sale in rec.sale_ids:
+                sale.action_confirm()
+
+    def write(self, vals):
+        res = super(SaleGenerator, self).write(vals)
```
bealdav:
```suggestion
        res = super().write(vals)
```

sale_generator/models/sale_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+    @api.model
+    def create(self, vals):
+        if vals.get("name", "/") == "/":
+            vals["name"] = (
+                self.env["ir.sequence"].next_by_code("sale.order.generator")
+                or "/"
+            )
+        return super(SaleGenerator, self).create(vals)
```
bealdav:
```suggestion
        return super().create(vals)
```

sale_generator/tests/test_sale_generator.py
```diff
@@ -0,0 +1,37 @@
+# 2017 EBII Monsieurb <monsieurb@saaslys.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleGenerator(TransactionCase):
+    def setUp(self):
+        super(TestSaleGenerator, self).setUp()
```
bealdav:
```suggestion
        super().setUp()
```

sale_generator/models/sale_order.py
```diff
@@ -12,4 +11,3 @@ class SaleOrder(models.Model):
     generator_id = fields.Many2one(
         comodel_name="sale.generator", string="Generator"
     )
-    is_template = fields.Boolean()
```
bealdav: Not sure it should be removed. But probably we could replace this field by an `active` field set to True in this case.
Then template exist from generator but not from sales. It avoid to hack sale.order domain action

cc @sebastienbeau

sale_generator/models/sale_generator.py
```diff
@@ -18,10 +18,7 @@ class SaleGenerator(models.Model):
         comodel_name="sale.order", inverse_name="generator_id", string="Sales"
     )
     tmpl_sale_id = fields.Many2one(
-        comodel_name="sale.order",
-        string="Sale Template",
-        required=True,
-        domain=[("is_template", "=", True)],
```
bealdav: It seems the domain is important here

sale_generator/views/sale_view.xml
```diff
@@ -1,56 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<odoo>
-    <record id="view_sale_order_form" model="ir.ui.view">
```
bealdav: Not sure it should be removed because it's required to set tmpl_sale_id in generator

bealdav (changes requested):

sale_generator/models/sale_order.py
kevinkhao: Yes, not sure either, @sebastienbeau could you help us understand how the client uses this, how do we solve the problem of differentiating whether SOs are used as templates or not and how to find them ? Does the client want to know ? If yes, how should they go about it ?

In my opinion, if they only want to use the generator as a simple copy-paste, they shouldn't care about it... but because we keep the generators and their source SO, they might misuse it (by trying to generate more copies of an outdated SO) and create confusion.

If they want to really track generators and SOs used as templates for the generators, then we need a more elaborate solution and menu items with filters etc

sale_generator/models/sale_generator.py
kevinkhao: thanks David, I must've done some bad force push operations

sale_generator/models/sale_order.py
kevinkhao: The new version uses is_template field again as well as active field to help separate sale order templates from sale orders

sale_generator/models/sale_generator.py
kevinkhao: Yes, new version fixes this

sale_generator/models/sale_order.py
```diff
@@ -0,0 +1,50 @@
...
+    is_template = fields.Boolean(string="Is a generator template")
+    active = fields.Boolean("Active", default=True)
+
+    @api.model
+    def create(self, vals):
+        if self.env.context.get("tmpl_mode"):
+            vals["name"] = self.env["ir.sequence"].next_by_code(
+                "sale.order.tmpl"
+            ) or _("New template")
+            if not vals.get("partner_id"):
```
sebastienbeau: I do not thinks that it's a good idea to take a random partner.
Indeed this can lead to some issue 
- in multi company case (partner can be not share between company)
- impossibility to delete the partner because is linked to a template (and no way to see it)

I prefers to have a inactive dedicated partner for that (like in 8)

sebastienbeau (commented):
ir.rule have been delete please add it back

https://github.com/OCA/sale-workflow/pull/1079/commits/0c384f48d465eea9fa3f0f73094f4900eca6426f#diff-6c660b04e614563767531a8b72e31475

sale_generator/models/sale_order.py
```diff
@@ -0,0 +1,50 @@
...
+from lxml import etree
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    generator_id = fields.Many2one(
+        comodel_name="sale.generator", string="Generator"
+    )
+    is_template = fields.Boolean(string="Is a generator template")
```
sebastienbeau: please add copy=False

sale_generator/models/sale_order.py
```diff
@@ -0,0 +1,50 @@
...
+            ) or _("New template")
+            if not vals.get("partner_id"):
+                # in tmpl_mode we hide partner_id and don't care about its value
+                vals["partner_id"] = (
+                    self.env["res.partner"].search([], limit=1).id
+                )
+        return super().create(vals)
+
+    @api.model
+    def fields_view_get(
```
sebastienbeau: I do not understand why you did this.
Not after testing If I go to the sale template I need to fill a customer the field should be hidden

sale_generator/views/sale_order_view.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record model="ir.actions.act_window" id="action_sale_order_template">
+        <field name="name">Sale Order</field>
+        <field name="res_model">sale.order</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form</field>
+        <field name="context">{'default_active': False, 'default_is_template': True, 'tmpl_mode': True}</field>
+        <field name="domain">[('is_template', '=', True), ('active', '=', False)]</field>
+    </record>
+
+    <menuitem name="Sale Order Templates"
```
sebastienbeau, line 66: I think we should move the menu in sale and not in configuratiojn

sale_generator/models/sale_generator.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    def _prepare_copy_vals(self, partner):
+        self.ensure_one()
+        vals = {
+            "partner_id": partner.id,
+            "generator_id": self.id,
+            "warehouse_id": self.warehouse_id.id,
+            "company_id": self.warehouse_id.company_id.id,
+        }
```
sebastienbeau, line 51: is_template: False

sale_generator/models/sale_order.py
sebastienbeau: maybe better to reuse view https://github.com/OCA/sale-workflow/commit/0c384f48d465eea9fa3f0f73094f4900eca6426f#diff-ab867746dba3d99c7ab6259ae0861977

sale_generator/views/sale_order_view.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="view_sale_order_tmpl_form" model="ir.ui.view">
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale_stock.view_order_form_inherit_sale_stock"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="is_template" invisible="1"/>
+            </field>
```
sebastienbeau, line 20: Can you hide the field sale_order_template_id ?

kevinkhao: ok, but this is a field completely unrelated to this module (that field is declared in sale_management). Just to be sure, can you confirm ?

kevinkhao: confirmed and done, https://github.com/OCA/sale-workflow/pull/1079/commits/4d7a86854612182b8bffad909a4dee598b29f214

sebastienbeau:
LGTM (code review and testing)

sebastienbeau (approved):

bealdav (approved):
LGTM, thanks a lot

PR #1081, opened 2020-02-27
[IMP] sale_order_type: run onchanges on create

chienandalu:
When an order was created outside UI (e.g.: website sale) the order type
values weren't properly fetched. We run them on create when no type is
given asumming that in that case the invoker didn't take this behavior
into account.

Also, we get rid of the default value for the field and keep although we
default to it in case none is given by the partner.

cc @Tecnativa TT22539


pedrobaeza (approved):

rafaelbn (approved):

pedrobaeza:
/ocabot merge patch

Please forward port it to v12

PR #1091, opened 2020-03-13
[13.0][MIG] sale_procurement_group_by_line

JoanSForgeFlow:
Standard **sale_procurement_group_by_line** migration to v13.

cc ~ @ForgeFlow @LoisRForgeFlow


sale_procurement_group_by_line/model/sale.py
```diff
@@ -20,16 +20,14 @@ def _prepare_procurement_group_by_line(self, line):
...
     def _get_procurement_group_key(self):
         """ Return a key with priority to be used to regroup lines in multiple
         procurement groups
 
         """
         return 8, self.order_id.id
 
-    @api.multi
-    def _action_launch_stock_rule(self):
+    def _action_launch_stock_rule(self, previous_product_uom_qty=False):
```
LoisRForgeFlow, line 30: This method is overriding standard one, right? if so, can you add a comment stating that?

MiquelRForgeFlow: well, it's calling super, so it's a bit obvious :P

sale_procurement_group_by_line/model/sale.py
```diff
@@ -0,0 +1,122 @@
...
+                        self.env.company,
+                        values,
+                    )
+                )
+                self.env["procurement.group"].run(procurements)
+            except UserError as error:
+                errors.append(error.name)
+        if errors:
+            raise UserError("\n".join(errors))
+        return super(SaleOrderLine, self)._action_launch_stock_rule()
```
MiquelRForgeFlow:
```suggestion
        return super()._action_launch_stock_rule(previous_product_uom_qty=previous_product_uom_qty)
```

MiquelRForgeFlow (commented):
```diff
@@ -99,15 +97,20 @@ def _action_launch_stock_rule(self):
- 42            qty = line._get_qty_procurement()
+ 42            qty = line._get_qty_procurement(previous_product_uom_qty)

sale_procurement_group_by_line/model/sale.py
LoisRForgeFlow: I'm a little bit confused, because there is a super at the end but the method seems to be a copy of the standard one with few changes, can you check what happens in supers, is the procurement executed twice?

MiquelRForgeFlow: This super is a tricky way to maintain inheritance. Overriding methods can arise problems, that's why its better to call super if possible.

MiquelRForgeFlow (approved):

JoanSForgeFlow:
@LoisRForgeFlow @MiquelRForgeFlow comments attended!

sale_procurement_group_by_line/model/sale.py
LoisRForgeFlow: @MiquelRForgeFlow Yes, I know that :laughing: My concern is that this call to super is execunting almost the same code just after. That's why I thought the method was overriding (meaning it has no call to super) but it is not, so the question is... should it be overriden? or can it be improved?

LoisRForgeFlow:
@JoanSForgeFlow Can you have a look at the discussion in my review and bring some light to it?

sale_procurement_group_by_line/model/sale.py
MiquelRForgeFlow: But I already said that it should not be overridden because is preferable to have `super`. This method is also called in `procurement_jit` (which have auto_install=True), and many modules depend from `sale_stock` but not from `procurement_jit`, so it may arise problems if we don't put the `super`. And it cannot be improved because the method is complex and there is no way do it easier. It's not something of this migration, but something we have in previous versions and something I already checked before. But of course, feel free to recheck.

JoanSForgeFlow:
Standard method `_action_launch_stock_rule` checks the moves of the line, if it's already done does nothing, because of this reason the procurement is not executed twice.

sale_procurement_group_by_line/model/sale.py
LoisRForgeFlow: @MiquelRForgeFlow you are not understanding my concern... Maybe I didn't explain myself correctly. Anyway, the issue has been discussed with @JoanSForgeFlow and it is clarified. Thanks for having a look!

LoisRForgeFlow (approved):
Thanks! LGTM now :+1:

AaronHForgeFlow:
> Standard method `_action_launch_stock_rule` checks the moves of the line, if it's already done does nothing, because of this reason the procurement is not executed twice.

But in this case, super will be always called and the custom code will never take effect, or  Am I missing something?

JoanSForgeFlow:
@AaronHForgeFlow custom code takes effect running the procurement, then calls super and the standard one checks if the procurement it's already done for running the procurement or not

AaronHForgeFlow (approved):
Code review :+1:

AaronHForgeFlow:
> @AaronHForgeFlow custom code takes effect running the procurement, then calls super and the standard one checks if the procurement it's already done for running the procurement or not

Yes, it's clear, thanks for the clarification.

PR #1097, opened 2020-03-18
[12.0][ADD] - Add new module Sale Order Transmit Method

sbejaoui:


sbejaoui:
cc/ @gva-acsone

rousseldenis (approved):
Code review. Seems great.

Wouldn't it be useful to store that info on partner level for default ?

sbejaoui:
> Code review. Seems great.
> 
> Wouldn't it be useful to store that info on partner level for default ?

thank you @rousseldenis for your review.

the `account_invoice_transmit_method` module manage this already

sale_order_transmit_method/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @api.multi
+    def _finalize_invoices(self, invoices, references):
+        res = super()._finalize_invoices(invoices, references)
+        for invoice in invoices.values():
+            transmit_methods = invoice.invoice_line_ids.mapped(
+                'sale_line_ids.order_id.transmit_method_id'
+            )
+            if len(set(transmit_methods)) == 1:
+                invoice.transmit_method_id = transmit_methods[0]
```
hparfr: why using a set and not `if len(transmit_methods)` directly ?

hparfr (approved):
LGTM

sale_order_transmit_method/models/sale_order.py
sbejaoui: to avoid duplicate

sbejaoui:
@OCA/core-maintainers , This PR is ready to merge

sale_order_transmit_method/models/sale_order.py
simahawk: you can use `.ids` which never contains duplicates. But in this case, using mapped ensure you get only one per value ;)

sale_order_transmit_method/tests/test_sale_order.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrder(TransactionCase):
```
simahawk: pls use SavepointCase and move whole setup records to setUpClass

simahawk (changes requested):

sale_order_transmit_method/tests/test_sale_order.py
sbejaoui: Is there any reason for that?

simahawk: very simple: `setUp` runs EVERY time for EVERY test. `setUpClass` only once.
Multiply this for all the tests we run in CI and you get tons of resources and time wasted because of this. Another recommendation is to disable tracking suite wise from setUpClass when not needed:
```py 
```
Sooner or later I'll add these tips to official OCA guidelines :sweat_smile: .

sale_order_transmit_method/models/sale_order.py
sbejaoui: Indeed, mapped return a list of values for each record even with duplicate, except when the result is a recordset. (no id duplication in a record set) 

I have a reflex to use `set` to avoid this.



As you say, mapped is enough in this case, I will make the change.

sbejaoui:
@simahawk , can you update you review please?

simahawk (approved):
@sbejaoui LG thanks!

Would you mind squashing?

sbejaoui:
> @sbejaoui LG thanks!
> 
> Would you mind squashing?

will do

PR #1104, opened 2020-03-26
[13.0] Add module sale_cutoff_time_delivery

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/partner-contact/pull/876

This module allows to define Cutoff times for delivery orders.

If no commitment date is set, any sale order confirmed after the applicable
cutoff time will have its delivery order postponed by one day at the time of
the applicable cutoff, and any sale order confirmed before the applicable
cutoff time will have its delivery order postponed to the time of the applicable
cutoff.

Application of cutoff is defined by partner, with default value being the
cutoff time of the sale order's warehouse.

Cut-off must however be applied on the warehouse by marking the checkbox
"Apply cutoff" so that the computation happens.

### DIFF
--- a/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
+++ b/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
@@ -1,0 +1,31 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class CutoffTimeMixin(models.AbstractModel):
+
+    _name = "cutoff.time.mixin"
+
+    order_delivery_cutoff_hours = fields.Selection(
+        selection="_selection_order_delivery_cutoff_hours", required=True, default="12"
+    )
+    order_delivery_cutoff_minutes = fields.Selection(
+        selection="_selection_order_delivery_cutoff_minutes",
+        required=True,
+        default="00",
+    )
+
+    @api.model
+    def _selection_order_delivery_cutoff_hours(self):
+        return [("{:02}".format(i), "{:02}".format(i)) for i in range(0, 24)]
+
+    @api.model
+    def _selection_order_delivery_cutoff_minutes(self):
+        return [("{:02}".format(i), "{:02}".format(i)) for i in range(0, 60)]
+
+    def get_cutoff_time(self):
+        return {
+            "hour": self.order_delivery_cutoff_hours,
+            "minute": self.order_delivery_cutoff_minutes,
+        }
--- a/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
+++ b/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
@@ -5,7 +5,7 @@ from odoo import api, fields, models
 
 class CutoffTimeMixin(models.AbstractModel):
 
-    _name = "cutoff.time.mixin"
+    _name = "sale.cutoff.time.mixin"
 
     order_delivery_cutoff_hours = fields.Selection(
         selection="_selection_order_delivery_cutoff_hours", required=True, default="12"
--- a/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
+++ b/sale_cutoff_time_delivery/models/cutoff_time_mixin.py
@@ -4,7 +4,6 @@ import math
 from datetime import time
 
 from odoo import api, fields, models
-from odoo.addons.base.models.res_partner import _tz_get
 
 
 class TimeCutoffMixin(models.AbstractModel):
@@ -12,9 +11,6 @@ class TimeCutoffMixin(models.AbstractModel):
     _name = "time.cutoff.mixin"
 
     cutoff_time = fields.Float()
-    tz = fields.Selection(
-        _tz_get, string='Timezone'
-    )
 
     def get_cutoff_time(self):
         hour, minute = self._get_hour_min_from_value(self.cutoff_time)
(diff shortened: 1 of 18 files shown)


sale_cutoff_time_delivery/models/cutoff_time_mixin.py
```diff
@@ -0,0 +1,31 @@
...
+
+class CutoffTimeMixin(models.AbstractModel):
+
+    _name = "cutoff.time.mixin"
+
+    order_delivery_cutoff_hours = fields.Selection(
+        selection="_selection_order_delivery_cutoff_hours", required=True, default="12"
+    )
+    order_delivery_cutoff_minutes = fields.Selection(
+        selection="_selection_order_delivery_cutoff_minutes",
```
rousseldenis: Please change to lambda

sale_cutoff_time_delivery/models/cutoff_time_mixin.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class CutoffTimeMixin(models.AbstractModel):
+
+    _name = "cutoff.time.mixin"
```
rousseldenis: Prefer sale.cutoff.time.mixin as this is applicable to sale. Either, externalize in another module

sale_cutoff_time_delivery/models/cutoff_time_mixin.py
grindtildeath: AFAICS, this is not needed, as a string arg is called using using `getattr`: https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L2170

sale_cutoff_time_delivery/models/cutoff_time_mixin.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class CutoffTimeMixin(models.AbstractModel):
+
+    _name = "sale.cutoff.time.mixin"
+
+    order_delivery_cutoff_hours = fields.Selection(
```
grindtildeath: Probably better to harmonize usage with #1101 and use Float

sale_cutoff_time_delivery/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+* Consider timezones
```
grindtildeath: This must be fixed as well, although I still don't know how to do it :sweat_smile:

sale_cutoff_time_delivery/models/sale_order.py
```diff
@@ -0,0 +1,123 @@
...
+        cutoff_time = time(hour=cutoff.get('hour'), minute=cutoff.get('minute'))
+        if cutoff_tz:
+            utc_cutoff_time = tz_utils.tz_to_utc_time(
+                cutoff_tz,
+                cutoff_time,
+                base_date=expected_date
+            )
+        else:
+            utc_cutoff_time = cutoff_time
+        if expected_date.time() <= utc_cutoff_time:
```
jbaudoux, line 148: You should not compare times. Apply here the same implementation you did in _prepare_procurement_values.
1:00 < 3:00 in CET so should remains the same after conversion to UTC: 23:00 <> 1:00

jbaudoux (changes requested):

sale_cutoff_time_delivery/models/sale_order.py
grindtildeath: I'm not sure to understand why this wouldn't work since `utc_cutoff_time` is computed according to `expected_date`. Can you please explain a bit more?

LoisRForgeFlow: @jbaudoux ping

jgrandguillaume (approved):
Functional test ok so far

sale_cutoff_time_delivery/models/sale_order.py
jbaudoux: Let's define a cutoff at 6AM AWST (Australia time). You order at 10AM, so you are after the cutoff.
Now I convert to UTC: Cutoff=22PM UTC. Order time is 2AM. I am before the cutoff.
If my warehouse is in Australia as well as my customers, I don't care about the time in London.

jbaudoux:
@jgrandguillaume I don't approve yet. To discuss next week.

sale_cutoff_time_delivery/models/cutoff_time_mixin.py
```diff
@@ -0,0 +1,40 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import math
+from datetime import time
+
+from odoo import api, fields, models
+
+
+class TimeCutoffMixin(models.AbstractModel):
+
+    _name = "time.cutoff.mixin"
```
guewen, line 11:
```suggestion
    _name = "time.cutoff.mixin"
    _description = "Time Cut-off Mixin"
```

simahawk:
@jbaudoux @grindtildeath any update on this?

sale_cutoff_time_delivery/models/cutoff_time_mixin.py
guewen: I think this is what you wanted to link: https://github.com/odoo/odoo/blob/45afa3a20693ebed11ad2fa7ebd1b73772ca5b3d/odoo/fields.py#L2242-L2243 ;)

grindtildeath:
@simahawk Not in my hands anymore :shrug:

simahawk:
replaced by #1378

PR #1105, opened 2020-03-26
[12.0][ADD] sale_commitment_lead_time

chafique-delli:
This module allows you to know if the time to prepare a sales order has been respected, based on an order limit time defined on the company.
For example, in e-commerce, some companies agree to prepare a customer order the same day if it has been confirmed before a certain time.


rousseldenis:
@chafique-delli WIP ?

chafique-delli:
@rousseldenis , no, finished work. Thanks, for a review.

sale_check_preparation_time/models/res_company.py
```diff
@@ -0,0 +1,22 @@
...
+import pytz
+
+
+def _tz_get(self):
+    return [(tz, tz) for tz in sorted(
+        pytz.all_timezones, key=lambda tz: tz if not tz.startswith(
+            'Etc/') else '_')]
+
+
+class resCompany(models.Model):
```
sebastienbeau: ResCompany

sale_check_preparation_time/models/stock.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.multi
+    def action_done(self):
+        res = super(StockPicking, self).action_done()
+        for picking in filter(lambda x: x.picking_type_id.code == 'outgoing' and
+                              x.sale_id, self):
+            if picking.company_id.tz:
+                local_tz = pytz.timezone(picking.company_id.tz)
+            else:
+                local_tz = pytz.timezone('UTC')
```
sebastienbeau: Maybe it's better to add a flag "check_preparation_time" on company and then put in mandatory the field
- tz and order_limit_hour.
So here the field will be always fill?
And if the check_preparation_time is not tic then we just skip this step?

sale_check_preparation_time/models/stock.py
```diff
@@ -0,0 +1,30 @@
...
+            confirmation_date = local_tz.localize(
+                picking.sale_id.confirmation_date)
+            order_hour = confirmation_date.hour
+            order_limit_hour = int(picking.company_id.order_limit_hour)
+            order_date = fields.Date.from_string(
+                picking.sale_id.confirmation_date)
+            picking_date = fields.Date.from_string(picking.date_done)
+            delta = picking_date - order_date
+            if order_hour < order_limit_hour and delta.days <= 1:
+                picking.sale_id.write({'timely_delivery': True})
```
sebastienbeau: I think it will be great to add on the sale order a field 'first_delivery_datetime' and also fill this field.
It can be usefull for statistic/dashboard

sebastienbeau (changes requested):

sale_check_preparation_time/models/stock.py
chafique-delli: @sebastienbeau , There is already the field 'effective_date' which is used for that.

sale_check_preparation_time/models/stock.py
chafique-delli: @sebastienbeau , Indeed, if you wish to use the functionality, the field tz and order_limit_hour will be always fill.

sale_check_preparation_time/models/res_company.py
chafique-delli: @sebastienbeau , I'm correcting.

chafique-delli:
ping @sebastienbeau

rousseldenis:
IMHO, I would change the module name to sale_order_time_limit as it is more representative of the real use case. What do you think ?

sale_check_preparation_time/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models, api
+import pytz
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    timely_delivery = fields.Boolean(
```
hparfr: When I read it, I assumed timely_delivery means this order should respect a time constraint.
It seems to be the opposit. I suggest to rename it to somethink like "delivered_in_time"

hparfr:
Thanks @chafique-delli 
About naming of the module, what do you think of "sale_commitment_lead_time" ?

Because, this module is mainly about telling a customer if he confirms in time, he can have is order processed the same day.

Lead time seems to me less ambiguous than  "preparation".

@rousseldenis , what do you think of sale_commitment_lead_time ?

rousseldenis:
> @rousseldenis , what do you think of sale_commitment_lead_time ?

LGTM!

chafique-delli:
Thanks @hparfr for the proposal.
So I'm renaming the module to 'sale_commitment_lead_time'.

hparfr (approved):
LGTM

tonygalmiche (approved):
I tested on my computer locally
I checked the code
Everything seems OK for me
Thanks for this new addon

sebastienbeau (approved):

PR #1107, opened 2020-03-27
[13.0] Add module sale_cutoff_time_weekday_delivery

grindtildeath:
Depends on:
 - [ ] #1101 
 - [ ] #1104 

This module adds some unittests to ensure Method Resolution Order of
`sale.order.line._prepare_procurement_values` for modules
`sale_cutoff_time_delivery` and `sale_partner_delivery_window` is correct.


jbaudoux (changes requested):
We need to ensure that cutoff computation is applied before time window. If after cutoff, we add a day. Then, if not in the time window, we add additional days.
When I tested, it executed in the contrary order

grindtildeath:
@jbaudoux This test (https://github.com/OCA/sale-workflow/pull/1107/files#diff-4d050161f1bf44a1ee247ea0e214f592R194-R230) must ensure MRO is correct. Can you check if you still have issues after the refresh of submodules?

sale_partner_cutoff_delivery_window/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["sale_cutoff_time_delivery", "sale_partner_delivery_window"],
+    "auto-install": True,
```
grindtildeath:
```suggestion
    "auto_install": True,
```

jgrandguillaume (approved):
Functional test, work great

simahawk:
replaced by #1379

PR #1109, opened 2020-03-31
[13.0][MIG] sale_mrp_link

traviswaelbro:


CasVissers-360ERP (approved):
LGTM!

pedrobaeza:
The way this "catches" the link is very weak, through a search of a string in the source field. You should migrate and use instead this module: #525 

For me, `sale_mrp_link` should be deprecated.

alyngsoe:
Can anyone please tell os if we have a v13 module to link Saleorder and manufacturing orders ? I cannot figure out  the outcome of thread.

pedrobaeza (changes requested):
This one shouldn't be migrated, but https://github.com/OCA/manufacture/tree/12.0/mrp_sale_info

traviswaelbro:
I don't think `mrp_sale_info` is a replacement for `sale_mrp_link`. Yes, it links the Manufacturing Order and Work Order to a Sales Order, but it doesn't actually provide any way for the Sales Order to display what Manufacturing Orders are related. I will continue using `sale_mrp_link` for its provided features.

pedrobaeza:
If that's the case, it's as simply as add such smart-button in the other module. Here the problem is not the button, but the method to get the link between the 2 records.

rousseldenis:
> I don't think `mrp_sale_info` is a replacement for `sale_mrp_link`. Yes, it links the Manufacturing Order and Work Order to a Sales Order, but it doesn't actually provide any way for the Sales Order to display what Manufacturing Orders are related. I will continue using `sale_mrp_link` for its provided features.

I agree with Pedro on the fact the way to retrieve sale order on production is not reliable. You should use procurement group.

I tend to say both modules must merge using behaviour described above.

gdgellatly:
@rousseldenis @pedrobaeza I don't know if this approach helps, it was a fork of an OCA module where we didn't like the origin searching I think so did it with stock moves instead.

https://github.com/odoonz/odoonz-addons/tree/12.0/mrp_sale_fields

This is the specifc code for getting sale on mrp, it looks a bit dodgy now I look at it, but it has never failed me. You can use whatever code you think is reasonable there.

```
@api.depends("move_dest_ids")
    def _compute_sale_order(self):
        self._cr.execute(
            """SELECT sm.created_production_id, sol.order_id
        FROM stock_move sm
        LEFT JOIN sale_order_line sol ON sm.sale_line_id = sol.id
        WHERE sm.sale_line_id IS NOT NULL
          AND sm.created_production_id IS NOT NULL
          AND sm.created_production_id IN %s """,
            (tuple(self.ids),),
        )
        production_data = self._cr.fetchall()
        if production_data:
            mapped_data = dict(production_data)
            for production in self:
                production.sale_id = mapped_data.get(production.id, False)

    sale_id = fields.Many2one(
        comodel_name="sale.order",
        string="Sale Order",
        readonly=True,
        compute="_compute_sale_order",
        store=True,
    )

    partner_id = fields.Many2one(
        related="sale_id.partner_id", string="Customer", store=True
    )
```

actually I see my guys already migrated to v13 https://github.com/odoonz/odoonz-addons/blob/13.0/mrp_sale_fields/models/mrp_production.py

pedrobaeza:
Thanks for that, @gdgellatly, but the code in the other module is not failing, so no need to change it. It's just that is another module and it seems to be missing a couple of smart-buttons that can be incorporated there. I don't have any problem in having this module at OCA, as I will continue with the other, but I think this one should be "merged" in the other for avoiding a false sensation that this module works for every case.

gdgellatly:
Yeah, the other one never really worked for us, as we their is a bunch of sale order revision code which bypassed stock rule as often the MO already existed.

danimv5 (approved):
Functional review!! LGTM!!

gdgellatly (changes requested):
@traviswaelbro For much the same logic as other experienced reviewers. Fundamentally you want a seperate module for a couple of smart buttons functionality over another existing ported module. Its not your fault, honestly over time reviewers expand out, modules get included that shouldn't and it is only after some time we see the silliness of our prior efforts.

I myself have been both the instigator and receiver of such changes and they are healthy for us to grow our quality in this area by having twice as many eyes on 1 module as we would have split.

It is far more advantageous if we can merge those buttons into the other module, which is already ported and maintains a more robust approach to linking the objects, rather than maintain 2 seperate modules. I really don't see anything blocking that occuring.

brendapaniagua:
Functional review. LGTM!!

PR #858, opened 2019-05-29
[12.0][MIG] sale_packaging

rousseldenis:
Depends on :

* [x] https://github.com/OCA/product-attribute/pull/486
* [ ] Complete coverage
* [x] Remove entry in oca_dependencies.txt

PR #1114, opened 2020-04-07
[13.0] Create module sale_order_line_packaging_qty

grindtildeath:
This module adds a "Package quantity" field on sale order line in order to
define a quantity according to the selected "Package".

Works better with:
 - [x] https://github.com/odoo/odoo/pull/49150

### DIFF
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -1,0 +1,85 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class SaleOrderLine(models.Model):
+
+    _inherit = "sale.order.line"
+
+    product_packaging_qty = fields.Float(
+        string="Package quantity",
+        compute="_compute_product_packaging_qty",
+        inverse="_inverse_product_packaging_qty",
+    )
+
+    @api.depends("product_uom_qty", "product_uom", "product_packaging", "product_packaging.qty")
+    def _compute_product_packaging_qty(self):
+        for sol in self:
+            if not sol.product_packaging or sol.product_uom_qty == 0 or sol.product_packaging.qty == 0:
+                sol.product_packaging_qty = 0
+                continue
+            # Consider uom
+            if sol.product_id.uom_id != sol.product_uom:
+                product_qty = sol.product_uom._compute_quantity(
+                    sol.product_uom_qty, sol.product_id.uom_id
+                )
+            else:
+                product_qty = sol.product_uom_qty
+            sol.product_packaging_qty = product_qty / sol.product_packaging.qty
+
+    def _inverse_product_packaging_qty(self):
+        for sol in self:
+            if not sol.product_packaging:
+                raise UserError(
+                    _(
+                        "You must define a package before setting a quantity "
+                        "of said package."
+                    )
+                )
+            if sol.product_packaging.qty == 0:
+                raise UserError(
+                    _("Please select a packaging with a quantity bigger than 0")
+                )
+            sol.write(
+                {
+                    "product_uom_qty": sol.product_packaging.qty * sol.product_packaging_qty,
+                    "product_uom": sol.product_packaging.product_uom_id.id,
+                }
+            )
+
+    @api.onchange('product_packaging')
+    def _onchange_product_packaging(self):
+        if self.product_packaging:
+            self.update(
+                {
+                    "product_packaging_qty": 1,
+                    "product_uom_qty": self.product_packaging.qty,
+                    "product_uom": self.product_id.uom_id,
+                }
+            )
+        else:
+            self.update(
+                {
+                    "product_packaging_qty": 0,
+                }
+            )
+        return super()._onchange_product_packaging()
+
+    @api.onchange('product_packaging_qty')
+    def _onchange_product_packaging_qty(self):
+        if self.product_packaging_qty and not self.product_packaging:
+            raise UserError(
+                _(
+                    "You must define a package before setting a quantity "
+                    "of said package."
+                )
+            )
+        else:
+            self.update(
+                {
+                    "product_uom_qty": self.product_packaging_qty * self.product_packaging.qty,
+                    "product_uom": self.product_id.uom_id,
+                }
+            )
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -83,3 +83,15 @@ class SaleOrderLine(models.Model):
                     "product_uom": self.product_id.uom_id,
                 }
             )
+
+    @api.onchange('product_uom_qty')
+    def _onchange_product_uom_qty(self):
+        """
+        Ensure a warning is raised when changing the package if the qty
+        is not a multiple of the package qty.
+        """
+        # TODO Drop once https://github.com/odoo/odoo/pull/49150/ is merged
+        res = super()._onchange_product_uom_qty()
+        if not res:
+            res = self._check_package()
+        return res
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -12,6 +12,7 @@ class SaleOrderLine(models.Model):
         string="Package quantity",
         compute="_compute_product_packaging_qty",
         inverse="_inverse_product_packaging_qty",
+        digits="Product Unit of Measure",
     )
 
     @api.depends(
@@ -35,6 +36,12 @@ class SaleOrderLine(models.Model):
                 product_qty = sol.product_uom_qty
             sol.product_packaging_qty = product_qty / sol.product_packaging.qty
 
+    def _prepare_product_packaging_qty_values(self):
+        return {
+            "product_uom_qty": self.product_packaging.qty * self.product_packaging_qty,
+            "product_uom": self.product_packaging.product_uom_id.id,
+        }
+
     def _inverse_product_packaging_qty(self):
         for sol in self:
             if not sol.product_packaging:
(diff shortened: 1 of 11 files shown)


rousseldenis:
@grindtildeath Are you aware of this one ?

https://github.com/OCA/sale-workflow/pull/858

grindtildeath:
@rousseldenis To me the usage looks slightly different and IMO both modules can co exist. 
cc @jgrandguillaume your call anyway :wink:

jgrandguillaume:
Hi @rousseldenis @grindtildeath,


I've checked this module already before asking this one, but it doesn't fit here (or I didn't got it right may be :) ?).

With that module we link a UoM with a packaging and this is absolutely not possible in our case. This is convenient if you work with similar package across your product. Here we have each product with its own packaging.

We do not want any link between uom and packaging here. e.g.

I sell gloves by pair, package in box of 100 or pallet of 10'000
I sell pants by unit, package in box of 20 or pallet of 2'000

I just need a feature in my SO line that allows me to compute in packaging, something like:
 * I order 10 boxes of gloves or
 * I order 5 pallets of pants
And thus must compute proper unit quantity behind.

Could we do this with you module as well ? Did I miss something here ? Also it pull the dependencies on https://github.com/OCA/product-attribute/pull/486

And using this force me in my case to define on UoM per packaging I have, to much work for tens of thousands of them.

May be I didn't understood it well... you tell me.

cc @jbaudoux

rousseldenis:
> Hi @rousseldenis @grindtildeath,
> 
> I've checked this module already before asking this one, but it doesn't fit here (or I didn't got it right may be :) ?).
> 
> With that module we link a UoM with a packaging and this is absolutely not possible in our case. This is convenient if you work with similar package across your product. Here we have each product with its own packaging.
> 
> We do not want any link between uom and packaging here. e.g.
> 
> I sell gloves by pair, package in box of 100 or pallet of 10'000
> I sell pants by unit, package in box of 20 or pallet of 2'000
> 
> I just need a feature in my SO line that allows me to compute in packaging, something like:
> 
> * I order 10 boxes of gloves or
> * I order 5 pallets of pants
>   And thus must compute proper unit quantity behind.
> 
> Could we do this with you module as well ? Did I miss something here ? Also it pull the dependencies on [OCA/product-attribute#486](https://github.com/OCA/product-attribute/pull/486)
> 
> And using this force me in my case to define on UoM per packaging I have, to much work for tens of thousands of them.
> 
> May be I didn't understood it well... you tell me.
> 
> cc @jbaudoux

Indeed, this is convenient for the way you want to use (we do as well for several customers).

e.g.: you can sell bottles of water per unit or per 6. So, in sale order line, you can choose the per unit package or the per 6 one.
You can easily adapt your case as if the packaging '10 boxes of gloves' is defined on your product, just select it in sale order line and everything is computed for you. I have no time now to explain furthermore but we use it for years now and fits nearly all cases we have.

You have also the purchase_packaging module that allow you to define the packaging for purchases.

jgrandguillaume:
@rousseldenis I've seen this since months already. But trying them forced me to define every packaging I have as a UoM as well which I find to much work for no value IMO.

e.g. If I have 10 pack of gloves and 10 pack of pants, I need one UoM, and then I I have pallet of 200 and 300 IU need two UoM linked to two packaging right ? -> I end up creating as must UoM as I have of packaging quantity.. what's the point ?

This what I don't want to do.

Our case [1] VS your case [2]


[1]

Screenshot from 2020-04-08 13-36-11
Screenshot from 2020-04-08 13-33-38
Screenshot from 2020-04-08 13-36-57
Screenshot from 2020-04-08 13-52-43


[2]

Other screen looks similar, but this one, the Unit of measure, where I need to define one for each packaging quantity I have !? I mean what is the win ?

Screenshot from 2020-04-08 13-51-24

jbaudoux:
Having seen both approach, the result from user experience is more or less the same.
The main question is what is the pro/con of having many UoM as in the Acsone approach https://github.com/OCA/sale-workflow/pull/858 ?
And for what should I use a unit of measure?
This has mainly implication on the business document (SO, PO, invoice, stock.picking) and product management.

## SO, PO, Invoice
If a pallet is 72 units, and I create a document line where I put that I want a pallet product packaging and store it visible on the line. Should I also write 1 (qty) "pallet of 72" (=uom) on the document or should I write 72 (qty) "units" (=uom) ?
I would tend to the second approach that prevents to create many uom in Odoo.
Also, if I have to exchange the document in EDI, the uom should ideally match a code from the UN/ECE list (https://docs.peppol.eu/poacc/upgrade-3/codelist/UNECERec20/). There I find that box, bag, pallet are recognized uom. But I do not find "pallet of 72", "pallet of 80", "pallet of 120" which are different packagings. So I will communicate 72 units packaged on a pallet where 72 is the quantity and units and pallets are known unit code. Of course you could convert any uom to known unit code in the EDI interface.

## Logistics
If a pallet is made of 10 transport boxes containing 8 small boxes containing 40 units. So 1 pallet = 3200 units.
Showing on the stock picking "1 pallet" or showing "3200 units" does not really help counting that there are effectively 10 transport boxes on the pallet. So both solutions need to show the product packaging and not only the qty and uom.

## Product Management
Having so many uom seems to me not convenient to select the right one as base uom for a new product. In Acsone approach, I would add a domain to hide all uom that are linked to a packaging to only be able to select base uoms when configuring the product.

## UoM naming
Creating many uoms, how do you name them?
Either you have many uom with the same name but different factors and I find it confusing at reading.
Or you have different name but then it becomes very long like: "Box of 8 small boxes of 40 units" (factor = 320)

Tonow-c2c (approved):
Code LGTM

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @api.onchange("product_uom_qty")
+    def _onchange_product_uom_qty(self):
+        """
+        Ensure a warning is raised when changing the package if the qty
+        is not a multiple of the package qty.
+        """
+        # TODO Drop once https://github.com/odoo/odoo/pull/49150/ is merged
+        res = super()._onchange_product_uom_qty()
+        if not res:
```
MiquelRForgeFlow, line 88:
```suggestion
        if not res and not (self.state == 'sale' and self.product_id.type in ['product', 'consu'] and self.product_uom_qty < product_uom_qty_origin and self.product_uom_qty < self.qty_delivered):
```
according to your PR to odoo

sale_order_line_packaging_qty/views/sale_order.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="view_order_form_inherit_sale_stock_inherit" model="ir.ui.view">
+        <field name="name">sale.order.form.sale.stock.inherit</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale_stock.view_order_form_inherit_sale_stock"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_packaging']" position="after">
```
MiquelRForgeFlow:
```suggestion
            <xpath expr="//form/group/group/field[@name='product_packaging']" position="after">
```

sale_order_line_packaging_qty/views/sale_order.xml
```diff
@@ -0,0 +1,16 @@
...
+                <field name="product_packaging_qty" groups="product.group_stock_packaging" attrs="{'invisible': [('product_packaging', '=', False)]}"/>
+            </xpath>
+            <xpath expr="//field[@name='product_packaging' and @optional='show']" position="before">
```
MiquelRForgeFlow:
```suggestion
            <xpath expr="//tree/field[@name='product_packaging']" position="before">
```

sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
```diff
@@ -0,0 +1,35 @@
...
+        order = self.env["sale.order"].create({"partner_id": self.partner.id})
+        order_line = self.env["sale.order.line"].create(
+            {
+                "order_id": order.id,
+                "product_id": self.product.id,
+                "product_uom": self.product.uom_id.id,
+                "product_uom_qty": 3.0
+            }
+        )
+        order_line.write({"product_packaging": self.packaging})
```
MiquelRForgeFlow, line 27:
```suggestion
        order_line.write({"product_packaging": self.packaging})
        order_line._onchange_product_packaging()
```

sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
```diff
@@ -0,0 +1,35 @@
...
+        self.assertEqual(order_line.product_uom_qty, 5.0)
+        self.assertEqual(order_line.product_packaging_qty, 1.0)
+        order_line.write({"product_packaging_qty": 3.0})
```
MiquelRForgeFlow, line 31:
```suggestion
        order_line.write({"product_packaging_qty": 3.0})
        order_line._onchange_product_packaging_qty()
```

sale_order_line_packaging_qty/models/sale_order_line.py
grindtildeath: I'm not sure to understand why we'd need this condition as we only want to ensure `_check_package` is called when `product_uom_qty` is changed :no_mouth:

sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
grindtildeath: Actually we don't need this onchange as it must be handled by the inverse function.

grindtildeath (commented):
@MiquelRForgeFlow Thx for your review.
I changed a bit the XML to make xpath conditions more explicit, but still not sure to understand why we'd need this condition before calling `_check_package`

sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
guewen: We do need it because the inverse is called only when the record is saved. I'll add it back

grindtildeath: My bad :facepalm: , got tricked by the write in unit test and removed it. Thanks for restoring.

jgrandguillaume:
@grindtildeath @guewen Would you check travis please ?

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,89 @@
...
+    def _inverse_product_packaging_qty(self):
+        for sol in self:
+            if not sol.product_packaging:
+                raise UserError(
+                    _(
+                        "You must define a package before setting a quantity "
+                        "of said package."
+                    )
+                )
+            if sol.product_packaging.qty == 0:
```
i-vyshnevska:
```suggestion
            if sol.product_packaging_qty == 0:
```
and you may need to check it every where

i-vyshnevska (changes requested):

MiquelRForgeFlow (approved):
LGTM

simahawk:
@i-vyshnevska good for you?
@grindtildeath the build needs some love ;)

i-vyshnevska (approved):

PR #1122, opened 2020-04-21
[10.0][ADD] sale_order_pending_payment

LRovira-PlanetaTIC:
Show amount total pending in sale order form and template when an advance invoice has already been paid.

cc @PlanetaTIC


rousseldenis:
@LRovira-PlanetaTIC Thanks for this. Could you check travis ?

Is this a real migration as the module appears to be not present in any main branch yet?

LRovira-PlanetaTIC:
> @LRovira-PlanetaTIC Thanks for this. Could you check travis ?
> 
> Is this a real migration as the module appears to be not present in any main branch yet?

@rousseldenis  You are right, I was wrong to put the MIG label
When I can I check the travis

PR #1123, opened 2020-04-23
[13.0][IMP][sale_order_type] Use computed stored writable fields instead of onchanges

ivantodorovich:
Switch to computed writable fields.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -14,10 +14,18 @@ class AccountMove(models.Model):
         readonly=False,
         store=True,
     )
+    invoice_payment_term_id = fields.Many2one(
+        compute="_compute_invoice_payment_term_id", store=True, readonly=False,
+    )
+    journal_id = fields.Many2one(
+        compute="_compute_journal_id", store=True, readonly=False,
+    )
 
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
         for record in self:
+            if not record.is_invoice():
+                continue
             if not record.partner_id:
                 record.sale_type_id = self.env["sale.order.type"].search([], limit=1)
             else:
@@ -32,10 +40,18 @@ class AccountMove(models.Model):
                 if sale_type:
                     record.sale_type_id = sale_type
 
-    @api.onchange("sale_type_id")
-    def onchange_sale_type_id(self):
-        # TODO: To be changed to computed stored readonly=False if possible in v14?
-        if self.sale_type_id.payment_term_id:
-            self.invoice_payment_term_id = self.sale_type_id.payment_term_id.id
-        if self.sale_type_id.journal_id:
-            self.journal_id = self.sale_type_id.journal_id.id
+    @api.depends("sale_type_id")
+    def _compute_invoice_payment_term_id(self):
+        if hasattr(super(), "_compute_invoice_payment_term_id"):
+            super()._compute_invoice_payment_term_id()
+        for rec in self:
+            if rec.sale_type_id:
+                rec.invoice_payment_term_id = rec.sale_type_id.payment_term_id
+
+    @api.depends("sale_type_id")
+    def _compute_journal_id(self):
+        if hasattr(super(), "_compute_journal_id"):
+            super()._compute_journal_id()
+        for rec in self:
+            if rec.sale_type_id.journal_id:
+                rec.journal_id = rec.sale_type_id.journal_id
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -8,10 +8,14 @@ from odoo import api, fields, models
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    def _default_type_id(self):
+        return self.env["sale.order.type"].search([], limit=1)
+
     type_id = fields.Many2one(
         comodel_name="sale.order.type",
         string="Type",
         compute="_compute_type_id",
+        default=lambda x: x._default_type_id(),
         readonly=False,
         store=True,
     )
@@ -32,19 +36,16 @@ class SaleOrder(models.Model):
     @api.depends("partner_id", "company_id")
     def _compute_type_id(self):
         for record in self:
-            if not record.partner_id:
-                record.sale_type_id = self.env["sale.order.type"].search([], limit=1)
-            else:
-                sale_type = (
-                    record.partner_id.with_context(
-                        force_company=record.company_id.id
-                    ).sale_type
-                    or self.partner_id.commercial_partner_id.with_context(
-                        force_company=record.company_id.id
-                    ).sale_type
-                )
-                if sale_type:
-                    record.type_id = sale_type
+            sale_type = (
+                record.partner_id.with_context(
+                    force_company=record.company_id.id
+                ).sale_type
+                or self.partner_id.commercial_partner_id.with_context(
+                    force_company=record.company_id.id
+                ).sale_type
+            )
+            if sale_type:
+                record.type_id = sale_type
 
     @api.depends("type_id")
     def _compute_warehouse_id(self):
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -1,5 +1,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2020 Druidoo - Iván Todorovich
 
 from odoo import api, fields, models
 
@@ -10,13 +11,26 @@ class SaleOrder(models.Model):
     type_id = fields.Many2one(
         comodel_name="sale.order.type",
         string="Type",
-        compute="_compute_sale_type_id",
+        compute="_compute_type_id",
         readonly=False,
         store=True,
     )
+    warehouse_id = fields.Many2one(
+        compute="_compute_warehouse_id", store=True, readonly=False,
+    )
+    picking_policy = fields.Selection(
+        compute="_compute_picking_policy", store=True, readonly=False,
+    )
+    payment_term_id = fields.Many2one(
+        compute="_compute_payment_term_id", store=True, readonly=False,
+    )
+    pricelist_id = fields.Many2one(
+        compute="_compute_pricelist_id", store=True, readonly=False,
+    )
+    incoterm = fields.Many2one(compute="_compute_incoterm", store=True, readonly=False)
 
     @api.depends("partner_id", "company_id")
-    def _compute_sale_type_id(self):
+    def _compute_type_id(self):
         for record in self:
             if not record.partner_id:
                 record.sale_type_id = self.env["sale.order.type"].search([], limit=1)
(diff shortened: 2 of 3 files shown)


pedrobaeza:
This is `[IMP]` and not `[ADD]` as you are improving an existing module, not adding one, and maybe you can split the move of the field into other commit. Is this using already computed writeable fields? That can be also a good improvement.

ivantodorovich:
Hi @pedrobaeza 

Thanks for your review!
I've just split the commits and changed the commit messages.

Regarding your comment to use computed writeable fields, I'm not clear yet on the approach.
Is it something like this?

ie:

```py
warehouse_id = fields.Many2one(
    computed="_compute_warehouse_id",
    readonly=False,
    store=True,
)

# ...

@api.depends('type_id')
def _compute_warehouse_id(self):
    for rec in self:
        if rec.type_id.warehouse_id:
            rec.warehouse_id = rec.type_id.warehouse_id
```

Just to be clear this will result in making all of these fields computed, and each of them with their own `_compute` method:

sale.order:
- warehouse_id
- picking_policy
- payment_term_id
- pricelist_id
- incoterm_id

sale.order.line:
- route_id


Is there a real benefict to doing that? Because in this particular case it will result in more lines of code and probably less readable :thinking: 

I'm also not sure how if affects compatibility with other modules. What if there's another module making the `warehouse_id` a computed field, naming the compute method differently? :thinking:

rousseldenis:
@pedrobaeza What do you think of last comment ?

ivantodorovich:
ping @pedrobaeza

pedrobaeza:
Sorry for not answering. Too much mail traffic...

Very good questions:

> Is there a real benefit to doing that? Because in this particular case it will result in more lines of code and probably less readable thinking

Yes, the benefit is to avoid glue modules or spaghetti code calling all the onchanges. Examples:

- This is not needed anymore: https://github.com/OCA/sale-workflow/blob/5d62f471021ed366e9fedb51012192484550145f/sale_order_type/tests/test_sale_order_type.py#L66
- This module disappears: https://github.com/OCA/e-commerce/tree/10.0/website_sale_order_type

> I'm also not sure how if affects compatibility with other modules. What if there's another module making the warehouse_id a computed field, naming the compute method differently?

This is my only doubt about the process. I haven't tested this. I hope this works as previous onchanges: you can accumulate then depending on the compute method name. If so, then we should put a guideline that states that a compute method should be called with the module name suffix (example: `_compute_warehouse_id_sale_order_type`) if the field is not defined in our module. If defined, then regular compute name is expected (`_compute_warehouse_id`). Is it too much to ask you to make the test to see how it works?

ivantodorovich:
@pedrobaeza unfortunately it's not accumulating

Take this scenario for instance:

Module A
-------------

```py
class MyModel(models.Model):
    _name = "my.model"

    field_a = fields.Integer()
    field_z = fields.Integer()
```

Module B
-------------
depends on Module A

```py
class MyModel(models.Model):
    _inherit = "my.model"

    field_b = fields.Integer()
    field_z = fields.Integer(compute="_compute_field_z_module_b", store=True, readonly=False)

    @api.depends('field_b')
    def _compute_field_z_module_b(self):
        for rec in self:
            rec.field_z = rec.field_b
```

Module C
-------------
depends on Module A

```py
class MyModel(models.Model):
    _inherit = "my.model"

    field_c = fields.Integer()
    field_z = fields.Integer(compute="_compute_field_z_module_c", store=True, readonly=False)

    @api.depends('field_c')
    def _compute_field_z_module_c(self):
        for rec in self:
            rec.field_z = rec.field_c
```

Result
---------

Only changes in `field_c` will trigger the update on `field_z`, because it's the last module loaded



So.. IMO this approach would raise compatibility issues. As I said, if there's another module trying to make the `warehouse_id` a computed method, only one of them would work and we'd need a glue module to make them work together :/ while with onchanges this is not required..

Sure, we also need glue modules like https://github.com/OCA/e-commerce/tree/10.0/website_sale_order_type, but at least it's more explicit

pedrobaeza:
@ivantodorovich thanks for the investigation. This is what you have to do for making it work:

Module B
-------------

```py
class MyModel(models.Model):
    _inherit = "my.model"

    field_z = fields.Integer(compute="_compute_field_z", store=True, readonly=False)

    @api.depends('field_b')
    def _compute_field_z(self):
        if hasattr(super(), '_compute_field_z'):
            super()._compute_field_z()
        # do stuff
```

Module C
-------------

```py
class MyModel(models.Model):
    _inherit = "my.model"

    field_z = fields.Integer(compute="_compute_field_z", store=True, readonly=False)

    @api.depends('field_c')
    def _compute_field_z(self):
        if hasattr(super(), '_compute_field_z'):
            super()._compute_field_z()
        # do stuff
```
I have checked and it works, and `depends` are aggregated, so I still think it's valuable. What do you think? We can create a guideline for this in OCA Guidelines for adjusting all to this standard.

ivantodorovich:
@pedrobaeza I've just pushed the change to deprecate onchanges

Functionally, it's working ok. But it broke the tests.. 
I thought computed fields were recomputed "on demand", but apparently they're not?

I've tried with `order.recompute()` after `create()`, but it's still not working.
https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5769

Any idea? :thinking:

ivantodorovich:
@pedrobaeza I fixed tests but I'm not quite happy with what I had to do. I would appreciate your input here.

Apparently computed writeable stored fields are not recomputed on `create` if the field is explicitly set on vals, or if there's a default value.

ie:

```py
    _inherit = "res.partner"

    test = fields.Char(
        'test', compute='_compute_test', readonly=False, store=True, default='xxx',
    )

    @api.depends('name')
    def _compute_test(self):
        for rec in self:
            rec.test = rec.name
```

```py
    def test_test(self):
        p = self.env['res.partner'].create({'name': 'asdasd'})
        self.assertEqual(p.name, p.test)
```

```py
    self.assertEqual(p.name, p.test)
AssertionError: 'asdasd' != 'xxx'
- asdasd
+ xxx
```

Tests were failing because of that, mainly:

- `warehouse_id` is set explicitly here: https://github.com/odoo/odoo/blob/13.0/addons/sale_stock/models/sale_order.py#L66

- `picking_policy` also has a default value:
https://github.com/odoo/odoo/blob/13.0/addons/sale_stock/models/sale_order.py#L27

- `journal_id` has a default value:
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L121

and others..
So they're not recomputed on record create.

I overloaded `create()` and to explicitly trigger the recomputations with:

```py
    @api.model_create_multi
    def create(self, vals_list):
        res = super().create(vals_list)
        # This will trigger recomputations and overload default values
        # for the new computed fields, if set on the sale.order.type
        res.modified(["sale_type_id"])
```

But for some reason `self.assertEqual(invoice.journal_id, sale_type.journal_id)` was still failing (https://github.com/druidoo-dev/sale-workflow/blob/13.0-sale_order_type/sale_order_type/tests/test_sale_order_type.py#L109) , so I had to manually recompute it again. I don't know why it's requiered but it doesn't work without it :/

I don't like much the fact that it's required to overload `create()` and add these lines for it to work..

--- 

Just to be clear this is not affecting the standard user flow : Going through form views. That works even without overloading `create()`.

This is only affecting the programatical creation of records through `create()`, like in the automated tests or code in other modules.

ivantodorovich:
ping @pedrobaeza :)

pedrobaeza:
Sorry, I haven't been able yet to check it

sale_order_type/models/account_move.py
```diff
@@ -32,10 +40,35 @@ def _compute_sale_type_id(self):
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        # This will trigger recomputations and overload default values
+        # for the new computed fields, if set on the sale.order.type
+        res.modified(["sale_type_id"])
+        # For some reason journal_id is not following this rule
+        # and needs to be explicitly recomputed
+        res._compute_journal_id()
+        return res
```
pedrobaeza: You can't launch `_compute_journal_id` on create. This way, you are overriding any specific `journal_id` passed in the vals, which is not correct. Having also `res.modifed...` seems not correct. Why do you need that? Please restore the previous situation and let me see the error in CI.

pedrobaeza (changes requested):
Sorry for my late reply. Please see my comment and I promise this time I will check it quickly.

sale_order_type/models/account_move.py
ivantodorovich: Yeah, you're right.

This is what I've tried to solve:

> Apparently computed writeable stored fields are not recomputed on create if the field is explicitly set on vals, or if there's a default value.
 Tests were failing because of that, mainly:
warehouse_id is set explicitly here: https://github.com/odoo/odoo/blob/13.0/addons/sale_stock/models/sale_order.py#L66
picking_policy also has a default value:
https://github.com/odoo/odoo/blob/13.0/addons/sale_stock/models/sale_order.py#L27
journal_id has a default value:
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L121
and others..
So they're not recomputed on record create.


I've just pushed a commit reverting this, so you can test.

pedrobaeza:
@ivantodorovich please give me permissions in your branch for pushing an updated version. I'm almost there. The main problem is that you are not mimicking the behavior in UI in tests. I'm doing it with SSF (server side forms), and this has discover an underlying problem that I'm debugging now.

ivantodorovich:
done ;)

pedrobaeza:
You can look for the temporary fix while I check the problem with `payment_term_id` and `pricelist_id` fields (checked also in UI that they are not automatically changed when changing the partner with a sale type).

ivantodorovich:
I'm not satisfied with the result, to be honest.
If we go back to the starting point, the main benefit of computed writeable field is to avoid modules like https://github.com/OCA/e-commerce/blob/10.0/website_sale_order_type/models/sale_order.py.

But in fact, if we need SSF to pass the tests, it means we'll still need glue modules like that one

I think I'll revert this PR to how it was before the computed writable fields. Someone financed this routes feature, and it's a shame it's not merged because of a not really necessary refactor. I'll create another PR to discuss / continue the work on computed writeable stored fields

pedrobaeza:
Hi, @ivantodorovich, I agree right now the switch is not satisfactory, but the culprit is not the computed writable fields. Anyway, I have done the following:

- Push to 13.0 branch directly your first 3 commits with valid changes adding the route and the UI improvement.
- Let this PR with the rest for switching to computed writable.
- I will continue here exploring the problems and having test for both UI and direct creation. The problem in direct creation may be due to code upstream that manipulates the dictionary in create/write methods.

ivantodorovich:
Thanks @pedrobaeza !

Re: the problem, yes it's because of that.
And also because computed stored writable fields are not recomputed on `create` if there are default values for these computed fields.

Look at the example I wrote here: https://github.com/OCA/sale-workflow/pull/1123#issuecomment-633639514

In upstream, many of the fields we're trying to compute have default values, or dictionary manipulation on create

PR #1134, opened 2020-05-07
Migrate portal_sale_personal_data_only to V13.0

chrisandrewmann:


pedrobaeza:
I'm afraid this is still suffering the same commit history problem. If the problem is that this module is not in 12.0, then you have to change the command for using the last source branch that contains the latest version (11.0? 10.0?).

chrisandrewmann:
Sorry i'm going to bail on doing any more of this as i'm clearly in over my head and will revoke my PR for this and any OCA modules. You're welcome to take it on but I'm at a loss.

pedrobaeza:
Well, I'm only trying to guide you on how to do it. I don't want to discourage you, but you have to understand that if the commit history is not preserved, the attribution of the whole source code is to you, which is not fair.

PR #1143, opened 2020-05-13
12.0 IMP sale_order_lot_selection: allow to confirm sale when lot unavailable

bealdav:
You may want use unavailable lot (i.e. you sell a product manufactured by a supplier
but with your requirements. Then you provide a lot number before to receive product

### DIFF
--- a/sale_order_lot_selection/models/sale.py
+++ b/sale_order_lot_selection/models/sale.py
@@ -1,5 +1,4 @@
-from odoo import api, fields, models, _
-from odoo.exceptions import UserError
+from odoo import api, fields, models
 
 
 class SaleOrderLine(models.Model):
@@ -30,27 +29,3 @@ class SaleOrderLine(models.Model):
         return {
             'domain': {'lot_id': [('id', 'in', available_lot_ids)]}
         }
-
-    def assign_move_with_lots(self):
-        for move in self.move_ids:
-            if move.state == 'confirmed':
-                move._action_assign()
-                move.refresh()
-            if (
-                move.state != "assigned"
-            ):
-                raise UserError(
-                    _("Can't reserve products for lot %s") % self.lot_id.name
-                )
-
-
-class SaleOrder(models.Model):
-    _inherit = "sale.order"
-
-    @api.multi
-    def action_confirm(self):
-        res = super().action_confirm()
-        for line in self.order_line:
-            if line.lot_id:
-                line.assign_move_with_lots()
-        return res
--- a/sale_order_lot_selection/models/sale.py
+++ b/sale_order_lot_selection/models/sale.py
@@ -31,52 +31,26 @@ class SaleOrderLine(models.Model):
             'domain': {'lot_id': [('id', 'in', available_lot_ids)]}
         }
 
-
-class SaleOrder(models.Model):
-    _inherit = "sale.order"
-
-    @api.model
-    def get_move_from_line(self, line):
-        move = self.env['stock.move']
-        # i create this counter to check lot's univocity on move line
-        lot_count = 0
-        for p in line.order_id.picking_ids:
-            for m in p.move_lines:
-                move_line_id = m.move_line_ids.filtered(
-                    lambda line: line.lot_id)
-                if move_line_id and line.lot_id == move_line_id[:1].lot_id:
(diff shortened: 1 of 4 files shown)


rvalyi:
I understand perfectly you may want to sell new lots (not available then) and this is a great improvement.

But then that would be an assumed CHOICE. With your PR what about a company who would still want to avoid selling unavailable lots in its own use case? I mean I feel there should be an option in the the company, settings or sale team and leave the former `_check_move_state()` as an option, no?

bealdav:
You're right, I'll put an overidable method to allow to skip

bealdav:
Fail comes from 12.0 sale_stock_secondary_unit and sale_elaboration

sale_order_lot_selection/models/sale.py
```diff
@@ -52,32 +52,28 @@ def get_move_from_line(self, line):
...
-                move.refresh()
-            if move.state != 'assigned':
-                raise UserError(_('Can\'t reserve products for lot %s') %
-                                line.lot_id.name)
+    def forbid_unavailable_lot_reservation(self):
+        """ You may want use unavailable lot in your stock move:
+            i.e. you sell a product manufactured by a supplier
+            but with your requirements including traceability on received lot.
+
+            In this case override this method to return True in your use case
```
rvalyi:
```suggestion
            In this case override this method to return False in your use case
```

sale_order_lot_selection/models/sale.py
```diff
@@ -52,32 +52,28 @@ def get_move_from_line(self, line):
...
-                    move.reserved_availability
-                )
-                if unreserved_moves:
+                move = self.get_move_from_line(line)
+                if move.state == 'confirmed':
+                    move._action_assign()
+                    move.refresh()
+                if (
+                    move.state != "assigned"
+                    and self.forbid_unavailable_lot_reservation()
```
rvalyi: would be better to pass the line and/or the move to this forbid_unavailable_lot_reservation hook (so we can have the lot and the product), so you can do the necessary checking that you are indeed in your special case, no?

rvalyi (dismissed):

sale_order_lot_selection/models/sale.py
rvalyi: (True is the default here)

bealdav:
Concerned by https://github.com/OCA/sale-workflow/issues/1166

bealdav:
Hi @eLBati @PCatinean I've made changes explained https://github.com/OCA/sale-workflow/issues/1166

Could you make a review please ?

PCatinean:
Hi @bealdav, either it's been ages since I've been working with Gitlab only and forgot how to use Github or something else is wrong. The diff shows just removal of functionality from sale.py, some demo data addition and changes in tests. Am I missing something?

bealdav:
You're right the diff show removing code as explained in #1166 
Because this code prevent a behaviour (you can't confirm a sale order when no lot is not available)

PCatinean:
Ah understood, I was referring to the conversation you had with rvalyi and was looking for the optional method you mentioned. Will you still be adding that?

bealdav:
Yes but developers (you and @eLBati) can't remember (even searching in their history) why this code was added. Then here 2 possibilities

- we keep these features for ever even nobody know what or who they are suppose to serve (see https://en.wikipedia.org/wiki/The_Cathedral_and_the_Bazaar)

- we clean undocumented features (problably needed when reach an older version for a specific project) with updated code matching with normal behavior of odoo at this version

If we want to keep OCA great for the future I think it's better to move towards the second strategy

PCatinean (approved):

sebastienbeau (approved):
LGTM, ok for dropping unused feature. In any case if we want to add teh contstrait it should be done in a separated module

sebastienbeau:
@rvalyi can you finish you review?

marcelsavegnago (approved):

sebastienbeau:
@rousseldenis Can you merge it ? (I do not have the right anymore)

bealdav:
Please @rousseldenis could you merge it ? Thank you.

bealdav:
ping @rousseldenis  could you have a look ?

PR #1147, opened 2020-05-19
[12.0][FIX] sale_elaboration: Tests

### LINKED ISSUE #1146: [12.0] Main branch is red on travis (sale_elaboration + sale_stock_secondary_unit).
https://travis-ci.org/github/OCA/sale-workflow/jobs/687769282#L2337

Sale order in tests is created without company_id (not required), so the invoice creation raises as it takes the one on sale order (and is required).

Last good build on 24th of April, first raise on 13th of May. 

https://github.com/OCA/sale-workflow/blob/12.0/sale_elaboration/tests/test_sale_elaboration.py#L125

Didn't find any relevant changes in Odoo code.

rousseldenis:
fixes #1146 partially

### DIFF
--- a/sale_elaboration/tests/test_sale_elaboration.py
+++ b/sale_elaboration/tests/test_sale_elaboration.py
@@ -104,6 +104,7 @@ class TestSaleElaboration(SavepointCase):
     def test_invoice_elaboration(self):
         so = self.env['sale.order'].new({
             'partner_id': self.partner.id,
+            'company_id': self.env.user.company_id.id,
             'order_line': [
                 (0, 0, {
                     'name': self.product_elaboration_A.name,


pedrobaeza:
Maybe it's better to use `common.Form` for populating records? Do you want me to propose such change?

rousseldenis:
> Maybe it's better to use `common.Form` for populating records? Do you want me to propose such change?

Nope at this time. See https://github.com/OCA/sale-workflow/issues/1146#issuecomment-630660203 Maybe in another PR

pedrobaeza:
Well, I see this solution very fragile, as next time it's going to be another field, so I prefer the long term solution.

rousseldenis:
> Well, I see this solution very fragile, as next time it's going to be another field, so I prefer the long term solution.

Note that the bug can be reproduced on runbot if company field is not filled in :

image


IMHO it's a gap in Odoo code. A better error message on core should be the solution (as inconsistency between not required field in sales and required in invoices)

pedrobaeza:
Yes, but with my solution you fix both: tests and runbot.

rousseldenis:
@pedrobaeza

pedrobaeza:
I will fix it today

PR #1148, opened 2020-05-19
[10.0][FIX] sale_promotion_rule: Enable sequence order

rousseldenis:
Sequence order was never applied du of missing _order.

The only one used was 'id'


rousseldenis:
@sebastienbeau @Cedric-Pigeon @lmignon

lmignon (approved):

Cedric-Pigeon (approved):

gurneyalex (approved):

PR #1150, opened 2020-05-19
[13.0][FIX] sale_order_product_recommendation: Sale order discount not applied. Helper to call all so line onchanges

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -7,6 +7,12 @@ from datetime import datetime, timedelta
 from odoo import api, fields, models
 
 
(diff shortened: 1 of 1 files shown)


rousseldenis:
@sergio-teruel Are you aware of https://github.com/OCA/server-tools/tree/13.0/onchange_helper ?

pedrobaeza:
From v12, you don't need that module nor this trick. Create an instance of `test.Form` and it will do the same of `onchange_helper`.

HaraldPanten (approved):
Functional review. Seems to be OK 👍

pedrobaeza:
With #1157 this is not needed

pedrobaeza:
Closing as not needed.

PR #1151, opened 2020-05-25
[IMP] sale_order_lot_selection: Select the same product in a sale order

manuelcalerosolis:
When trying to sell on two different order lines, the same product and with the same batch number, (for example, one unit with a price and the other with a zero price), the error "Can't retrieve lot on stock" appears. 

This PR corrects that error and also provides more information about the line item that is causing the problem.

### DIFF
--- a/sale_order_lot_selection/models/sale.py
+++ b/sale_order_lot_selection/models/sale.py
@@ -47,9 +47,10 @@ class SaleOrder(models.Model):
                 if move_line_id and line.lot_id == move_line_id[:1].lot_id:
                     move = m
                     lot_count += 1
-                    # if counter is 0 or > 1 means that something goes wrong
-                    if lot_count != 1:
-                        raise UserError(_('Can\'t retrieve lot on stock'))
+                    # if counter is 0 means that something goes wrong
+                    if lot_count == 0:
+                        raise UserError(_('Can\'t retrieve product stock %s, for lot %s') %
+                                (line.product_id.name, line.lot_id.name))
         return move
 
     @api.model


bealdav:
Hi @manuelcalerosolis could you have a look here
https://github.com/OCA/sale-workflow/pull/1143/files#diff-a61bf45f290b08d6f9db336dcb39ff70L55

Don't you think my patch can also solve your problem ?
If yes, could we only go on with it because it solve the problem in a more generic way ? 
Are you ok to review it ?

manuelcalerosolis:
HI @bealdav I will review it, if it solves my problem, perfect.

Thanks

manuelcalerosolis:
@bealdav your PR not solves my problem.

Tray this

image

I get this error 

image

I think they solve different situations.

Regards

pedrobaeza:
@manuelcalerosolis you are still committing with the wrong email. Remember you have to change it at your compute level through git commands.

See https://www.google.com/search?client=ubuntu&channel=fs&q=git+email+address&ie=utf-8&oe=utf-8

sale_order_lot_selection/models/sale.py
```diff
@@ -47,9 +47,10 @@ def get_move_from_line(self, line):
                 if move_line_id and line.lot_id == move_line_id[:1].lot_id:
                     move = m
                     lot_count += 1
-                    # if counter is 0 or > 1 means that something goes wrong
-                    if lot_count != 1:
-                        raise UserError(_('Can\'t retrieve lot on stock'))
+                    # if counter is 0 means that something goes wrong
+                    if lot_count == 0:
+                        raise UserError(_('Can\'t retrieve product stock %s, for lot %s') %
```
bealdav, line 52: I read again your code.
It seems that `loc_count` can't be equals to 0 because in line 49 it's set from 0 to 1 or upper.
Exception is never raised.

But I need to solve this case. I'll try to fix in another PR. Thanks

rousseldenis:
@manuelcalerosolis @bealdav What's the status of this ?

bealdav:
Probably need to stay as it for now because no time to deal with it

PR #1152, opened 2020-05-25
[MIG][13.0] sale_tier_validation: Migration to v13.0

sergio-teruel:
cc @Tecnativa TT24064
@carlosdauden @chienandalu Can you review?


HaraldPanten (changes requested):
Hi @sergio-teruel ,

It's functionally working fine, but I would change something. Look:

Captura de pantalla 2020-05-26 a las 9 16 07

When you try to configure the domain, the options are not shown properly. If you look at my screenshot, user can not see the fields in the "drop-down" menu as it appears mostly "cut".

Could you have a look?

HaraldPanten:
I also found a kind of bug. Look:

Captura de pantalla 2020-05-26 a las 10 16 57

Once I saved a tier validation, If I open it and I try to edit the model, I can select any model. It's not correct. I should only be able to select sale.order model.

sergio-teruel:
Hi @HaraldPanten These errors are not related to this module.
For reference model weird behavior is due to this onchange https://github.com/OCA/server-ux/blob/17da52b2e256745c1b59f57c29457ea2e4c1ba89/base_tier_validation/models/tier_definition.py#L64-L63.
For selector domain issue is due to Odoo core widget and I known it.
These fixes have not must fixed in this PR.

HaraldPanten:
> Hi @HaraldPanten These errors are not related to this module.
> For reference model weird behavior is due to this onchange https://github.com/OCA/server-ux/blob/17da52b2e256745c1b59f57c29457ea2e4c1ba89/base_tier_validation/models/tier_definition.py#L64-L63.
> For selector domain issue is due to Odoo core widget and I known it.
> These fixes have not must fixed in this PR.

Hi @sergio-teruel I understand...

Even though, why not adding a "notes" field at the bottom, after the domain selector? Wouldn't we save part of this issue like that, adding a small [IMP]? With the "notes" field we could offer better user functionalities, as well. What do you think?

I'll have a look at the onchange problem 👍 

THX

sergio-teruel:
@HaraldPanten I think it's better create a Odoo PR for fix the issue...
@LoisRForgeFlow What do you think??

HaraldPanten (approved):
> @HaraldPanten I think it's better create a Odoo PR for fix the issue...
> @LoisRForgeFlow What do you think??

Maybe you're right. Will you manage this, or... How do you use to proceed in these cases?

LoisRForgeFlow:
@sergio-teruel @HaraldPanten Didn't know about the issue with the domain widget, thanks for pointing. I agree with Sergio, I rather do a PR to Odoo fixing that than dirty workarounds, this is an issue that might be present in more modules than just tier validation. Refer the PR to us here when you have it ready!

chienandalu (commented):
Maybe it's something from the base module, but I get the validation messages duplicated:
image

sergio-teruel:
@chienandalu this occours because you have defined two definition for the same user with the same domain...
Usualy you will have diferent domains..
Anyway this is manage in base module.
https://github.com/OCA/server-ux/blob/4865fac38e8598b3faea1110fd80b387cdecf177/base_tier_validation/models/tier_validation.py#L200

rousseldenis (approved):
Code review. Thanks for the tests!

chienandalu (approved):
Oh, that was it :sweat_smile:

PR #1047, opened 2020-01-16
[13.0][MIG] sale_order_price_recalculation

bjornbillen:

PR #1155, opened 2020-05-28
[13.0][MIG] sale_order_price_recalculation

rousseldenis:
Supersedes #1047


HaraldPanten (changes requested):
It's functionally working fine, but could you check travis?

rousseldenis:
> It's functionally working fine, but could you check travis?

Yes indeed. I hesitated to pick the code changed in former PR : https://github.com/OCA/sale-workflow/pull/1047/commits/8f8b0cbd6951bf735a005827861f0f9b30704140#diff-55013901e41dac5691e86b337c0be50aR20

I try to find a solution

rousseldenis:
@Cedric-Pigeon @lmignon  @pedrobaeza What do you think (last commit) ?

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+# Copyright 2020 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _recalculate_prices_get_fields(self):
+        return ["price_unit", "discount"]
```
lmignon, line 15: @rousseldenis Your change to use `onchange_helper` LGTM nevertheless can you add a comment to explain the purpose of this methods and where these fields are used.

pedrobaeza: Please don't use `onchange_helper` and add a new dependency. You just need to use core `odoo.tests.Form`, that makes the same. Example:

https://github.com/OCA/sale-workflow/pull/1157/commits/458ef3a4796b935d3e05f79b7a0481a592a43d6b

lmignon: I am reluctant to use a method written for testing in business code. Especially since it degrades performance quite strongly.

pedrobaeza: It's already used in business code in Odoo itself, and it degrades the same as `onchange_helper`, as it triggers onchanges.

rousseldenis: I was sure you want to use it but I agree with @lmignon. That's not because Odoo uses it in their code, that's a good idea...

But after a little search, they use it three times only, twice in localized or external contribution modules.

rousseldenis: Moreover, the import speaks by itself... 

`from odoo.tests.common import Form`

rousseldenis: The only thing I found is that product.price is not in tocompute, so this is the cache value which is used (wrong one).

The only 'hack' I found is to add price field. But this is not sustainable as we cannot figure out which new computed field could be added to the onchange.

But as we know which one we want to modify, onchange_helper is definetely the solution.

image

pedrobaeza: About the `Form` being in `odoo.tests`, it's because historical reasons, starting as a helper for tests, but then seeing a convenient way to populate all values, and spreading its use across business code. You can see several examples of usage in latest developed modules:

https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/account_facturx/models/account_move.py#L99
https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/l10n_be_edi/models/account_invoice.py#L49
https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/l10n_it_edi/models/account_invoice.py#L361

The problem you mention you have in `onchange_helper` may not happen with this helper.

The reason I prefer it is for reducing maintenance burden of the extra module and having to deal with different logics depending on the source.

Luckily, both things will go away in v14 with the refactoring to computed writeable fields, but meanwhile...

Would you accept an alternate PR using the method I comment to compare?

HaraldPanten (approved):

sale_order_price_recalculation/models/sale_order.py
rousseldenis: > About the `Form` being in `odoo.tests`, it's because historical reasons, starting as a helper for tests, but then seeing a convenient way to populate all values, and spreading its use across business code. You can see several examples of usage in latest developed modules:
> 
> https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/account_facturx/models/account_move.py#L99
> https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/l10n_be_edi/models/account_invoice.py#L49
> https://github.com/odoo/odoo/blob/2b4b2ce8660326b1cb53b5624690718b89aaceda/addons/l10n_it_edi/models/account_invoice.py#L361
> 
> The problem you mention you have in `onchange_helper` may not happen with this helper.
> 
> The reason I prefer it is for reducing maintenance burden of the extra module and having to deal with different logics depending on the source.
> 
> Luckily, both things will go away in v14 with the refactoring to computed writeable fields, but meanwhile...
> 
> Would you accept an alternate PR using the method I comment to compare?

IMHO, onchange_helper would not be such a burden as already migrated and tested so much in previous versions. As you said, for v14, it will disappear, so... The reason whyForm() as been used in code is not clear.

As said in https://github.com/odoo/odoo/pull/25508#pullrequestreview-425975320, you cannot rely logic on that Form() implementation. I'm not a fool.

pedrobaeza: Why can't you rely on it? It's totally trustable. And I can say that because it's used in tests as well as in logic. I have been using it for a while now without too much problems. I'm not saying that in v14 it will disappear. What will disappear is the need to use it (the same as `onchange_helper`), due to the move to computed writeable, but meanwhile in v12/v13, I prefer to use SSF (acronym of Server Side Form, which is the official name for this) than to add a new piece in the puzzle through the OCA module, that has had also problems in the past and I remember a PR that lasted a lot.

benwillig (changes requested):
Functionnal tests ok for me.

Could you please just add a class on button in the form view to make them a bit more visible?
It's hard to see them because it's just text at the moment.

image

The PR also need a rebase.

sergio-teruel:
Please rebase to fix conflicts

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -0,0 +1,51 @@
...
+                # we make this to isolate changed values:
+            line2 = self.env["sale.order.line"].new(vals)
+            # we make this to isolate changed values:
+            line2.env.add_to_compute(
+                line2.product_id._fields["price"], line2.product_id
+            )
+            line2.product_uom_change()
+            line2._onchange_discount()
+            line.write({"price_unit": line2.price_unit, "discount": line2.discount})
+            # line2 = self.env["sale.order.line"].new(vals)
```
simahawk, line 31: @rousseldenis took a decision?
FWIW I'd go for `odoo.tests.Form` because despite were it's placed, it's reliable enough for this. 
In any case, as mentioned, the need for it will likely go away w/ v14. Just KISS :wink:

sale_order_price_recalculation/tests/test_sale_order_price_recalculation.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2015 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+# Copyright 2016 Vicent Cubells <vicent.cubells@tecnativa.com>
+# Copyright 2017 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2018 Duc Dao Dong <duc.dd@komit-consulting.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo.tests import common
+
+
+class TestSaleOrderPriceRecalculation(common.TransactionCase):
```
simahawk, line 9: Please:
1. use SavepointCase, seems pointless to re-create these records for each test
2. disable tracking globally using `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

sale_order_price_recalculation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
+This module add 2 buttons on sale orders (below sale order lines) that:
```
simahawk, line 1:
```suggestion
This module adds 2 buttons on sale orders (below sale order lines) that:
```

sale_order_price_recalculation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+
+* Recalculates the prices of the order lines that contain a product in them.
```
simahawk, line 3:
```suggestion
* Recalculate the prices of the order lines that contain a product
```

sale_order_price_recalculation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+* Reset product descriptions from current product information.
```
simahawk, line 4:
```suggestion
* Reset product descriptions using current product information
```

sale_order_price_recalculation/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* In a sale order with lot of lines, the recalculation may slow down, because
```
simahawk, line 1:
```suggestion
* In a sale order with a lot of lines, the recalculation may slow down, because
```

sale_order_price_recalculation/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
...
+  sale general data (amount untaxed, amount taxed...) are recalculated for
+  each line.
```
simahawk, line 3:
```suggestion
  each line. We could think about using queue.job to do it asynchronously.
```

simahawk (changes requested):

sergio-teruel:
@rousseldenis Can you do a rebase to fix conflict???

sale_order_price_recalculation/tests/test_sale_order_price_recalculation.py
```diff
@@ -0,0 +1,103 @@
...
+        self.sale_order.recalculate_prices()
+        self.assertEqual(self.sale_order_line.discount, 50.0)
+
+    def test_name_recalculation(self):
+        self.sale_order_line.price_unit = 150.0
+        initial_price = self.sale_order_line.price_unit
+        self.assertEqual(self.sale_order_line.name, self.product.name)
+        self.sale_order_line.name = "Custom Jacket"
+        self.sale_order.recalculate_names()
+        self.assertEqual("Jacket - Color: Black - Size: XL", self.sale_order_line.name)
```
sergio-teruel, line 101: @rousseldenis Take into account that if the product has a default_code the name would be "[00001]......" so the test fails...
If product_sequence is installed that always creates a product ref this test fails
Maybe the best option is do the test with a product with default_code

rousseldenis: @sergio-teruel I didn't even touched to tests because for a migration, tests should pass as is.

pedrobaeza: @rousseldenis what my colleague is talking about is the resilience of the tests when combined with other OCA modules, as our CI reveals. In this case, combining this module with `product_sequence`. Indeed, the tests weren't resilient before and it's not your fault. But the proposal is to change it for that. I propose an even more resilient option, as if you have the module `sale_line_description`, Sergio's proposal will fail as well:

```suggestion
        self.assertIn("Jacket - Color: Black - Size: XL", self.sale_order_line.name)
```

pedrobaeza: Or even a better one:
```suggestion
        self.assertNotEqual("Custom Jacket", self.sale_order_line.name)
```
the test is satisfied as it checks that the description has been recomputed, no matter to what value.

rousseldenis: Ok agree with proposal, makes sense. I need time to get back on this, but if you want to make a PR on my branch, don't hesitate.

pedrobaeza: I can supersede the PR if you want, but would you accept a PR removing the dependency on `onchange_helper` and switching to `Form`?

rousseldenis: Yes, do it.

pedrobaeza:
Superseded by #1239

PR #1100, opened 2020-03-23
[12.0][IMP] sale_order_product_recommendation: Mobile views

pedrobaeza:
Add kanban + form for proper mobile usage.

Selección_036

Selección_037

cc @Tecnativa TT21495

PR #1157, opened 2020-06-02
[13.0] sale_order_product_recommendation: Forward port of #1100

pedrobaeza:
Forward-port of #1100 

- Mobile views: Add kanban + form for proper mobile usage.
- Remove useless flag `is_modified`: It's better to simply check the flag condition than to keep that field.
- Don't tie second. unit to product default one: Having a related field to default sale secondary UoM makes this recommendator to not fit possible different secondary UoM to be used. This commit fixes this behavior. A refactoring on the main module having a prepare method hook has been needed.
- Use Form for populating onchanges: Previous code was not very resilient to possible extra adaptations or require specific code to be added to the old hook `_trigger_so_line_onchanges`. With this, we make sure all onchanges are executed thanks to the odoo.tests.Form class. We lose a bit of performance, but it's better than to have issues with data not correctly synchronized.

cc @Tecnativa TT21495

### DIFF
--- a/sale_order_product_recommendation/README.rst
+++ b/sale_order_product_recommendation/README.rst
@@ -73,6 +73,7 @@ Contributors
   * Jairo Llopis
   * David Vidal
   * Alexandre Díaz
+  * Pedro M. Baeza
 
 Maintainers
 ~~~~~~~~~~~
--- a/sale_order_product_recommendation/__manifest__.py
+++ b/sale_order_product_recommendation/__manifest__.py
@@ -1,4 +1,5 @@
 # Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Sale Order Product Recommendation",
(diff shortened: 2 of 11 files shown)


sergio-teruel (approved):

HaraldPanten:
> * Don't tie second. unit to product default one: Having a related field to default sale secondary UoM makes this recommendator to not fit possible different secondary UoM to be used. This commit fixes this behavior. A refactoring on the main module having a prepare method hook has been needed.

Hi @pedrobaeza , if I understood well, you removed the default secondary unit relation in wizard mobile view, to make it fit properly, right? 

With this performance, user has to set manually the secondary unit in the sale order lines instead of the wizard. Is that what you did? Thx

pedrobaeza:
Well, that's the same patch as in v12. In a nutshell:

- You can have several secondary units per product, and one or none default sale secondary unit.
- On sales order line, you can put any of these secondary UoM, defaulting to the other.
- When opening the recommendation wizard, you lost before this commit that secondary UoM if different from the default one.

HaraldPanten (approved):
> Well, that's the same patch as in v12. In a nutshell:
> 
> * You can have several secondary units per product, and one or none default sale secondary unit.
> * On sales order line, you can put any of these secondary UoM, defaulting to the other.
> * When opening the recommendation wizard, you lost before this commit that secondary UoM if different from the default one.

OK! 👍

chienandalu (approved):
Code and functional review :+1:

PR #1167, opened 2020-06-16
[12.0] Add sale_automatic_workflow_job

guewen:
Use Queue Jobs to process the Sales Automatic Workflow actions.

The default behavior of the automatic workflow module is to use a
scheduled action that searches all the record that need a workflow
action and sequentially process all of them.

It can hit some limits when the number of records is too high.

This module keeps the scheduled action to search the records, but
instead of directly executing the actions (confirm a sales order,
create invoices for a sales order, validate invoices, ...), it
creates one job per operation to do.

It uses an identity key on the jobs so it will not create the same
job for the same record and same operation twice.

~

I needed to extract methods in `sale_automatic_workflow` in order
to be able to execute them as jobs.

A new decorator "job_auto_delay" (to eventually move in queue_job) makes
these methods automatically delayed when called.

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -7,3 +7,4 @@ server-ux
 web
 bank-payment
 stock-logistics-warehouse
+queue
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -44,15 +44,25 @@ class AutomaticWorkflowJob(models.Model):
         ' invoices, pickings...'
     )
 
+    def _do_validate_sale_order(self, sale):
+        """Validate a sales order"""
+        with savepoint(self.env.cr), force_company(self.env, sale.company_id):
+            sale.action_confirm()
+
     @api.model
     def _validate_sale_orders(self, order_filter):
         sale_obj = self.env['sale.order']
         sales = sale_obj.search(order_filter)
         _logger.debug('Sale Orders to validate: %s', sales.ids)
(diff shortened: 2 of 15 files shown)


guewen:
Travis status is not refreshed, but the build is green...

rousseldenis:
@guewen To do this easily, you could use https://github.com/OCA/queue/tree/13.0/queue_job_cron

Just enable feature on automatic workflow cron and that's it ! In production for months with lots of sales.

But indeed, if you want to split per operation, this can be an alternative

guewen:
Hi @rousseldenis, yes I advertise (and use of course) this module when I can. But it's not the same thing here. In automatic workflow there is only one big cron that:

* search all sales orders matching the domain "sales to confirm", and confirm them
* search all sales matching the domain "invoices to generate" and create them
* search all invoices matching the domain "invoices to validate" and validate them
* etc

If you activate the "run as job" option, you have some improvements, such as you see if the job fails for instance. But still, it is one big transaction which can last dozens of minutes.

This module keeps the cron, which can be run as job or not, but the behavior of the cron is now:


* search all sales orders matching the domain "sales to confirm", and create jobs to confirm them
* search all sales matching the domain "invoices to generate" and create jobs to create them
* search all invoices matching the domain "invoices to validate" and create jobs to validate them

So the cron itself is short, every action is done in its own transaction, and it's much easier to trace why a sales order was not validated for instance.

gurneyalex (approved):

rousseldenis (approved):
LGTM

PR #1168, opened 2020-06-16
[10.0][IMP] sale_triple_discount: Improve compute_amount perf

lmignon:
Call the preprocess only once before the loop in order to generate one query for each line processed when caching the original values
@gurneyalex @tafaRU

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -55,10 +55,12 @@ class SaleOrderLine(models.Model):
         # case the work done in preprocess would be lost as soon as the 1st
         # line is updated, causing the following lines to ignore the discount2
         # and discount3 fields.
+        prev_values = self._get_triple_discount_preprocess_values()
+        to_process_values = self._get_triple_discount_process_values()
         for line in self:
-            prev_values = line.triple_discount_preprocess()
+            line._cache.update(to_process_values[line])
             super(SaleOrderLine, line)._compute_amount()
-            line.triple_discount_postprocess(prev_values)
+            line.triple_discount_postprocess({line: prev_values[line]})
 
     discount2 = fields.Float(
         'Disc. 2 (%)',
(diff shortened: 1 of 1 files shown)


jbaudoux:
@lmignon In the comment block above this for loop, it is explicitly written that you cannot do this change and why you cannot do this.

lmignon:
@jbaudoux The sensitive part of the code is to restore the values collected during the pre-process on each loop. With this change we collect all the values at once and we restore these values after the compute on each line. I could may be improve the code by avoiding to restore all the lines on each loop but only the one of the loop. If we collect the values on each loop, we'll generate 1 SQL query by line since the cache is invalidated line by line and therefore values read line by line....

jbaudoux:
@lmignon When reading the comment, I understood that the sensitive part is this line https://github.com/acsone/sale-workflow/blob/10.0-sale_triple_discount_perf-lmi/sale_triple_discount/models/sale_order_line.py#L133 and that the cache of all the lines set there can be reset on the call of https://github.com/acsone/sale-workflow/blob/10.0-sale_triple_discount_perf-lmi/sale_triple_discount/models/sale_order_line.py#L60 on the first line. So you have to reset the cache of the next line before calling https://github.com/acsone/sale-workflow/blob/10.0-sale_triple_discount_perf-lmi/sale_triple_discount/models/sale_order_line.py#L60 on the next line. This is how I interpret the comment and I may be wrong as I haven't tested.

lmignon:
@jbaudoux You're right. I modified the code by adding two methods. The first collects the current field values. The second collects the values needed for the `compute_amount` for these fields.
From there, in the iteration for each loop, I set the cache with the values needed for the `compute_amount` method and reset it after the call to the method with the original values. 
I am very, very surprised to see that the tests all passed with the previous change that was wrong. So there's some highly sensitive code in this module that's untested. Also, I can't understand why, we don't have a problem with the [`amount_all`](https://github.com/acsone/sale-workflow/blob/10.0-sale_triple_discount_perf-lmi/sale_triple_discount/models/sale_order.py#L13) method on sale.order...

lmignon:
ping @jbaudoux

gurneyalex:
Just commenting here, because, although the code of the PR looks ok to me, so was the version that I had to fix during a horrible debugging night, under pressure. I think we all know which project we are talking about here... 

You probably checked https://github.com/OCA/sale-workflow/commit/d1ee7c6db92276eeb8bf47d22b1e0c6804b1e7ac The cache invalidation does not happen every time, and I could not find the root cause of this. So my advice is test test test test and test again, on this well known project, and get wild with the tests combinations, and various loads. I remember running the same computation and tests in a loop, to get 1 failure after 100 test runs. I could not get my patched version to fail after 10000 runs and decided I was happy with this (and I needed sleep). 

Rant mode:
I am really unhappy with the implementation of this module and I wish we had discount1, discount2 and discount3 as new field, and that changing them would update the main discount field which would be hidden from the user. And things would be sooo simpler and easier, without resorting to dirty tricks such as writing pseudo values in the cache.

lmignon:
Thank you for your review @gurneyalex 
I share your feeling about this ugly implementation. 
The purpose of the proposed changes is to ensure that the cache remains consistent between each loop while collecting the original data from the cache and the data needed for the right computation by `compute_amount` in batch. Without these changes, you end up with at least 1 SQL query per loop. And as you know, on a project we know well, any performance gain is very important to take.

gurneyalex (approved):

jbaudoux (approved):

rousseldenis:
@lmignon Ready ?

PR #1171, opened 2020-06-22
[12.0][IMP] sales_team_security: Contacts permissions revamp

pedrobaeza:
Restrict contacts by code, removing the need of manually manipulating the existing record rules, that can also have some side effects.

@Tecnativa TT24290

### DIFF
--- a/sales_team_security/models/ir_rule.py
+++ b/sales_team_security/models/ir_rule.py
@@ -1,0 +1,55 @@
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import api, models, tools, SUPERUSER_ID
+from odoo.osv import expression
+from odoo.tools import config
+
+
+class IrRule(models.Model):
+    _inherit = 'ir.rule'
+
+    @api.model
+    @tools.conditional(
+        'xml' not in config['dev_mode'],
+        tools.ormcache(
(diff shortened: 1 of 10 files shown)


sergio-teruel (approved):
:+1: Code review... ok form me. Waiting travis

HaraldPanten (approved):
Seems to be OK 👍

sales_team_security/models/ir_rule.py
```diff
@@ -0,0 +1,55 @@
...
+    def _compute_domain(self, model_name, mode="read"):
+        """Inject extra domain for restricting partners when the user
+        has the group 'Sales / User: Own Documents Only'.
+        """
+        res = super()._compute_domain(model_name, mode=mode)
+        user = self.env.user
+        group1 = "sales_team.group_sale_salesman"
+        group2 = "sales_team_security.group_sale_team_manager"
+        group3 = "sales_team.group_sale_salesman_all_leads"
+        if model_name == "res.partner":
```
yajo: :thinking: If you're affecting only the `res.partner` model, don't you think `res.partner.check_access_rule()` would be the appropriate hook to override?

pedrobaeza: That method only serves for direct operations over the record, but for example it doesn't limit the number or records returned by a `search`. But this low level method is call for all of them.

yajo: Interesting, today I learned that.

yajo (approved):

PR #1170, opened 2020-06-19
[12.0] [ADD] sale_product_email module

chafique-delli:

PR #1172, opened 2020-06-22
[12.0][FIX] sale_order_generator: Fix view

rousseldenis:
Due to https://github.com/OCA/sale-workflow/commit/ea6fceb6aceb6f2c9ee146c09b9e87176e53d0ae commited directly in 12.0 branch.

See error: https://github.com/OCA/sale-workflow/pull/1170#issuecomment-647414530

@pedrobaeza You cannot use that kind of domain in a view.

### DIFF
--- a/sale_generator/views/sale_order_view.xml
+++ b/sale_generator/views/sale_order_view.xml
@@ -6,7 +6,7 @@
         <field name="inherit_id" ref="sale_management.sale_order_form_quote"/>
         <field name="arch" type="xml">
             <xpath expr="//field[@name='sale_order_template_id']" position="attributes">
-                <attribute name="attrs">{'invisible': [(1, '=', 1)]}</attribute>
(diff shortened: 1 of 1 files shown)


rousseldenis:
To be tested in runbot before merge

pedrobaeza:
Why? It was not giving any error, and the previous one was incorrect and making to fail whole branch.

pedrobaeza:
Take into account that it was overwriting other `attrs`, so I think your solution is not going to work

rousseldenis:
> Take into account that it was overwriting other `attrs`, so I think your solution is not going to work

attrs is not present in base module, so this is correct.

Put your debug mode in runbot, you'll see

pedrobaeza:
OK then

HaraldPanten (approved):
LGTM 👍

PR #1174, opened 2020-06-24
[IMP][sale_order_line_packaging_qty] Sell product / unit or packaging

Tonow-c2c:
**Context:**

We want to be able to only sell by packaging some product (forbidden to
 sell by unit). Also, we want to choose which packaging is eligible to
 sell.

**Usage**

When I encode a sale order line, I must fill up the packaging and
related quantity if the product is set as "Only sell by packaging".
I cannot enter a quantity in the standard Odoo field quantity that
is not a multiple of a sellable packaging
(product.packaging can be sold = True).

Is the product is not configured as "Only sell by packaging",
then I am allowed to fill up a line with no packaging.

On every packaging I can chose if it is sellable or not. By default it is.

I can chose at product level if i restrict the quantity to sell
to existing sellable packaging.

**Dev**

* Add a check box in product "Only sell by packaging multiple" (default False), aside of "Can be sold" in the header
* Add a boolean on the packaging "Can be sold" (default True)
* On the list of available packaging in a SO line, only display packaging with "can be sold = True"
* When a SO line is encoded *with a product that can only be sold by packaging*:
    * The quantity field cannot contain a quantity that is not a multiple of a sellable packaging (can be sold = True) -> raise a warning (triggered on the on_change please, not create/write as we still want to be able to save it e.g. in case of an EDI order) -> "The quantity is not a multiple of a defined sellable packaging. Please complete the packaging first."
    * On create / write method, *if packaging and packaging quantity is empty*, fulfill the info if a multiple of an existing sellable packaging (can be sold = True) is found
        * e.g. if my product is sold by packaging only and I have packaging of box: 10, Karton: 200, and pallet:1000, if the quantity entered is 400, I expect the system to complete the packaging with Karton and packaging quantity with 2
        * e.g. if my product is sold by packaging only and I have packaging of box: 10, Karton: 200, and pallet:1000, if the quantity entered is 420, I expect the system to complete the packaging with box and packaging quantity with 42
        * e.g.. if my product is sold by packaging only and I have packaging of box: 10, Karton: 200, and pallet:1000, if the quantity entered is 405, I expect the system to leave it empty
    * Make the message "You must define a package before setting a quantity of said package." a user warning and not a blocking point AND only raise this warning for product that have "Only sell by packaging multiple = True"

-----

OCA 2020-06-25 08-40

### DIFF
--- a/sale_order_line_packaging_qty/models/product_product.py
+++ b/sale_order_line_packaging_qty/models/product_product.py
@@ -1,0 +1,21 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _which_pack_multiple(self, qty):
+        """ Return multiple of product packaging for one quantity if exist.
+        """
+        self.ensure_one()
+        pack_multiple = False
+        if qty:
+            for pack in self.packaging_ids.sorted("qty", reverse=True):
+                if pack.packaging_type_id.can_be_sold and pack.qty:
+                    if (qty % pack.qty) == 0:
+                        pack_multiple = pack
+                        break
+        return pack_multiple
--- a/sale_order_line_packaging_qty/models/product_template.py
+++ b/sale_order_line_packaging_qty/models/product_template.py
@@ -1,0 +1,11 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    sell_only_by_packaging = fields.Boolean(
+        string="Only sell by packaging", default=False,
+    )
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -37,6 +37,7 @@ class SaleOrderLine(models.Model):
             sol.product_packaging_qty = product_qty / sol.product_packaging.qty
 
     def _prepare_product_packaging_qty_values(self):
+        self.ensure_one()
         return {
             "product_uom_qty": self.product_packaging.qty * self.product_packaging_qty,
             "product_uom": self.product_packaging.product_uom_id.id,
@@ -86,5 +87,60 @@ class SaleOrderLine(models.Model):
         # TODO Drop once https://github.com/odoo/odoo/pull/49150/ is merged
         res = super()._onchange_product_uom_qty()
         if not res:
-            res = self._check_package()
+            res = self._check_package() or self._check_qty_is_pack_multiple()
         return res
+
+    @api.onchange("product_id")
+    def product_id_change(self):
+        res = super().product_id_change()
+        if self.product_id.sell_only_by_packaging:
+            self.product_uom_qty = min(self.product_id.packaging_ids.mapped("qty"))
+        return res
+
+    def _check_qty_is_pack_multiple(self):
+        """ Check only for product with sell_only_by_packaging
+        """
+        # and we dont want to have this warning when we had the product
+        if self.product_id.sell_only_by_packaging:
+            if not self._is_pack_multiple():
+                warning_msg = {
+                    "title": _("Product quantity can not be packed"),
+                    "message": _(
+                        "For the product {prod}\n"
+                        "The {qty} is not the multiple of any pack.\n"
+                        "Please add a package"
+                    ).format(prod=self.product_id.name, qty=self.product_uom_qty),
+                }
+                return {"warning": warning_msg}
+        return {}
+
+    def _is_pack_multiple(self):
+        return bool(self.product_id._which_pack_multiple(self.product_uom_qty))
+
+    def write(self, vals):
+        # Fill the packaging if they are empty and the quantity is a multiple
+        for line in self:
+            product_uom_qty = vals.get("product_uom_qty")
+            product_packaging = vals.get("product_packaging")
+            if line.product_id.sell_only_by_packaging and (
+                not line.product_packaging
+                or ("product_packaging" in vals and not product_packaging)
+            ):
+                pack_multiple = line.product_id._which_pack_multiple(product_uom_qty)
+                if pack_multiple:
+                    vals.update({"product_packaging": pack_multiple.id})
+        return super().write(vals)
+
+    @api.model
+    def create(self, vals):
+
+        # Fill the packaging if they are empty and the quantity is a multiple
+        product = self.env["product.product"].browse(vals.get("product_id"))
+        product_uom_qty = vals.get("product_uom_qty")
+        product_packaging = vals.get("product_packaging")
+
+        if product and product.sell_only_by_packaging and not product_packaging:
+            pack_multiple = product._which_pack_multiple(product_uom_qty)
+            if pack_multiple:
+                vals.update({"product_packaging": pack_multiple.id})
+        return super().create(vals)
--- a/sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
+++ b/sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
@@ -30,3 +30,98 @@ class TestSaleOrderLinePackagingQty(SavepointCase):
         self.assertEqual(order_line.product_packaging_qty, 1.0)
         order_line.write({"product_packaging_qty": 3.0})
         self.assertEqual(order_line.product_uom_qty, 15.0)
+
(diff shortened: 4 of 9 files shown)


sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
```diff
@@ -30,3 +30,58 @@ def test_product_packaging_qty(self):
...
+        order_line.product_id_change()
+        self.assertFalse(order_line._onchange_product_uom_qty())
+
+        order_line.write({"product_uom_qty": 3.0})
+        self.assertTrue(order_line._onchange_product_uom_qty())
+
+        order_line.write({"product_uom_qty": self.packaging.qty * 2})
+        self.assertFalse(order_line._onchange_product_uom_qty())
+
+    # def test_auto_fill_packaging(self):
```
rousseldenis: Remove what you don't use

rousseldenis:
Makes sense to me

sale_order_line_packaging_qty/tests/test_sale_order_line_packaging_qty.py
Tonow-c2c: I will fix it's it's reason why I added on WIP on the description :wink:

sale_order_line_packaging_qty/models/product_product.py
```diff
@@ -0,0 +1,18 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
i-vyshnevska, line 2: copyrights

sale_order_line_packaging_qty/models/product_product.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _which_pack_multiple(self, qty):
```
i-vyshnevska, line 10: a comment?

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+        is_pack_multiple = False
+        for pack in self.product_id.packaging_ids:
+            if pack.packaging_type_id.can_be_sold and pack.qty:
+                if (self.product_uom_qty % pack.qty) == 0:
+                    is_pack_multiple = True
+                    break
+        return is_pack_multiple
+
+    def write(self, vals):
+        # Fill the packaging if they is empty and the quantity is a multiple
```
i-vyshnevska:
```suggestion
        # Fill the packaging if they are empty and the quantity is a multiple
```

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+            ):
+                pack_multiple = line.product_id._which_pack_multiple(product_uom_qty)
+                if pack_multiple:
+                    vals.update({"product_packaging": pack_multiple.id})
+        return super().write(vals)
+
+    @api.model
+    def create(self, vals):
+
+        # Fill the packaging if they is empty and the quantity is a multiple
```
i-vyshnevska:
```suggestion
        # Fill the packaging if they are empty and the quantity is a multiple
```

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+    def product_id_change(self):
+        res = super().product_id_change()
+        if self.product_id.only_sell_by_pack:
+            self.product_uom_qty = min(self.product_id.packaging_ids.mapped("qty"))
+        return res
+
+    def _check_qty_is_pack_multiple(self):
+        """ Check only for product with only_sell_by_pack
+        """
+        # self.product_uom_qty != 1 cause is the defaut value
```
TDu: Is this a todo that should be implemented ?

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+                        "Please add a package".format(
+                            prod=self.product_id.name, qty=self.product_uom_qty
+                        )
+                    ),
+                }
+                return {"warning": warning_mess}
+        return {}
+
+    def _is_pack_multiple(self):
+        is_pack_multiple = False
```
TDu: This function is really closed to `def _which_pack_multiple(self, qty)` on `product.product` 
I wonder if you could just do something like this
```suggestion
        return bool(self.product_id._which_pack_multiple(self.product_uom_qty))
```

sale_order_line_packaging_qty/models/product_product.py
simahawk: a docstring! :)

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+        # self.product_uom_qty != 1 cause is the defaut value
+        # and we dont want to have this warning when we had the product
+        if self.product_id.only_sell_by_pack:
+            if not self._is_pack_multiple():
+                warning_mess = {
+                    "title": _("Product quantity can not be packed"),
+                    "message": _(
+                        "For the product {prod}\n"
+                        "The {qty} is not the multiple of any pack.\n"
+                        "Please add a package".format(
```
simahawk: translation broken if you format it on the string. Use `format` on the gettext object: `_(...).format(...)`

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,69 @@ def _onchange_product_uom_qty(self):
...
+        return res
+
+    def _check_qty_is_pack_multiple(self):
+        """ Check only for product with only_sell_by_pack
+        """
+        # self.product_uom_qty != 1 cause is the defaut value
+        # and we dont want to have this warning when we had the product
+        if self.product_id.only_sell_by_pack:
+            if not self._is_pack_multiple():
+                warning_mess = {
```
simahawk: "mess" != "msg" :smile:

sale_order_line_packaging_qty/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    only_sell_by_pack = fields.Boolean(string="Only sell by packaging", default=False,)
```
simahawk:
```suggestion
    sell_only_by_packaging = fields.Boolean(string="Only sell by packaging", default=False,)
```

simahawk (changes requested):

jgrandguillaume (approved):
Unless you fix what @simahawk asked I'm good, thx !

sale_order_line_packaging_qty/models/sale_order_line.py
Tonow-c2c: yes should be remove it's was comment for me in the dev phase sorry for the noise

Tonow-c2c:
```suggestion
```

simahawk:
@Tonow-c2c thanks! Please squash! :)

simahawk (approved):

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -86,5 +87,60 @@ def _onchange_product_uom_qty(self):
...
+        for line in self:
+            product_uom_qty = vals.get("product_uom_qty")
+            product_packaging = vals.get("product_packaging")
+            if line.product_id.sell_only_by_packaging and (
+                not line.product_packaging
+                or ("product_packaging" in vals and not product_packaging)
+            ):
+                pack_multiple = line.product_id._which_pack_multiple(product_uom_qty)
+                if pack_multiple:
+                    vals.update({"product_packaging": pack_multiple.id})
```
grindtildeath, line 131: How could this go unnoticed? :astonished: 

1. If `product_id` is defined in `vals` accessing `line.product_id` to select a packaging would use packaging of a product that is not the one written on the lines.
2. If `product_id` is not defined in `vals`, the for loop on the lines will redefine `product_packaging` on each iteration that finds a packaging, and the value written on all the lines will be the last packaging to be found.

grindtildeath (commented):
I have two issues with this PR:
- Mix of multiple features into the same module
- Buggy code (cf below)

That's why I did open https://github.com/OCA/sale-workflow/pull/1191 to revert this.

PR #1176, opened 2020-06-29
sale_order_type: No type changes once validated

houzefa-abba:
This follows other sales order / invoice fields which switch to
read-only once validated.


houzefa-abba:
Hmm as I expected tests failed but that's not due to these changes.

The last changes on <https://github.com/OCA/sale-workflow/commits/11.0> were on translations, some passed, some failed.

I ran tests of this `sale_order_type` module in my local DB to make sure; they do work fine.

PR #1177, opened 2020-06-30
[11.0][FIX] sale_global_discount: wrong property name

chienandalu:
cc @Tecnativa TT24526


pedrobaeza (approved):

pedrobaeza:
Fixed error upstream, so let's fast-track this:

/ocabot merge patch

PR #1178, opened 2020-07-02
[FIX][sale_order_line_packaging_qty] Prevent product.packaging without quantity case

Tonow-c2c:
Issue to fix by this PR

* Flag a product as "Only sell by packaging"
* Ensure there is no defined packaging for this product
* Create a SO and add a line you have a crash

```python
File "/odoo/external-src/sale-workflow/sale_order_line_packaging_qty/models/sale_order_line.py", line 97, in product_id_change
self.product_uom_qty = min(self.product_id.packaging_ids.mapped("qty"))
ValueError: min() arg is an empty sequence
```

### DIFF
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -94,7 +94,9 @@ class SaleOrderLine(models.Model):
     def product_id_change(self):
         res = super().product_id_change()
         if self.product_id.sell_only_by_packaging:
-            self.product_uom_qty = min(self.product_id.packaging_ids.mapped("qty"))
+            quantites = self.product_id.packaging_ids.mapped("qty")
+            if quantites:
+                self.product_uom_qty = min(quantites)
         return res
 
     def _check_qty_is_pack_multiple(self):


jgrandguillaume (approved):

jbaudoux (approved):

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -94,7 +94,9 @@ def _onchange_product_uom_qty(self):
     def product_id_change(self):
         res = super().product_id_change()
         if self.product_id.sell_only_by_packaging:
-            self.product_uom_qty = min(self.product_id.packaging_ids.mapped("qty"))
+            quantites = self.product_id.packaging_ids.mapped("qty")
+            if quantites:
+                self.product_uom_qty = min(quantites)
```
grindtildeath, line 99:
```suggestion
            quantities = self.product_id.packaging_ids.mapped("qty")
            if quantities:
                self.product_uom_qty = min(quantities)
```

grindtildeath (changes requested):
I find this fix a bit illogical. 
If we want to restrict the sale of a product with its packaging, IMO an error should be raised when the flag is set on a product and no packaging is defined.

Moreover, I tend to think the restriction of selling by packaging should have been done in a separated module to avoid a mix of features in this one that had a well defined perimeter...
cc @guewen  @jgrandguillaume

guewen:
Why the option to force using a packaging on a sale line is in the module that adds the package qty on lines? Seems unrelated?

grindtildeath:
This PR shouldn't be needed after merging #1191 especially considering https://github.com/OCA/sale-workflow/pull/1174#pullrequestreview-447241590

simahawk:
closing in favour of https://github.com/OCA/sale-workflow/pull/1191

PR #1181, opened 2020-07-03
[11.0][FIX] sale_global_discount: Constraint tax lines combinations

pedrobaeza:
Not all taxes combinations are possible with global discounts when they are later invoiced and posted, so we need to constraint them early.

OCA/account-invoicing#745 addresses the problem at invoice level, posting correctly taxes for supported cases, and constraining the rest.

We apply here the same constraints at sales order level. Existing tests have been changed because they were using one of the forbidden taxes combination, and 2 new tests have been added for covering the new constraints.

@Tecnativa TT24526

### DIFF
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import api, fields, models
+from odoo import _, api, exceptions, fields, models
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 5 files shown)


sale_global_discount/models/sale_order.py
```diff
@@ -42,10 +42,34 @@ def get_discounted_global(self, price=0, discounts=None):
...
+        taxes_keys = {}
+        for line in self.order_line:
+            if not line.tax_id:
+                raise exceptions.UserError(_(
+                    "With global discounts, taxes in lines are required."
+                ))
+            for key in taxes_keys:
+                if key == line.tax_id:
+                    break
+                elif key & line.tax_id:
```
chienandalu, line 61: why is this bitwise operation needed to perform the check?

pedrobaeza: You have approved the same operation in the other PR :stuck_out_tongue: 

The idea is to constraint non valid taxes combinations that will lead to incorrect global discount. Example:

- Line 1: tax 1 and tax 2
- Line 2: tax 2

As (tax 1, tax 2) is not the same as (tax 2), first condition doesn't break the loop, but the overlapping of both (the `&` operation) gives you a common tax, raising the constraint.

chienandalu: Interesting technique :)

chienandalu (approved):

rafaelbn (approved):
:+1:

PR #1188, opened 2020-07-08
[ENH] sale_order_type, better work with multi-company

watcharaporn-git:


kittiu:
@watcharaporn-git 
1. Use same review as [here](https://github.com/OCA/manufacture/pull/499/commits/c2568f98d31c03e3ca44b74d99dbc1ecd02a482f)
2. Make sure Travis is green https://travis-ci.org/github/OCA/sale-workflow/jobs/706193377 (please use pre-commit)

kittiu:
Note: you can always refer to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 for migration
and for coding standard, https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

sale_order_type/models/sale.py
```diff
@@ -73,6 +74,16 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    @api.constrains('company_id', 'type_id.company_id')
```
sswapnesh:
```suggestion
    @api.constrains('company_id', 'type_id')
```
https://github.com/odoo/odoo/blob/c89b4c25ed92542f6fd0279cf3a9e4ef1bac9fa5/odoo/api.py#L119

sale_order_type/models/sale.py
```diff
@@ -13,6 +13,7 @@ class SaleOrder(models.Model):
         compute="_compute_sale_type_id",
         readonly=False,
         store=True,
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
```
sswapnesh:
```suggestion
        domain="[('company_id', 'in', [False, company_id])]",
```

sale_order_type/models/sale.py
```diff
@@ -73,6 +74,16 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    @api.constrains('company_id', 'type_id.company_id')
+    def _constrains_company(self):
+        if self.company_id not in self.type_id.company_id:
+            raise exceptions.ValidationError("Error : Company and Type No Matching !!!")
+
+    @api.onchange('company_id', 'type_id.company_id')
+    def check_company(self):
```
sswapnesh:
```suggestion
    def _check_company(self):
```

and also it should be `@api.onchange('company_id')`

kittiu: I would add to change the method name to `_check_company_order_type` to be more understandable.

kittiu:
Ready for review,
cc @sswapnesh @ps-tubtim @watcharaporn-git

sale_order_type/models/sale.py
```diff
@@ -73,6 +74,18 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    @api.constrains("company_id")
+    def _check_so_type_company(self):
+        for record in self.filtered("type_id.company_id"):
```
sswapnesh:
```suggestion
        if self.filtered(lambda r: r.type_id.company_id != r.company_id):
```
Maybe something like this

kittiu: My point also that, if user do not select type or type has no company specified, we don't want to check.
On your point, it still check, isn't it?

sswapnesh: You are right, Maybe in that case we can check something like this


```suggestion
        if self.filtered(lambda r: r.type_id.company_id and r.company_id and r.type_id.company_id != r.company_id):
```

kittiu: Done, thank you!

sswapnesh (approved):

ps-tubtim (approved):

marcelsavegnago (approved):

newtratip:
@watcharaporn-git Can you check the conflict file ?

sale_order_type/models/sale.py
```diff
@@ -12,6 +12,7 @@ class SaleOrder(models.Model):
         string="Type",
         compute="_compute_sale_type_id",
         store=True,
+        domain="[('company_id', 'in', [False, company_id])]",
```
ivantodorovich, line 15: Why not using the new `check_company=True` ?

https://www.odoo.com/documentation/13.0/howtos/company.html#multi-company-consistency

sale_order_type/models/sale.py
```diff
@@ -74,6 +75,19 @@ def _prepare_invoice(self):
...
         return res
 
+    @api.constrains("company_id")
+    def _check_so_type_company(self):
+        if self.filtered(
+            lambda r: r.type_id.company_id
+            and r.company_id
+            and r.type_id.company_id != r.company_id
+        ):
+            raise ValidationError(_("Document's company and type's company mismatch"))
```
ivantodorovich, line 85:
```suggestion
```

If you use `check_company=True`, this is not needed.

https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L3132

sale_order_type/models/sale.py
```diff
@@ -74,6 +75,19 @@ def _prepare_invoice(self):
...
+
+    @api.onchange("company_id")
+    def _onchange_company(self):
+        self.type_id = False
+
```
ivantodorovich, line 90:
```suggestion
```

`_compute_sale_type_id` already depends on `company_id`, and it's supposed to do a better job (setting the sale_type of the partner)

Why would we need this onchange?

ivantodorovich (changes requested):

pedrobaeza:
See #1193 for changes in the computation of the sale type.

kittiu:
> See #1193 for changes in the computation of the sale type.

TODO @kittiu

kittiu:
Note: https://github.com/OCA/sale-workflow/pull/1203 something on v12 to lean on. @kittiu

PR #1190, opened 2020-07-10
[13.0][IMP][sale_order_type] Sale Order Type: Fix view and readonly

ivantodorovich:
Hello,

This PR addresses two topics:

### 1. Fixes the sale_type_id field in account.move (Invoice) form view.

Before:
befoer

After:
image

It also makes sure the field is only shown on actual invoices, and not on other account.moves

------

### 2. Makes the sale_type_id field in both sale.order and account.move models readonly depending on the state, to be consistent with Odoo workflows.

Before, it was possible to change the sale_type_id in Invoices that where already posted, for instance. 

That's undesireable because the sale.order.type may try to change the journal, for instance.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -11,8 +11,9 @@ class AccountMove(models.Model):
         comodel_name="sale.order.type",
         string="Sale Type",
         compute="_compute_sale_type_id",
-        readonly=False,
         store=True,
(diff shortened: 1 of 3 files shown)


rousseldenis (approved):
LGTM

victoralmau (approved):

pedrobaeza (approved):

PR #1191, opened 2020-07-13
[13.0] split module sale_order_line_packaging_qty

grindtildeath:
PR #1174 did introduce multiple features into the same module (`sale_order_line_packaging_qty`) which had a well defined name and scope.
This PR aims to split the different features that were merged in `sale_order_line_packaging_qty` into a different module `sale_by_packaging` to provide well packaged features among modules.

- sale_order_line_packaging_qty

This module adds a "Package quantity" field on sale order line in order to
define a quantity according to the selected "Package".

- sale_by_packaging

This module provides different configuration option to manage packagings on
sale orders.

Following options are available to define which packaging type can be sold and
which product can only be sold by packaging.

* Can be sold: On product packaging type model, this checkbox defines if product
  packagings from this particular type are available to be selected on sale
  order line.

* Sell only by packaging: On product template model, this checkbox restricts
  sales of these products if no packaging is selected on the sale order line.
  If no packaging is selected, it will either be auto-assigned if the quantity
  on the sale order line matches a packaging quantity or an error will be raised.


sale_product_by_packaging_only/models/sale_order_line.py
```diff
@@ -0,0 +1,92 @@
...
+        for line in self:
+            product_uom_qty = vals.get("product_uom_qty")
+            product_packaging = vals.get("product_packaging")
+            if line.product_id.sell_only_by_packaging and (
+                not line.product_packaging
+                or ("product_packaging" in vals and not product_packaging)
+            ):
+                pack_multiple = line.product_id._which_pack_multiple(product_uom_qty)
+                if pack_multiple:
+                    vals.update({"product_packaging": pack_multiple.id})
```
guewen: Would be worth to extract the behavior added in `write` in another method: easier to customize and extend this behavior if it's in another method compared to being in the middle of a write chain.

sale_product_by_packaging_only/models/sale_order_line.py
```diff
@@ -0,0 +1,92 @@
...
+        # Fill the packaging if they are empty and the quantity is a multiple
+        product = self.env["product.product"].browse(vals.get("product_id"))
+        product_uom_qty = vals.get("product_uom_qty")
+        # TODO Consider UOM
+        product_packaging = vals.get("product_packaging")
+
+        if product and product.sell_only_by_packaging and not product_packaging:
+            pack_multiple = product._which_pack_multiple(product_uom_qty)
+            if pack_multiple:
+                vals.update({"product_packaging": pack_multiple.id})
```
guewen: same remark as for the write above

grindtildeath:
ping @Tonow-c2c @jgrandguillaume @simahawk since you contributed or reviewed #1174

guewen:
As discussed, we should keep sale_order_line_packaging_qty for the quantity only, reflecting what's done in `purchase_order_line_packaging_qty`, and have a single module `sale_product_by_packaging` that depends on the qty one for the other use cases (packaging & packaging type can be sold + sell product by packaging only + filter list of available packaging).

jgrandguillaume (approved):
That looks good and in-line with the purchase part. Thx for the job !

sale_by_packaging/models/sale_order_line.py
```diff
@@ -0,0 +1,160 @@
...
+    def write(self, vals):
+        """Auto assign packaging if needed"""
+        fields_to_check = ["product_id", "product_uom_qty", "product_uom"]
+        if vals.get("product_packaging") or not any(
+            fname in vals for fname in fields_to_check
+        ):
+            return super().write(vals)
+        for line in self:
+            line_vals = vals.copy()
+            line_vals.update(self._write_auto_assign_packaging(line_vals))
```
simahawk: shouldn't be called on line?
```suggestion
            line_vals.update(line._write_auto_assign_packaging(line_vals))
```

sale_by_packaging/models/sale_order_line.py
```diff
@@ -0,0 +1,160 @@
...
+    def _check_qty_is_pack_multiple(self):
+        """ Check only for product with sell_only_by_packaging
+        """
+        # and we dont want to have this warning when we had the product
+        if self.product_id.sell_only_by_packaging:
+            if not self._get_product_packaging_having_multiple_qty(
+                self.product_id, self.product_uom_qty, self.product_uom
+            ):
+                warning_msg = {
+                    "title": _("Product quantity can not be packed"),
```
simahawk:
```suggestion
                    "title": _("Product quantity cannot be packed"),
```

sale_by_packaging/tests/test_sale_only_by_packaging.py
```diff
@@ -0,0 +1,114 @@
...
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.partner = cls.env.ref("base.res_partner_12")
+        cls.product = cls.env.ref("product.product_product_9")
+        cls.packaging = cls.env["product.packaging"].create(
+            {"name": "Test packaging", "product_id": cls.product.id, "qty": 5.0}
+        )
+
+    def test_onchange_qty_is_pack_multiple(self):
+        order = self.env["sale.order"].create({"partner_id": self.partner.id})
```
simahawk: the order can be created in setupClass only I think

sale_by_packaging/tests/test_sale_only_by_packaging.py
```diff
@@ -0,0 +1,114 @@
...
+        order_line_1 = self.env["sale.order.line"].create(
+            {
+                "order_id": order.id,
+                "product_id": self.product.id,
+                "product_uom": self.product.uom_id.id,
+                "product_uom_qty": self.packaging.qty * 2,
+            }
+        )
+        self.assertTrue(order_line_1.product_packaging)
+        # self.assertTrue(order_line_1.product_packaging_qty)
```
simahawk: any dead code to cleanup? :)

simahawk (changes requested):
@grindtildeath good work, tnx! Can you squash when done w/ my comments?

sale_by_packaging/models/sale_order_line.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @api.onchange("product_id")
+    def product_id_change(self):
+        res = super().product_id_change()
+        if self.product_id.sell_only_by_packaging:
+            first_packaging = fields.first(self.product_id.packaging_ids)
+            self.update(
+                {
+                    "product_packaging": first_packaging.id,
+                    "product_uom_qty": first_packaging.qty,
```
simahawk: here you could have ZERO as qty. Can we do anything better? Eg: not setting the qty and / or show a warning?

grindtildeath: I filtered out packagings with a qty of zero and added a note in roadmap :wink:

simahawk (approved):

sale_by_packaging/models/product_template.py
```diff
@@ -0,0 +1,62 @@
...
+    _inherit = "product.template"
+
+    sell_only_by_packaging = fields.Boolean(
+        string="Only sell by packaging",
+        default=False,
+        help="Restrict the usage of this product on sale order lines without "
+        "packaging defined",
+    )
+
+    @api.constrains("sell_only_by_packaging", "sale_ok", "packaging_ids")
```
i-vyshnevska, line 17: do you need a "packaging_ids" field here?

grindtildeath: Not needed indeed, good catch

PR #1192, opened 2020-07-13
[13.0][ADD] sale_product_set_packaging_qty

grindtildeath:
This module allows to define packagings and packaging quantities on product set lines.

### DIFF
--- a/sale_product_set_packaging_qty/__manifest__.py
+++ b/sale_product_set_packaging_qty/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Product Set Packaging Quantity",
+    "summary": "Manage packaging and quantities on product set lines",
+    "version": "13.0.1.1.0",
+    "development_status": "Alpha",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["sale_product_set"],
+    "data": ["views/product_set.xml"],
+}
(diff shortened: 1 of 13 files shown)


simahawk (commented):
LG. Would you mind adding some tests?

sale_product_set_packaging_qty/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Product Set Packaging Quantity",
+    "summary": "Manage packaging and quantities on product set lines",
+    "version": "13.0.1.1.0",
```
simahawk:
```suggestion
    "version": "13.0.1.0.0",
```

sale_product_set_packaging_qty/models/product_set.py
```diff
@@ -0,0 +1,76 @@
...
+            if (
+                not line.product_packaging_id
+                or float_is_zero(line.quantity, precision_rounding=uom_rounding)
+                or float_is_zero(
+                    line.product_packaging_id.qty, precision_rounding=uom_rounding
+                )
+            ):
+                line.product_packaging_qty = 0
+                continue
+            line.product_packaging_qty = line.quantity / line.product_packaging_id.qty
```
simahawk, line 32: If the packaging is not set we should not have any qty here. No matter the rest.
Right now the inverse method is broken on `update` because of this.

Use case:
* 1 line w/ no packaging -> pkg qty gets computed
* you call `update` to set packaging + qty -> you get an error

BTW the validation should be done in a constraint IMO.

simahawk (changes requested):

sale_product_set_packaging_qty/models/product_set.py
simahawk: Hmm testing this a bit further it's quite weird: when using `update` by passing pkg qty and pkg, `product_packaging_qty` gets the new value immediately (hence the check on inverse is satisfied) but the pkg is not.
If you use `write`, both values are already updated when hitting the inverse method :thinking:

guewen:
Added some tests

guewen (approved):

simahawk (approved):

simahawk:
@guewen can you check travis?

PR #1193, opened 2020-07-14
[13.0][FIX] sale_order_type: Fix compute writable cache miss error

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @Tardo


pedrobaeza (changes requested):
I think you should put a default instead, as the idea is to not overwrite the sale type if when changing partner, there's no associated order type.

Tardo (approved):

sale_order_type/models/sale.py
```diff
@@ -25,12 +25,11 @@ def _compute_sale_type_id(self):
...
                         force_company=record.company_id.id
                     ).sale_type
-                    or self.partner_id.commercial_partner_id.with_context(
+                    or record.partner_id.commercial_partner_id.with_context(
                         force_company=record.company_id.id
                     ).sale_type
                 )
-                if sale_type:
-                    record.type_id = sale_type
+                record.type_id = sale_type
```
carlosdauden, line 32: Try this:
```suggestion
                record.type_id = record.default_get(['type_id'])['type_id']
```

carlosdauden (approved):

sale_order_type/models/sale.py
carlosdauden: Try this:
```suggestion
                record.type_id = sale_type or record.default_get(['type_id'])['type_id']
```
¿Same in account_move.py?

carlosdauden (approved):

sergio-teruel:
@pedrobaeza @carlosdauden Changes done!!

rafaelbn (approved):
:+1:

ernestotejeda (commented):
Code review :+1:

pedrobaeza:
There's a conflict.

sale_order_type/models/account_move.py
```diff
@@ -12,26 +12,25 @@ class AccountMove(models.Model):
         string="Sale Type",
         compute="_compute_sale_type_id",
         store=True,
-        readonly=True,
-        states={"draft": [("readonly", False)]},
+        readonly=False,
+        states={"posted": [("readonly", True)], "cancel": [("readonly", True)]},
```
ivantodorovich, line 16: Is there a benefict in doing it like this?
It's basically the same but reversed, no? :thinking:

sale_order_type/models/sale.py
```diff
@@ -12,26 +12,29 @@ class SaleOrder(models.Model):
...
         compute="_compute_sale_type_id",
         store=True,
-        readonly=True,
-        states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
+        readonly=False,
+        states={
+            "sale": [("readonly", True)],
+            "done": [("readonly", True)],
+            "cancel": [("readonly", True)],
+        },
```
ivantodorovich, line 20: Same here, why? :thinking:

sale_order_type/models/account_move.py
```diff
@@ -12,26 +12,25 @@ class AccountMove(models.Model):
...
     )
 
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
         for record in self:
             if not record.partner_id:
```
ivantodorovich: Shouldn't we also check if `record.is_invoice()`, to avoid setting the sale_order_type if the move is not an invoice?

```py
if not record.is_invoice():
    continue
```

sale_order_type/models/account_move.py
carlosdauden: Odoo does not correctly manage readonly states in certain cases with computed writable fields. So if we set True to the readonly parameter, the value is not saved.
That is why the logic has been reversed, so that the value of the readonly attribute in the field is False and works correctly.
When Odoo fixes the problem the logic can be simplified again.

pedrobaeza: Not sure if the problem is an Odoo bug or a mangling in inheritance of a mix of attrs and states, but anyway, please use `force_save="1"` in the view instead of changing this and everything solved.

pedrobaeza:
I have rebased for fixing conflicts, but there are still opened several points from me and Iván.

sale_order_type/views/account_move_views.xml
```diff
@@ -4,7 +4,7 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
-            <field name="journal_id" position="before">
+            <field name="journal_id" position="after">
```
pedrobaeza: Why changing this order?

sale_order_type/models/account_move.py
pedrobaeza: Please don't change this and use `force_save="1"` as stated

pedrobaeza:
@sergio-teruel please see my comments

sale_order_type/views/account_move_views.xml
sergio-teruel: In the other way field is not displyed correctly

sale_order_type/models/account_move.py
carlosdauden: `force_save="1"` works at JS view level and the issue is that the modifier is not keep in create method https://github.com/odoo/odoo/blob/27257d34c8b7374bfd8037de71f175f4ecb2913e/addons/web/static/src/js/views/basic/basic_model.js#L3690

`force_save="1"` don't work because `_fields` method:
https://github.com/odoo/odoo/blob/ed89c098b35405cf7c1da98f125761bb354e1297/odoo/models.py#L3701
gets field model definition and this evaluation:
https://github.com/odoo/odoo/blob/ed89c098b35405cf7c1da98f125761bb354e1297/odoo/models.py#L3722
avoid that the field value are protected.

yajo:
This has conflicts @sergio-teruel

sergio-teruel:
Can you fix it??

pedrobaeza:
I'm on it

sale_order_type/models/account_move.py
pedrobaeza: I finally got it: this is not a computed writable field for ORM if not:

- store=True
- readonly=False
- with compute method

so it's clear now and it doesn't matter that the condition is longer this way. It must be this.

Thanks for the patience and the explanations.

sale_order_type/models/sale.py
pedrobaeza: You can check https://github.com/OCA/sale-workflow/pull/1193/files#r489781694

pedrobaeza (approved):
Rebased and adapted to latest changes

carlosdauden (approved):
Unrelated travis error, restart build after 00:00 GMT

ivantodorovich (approved):

carlosdauden:
This improvement is missed?:
https://github.com/OCA/sale-workflow/compare/86283d05fc925129b53f6d74f6b9acd76e3bb5b0..b967dfa5bf5cd0fe5112448295f9eea9566319df#diff-22a8bba379372ac7eaddce1f914a1b55L31

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -88,18 +89,28 @@ def setUp(self):
...
-        sale_line_dict = {
-            "product_id": self.product.id,
-            "name": self.product.name,
-            "product_uom_qty": 1.0,
-            "price_unit": self.product.lst_price,
-        }
-        return {"partner_id": self.partner.id, "order_line": [(0, 0, sale_line_dict)]}
+    def create_sale_order(self, partner=False):
+        sale_form = Form(self.env["sale.order"])
+        sale_form.partner_id = partner or self.partner
```
victoralmau, line 94:
```suggestion
        sale_form.partner_id = partner or self.partner
        sale_form.date_order = fields.Datetime.now() - timedelta(days=7)
```
Now, show travis error related to https://github.com/OCA/sale-workflow/blob/13.0/sale_automatic_workflow/tests/test_automatic_workflow.py#L54
Maybe defining `date_order` that `sale_automatic_workflow` tests will be checked.

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -4,6 +4,7 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 import odoo.tests.common as common
+from odoo.tests import Form
```
victoralmau, line 7:
```suggestion
from datetime import timedelta

from odoo import fields
import odoo.tests.common as common
from odoo.tests import Form
```
Related to previous suggestion it will be necessary to.

PR #1051, opened 2020-01-20
[13.0][MIG] sale_exception

jaredkipe:
Requires https://github.com/OCA/server-tools/pull/1760

Cleanup per docs including squashing "UPD" commits.
Unit tests run and pass.

PR #1194, opened 2020-07-14
[13.0] Add sale exception details on lines

guewen:
For now in draft, because it depends on https://github.com/OCA/sale-workflow/pull/1051
Only the last commit is relevant.

## Keep details of sale exceptions on lines

So we can:

* display the lines with an exception in red
* show the exception messages when we click on a line

It's a huge usability improvement when users have dozens or hundreds of lines.

Screenshot of highlighted line


leemannd (approved):
Thanks!

simahawk:
replaced by #1380

PR #1057, opened 2020-02-03
[UPG] sale_quotation_number 11->12

richard-willdooit:

PR #1196, opened 2020-07-16
[12.0][MIG] sale_quotation_number: Migration to 12.0

manuelregidor:
Standard migration. Supersedes https://github.com/OCA/sale-workflow/pull/1057


manuelregidor:
@ValentinVinagre @HaraldPanten Hello, can you have a look? Thank you.

HaraldPanten (changes requested):
I tested the module on runbot and I realized of some details:

1- There seems to be a strange behaviour when you work with _sale_order_line_input_ module. When you validate the quotations, some gaps are being generated in sales order sequence. IMO That's not a mistake in your module, but you should mention this as a "known issue" in the module readme.

2- Why did you move the "Use Same Enumeration" option from Sales settings to CRM settings? 🤔

sale_quotation_number/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+    def create(self, vals):
+        company = self.env['res.company']._company_default_get('sale.order')
+        if not company.keep_name_so:
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'sale.quotation') or '/'
+        return super(SaleOrder, self).create(vals)
+
+    @api.multi
+    def action_confirm(self):
+        if super(SaleOrder, self).action_confirm():
```
richard-willdooit: @manuelregidor 

There is a small problem with this process.

If you confirm the order, will put the number on the procurement, and will be the reference on Manufacturing Orders.

If you then change the number after it is confirmed, it will not match the procurement references.

We found you really need to allocate the number and put it on the order, and then call action confirm, so that the downstream references match the new number, not the quote number.

sale_quotation_number/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+        related='company_id.keep_name_so',
+        readonly=False
+    )
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.multi
+    def copy(self, default=None):
```
richard-willdooit, line 26: @manuelregidor Copying the origin seems to be confusing at times 

if Order A has an origin of "SQxxx", and it gets copied, then the new quote should not have that as it's origin.  This is in line with standard Odoo - If I duplicate an order, it does not keep on the order a reference to the order or quote it was duplicated from, so why do it here?

This is easy enough to override, but I feel it is against standard Odoo

richard-willdooit (changes requested):

sale_quotation_number/models/sale_order.py
manuelregidor: Hello. So do you suggest just putting in the SO's origin field the name of the SQ it comes from without including the rest of SQs and SOs it might have been coppied from? Thank you.

HaraldPanten (approved):

sale_quotation_number/models/sale_order.py
richard-willdooit: @manuelregidor 

We went with:

```suggestion

    @api.multi
    @api.returns('self', lambda value: value.id)
    def copy(self, default=None):
        default = dict(default or {})
        if not 'origin' in default:
        def copy(self, default=None):
            default['origin'] = False
        return super().copy(default)
```

Just seems that it is better to put nothing than a bad accumulating origin, especially when you consider that a manually created document rarely has a meaningful "origin"

As I said, this one I am not so hung up on, as it is an easy method to write our own extension.

manuelregidor: Ok, I understand what you mean, but I think that's more of an improvement.

sale_quotation_number/models/sale_order.py
manuelregidor: I have tried to sort out that issue. You can check it out in my last commit. 

However, there are still a few problems. Everything works fine as long as just one SQ is confirmed. If multiple SQs get confirmed altogether, the value of number_next_actual (ir.sequence model), which I use, does not get updated until all the SQs have been confirmed.

Maybe you can check it out and find out a way of solving the problem. Otherwise, I will just return to the way the module works in previous versions.

sale_quotation_number/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+                quo = order.name
+            new_name = sequence_id.get_next_char(sequence_id.number_next_actual)
+            # save old values
+            old_origin = order.origin
+            old_name = order.name
+            order.write({
+                'origin': quo,
+                'name': new_name
+            })
+            if not super(SaleOrder, order).action_confirm():
```
richard-willdooit, line 57: @manuelregidor Action confirm always returns True, does it not?  If it fails, it always does a raise, so this seems like a lot of overhead?

sale_quotation_number/models/sale_order.py
richard-willdooit: @manuelregidor Seems very convoluted - confirm always returns true, unless it raises an error, in which case it should roll back.


```
    @api.multi
    def action_confirm(self):
        for order in self:
            if order.state in ('draft', 'sent') and not order.company_id.company.keep_name_so:
                if order.origin and order.origin != '':
                    quo = order.origin + ', ' + order.name
                else:
                    quo = order.name
                order.write({
                    'origin': quo,
                    'name': self.env['ir.sequence'].next_by_code(
                        'sale.order')
                })
```

Should be sufficient - if any order can't confirm it will roll back?

rousseldenis:
@manuelregidor Could you check why travis is red?

sale_quotation_number/models/sale_order.py
manuelregidor: If an order can't confirm, the sequence will go forward anyway and there will be gaps.

richard-willdooit: @manuelregidor 

No, if an order can't confirm, it raises an error.

The code in sale.py returns True every time, unless it raises an error on the way.

I am happy to be shown wrong, but can you point out the code in Odoo that returns anything other than True (or the result of self.write, from action_done, which is always True, too).

manuelregidor: @richard-willdooit 
Yes, it always returns True inless it raises an error. That's why, when an order can't confirm because an error is raised in action_confirm (from sale_order parent), next_by_code has already been done, so the code has moved to the next number. Next time an order is succesfully confirmed, there will be a gap in the sequence.

richard-willdooit: @manuelregidor No, when an error is raised, the transaction rolls back, and the increment is undone.  The sql issues a rollback, and it is in the same transaction.

richard-willdooit: @manuelcalerosolis The increment of the number is inside the same sql transaction

manuelregidor: @richard-willdooit Yes, you're right. I've tested it with "real" errors (not forcing exceptions as I was doing before) and it works fine. Thank you.

manuelregidor:
> @manuelregidor Could you check why travis is red?

Sorted. Thank you.

manuelregidor:
@richard-willdooit @rousseldenis Could you review, please?

sale_quotation_number/models/sale_order.py
```diff
@@ -2,29 +2,23 @@
...
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    @api.model
+    def create(self, vals):
+        company = self.env['res.company']._company_default_get('sale.order')
+        if not company.keep_name_so:
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'sale.quotation') or '/'
+        return super(SaleOrder, self).create(vals)
```
richard-willdooit, line 20: company might be in vals - if so, should use that in preference...

sale_quotation_number/models/sale_order.py
```diff
@@ -0,0 +1,47 @@
...
+
+from odoo import models, api
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def create(self, vals):
+        company = self.env['res.company']._company_default_get('sale.order')
```
richard-willdooit: company might be in vals - if so, must use that in preference...

richard-willdooit (changes requested):

sale_quotation_number/tests/test_sale_order.py
```diff
@@ -35,34 +37,55 @@ def test_enumeration(self):
...
 
-    def test_with_origin(self):
-        origin = 'origin'
-        order1 = self.sale_order_model.create({
-            'origin': origin,
-            'partner_id': self.env.ref('base.res_partner_1').id
-        })
-        quotation1_name = order1.name
-        order1.action_confirm()
+    # def test_with_origin(self):
```
rousseldenis: If you don't use code anymore, please remove it, don't comment.

But, in this case, you cannot delete tests. Please adapt them if code has changed.

rousseldenis (changes requested):

sale_quotation_number/tests/test_sale_order.py
manuelregidor: Done.

sale_quotation_number/models/sale_order.py
manuelregidor: Done.

sale_quotation_number/tests/test_sale_order.py
rousseldenis: Not really. Some tests are still commented.

manuelregidor: Sorry about that. Sorted now.

richard-willdooit (approved):

sale_quotation_number/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     'name': 'Sale Quotation Numeration',
     'summary': "Different sequence for sale quotations",
-    'version': '11.0.1.1.0',
+    'version': '12.0.1.1.0',
```
nikul-serpentcs: version should be `12.0.1.0.0`

sale_quotation_number/__manifest__.py
```diff
@@ -7,7 +7,7 @@
...
     'category': 'Sales Management',
     'website': 'https://odoo-community.org/',
```
nikul-serpentcs: Website URL should be link this 'https://github.com/OCA/sale-workflow/'

sale_quotation_number/models/sale_order.py
```diff
@@ -2,29 +2,27 @@
...
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    @api.model
+    def create(self, vals):
+        company = vals.get('company_id', False)
+        if company:
+            company = self.env['res.company'].browse(company)
+        else:
+            company = self.env['res.company']._company_default_get('sale.order')
```
nikul-serpentcs, line 19: @manuelregidor I think here you should add direct `company = self.env.user.company_id`

nikul-serpentcs (commented):
nitpicking

sale_quotation_number/models/sale_order.py
manuelregidor: That's what function _company_default_get() does, so I'd leave it the way it is now.

sale_quotation_number/__manifest__.py
manuelregidor: Done

sale_quotation_number/__manifest__.py
manuelregidor: Done

sale_quotation_number/models/sale_order.py
nikul-serpentcs: @manuelregidor  Yes you are right 
But IMO `_company_default_get()` method called `self.env['res.users']._get_company()` method and `return self.env.user.company_id`.

That's why I am asking to put directly `self.env.user.company_id`

ValentinVinagre (approved):
Delete the `# -*- coding: utf-8 -*-` and LGTM 👍

sale_quotation_number/models/sale_order.py
ValentinVinagre: If other module inherits the `_company_default_get `, IMO it's not a bad solution.

manuelregidor:
@rousseldenis Is it ready to merge?

manuelregidor:
Hi @nikul-serpentcs, could you review, please?

rousseldenis (approved):

rousseldenis:
@manuelregidor Is it ready for you ?

manuelregidor:
@rousseldenis Yes, it is ready for me.

HaraldPanten:
@rafaelbn @pedrobaeza could you merge this one, please?

HaraldPanten:
> /ocabot merge nobump

THX!

PR #1200, opened 2020-07-23
[13.0] sale_by_packaging: Fix constrains using dotted path

guewen:
Following warning is shown at runtime:

WARNING odoodb_test odoo.models: method product.template._check_sell_only_by_packaging_can_be_sold_packaging_ids: @constrains parameter 'packaging_ids.can_be_sold' is not a field name

Only fields of the current model can be used in a constrains.

### DIFF
--- a/sale_by_packaging/models/product_packaging_type.py
+++ b/sale_by_packaging/models/product_packaging_type.py
@@ -1,10 +1,31 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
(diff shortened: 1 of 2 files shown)


grindtildeath (approved):
LG, but maybe we should add the constraint on product.packaging? Or at least add in roadmap?

guewen:
@grindtildeath Added a constrains on product.packaging.type.
And removed `_check_sell_only_by_packaging_packaging_ids` which checks basically the same thing as `_check_sell_only_by_packaging_can_be_sold_packaging_ids` ("no packaging with can be sold" includes "no packaging").

rousseldenis (approved):
LGTM. @guewen Maybe a test that highlight that ?

PR #1204, opened 2020-07-28
[FIX] 11.0 Sale Fixed Discount

j-zaballa:
Fix
===
`_onchange_discount` method was shadowing the original _onchange_discount
and all its field triggers had been replaced by `discount`. As the new method's
trigger is the field `discount`, the name should be `_onchange_discount`.
But the name is already taken. Hence `_onchange_discount2`. I'm not aware of
a better naming convention in cases like this.

**EDIT:** checked 12.0 and 13.0 branches and saw that `_onchange_discount`
has been renamed to `_onchange_discount_percent`. Already changed it.

**EDIT2** changed `_onchange_discount` call on tests for `_onchange_discount_percent`.

Also, the way the method was implemented, it was impossible to change the
discount manually if the pricelist had a discount, because that would
trigger the `_onchange_discount`, call to super, and get the discount
from the pricelist.

Improve
=======
Override orginal `_onchange_discount` in order to get the value for the
fixed discount in case the price rule method is `fixed`. It basically
repeats what the original method does, so not very DRYish. I don't know
if there is a better approach (I can only think of monkey path, but I
don't like it either)

### DIFF
--- a/sale_fixed_discount/models/sale_order.py
+++ b/sale_fixed_discount/models/sale_order.py
@@ -38,16 +38,68 @@ class SaleOrderLine(models.Model):
         help="Fixed amount discount.")
 
     @api.onchange('discount')
-    def _onchange_discount(self):
-        res = super(SaleOrderLine, self)._onchange_discount()
+    def _onchange_discount2(self):
         if self.discount:
             self.discount_fixed = 0.0
-        return res
 
     @api.onchange('discount_fixed')
     def _onchange_discount_fixed(self):
         if self.discount_fixed:
             self.discount = 0.0
+
+    @api.onchange('product_id', 'price_unit', 'product_uom', 'product_uom_qty',
+                  'tax_id')
+    def _onchange_discount(self):
+        res = super(SaleOrderLine, self)._onchange_discount()
+
+        if not (self.product_id and self.product_uom and
+                self.order_id.partner_id and self.order_id.pricelist_id and
+                self.order_id.pricelist_id.discount_policy == 'without_discount' and
+                self.env.user.has_group('sale.group_discount_per_so_line')):
+            return res
+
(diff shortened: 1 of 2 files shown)


rousseldenis:
@j-zaballa Could you add tests for compute_price == 'fixed' as this is not covered ?

See https://codecov.io/gh/OCA/sale-workflow/compare/6b7b805bc151f20bad744dbc595b79ecd02bcc52...7cf1be65993a4c933b653415517d19fd11a1ce88

j-zaballa:
> @j-zaballa Could you add tests for compute_price == 'fixed' as this is not covered ?

@rousseldenis done!

rousseldenis:
@j-zaballa Can you fix tests as they are failing ?

j-zaballa:
@rousseldenis Yes, I see. I think the issue is that `self.env.user.has_group('sale.group_discount_per_so_line'))` is `False` and `_onchange_discount` returns without doing its job. Not quite sure how to make it `True`. Just adding the group to the user?

Thanks for the feedback, by the way :)

rousseldenis:
> @rousseldenis Yes, I see. I think the issue is that `self.env.user.has_group('sale.group_discount_per_so_line'))` is `False` and `_onchange_discount` returns without doing its job. Not quite sure how to make it `True`. Just adding the group to the user?
> 
> Thanks for the feedback, by the way :)

Indeed, add the group to user before testing.

j-zaballa:
> Indeed, add the group to user before testing.

Thanks! Hope it works now.

rousseldenis:
@j-zaballa Could you squash your commits and rebase ? Thanks

j-zaballa:
@rousseldenis sorry but I have been out of the Odoo world for a while...

PR #1210, opened 2020-08-04
[13.0] Create module sale_order_carrier_auto_assign - alpha

grindtildeath:
This module assigns automatically delivery carrier on sale order confirmation.

### DIFF
--- a/sale_order_carrier_auto_assign/models/sale_order.py
+++ b/sale_order_carrier_auto_assign/models/sale_order.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import models
+from odoo.tools import config
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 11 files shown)


grindtildeath:
@ivantodorovich @pedrobaeza @kongrattapong @nicomacr @cubells @chienandalu @carlosdauden As you guys were last contributors on module sale_order_type, can you please check ea817bf ? :pray: :hugs: 

Also please let me know if I should better extract this commit in a separated PR.

pedrobaeza:
@grindtildeath I think this module should go to OCA/delivery-carrier, so with that, you don't have the problem. Anyway, I think the default was removed previously by a reason. Other thing you can do is to force the sale.order view to use in `Form`.

jbaudoux (approved):
LGTM

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+            "test_carrier_auto_assign"
+        ):
+            return super().action_confirm()
+        self._add_delivery_carrier_on_confirmation()
+        return super().action_confirm()
+
+    def _add_delivery_carrier_on_confirmation(self):
+        """Automatically add delivery.carrier on sale order confirmation"""
+        for order in self:
+            if order.carrier_id or any(line.is_delivery for line in order.order_line):
```
jbaudoux, line 22: I would use here "if order.delivery_set"
This is used to display on the SO form if the delivery method is set.

jbaudoux (changes requested):

jgrandguillaume (approved):
LGTM thx

jgrandguillaume:
I suggest to merge as it's alpha.

PR #1214, opened 2020-08-19
[13.0] Compute shipping rate on carrier auto-assign

guewen:
The computation of the rate was not called, so the line would always be free.
Include also this suggestion: https://github.com/OCA/sale-workflow/pull/1210#discussion_r466378910


grindtildeath (approved):
LGTM :+1: 
Hopefully this wizard's code will be improved in next version :sweat_smile:

jgrandguillaume (approved):
LGTM

PR #1215, opened 2020-08-20
[12.0][FIX] sale_order_line_price_history: set 'unit price' + 'discount'

ernestotejeda:
Now when the unit price of a previous sales order line is set, the discount of that previous line is set too

cc @Tecnativa TT25012

### DIFF
--- a/sale_order_line_price_history/wizards/sale_order_line_price_history.py
+++ b/sale_order_line_price_history/wizards/sale_order_line_price_history.py
@@ -106,7 +106,17 @@ class SaleOrderLinePriceHistoryline(models.TransientModel):
         related="sale_order_line_id.discount",
     )
 
+    def _prepare_set_price_history_vals(self):
+        """ Hook method to prepare the values to update the
+        sales order line in context.
+
+        This method is invoke by action_set_price method in this model.
+        """
+        self.ensure_one()
+        return {'price_unit': self.price_unit, 'discount': self.discount}
+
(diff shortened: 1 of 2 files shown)


sale_order_line_price_history/wizards/sale_order_line_price_history.py
```diff
@@ -109,4 +109,7 @@ class SaleOrderLinePriceHistoryline(models.TransientModel):
     @api.multi
     def action_set_price(self):
         self.ensure_one()
-        self.history_sale_order_line_id.price_unit = self.price_unit
+        self.history_sale_order_line_id.write({
```
pedrobaeza: Put in on a hook for possible inheritance (for example, `sale_triple_discount`).

sale_order_line_price_history/wizards/sale_order_line_price_history.py
```diff
@@ -106,7 +106,12 @@ class SaleOrderLinePriceHistoryline(models.TransientModel):
         related="sale_order_line_id.discount",
     )
 
+    def _prepare_sale_order_line_vals(self):
```
pedrobaeza: This method name is too generic and can collapse with others. Call it `_prepare_set_price_history_vals`, and add a proper docstring. Don't lose good habits.

pedrobaeza (approved):

marcelsavegnago (approved):

PR #1219, opened 2020-08-25
[IMP] sale_promotion_programs: Apply coupon on first sale order only

ajaniszewska-dev:
Aim of this PR is to add possibility to use promotion on first sale order only for each customer.
To achieve that new field is added to promotion form view.

As per behavior in original feature, promotion can be used when predefined code is entered in Sale Order or without that (automatically). 

in case of code usage we need to ensure that this is the first SO of this customer  -> added condition in _check_promo_code()
In case of promotion applied without code,  _filter_programs_from_common_rules() is called. This method filters promotions relevant for particular sale order, based on the requirements -> there we check if there are already some existing sale orders for this customer, if not, this is the first SO so promotion can be applied.

promo

promo_code

### DIFF
--- a/sale_coupon_advanced_option/models/sale_coupon_program.py
+++ b/sale_coupon_advanced_option/models/sale_coupon_program.py
@@ -1,0 +1,52 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import _, api, fields, models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    # Add possibility to use discount only on first order of a customer
+    first_order_only = fields.Boolean(
+        string="Apply only first",
+        default=False,
+        help="Apply only on the first order of each client matching the conditions",
+    )
+
+    def _check_promo_code(self, order, coupon_code):
+        if (
+            self.first_order_only
+            and len(
+                self.env["sale.order"].search(
+                    [("partner_id", "=", order.partner_id.id),("state","!=","cancel")]
+                )
+            )
+            > 1
+        ):
+            return {"error": _("Coupon can be used only for the first sale order!")}
+        else:
+            return super()._check_promo_code(order,coupon_code)
+
+    def _filter_programs_by_sale_order_count(self, order):
+        if (
+            self.env["sale.order"].search_count(
+                [
+                    ("partner_id", "=", order.partner_id.id),
+                    ("state","!=", "cancel")
+                ]
+            )
+            <= 1
+        ):
+            return self.filtered(lambda program: program.first_order_only is True)
+        else:
+            return self.filtered(lambda program: not program.first_order_only)
+
+    @api.model
+    def _filter_programs_from_common_rules(self, order, next_order=False):
+        """ Return the programs if every conditions is met
+            :param bool next_order: is the reward given from a previous order
+        """
+        programs = super()._filter_programs_from_common_rules(order, next_order)
+        programs = programs and programs._filter_programs_by_sale_order_count(order)
+        return programs
--- a/sale_coupon_advanced_option/tests/test_first_so_program_only.py
+++ b/sale_coupon_advanced_option/tests/test_first_so_program_only.py
@@ -1,0 +1,255 @@
+from odoo.exceptions import UserError
+
+from odoo.addons.sale_coupon.tests.common import TestSaleCouponCommon
+
+
+class TestProgramForFirstSaleOrder(TestSaleCouponCommon):
+    def setUp(self):
+        super(TestProgramForFirstSaleOrder, self).setUp()
+
+        self.env["sale.coupon.program"].search([]).write({"active": False})
+
+        self.product_A = self.env["product.product"].create(
+            {"name": "Product A", "list_price": 60, "sale_ok": True}
+        )
+
+        self.program1 = self.env["sale.coupon.program"].create(
+            {
+                "name": "Get 10% discount for 1st order",
+                "program_type": "promotion_program",
+                "reward_type": "discount",
+                "discount_type": "percentage",
+                "discount_percentage": 10.0,
+                "rule_min_quantity": 1,
+                "rule_minimum_amount": 100.00,
+                "promo_applicability": "on_current_order",
+                "promo_code_usage": "no_code_needed",
+                "maximum_use_number": 0,
+                "first_order_only": True,
+                "active": True,
+            }
+        )
+
+        self.program2 = self.env["sale.coupon.program"].create(
+            {
+                "name": "Get 50% discount",
+                "program_type": "promotion_program",
+                "reward_type": "discount",
+                "discount_type": "percentage",
+                "discount_percentage": 50.0,
+                "rule_min_quantity": 1,
+                "rule_minimum_amount": 100.00,
+                "promo_applicability": "on_current_order",
+                "promo_code_usage": "no_code_needed",
+                "maximum_use_number": 0,
+                "active": False,
+            }
+        )
+
+        self.program3 = self.env["sale.coupon.program"].create(
+            {
+                "name": "Get 30% discount with code",
+                "program_type": "promotion_program",
+                "reward_type": "discount",
+                "discount_type": "percentage",
+                "discount_percentage": 30.0,
+                "rule_min_quantity": 1,
+                "rule_minimum_amount": 100.00,
+                "promo_applicability": "on_current_order",
+                "promo_code_usage": "code_needed",
+                "promo_code": "30_discount",
+                "maximum_use_number": 0,
+                "first_order_only": True,
+                "active": True,
+            }
+        )
+
+        self.partner1 = self.env["res.partner"].create({"name": "Jane Doe"})
+
+        self.partner2 = self.env["res.partner"].create({"name": "John Doe"})
+
+        self.partner3 = self.env["res.partner"].create({"name": "John Deere"})
+
+    def test_first_sale_order_program(self):
+
+        #  test first sale order program
+
+        order1 = self.env["sale.order"].create({"partner_id": self.partner1.id})
+        order1.write(
+            {
+                "order_line": [
+                    (
+                        0,
+                        False,
+                        {
+                            "product_id": self.product_A.id,
+                            "name": "Product A",
+                            "product_uom_qty": 2.0,
+                            "product_uom": self.uom_unit.id,
+                        },
+                    ),
+                ]
+            }
+        )
+
+        order1.recompute_coupon_lines()
+        discounts = set(order1.order_line.mapped("name")) - {"Product A"}
+        self.assertEqual(len(discounts), 1, "Order should contain one discount")
+        self.assertTrue(
+            "Get 10% discount for 1st order" in discounts.pop(),
+            "The discount should be a 10% discount",
+        )
+        self.assertEqual(
+            len(order1.order_line.ids), 2, "The order should contain 2 lines"
+        )
+
+        #  test non first sale order
+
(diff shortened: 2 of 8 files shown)


sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import _, api, fields, models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    # Add posibility to use discount only on first customer"s sale order
```
gurneyalex:
```suggestion
    # Add possibility to use discount only on first order of a customer
```

sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
...
+    first_order_only = fields.Boolean(
+        string="Apply only on the first order of each client matching the conditions",
+        default=True,
```
gurneyalex:
```suggestion
        default=False,
```

sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    def _check_promo_code(self, order, coupon_code):
```
gurneyalex: you should not reimplement completely the method -> you need to call super() for the checks which are part of the sale_coupon module and only add checks specific to this module in the overriden method.

sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
...
+                        to get the reward!"
+                    )
+                }
+        return message
+
+    def _filter_programs_by_sale_order_amount(self, order):
+        if (
+            len(
+                self.env["sale.order"].search(
+                    [("partner_id", "=", order.partner_id.id)]
```
gurneyalex:
```suggestion
                    [("partner_id", "=", order.partner_id.id)], limit=1
```
We only care that there is at least one.

sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
...
+            ):
+                message = {
+                    "error": _(
+                        "At least one of the required conditions is not met \
+                        to get the reward!"
+                    )
+                }
+        return message
+
+    def _filter_programs_by_sale_order_amount(self, order):
```
gurneyalex:
```suggestion
    def _filter_programs_by_sale_order_count(self, order):
```

language issue: reading sale_order_amount, people will thing of the total price of the sale order, not the number of sale orders.

sale_promotion_programs/models/sale_coupon_program.py
```diff
@@ -0,0 +1,116 @@
...
+        else:
+            return self.filtered(lambda program: not program.first_order_only)
+
+    @api.model
+    def _filter_programs_from_common_rules(self, order, next_order=False):
+        """ Return the programs if every conditions is met
+            :param bool next_order: is the reward given from a previous order
+        """
+        programs = super()._filter_programs_from_common_rules(order, next_order)
+        programs = programs and programs._filter_programs_by_sale_order_amount(order)
```
gurneyalex:
```suggestion
        programs = programs and programs._filter_programs_by_sale_order_count(order)
```

gurneyalex (changes requested):
I would prefer the module to be named sale_coupon_advanced_option

sale_promotion_programs/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Sale Promotion Programs",
```
gurneyalex:
```suggestion
    "name": "Sale Coupon First Order Only",
```

sale_promotion_programs/__manifest__.py
```diff
@@ -0,0 +1,12 @@
...
+    "summary": """Sale Promotion Programs""",
```
gurneyalex:
```suggestion
    "summary": """Sale Coupon applies only on customer's first order""",
```

gurneyalex (changes requested):
also you need to add a readme (check the template in https://github.com/OCA/maintainer-tools/tree/master/template/module/readme : copy the files which are relevant: description, usage, roadmap and contributors and adapt their content)

sale_promotion_programs/models/sale_coupon_program.py
ajaniszewska-dev: I must disagree, test started to fail here: https://github.com/OCA/sale-workflow/pull/1219/files#diff-1770ad75191c769f8025a7cba19742d7R135   so I used pdb. with this change, even if search return for example 2 sale orders we never go to else block because of` limit=1`. that causes that promotions which should be valid for non-first orders won't be returned and applied.

sale_coupon_first_order/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+    _inherit = "sale.coupon.program"
+
+    # Add possibility to use discount only on first order of a customer
+    first_order_only = fields.Boolean(
+        string="Apply only on the first order of each client matching the conditions",
+        default=False,
+    )
+
+    def _check_promo_code(self, order, coupon_code):
+        message = {}
```
i-vyshnevska:
```suggestion
```

sale_coupon_first_order/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+        if (
+            self.first_order_only
+            and len(
+                self.env["sale.order"].search(
+                    [("partner_id", "=", order.partner_id.id)]
+                )
+            )
+            > 1
+        ):
+            message = {"error": _("Coupon can be used only for the first sale order!")}
```
i-vyshnevska:
```suggestion
            return {"error": _("Coupon can be used only for the first sale order!")}
```

sale_coupon_first_order/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            message = super()._check_promo_code(order,coupon_code)
+        return message
```
i-vyshnevska:
```suggestion
        
        return super()._check_promo_code(order, coupon_code)
```

sale_coupon_first_order/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+
+from odoo import _, api, fields, models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    # Add possibility to use discount only on first order of a customer
+    first_order_only = fields.Boolean(
+        string="Apply only on the first order of each client matching the conditions",
```
i-vyshnevska:
```suggestion
        string="Apply on first",
        help="Apply only on the first order of each client matching the conditions",
```

sale_coupon_advanced_option/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+from odoo import _, api, fields, models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    # Add possibility to use discount only on first order of a customer
+    first_order_only = fields.Boolean(
+        string="Apply only first",
+        default=False,
```
i-vyshnevska, line 13:
```suggestion
```
it is false by default

sale_coupon_advanced_option/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+            and len(
+                self.env["sale.order"].search(
+                    [("partner_id", "=", order.partner_id.id),("state","!=","cancel")]
+                )
+            )
+            > 1
+        ):
+            return {"error": _("Coupon can be used only for the first sale order!")}
+        else:
+            return super()._check_promo_code(order,coupon_code)
```
i-vyshnevska, line 29:
```suggestion
        return super()._check_promo_code(order,coupon_code)
```
you don't need an else block, cos you have a return in if block

sale_coupon_advanced_option/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+    def _filter_programs_by_sale_order_count(self, order):
+        if (
+            self.env["sale.order"].search_count(
+                [
+                    ("partner_id", "=", order.partner_id.id),
+                    ("state","!=", "cancel")
+                ]
+            )
+            <= 1
+        ):
```
i-vyshnevska, line 40: same block as on lines 19-25 move it to function and do call in both cases

sale_coupon_advanced_option/models/sale_coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            return self.filtered(lambda program: not program.first_order_only)
+
+    @api.model
+    def _filter_programs_from_common_rules(self, order, next_order=False):
+        """ Return the programs if every conditions is met
+            :param bool next_order: is the reward given from a previous order
+        """
+        programs = super()._filter_programs_from_common_rules(order, next_order)
+        programs = programs and programs._filter_programs_by_sale_order_count(order)
```
i-vyshnevska, line 51: why do you need "and" here?

sale_coupon_advanced_option/tests/test_first_so_program_only.py
```diff
@@ -0,0 +1,255 @@
...
+
+        self.partner1 = self.env["res.partner"].create({"name": "Jane Doe"})
+
+        self.partner2 = self.env["res.partner"].create({"name": "John Doe"})
+
+        self.partner3 = self.env["res.partner"].create({"name": "John Deere"})
+
+    def test_first_sale_order_program(self):
+
+        #  test first sale order program
```
i-vyshnevska, line 75: the comment is the same as a test method name :)

sale_coupon_advanced_option/tests/test_first_so_program_only.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertEqual(len(discounts), 1, "Order should contain one discount")
+        self.assertTrue(
+            "Get 10% discount for 1st order" in discounts.pop(),
+            "The discount should be a 10% discount",
+        )
+        self.assertEqual(
+            len(order1.order_line.ids), 2, "The order should contain 2 lines"
+        )
+
+        #  test non first sale order
```
i-vyshnevska, line 106: can you split test case on smaller ones?

sale_coupon_advanced_option/tests/test_first_so_program_only.py
```diff
@@ -0,0 +1,255 @@
...
+    def setUp(self):
+        super(TestProgramForFirstSaleOrder, self).setUp()
+
+        self.env["sale.coupon.program"].search([]).write({"active": False})
+
+        self.product_A = self.env["product.product"].create(
+            {"name": "Product A", "list_price": 60, "sale_ok": True}
+        )
+
+        self.program1 = self.env["sale.coupon.program"].create(
```
i-vyshnevska, line 16: I am afraid test case is not working as you expected, with "discount_type": "percentage"  they have been filtered to have only the best one 
https://github.com/odoo/odoo/blob/114b9d2c7b858b9cc8bc8d82239f6fe53290d58f/addons/sale_coupon/models/sale_coupon_program.py#L150 
that is why I had different types of programs in my test case to be sure they all have been applied to my order
https://github.com/camptocamp/sale-workflow/blob/5e55e1686fb8579a71ce905dbd0b31c12d979083/sale_coupon_advanced/tests/test_sale_coupon_computation.py#L23
and in one test case you introducing previous behavior than a new one

PR #1222, opened 2020-08-27
[13.0][IMP] sale_discount_display_amount. Update hooks

acsonefho:
In some case, the hook is triggered and columns could already exists. That raise an exception and block the Odoo start/update process. It could happens when you already use this module on a previous version (9.0 for example) then your DB is migrated to 13.0 and you re-use this module.


danimv5 (approved):
Functional review! LGTM. Good idea.

eduardgm (approved):
Good job @acsonefho ! Functionally checked, LGTM 👍🏻

PR #1231, opened 2020-09-07
[13.0] sale_by_packaging: no pkging override if already set

simahawk:
Before this change the packaging was always overridden
hence the user had not control on the packaging to be used on SO line.

TODO:

- [x] rewrite tests

### DIFF
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -61,8 +61,9 @@ class SaleOrderLine(models.Model):
                     )
                 )
             )
-            if first_packaging:
-                self.update(
+            if first_packaging and self.product_packaging != first_packaging:
+                # Force packacing and its qty if not set already
+                self.with_context(_force_auto_assign=True).update(
                     {
                         "product_packaging": first_packaging.id,
                         "product_uom_qty": first_packaging.qty,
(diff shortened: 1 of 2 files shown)


sale_by_packaging/models/sale_order_line.py
```diff
@@ -101,14 +101,22 @@ def _get_product_packaging_having_multiple_qty(self, product, qty, uom):
...
-        if vals.get("product_packaging") or not any(
-            fname in vals for fname in fields_to_check
-        ):
+        if vals.get("product_packaging") or self.env.context.get("_skip_auto_assign"):
+            # setting the packaging directly, skip auto assign
             return super().write(vals)
         for line in self:
+            if line.product_packaging:
+                # don't touch it if set already
+                continue
```
TDu, line 114: This is problematic in my view, I open a quotation change the quantity on a line, click save, there is no message and the changes are not saved

jgrandguillaume (approved):
LGTM

sale_by_packaging/models/sale_order_line.py
sebalix: fixed

TDu (approved):

TDu:
@simahawk  Looks good as it is now thanks to @sebalix for the last fix. Does it still need the `need fixing` and `work in progress` labels ?

simahawk:
fixes integrated into https://github.com/OCA/sale-workflow/pull/1276

PR #1232, opened 2020-09-08
[IMP] : sale_order_price_recalculation : homogenize button displays

legalsylvain:
Trivial patch.
- homogenize displays of buttons of the module ``sale_order_price_recalculation``. Both modules are moving in the standard header part.

@SandieFavre, @quentinDupont

#GRAPOCA

### DIFF
--- a/sale_order_price_recalculation/views/sale_order_view.xml
+++ b/sale_order_price_recalculation/views/sale_order_view.xml
@@ -6,20 +6,14 @@
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form"/>
         <field name="arch" type="xml">
-            <field name="note" position="after">
-                <group name="sorp">
-                    <div class="btn-group">
-                        <button name="recalculate_prices"
-                                string="Recalculate prices"
-                                attrs="{'invisible': [('state', 'not in', ['draft', 'sent'])]}"
-                                type="object" colspan="4" class="ml-2 btn btn-default"/>
-                        <button name="recalculate_names"
-                                string="Reset descriptions"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I prefer buttons on bottom, as they are not so important to be before lines.

legalsylvain:
Hi @pedrobaeza 

> I prefer buttons on bottom, as they are not so important to be before lines.

Well, I think it depends ;-). For some odoo instances, the button is not so important, but for others, with a big use of pricelists, it is very important.

but my point was not about if it is important or not. It's just about UX. 
1) when there are a lot lines, the button are at the bottom of the page, so changing pricelist then update prices is not a user friendly process.
2) this module introduces a concept that is totally new in all Odoo, "a button at the bottom of a form". In fact, when we showed the sale module to newcomers in GRAP, they just did'nt found the button. That's a UX issue IMO.

pedrobaeza:
Well, I'm only using the recalculation module, and I don't want to have it on top, as it impacts also on the mobile view. If you find a better position (maybe in the header), it's OK for me. If not, I don't accept the change.

legalsylvain:
> If you find a better position (maybe in the header), it's OK for me

The header is great also. I'm totally agree. I was just saying that putting a button on the bottom of a form is very bad.

hparfr:
Thanks for the harmonization. 
I prefer when buttons are all on top.

hparfr (approved):

pedrobaeza (changes requested):
Change both to header then.

pedrobaeza (approved):

legalsylvain:
thanks you both for your review !

PR #1237, opened 2020-09-09
[FIX] sale_order_product_recommendation: wizard lines without wizard_id reference.

sergio-teruel:
cc @Tecnativa TT25430
@carlosdauden @chienandalu Can you review?

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -125,6 +125,7 @@ class SaleOrderRecommendation(models.TransientModel):
             i += 1
             if i >= self.line_amount:
                 break
+        recommendation_lines.wizard_id = self
         self.line_ids = recommendation_lines
 
     def action_accept(self):
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -125,8 +125,13 @@ class SaleOrderRecommendation(models.TransientModel):
             i += 1
             if i >= self.line_amount:
                 break
-        recommendation_lines.wizard_id = self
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

chienandalu (approved):

sergio-teruel:
pre-commit fixed on sale_quotation_number module

pedrobaeza:
I'm afraid there's an error in tests

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -125,6 +125,7 @@ def _generate_recommendations(self):
             i += 1
             if i >= self.line_amount:
                 break
+        recommendation_lines.wizard_id = self
```
pedrobaeza: This shouldn't be needed, as the assignation to the one2many should assign this inverse field.

chienandalu: Then maybe it's missing in the prepare: https://github.com/OCA/sale-workflow/pull/1237/files#diff-bf4bc078958d60f65a62f59551d2e6e1R65-R78

pedrobaeza: That field is filled in the next line, as part of the one2many assignation (the inverse one is assigned). If @sergio-teruel tells us the initial problem, then we can check the source of the problem.

pedrobaeza: Checked with him, and seems a fault in the ORM and he has put the more resilient way of fixing the problem and allow the method to be called from onchange or from real records. The ORM flaw might be solved with https://github.com/odoo/odoo/pull/55995 in v14.

pedrobaeza:
/ocabot merge patch

Is there the same problem in `purchase_order_product_recommendation`?

PR #1239, opened 2020-09-09
[13.0][MIG] sale_order_price_recalculation: Migration to 13.0

pedrobaeza:
- Standard procedure
- Adapt to `Form` commodity

@Tecnativa TT22088


pedrobaeza:
@sergio-teruel @rousseldenis if you want to check. The code has become pretty simple.

rousseldenis:
> The code has become pretty simple.

Not sure it is more simple as it was already (not so much less lines) :smiley: 

But you could have taken the two first commits I did in the other PR.

rousseldenis (approved):

pedrobaeza:
Yeah, sorry about that, but I saw a lot of noise to add the diff with onchange_helper code and then remove all that lines again.

sergio-teruel (approved):

legalsylvain (changes requested):
Hi. 

> Yeah, sorry about that, but I saw a lot of noise to add the diff with onchange_helper code and then remove all that lines again.

I don't think it's noise. For me, the use of the class ``odoo.tests.Form`` is not an OCA convention, and there a lot of cons. See recent reactions on the tweet of @yenthe666 https://twitter.com/Yenthe666/status/1299715778447650816

I'm personaly not agree to use it out of test section. I think that we should have a talk regarding that point. If the talk has been already done, please point the URL, and I'll remove my blocking point.

**Pro:**
- very synthetic code

**Cons:**
- Using a "tests" tools in a non tests part is not a good practice.
- In fact, odoo himself just use this tools for the test, not for the code of the module.
- It is very, very slower. I just installed ``purchase_quick`` on my Odoo V12 instance. All the user are very happy with that new feature, but in the same time, we are facing slowness. Investigating the reason. all the time is consumed by the call of ``odoo.tests.Form``. 

I'm currently working to remove this call manually. I'd like to avoid to remove manually this call in all the new OCA modules I'm installing ;-)


CC : @guewen, @lmignon 

kind regards.

pedrobaeza:
Please don't start again the same debate. It's not a test tool. It's a tool that started for tests and the expanded, but Odoo didn't want to move and change all the imports. It's a bit slower, but ensure correctness, and in v13, its performance is OK. Besides, this button is pressed manually, and not in a massive way, so there shouldn't be a problem.

Don't lead to a fork inside OCA having a module with it and another without, as I want to have it and forget all other dependencies or side effects, and if you don't accept it, I will create a `sale_order_price_recalculation_with_form` or similar module.

rousseldenis:
So, my feelings about that were not so bad... If performance problems are confirmed, we cannot rely on it in this as price computation has to be as lean as possible and as .. performant as possible.

Moreover, we can deduce that Odoo will NEVER fix Form() performance problems and cannot take time to enhance it (maybe) in OCB.

onchange_helper as first suggested is very well tested, used in production in several projects by people across OCA.

So, we need to decide what to do. You know also my point of view. See https://github.com/OCA/sale-workflow/pull/1155

pedrobaeza:
"Performance" problems are because more things are done (onchanges, etc), and a little overhead (Form preparation) against direct field assignation, but in v13 the performance is good enough, and as said, business logic is more important. We all know that this comes from a bad Odoo design since the beginning coupling view and controller layers, but for now, we can't do other thing. The switch to computed writable fields will allow us to avoid this entirely, but that simply can't happen now.

rousseldenis:
> business logic is more important

Yes, but if you want to use it on an ecommerce website for instance, performance is needed too. You cannot decrease customer UX, no ?

pedrobaeza:
Of course, but the optimizations should be done in the proper layer or hack that instance. For me, the path to optimization is clear: go to v13 ASAP, switch to computed writable fields in new modules (I have been doing such effort in my migrated modules), and use `Form` when not possible (which seems anti-optimization, but it's for avoiding side effects and a bunch of glue modules).

legalsylvain:
> If performance problems are confirmed

OK, i'll try to provide benchmark. In fact, I only use even odoo version, so don't know if in V13, performance issue are minor. I'll say you in a couple of days. 

kind regards.

Cedric-Pigeon:
@pedrobaeza I do not really agree on this argument:
`Besides, this button is pressed manually, and not in a massive way, so there shouldn't be a problem.`
It is your point of view. On one project we use it to recalculate all opened quotations in case of pricelist changes.  No more massive process than that.

I do not have a clear opinion of using it or not within OCA addons but at least I think that a position has to be taken before spreading it everywhere. If worries about performance are founded, we should avoid using it. Do we have any case where Odoo uses it in the core for business code? If not I would be in favour of avoiding it in OCA too.
If performances are ok and Odoo tends to use it, than let's go. If not let's limit its usage within tests and not in business code.

pedrobaeza:
OK about the potential usage of this module as base for massive operations.

Form is used in several places in Odoo business logic in new modules done since its inclusion, like:

https://github.com/odoo/odoo/blob/f89b715f05acda64ce6db73010d16487db9b269a/addons/account_facturx/models/account_move.py#L142
https://github.com/odoo/odoo/blob/f89b715f05acda64ce6db73010d16487db9b269a/addons/l10n_it_edi/models/account_invoice.py#L361

I'm performing speed tests in v12/v13 for comparing and will provide results ASAP.

pedrobaeza:
I have done the performance test, and I have discovered that my code contained a bug and was not optimized as well, so thanks for pushing me to re-check.

I have taken 2 DBs with customer data and heavy additional layers (275 in v12 and 256 modules in v13) for testing in more real conditions. They can't be exactly the same, but I think the environments are good enough for this test.

These are the numbers for v12:

Selección_006

And for v13, these are the numbers:

Selección_007

My conclusions are:

- v12 is a bit more performant, as it was not executing all the onchanges.
- v13, even performing all onchanges associated to all the extension modules we have in both installations (sale_secondary_unit, sale_margin, and a big etc), performs good enough thanks also to the ORM optimizations.
- On aggregated data/big sales orders, differences are not so big.
- You have to take care on how do you use the `Form` (but the same as developing under-performant computed fields or other ORM elements).

So I get to the same summary: "Form is a bit slower than direct code as expected, but it's performant enough for the benefit of respecting inherited business logic". Do you agree with me?

HaraldPanten:
IMO performance is not perfect, but it's good enough. Thx for testing 👍

legalsylvain:
Hi @pedrobaeza. 
Your test are mixing three things. 
- Odoo versions (12 / 13), 
- two differents databases, 
- the use (or not) of ``odoo.tests.Form``

Isolating each changes, the result are quit different.

I so tried to introduce your refactoring in V12. here are [the result](https://github.com/OCA/sale-workflow/pull/1241)
-> Using odoo.tests.Form, the processing time is multiplied by 8.

Against your branch V13, I removed your changes and keeped the previous code. Here are [the result](https://github.com/Tecnativa/sale-workflow/pull/1) : 
-> Removing odoo.tests.Form, the processing time is divide by 11.

Did I miss something ? Is there anything I'm doing wrong ?

pedrobaeza:
Hi, @legalsylvain,

I think you have lost the context of the history that brings us here. Let me introduce it, as there has been several threads that are not easy to follow:

- First of all, let me tell you that I was one of the first contributors of this module with the approach of calling the onchange in 8.0: https://github.com/OCA/sale-workflow/blob/8.0/sale_order_price_recalculation/models/sale_order.py, knowing its limitation but being efficient and performant.
- Acsone, when migrating the module to v13, probably saw/were affected by its limitations:

  * if you instantiate another onchange method (one with other name but the same decorator), it's not called,
  * And also if there's chained onchanges, they are not called as well.

- Then they decided to switch to a dependency on `onchange_helper`, for overpassing the limitations.
- It was in that moment when I commented the inconvenience of adding such "big" dependency for this little module, and the existence of a native way of overpassing the limitations through `Form`. Why my reluctance on this? Because our maintenance model includes support for all the code installed in the customers, and when migrating customers that have this module installed from v12 to v13, we pass from maintaining few lines of this module to a bunch of more with the new dependency, which potentially we will need to:

  * migrate (we are not doing that for now as it's being maintained by others - thanks for that-, but we are contractually forced to migrate it if others retire its support or do it in late terms).
  * support any side effect on current version
  * etc

- Then this thread becomes a debate of using `Form` in general in OCA modules or not using it (in favor of other solutions).
- What I tried to showcase in my performance tests, was that the performance was near the one we had in v12, which is true. That's why I compare a v12 with previous code with an v13 with `Form` code. We all know that the previous v12 equivalent code is blazing fast compared to `Form` option, but we try to harmonize both options.

Now, the options are:

1. Keep this with `Form` with what it can be considered as acceptable performance.
2. Switch again to `onchange_helper`, which is very inconvenient for us as explained and we will need to have an alternative module.
3. Back again to the v12 code with its limitation. This can be workarounded changing involved modules for using the same onchange method instead of new ones, and including all the price logic in the main product onchange.
4. An intermediate option that can be to called `model.onchange` core method that at least execute all the onchanges (but not chained).

And for the general debate, my point was to prove that losing a bit of performance, using `Form` we ensure business logic, which is critical for example in record creation (not specially for this module).

I hope now you have the whole picture clearer. @rousseldenis can tell us which was their specific problems to focus more the final solution.

lmignon:
Hi all,

I am a little bit concerned in the current discussion to see that, in the end, internal constraints specific to a major and important actor/contributor for the community should guide the choices of good practices for the whole community. 
As a main contributor and user of onchange_helper, we have continuously maintained and migrated this module. It is available in all versions of odoo. So I don't see why it couldn't be used even though it brings a significant performance gain compared to what Odoo is capable of doing natively. I will be happy to remove this module once Odoo has removed all the onchanges. Unfortunately this is not yet the case for the moment. 
It's a game for all open source players to ensure the maintenance of the modules that are installed at the client's site. In this game, I think we all participate in the correction/improvement/maintenance of modules produced by others if this is necessary for our clients. We don't ask Technativa to take responsibility for maintaining modules that we have developed. I therefore believe that this element cannot be taken into consideration here.

kind regards

lmigon

pedrobaeza:
@lmignon I see that you haven't taken my point on how we work, and I didn't want to add `onchange_helper` dependency nor `Form` processing. It was your colleague who added it, and I tried to find a solution that satisfies both: not adding that dependency burden to this module, but not losing the reason why you added it. If there's no problem with the second point, we rolled-back again to the approach of v12. For me, it's not problem and I prefer it.

Cedric-Pigeon:
@pedrobaeza Sorry but I totally agree with @legalsylvain. To be relevant and comparables, performances tests must be done in the same context: same database and same version.

pedrobaeza:
@Cedric-Pigeon, I repeat that I prefer not to depend on `onchange_helper` nor `Form`, but it was your company who introduced that need, so for me it's OK to rolled-back to previous approach.

Comparing same version with previous approach with current version with `Form` obviously will be fatal for `Form` approach, as it's doing a lot of things against doing almost nothing, but I tried to say that we will have something similar to 12.0.

Can I switch to v12 approach?

Cedric-Pigeon:
@pedrobaeza Yes, I know but what I really don't catch is why is it a problem? `onchange_helper` is highly supported, tested and often used.

rousseldenis:
> @Cedric-Pigeon, I repeat that I prefer not to depend on `onchange_helper` nor `Form`, but it was your company who introduced that need, so for me it's OK to rolled-back to previous approach.
> 
> Comparing same version with previous approach with current version with `Form` obviously will be fatal for `Form` approach, as it's doing a lot of things against doing almost nothing, but I tried to say that we will have something similar to 12.0.
> 
> Can I switch to v12 approach?

For history: 

The reason why onchange_helper was introduced was that onchange methods (v12) produced wrong result in v13. I did at that time in a well documented commit the change to onchange_helper. The tests were passing very well too.

pedrobaeza:
@Cedric-Pigeon Think in the way that now Odoo works supporting third-party modules: you have a price per 1k lines of code. We have customers with `sale_order_price_recalculation` module in v12, but not with `onchange_helper`. Migrating them to v13 will mean to raise the number of lines to maintain from a few to a lot of more due to the new dependency. It's something not affordable. You say that the module is maintained, and indeed it is... for now. I'm forced to migrate it/maintain it, you doing it for me or not. Maybe you decide that you skip v14 and don't migrate it, or even migrating it, you do it very late and I have to migrate it before for being able to migrate my customer. And not only that, if there's a side effect in other point because the module performs something that breaks an standard behavior, and I need to debug and fix that, as you haven't experimented that bug, it costs me in my maintenance balance.

I think this is not a Tecnativa (why all of you write Technativa :stuck_out_tongue_closed_eyes: ?) question, but a general one in an ecosystem where maintenance is relevant. And that's why I keep saying that if you don't accept to move to `Form` and keep `onchange_helper`, we will need a "light fork" for our customers. All with open source rules, not imposing anything, but not being forced also to anything.

@rousseldenis thanks for explaining the reason why `onchange_helper` was introduced. Then would you all be OK if I touch the code for not depending on `onchange_helper` nor on `Form`, but keep this working?

rousseldenis:
> @rousseldenis thanks for explaining the reason why `onchange_helper` was introduced. Then would you all be OK if I touch the code for not depending on `onchange_helper` nor on `Form`, but keep this working?

Sure. If you find it, go ahead! But I remember that it took it long debugging time already.

pedrobaeza:
@rousseldenis another question about that: the problem was risen by current tests?

Cedric-Pigeon:
@pedrobaeza your argument sounds totally crazy to me. If I well follow you, you only include addons which only depends on add-ons maintained by you? Sorry, but where is the community spirit in that? Yes sometimes you need to migrate dependencies, but sometimes it is already done by others. That's the case for everyone. 
And now what if Odoo decides to disallow `Form` usage in business code in V14.0? The problem is exactly the same.

pedrobaeza:
@Cedric-Pigeon you are not following me: I'm talking about number of lines of code to maintain: it's not the same (fictional numbers) 100 lines for `sale_order_price_recalculation` than 100 + 1000 for `onchange_helper`, but indeed, we assume the support for our customer for the whole code they have installed in the DB, not only the modules we have created. Why is that crazy? If you want, we can talk about this in a live session for clarifying.

Cedric-Pigeon:
@pedrobaeza Sorry I missed the Odoo contract based on the number of lines so I misunderstood your arguments. 
My position is clear:
At an equal performance (or really poor degradation on high volumes), I am ok to switch to `Form` otherwise I would rather be in favour of using `on_change_helper`. We could not impact the efficiency of a solution for contractual reasons.

pedrobaeza:
OK, I will try to make this work with neither of both. I'll tell you soon.

rousseldenis:
> @Cedric-Pigeon you are not following me: I'm talking about number of lines of code to maintain: it's not the same (fictional numbers) 100 lines for `sale_order_price_recalculation` than 100 + 1000 for `onchange_helper`, but indeed, we assume the support for our customer for the whole code they have installed in the DB, not only the modules we have created. Why is that crazy? If you want, we can talk about this in a live session for clarifying.

It's clear for me, but IMHO your internal strategic choices must not block OCA developments. Indeed, when we develop something, we should take care of dependencies. Concerning this point, onchange_helper is a transversal module you can find in dependencies in other modules through OCA.

I understand more the reaction you have in other PR's with dependencies adding. But my opinion is, if a bunch of code, even external, is doing things in a good way, why not using it.

Indeed, it's not the Tecnativa way, it's the OCA one.

pedrobaeza:
It's not only our internal strategies, but now also Odoo ones with their maintenance support for third-party modules. We both can be wrong or not, but it's a factor to be taken from now on the whole Odoo ecosystem.

Yenthe666:
@pedrobaeza about the maintenance support: it's indicative and not required so I can assure you almost nobody runs the Odoo maintenance way.
On a sidenote: damn you guys argue a lot.

rousseldenis:
> @pedrobaeza about the maintenance support: it's indicative and not required so I can assure you almost nobody runs the Odoo maintenance way.
> On a sidenote: damn you guys argue a lot.

Whooo, you here ? Cool, you come and contribute ?

pedrobaeza:
@Yenthe666 indeed we don't have any customer with Odoo third-party maintenance because... we are doing that since 5 years! It's our business model that Odoo "has copied", so it's very relevant for us.

Yenthe666:
> > @pedrobaeza about the maintenance support: it's indicative and not required so I can assure you almost nobody runs the Odoo maintenance way.
> > On a sidenote: damn you guys argue a lot.
> 
> Whooo, you here ? Cool, you come and contribute ?

Not sure if you're being sarcastic here. I shall take it as a happy & thankful comment?

rousseldenis:
> > > @pedrobaeza about the maintenance support: it's indicative and not required so I can assure you almost nobody runs the Odoo maintenance way.
> > > On a sidenote: damn you guys argue a lot.
> > 
> > 
> > Whooo, you here ? Cool, you come and contribute ?
> 
> Not sure if you're being sarcastic here. I shall take it as a happy & thankful comment?

No, no really. I know your technical skills and I think it's dommageable for OCA not to have you in contributors. It was maybe a little at the beginning, but my real will is to have you in the pack! So, welcome when you want!

lmignon:
My aim is not to question the business models of the various people here. However, I do wonder about the deep motivations of the various parties in this debate. 
The idea in our developments is to offer the best to our customers. If the best goes through more lines of code, what's the problem? Where I personally am starting to have a problem is realizing that commercial constraints of some people should govern our choices. For me, this becomes a brake on innovation. If we push the reasoning even further, tomorrow we'll have to do without queue_job because odoo is developing its own system, even though queue job has more functionalities and is more efficient. 
I think that like many members of the community, we participate significantly in the migration and maintenance effort of the modules used by our customers. This effort is a shared responsibility and it seems to me that no one can claim a greater responsibility than the others.
If tomorrow, the cost of maintenance proposed by some must guide the technical choices and restrict the community's possibilities for innovation, the spirit of open source development will be lost and it will become impossible to collaborate on a solution common to all. 
The constraints of some are not the constraints of others. This is why it is necessary, in my opinion, to focus on what is important, namely the user experience, without introducing elements specific to one's own constraints. In this case, in my opinion, the use of the onchange_helper benefits the end user by offering a higher level of performance. There is therefore no reason to do without it. 
Kind regards.

legalsylvain:
Hi @pedrobaeza and others. First thanks for all that contributions and remarks.

### Performance remarks

@pedrobaeza said : 

> v12 is a bit more performant, as it was not executing all the onchanges.

You are observing a loss of 25 % ((53 - 66) / 53), passing from 12.0 without Form to 13.0 with Form. --> It could be OK.

When I test with just demo data, I observe a loss of 518 % ((2.874 - 17.79) / 2.874) passing from 12.0 without Form to 13.0 with Form. --> It is not acceptable.

I can not figure out how it is possible to have such a big difference.

V12 and V13 seems to have similar demo data. (5 sale.order and 11 sale.order.line)

Any insights ? 

### Business model

Regarding the business model, I think that the business model of Tec(h)nativa is great. Charge customers for maintenance and migration of custom / community code is very sustainable. In the meantime, I don't think that choices regarding performances and design should be affected by such an invoicing concept.

> It's not only our internal strategies, but now also Odoo ones with their maintenance support for third-party modules.

Not exactly. When you engage you to port module of third-party (OCA modules) Tecnativa shares the result and makes a PR on the OCA. (that is great). I never saw Odoo SA workers doing the same. I think that is not the same work, and not the same quality.

**Question**

I don't see ``onchanger_helper`` as a huge module : 
```
>>> cloc ./grap-odoo-env-12.0/src/server-tools/onchange_helper/ --exclude-dir=tests,i18n,description,demo

-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                           4             10             22             58
-------------------------------------------------------------------------------
SUM:                             4             10             22             58
-------------------------------------------------------------------------------
>>> cloc ./grap-odoo-env-13.0/src/server-tools/onchange_helper/ --exclude-dir=tests,i18n,description,demo
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                           4             10             22             55
-------------------------------------------------------------------------------
SUM:                             4             10             22             55
-------------------------------------------------------------------------------
```
In the meantime, I guess that all CE instance of Tecnativa and other integrators, are installing the must have ``web_responsive`` module (thanks to all the developpers BTW), and that is not the same game : 

```
>>> cloc ./grap-odoo-env-12.0/src/web/web_responsive/ --exclude-dir=tests,i18n,description,demo
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Sass                             1             85             50            442
JavaScript                       1             38            197            291
XML                              6             34             41            217
Python                           4              6             11             40
-------------------------------------------------------------------------------
SUM:                            12            163            299            990
-------------------------------------------------------------------------------
>>> cloc ./grap-odoo-env-13.0/src//web/web_responsive/ --exclude-dir=tests,i18n,description,demo
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Sass                             1            101             56            592
JavaScript                       2             53            302            510
XML                              8              0             45            303
Python                           4              6             12             37
-------------------------------------------------------------------------------
SUM:                            15            160            415           1442
-------------------------------------------------------------------------------
```

55 lines of code, including manifest is very small. I never said that it was 55 easy lines to migrate ;-), but if the metric is the number of lines, well, I don't see any problem with that.

pedrobaeza:
@legalsylvain it's not only a question of the number of lines (although it can be also significative), but the change across versions. Anyway, you can see in the PR history that it has not been easy to get to that number of lines: https://github.com/OCA/server-tools/pulls?q=is%3Apr+onchange+is%3Aclosed, and even on v9 the module was never merged: https://github.com/OCA/server-tools/pull/1579. So my general point is that it's a risk to add more modules/code. But I want to let that debate that is unfocusing us of this being merged.

I have already switched to the direct approach that was in v12 without neither of `onchange_helper` nor `Form`, and tests are working properly, so I hope this can finally get merged.

legalsylvain (approved):
thanks for the changes, and for all the discussion around ``odoo.tests.Form``, ``onchange_helper``, etc...

Cedric-Pigeon (approved):
Thanks @pedrobaeza

rvalyi:
Hello folks. Interesting talk. I have a very naive question:
would it bit possible/reasonable to have a 1 line test, similar to the `hasattr(someobject, some_column)` hack that would test if `onchange_helper` module  is installed and call the onchange_helper faster implementation if available. But without hard dependency on it. Would it be simple enough? Would it behave in ways deterministic enough in both cases?

rousseldenis (approved):

pedrobaeza:
@rvalyi current code is faster than with `onchange_helper`. The potential problem it has is a third party module modifying the price/discount/name in chained onchanges.

rvalyi:
> @rvalyi current code is faster than with `onchange_helper`. The potential problem it has is a third party module modifying the price/discount/name in chained onchanges.

I got that and I have nothing against the current PR. In fact my question was more generic about the onchange_helper vs odoo.tests.Form debate (@legalsylvain also just reported huge perf win of removing odoo.tests.Form from other modules in a French user group). Would it be reasonable to use onchange_helper when available and fall back on odoo.tests.Form if not (so without hard dependency). Disclaimer I didn't think too much about it so might sound a dumb question.

pedrobaeza:
Well, it has significant structural changes, so it will mean to double the amount of code. Its scope is also not the same. `Form` is for filling a complex record (even with o2m) as if you are doing it in UI backend. `onchange_helper` serves for ensuring certain consistency with UI, but not a global one, and it depends on the internals of each model to get the same result (onchange application order for example). I still prefer `Form` for having more accurate results, but the performance can be a problem in certain kind of operations. If only computed writable fields were used everywhere already...

legalsylvain (changes requested):
Could you just cherry pick this trivial v12 commit ?
https://github.com/OCA/sale-workflow/pull/1232

pedrobaeza:
@legalsylvain cherry-picked. Please merge this.

lmignon (changes requested):
Sorry, but the current implementation is not right and will not work if others onchanges are defined in custom addons. I completely disagree that the commercial interests of some prevail over the quality, genericity and performance of the code produced in OCA. For me, this kind of argument is the beginning of the end of OCA. OCA is not the repository of one company, it is the repository of a community.
Last but not least, the commit history is not preserved since it doesn't contains the initial work of @rousseldenis

pedrobaeza:
@lmignon here there are no commercial interest, but the implementation as it has been since 4 versions. Please stop sabotaging this. All of your colleagues have agreed to continue this way. The commit history is totally preserved from previous versions, so also nothing to argue about OCA conventions. If you want to add a lot of intermediate diff lines for nothing (as it will be added and then removed), I can make you happy for finally finishing this, but again your colleague has accepted.

legalsylvain:
@pedrobaeza it seems that you have to change the name of the button.

see : https://travis-ci.org/github/OCA/sale-workflow/jobs/726941492#L1221

lmignon:
@pedrobaeza In a democracy everyone can express themselves. My colleagues got tired of endless discussions and finally decided to leave it as is. Once again, it's not because the implementation is the same from 4 versions that it can't evolve. I'm sorry, but here it's not me who is sabotaging, but you who are against all evolution even though it brings more genericity. 
It's therefore important to me to express my opinion even if the decision that'll be taken is not the better one IMO.

pedrobaeza:
@legalsylvain I have blindly cherry-picked the commit, without checking its correctness in this version. I'm fixing now.

@lmignon expressing your opinion, of course. Blocking the merge of this, I think you shouldn't, as I have been answering all the requests, consuming a lot of time, for getting a minimum agreement. If you don't like this implementation, don't use it. Create another module (OCA or not), using `onchange_helper`. I was ready for doing the same on the contrary case, but it's not fair now after all that not being you the one that have worked on this, you block it. If it has been implemented this way during 4 versions, the not perfect implementation is not so bad at all.

legalsylvain:
> @legalsylvain I have blindly cherry-picked the commit, without checking its correctness in this version. I'm fixing now.

thanks !

Second point : 
Well, migrate from a version to another is a good period to improve modules, fixing problems. But in the other hand, it's an extra cost / extra work that is not fair to ask to Pedro to do, as he's against this design, and seems to don't like ``onchange_helper`` module.

solution A) We merge this PR as it.
solution B) @lmignon, (or others), adds the dependency to ``onchange_helper`` module, improving the current PR, before merging this one.
solution C) We add in a second time, a glue module ``autoinstall=True``, named ``sale_order_price_recalculation_onchange_helper``, that overwrite ``_recompute_form_line_vals`` function. 

My preference goes to solution B, because I'm agree with this point _"but the current implementation is not right and will not work if others onchanges are defined in custom addons"_ so there is no reason to use ``sale_order_price_recalculation`` without ``sale_order_price_recalculation_onchange_helper`` IMO.

kind regards.

lmignon:
@legalsylvain The initial migration and enhancement has been done by @rousseldenis https://github.com/OCA/sale-workflow/pull/1155 and refused/closed by @pedrobaeza. We never asked Pedro to do the job and the migration was done since May 28.

pedrobaeza:
@lmignon I didn't closed the PR on my own, as I asked @rousseldenis if I can resume the work as we need this module, with his approval, but the PR he did was not finished (and he said that he can't finish it because he was busy) and with Travis red.

rousseldenis:
> @lmignon I didn't closed the PR on my own, as I asked @rousseldenis if I can resume the work as we need this module, with his approval, but the PR he did was not finished (and he said that he can't finish it because he was busy) and with Travis red.

Travis was not red with just onchange helper. You didn't want it as is. https://github.com/OCA/sale-workflow/pull/1155#discussion_r436517469

I've just removed the temporary commit that tried to restore v12 behaviour. Waiting for travis there https://github.com/OCA/sale-workflow/pull/1245

rousseldenis:
Let's merge this as we're a little bit tired. Discussion is often constructive but not in this case.

That highlights even so the ownership of some on something that should be contributive and open... Not the idea of what I think OCA should be.

legalsylvain:
@rousseldenis : I don't understand. #1245 is failing ?

rousseldenis:
> @rousseldenis : I don't understand. #1245 is failing ?

Nope, it's green, but we are out of time to get tests on that.

pedrobaeza:
Last Travis of the other PR was red: https://travis-ci.org/github/OCA/sale-workflow/builds/725179536 (but I didn't check why), and again it was under your approval: https://github.com/OCA/sale-workflow/pull/1155#discussion_r485402086

rousseldenis:
> and again it was under your approval: [#1155 (comment)](https://github.com/OCA/sale-workflow/pull/1155#discussion_r485402086)

I was already fed up with that but that was before @legalsylvain taking interest of this as no one came.

pedrobaeza:
@rousseldenis Are you debating my OCA spirit, answering all your questions, changing code for reaching an agreement and consuming a lot of time?

rousseldenis:
> @rousseldenis Are you debating my OCA spirit, answering all your questions, changing code for reaching an agreement and consuming a lot of time?

Nope, don't make me say something I didn't. The thing is sometimes it is quite complex to know why you are refusing some people developments and the way you do it. It's sometimes a little bit harsh. But IMHO just incomprehension/communication problem. Let's go further and have a beer :smiley:

pedrobaeza:
Adding points to the debate is not refusing/blocking them. @sebastienbeau can say for example that I have told him about some concerns in their PRs, but don't block and even blessing them after the corresponding explanations. I think I have been very patient in this PR with all your concerns as well.

sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_sales_order_auto_done_setting" model="ir.ui.view">
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_sales_order_auto_done_setting" />
```
Cedric-Pigeon, line 5: @pedrobaeza Why did you change the original anchor for inheritance? See https://github.com/OCA/sale-workflow/pull/1155/files#diff-5a3cebb009ef27460435af3553b111aaR6
image

Is there a particular functional reason? Because if you do not activate the option "Lock Confirmed Sales", users won't see the recalculation buttons.

pedrobaeza: Sorry for the late answer. Previous xpath was not working anymore since @legalsylvain's change in  38ffb1389098431d63887a47ff8dbde19042d15e (which both the one you mentioned -  #1155 - and #1245 don't include) and I looked for an alternative without checking the point you mention. I can change it to a more suitable one if this finally goes on.

chienandalu (approved):
This has been working on production for months already. I think it can be safely merged

chienandalu:
The merge didn't success at the moment :confused: Can you recall the bots?

chienandalu:
I didn't merge this time either...

chienandalu:
At last :)

PR #1240, opened 2020-09-10
[13.0] [FIX] sale_order_type: Only show the type field in sales

nicomacr:
Actualy the field type is showed in all of the invoice types (customer and vendors) however, for the vendors invoices it's not necesary to show it . Also the same apply for the compute method for that field.

Screenshot (25)

### DIFF
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -11,7 +11,9 @@ class TestSaleOrderType(common.TransactionCase):
         super(TestSaleOrderType, self).setUp()
         self.sale_type_model = self.env["sale.order.type"]
         self.sale_order_model = self.env["sale.order"]
-        self.invoice_model = self.env["account.move"]
+        self.invoice_model = self.env["account.move"].with_context(
+            default_type="out_invoice"
+        )
         self.partner = self.env.ref("base.res_partner_1")
         self.partner_child_1 = self.env["res.partner"].create(
             {"name": "Test child", "parent_id": self.partner.id, "sale_type": False}
@@ -27,6 +29,7 @@ class TestSaleOrderType(common.TransactionCase):
         self.journal = self.env["account.journal"].search(
             [("type", "=", "sale")], limit=1
         )
+        self.default_sale_type_id = self.env["sale.order.type"].search([], limit=1)
         self.warehouse = self.env["stock.warehouse"].create(
             {"name": "Warehouse Test", "code": "WT"}
         )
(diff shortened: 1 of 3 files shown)


nicomacr:
@pedrobaeza could you check this out?

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -117,17 +119,24 @@ def test_sale_order_change_partner(self):
 
     def test_invoice_onchange_type(self):
         sale_type = self.sale_type
-        invoice = self.invoice_model.new({"sale_type_id": sale_type.id})
+        invoice = self.invoice_model.with_context(default_type="out_invoice").new(
```
pedrobaeza: Put the `with_context` in the `invoice_model` declaration

sale_order_type/views/account_move_views.xml
```diff
@@ -7,7 +7,7 @@
             <field name="journal_id" position="before">
                 <field
                     name="sale_type_id"
-                    attrs="{'invisible': [('type', 'not in', ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'])]}"
```
pedrobaeza, line 10: I wonder which was the intention of the original contributor including explicitly the vendor bills.

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -27,6 +27,8 @@ def setUp(self):
         self.journal = self.env["account.journal"].search(
             [("type", "=", "sale")], limit=1
         )
+
```
pedrobaeza: Don't add empty lines inside a method.

sale_order_type/views/account_move_views.xml
nicomacr: I was wondering the same

nicomacr:
@pedrobaeza Done

sale_order_type/views/account_move_views.xml
pedrobaeza: @ivantodorovich any specific reason for putting the type also on vendor bills that we are missing?

pedrobaeza (approved):

sale_order_type/views/account_move_views.xml
ivantodorovich: It's a mistake :sweat_smile: 
this change is correct :+1:

ivantodorovich (approved):

HaraldPanten (approved):

PR #1244, opened 2020-09-14
[12.0][FIX] fill type if null by default as it is now required

sergiocorato:
`type_id` is required but can be null for migrated db, this fix it.
n.b.: TODO at-install hook to fill the field? (in other PR)
n.b.: merge with `nobump`

### DIFF
--- a/sale_order_type/migrations/12.0.1.1.1/post-migration.py
+++ b/sale_order_type/migrations/12.0.1.1.1/post-migration.py
@@ -1,0 +1,24 @@
+# Copyright 2020 Sergio Corato <https://github.com/sergiocorato>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def _fill_sale_order_type(cr):
+    _logger.info("Fill type_id in sale order if null as it is now required")
+
+    cr.execute(
+        """
+    UPDATE sale_order
+    SET
+        type_id=(SELECT id FROM sale_order_type ORDER BY id LIMIT 1)
+    WHERE
+        type_id is null
+        """
+    )
+
+
+def migrate(cr, version):
+    _fill_sale_order_type(cr)
--- a/sale_order_type/__manifest__.py
+++ b/sale_order_type/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     "name": "Sale Order Type",
-    "version": "12.0.1.1.0",
+    "version": "12.0.1.1.1",
     "category": "Sales Management",
     "author": "Grupo Vermon,"
               "AvanzOSC,"


sale_order_type/migrations/12.0.1.1.1/post-migration.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 Sergio Corato <https://github.com/sergiocorato>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from openupgradelib import openupgrade
```
astirpe: When upgrading this module, from `12.0.1.1.0` to `12.0.1.1.1`, you are assuming that `openupgradelib` is installed in the system. But in most of the cases, for example in odoo.sh, this will cause an error.
Could you avoid the use of `openupgradelib`?

sergiocorato: And what about the other n OCA modules that use `openupgradelib`?

astirpe: Same discourse.
The use of `openupgradelib` should be avoided for just bug fixing in general.

sergiocorato: As this method is used everywhere for patch migration in OCA, I wonder if this is a new guideline generally agreed by OCA contributors.

pedrobaeza: AFAIK, odoo.sh can import openupgradelib, so I don't see it as a big deal, isn't it?

astirpe: odoo.sh was just an example. Would you force all the customer running a live instance of the module to install openupgradelib just because they upgraded the module? Expect someone opening an issue one day, we already got such kind of issues in the past. Here is one https://github.com/OCA/hr/issues/754

In this particular PR, the removal of the openupgradelib is quite simple, so I would recommend it.

pedrobaeza: Well, it's the balance between the amount of code that you need to add for avoiding `openupgradelib`. Most OCA users are aware of this lib and have it installed as it's useful for lot of things, but I'm OK if someone wants to avoid as well.

sergiocorato: I remove it without problem, however it seems difficult to me not to have it installed.

astirpe: In this particular case I don't see so much increasing amount of code.
And we should not always assume that OCA users are technically prepared to prevent such issues when just upgrading a module.

sergiocorato: Done!

astirpe (approved):

pedrobaeza (approved):

marcelsavegnago (approved):

PR #1245, opened 2020-09-14
[13.0][MIG] sale_order_price_recalculation with onchange_helper

rousseldenis:
Reopening of https://github.com/OCA/sale-workflow/pull/1155 for testing


legalsylvain:
Hi ! don't forget to cherry pick https://github.com/OCA/sale-workflow/pull/1232 and adapt to V13. Thanks !

rousseldenis:
> Hi ! don't forget to cherry pick #1232 and adapt to V13. Thanks !

Ok, I try to make it green first, fix @simahawk remarks, then yours. We're quite challenging here :smiley:

danimv5 (approved):
Functional review! LGTM

eduardgm (approved):
Functionally checked, LGTM 👍🏻

legalsylvain:
Hi @Cedric-Pigeon. I don't understand your last commit.
I don't understand why protect a feature that can be done manually, without the group. It will only make users loosing time. don't you think ?

Cedric-Pigeon:
> Hi @Cedric-Pigeon. I don't understand your last commit.
> I don't understand why protect a feature that can be done manually, without the group. It will only make users loosing time. don't you think ?

I had issues on big sale orders where users changed the pricelist and clic on recalculate prices without being allowed to do it.
So I thought it could be the case for other people and made this commit.
If it is a problem I can only do it on my project.

legalsylvain:
Well, I understand your need. however I see it as a regression.

- people installing the module expect that the feature is enabled. With such commit, user have to affect the group to all existing users, to make the features working.
- If a user has the possibility to change the pricelist and add / update sale.order.line, so he should have the possilibility to use the "recalculate prices", for consistency reasons.

so your need is valid, but I see it as a custom need, that should be set in the customer project.

Cedric-Pigeon:
@legalsylvain Ok, I understand your point of view. I'll rollback this commit and put it in my project.

gva-acsone (approved):
LGTM Functional test

wtaferner:
@rousseldenis @pedrobaeza 
Any reason why it is still marked as draft?
I have read now the discussions on https://github.com/OCA/sale-workflow/pull/1239, but I guess this one or the other can be finally merged after resolving the conflict, right? Not sure if stalling over a dispute would help the community and anyway, the onchanges are removed/replaced in 14.0 in Odoo Core (or to put it differently need the computed field workaround)

I like the module, so it should be part of 13.0 asap instead of splitting by using one or the other "fork".

pedrobaeza:
As you know, I prefer the other one as it was the same as in v12, but I gave up on this and use my PR in my deployments.

bealdav:
I confirm that implementation using Form is very very very slow in my customer instance with account_invoice_intercompany with big list of products in invoice.

Please do no use it again in OCA modules. Onchange helper is really more eficient (previously implemented in account_invoice_intercompany).

Just a non opinionated customer feedback ! Thanks

mourad-ehm (approved):
I prefer Implementation using onchange_helper because using Form causes perf propblem.

rousseldenis:
@bealdav @mourad-ehm @wtaferner I reactivated this one as it is currently migrating to 14.

rousseldenis:
sale_order_general_discount has changed discount field behaviour making it computed.

https://github.com/OCA/sale-workflow/pull/1281

IMHO the way done there is too restrictive!

manuelcalerosolis:
Please review Travis ans CI errors.

tafaRU:
@rousseldenis just noticed that `sale_order_price_recalculation` was already migrated to `13.0` so this PR would seem to close, isn'it?

rousseldenis:
> @rousseldenis just noticed that `sale_order_price_recalculation` was already migrated to `13.0` so this PR would seem to close, isn'it?

Yes, it is.

People who need performances should choose this anyway. :-)

pedrobaeza:
Merged option is more performant AFAIK (it's not using Form).

PR #1249, opened 2020-09-17
[13.0][IMP] sale_order_line_packaging_qty - packaging selection

TDu:
On sale order line filter out packaging that have a quantity of zero.
Also remove the option to create a new packaging from the sale order
line.

### DIFF
--- a/sale_order_line_packaging_qty/views/sale_order.xml
+++ b/sale_order_line_packaging_qty/views/sale_order.xml
@@ -38,9 +38,10 @@
                     name="product_packaging"
                     attrs="{'invisible': [('product_id', '=', False)]}"
                     context="{'default_product_id': product_id, 'tree_view_ref':'product.product_packaging_tree_view', 'form_view_ref':'product.product_packaging_form_view'}"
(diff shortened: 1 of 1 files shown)


sale_order_line_packaging_qty/views/sale_order.xml
```diff
@@ -38,9 +38,10 @@
                     name="product_packaging"
                     attrs="{'invisible': [('product_id', '=', False)]}"
                     context="{'default_product_id': product_id, 'tree_view_ref':'product.product_packaging_tree_view', 'form_view_ref':'product.product_packaging_form_view'}"
-                    domain="[('product_id','=',product_id)]"
+                    domain="[('product_id','=',product_id), ('qty', '>', 0)]"
```
rousseldenis, line 41: In which case do you have packaging == 0?

TDu: Not clear how, but thanks for asking the why ? After internal discussion this is not needed any more. And it would be better to fix the data model with a constraint not allowing zero quantity on the product.packaging.

PR #1251, opened 2020-09-21
[13.0][ADD] sale_order_product_availability_inline: new module

ernestotejeda:
Show product availability (free_qty) in sales order lines product drop-down.

Screenshot 2020-09-22 18:04:40

cc @Tecnativa TT25486


pedrobaeza:
Please remove `Free to use: ` word there, and I'm debating about putting before or after the quantity between parenthesis (as there are very long product names).

ernestotejeda:
@pedrobaeza , `Free to use:` removed.


> I'm debating about putting before or after the quantity between parenthesis (as there are very long product names).

Maybe before is better but I'm not sure so let me know if you think it's better.

rafaelbn (approved):
Functional test un runbot

sale_order_product_availability_inline/views/sale_views.xml
```diff
@@ -0,0 +1,44 @@
...
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/tree//field[@name='product_id']"
+                position="attributes"
+            >
+                <attribute name="context">{
+                        'so_product_stock_inline': True,
```
youring: IMHO, is it better to using python_dict(depends on base_view_inheritance_extension) to append context, other than overriding? This way keeps other modules' context working.

pedrobaeza: Can you please do a PR against our branch for adding this?

victoralmau (approved):

HaraldPanten (changes requested):
Units availability are shown in the quotation report. IMO they shouldn't appear there, as it's an internal information and not an information for customers.

Steps:

1- Create a quotation for a random partner.
2- Add several lines with several stockable products. These products need to have several "free to use" stock units.
3- Save and print the quotation.

**This is the result:**

Captura de pantalla 2020-09-22 a las 9 57 28

pedrobaeza:
The problem is not printing the report, but on onchange product that keeps the context and thus shows display_name with quantities. @ernestotejeda please overwrite the onchange and reset that context key before calling to super.

ernestotejeda:
@pedrobaeza , @HaraldPanten  Changes done

sale_order_product_availability_inline/views/sale_views.xml
youring: Done. Just did my first PR to OCA, I'm not a developer, correct me if wrong.

pedrobaeza: I have just merged it. You can check if it works soon on runbot.

youring:
One not funcional related question here: why the free_qty field shows one decimal place, not respecting the digits (3) of "Product Unit of Measure"?

pedrobaeza:
Because the quantity is shown as delivered in Python. Expect that if you have 3 decimals, they will be shown as well.

HaraldPanten (approved):
Functional review 👍

youring:
> 
> 
> Because the quantity is shown as delivered in Python. Expect that if you have 3 decimals, they will be shown as well.

Then how to show as integer while digit for "Product Unit of Measure" is 0?

pedrobaeza:
You need to investigate that, as we don't have more budget for this.

PR #1252, opened 2020-09-22
[13.0][IMP] sale_automatic_workflow, add _do function

kittiu:
Call _do_xxx function, i.e., _do_validate_sale_order, as this is required if we need to use queue_job later on.
```
    def _do_validate_sale_order(self, sale):
        """Validate a sales order"""
        with savepoint(self.env.cr), force_company(self.env, sale.company_id):
            sale.action_confirm()

    @api.model
    def _validate_sale_orders(self, order_filter):
        sale_obj = self.env["sale.order"]
        sales = sale_obj.search(order_filter)
        _logger.debug("Sale Orders to validate: %s", sales.ids)
        for sale in sales:
            self._do_validate_sale_order(sale)
```

### DIFF
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -45,6 +45,15 @@ class AutomaticWorkflowJob(models.Model):
         " invoices, pickings..."
     )
 
+    def _do_validate_sale_order(self, sale, domain_filter):
+        """Validate a sales order, filter ensure no duplication"""
+        if not self.env["sale.order"].search_count(
+            [("id", "=", sale.id)] + domain_filter
+        ):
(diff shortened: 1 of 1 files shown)


kittiu:
After looking around, my guess is that, the first [13.0][MIG] of this module was following v11
https://github.com/OCA/sale-workflow/blob/11.0/sale_automatic_workflow/models/automatic_workflow_job.py#L44

While, the v12 which already has job already has _do_xxxx function to be used in `sale_automatic_workflow_job`

So, I think this IMP will bring the job capability back.

rousseldenis (approved):

newtratip (approved):

kittiu:
@OCA/crm-sales-marketing-maintainers

PR #1255, opened 2020-09-23
[13.0][MIG] sale_order_action_invoice_create_hook

SurachetS:


sale_order_action_invoice_create_hook/model/sale_order.py
```diff
@@ -93,8 +92,8 @@ def new_action_invoice_create(self, grouped=False, final=False):
                         invoice = invoices[group_key]
                         # END HOOK
                         vals = {}
-                        if order.name not in invoice.origin.split(", "):
-                            vals["origin"] = invoice.origin + ", " + order.name
+                        if order.name not in invoice.invoice_origin.split(", "):
+                            vals["origin"] = invoice.invoice_origin + ", " + order.name
```
willer-mo, line 96:
```suggestion
                            vals["invoice_origin"] = invoice.invoice_origin + ", " + order.name
```

sale_order_action_invoice_create_hook/model/sale_order.py
```diff
@@ -104,13 +103,19 @@ def new_action_invoice_create(self, grouped=False, final=False):
...
+                                }
                             )
                             pending_section = None
-                        line.invoice_line_create(
-                            invoices[group_key].id, line.qty_to_invoice
+                        invoice.write(
+                            {"invoice_line_ids": [(0, 0, line._prepare_invoice_line())]}
                         )
+
+                        # pp(line._prepare_invoice_line())
```
willer-mo, line 118:
```suggestion
```
Remove comment

sale_order_action_invoice_create_hook/model/sale_order.py
```diff
@@ -137,27 +145,30 @@ def new_action_invoice_create(self, grouped=False, final=False):
                     for line in invoice.invoice_line_ids:
                         line.quantity = -line.quantity
                 # Use additional field helper function (for account extensions)
-                for line in invoice.invoice_line_ids:
-                    line._set_additional_fields(invoice)
+                # for line in invoice.invoice_line_ids:
+                #    line._set_additional_fields(invoice)
```
willer-mo: If that code is not going to be used remove it too.

etobella:
The current function of odoo has changed a lot, I think you are loosing part of its logic

pedrobaeza:
Is this still needed? For grouping criteria I did a module that don't need this one.

etobella:
@pedrobaeza Which module?

pedrobaeza:
https://odoo-community.org/shop/product/sales-order-invoicing-grouping-criteria-5103

etobella:
interesting approach!! The main difference between this modules is done when you are trying to attach the sale order to an existing opened invoice. This module tried to solve this kind of situations (and inserted several hooks)

pedrobaeza:
Yeah, but the hooks don't worth if you don't use them. About attaching to the same invoice, I don't like that approach sincerely, but I think we have already talked about this.

PR #1253, opened 2020-09-22
[13.0][MIG] sale_automatic_workflow_job

kittiu:
Standard migration

Depends on:
- [ ] https://github.com/OCA/sale-workflow/pull/1252
- [ ] https://github.com/OCA/queue/pull/255 (add @job_auto_delay)

PR #1256, opened 2020-09-23
[REF] sale_automatic_workflow|_stock: separate

SButko:
There is an issue for service companies that do not need the stock
module to be installed when using `sale_automatic_workflow`. Module
depended on `stock`, therefore it was auto installed. Now, stock related
part is moved to a separate module - `sale_automatic_workflow_stock` and
can be installed separately.


rousseldenis:
Indeed, that's a caveat as it's better to split different functional parts in different modules.

I'm not sure this refactoring should be merged in stable version. People don't manage every time module versions in their projects.

mmequignon:
@rousseldenis that pr supersedes this one : https://github.com/OCA/sale-workflow/pull/1257

kittiu:
Hi, I also agree that this can be split. But not sure it should be done in this version per say. V14 is out, should it be there?
In any case, if you will proceed (and if the maintainer allowed), we are fixing issues when dealing with job, with sale_automatic_workflow_job. Can you check this?
https://github.com/OCA/sale-workflow/pull/1252
https://github.com/OCA/sale-workflow/pull/1253
Note: the IMP was required in case dealing with big amount of sale orders, where each job should do a bit thing, or the system will hault.
Thanks!

PR #1257, opened 2020-09-24
[REF] sale-automatic-workflow

mmequignon:
replaces https://github.com/OCA/sale-workflow/pull/1256

description :

There is an issue for service companies that do not need the stock
module to be installed when using sale_automatic_workflow. Module
depended on stock, therefore it was auto installed. Now, stock related
part is moved to a separate module - sale_automatic_workflow_stock and
can be installed separately.


i-vyshnevska (approved):

sale_automatic_workflow/tests/test_automatic_workflow.py
```diff
@@ -13,33 +13,19 @@
...
         self.assertEqual(sale.state, "sale")
-        self.assertTrue(sale.picking_ids)
         self.assertTrue(sale.invoice_ids)
         invoice = sale.invoice_ids
         self.assertEqual(invoice.state, "posted")
-        picking = sale.picking_ids
-        self.run_job()
-        self.assertEqual(picking.state, "done")
-
-    def test_onchange(self):
```
grindtildeath, line 32: I guess we could keep this test on the onchange and rely on the `team_id` field instead of `picking_policy`

sale_automatic_workflow/views/sale_order_views.xml
```diff
@@ -19,14 +19,4 @@
             </xpath>
         </field>
     </record>
-    <record id="view_order_form_inherit_sale_stock" model="ir.ui.view">
-        <field name="name">sale.order.form.sale.stock</field>
-        <field name="model">sale.order</field>
-        <field name="inherit_id" ref="sale_stock.view_order_form_inherit_sale_stock" />
-        <field name="arch" type="xml">
-            <xpath expr="//group[@name='sale_shipping']" position="inside">
-                <field name="all_qty_delivered" />
```
grindtildeath, line 28: I'm not sure this field must be removed from the view as it's still in the model. But as service product can still be delivered through timesheets without a picking, I guess we still need it.

BTW: `_compute_all_qty_delivered` might need some fixing.

sale_automatic_workflow_stock/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,38 @@
...
+        for picking in pickings:
+            with savepoint(self.env.cr):
+                picking.validate_picking()
+
+    @api.model
+    def run_with_workflow(self, sale_workflow):
+        """Override to add stock picking validation."""
+        super().run_with_workflow(sale_workflow)
+        workflow_domain = [("workflow_process_id", "=", sale_workflow.id)]
+        if sale_workflow.validate_picking:
```
grindtildeath, line 35: There might be an issue here if the picking is validated after the creation of the invoice (depending on the invoice policy of products).
Looks like we need a hook in the super method to call picking validation between the validation of sale order and the creation of the invoice.

grindtildeath (changes requested):
Looks very good overall, just two small things

sale_automatic_workflow/views/sale_order_views.xml
mmequignon: It's because `sale_stock` isn't installed anymore, so that overriding does not makes sense anymore.

rousseldenis:
@mmequignon Is this still applicable as there have been already changes since migration ?

PR #1265, opened 2020-09-30
[12.0][ADD] sale_order_line_serial_unique

rousseldenis:


lmignon:
@rousseldenis I Why do you need to restrict the number of products on the SOL. The serial number is not linked to the SOL?

HaraldPanten (approved):
LGTM

len-foss:
Why not replace this by auto_split? 
If you forced to input the serial number per line then the constraint would make sense; however here you want to split them for invoicing (technical) purposes.
So you could for example simply split the line when saving instead of breaking the user-flow.

TimLai125:
I think the module might make more sense if we add a new numeric field to product to control this constraints instead of using tracking field which could be more flexible, e.g. the quantity of the sales line cannot exceed the set number.

PR #1267, opened 2020-09-30
[12.0][ADD] sale_expense_receipt

Khalid-SerpentCS:
Sale Expense Receipt
===================================================================================================
This module allows you to attach all the receipts of the expenses when invoicing a sale order and send them with the invoice.

Usage
=================================================================================================
* Go to Expenses
* Create an expense with a product that is re-invoiced (Re-invoicing policy: At cost)
* Select a sale order and attach the receipt
* Submit, approve and post the journal entries
* Go to Sales > Sales Orders
* Open the sale orders selected on the expense and create the invoice. A file "Expense_Receipt.zip" should be attached to the 
  invoice and should contain the receipt of the expense.


sale_expense_receipt/models/sale.py
```diff
@@ -0,0 +1,64 @@
...
+                lambda e: e.sheet_id.state == 'post' and
+                e.product_id == line.product_id and
+                e.unit_amount == line.price_unit)
+
+        attachments = self.env['ir.attachment'].search([
+            ('res_id', 'in', expenses.ids)])
+
+        if attachments:
+            file_path = pathlib.Path(temp_dir + '/Expense_Receipt.zip')
+            receipt_zip_file = zipfile.ZipFile(
```
len-foss, line 35: Why not use a `with` here?

sale_expense_receipt/models/sale.py
```diff
@@ -0,0 +1,64 @@
...
+                file_path, mode='w')
+            for attachment in attachments:
+                receipt_zip_file.writestr(attachment.name,
+                                          base64.b64decode(attachment.datas))
```
len-foss, line 39: Maybe `from base64 import b64decode` to lighten a bit the lines?

len-foss (approved):
Code review: LGTM
However, you would need to add some tests to pass the required code coverage.

rousseldenis:
@JayVora-SerpentCS

JayVora-SerpentCS:
> @JayVora-SerpentCS

do we need this feature? It was long back migrated.

PR #1276, opened 2020-10-05
[13.0][FIX] sale_order_line_packaging_qty

TDu:
When a quantity is set on the sale order line before or at the same time
than the product itself, the onchange triggered will replace the line
set quantity with the default size of the package.
This is especially problematic for sales order created programmatically.
So to improve the solution, if the quantity fits with the package
quantity it is kept.

### DIFF
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -39,7 +39,9 @@ class SaleOrderLine(models.Model):
             }
         return super()._onchange_product_packaging()
 
-    @api.constrains("product_id", "product_packaging", "product_packaging_qty")
+    @api.constrains(
+        "product_id", "product_packaging", "product_packaging_qty", "product_uom_qty"
+    )
     def _check_product_packaging_sell_only_by_packaging(self):
         for line in self:
             if not line.product_id.sell_only_by_packaging:
@@ -90,15 +92,20 @@ class SaleOrderLine(models.Model):
             SaleOrderLine, self.with_context(_skip_auto_assign=True)
         )._inverse_product_packaging_qty()
 
+    def _inverse_qty_delivered(self):
+        # Force skipping of auto assign
+        super(
+            SaleOrderLine, self.with_context(_skip_auto_assign=True)
+        )._inverse_qty_delivered()
+
     def write(self, vals):
         """Auto assign packaging if needed"""
-        if vals.get("product_packaging") or self.env.context.get("_skip_auto_assign"):
+        if "product_packaging" in vals.keys() or self.env.context.get(
+            "_skip_auto_assign"
+        ):
             # setting the packaging directly, skip auto assign
             return super().write(vals)
         for line in self:
-            if line.product_packaging:
-                # don't touch it if set already
-                continue
             line_vals = vals.copy()
             line_vals.update(line._write_auto_assign_packaging(line_vals))
             super(SaleOrderLine, line).write(line_vals)
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -106,12 +106,17 @@ class SaleOrderLine(models.Model):
             qty = uom._compute_quantity(qty, product.uom_id)
         return product.get_first_packaging_with_multiple_qty(qty)
 
+    def _inverse_product_packaging_qty(self):
+        # Force skipping of auto assign
+        # if we are writing the product_uom_qty directly via inverse
+        super(
+            SaleOrderLine, self.with_context(_skip_auto_assign=True)
+        )._inverse_product_packaging_qty()
+
     def write(self, vals):
         """Auto assign packaging if needed"""
-        fields_to_check = ["product_id", "product_uom_qty", "product_uom"]
-        if vals.get("product_packaging") or not any(
-            fname in vals for fname in fields_to_check
-        ):
+        if vals.get("product_packaging") or self.env.context.get("_skip_auto_assign"):
+            # setting the packaging directly, skip auto assign
             return super().write(vals)
         for line in self:
             if line.product_packaging:
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -109,8 +109,8 @@ class SaleOrderLine(models.Model):
                     "title": _("Product quantity cannot be packed"),
                     "message": _(
                         "For the product {prod}\n"
-                        "The {qty} is not the multiple of any pack.\n"
-                        "Please add a package"
+                        "The quantity {qty} is not the multiple of any packaging.\n"
+                        "Please add a packaging."
                     ).format(prod=self.product_id.name, qty=self.product_uom_qty),
                 }
                 return {"warning": warning_msg}
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -72,12 +72,7 @@ class SaleOrderLine(models.Model):
                 )
             )
             if first_packaging:
-                self.update(
-                    {
-                        "product_packaging": first_packaging.id,
-                        "product_uom_qty": first_packaging.qty,
-                    }
-                )
+                self.update({"product_packaging": first_packaging.id})
         return res
 
     def _force_qty_with_package(self):
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -60,10 +60,6 @@ class SaleOrderLine(models.Model):
                     )
                 )
 
-    @api.onchange("product_id")
-    def product_id_change(self):
-        return super().product_id_change()
-
     def _force_qty_with_package(self):
         """
 
--- a/sale_by_packaging/models/sale_order_line.py
+++ b/sale_by_packaging/models/sale_order_line.py
@@ -1,8 +1,8 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from odoo import _, api, fields, models
+from odoo import _, api, models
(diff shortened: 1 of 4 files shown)


sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -66,10 +66,16 @@ def _onchange_product_packaging_qty(self):
     @api.onchange("product_packaging")
     def _onchange_product_packaging(self):
         if self.product_packaging:
+            pack_qty = 1
```
lmignon, line 75: @TDu Why don't you rely on the previous state of your record available provided by `self._origin`? You could therefore check `if self.product_uom_qty == self._origin.product_uom_qty:`

TDu: @lmignon Because I did not think of it. So I guess you propose to do the check if existing product_uom_qty can be kept with the new packaging after the update (to save on some variable creation) ?
Well I did some tests on this but funny thing is, in my tests the `onchange` is not called following a change in a Form but by a `play_onchanges` function and there is no record in `self._origin` on every calll !?

lmignon: @TDu I suppose that the play_onchanges is called on a record initialized with self.new()... If it's the case, to be consistent with odoo, the self.new() should be called with the origin args.... see https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5510.

TDu: @lmignon Thanks for your answer, from what I can tell the sale.order is created with the `create` function. Done in an other OCA module https://github.com/OCA/edi/blob/cc04afc899246919a4ac90373d92ced0a398641c/sale_order_import/wizard/sale_order_import.py#L248

lmignon: Something is missing into this puzzle... Nevertheless your proposed change makes sense even if I would prefer to not rely on some implicit rules to determine the context in which the onchange method is called. I'm curious to discover the method triggering the call to this onchange....

guewen: `play_onchanges` comes from: https://github.com/OCA/server-tools/blob/5e432fdd9b33a18badff88119e9d8f0791736369/onchange_helper/models/base.py#L24

Is it possible that we have no `_origin` when going through this method?

lmignon: @guewen Indeed, IMO, whe should initialize the "_origin" field on the current recordset before the call to `onchange` In this way we are consistent with the way the onchange methods are called from the UI. https://github.com/OCA/server-tools/blob/13.0/onchange_helper/models/base.py#L57
somethng like:

```python
    ....
     with self._set_origin():
         for field in onhcange_fields:
            onchange_values = self.onchange(all_values, field, onchange_specs)
    ...

@contextmanager
def _set_origin(self):
    if self:
        self.ensure_one()
        setattr(self, '_origin', self)
    yield
    if self:
        delattr(self, "_origin")
```

lmignon: @guewen I'm not sure this code will work in odoo 13. The poposed PR is probably the safer way to solve your issue

lmignon: Reading the code in models.py for odoo13. I'm sure that the proposed change will not work.

jbaudoux: @TDu I tested with sale_order_import and the qty is wrong and/or the deducted packaging is wrong.

In my tests, I have a product with a retail box=22 units and transport box of 176 units.
Test 1: I order a qty not multiple of a packaging. Then SO line qty is always 1 retail box instead of the order qty
Test 2: I order a qty multiple of a transport box. Then SO line packaging is not expressed with the biggest packaging (transport box) but as a multiple of retail box.

When doing those 2 tests with odoo SO interface, it works properly

TDu: @jbaudoux In the UI if the quantity is set first (with not a multiple of packaging quantity) and then the product is set, then the quantity is lost and 1 retail box is selected and the quantity contains the retail box quantity.

TDu: This is what happens on the import. To change this behaviour would mean that a default package is not selected automatically. As by default Odoo sets 1 as a quantity I see no way to know if this is a default 1 or required 1 on the line.

jbaudoux: @TDu I would remove the default packaging. It is not convenient in the UI as you do not see the possibilities without emptying the field. So better leave it empty by default and the user is free to select one. It will also be easier this way if the user does not want to select one as emptying that field is not that easy in the UI (it sometimes comes back).

TDu: @jbaudoux I checked the code and the change happens in this onchange https://github.com/OCA/sale-workflow/blob/35feda502a35c55f0074232da1041926c0277902/sale_by_packaging/models/sale_order_line.py#L56
So only if the product is to be sold only by packaging. (This was the case for the product I used for my test).
Disabling the `Only sell by packaging`  on the product and the problem is gone.
So it is actually a feature of the module.

jbaudoux: @TDu sell_only_by_packaging should require a packaging in the UI but not fill with default value

sale_by_packaging/models/sale_order_line.py
```diff
@@ -52,18 +51,7 @@ def _check_product_packaging_sell_only_by_packaging(self):
...
-                self.product_id.packaging_ids.filtered(
-                    lambda p: not float_is_zero(
-                        p.qty, precision_rounding=p.product_uom_id.rounding
-                    )
-                )
-            )
-            if first_packaging:
-                self.update({"product_packaging": first_packaging.id})
-        return res
+        return super().product_id_change()
```
i-vyshnevska: why not remove it completely?

TDu: Yes why not ? Removed now.

.pre-commit-config.yaml
```diff
@@ -18,8 +18,8 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-  - repo: https://github.com/prettier/pre-commit
-    rev: "v1.19.1"
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v2.1.2"
```
TDu: This was to get the pre-commit hooks working again locally

guewen: reference: https://github.com/OCA/oca-addons-repo-template/pull/41

sale_order_line_packaging_qty/models/sale_order_line.py
guewen: What is the status here? Is there something to change or is this thread resolved?

TDu: Yes by this https://github.com/OCA/sale-workflow/pull/1276/commits/6c291697db159c6c7d518526bae1eba43b9c7ccc

TDu: It is a commit below

TDu: The requirement is done by a constraint

.pre-commit-config.yaml
yajo: You can see here that prettier was working fine with the old config: https://travis-ci.com/github/OCA/sale-workflow/jobs/443358674#L284-L369

Could you please leave it as it was?

simahawk:
@TDu what's missing here?

TDu:
@simahawk Removing the change on the pre-commit config, done now.

simahawk:
@TDu https://github.com/TDu/sale-workflow/pull/1 integrates https://github.com/OCA/sale-workflow/pull/1231

sale_by_packaging/tests/test_sale_only_by_packaging.py
```diff
@@ -121,8 +66,11 @@ def test_create_auto_fill_packaging(self):
                 "product_uom_qty": self.packaging.qty * 2,
             }
         )
-        self.assertFalse(order_line_1.product_packaging)
-        self.assertFalse(order_line_1.product_packaging_qty)
+        # Why should it be false if the quantity fits ?
```
simahawk: @TDu is this something still needed?

simahawk (approved):
LG

sale_by_packaging/tests/test_sale_only_by_packaging.py
TDu: I kept it commented because, I did not understand the assertion.
For me it can be removed.

TDu: removed

sale_by_packaging/models/sale_order_line.py
```diff
@@ -96,42 +78,26 @@ def _force_qty_with_package(self):
...
-                warning_msg = {
-                    "title": _("Product quantity cannot be packed"),
-                    "message": _(
-                        "For the product {prod}\n"
-                        "The {qty} is not the multiple of any pack.\n"
-                        "Please add a package"
-                    ).format(prod=self.product_id.name, qty=self.product_uom_qty),
-                }
-                return {"warning": warning_msg}
-        return {}
```
sebalix, line 122: As this code is removed, is the documentation needs to be updated?
https://github.com/OCA/sale-workflow/blob/13.0/sale_by_packaging/readme/DESCRIPTION.rst

TDu: Readme has been updated.

sebalix (approved):

simahawk:
@TDu pls squash

TDu:
Squashed, but only the fixup.

Issue #1260, opened 2020-09-28
[13.0] sale_order_type: sale_type_id empty on invoices


rgl1234:
Is it by design that the field "sale_type_id" stays empty after creating an invoice?
![image](https://user-images.githubusercontent.com/11720329/94402677-a1f1b480-016c-11eb-885a-98f1f20d82ca.png)

pedrobaeza:
Complete steps to reproduce the "problem"?

rgl1234:
1) create SO with "Type = A"
2) deliver products
3) create invoice
4) "Type = empty" on invoice created from SO, expected would be "Type = A" also on invoice.....

nikul-serpentcs:
@rgl1234 Yes, You are right.

@pedrobaeza I think here we add a condition, please check https://github.com/OCA/sale-workflow/blob/13.0/sale_order_type/models/account_move.py#L25

when we added a customer that time not set sale type then we manage like that.
`if not record.partner_id` -> `if not record.partner_id and not record.partner_id.sale_type`

When it's  approved, I will create PR for that

pedrobaeza:
Well, according to that, the first type record will be selected, not empty.

PR #1280, opened 2020-10-07
[13.0] [FIX] sale_order_type: sale_type_id empty on invoices

nikul-serpentcs:
Issue #1260


pedrobaeza:
Travis is red, and not sure this is the ideal solution. I will check the issue personally later for a similar case in one of my customers.

PR #1287, opened 2020-10-12
[IMP] sale_order_line_date: improving commitment_date tree column

dennybiasiolli:
allowing users to hide the column from the tree

### DIFF
--- a/sale_order_line_date/views/sale_order_view.xml
+++ b/sale_order_line_date/views/sale_order_view.xml
@@ -20,7 +20,7 @@
                 expr="//field[@name='order_line']/tree/field[@name='price_unit']"
                 position="after"
             >
-                <field name="commitment_date" />
+                <field name="commitment_date" optional="show" />
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


len-foss (commented):
AFAICT, optional="show" is not documented anywhere, that's interesting. 
Can you post a before/after screenshot for the feature, since it's a visual change?
(small nitpick, your commit would be an [IMP] not an [UPD])

dennybiasiolli:
@len-foss changed commit message ;)

The `optional="show"` parameters allows to act in tree view columns in this way:

- always on, when not specified in tree view field

- default on, but hideable, when `optional="show"`

- default off, but showable, when `optional="hide"`

You can find a detailed explanation here: https://www.youtube.com/watch?v=UfWNwupq2CE
This PR allows users to hide the column `commitment_date` from the sale order lines for improving appearance of the table.

Before:
image


After:
image

len-foss (approved):
Thanks for the info!

rousseldenis (approved):
LGTM. Tested on runbot

bodedra (approved):

lmignon (approved):

PR #1288, opened 2020-10-12
[14.0] [MIG] sale_discount_display_amount module

chafique-delli:


tonygalmiche (approved):
LGTM : Code review and functional review with runbot

image

sale_discount_display_amount/hooks.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2018 ACSONE SA/NV
```
TimLai125, line 1: Why hook is needed for these fields? Shouldn't Odoo already take care of computing the initial values of those fields during installation?

TimLai125 (changes requested):
I tested the module, yet I think the presentation of the fields are a bit confusing to the users.

By default, user interprets Untaxed Amount + Taxes = Total, after adding two fields in the middle, it may give false impression to users that those two fields are also contributing in the calculation of Total. Maybe we can put somewhere after the "Total" field like Margin?
image

danimv5 (approved):
Functional review!! LGTM!!

sale_discount_display_amount/hooks.py
rousseldenis: I think this is was done because if you have an existing environment with plenty of orders, the compute with ORM is quite long.

rousseldenis:
@chafique-delli Can you squash your commits ?

HaraldPanten (approved):
Tested on runbot LGTM.

bodedra (approved):

lmignon (approved):

rousseldenis:
@TimLai125 Could you update your review ?

TimLai125 (approved):

newtratip (approved):
:+1: LGTM

rousseldenis:
I think there is plenty of approvers :smile: 

/ocabot merge patch

PierrickBrun (approved):

PR #1082, opened 2020-02-29
[MIG] Migrated sale_order_line_price_history to v13.

sudhir-erpharbor:

PR #1289, opened 2020-10-12
[13.0] [MIG] sale_order_line_price_history

dennybiasiolli:
Taking over the PR #1082 as said here: https://github.com/OCA/sale-workflow/pull/1082#issuecomment-707031623

Ref #1082


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

dennybiasiolli:
Thanks @pedrobaeza , sorry but these are my first contributions. I'll take all the steps you linked. Should I make another PR with the correct branch name or I can only "fix" the commits history in this PR?

pedrobaeza:
You can reuse same branch and PR. Take care also about Travis red.

dennybiasiolli:
Thanks @pedrobaeza , everything's ok now.

CasVissers-360ERP (approved):
Functional 👍

ernestotejeda (commented):
@dennybiasiolli I tested it in runbot and it doesn't work as expected. When the wizard is launched, the partner of the sale order must be selected by default. Now, the partner field appears empty and when I select one, even the right partner, the lines disappear.
Could you check that, please?

ernestotejeda (commented):
In the wizard lines the partner column are empty, maybe that has to do with what I mentioned  above

dennybiasiolli:
@ernestotejeda thanks for the feedback, I'm trying in a clean `Odoo 13.0` environment with only `sale_management` module and demo data and it's working as expected.
The `partner` field it's pre-populated with the partner defined in the sale order.

What are the modules that are installed in your environment? So I can reproduce exactly the situation.

ernestotejeda:
I think you are testing it locally and I am testing it in runbot. You commentted that this is your first contrinuton, so you may not have seen the runbot. You can access this in this way:
Peek 2020-10-15 10-30

Now, this is what I see when I test on the runbot:
Peek 2020-10-15 10-22

dennybiasiolli:
Thanks for the detailed explanation :)
I'm checking on runbot, is there a way to take a look at the logs? Because it keeps disconnecting and reconnecting in almost every action I take...
Anyway I created a new customer with two orders and the bug you mentioned appears also on the "New Customer 1", but if I try with an order of "Gemini Furniture" everything works perfectly...maybe it has something conflicting with another module?

dennybiasiolli:
BTW I'm trying locally installing all modules from this repo, but a lot of modules requires external modules to be installed

pedrobaeza (approved):
Tried on local and working, so not sure about the glitch, but it's not by the module itself.

PR #1293, opened 2020-10-13
[FIX][12.0] sale_global_discount: sale order templates compatibility

pedrocs-exo:
If you happen to use line sections or notes on sales order lines, you can now avoid the '... taxes in lines are required.' error and actually use global discounts.


pedrobaeza (approved):
I suppose the same apply to `account_global_discount`.

pedrocs-exo:
Yes, thank you! The related PR is on the way

PR #1295, opened 2020-10-19
[13.0][ADD] sale_by_packaging: add possibility to force the qty during SO line create/update

acsonefho:
Add feature in the module `sale_by_packaging`.

For now, the module display a warning (during onchange) if the qty set in the SO line is not a multiple of the packaging.
This module add the possibility to force the qty of the SO line with this packaging's multiple.

**Ex:**
You set your packaging to sell only by 5 units.
If a user set `qty = 3`, the onchange is triggered and the `qty` will be 5.

### DIFF
--- a/sale_by_packaging/models/product_product.py
+++ b/sale_by_packaging/models/product_product.py
@@ -2,11 +2,40 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import fields, models
-from odoo.tools import float_is_zero
+from odoo.tools import float_compare, float_is_zero, float_round
 
 
 class ProductProduct(models.Model):
     _inherit = "product.product"
+
+    def _convert_packaging_qty(self, qty, uom, packaging):
+        """
+        Convert the given qty with given UoM to the packaging uom.
+        To do that, first transform the qty to the reference UoM and then
+        transform using the packaging UoM.
+        The given qty is not updated if the product has sell_only_by_packaging
+        set to False or if the packaging is not set.
+        Inspired from sale_stock/models.sale_order.py _check_package(...)
+        :param qty: float
+        :return: float
+        """
+        if not self or not packaging:
+            return qty
+        self.ensure_one()
+        if self.sell_only_by_packaging and packaging.force_sale_qty:
+            q = self.uom_id._compute_quantity(packaging.qty, uom)
+            if (
+                qty
+                and q
+                and float_compare(
+                    qty / q,
+                    float_round(qty / q, precision_rounding=1.0),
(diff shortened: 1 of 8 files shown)


guewen:
Doesn't it overlap with https://github.com/OCA/sale-workflow/tree/13.0/sale_order_line_packaging_qty?

lmignon:
@guewen It seems that `sale_by_packaging` depends on `sale_order_line_packaging_qty` https://github.com/OCA/sale-workflow/blob/13.0/sale_by_packaging/__manifest__.py#L14

Cedric-Pigeon:
@guewen no, this PR adds an extra feature to convert automatically the quantity encoded by the user in complete package. Currently, the feature only displays a message.

guewen (approved):
Got it.
Can you add a description of the option in the readme?

sale_by_packaging/models/product_product.py
```diff
@@ -2,12 +2,41 @@
...
+            return qty
+        self.ensure_one()
+        if self.sell_only_by_packaging and packaging.force_sale_qty:
+            q = self.uom_id._compute_quantity(packaging.qty, uom)
+            if (
+                qty
+                and q
+                and float_compare(
+                    qty / q,
+                    float_round(qty / q, precision_rounding=1.0),
```
guewen, line 32: shoudn't `precision_rounding` be the uom rounding?

sale_by_packaging/models/product_product.py
```diff
@@ -2,12 +2,41 @@
...
+                    precision_rounding=0.001,
```
guewen, line 33: shoudn't `precision_rounding` be the uom rounding?

sale_by_packaging/models/product_product.py
guewen, line 32: Why don't you compare only qty vs q?

sale_by_packaging/models/product_product.py
acsonefho: Nope because if I use the UoM's rounding, (often 0.01) it doesn't work as it's too precise.
This part is a copy/paste from Odoo (the message who displays the quantity who should be sold).
And I prefer to keep the same computation.

sale_by_packaging/models/product_product.py
acsonefho: As mentioned before, it's just because it's a copy/paste from Odoo (only the computation part) and I prefer to keep the same computation.

lmignon (approved):
LGTM (Code review only)

Cedric-Pigeon (approved):
Code review and functional tests

rousseldenis (approved):
LGTM

PR #1296, opened 2020-10-19
Update es_ES.po

jorgecurso1:
translation to Spanish


HaraldPanten:
Hi @jorgecurso1 I think you should use weblate for translating terms: https://translation.odoo-community.org/

PR #1301, opened 2020-10-23
[ADD] sale_tiered_pricing: add tiered pricing

len-foss:
This PR adds support for tiered pricing.

~~(It depends on https://github.com/odoo/odoo/pull/60519)~~ already merged!

Info/Usage/... can be found in the description file.


sale_tiered_pricing/models/sale_order_line.py
```diff
@@ -0,0 +1,72 @@
...
+        res = super(SaleOrderLine, self).get_sale_order_line_multiline_description_sale(
+            product
+        )
+        if self.is_tier_priced:
+            tier_rule = self.order_id.pricelist_id.get_tier_rule(product, self)
+            qps = tier_rule.tiered_pricelist_id.get_quantities_and_prices(
+                self.price_unit, self.product_uom, product, self.product_uom_qty
+            )
+            computation = " + ".join("{} x {}".format(q, p) for q, p in qps)
+            tier_description = _("Tiers: {}").format(computation)
```
lmignon, line 44: @len-foss `_()` will not be translated into the partner lang..... How can we force the use of the partner lang when calling the translate method?

lmignon (approved):
LGTM (Code + functional)  A nice to have would be to be able to get access to the tiered definitions from the product....

sale_tiered_pricing/models/sale_order_line.py
len-foss: Damn, I thought getting the product in the right context would do that since it's what's done in standard.
I'll write a test case and check that, thanks for bringing that up.

ccasula:
Test case 1: Failed: SO (Ordered quantity = 0, description field displays 1)
TC001

ccasula:
Test case 2: Failed: SO: Rounding issue in the calculated price (it should be 1.157 instead of 1.156,95).

TC002

len-foss:
Regarding the rounding issue: this is a known problem with this approach (I'd have to check in details the decimal precisions etc to see if there is a way to mitigate this). 
But I believe the trade-off regarding the simplicity that if confers with editing sale orders (no need to go through a wizard) as well as the code complexity if we had to split the lines makes it up for the issue.

For the description, the case "zero units" is degenerate and may add an extra if, but the overall multiline_description_sale is still TODO (a bit too spartan for the moment).

sale_tiered_pricing/models/sale_order_line.py
rousseldenis: Ok, this case is tested

sale_tiered_pricing/wizard/tiered_pricing_wizard.py
```diff
@@ -0,0 +1,108 @@
...
+    tiered_pricing_id = fields.Many2one(
+        "product.pricelist",
+        string="Tiered Pricing",
+        default=_default_tiered_pricing,
+        domain=[("is_tiered_pricing", "=", True)],
+    )
+
+    def create_pricelist_items(self):
+        if not self.pricelist_ids:
+            raise OdooWarning(_("You should select at least one target pricelist!"))
```
rousseldenis, line 75: @len-foss Warning is deprecated. You should use UserError

rousseldenis (changes requested):
This seems great!

Little change

PR #1303, opened 2020-10-27
[13.0][ADD] sale_stock_picking_validation_blocking

qgroulard:
This module adds the opportunity to prevent the validation of delivery order from the SO.

For instance, it can be useful if you wish to reserve the stocks when the SO is confirmed but you want to wait for payment before validating the related pickings.


rousseldenis:
As far as I see, this could be complementary of this no ? https://github.com/OCA/sale-workflow/tree/12.0/sale_stock_picking_blocking

That one would block delivery creation instead.

But, in your case, do you really want to create all the picking flow and block reserved products (even if the time period is long) ?

qgroulard:
> As far as I see, this could be complementary of this no ? https://github.com/OCA/sale-workflow/tree/12.0/sale_stock_picking_blocking
> 
> That one would block delivery creation instead.

It is not quite the same behavior, but yes it could be.
Whether we want to merge these two modules is a matter of more modularity vs more complete modules. I would be glad to read more opinions about that.

> But, in your case, do you really want to create all the picking flow and block reserved products (even if the time period is long) ?

That's what our client needs at least, hopefully they are other clients like ours who could benefit from this module.

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -0,0 +1,43 @@
...
+class TestSaleStock(TransactionCase):
+    def setUp(self):
+        super(TestSaleStock, self).setUp()
+        partner = self.env.ref("base.res_partner_1")
+        product = self.env.ref("product.product_delivery_01")
+        self.sale_order = self.env["sale.order"].create(
+            {
+                "partner_id": partner.id,
+                "partner_invoice_id": partner.id,
+                "partner_shipping_id": partner.id,
```
yostashiro:
```suggestion
```

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -0,0 +1,43 @@
...
+                "order_line": [
+                    (
+                        0,
+                        0,
+                        {
+                            "name": product.name,
+                            "product_id": product.id,
+                            "product_uom_qty": 10.0,
+                            "product_uom": product.uom_id.id,
+                            "price_unit": product.list_price,
```
yostashiro:
```suggestion
```

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -0,0 +1,43 @@
...
+                        },
+                    )
+                ],
+                "pricelist_id": self.env.ref("product.list0").id,
```
yostashiro:
```suggestion
```

sale_stock_picking_validation_blocking/views/sale_order.xml
```diff
@@ -0,0 +1,31 @@
...
+<odoo>
+    <record model="ir.ui.view" id="sale_order_form_view">
+        <field
+            name="name"
+        >sale.order.form (in sale_stock_picking_validation_blocking)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale_stock.view_order_form_inherit_sale_stock" />
+        <field name="arch" type="xml">
+            <button name="action_draft" position="after">
+                <field name="picking_validation_blocked" invisible="1" />
```
yostashiro: Maybe it's better to show this field somewhere on the form. I thought the blocked state was not visually very obvious with the buttons below. Besides these buttons are not visible to people without the sales manager group.

yostashiro (commented):
Nice module. 👍  I may want to backport this to 12.0.  One of our customers does have this requirement with only difference being that it's the financial manager who should be able to unblock the delivery validation.

sale_stock_picking_validation_blocking/views/sale_order.xml
qgroulard: Our customer suggested to track this field, would that do the job for you ?
Thanks for the review.

yostashiro: That could do, but the track message might be buried deep down in case there is a lot of communication on the sales order. Another idea is to add a smart button that's visible to everyone, and show a warning message when someone with no access right clicks on it, if that makes sense.

qgroulard: I've pushed your suggestions.
I didn't quite liked to have a button which you can't click on, but the field is now displayed in the 'Delivery' section (+ tracked).

sale_stock_picking_validation_blocking/views/sale_order.xml
```diff
@@ -0,0 +1,34 @@
...
```
yostashiro: This line can be removed now that the field is added in the lower part.

yostashiro (approved):
LGTM 👍

rousseldenis (approved):
Code review

PR #1305, opened 2020-10-28
[13.0][FIX] sale_sourced_by_line: Fix routes

rousseldenis:
Since Odoo commit https://github.com/odoo/odoo/commit/51b7960f7e30aab471bb846cb890efd95c7b2dcc
the stock rules are enforced for admin user to be filtered
by current company.

So, no convenient rule was found when evaluating https://github.com/odoo/odoo/commit/51b7960f7e30aab471bb846cb890efd95c7b2dcc#diff-a8e268c29293c3896cc7bf77c5a0809a394ffe929388180b6e96cce485177956R434


pedrobaeza:
Thanks for the quick patch

PR #1307, opened 2020-11-02
[13.0][FIX] sale_order_product_recommendation: show the lines as soon as the wizard is launched

ernestotejeda:
cc @Tecnativa TT26388


pedrobaeza (approved):

sergio-teruel (approved):
If run ok... good...
I think that virtual records have not work properly... :smile:

PR #1312, opened 2020-11-03
[12.0][ADD] sale_product_multi_add: include es.po translation

jvargass:
Include translation to Spanish.


pedrobaeza:
Please use translation platform instead:

https://odoo-community.org/page/translate

PR #1318, opened 2020-11-06
[13][IMP] product_set also duplicate lines

TDu:


sale_product_set/models/product_set.py
```diff
@@ -14,7 +14,7 @@ class ProductSet(models.Model):
         string="Internal Reference", help="Product set internal reference", copy=False
     )
     set_line_ids = fields.One2many(
-        "product.set.line", "product_set_id", string="Products"
+        "product.set.line", "product_set_id", string="Products", copy=True
```
rousseldenis, line 17: copy=True by default, so why this ?

TDu: @rousseldenis Yes, copy is True by default except for One2Many and computed fields.

rousseldenis (approved):

bodedra (approved):
LG :+1:

PR #1319, opened 2020-11-06
[13.0] [MIG] sale_manual_delivery

phuctranfxvn:
Migrated from https://github.com/OCA/sale-workflow/pull/987


newtratip:
@rousseldenis Could you change this milestone from 14.0 to 13.0 please ?

sale_manual_delivery/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+        string="Manual Delivery",
+        default=False,
+        help="If Manual, the deliveries are not created at SO confirmation.\
+        You need to use the Create Delivery button in order to reserve and \
+        ship the goods.",
+    )
+
+    @api.onchange("team_id")
+    def _onchange_team_id(self):
+        for so in self:
```
i-vyshnevska: onchange runs on a single record no need to iterate here

sale_manual_delivery/models/stock_move.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2017 Denis Leemann, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    carrier_id = fields.Many2one("delivery.carrier", string="Delivery Method")
```
i-vyshnevska, line 10: I think it is better to have the carrier on stock.picking level, I don't know cases when moves belong to one picking and served by different carriers

phuctranfxvn: Hi @i-vyshnevska
Agree with you. However, this module is migrated from https://github.com/OCA/sale-workflow/pull/987. I think it is better to have this reviewed in mentioned PR first. I'll notify this and tag you to follow up

sale_manual_delivery/models/sale_order.py
phuctranfxvn: fixed in https://github.com/OCA/sale-workflow/pull/1319/commits/c31db4ce252ff937e9014fb4d5626e4116b5e5fe

rousseldenis:
@phuctranerp Do you mind if we switch to #1566 ?

phuctranfxvn:
> @phuctranerp Do you mind if we switch to #1566 ?

@rousseldenis Yes, that's fine for me, thanks

PR #1323, opened 2020-11-09
[ADD] sale_coupon_multi_use: init

oerp-odoo:
Allows to consume same coupon multiple times. Can only be used when
discount has fixed amount. Discount fixed amount can't be changed if
there are coupons with `multi_use` flag set.

Can specify value of coupon and that value can be consumed per multiple
sale orders.

If discount is removed from sale order, so is consumption line from
coupon.

### DIFF
--- a/sale_coupon_multi_use/i18n/fr.po
+++ b/sale_coupon_multi_use/i18n/fr.po
@@ -1,0 +1,129 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#   * sale_coupon_multi_use
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 13.0+e\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-11-06 12:26+0000\n"
+"PO-Revision-Date: 2020-11-06 12:26+0000\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__amount
+msgid "Amount"
+msgstr "Montant"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon__consumption_line_ids
+msgid "Consumption Lines"
+msgstr "Lignes de consommation"
+
+#. module: sale_coupon_multi_use
+#: code:addons/sale_coupon_multi_use/models/sale_coupon.py:0
+#, python-format
+msgid ""
+"Consumption Lines can't be deleted directly. To do that, delete related sale"
+" order line."
+msgstr "Les lignes de consommation ne peuvent pas être modifiées via cette interface. Pour cela, il faut modifier les lignes correspondantes sur les bons de commande."
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__coupon_id
+msgid "Coupon"
+msgstr "Bon de réduction"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__create_uid
+msgid "Created by"
+msgstr "Créé par"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__create_date
+msgid "Created on"
+msgstr "Créé le"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__display_name
+msgid "Display Name"
+msgstr "Nom"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon__discount_fixed_amount_delta
+msgid "Fixed Amount Delta"
+msgstr "Montant restant"
+
+#. module: sale_coupon_multi_use
+#: code:addons/sale_coupon_multi_use/models/sale_coupon_program.py:0
+#, python-format
+msgid ""
+"Fixed Amount can't be changed when there are Multi Use coupons already."
+msgstr "Le montant restant ne peut pas être modifié quand un bon de réduction a déjà été utilisé plusieurs fois."
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__id
+msgid "ID"
+msgstr "ID"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line____last_update
+msgid "Last Modified on"
+msgstr "Dernière modification le"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__write_uid
+msgid "Last Updated by"
+msgstr "Dernière mise à jour par"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__write_date
+msgid "Last Updated on"
+msgstr "Dernière mise à jour le"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon__multi_use
+msgid "Multi Use"
+msgstr "Sécable"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_program__coupon_multi_use
+msgid "Multi Use Coupons"
+msgstr "Bon de réduction sécable"
+
+#. module: sale_coupon_multi_use
+#: code:addons/sale_coupon_multi_use/models/sale_coupon_program.py:0
+#, python-format
+msgid ""
+"Multi Use Coupons program must have Coupon Program Type, Discount Reward and"
+" Fixed Amount as Apply Discount"
+msgstr "Les bons de réductions sécables doivent être de type Remise et Montant fixe"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model,name:sale_coupon_multi_use.model_sale_coupon_consumption_line
+msgid "Sale Coupon Consumption Line"
+msgstr "Lignes de consommation du bon de réduction"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_coupon_consumption_line__sale_order_line_id
+msgid "Sale Order Line"
+msgstr "Ligne de bon de commande"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model,name:sale_coupon_multi_use.model_sale_coupon
+msgid "Sales Coupon"
+msgstr "Bons de réduction"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model,name:sale_coupon_multi_use.model_sale_coupon_apply_code
+msgid "Sales Coupon Apply Code"
+msgstr "Code du bon de réduction"
+
+#. module: sale_coupon_multi_use
+#: model:ir.model,name:sale_coupon_multi_use.model_sale_coupon_program
+msgid "Sales Coupon Program"
+msgstr "Campagne de bons de réduction"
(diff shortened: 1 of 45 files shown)


sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class SaleCoupon(models.Model):
+    """Extend to implement multi use coupon rule."""
+
+    _inherit = "sale.coupon"
+
+    # Takes value from related program (coupon_multi_use field), when
+    # it is generated.
+    multi_use = fields.Boolean("Multi Use", readonly=True)
```
rousseldenis:
```suggestion
    multi_use = fields.Boolean(readonly=True)
```
fields string would be composed automatically like that

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+            coupon_order_data = self._context.get("coupon_order_data")
+            if coupon_order_data:
+                coupon_consumed = self._handle_multi_use(coupon_order_data)
+                if not coupon_consumed:
+                    # Makes it so coupon that coupon is not used up yet.
+                    del vals["state"]
+        return super().write(vals)
+
+
+class SaleCouponConsumptionLine(models.Model):
```
rousseldenis: Please put one model per file.

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+    _name = "sale.coupon.consumption_line"
+    _description = "Sale Coupon Consumption Line"
+
+    coupon_id = fields.Many2one("sale.coupon", "Coupon", required=True, index=True)
+    # ondelete takes care of automatically removing consumption line,
+    # when discount line is removed on related sale order.
+    sale_order_line_id = fields.Many2one(
+        "sale.order.line", "Sale Order Line", required=True, ondelete="cascade"
+    )
+    amount = fields.Float()
```
rousseldenis: Please use digits precision

sale_coupon_multi_use/models/sale_coupon_program.py
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+    @api.onchange("coupon_multi_use")
+    def _onchange_coupon_multi_use(self):
+        if self.coupon_multi_use:
+            self.update(self._prepare_multi_use_vals())
+
+    def _get_multi_use_coupons(self):
+        self.ensure_one()
+        return self.coupon_ids.filtered(lambda r: r.multi_use)
```
rousseldenis:
```suggestion
        return self.coupon_ids.filtered("multi_use")
```

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+    def _normalize_discount(self):
+        """Adjust SOL discount to match consumed discount."""
+        self.ensure_one()
+        # Discount initially won't match, when standard functionality
+        # applies full discount from coupon. But because we split
+        # coupon amount, we want to apply maximum possible discount.
+        sol = self.sale_order_line_id
+        sol_discount = abs(sol.price_unit)
+        amount_to_adjust = sol_discount - self.amount
+        if amount_to_adjust > 0:
```
rousseldenis: Please use float_compare()

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+
+    _inherit = "sale.coupon"
+
+    # Takes value from related program (coupon_multi_use field), when
+    # it is generated.
+    multi_use = fields.Boolean("Multi Use", readonly=True)
+    consumption_line_ids = fields.One2many(
+        "sale.coupon.consumption_line", "coupon_id", "Consumption Lines", readonly=True,
+    )
+    discount_fixed_amount_delta = fields.Float(
```
rousseldenis, line 19: Please use digits precision.

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+        for rec in self:
+            rec.discount_fixed_amount_delta = rec._get_discount_fixed_amount_delta()
+
+    def _is_multi_use_triggered(self, vals):
+        # We expect multi_use coupon will be updated one by one only.
+        return (
+            len(self) == 1
+            and self.multi_use
+            # Must have amount to split.
+            and self.discount_fixed_amount_delta > 0
```
rousseldenis: Please use float_compare()

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+        consumption_line = self._create_consumption_line(amount_consumed, sale_order)
+        consumption_line._normalize_discount()
+        return consumption_line
+
+    def _handle_multi_use(self, coupon_order_data):
+        self.ensure_one()
+        new_delta, discount = self._get_compared_discount_with_delta(coupon_order_data)
+        self._consume_line(discount, coupon_order_data["order"])
+        # Indicates whether coupon was fully consumed.
+        return new_delta <= 0
```
rousseldenis: Please use float_compare()

sale_coupon_multi_use/wizards/sale_coupon_apply_code.py
```diff
@@ -0,0 +1,19 @@
...
+
+class SaleCouponApplyCode(models.TransientModel):
+    """Extend to modify apply_coupon method for multi coupon usage."""
+
+    _inherit = "sale.coupon.apply.code"
+
+    def apply_coupon(self, order, coupon_code):
+        """Extend to pass order coupon ctx for multi coupon usage."""
+        self = self.with_context(
+            coupon_order_data={
```
rousseldenis: I don't really like using context to pass arguments. It was not done for that purpose and should be avoided as much as possible (I know the original apply_coupon is not easily overridable)

rousseldenis (changes requested):
@oerp-odoo Thanks for your contribution.

Code review

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: I know, just like to be explicit here:)

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: Hm, I was using this guidelines: https://www.odoo.com/documentation/13.0/reference/guidelines.html#file-naming
It allows to group models, following main one, (like putting sale.order.line on sale_order.py, if there is also sale.order used - which is main model).

You explicitly split every model?

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: Thing is, I did this because float fields on sale_coupon, also do not use precision. So used same pattern here. Now then if I would use precision, I should probably use configurable decimal accuracy. But then its not clear which one would fir here, Price?

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: And precision to use? The one from related product UOM?

sale_coupon_multi_use/wizards/sale_coupon_apply_code.py
oerp-odoo: So any suggestion, to extend it better? 
Because in my opinion otherwise you would need to overwrite copy/paste blocks of code, which is ugly and not great.
Also, for what else, context is meant, if not passing stuff around? It does just pass stuff:) Though I agree its not as obvious as if you would pass proper argument from method.

I use this rule of thumb (priority):
- Extend method normally and pass/modify it without context (like for example, you can update result dictionary or something).
- If can't reach needed insides of a method by extending normally, pass context to intercept wanted parts.
- if context can't be used (not reachable, poor original implementation etc), then use last resort override whole method.

rousseldenis: No, I've not digged into code too much. If you estimate this is the only way to do it, let's go. Just change the other things.

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+    consumption_line_ids = fields.One2many(
+        "sale.coupon.consumption_line", "coupon_id", "Consumption Lines", readonly=True,
+    )
+    discount_fixed_amount_delta = fields.Float(
+        "Fixed Amount Delta", compute="_compute_discount_fixed_amount_delta"
+    )
+
+    def _get_discount_fixed_amount_delta(self):
+        self.ensure_one()
+        amount_total = self.program_id.discount_fixed_amount
```
i-vyshnevska: for additional saftines I would make "discount_fixed_amount" readonly if program has one used coupon

sale_coupon_multi_use/models/sale_coupon.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def write(self, vals):
+        """Extend to manage multi_use coupons."""
+        if self._is_multi_use_triggered(vals):
+            coupon_order_data = self._context.get("coupon_order_data")
+            if coupon_order_data:
+                coupon_consumed = self._handle_multi_use(coupon_order_data)
+                if not coupon_consumed:
+                    # Makes it so coupon that coupon is not used up yet.
+                    del vals["state"]
```
i-vyshnevska: do we need to add a new state option to state field? smth kind of "started"

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: There is constraint to not allow changing it, if there is at least one. 

Though the way Odoo implemented program and coupons relations is quite clunky. I think better way would be, if program would act just as a template and then coupons would copy values. So it would be independent from later program changes. But that would require some more improving.

For example in standard sale_coupon, when you apply some coupon discount on SO, it creates SOL with product on program. If later you change discount amount on program, it updates related product name, mentioning different amount. And then when you open that old SO again, it will show that new amount on product name as if that discount was applied, when actually it was not. So its kind of confusing to user I think.

sale_coupon_multi_use/models/sale_coupon.py
oerp-odoo: That could work, but then we would introduce additional complexity, because sale_coupon does not expect new state. That would mean, we would need to make sure it would understand what that state means (might be multiple places where we would need to intercept).

i-vyshnevska (approved):

jcoux:
Changed since last reviews:
* Totally rewrite the module `sale_coupon_multi_use` to make it more easy to read and maintainable.
* Fix bugs on module `sale_coupon_multi_use`.
* Add module `sale_coupon_multi_currency`.
* Add module `sale_coupon_multi_use_currency` to have a glue between 2 other modules.

sale_coupon_multi_use/i18n/fr.po
```diff
@@ -0,0 +1,164 @@
...
+#. module: sale_coupon_multi_use
+#: model:ir.model.fields,field_description:sale_coupon_multi_use.field_sale_order_line__coupon_consumption_line_ids
+msgid "Coupon Consumption Line"
+msgstr "Lignes de consommation des bons de réduction"
+
+#. module: sale_coupon_multi_use
+#: code:addons/sale_coupon_multi_use/models/sale_coupon_program.py:0
+#, python-format
+msgid "Coupon multi use can't be changed with existing coupons."
+"Le bon de réduction sécable ne peut pas être modifié avec des bons de réduction existants."
```
damdam-s:
```suggestion
msgstr "Le bon de réduction sécable ne peut pas être modifié avec des bons de réduction existants."
```

damdam-s (approved):
LG

i-vyshnevska (approved):

epanisset (commented):
Functional test done: Ok approved.

jcoux:
> Changed since last reviews:
> 
> * Totally rewrite the module `sale_coupon_multi_use` to make it more easy to read and maintainable.
> * Fix bugs on module `sale_coupon_multi_use`.
> * Add module `sale_coupon_multi_currency`.
> * Add module `sale_coupon_multi_use_currency` to have a glue between 2 other modules.

@rousseldenis As said in my previous comment: a lot of changes on this PR.
And I now have 2 technical approvals + 1 functional approval.
Can you update (or maybe view the changes redo) a review to allow this PR can be merged?

Thanks in advance! :)

jcoux:
Hello @rousseldenis
If you could update your review, it would can very cool: I would like to merge this PR!
Thanks in advance.

sale_coupon_multi_use/models/sale_coupon.py
rousseldenis: @oerp-odoo Please read OCA conventions https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#15file-naming

rousseldenis (approved):
Code review

rousseldenis:
@jcoux Are these one ready ?

jcoux:
@rousseldenis 
Thanks a lot for reviews! :)
For me yes, it's ready.
Just asked me if it's better to squash all commits or not.
What is your opinion about this?

rousseldenis:
> @rousseldenis
> Thanks a lot for reviews! :)
> For me yes, it's ready.
> Just asked me if it's better to squash all commits or not.
> What is your opinion about this?

My opinion is that commits should reflect development flow. So, if they make sense, keep them, if some can be squashed together, do it. Some people don't like too many commits, I'm not if they means something.

jcoux:
As this module is relatively complex and I rewritten a big part of it, I think it's better to keep commit history like we have now: for me we can merge ;)

PR #1327, opened 2020-11-11
[FIX] sale_commercial_partner: compute_sudo

chienandalu:
In a multi-company scenario, a partner could have sale orders belonging to
different companies, wich will raise an security exception due to record
rules

cc @Tecnativa TT26729

### DIFF
--- a/sale_commercial_partner/models/sale.py
+++ b/sale_commercial_partner/models/sale.py
@@ -12,5 +12,6 @@ class SaleOrder(models.Model):
         related='partner_id.commercial_partner_id',
         store=True,
         string='Commercial Entity',
-        index=True
+        compute_sudo=True,
+        index=True,
     )


pedrobaeza:
s/user/partner

pedrobaeza (approved):

sergio-teruel (approved):

sale_commercial_partner/models/sale.py
```diff
@@ -12,5 +12,6 @@ class SaleOrder(models.Model):
         related='partner_id.commercial_partner_id',
         store=True,
         string='Commercial Entity',
-        index=True
+        compute_sudo=True,
```
carlosdauden, line 15: related_sudo can be used instead, If you want:
https://github.com/odoo/odoo/blob/7b986587956c23b77b0da5043410eab11e0d2f28/odoo/fields.py#L368

carlosdauden (approved):

sale_commercial_partner/models/sale.py
chienandalu: The linked code is for v13. It looks to be different in v12, where that evaluation isn't made

PR #1328, opened 2020-11-11
[13.0] [FIX] sale_order_type: Not allowed to delete a sale order type if them is used in an sale order.

nicomacr:
So keep in mind that if you accidentally delete a sales type, it is protected until you remove it from the sales order.
In case you need to perform an operation such as printing the report and the sales order is already confirmed. Because the field is required only per view.


pedrobaeza (approved):

HaraldPanten (approved):

jjscarafia (approved):
LGTM

PR #1330, opened 2020-11-12
[12.0][IMP] sale_order_product_recommendation: Allow to display price unit from pricelist or from last sale order

sergio-teruel:
cc @Tecnativa TT26727
ping @carlosdauden @chienandalu

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -37,6 +37,11 @@ class SaleOrderRecommendation(models.TransientModel):
         help="The less, the faster they will be found.",
     )
     last_compute = fields.Char()
+    # Get default value from config settings
+    sale_recommendation_price_origin = fields.Selection([
+        ('pricelist', 'Pricelist'),
+        ('last_sale_price', 'Last sale price')
+    ], string="Product price origin", default="pricelist")
 
     @api.model
     def _default_order_id(self):
@@ -197,14 +202,28 @@ class SaleOrderRecommendationLine(models.TransientModel):
     )
     sale_uom_id = fields.Many2one(related="sale_line_id.product_uom")
 
-    @api.depends("partner_id", "product_id", "pricelist_id", "units_included")
+    @api.depends("partner_id", "product_id", "pricelist_id", "units_included",
+                 "wizard_id.sale_recommendation_price_origin")
     def _compute_price_unit(self):
+        """
+        Get product price unit from product list price or from last sale price
+        """
+        price_origin = (
+            fields.first(self).wizard_id.sale_recommendation_price_origin or
+            "pricelist"
+        )
+        if price_origin == "last_sale_price":
+            last_sale_price_dic = self._get_last_sale_price_product()
         for one in self:
-            one.price_unit = one.product_id.with_context(
-                partner=one.partner_id.id,
-                pricelist=one.pricelist_id.id,
-                quantity=one.units_included,
-            ).price
+            if price_origin == "pricelist":
+                one.price_unit = one.product_id.with_context(
+                    partner=one.partner_id.id,
+                    pricelist=one.pricelist_id.id,
+                    quantity=one.units_included,
+                ).price
+            else:
+                one.price_unit = last_sale_price_dic.get(
+                    one.product_id.id, 0.0)
 
     def _prepare_update_so_line(self, line_form):
         """So we can extend SO update"""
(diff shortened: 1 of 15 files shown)


rafaelbn (approved):
:+1:  functionally

Travis is :red_circle:

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -215,3 +236,27 @@ def _prepare_new_so_line(self, line_form, sequence):
...
+                LIMIT 1
+            ) AS price_unit
+            FROM product_product pp
+            WHERE pp.id in %s
+        """
+        self.env.cr.execute(sql, (
+            fields.first(self).partner_id.id,
+            self.env.user.company_id.id, tuple(self.mapped("product_id").ids))
+        )
+        return self.env.cr.dictfetchall()
```
carlosdauden: You can return this to be assigned to last_sale_price_dic directly 
```suggestion
        return dict(self.env.cr.fetchall())
```

sergio-teruel:
@carlosdauden Changes done!!! and fix tests

carlosdauden (approved):

pedrobaeza (changes requested):
The SQL query is bypassing record rules, so won't return proper last price in case of using "User (own documents only)" permission or custom record rules.

Please do it with ORM, as we drafted in the chat, using `limit=1` and `order` clauses of `search` method, and also you should consider `price_reduce` instead of `price_unit` for including the discount:

```python
so = self.env["sale.order"].search([("order_line.product_id", "=", x)], limit=1, order="confirmation_date desc")
so_line = self.env["sale.order.line"].search([("order_id", "=", so.id), ("product_id", "=", x)])
return so_line.price_reduce
```

sergio-teruel:
> The SQL query is bypassing record rules, so won't return proper last price in case of using "User (own documents only)" permission or custom record rules.

I want the last price regardless of the commercial that sold it... 

> Please do it with ORM, as we drafted in the chat, using limit=1 and order clauses of search method, a

Ok. I will do it although it is worse for performance and the customer has already complained about the performance of this module.

> consider price_reduce instead of price_unit for including the discount:

I am not sure ... I will chek it with customer.

pedrobaeza:
Yeah, I know this is not the best for performance, but not sure if it will affect a lot. We can do some performance test. But it's more important to avoid data leak.

sergio-teruel:
@pedrobaeza I have done some performance tests and and there is no big difference in performance so I do with ORM search.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -215,3 +231,21 @@ def _prepare_new_so_line(self, line_form, sequence):
...
         line_form.product_uom_qty = self.units_included
+
+    def _get_last_sale_price_product(self):
+        """
+        Get last price from last order.
+        Use sudo to read sale order from other users like as other commercials.
+        """
+        self.ensure_one()
+        so = self.env["sale.order"].sudo().search([
+            ("partner_id", "=", fields.first(self).partner_id.id),
```
carlosdauden: first() method is unneeded If you use ensure_one
```suggestion
            ("partner_id", "=", self.partner_id.id),
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -215,3 +231,21 @@ def _prepare_new_so_line(self, line_form, sequence):
...
+            ("confirmation_date", "!=", False),
+            ("state", "not in", ('draft', 'sent', 'cancel')),
+            ("order_line.product_id", "=", self.product_id.id)
+        ], limit=1, order="confirmation_date DESC, id DESC")
+        so_line = self.env["sale.order.line"].sudo().search([
+            ("order_id", "=", so.id),
+            ("product_id", "=", self.product_id.id)
+        ], limit=1)
```
carlosdauden: You can modifiy this to get last line created if more than one and only fetch price_unit field:
```suggestion
        ], limit=1, order="id DESC").with_context(prefetch_fields=False)
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -215,3 +231,21 @@ def _prepare_new_so_line(self, line_form, sequence):
...
         line_form.sequence = sequence
         line_form.product_uom_qty = self.units_included
+
+    def _get_last_sale_price_product(self):
+        """
+        Get last price from last order.
+        Use sudo to read sale order from other users like as other commercials.
+        """
+        self.ensure_one()
+        so = self.env["sale.order"].sudo().search([
```
carlosdauden, line 244: If you use sudo() you must check company
```suggestion
        so = self.env["sale.order"].sudo().search([
            ("company_id", "=", self.env.user.company_id.id),        
```

carlosdauden (changes requested):
Minor suggestions

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -215,3 +231,21 @@ def _prepare_new_so_line(self, line_form, sequence):
...
+            ("partner_id", "=", fields.first(self).partner_id.id),
+            ("confirmation_date", "!=", False),
+            ("state", "not in", ('draft', 'sent', 'cancel')),
+            ("order_line.product_id", "=", self.product_id.id)
```
carlosdauden, line 249: Optimal performance because order_line field uses auto_join:
https://github.com/odoo/odoo/blob/8c54ed42eef16817f198f10697820eddcc7d0c74/addons/sale/models/sale.py#L174

sergio-teruel:
@carlosdauden changes done!!

sergio-teruel:
@carlosdauden Changes done!!

carlosdauden (approved):

carlosdauden:
This is ready yo merge?

pedrobaeza:
Please forward-port it to v13

PR #1337, opened 2020-11-17
[12.0][FIX] remove company contextmanager

i-vyshnevska:
related to https://github.com/OCA/sale-workflow/pull/493

update on user causes locking for jobs, so only 1 tread is allowed
for the channel, but core implementation was updated and now it relying on
first on given context instead of the user

### DIFF
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -24,16 +24,6 @@ def savepoint(cr):
         _logger.exception('Error during an automatic workflow action.')
 
 
-@contextmanager
-def force_company(env, company_id):
-    user_company = env.user.company_id
-    env.user.update({'company_id': company_id})
-    try:
-        yield
-    finally:
-        env.user.update({'company_id': user_company})
-
-
 class AutomaticWorkflowJob(models.Model):
     """ Scheduler that will play automatically the validation of
     invoices, pickings...  """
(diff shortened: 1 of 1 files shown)


guewen:
Why did you not use `force_company` too in every place the context manager was used? I guess you audited the methods and nothing used the context key, but third-party modules could still rely on it, no?

i-vyshnevska:
@guewen as multi-company tests was running fine I thought we can relly only on company_id of the object, suppose force_company is not malicious so we can have it in context
updated

guewen (approved):
Thanks!

sebalix (approved):

marcelsavegnago (approved):

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -74,12 +69,14 @@ def _create_invoices(self, create_filter):
...
-            invoice.with_context(
-                force_company=invoice.company_id.id).action_invoice_open()
+        with savepoint(self.env.cr):
+            try:
+                invoice.with_context(
+                    force_company=invoice.company_id.id
+                ).action_invoice_open()
+            except Exception:
+                _logger.exception('Error on validating invoice {}'.format(invoice.id))
+                raise Exception
```
guewen: @i-vyshnevska to reraise something without shadowing the exception, you should just use `raise`.
```suggestion
                raise
```

atchuthan:
Are we planning to remove the context_manger to update the user's default company for v13 in the future?

PR #1338, opened 2020-11-17
[ADD] sale_coupon_product_management: init

oerp-odoo:
Implemented related fields update from program to product. Changed the
product name value passed from program to product (no longer uses
reward record display_name) - now it only uses name field from program
directly.

Implemented constraints and onchanges to fill up data from related
program options for program and its product.

Implemented product option values passing when program is created,
added boolean to specify default category for next order promotion program.

Added access rights for sale.coupon.program.option, views for module.

### DIFF
--- a/sale_coupon_product_management/models/product_category.py
+++ b/sale_coupon_product_management/models/product_category.py
@@ -1,0 +1,21 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductCategory(models.Model):
+
+    _inherit = "product.category"
+
+    is_program_category = fields.Boolean()
+    program_product_sale_ok = fields.Boolean()
+    program_product_discount_fixed_amount = fields.Boolean()
+    default_promotion_next_order_category = fields.Boolean(copy=False)
+
+    @api.onchange("is_program_category")
+    def _onchange_is_program_category(self):
+        if not self.is_program_category:
+            self.program_product_sale_ok = False
+            self.program_product_discount_fixed_amount = False
+            self.default_promotion_next_order_category = False
--- a/sale_coupon_product_management/models/product_product.py
+++ b/sale_coupon_product_management/models/product_product.py
@@ -1,0 +1,43 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class ProductProduct(models.Model):
+
+    _inherit = "product.product"
+
+    program_product_sale_ok = fields.Boolean(related="categ_id.program_product_sale_ok")
+    program_product_discount_fixed_amount = fields.Boolean(
+        related="categ_id.program_product_discount_fixed_amount"
+    )
+
+    @api.onchange("categ_id")
+    def _onchange_program_categ_id(self):
+        category = self.categ_id
+        if category.is_program_category:
+            self.sale_ok = self.program_product_sale_ok
+            # In case of existing product, we don't reset the price.
+            if (
+                self.program_product_discount_fixed_amount
+                and isinstance(self.id, models.NewId)
+                and not self.id.origin
+            ):
+                self.lst_price = False
+
+    @api.constrains("categ_id", "sale_ok")
+    def _check_program_product_sale_ok(self):
+        for rec in self:
+            category = rec.categ_id
+            if (
+                category.is_program_category
+                and rec.sale_ok != category.program_product_sale_ok
+            ):
+                raise UserError(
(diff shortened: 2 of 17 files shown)


sale_coupon_product_management/models/product_category.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _predicate_product_categ_with_opts(self):
+        # Expects empty or singleton record.
+        return (
+            self.is_promotion_category or self.is_coupon_category
+        ) and self.program_option_ids
+
+    @api.constrains("default_promotion_next_order_category")
+    def _check_default_promotion_next_order_category(self):
+        if self.default_promotion_next_order_category:
```
i-vyshnevska: can self be a recordset?

sale_coupon_product_management/models/product_product.py
```diff
@@ -0,0 +1,47 @@
...
+        if self._predicate_product_categ_with_opts():
+            # Updating only product.product values here.
+            self.update(
+                self.program_option_ids.get_program_values()[DISCOUNT_PRODUCT_FNAME]
+            )
+
+    @api.constrains("categ_id", "sale_ok")
+    def _check_product_options(self):
+        for rec in self:
+            categ = self.categ_id
```
i-vyshnevska:
```suggestion
            categ = rec.categ_id
```

sale_coupon_product_management/models/sale_coupon_program.py
```diff
@@ -0,0 +1,122 @@
...
+            return forced_product_vals
+
+        self = self.with_context(forced_product_vals=prepare_forced_product_vals())
+        program = super(SaleCouponProgram, self).create(vals)
+        product_extra_vals = program._get_reward_line_product_extra_create_vals(vals)
+        if product_extra_vals:
+            program.discount_line_product_id.write(product_extra_vals)
+        return program
+
+    def write(self, vals):
```
i-vyshnevska, line 139: I am a little bit confused why we need such complexity to just update the product? 
we can make discount_line_product_id visible on sale.coupon.program and sale.coupon so customer can open a view and edit fields properly
cc @epanisset

oerp-odoo: That field is already visible. And user can update it normally using standard functionality.

But requirement was to make it easier to update/manage program and its related product. Also program options allow to automate some things, so you only need to set category and then don't need to check if fields in question are set correctly.

sale_coupon_product_management/models/product_category.py
oerp-odoo: Yes, my mistake.

sale_coupon_product_management/models/sale_coupon_program.py
epanisset: Yes, to avoid any mistake or missing information on their side + Because of the volume that would need to be handle, better to make it automatic :)

i-vyshnevska (approved):

sale_coupon_product_management/models/product_category.py
```diff
@@ -0,0 +1,42 @@
...
+            rec.program_option_ids.validate_compatibility()
+
+    def _predicate_product_categ_with_opts(self):
+        # Expects empty or singleton record.
+        return (
+            self.is_promotion_category or self.is_coupon_category
+        ) and self.program_option_ids
+
+    @api.constrains("default_promotion_next_order_category")
+    def _check_default_promotion_next_order_category(self):
```
damdam-s: every other `constrains` does `for rec in self` (https://github.com/OCA/sale-workflow/pull/1338/files#diff-8f89f1a2aaae6121bdec8d0b8ee8ac3b3f22daeddfdb20344a977b823f5a08b0R24 and https://github.com/OCA/sale-workflow/pull/1338/files#diff-b2ed4ee78eb41076a4a0f93796c236b525bf256972c3ba6d15da12bbd13565f6R26)
Why not there?

damdam-s (changes requested):

epanisset (changes requested):
Functionnal tests provided: 
1. Issue with related_product_default_code (Mandatory for the product chosen on discount_line_product_id when discount_line_product_chosen = true)
2. Issue with the product name in discount_line_product_id, in case of discount_line_product_chosen = true (the product name is changed for the name of the sale.coupon.program). The program name has to be recorded on the product created only when discount_line_product_chosen = false.   
3. It's currently possible to link a product to multiple sale.coupon.program. (in this case, only the last active promotion (in the sequence) is taken into consideration for the promotion application. It could become a limit if a product is linked to more than one promotion active. What I will do = Limit the possibility to link a product to only one sale.coupon.program. 
4. I would suggest changing the related_product_default_code in not mandatory in every case.

jcoux:
Currently rewrite and simplify this PR (and do few fixes): new commits will arrive.

jcoux:
Replaced by https://github.com/OCA/sale-workflow/pull/1479

PR #1342, opened 2020-11-19
[ADD] sale_coupon_most_expensive: init

oerp-odoo:
Adds reward discount apply option for most expensive product.

Logic is reused same way as it is used for cheapest product, just it
uses most expensive product line.

### DIFF
--- a/sale_coupon_most_expensive/models/sale_order.py
+++ b/sale_coupon_most_expensive/models/sale_order.py
@@ -1,0 +1,45 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+from odoo.tools import float_compare
+
+
+class SaleOrder(models.Model):
+    """Extend to implement most_expensive_product disc option."""
+
+    _inherit = "sale.order"
+
+    def _get_most_expensive_line(self):
+        # Mimic _get_cheapest_line method.
+        price_reduce_digits = self.env["decimal.precision"].precision_get(
+            self.order_line._fields["price_reduce"]._digits
+        )
+        lines = self.order_line.filtered(
+            lambda r: not r.is_reward_line
+            and float_compare(r.price_reduce, 0, precision_digits=price_reduce_digits)
+            > 0
+        )
+        return max(lines, key=lambda r: r["price_reduce"])
+
+    def _get_cheapest_line(self):
+        if self._context.get("use_most_expensive_line"):
+            return self._get_most_expensive_line()
+        return super()._get_cheapest_line()
+
+    def _get_reward_values_discount(self, program):
+        if (
+            # Make sure we don't overwrite checks order.
+            program.discount_type != "fixed_amount"
+            and program.discount_apply_on == "most_expensive_product"
+        ):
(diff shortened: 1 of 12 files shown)


mmequignon (approved):
LGTM

sale_coupon_most_expensive/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+
+class SaleOrder(models.Model):
+    """Extend to implement most_expensive_product disc option."""
+
+    _inherit = "sale.order"
+
+    def _get_most_expensive_line(self):
+        # Mimic _get_cheapest_line method.
+        lines = self.order_line.filtered(
+            lambda r: not r.is_reward_line and r.price_reduce > 0
```
rousseldenis: Please use float_compare on price_reduce

rousseldenis (changes requested):
Code review.

sale_coupon_most_expensive/models/sale_order.py
oerp-odoo: is it really necessary?

This method mimics:

```
    def _get_cheapest_line(self):
        # Unit prices tax included
        return min(self.order_line.filtered(lambda x: not x.is_reward_line and x.price_reduce > 0), key=lambda x: x['price_reduce'])
```

Here `float_compare` is not used either. Also if it would be used, then you need to specify precision. And what precision would be used here then?

damdam-s (approved):
code LGTM. no functionnal tests

sale_coupon_most_expensive/models/sale_order.py
jcoux: @rousseldenis Fixed! Can you update your review please?

epanisset (approved):
Functionnal review provided:
- Automatic application of the promotion : ok 
- Application using a coupon : ok 
- Application if multi-promotion : ok

sale_coupon_most_expensive/models/sale_order.py
```diff
@@ -0,0 +1,41 @@
...
+class SaleOrder(models.Model):
+    """Extend to implement most_expensive_product disc option."""
+
+    _inherit = "sale.order"
+
+    def _get_most_expensive_line(self):
+        # Mimic _get_cheapest_line method.
+        lines = self.order_line.filtered(
+            lambda r: not r.is_reward_line
+            and float_compare(r.price_reduce, 0, precision_digits=2) > 0
```
rousseldenis: You cannot hardcode precision_digits like that. You need to get value from 'Product Price' as defined in field : https://github.com/odoo/odoo/blob/13.0/addons/sale/models/sale.py#L1258

jcoux: @rousseldenis Thanks for your feedback: it's now fixed.

jcoux:
Hello @rousseldenis
If you could update your review, it would can very cool: I would like to merge this PR!
Thanks in advance.

rousseldenis (approved):
Code review

jcoux:
Will squash this PR to have a proper commit history.

jcoux:
PR squashed: for me ready to merged ;)

jcoux:
I also rebased to fix travis errors (from other modules)

PR #1348, opened 2020-11-23
[14.0][MIG] sale_partner_version: Migration to v14.0

kevinkhao:
requires https://github.com/OCA/partner-contact/pull/1021


atchuthan (approved):
Code review :+1:

hparfr:
@kevinkhao can you please rebase ?

atchuthan:
@kevinkhao 
There seem to be conflicts.  Please resolve it

Conflicting files
sale_exception/README.rst
sale_exception/__manifest__.py
sale_exception/data/sale_exception_data.xml
sale_exception/i18n/es.po
sale_exception/i18n/sale_exception.pot
sale_exception/models/sale.py
sale_exception/static/description/index.html

kevinkhao:
@hparfr @atchuthan done

kevinkhao:
@hparfr should be good
could you pls help check if this follows convention https://github.com/OCA/sale-workflow/pull/1348/commits/7b3ea80c2a990ebbda494a7262e3c903759dd2cd

hparfr:
You can may be populate the `excludes` key in the manifest like in account_e-invoice_generate. About the .travis file I do not know. 
https://github.com/OCA/edi/blob/12.0/account_e-invoice_generate/__manifest__.py#L15

kittiu:
@kevinkhao I functional test this, and it works fine. I got some trivial comments,

* Should we remove the word (copy) from the archived address. Because sometime, it will also get printed in the form.
* The readme is rather minimum, it took me a while to make sense of it. Can we add more explanation, i.e.,
```
This module uses the partner_address_version module to create a partner version when confirming a sale order.

By default, this module ensure that the address fields (Invoice Address, Shipping address) are immutable.
This works by ensure that, whenever a sale order is confirmed, the address fields of the sale order will be copied (and immediately archived), so it will used by this sale order only.
```

kittiu:
By the way, is this technique being implemented in other modules yet? I.e., purchase order. I am interested to make one.

kevinkhao:
> * Should we remove the word (copy) from the archived address. Because sometime, it will also get printed in the form.


I think if we remove (copy) it is less easy to differentiate between archived versions, both for a user and for other module that maybe uses the name of a partner as a key or something like that. Then we could add (archived) or something like that ?


> * The readme is rather minimum, it took me a while to make sense of it. Can we add more explanation, i.e.,

Yes, thanks for the suggestion

> This module uses the partner_address_version module to create a partner version when confirming a sale order.
> 
> By default, this module ensure that the address fields (Invoice Address, Shipping address) are immutable.
> This works by ensure that, whenever a sale order is confirmed, the address fields of the sale order will be copied (and immediately archived), so it will used by this sale order only.

kittiu (approved):

kittiu:
> I think if we remove (copy) it is less easy to differentiate between archived versions, both for a user and for other module that maybe uses the name of a partner as a key or something like that. Then we could add (archived) or something like that ?

Thanks, I see it is very subjective. Let's leave it as is.

kevinkhao:
@kittiu also I forgot to answer the part about using it elsewhere: as far as I know partner_address_version is not used elsewhere in OCA, we have some custom modules that use it though
https://github.com/akretion/sale-import
for creating sale orders through a rest API

atchuthan (approved):

newtratip (approved):

PR #1354, opened 2020-11-26
[13.0][MIG] sale order revision

tbaden:
superseed of [1021](https://github.com/OCA/sale-workflow/pull/1021)


tbaden:
how to handle travis fail on prettier xml plugin?

tbaden:
Test failing sale_automatic_workflow:
FAIL: TestAutomaticWorkflow.test_date_invoice_from_sale_order
  File "/home/travis/build/OCA/sale-workflow/sale_automatic_workflow/tests/test_automatic_workflow.py", line 54, in test_date_invoice_from_sale_order
    self.assertEqual(invoice.invoice_date, last_week_time.date())
AssertionError: datetime.date(2020, 11, 20) != datetime.date(2020, 11, 19)
 
Module sale_automatic_workflow: 1 failures, 0 errors

@rousseldenis 
How to handle as it's not part of the migration?

tbaden:
> Test failing sale_automatic_workflow:
> FAIL: TestAutomaticWorkflow.test_date_invoice_from_sale_order
> Traceback (most recent call last):
> File "/home/travis/build/OCA/sale-workflow/sale_automatic_workflow/tests/test_automatic_workflow.py", line 54, in test_date_invoice_from_sale_order
> self.assertEqual(invoice.invoice_date, last_week_time.date())
> AssertionError: datetime.date(2020, 11, 20) != datetime.date(2020, 11, 19)
> 
> Module sale_automatic_workflow: 1 failures, 0 errors
> 
> @rousseldenis
> How to handle as it's not part of the migration?

seems to be a timezone bug. running between 0:00 and 1:00 CEST(GMT+1) will result in an error, as the invoice date uses GMT+0, which will use the day before instead.
tested this with different timedeltas to confirm this. will rerun travis after 1:00

MiquelRForgeFlow:
@tbaden thanks for doing the superseed and finish the migration, but please could you maintain the commits of @jeroen7s? If not, then it cannot be called a superseed but a substitution instead :S

tbaden:
@MiquelRForgeFlow thanks for the hint, I somehow dropped them in my travis tests/reruns, will add them

tbaden:
@MiquelRForgeFlow his commit history is now back where it belongs and Mr. Travis is happy as well.

marylla (approved):
I made a functional test, it works and looks good to me. :)

What I did:
* choose a quotation
* cancel the quotation
* create a new revision from the cancelled quotation
-> a new quotation is created and named as the origin order with the suffix -01
* cancel this revision quotation
* create a new revision
-> a new quotation is created and named as the origin order with the suffix -02

I still think it's bad that I cannot create a new second revision from the first cancelled order because it tries to create a new one with -01 instead of choosing the next free number. 
But ... this is also the case in previous versions and therefore, does not interfere with my review here. ;)

MiquelRForgeFlow (approved):

HaraldPanten (approved):
Functional review

sale_order_revision/tests/test_sale_order_revision.py
```diff
@@ -7,23 +7,23 @@
 
 
 class TestSaleOrderRevision(common.SavepointCase):
-
     @classmethod
     def setUpClass(cls):
         super(TestSaleOrderRevision, cls).setUpClass()
-        cls.sale_order_model = cls.env['sale.order']
```
simahawk: Could you add this? :pray: 

cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)

simahawk: it can be done in another PR if you want... we can merge this.

tbaden: how to reference this for a new pr? creating an issue related to this comment?

PR #1357, opened 2020-12-02
[13.0][FIX] sale_order_type, type not transferred from SO to invoice

CasVissers-360ERP:
When creating an invoice from a sale order the sale order type was overwritten by the compute method on account.move.

@pedrobaeza @HaraldPanten


pedrobaeza:
This change is not valid, as it's invalidating the change the partner and there's s previous type. You need to locate the invoice generation code, and when this compute is triggered, and force the type afterwards. I remember there was a test for this, isn't it? Maybe it's not reproducing the exact problem?

PR #1362, opened 2020-12-03
[13.0] [FIX] sales_team_security: Show the group in the same droplist of sales.

nicomacr:
In this case the group "Channel Manager" doesn't showed in the list of right of sales.
The category is different than the other gruops, we fix that in this commit.


pedrobaeza (approved):

bodedra (approved):
LG :+1:

PR #1366, opened 2020-12-07
[14][MIG] sale_shipping_info_helper

bguillot:
Migration of sale shipping info helper


sale_shipping_info_helper/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.depends("amount_total", "amount_untaxed")
+    def _compute_shipping(self):
+        for record in self:
+            record.shipping_amount_untaxed = (
+                record.shipping_amount_total
+            ) = record.shipping_amount_tax = 0
+            for line in record.order_line:
+                if not line.is_delivery:
+                    continue
+                record.update(
```
atchuthan: If we have partial DO and shipped later, then we might have multiple shipping line, so we need to append the values of each shipping line right?

atchuthan (changes requested):

dsolanki-initos (approved):
Code review no major change.

sale_shipping_info_helper/models/sale_order.py
bguillot: Indeed you can have several shipping lines.

Thanks

kevinkhao (approved):
Code LGTM

dsolanki-initos:
It would be great if somone can merge this PR.

atchuthan (approved):

PR #1372, opened 2020-12-11
[MIG] 14.0 sale_delivery_state

PierrickBrun:
I removed a test which was declared twice


danimv5 (commented):
The module works fine. But I would put the new field: "Delivery State", within the "Other information" tab, in the "Delivery" block, so it does not affect the sales order's fields.

If you do not like where it is placed, you should put the field under "Payment terms", since right now it is affecting the field's display.

PierrickBrun:
Hi @danimv5 ,

Thank you, I did not see the broken display. This is fixed now.

I think it is important to have it there since this is something you would want to see at the first glance (for instance if someone calls and asks if their package was shipped)

sebastienbeau (approved):
LGTM test and code review

sale_delivery_state/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.tools import float_compare, float_is_zero
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    delivery_state = fields.Selection(
```
simahawk, line 13: this module seems to clash or overlap w/ https://github.com/OCA/sale-workflow/blob/14.0/sale_automatic_workflow/models/sale_order.py#L18

Could you change that module to depends on this and refactor that field to be computed on `delivery_state == done` (and make it not stored)? We could trash the field but would be better for backward compat.

Computing the same info and storing it twice seems useless.

sale_delivery_state/tests/test_delivery_state.py
```diff
@@ -0,0 +1,91 @@
...
+# Copyright 2018 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import SavepointCase
+
+
+class TestDeliveryState(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 14:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (changes requested):

sebastienbeau:
@PierrickBrun  can you fix it ?

simahawk:
@PierrickBrun gentle ping :wink:

PierrickBrun:
@simahawk Done
Travis is failing but I suspect this is related to the rebase and not the last commit

sale_automatic_workflow/models/sale_order.py
```diff
@@ -18,23 +17,13 @@ class SaleOrder(models.Model):
...
         for order in self:
-            order.all_qty_delivered = all(
-                line.product_id.type not in ("product", "consu")
-                or float_compare(
-                    line.qty_delivered, line.product_uom_qty, precision_digits=precision
-                )
-                == 0
-                for line in order.order_line
-            )
+            order.all_qty_delivered = order.delivery_sate == "done"
```
sebastienbeau:
```suggestion
            order.all_qty_delivered = order.delivery_state == "done"
```

sale_automatic_workflow/models/sale_order.py
```diff
@@ -18,23 +17,13 @@ class SaleOrder(models.Model):
     all_qty_delivered = fields.Boolean(
         compute="_compute_all_qty_delivered",
         string="All quantities delivered",
-        store=True,
+        store=False,
```
simahawk: I don't think you should change this now because you could break existing installations. If you do it, at least, add a search method which logs a deprecation warning.

What do you think?

simahawk (commented):
True, the error on travis seems related to `sale_product_seasonality` but the build on 14.0 is green. Would you mind to rebase?

sale_automatic_workflow/models/sale_order.py
simahawk: @PierrickBrun @sebastienbeau ping

sale_automatic_workflow/__manifest__.py
```diff
@@ -5,15 +5,15 @@
 
 {
     "name": "Sale Automatic Workflow",
-    "version": "14.0.1.1.1",
+    "version": "14.0.1.2.1",
```
simahawk:
```suggestion
    "version": "14.0.1.2.0",
```

sale_automatic_workflow/models/sale_order.py
PierrickBrun: Hi @simahawk I think it may be better to just let it stored and add a comment & roadmap message to make it non-stored in a future version

simahawk: Salut! 
As you prefer. So, please:

1. restore this flag
2. add a comment here and add a note to roadmap
3. squash the last commits of fixes done via GH on top of the migration commit
4. reword the commit for sale_automatic_workflow as this: `sale_automatic_workflow: depend on sale_delivery_state`

Merci! :)

sale_delivery_state/tests/test_delivery_state.py
```diff
@@ -0,0 +1,92 @@
...
+        self.order.order_line[0].qty_delivered = 2
+        self.assertEqual(self.order.delivery_state, "partially")
+
+    def test_delivery_done(self):
+        self.order.action_confirm()
+        for line in self.order.order_line:
+            line.qty_delivered = line.product_uom_qty
+        self.assertEqual(self.order.delivery_state, "done")
+
+    def mock_delivery(self):
```
simahawk: I didn't notice this before...  why don't you use `mock`? 
If you need to patch a field value you can use `PropertyMock`
Check this example https://github.com/OCA/wms/blob/63620cdd6abf14198c11967d4660c0e421f3a877/shopfloor/tests/test_checkout_scan_package_action.py#L392

sale_delivery_state/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        Returns True if all line have qty_delivered >= to ordered quantities
+
+        If `delivery` module is installed, ignores the lines with delivery costs
+
+        :returns: boolean
+        """
+        self.ensure_one()
+        # Skip delivery costs lines
+        sale_lines = self.order_line.filtered(lambda rec: not rec._is_delivery())
```
simahawk, line 46: @PierrickBrun wondering: does it make any sense to rely on `_is_delivery` only?
IMO we should depend on `delivery` and then use `is_delivery` flag instead (which is what is used by `_is_delivery` when `delivery` module is installed. 
I mean, what's the point of having a delivery state that won't be ever a delivery if you don't install `delivery`? :thinking: 
This would also make testing easier (no mocking required).

sale_delivery_state/tests/test_delivery_state.py
simahawk: please cherry-pick my commit here https://github.com/OCA/sale-workflow/pull/1649/commits/1464c53fae5ab43b04136803ae9bf5e86254a40f

sale_delivery_state/models/sale_order.py
PierrickBrun: I think the module was named something else when this decision was made but it may be pointless now.

Since this module is becoming a dependecy of sale_automatic_workflow, maybe it is a good idea to keep it that way to avoid having `sale_automatic_workflow` depend on `delivery` ?

simahawk: good point! :+1: 

Can you address this https://github.com/OCA/sale-workflow/pull/1372#discussion_r665419433 and cherry-pick my tests refactoring? :pray:

PierrickBrun: Done, thank you for the tests refactoring

sale_automatic_workflow/__manifest__.py
```diff
@@ -5,15 +5,15 @@
...
+    "version": "14.0.1.2.0",
     "category": "Sales Management",
     "license": "AGPL-3",
     "author": "Akretion, "
     "Camptocamp, "
     "Sodexis, "
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
-    "depends": ["sale_stock", "sales_team"],
+    "depends": ["sale_stock", "sales_team", "sale_delivery_state"],
```
pedrobaeza: You can't add such new dependency in a stable module.

pedrobaeza (dismissed):

sale_automatic_workflow/__manifest__.py
simahawk: The only way I see around this is that for v13 and v14 we add another module w/ auto_install=True that overrides sale_automatic_workflow behavior.

PierrickBrun: Hi @pedrobaeza I removed the dependency and added a glue module (with auto_install=True)

sale_automatic_workflow_delivery_state/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module make `sale_automatic_workflow`'s `all_qty_delivered` field be computed using
```
simahawk, line 1: Please:
1. add to the roadmap of this module "trash it for v 14"
2. add a comment in sale_automatic_workflow on top of `_compute_all_qty_delivered` that says "make it dependent on sale_delivery_state in v15 and pick the computation from ` sale_automatic_workflow_delivery_state`" or something similar

simahawk: @pedrobaeza good for you?

pedrobaeza: OK

PierrickBrun: Done

sale_delivery_state/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,15 @@
...
+The state of the sale order can be forced to fully delivered in case
+some quantities were cancelled by the customer and you consider you have
+nothing more to deliver.
+
+Sale order lines can have products or services, as long as the field `qty_delivered`
+is set, it will trigger the computation of delivery state.
+
+This module also works with delivery.carrier fees that are added as a
+sale order line. Thoses line are special as they will never be considered delivered.
+Delvery fees lines are ignored in the computation of the delivery state.
```
simahawk:
```suggestion
Delivery fees lines are ignored in the computation of the delivery state.
```

simahawk (approved):
@pedrobaeza fine for you?

pedrobaeza:
One thing about the naming, and sorry to come a bit late about that: on Odoo's technical world, `delivery` is used for all the carrier thing, so using here in the module name can be misleading. What about `sale_picking_state`? Note that there's a twin module https://github.com/OCA/purchase-workflow/tree/12.0/purchase_picking_state for the same on purchases.

simahawk:
but this module depends only on `sale` not on `delivery` so there's no picking here.
The hook is `_is_delivery` which is overridden `delivery`module to use `is_delivery` flag.

pedrobaeza:
But the main purpose of the module is to indicate if an order is "delivered", understood as the delivered quantity part, but not carrier delivered part. Note also that we have  https://github.com/OCA/delivery-carrier/tree/13.0/delivery_state for that second part. Then, although not depending on `sale_stock`, the meaning goes with picking part, not with carrier part.

pedrobaeza:
Anyways, what is the use case for hooking into `sale_automatic_workflow`? Doesn't already existing flag `invoice_service_delivery`  working for you?

simahawk:
> But the main purpose of the module is to indicate if an order is "delivered", understood as the delivered quantity part, but not carrier delivered part. Note also that we have https://github.com/OCA/delivery-carrier/tree/13.0/delivery_state for that second part. Then, although not depending on `sale_stock`, the meaning goes with picking part, not with carrier part.

Indeed, I would argue that https://github.com/OCA/delivery-carrier/tree/13.0/delivery_state name is wrong: it should be `delivery_carrier_state` or `delivery_picking_state` because that one is really tied to those models and behavior :wink:  

> Anyways, what is the use case for hooking into sale_automatic_workflow? Doesn't already existing flag invoice_service_delivery working for you?

Only perf issue: s_auto_workflow computes the same info and stores it. It's pointless to compute the very same info twice.

pedrobaeza:
> Indeed, I would argue that https://github.com/OCA/delivery-carrier/tree/13.0/delivery_state name is wrong: it should be delivery_carrier_state or delivery_picking_state because that one is really tied to those models and behavior

`carrier` word is not used in Odoo naming schema. About the last one, it was debated, but as `delivery` is depending on both `sale` + `stock`, if someone wants to expand the module to bring-down the delivery state information to the sales order, it can be done in the same module.

> Only perf issue: s_auto_workflow computes the same info and stores it. It's pointless to compute the very same info twice.

Then put the same name for both fields, and the computation will be done only once, but don't force to install a module with a debatable name, or let's go for a good module name. Anyways, I think this repository is too crowded with this kind of mini-modules.

sebastienbeau:
Hi regarding the naming 

If you look at the code of Odoo in sale module https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py.
You will see the wording delivery/delivered a lot of time because for a service or a physical product when we deliver it physical or virtually we deliver it.

For sure, I understand the "conflict" regarding the module "delivery" of Odoo SA but sadly we can not ask Odoo SA to rename to delivery_carrier (which will be really better). One thing that we can (but it's required a lot of work) it to prefix all delivery module with delivery_carrier when it's about carrier thing (but this is an other debate)

If nobody have a incredible good idea of module name, I think we should go for the current one (and rename in Vx if one day we have a better idea)

Regarding the glue module

sale workflow and sale_delivery_state do not have the same implementation
- in sale workflow if have a sale order with service even if you didn't deliver it, it will be considered as delivered
- in sale_delivery_state you need to set the delivered qty the service to make it delivered (only shipping cost are ignored)

So we need it if we want a version of sale-workflow that support service correctly. And as we do not want to change the dependency of sale workflow then we can keep a broken sale-workflow and if you need the service support you can install sale_delivery_state with the glue module.


So It's ok for me to merge this PR

pedrobaeza:
OK, I have dismissed my review then for not blocking.

sebastienbeau:
Ok thank you Pedro

PR #1375, opened 2020-12-15
[13.0][IMP] sale_by_packaging can be sold

TDu:
This change allows to be more specific on which product.packaging can be
sold or not.
By default the value from the product.packaging.type is used. But it is
editable afterwards, with this change.

### DIFF
--- a/sale_by_packaging/models/product_packaging.py
+++ b/sale_by_packaging/models/product_packaging.py
@@ -1,14 +1,15 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
 
     can_be_sold = fields.Boolean(
-        related="packaging_type_id.can_be_sold", readonly=True,
+        string="Can be sold", compute="_compute_can_be_sold", readonly=False, store=True
     )
+
     force_sale_qty = fields.Boolean(
         string="Force sale quantity",
         help="Determine if during the creation of a sale order line, the "
@@ -18,3 +19,8 @@ class ProductPackaging(models.Model):
         "When the user will put 3 as quantity, the system can force the "
         "quantity to the superior unit (5 for this example).",
     )
+
+    @api.depends("packaging_type_id")
+    def _compute_can_be_sold(self):
+        for record in self:
+            record.can_be_sold = record.packaging_type_id.can_be_sold
--- a/sale_by_packaging/tests/test_packaging_type_can_be_sold.py
+++ b/sale_by_packaging/tests/test_packaging_type_can_be_sold.py
@@ -38,19 +38,48 @@ class TestPackagingTypeCanBeSold(SavepointCase):
                 "packaging_type_id": cls.packaging_type_cannot_be_sold.id,
             }
         )
-
-    def test_packaging_type_can_be_sold(self):
-        order = self.env["sale.order"].create({"partner_id": self.partner.id})
-        order_line = self.env["sale.order.line"].create(
+        cls.order = cls.env["sale.order"].create({"partner_id": cls.partner.id})
+        cls.order_line = cls.env["sale.order.line"].create(
             {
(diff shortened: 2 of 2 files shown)


sale_by_packaging/models/product_packaging.py
```diff
@@ -18,3 +17,10 @@ class ProductPackaging(models.Model):
         "When the user will put 3 as quantity, the system can force the "
         "quantity to the superior unit (5 for this example).",
     )
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
simahawk: wouldn't be better a computed field w/ readonly=False? :thinking:

simahawk: also, any test?

TDu: > wouldn't be better a computed field w/ readonly=False? :thinking:

It was my first implementation.
But during the discussion it was decided that a change on the product.packaging.type should not cascade to the product.packaging. As the product_packaging_type_id is a required field it is better done like this IMO

simahawk: hmm but then do not depend on it, no?

TDu: I don't get what you mean, do not depend on what ?

TDu:
I have added a unit test.

sale_by_packaging/models/product_packaging.py
jbaudoux: Implement it with an onchange instead of create

TDu: So finally I implemented it with a compute readonly=False and store=True
Seems to work as expected.

sale_by_packaging/tests/test_packaging_type_can_be_sold.py
```diff
@@ -38,19 +38,44 @@ def setUpClass(cls):
...
+            self.order_line.write(
+                {"product_packaging": self.packaging_cannot_be_sold.id}
+            )
+            onchange_res = self.order_line._onchange_product_packaging()
             self.assertIn("warning", onchange_res)
+
+    def test_product_packaging_can_be_sold(self):
+        """Check that a product.packaging can be independently set as can be sold.
+        """
+        with self.assertRaises(ValidationError):
```
simahawk: can you test which exception msg you get here? You can use `with self.assertRaisesRegex(TypeError, msg)`

simahawk (changes requested):
LG but small improvement required on tests.

sale_by_packaging/tests/test_packaging_type_can_be_sold.py
TDu: Done

sale_by_packaging/models/product_packaging.py
```diff
@@ -18,3 +19,8 @@ class ProductPackaging(models.Model):
         "When the user will put 3 as quantity, the system can force the "
         "quantity to the superior unit (5 for this example).",
     )
+
+    @api.depends("packaging_type_id")
```
rousseldenis, line 23:
```suggestion
    @api.depends("packaging_type_id.can_be_sold")
```

rousseldenis (changes requested):
Code review

sale_by_packaging/models/product_packaging.py
simahawk: @rousseldenis I think this is done on purpose:  you don't want this value to change if the type changes its value.

TDu: Yes that is how it was requested.
You get the default "can_be_sold" value when selecting a packaging_type. But afterwards you can change it on the product.packaging.

simahawk (approved):

simahawk:
@TDu pls squash :)

TDu:
@simahawk Squash done

sale_by_packaging/models/product_packaging.py
simahawk: @rousseldenis ping :)

simahawk:
@rousseldenis piiiing :)

rousseldenis (approved):

PR #1378, opened 2020-12-16
[13.0] Add sale_cutoff_time_delivery

simahawk:
Replace #1104

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -2,3 +2,4 @@ product-attribute
 stock-logistics-warehouse
 account-invoicing
 server-tools
+partner-contact
--- a/sale_cutoff_time_delivery/__init__.py
+++ b/sale_cutoff_time_delivery/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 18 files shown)


simahawk:
@jbaudoux as per your comment here https://github.com/OCA/sale-workflow/pull/1104#pullrequestreview-390732343 if it's still valid I won't keep this blocked because it's an alpha module and we are using in it prod since quite some time now.
Let's discuss further improvements for another round.

jgrandguillaume (approved):
LGTM

TDu (approved):

jbaudoux:
> @jbaudoux as per your comment here [#1104 (review)](https://github.com/OCA/sale-workflow/pull/1104#pullrequestreview-390732343) if it's still valid I won't keep this blocked because it's an alpha module and we are using in it prod since quite some time now.
> Let's discuss further improvements for another round.

Can you at least put a FIXME that the comparison should be in warehouse timezone and not UTC.

simahawk:
> > @jbaudoux as per your comment here [#1104 (review)](https://github.com/OCA/sale-workflow/pull/1104#pullrequestreview-390732343) if it's still valid I won't keep this blocked because it's an alpha module and we are using in it prod since quite some time now.
> > Let's discuss further improvements for another round.
> 
> Can you at least put a FIXME that the comparison should be in warehouse timezone and not UTC.

done

jbaudoux (approved):

PR #1380, opened 2020-12-16
[13.0] sale_exception: add details on lines

simahawk:
Take over and complete #1194
 
So we can:

* display the lines with an exception in red
* show the exception messages when we click on a line

It's a huge usability improvement when users have dozens or hundreds of
lines.


jgrandguillaume (approved):
Functional test ok

guewen (approved):

PR #1381, opened 2020-12-16
[13.0][IMP] Add filter on sale_cutoff_time_delivery

TDu:
Add cutoff filter on stock.picking views to allow filtering on picking
before yesterday's cutoff, inside today's cutoff and in future cuttofs.


simahawk (approved):

jgrandguillaume (approved):
Functional test ok

PR #1382, opened 2020-12-16
14.0 MIGR sale isolated quotation

Kev-Roche:


bealdav:
cc @kittiu

bealdav (approved):
Just tested, it works

youring:
Just found a small issue on runbot, maybe also exist in previous versions.
Create a quotation, then convert to order, refresh page(browser refresh F5), the order form status bar is gone.

atchuthan (approved):

bealdav:
@youring you're right it's a limitation in this module but also in odoo in general.
Some informations are lost when refreshing. It's important to make think your customer to use menu after any refresh

bealdav:
Thanks a lot @dreispt

Issue #1371, opened 2020-12-11
[12.0] sale_tier_validation: Sale order cannot be processed when there are followers 


TimLai125:
Steps to reproduce:
1. Create a "Tier Validation" record for the sale order.
![image](https://user-images.githubusercontent.com/27185427/101875435-96d3ef80-3bc5-11eb-894c-544ad25fe3a7.png)
2. Create a SO with Admin user and assign Demo user as the Salesperson.
![image](https://user-images.githubusercontent.com/27185427/101875882-617bd180-3bc6-11eb-9c53-4577f9d5c759.png)
3. Press "Request Validation" 

Current behavior:
Odoo Server Error is raised and the order cannot be processed.
![image](https://user-images.githubusercontent.com/27185427/101875828-47da8a00-3bc6-11eb-854d-b625ed662b58.png)

TimLai125:
To resolve this situation, I figured out that we need to remove the followers from the record (Only the user who presses the "Request Validation" can be the follower, when there are other followers, the order is "stuck").

The cause of this issue seems to be the `portal` module is trying to update the `access_token` of the record when "Request Validation" is being pressed.

@mgosai  Do you have any comment on this issue?

PR #1386, opened 2020-12-19
[12.0][FIX] sale_tier_validation add access_token to exception field

kh-badep:
When confirming a sale order by a user who can do reviews, automatic validation is not done because immediately after that, the system generate an access token for portal access.

Since the access token does not affect the content of the sale order, it is best to add it to exception fields

### DIFF
--- a/sale_tier_validation/models/sale_order.py
+++ b/sale_tier_validation/models/sale_order.py
@@ -1,7 +1,7 @@
 # Copyright 2019 Open Source Integrators
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import models, api
 
 
 class SaleOrder(models.Model):
@@ -9,3 +9,9 @@ class SaleOrder(models.Model):
     _inherit = ['sale.order', 'tier.validation']
     _state_from = ['draft', 'sent', 'to approve']
     _state_to = ['sale', 'approved']
+
+    @api.model
+    def _get_under_validation_exceptions(self):
+        res = super()._get_under_validation_exceptions()
+        res.append('access_token')
+        return res


yostashiro (approved):
LGTM. Works as expected in my local environment. 👍

yostashiro:
This PR fixes https://github.com/OCA/sale-workflow/issues/1371.

HaraldPanten (approved):

TimLai125 (approved):
Functional test is done and I confirm this will fix https://github.com/OCA/sale-workflow/issues/1371

rousseldenis (approved):
LGTM

kittiu:
LGTM too. Don't Not sure why runbot is red.

kittiu (approved):

marcelsavegnago (approved):

jonerikceberio:
LGTM, referenced in https://github.com/OCA/account-invoicing/pull/773 because it is the same problem with account.invoice

kittiu:
Hi, all.

It seem access_token error affect many tier_validation modules, not only this sale_tier_validation.
I think, instead, we should fix this in `base_tier_validation` module.

Note: I see the same problem now in v13 and v14 as well.

WDYT?

Note: I made a quick fix for these 3 versions.
* https://github.com/OCA/server-ux/pull/281
* https://github.com/OCA/server-ux/pull/282
* https://github.com/OCA/server-ux/pull/283

jonerikceberio:
It would be great solving it from base_tier_validation, as any class that inherits portal is going to show the same problem.

LoisRForgeFlow:
The upstream fixes has been merged, including the [one for v12](https://github.com/OCA/server-ux/pull/281). This can be closed IMO.

PR #1388, opened 2020-12-21
[IMP] sale_order_type: Find sale order type based on the products of the sale order

chienandalu:
Get features that weren't forward-ported from v11 (aadd9f72dfbb94d104bb3a558a1cd9d553922b10)

cc @Tecnativa TT27435


pedrobaeza:
CI is red. Why do you need this?

chienandalu:
> Why do you need this?

I detected the missing feature from 11.0 to 12.0 of being able to have a sequence in `sale.order.type` to decide wich one comes first. It was attached to all this other stuff that is missing as well

pedrobaeza:
OK, the rule matching is not very needed by any of us, but if you find it working in a simply way, go ahead. If not, please extract sequence part only.

PR #1394, opened 2020-12-22
[14.0] [MIG] sale_partner_incoterm

ReedHayashikawa:
sale_partner_incoterm migration from 13.0 to 14.0


ReedHayashikawa:
Can you review this pull request?
@dreispt

sale_partner_incoterm/views/res_partner.xml
```diff
@@ -8,7 +8,7 @@
         <field name="inherit_id" ref="stock.view_partner_stock_form" />
         <field name="arch" type="xml">
             <field name="property_payment_term_id" position="after">
-                <field name="customer_rank" invisible="1" />
+                <field name="customer_rank" attrs="{'invisible':1}" />
```
atchuthan: Why do we need to change it from invisible to attrs?

sale_partner_incoterm/models/sale_order.py
```diff
@@ -0,0 +1,17 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
+        res = super().onchange_partner_id()
+        if not self.partner_id:
+            self.incoterm = False
+            return res
+        self.incoterm = self.partner_id.sale_incoterm_id
```
atchuthan, line 13: We don't need this code
```
        if not self.partner_id:
            self.incoterm = False
            return res
```
As when partner_id is not available then self.partner_id.sale_incoterm_id will result as False by default with the New ORM concept of recordset

atchuthan (changes requested):
minor changes

sale_partner_incoterm/views/res_partner.xml
dreispt: This change is not needed, but not a blocker

dreispt (approved):

atchuthan (approved):

HaraldPanten (approved):
LGTM

PR #1395, opened 2020-12-23
[14.0][MIG] sale_start_end_dates

alexis-via:


lfreeke:
Thank you for migrating this module. 

I do have a question, is it correct when setting a default start and end date it does't show on the sales order lines?

alexis-via:
@lfreeke When setting a default start and end date on the order AND your product has "Must have start/end dates", then you will have the start/end date on the lines. I just tested it again on v14 and it works as expected.

lfreeke:
> @lfreeke When setting a default start and end date on the order AND your product has "Must have start/end dates", then you will have the start/end date on the lines. I just tested it again on v14 and it works as expected.

@alexis-via Got it. Thanks. 
LGTM :+1:

lfreeke (approved):
Functional test :+1:

rvalyi (approved):

PR #1400, opened 2020-12-28
[12.0][FIX] sale_order_line_input: Company_id does not get the default value for new records

sergio-teruel:
cc @Tecnativa TT27512

### DIFF
--- a/sale_order_line_input/models/sale_order.py
+++ b/sale_order_line_input/models/sale_order.py
@@ -22,7 +22,7 @@ class SaleOrderLine(models.Model):
         store=True,
     )
 
-    @api.depends("order_id")
+    @api.depends("order_id", "order_id.company_id")
     def _compute_company_id(self):
         for line in self:
             line.company_id = (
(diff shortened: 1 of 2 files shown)


sale_order_line_input/models/sale_order.py
```diff
@@ -11,18 +11,40 @@ class SaleOrderLine(models.Model):
...
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        compute="_compute_company_id",
+        inverse=lambda self: self,
+        related=False,
+        readonly=True,
+        store=True,
+    )
+
+    @api.depends("order_id")
```
carlosdauden, line 25:
```suggestion
    @api.depends("order_id", "order_id.company_id")
```

sale_order_line_input/models/sale_order.py
```diff
@@ -11,18 +11,40 @@ class SaleOrderLine(models.Model):
         related='order_id.pricelist_id',
         readonly=True,
     )
+    # HACK: Overwrite field because the related fields do not get the default
+    # value correctly when user adds a new record.
+    company_id = fields.Many2one(
```
pedrobaeza, line 16: This hack is very conflicting and will affect all the sales orders, normal and input ones. Don't you have any other way (through `default_get` or similar) to get the same goal?

carlosdauden:
Is it still necessary?

PR #1405, opened 2021-01-03
[14.0][ADD] add module for add note on sale order line

sebastienbeau:


legalsylvain (approved):

yostashiro (approved):
Functional review. LGTM. 👍 

Just in case it is relevant, we made a similar module before and passed the note value to the corresponding field in stock move: https://github.com/qrtl/hls-custom/blob/12.0/sale_line_note/models/stock_rule.py#L10

PR #1407, opened 2021-01-04
[13.0] [MIG] sale_order_product_assortment: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT27472

Please @joao-p-marques  @chienandalu  review this.


sale_order_product_assortment/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+        string="Blacklisted Products",
+        compute="_compute_product_assortment_ids",
+    )
+    has_whitelist = fields.Boolean(compute="_compute_product_assortment_ids")
+    has_blacklist = fields.Boolean(compute="_compute_product_assortment_ids")
+
+    @api.depends("partner_id")
+    def _compute_product_assortment_ids(self):
+        # If we don't initialize the fields we get an error with NewId
+        self.whitelist_product_ids = self.env["product.product"]
```
chienandalu, line 25: Don't you loop over the recorset?

CarlosRoca13: In this case I want to initialize the field as a empty recordset. I don't know if this is the correct way

joao-p-marques: If that's what you want, that seems like the correct way, according to the [documentation](https://www.odoo.com/documentation/13.0/reference/orm.html#environment).
However, I don't really understand the logic... AFAICS, you are reassigning it [bellow](https://www.odoo.com/documentation/13.0/reference/orm.html#environment), right? What was the error with `NewId` you refer to in the comment?

chienandalu: But what about later?

CarlosRoca13: I just want to have records if a partner is defined... so here you have were we add the records https://github.com/OCA/sale-workflow/pull/1407/files#diff-d857107550de88517b0f4cb32ae7770a25a630dc254b97191e7f71f262183089R52-R57

chienandalu: Can you write over multiple records in v13 with a direct assignment?

CarlosRoca13: @joao-p-marques that's the error that i catch if I don't use this...
image

joao-p-marques: Ok, does it give any more information in the logs?

chienandalu (approved):

chienandalu:
I resolved my doubt :)

sale_order_product_assortment/models/sale_order.py
CarlosRoca13: @joao-p-marques the logs just give me this info `2021-01-04 09:43:48,203 1 WARNING devel odoo.http: ('sale.order(<NewId 0x7f7c50380ef0>,).whitelist_product_ids', None)`

joao-p-marques: Ok. Your latest push also cleared it for me :smile: Good job!

joao-p-marques (approved):

victoralmau (approved):

HaraldPanten (approved):

Issue #1391, opened 2020-12-22
[12.0] sale_triple_discount incorrect result


ioans73:
When creating a line with 3 different discounts in a Sale Order:

Unit price: $150
Discount 1: 20%
Discount 2: 10%
Discount 3: 5%

We save and the subtotal is set to: $102.06

Now once we modify the value of Discount 2, setting it to 5% and saving the line, we can see that the field does not update, maintaining its initial value and therefore, the subtotal doesn't re-calculate either.

Testing was done on a Runbot instance.

Does anyone know anything about this issue?

![sale_triple_discount](https://user-images.githubusercontent.com/29223264/102866734-094f9580-4438-11eb-864e-ece8df0b14ac.gif)

rousseldenis:
@eLBati @tafaRU @SimoRubi Could you help here ?

SimoRubi:
There is some history about triple discount cache management but as far as I know, it is for v10: https://github.com/OCA/sale-workflow/issues/817.

I don't know/recall of any similar issue in v12 (until now).

ioans73:
To solve the problem, at least temporarily, we've refactored some of the module's code.
Our clients haven't complained since we made these changes.
If you think it could be helpful, we can make a PR to propose the changes.

SimoRubi:
If the module has an issue, sharing a fix is helpful for sure :) thanks!

ioans73:
> If the module has an issue, sharing a fix is helpful for sure :) thanks!

PR done https://github.com/OCA/sale-workflow/pull/1410

SimoRubi:
@ioans73 sorry to bother again, since this is now solved, can you close?

PR #1410, opened 2021-01-07
[FIX] sale_triple_discount: Onchange cache management

ioans73:
Fix [OCA/sale-workflow#1391](https://github.com/OCA/sale-workflow/issues/1391)

We've refactored and deleted some functions, based on this old commit [10e71eb69a63ffb356f8beb14e8137a514eb28df](https://github.com/OCA/sale-workflow/commit/10e71eb69a63ffb356f8beb14e8137a514eb28df#diff-4582452e086bfb5976b52fbcd310a94a933ef3f2ea620a7f9432784935ef3002R15).
We don't know exactly why the method to calculate the discounts was modified, but with these changes it works correctly for us

### DIFF
--- a/sale_triple_discount/models/sale_order.py
+++ b/sale_triple_discount/models/sale_order.py
@@ -9,14 +9,6 @@ from odoo import api, models
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
 
-    @api.depends('order_line.price_total')
-    def _amount_all(self):
-        prev_values = dict()
-        for order in self:
-            prev_values.update(order.order_line.triple_discount_preprocess())
-        super(SaleOrder, self)._amount_all()
-        self.env['sale.order.line'].triple_discount_postprocess(prev_values)
-
     @api.multi
     def _get_tax_amount_by_group(self):
         # Copy/paste from standard method in sale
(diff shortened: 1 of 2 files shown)


SimoRubi (changes requested):
Hi, you are basically rolling back https://github.com/OCA/sale-workflow/pull/768 but I'm happy to see that those changes aren't needed anymore (since the added test doesn't fail in your PR).

I have found one problem in your approach though, here the steps to reproduce it in your PR's runbot:

1. In Sales > Configuration > Settings, enable 'discounts'
2. Create a sale order with one line having:
    - price unit = 25
    - quantity = 65
    - discount = 50
    - discount2 = 13

**Actual behavior**
The line's subtotal is 707

**Expected behavior**
The line's subtotal is 706.88 

I have found out that this happens because the discounted price unit is 28,275 but it is rounded to 28.28 as soon as you save it in the order line, due to the field's precision.

EDIT: I'm addressing this exact issue in https://github.com/OCA/account-invoicing/pull/876 for account invoice's triple discount

SimoRubi:
@ioans73 since the same issue has already been solved from https://github.com/OCA/sale-workflow/pull/1496, can you close this one?

ioans73:
Yes, I close it.
Thanks

PR #1365, opened 2020-12-04
[12.0][ADD] sale_pwa_cache

Tardo:
Depends on: https://github.com/OCA/web/pull/1756 & https://github.com/OCA/product-attribute/pull/735

cc @tecnativa TT23822

PR #1418, opened 2021-01-18
[ADD] sale_order_product_recommendation_pwa_cache: Glue Module

chienandalu:
- [ ] Depends on https://github.com/OCA/sale-workflow/pull/1365 `sale_pwa_cache`

cc @Tecnativa TT27734

### DIFF
--- a/sale_order_product_recommendation_pwa_cache/readme/CONTRIBUTORS.rst
+++ b/sale_order_product_recommendation_pwa_cache/readme/CONTRIBUTORS.rst
@@ -1,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_
+
+  * David Vidal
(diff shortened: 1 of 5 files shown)


sale_order_product_recommendation_pwa_cache/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Firstname Lastname <email.address@example.org> (optional company website url)
```
pedrobaeza: Change this

chienandalu: done

rousseldenis:
@pedrobaeza @chienandalu  I'm wondering if the pwa mechanism couldn't be transversal to be applied in the whole instance to avoid too much glue modules like this ?

chienandalu:
The thing is that now the approach is to have a minimum sales form and from there add just the needed elements for the salesmen to work properly (or having a custom one)

pedrobaeza:
The main amount of work is in the base module (`web_pwa_cache`), but each area can require a module for:

- Data cache specification.
- Specific optimized views.

That's why we have some glue modules for optional things. You can't force everyone wanting to have sales to put the product recommendation or the secondary unit.

chienandalu:
As #1365 is dismissed in favour of the latter v13 developement I think this can be as well (as it's migrated in that version as well)

PR #1427, opened 2021-01-22
[13.0][FIX] sale_order_type: Enable copying a sale order with different type_id

ntsirintanis:
Enable maintaining proper sequencing when copying a sale order and then change type_id. Do all this while keeping type_id copy=True

### DIFF
--- a/sale_order_type/__manifest__.py
+++ b/sale_order_type/__manifest__.py
@@ -17,8 +17,6 @@
     "Tecnativa,"
     "Agile Business Group,"
     "Niboo,"
-    "Therp BV,"
-    "Sunflower IT,"
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "license": "AGPL-3",
--- a/sale_order_type/__manifest__.py
+++ b/sale_order_type/__manifest__.py
@@ -4,6 +4,8 @@
 # Copyright 2016 Lorenzo Battistini
 # Copyright 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
 # Copyright 2018 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2021 Therp BV <https://therp.nl>.
+# Copyright 2021 Sunflower IT < https://sunflowerweb.nl>.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
@@ -15,6 +17,8 @@
     "Tecnativa,"
     "Agile Business Group,"
     "Niboo,"
+    "Therp BV,"
+    "Sunflower IT,"
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "license": "AGPL-3",
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -1,5 +1,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Therp BV <https://therp.nl>.
+# Copyright 2021 Sunflower IT < https://sunflowerweb.nl>.
 
 from datetime import datetime, timedelta
 
@@ -106,11 +108,9 @@ class SaleOrder(models.Model):
     @api.model
     def create(self, vals):
         if vals.get("name", "/") == "/" and vals.get("type_id"):
-            sale_type = self.env["sale.order.type"].browse(vals["type_id"])
-            if sale_type.sequence_id:
-                vals["name"] = sale_type.sequence_id.next_by_id(
-                    sequence_date=vals.get("date_order")
-                )
+            next_sequence = self._get_next_sequence(vals["type_id"])
+            if next_sequence:
+                vals["name"] = next_sequence
         return super(SaleOrder, self).create(vals)
 
     def write(self, vals):
@@ -150,6 +150,13 @@ class SaleOrder(models.Model):
         if self.type_id:
             res["sale_type_id"] = self.type_id.id
         return res
+
+    def _get_next_sequence(self, type_id):
+        sale_type = self.env["sale.order.type"].browse(type_id)
+        sequence = False
+        if sale_type.sequence_id:
+            sequence = sale_type.sequence_id.next_by_id()
+        return sequence
 
 
 class SaleOrderLine(models.Model):
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -108,7 +108,7 @@ class SaleOrder(models.Model):
     @api.model
     def create(self, vals):
         if vals.get("name", "/") == "/" and vals.get("type_id"):
-            next_sequence = self._get_next_sequence(vals["type_id"])
+            next_sequence = self._get_next_sequence(vals)
             if next_sequence:
                 vals["name"] = next_sequence
         return super(SaleOrder, self).create(vals)
@@ -151,11 +151,13 @@ class SaleOrder(models.Model):
             res["sale_type_id"] = self.type_id.id
         return res
 
-    def _get_next_sequence(self, type_id):
-        sale_type = self.env["sale.order.type"].browse(type_id)
+    def _get_next_sequence(self, vals):
+        sale_type = self.env["sale.order.type"].browse(vals.get("type_id"))
         sequence = False
         if sale_type.sequence_id:
-            sequence = sale_type.sequence_id.next_by_id()
+            sequence = sale_type.sequence_id.next_by_id(
+                sequence_date=vals.get("date_order")
+            )
         return sequence
 
 
--- a/sale_order_type/readme/CONTRIBUTORS.rst
+++ b/sale_order_type/readme/CONTRIBUTORS.rst
@@ -39,4 +39,12 @@
 
   * Giovanni Serra <giovanni@gslab.it>
 
+* `Therp BV <https://therp.nl>`_
+
+  * Nikos Tsirintanis <ntsirintanis@therp.nl>
+
+* `Sunflower IT <https://sunflowerweb.nl>`_
+
+  * Tom Blauwendraat <info@sunflowerweb.nl>
+
 Do not contact contributors directly about support or help with technical issues.
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -194,7 +194,7 @@ class TestSaleOrderType(common.TransactionCase):
         order.onchange_type_id()
         self.assertEqual(order.type_id.route_id, order.order_line[1].route_id)
 
-    def _test_sale_order_in_draft_state_update_name(self):
+    def test_sale_order_in_draft_state_update_name(self):
         order = self.create_sale_order()
         order.onchange_partner_id()
         order.onchange_type_id()
@@ -207,7 +207,7 @@ class TestSaleOrderType(common.TransactionCase):
         self.assertEqual(order.type_id, self.sale_type_quot)
         self.assertTrue(order.name.startswith("TQU"))
 
-    def _test_sale_order_in_sent_state_update_name(self):
+    def test_sale_order_in_sent_state_update_name(self):
         order = self.create_sale_order()
         order.onchange_partner_id()
         order.onchange_type_id()
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -194,7 +194,7 @@ class TestSaleOrderType(common.TransactionCase):
         order.onchange_type_id()
         self.assertEqual(order.type_id.route_id, order.order_line[1].route_id)
 
-    def test_sale_order_in_draft_state_update_name(self):
+    def _test_sale_order_in_draft_state_update_name(self):
         order = self.create_sale_order()
         order.onchange_partner_id()
         order.onchange_type_id()
@@ -207,7 +207,7 @@ class TestSaleOrderType(common.TransactionCase):
         self.assertEqual(order.type_id, self.sale_type_quot)
         self.assertTrue(order.name.startswith("TQU"))
 
-    def test_sale_order_in_sent_state_update_name(self):
+    def _test_sale_order_in_sent_state_update_name(self):
         order = self.create_sale_order()
         order.onchange_partner_id()
         order.onchange_type_id()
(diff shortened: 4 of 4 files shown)


ntsirintanis:
@NL66278 @gfcapalbo @lfreeke please review.

sale_order_type/models/sale.py
```diff
@@ -73,12 +74,27 @@ def onchange_type_id(self):
 
     @api.model
     def create(self, vals):
+        if self:
+            # This is a copy
+            # Give it a generic name
+            vals["name"] = "Copy of %s" % self[0].name
```
thomaspaulb: `_("Copy of %s") translates this

sale_order_type/models/sale.py
```diff
@@ -73,12 +74,27 @@ def onchange_type_id(self):
...
+            # Force user to select type_id, so proper
+            # sequencing will be maintained
+            vals["type_id"] = False
```
thomaspaulb: Why not just set [copy=False](https://github.com/OCA/sale-workflow/blob/13.0/sale_order_type/models/sale.py#L23) ?

sale_order_type/models/sale.py
```diff
@@ -73,12 +74,27 @@ def onchange_type_id(self):
 
     @api.model
     def create(self, vals):
+        if self:
+            # This is a copy
```
thomaspaulb: This seems strange to me, `create` is an `api.model` so I don't think that any record should be in `self`? Or there is something I don't know about.

My gut feeling says to do this in [copy](https://github.com/OCA/sale-workflow/blob/13.0/sale_order_revision/models/sale_order.py#L47), checking if `name` is filled with something other then `/`, if yes, prepend the "Copy of " to `default['name']`.

sale_order_type/models/sale.py
```diff
@@ -73,12 +74,27 @@ def onchange_type_id(self):
...
+            # Give it a generic name
+            vals["name"] = "Copy of %s" % self[0].name
+            # Force user to select type_id, so proper
+            # sequencing will be maintained
+            vals["type_id"] = False
         if vals.get("name", "/") == "/" and vals.get("type_id"):
-            sale_type = self.env["sale.order.type"].browse(vals["type_id"])
-            if sale_type.sequence_id:
-                vals["name"] = sale_type.sequence_id.next_by_id()
+            vals["name"] = self._get_sequencing(vals["type_id"]) or vals["name"]
```
thomaspaulb: The `or vals['name']` part is risky, because in weird cases this will result in an IndexError because `name` is not in vals. Better to be safe and go for:

```
next_sequence = self._get_sequencing(vals["type_id"])
if next_sequence:
    vals["name"] = next_sequence
```

thomaspaulb (changes requested):

sale_order_type/models/sale.py
```diff
@@ -73,12 +74,27 @@ def onchange_type_id(self):
...
 
+    def write(self, vals):
+        """
+        Maintain proper sequencing when copying
+
+        """
+        type_id = vals.get("type_id")
+        if type_id:
+            vals["name"] = self._get_sequencing(type_id) or self.name
+        super(SaleOrder, self).write(vals)
```
thomaspaulb: I gather this is for when the `type_id` changes and the number needs to change as well. That's cool but, this looks like it will get incremented every time the `type_id` is written, which is not what you want.

Also, if records in `self` are many, then many sale orders will get the same sequence, which is also not good.

You can get this right by, before `super`, looping the records "for this in self", compare their type_id with what it's going to be, and if you see a change, add it to a {so: new_type_id, so: new_type_id} style dict.

On the remaining sale orders (self - dict.keys()), you call super like you call it now, but without vals["name"].

Afterwards, for each one that had their type changed, you add the new sequence to vals and call super(SaleOrder, this).write(vals).

That way you won't slow down regular writes by adding a loop when it's not needed.

thomaspaulb (changes requested):

sale_order_type/models/sale.py
ntsirintanis: This was to avoid changing the field definition, just in case copying is needed someplace else in the code.

sale_order_type/models/sale.py
ntsirintanis: Basically i pdbed my way through this, and noticed that as soon as we duplicate a record, self in create() is basically the record we are duplicating. I'll look through this further.

thomaspaulb: You're right actually, it's passed [here](https://github.com/OCA/OCB/blob/13.0/odoo/models.py#L4664). It looks more by accident though than on purpose... and what if other module authors try to be smart, save some lines of Python, and use a filled recordset to call `create` on? Then they'll get their type deleted.

sale_order_type/models/sale.py
thomaspaulb: Well, but if people call copy someplace else in the code, `copy` in turn calls `create`, so they'll also run into the code here?

sale_order_type/models/sale.py
ntsirintanis: Yes you are right, I should multi this and your suggestion is absolutely the way to do it.

sale_order_type/models/sale.py
ntsirintanis: Basically you are right, I should just set copy=False to type_id and be done with it.

sale_order_type/models/sale.py
NL66278: @ntsirintanis I agree with @thomaspaulb here. That copy calls create with self filled is an implementation detail, not part of the API, and therefore subject to change without notice. The canonical way is to override the copy method.

sale_order_type/models/sale.py
NL66278: For usability it is much better to leave copy=True as is. Users are copying existing sale orders to re-use as much of the data from the existing record. That is including the type.

So I would go for a totally different approach. Clearly the `_("Copy of")` approach is valid for types that are not linked to a sequence. Otherwise get the next sequence for the type_id. Implement an on_change handler for type_id, and if the value really changes (compare with self._origin), get the next sequence if the new value also is linked to a sequence. If the old name was linked to a sequence, but a new one not, clear the name and force the user to enter a value.

sale_order_type/models/sale.py
```diff
@@ -74,11 +75,34 @@ def onchange_type_id(self):
...
     def create(self, vals):
         if vals.get("name", "/") == "/" and vals.get("type_id"):
-            sale_type = self.env["sale.order.type"].browse(vals["type_id"])
-            if sale_type.sequence_id:
-                vals["name"] = sale_type.sequence_id.next_by_id()
+            vals["name"] = self._get_sequencing(vals["type_id"]) or self.name or "/"
         return super(SaleOrder, self).create(vals)
 
+    def write(self, vals):
+        """
```
NL66278: doc string can be a oneliner

sale_order_type/models/sale.py
```diff
@@ -74,11 +75,34 @@ def onchange_type_id(self):
...
+        if "type_id" not in vals:
+            return super(SaleOrder, self.write(vals))
+        type_id = vals.get("type_id")
+        # Store orders that type_id has changed
+        updated_orders = self.env["sale.order"]
+        for this in self:
+            # type_id remains the same.
+            if this.type_id.id == type_id:
+                continue
+            sequencing = self._get_sequencing(type_id)
```
NL66278: I don't like the name of this variable. It is a verb. Why not stick with next_sequence?

sale_order_type/models/sale.py
```diff
@@ -74,11 +75,34 @@ def onchange_type_id(self):
...
+        """
+        if "type_id" not in vals:
+            return super(SaleOrder, self.write(vals))
+        type_id = vals.get("type_id")
+        # Store orders that type_id has changed
+        updated_orders = self.env["sale.order"]
+        for this in self:
+            # type_id remains the same.
+            if this.type_id.id == type_id:
+                continue
```
NL66278: Also continue if the type_id is not connected to a sequence.

ntsirintanis: yes, this is checked in _get_next_sequence()

NL66278 (approved):
:+1: LGTM And tests cover the different cases.

sale_order_type/models/sale.py
```diff
@@ -74,11 +76,33 @@ def onchange_type_id(self):
...
+            next_sequence = self._get_next_sequence(type_id)
+            # No next sequence for this type_id
+            if not next_sequence:
+                continue
+            # Change the name
+            new_vals = vals.copy()
+            new_vals["name"] = next_sequence
+            super(SaleOrder, this).write(new_vals)
+            updated_orders += this
+        super(SaleOrder, self - updated_orders).write(vals)
```
thomaspaulb: Add a `return` in front of this just in case someone relies on it

thomaspaulb (approved):

rafaelbn:
Please @ntsirintanis could you review conflicts? Thank you!

etobella:
@ntsirintanis there are some conflicts. Can you rebase and fix them?

ntsirintanis:
> @ntsirintanis there are some conflicts. Can you rebase and fix them?

I see that another PR introduced the same changes in write(). So I removed mine, and kept edits in create() only.

sale_order_type/models/sale.py
```diff
@@ -151,6 +151,13 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    def _get_next_sequence(self, type_id):
+        sale_type = self.env["sale.order.type"].browse(type_id)
+        sequence = False
+        if sale_type.sequence_id:
+            sequence = sale_type.sequence_id.next_by_id()
```
rousseldenis: @ntsirintanis Your implementation removes the sequence_date.

ntsirintanis:
@rousseldenis thank you very very much for spotting and pinging me about this.

thomaspaulb:
@ntsirintanis Can you rebase on latest 13.0 so we can see if the tests fail?

thomaspaulb:
Currently it's:

```
2023-06-06 07:52:27,476 251 INFO odoo odoo.modules.module: odoo.addons.sale_advance_payment.tests.test_sale_advance_payment running tests. 
2023-06-06 07:52:27,853 251 ERROR odoo odoo.sql_db: bad query: INSERT INTO "res_currency_rate" ("id", "create_uid", "create_date", "write_uid", "write_date", "company_id", "currency_id", "name", "rate") VALUES (nextval('res_currency_rate_id_seq'), 1, (now() at time zone 'UTC'), 1, (now() at time zone 'UTC'), 1, 2, '2023-06-06', 1.2) RETURNING id
ERROR: duplicate key value violates unique constraint "res_currency_rate_unique_name_per_day"
DETAIL:  Key (name, currency_id, company_id)=(2023-06-06, 2, 1) already exists.
 
2023-06-06 07:52:27,855 251 INFO odoo unittest.suite: ====================================================================== 
2023-06-06 07:52:27,855 251 ERROR odoo unittest.suite: ERROR: setUpClass (odoo.addons.sale_advance_payment.tests.test_sale_advance_payment.TestSaleAdvancePayment)
  File "/opt/odoo/odoo/tests/common.py", line 190, in _handleClassSetUp
    setUpClass()
  File "/__w/sale-workflow/sale-workflow/sale_advance_payment/tests/test_sale_advance_payment.py", line 74, in setUpClass
    {"rate": 1.20, "currency_id": cls.currency_usd.id}
  File "<decorator-gen-3>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 338, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/odoo/models.py", line 3825, in create
    records = self._create(data_list)
  File "/opt/odoo/odoo/models.py", line 3931, in _create
    cr.execute(query, params)
  File "/opt/odoo/odoo/sql_db.py", line 173, in wrapper
    return f(self, *args, **kwargs)
  File "/opt/odoo/odoo/sql_db.py", line 250, in execute
    res = self._obj.execute(query, params)
psycopg2.IntegrityError: duplicate key value violates unique constraint "res_currency_rate_unique_name_per_day"
DETAIL:  Key (name, currency_id, company_id)=(2023-06-06, 2, 1) already exists.
```

ntsirintanis:
> Currently it's:
> 
> ```
> 2023-06-06 07:52:27,476 251 INFO odoo odoo.modules.module: odoo.addons.sale_advance_payment.tests.test_sale_advance_payment running tests. 
> 2023-06-06 07:52:27,853 251 ERROR odoo odoo.sql_db: bad query: INSERT INTO "res_currency_rate" ("id", "create_uid", "create_date", "write_uid", "write_date", "company_id", "currency_id", "name", "rate") VALUES (nextval('res_currency_rate_id_seq'), 1, (now() at time zone 'UTC'), 1, (now() at time zone 'UTC'), 1, 2, '2023-06-06', 1.2) RETURNING id
> ERROR: duplicate key value violates unique constraint "res_currency_rate_unique_name_per_day"
> DETAIL:  Key (name, currency_id, company_id)=(2023-06-06, 2, 1) already exists.
>  
> 2023-06-06 07:52:27,855 251 INFO odoo unittest.suite: ====================================================================== 
> 2023-06-06 07:52:27,855 251 ERROR odoo unittest.suite: ERROR: setUpClass (odoo.addons.sale_advance_payment.tests.test_sale_advance_payment.TestSaleAdvancePayment)
> Traceback (most recent call last):
>   File "/opt/odoo/odoo/tests/common.py", line 190, in _handleClassSetUp
>     setUpClass()
>   File "/__w/sale-workflow/sale-workflow/sale_advance_payment/tests/test_sale_advance_payment.py", line 74, in setUpClass
>     {"rate": 1.20, "currency_id": cls.currency_usd.id}
>   File "<decorator-gen-3>", line 2, in create
>   File "/opt/odoo/odoo/api.py", line 338, in _model_create_multi
>     return create(self, [arg])
>   File "/opt/odoo/odoo/models.py", line 3825, in create
>     records = self._create(data_list)
>   File "/opt/odoo/odoo/models.py", line 3931, in _create
>     cr.execute(query, params)
>   File "/opt/odoo/odoo/sql_db.py", line 173, in wrapper
>     return f(self, *args, **kwargs)
>   File "/opt/odoo/odoo/sql_db.py", line 250, in execute
>     res = self._obj.execute(query, params)
> psycopg2.IntegrityError: duplicate key value violates unique constraint "res_currency_rate_unique_name_per_day"
> DETAIL:  Key (name, currency_id, company_id)=(2023-06-06, 2, 1) already exists.
> ```

It looks like the error is still the same

thomaspaulb:
@ntsirintanis Could you do yet another rebase? It seems the pipeline was fixed on June 13: 

https://github.com/OCA/sale-workflow/pull/2301

thomaspaulb:
@rousseldenis Green! Merge is now ready to be retried.

sale_order_type/__manifest__.py
```diff
@@ -15,6 +17,8 @@
     "Tecnativa,"
     "Agile Business Group,"
     "Niboo,"
+    "Therp BV,"
```
rousseldenis: @ntsirintanis Maybe too few changes to take authorship ?

ntsirintanis: @rousseldenis actually this was a new feature, when I opened the PR 2.5 years ago. Then, another PR with a similar feature was merged, thus making this PR less significant. No matter, you are right, I will remove this.

rousseldenis: :sweat_smile: Too much opened PR's on this repo. Harsh work to decrease their number.

rousseldenis (approved):
Code review

sale_order_type/__manifest__.py
```diff
@@ -4,6 +4,8 @@
 # Copyright 2016 Lorenzo Battistini
 # Copyright 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
 # Copyright 2018 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2021 Therp BV <https://therp.nl>.
+# Copyright 2021 Sunflower IT < https://sunflowerweb.nl>.
```
gfcapalbo, line 8: '23

PR #1439, opened 2021-02-03
[12.0][IMP] sales_team_security: Auto-followers (create and remove) in child partners

victoralmau:
Auto-followers (create and remove) in child partners when salesman is changed (write function)

Please @pedrobaeza review it.

Locked by: 
- [x] https://github.com/OCA/sale-workflow/issues/1446 `sale_order_lot_selection`

@Tecnativa TT27986

### DIFF
--- a/partner_contact_sale_info_propagation/models/res_partner.py
+++ b/partner_contact_sale_info_propagation/models/res_partner.py
@@ -1,17 +1,27 @@
 # Copyright 2019 Tecnativa - Ernesto Tejeda
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
 from lxml import etree
+from odoo.tools import config
 
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def _check_propagation_allowed(self):
+        return bool(
+            not config["test_enable"]
+            or (config["test_enable"] and self.env.context.get("test_propagation"))
+        )
+
     @api.multi
     def write(self, vals):
         """Propagate Salesperson and Sales Channel change in the partner to the
         child contacts."""
+        if not self._check_propagation_allowed():
+            return super().write(vals)
         for record in self:
             if 'user_id' in vals:
                 childs = record.mapped('child_ids').filtered(
@@ -23,22 +33,24 @@ class ResPartner(models.Model):
                     lambda r: not r.team_id or r.team_id == record.team_id)
                 if childs:
                     childs.write({'team_id': vals['team_id']})
-        return super(ResPartner, self).write(vals)
+        return super().write(vals)
 
     @api.model
     def create(self, vals):
+        if not self._check_propagation_allowed():
+            return super().create(vals)
         if 'parent_id' in vals:
             if 'user_id' not in vals:
                 vals.update(user_id=self.browse(vals['parent_id']).user_id.id)
             if 'team_id' not in vals:
                 vals.update(team_id=self.browse(vals['parent_id']).team_id.id)
-        return super(ResPartner, self).create(vals)
+        return super().create(vals)
 
     @api.onchange('parent_id')
     def onchange_parent_id(self):
         """Change Salesperson or Sales Channel if the parent company changes
         and there's no Salesperson or Sales Channel defined yet"""
-        res = super(ResPartner, self).onchange_parent_id()
+        res = super().onchange_parent_id()
         if self.parent_id and self.parent_id != self:
             parent = self.parent_id
             if not self.user_id:
--- a/sales_team_security/models/res_partner.py
+++ b/sales_team_security/models/res_partner.py
@@ -1,12 +1,17 @@
 # Copyright 2016-2018 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import api, models
+from odoo import api, fields, models
 from lxml import etree
 
 
 class ResPartner(models.Model):
     _inherit = 'res.partner'
+
+    # add indexes for better performance on record rules
+    user_id = fields.Many2one(index=True)
+    team_id = fields.Many2one(index=True)
 
     @api.model
     def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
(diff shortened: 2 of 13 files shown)


victoralmau:
Now exists some incompatibility with `partner_contact_sale_info_propagation` addon in some tests case.
I will change to prevent it but IMO it's need to define incompatibility addon in manifest.
What do you think @pedrobaeza ?

rousseldenis:
@victoralmau could you add a little explanation of this addition in DESCRIPTION.rst ?

sales_team_security/tests/test_sales_team_security.py
```diff
@@ -34,6 +51,17 @@ def setUpClass(cls):
                 (4, cls.env.ref("sales_team.group_sale_salesman").id)],
             "sale_team_id": cls.team.id,
         })
+        cls.partner_contact_sale_info_propagation_installed = cls._is_module_installed(
```
pedrobaeza: You should make tests independently from this module. If it's interfering, then the problem in in the other module, and would require a mitigation method like this one:

https://github.com/OCA/bank-payment/blob/3c96fedbb95060c9cd3a2aa87d383324d486bee1/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

sales_team_security/migrations/12.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """At installation time, need to followers sync"""
+    partners = env['res.partner'].search([
```
pedrobaeza: This should be a post_init_hook, but also being called from this migration script.

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
         """
         if self.parent_id and self.parent_id.team_id and not self.team_id:
             self.team_id = self.parent_id.team_id.id
+
+    @api.depends('company_type', 'child_ids', 'child_ids.user_id')
```
pedrobaeza: Depends is only for computed field, so this doesn't add anything.

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
...
+            partners = self.env['res.partner'].search([
+                ('id', 'in', (item.child_ids.ids + item.ids))
+            ])
+            user_ids = partners.mapped('user_id')
+            if user_ids:
+                partners.message_subscribe(
+                    partner_ids=user_ids.mapped('partner_id').ids
+                )
+
+    @api.depends('parent_id', 'child_ids')
```
pedrobaeza: Remove it

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
...
+    def change_user_id_sales_team_security(self):
+        for item in self:
+            partner = item.get_partner_to_sync()
+            if partner:
+                partner.sync_followers_in_childs()
+
+    @api.depends('company_type', 'parent_id', 'child_ids')
```
pedrobaeza: Remove it

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
...
+    def get_partner_to_sync(self):
```
pedrobaeza: Why not using `self.commercial_partner_id`?

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
...
+        2- Call change_user_id_sales_team_security function to
+        sync followers
+        """
+        for item in self:
+            if 'user_id' in vals and item.user_id:
+                partner_to_sync = item.get_partner_to_sync()
+                if partner_to_sync:
+                    ids = (partner_to_sync.child_ids.ids + partner_to_sync.ids)
+                    partners = self.env['res.partner'].search([('id', 'in', ids)])
+                    partners.message_unsubscribe(
```
pedrobaeza: Why this code is not in `sync` method? If you call it sync, do the real sync, not only half of the task.

sales_team_security/models/res_partner.py
```diff
@@ -40,3 +41,54 @@ def _onchange_parent_id_sales_team_security(self):
         """
         if self.parent_id and self.parent_id.team_id and not self.team_id:
             self.team_id = self.parent_id.team_id.id
+
+    @api.depends('company_type', 'child_ids', 'child_ids.user_id')
+    def sync_followers_in_childs(self):
+        for item in self.filtered(lambda x: x.company_type == 'company'):
+            partners = self.env['res.partner'].search([
```
pedrobaeza: I don't see here any extraction of the delivery / invoicing contacts. There's a method for that in the partner.

pedrobaeza (changes requested):

sales_team_security/models/res_partner.py
victoralmau: Changed it to `commercial_partner_id`

sales_team_security/models/res_partner.py
victoralmau: Function apply to `child` ids

sales_team_security/tests/test_sales_team_security.py
victoralmau: I understand what you say and your proposal, but I don't really know how to interfere with it.
The functionality that is added in this PR auto-add followers of the salesmans (`user_id`) of the contacts. (only in company contacts). The tests check this and work fine.
The `partner_contact_sale_info_propagation` addon auto-define salesman (`user_id`) for all parent contacts, which causes the expected result of the tests to change. Any proposal?

pedrobaeza: Yes, to avoid such propagation on test mode if you are not testing the other module, which style is the link I sent you.

pedrobaeza (approved):
Adjusted for syncing followers according invoice and delivery address + main contact.

victoralmau:
Actually, test fails when `partner_contact_sale_info_propagation` installed. 
I'm not really sure if both addons can work fine together and maybe solution can be add incompatibility in manifest file.
What do you think about it?

chienandalu (approved):

victoralmau:
It's ready to merge now, right?

pedrobaeza:
@victoralmau Travis is red

victoralmau:
> @victoralmau Travis is red

I have added a commit to resolve Travis. It is related to https://github.com/OCA/sale-workflow/pull/1439#issuecomment-866634495. (This will be necessary in v13 as well).

pedrobaeza:
The patch for making this green is not correct, as you should instead prevent in the other module to propagate the information if you are on test mode and not testing your module (through the context key).

victoralmau:
Thanks Pedro, would the current change be correct?

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -1,50 +1,64 @@
...
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def _check_propagation_allowed(self):
+        return bool(
+            not config["test_enable"]
+            or (
+                config["test_enable"]
+                and not self.env.context.get("prevent_propagation", False)
```
pedrobaeza: This should be the exact contrary:
```suggestion
                and not self.env.context.get("test_propagation")
```
and only enable such context on this module tests.

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -1,50 +1,64 @@
...
-                    lambda r: not r.user_id or r.user_id == record.user_id)
-                if childs:
-                    childs.write({'user_id': vals['user_id']})
-            if 'team_id' in vals:
-                childs = record.mapped('child_ids').filtered(
-                    lambda r: not r.team_id or r.team_id == record.team_id)
-                if childs:
-                    childs.write({'team_id': vals['team_id']})
-        return super(ResPartner, self).write(vals)
+        if self._check_propagation_allowed():
```
pedrobaeza: Avoid this diff doing an early return:
```suggestion
        if not self._check_propagation_allowed():
            return super().write(vals)
```

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -1,50 +1,64 @@
...
         and there's no Salesperson or Sales Channel defined yet"""
-        res = super(ResPartner, self).onchange_parent_id()
-        if self.parent_id and self.parent_id != self:
-            parent = self.parent_id
-            if not self.user_id:
-                res.setdefault('value', {}).update(user_id=parent.user_id)
-            if not self.team_id:
-                res.setdefault('value', {}).update(team_id=parent.team_id)
+        res = super().onchange_parent_id()
+        if self._check_propagation_allowed():
```
pedrobaeza: Is this also needed?

pedrobaeza (changes requested):

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -1,44 +1,57 @@
...
-                    lambda r: not r.user_id or r.user_id == record.user_id)
-                if childs:
-                    childs.write({'user_id': vals['user_id']})
-            if 'team_id' in vals:
-                childs = record.mapped('child_ids').filtered(
-                    lambda r: not r.team_id or r.team_id == record.team_id)
-                if childs:
-                    childs.write({'team_id': vals['team_id']})
-        return super(ResPartner, self).write(vals)
+        if not self._check_propagation_allowed():
```
pedrobaeza, line 23: You are still putting a big diff as not doing the early return.

partner_contact_sale_info_propagation/models/res_partner.py
victoralmau: I have verified that this is not necessary.

partner_contact_sale_info_propagation/models/res_partner.py
victoralmau: Sorry, now the diff is smaller.

pedrobaeza (approved):

PR #1440, opened 2021-02-04
[13.0] sale_fixed_discount - wrong value for the field price_reduce on SO line #1411 + Wrong tax_amount displayed on reporting #1424

mmarcot:
Hello,
This pull request aims to fix these issues on the branch 13.0:

- [12.0] sale_fixed_amount - Wrong tax_amount displayed on reporting #1424
- [13.0] sale_fixed_discount: price_reduce in database is wrong #1411

Im new to OCA contributing. This is my first PR. I sent the signed CLA yesterday, not sure if it is already taken in account.

Hope it's not a problem to fix 2 issues with the same PR

pull request for the branch 12.0 => #1443

Issue #1424, opened 2021-01-21
[12.0] sale_fixed_discount - Wrong tax_amount displayed on reporting


MAG-Florian-Chevalier:
Hello,

On sale order reports, the field tax_amount seems to be incorrect.
The tax amount is calculated on the undiscounted price.
Untaxed total and total are correct, and tax amount is only wrong on report.

Regards,

rousseldenis:
@MAG-Florian-Chevalier Could you change description if it's related to sale_fixed_**discount**

Issue #1411, opened 2021-01-10
[13.0] sale_fixed_discount: price_reduce in database is wrong


mehdi-ghezal:
Steps to reproduce:

Create a sale order with 2 line.
* 1st line: price=10, discount=50%
* 2nd line: price=10, discount=5 (fixed)

Check in the database:
* 1st line: price_reduce=5
* 2nd line: price_reduce=10

PR #1443, opened 2021-02-04
[12.0][FIX] sale_fixed_discount -Wrong tax_amount displayed on reporting + wrong price_reduce value

mmarcot:
Hello,
This pull request aims to fix these issues on the branch **12.0**:
- sale_fixed_amount - Wrong tax_amount displayed on reporting #1424
- sale_fixed_discount: price_reduce in database is wrong #1411

The same PR for branch 13.0 is available here => #1440

### DIFF
--- a/sale_fixed_discount/models/sale_order.py
+++ b/sale_fixed_discount/models/sale_order.py
@@ -1,9 +1,12 @@
 # Copyright 2017-18 Eficent Business and IT Consulting Services S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from functools import partial
+
 from odoo import api, fields, models, _
 import odoo.addons.decimal_precision as dp
 from odoo.exceptions import ValidationError
+from odoo.tools.misc import formatLang
 
 
 class SaleOrderLine(models.Model):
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (commented):
Could you add a test case?

MAG-Florian-Chevalier:
Hello,

I have just tested this fix on runbot and everything works fine.
I had an issue with the amount_tax which wasn't correct on portal with if I had some lines with fixed and other with percentage discount and this fix also corrects it.
I attach the screenshots of the sale order and its portal view for the test I've made on runbot.
Please note that the tax I've selected here is included in price.
image
Screenshot_20210914_095530

Regards,
Florian

MAG-Florian-Chevalier (approved):
Tests on runbot showed that this code works, override _amount_by_group seems a good idea to me so it will fix the price everywhere including on portal and reports.

flotho (approved):
code review OK
Functionnal test with @MAG-Florian-Chevalier  for fixed discount OK

rafaelbn:
@mmarcot hello! Thank you! the point is to test and report the use case on runbot before this PR to reproduce the issue before and after. I couldn't reproduce the issue in runbot without this PR. That why I think @LoisRForgeFlow ask it.

Maybe the issue comes from taxes included in[price and in other case it works?

eLBati:
new version, with super: https://github.com/OCA/sale-workflow/pull/3983

PR #1449, opened 2021-02-08
[13.0][FIX] sale_product_multi_add: fix missing call to _onchange_discount()

Cedric-Pigeon:
When using the addon feature, discounts defined on list price are not correctly applied


Cedric-Pigeon:
@acsonefho can you have a look please?

acsonefho (approved):
Code LGTM

PR #1451, opened 2021-02-08
[13.0][FIX] sale_product_set: apply discount from pricelist

Cedric-Pigeon:


simahawk:
@Cedric-Pigeon I think this is now fixed via https://github.com/OCA/sale-workflow/pull/1589 for all possible onchanges

Issue #1446, opened 2021-02-05
[12.0] sale_order_lot_selection: failing tests


CarlosRoca13:
The tests on travis version 12.0 are failing by sale_order_lot_selection module, here you have a link to last travis run https://travis-ci.com/github/OCA/sale-workflow/jobs/479997013#L2198

ping @PCatinean @bealdav

chienandalu:
It's funny this didn't fail until now :thinking:

pedrobaeza:
Fix in #1458

PR #1456, opened 2021-02-10
[12.0][FIX] sale_order_lot_generator: Prevent error in test with other addons that use same base products

victoralmau:
Related to https://github.com/OCA/sale-workflow/issues/1446 (It solve only one of these errors)

Locked by:
- [ ] https://github.com/OCA/sale-workflow/pull/1458

Please @CarlosRoca13 and @chienandalu can you review it?

@Tecnativa TT28247


victoralmau:
Included these commit in https://github.com/OCA/sale-workflow/pull/1458

PR #1458, opened 2021-02-10
[12.0][FIX] sale_order_lot_selection + sale_order_lot_generator: Prevent error in test with other addons

victoralmau:
Related to https://github.com/OCA/sale-workflow/issues/1446 (`sale_order_lot_selection` + `sale_order_lot_generator`)

Please @CarlosRoca13 and @chienandalu can you review it?

@Tecnativa TT28247

### DIFF
--- a/sale_order_lot_generator/readme/CONTRIBUTORS.rst
+++ b/sale_order_lot_generator/readme/CONTRIBUTORS.rst
@@ -4,3 +4,7 @@ Akretion:
 * Florian da Costa <florian.dacosta@akretion.com>
 * Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
 * David BEAL <david.beal@akretion.com>
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+    * Víctor Martínez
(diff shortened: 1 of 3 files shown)


CarlosRoca13 (approved):

lmignon (changes requested):
The problem is somewhere else. If you run the tests alone, everything is OK

lmignon (changes requested):
The problem is somewhere else. If you run the tests alone, everything is OK
:-1:

lmignon (changes requested):
The problem is somewhere else. If you run the tests alone, everything is OK
:-1:

pedrobaeza:
@victoralmau instead of changing this, please change in the setup that the involved product is `consumable` instead. This problem comes from using demo data (that other modules may change), and that's why this is discouraged and not done in latest Odoo versions. So you can still reference the demo product, but writing explicitly such value, or use a newly created product for this test.

chienandalu (approved):

pedrobaeza:
@lmignon do you agree now?

lmignon (approved):

lmignon:
thank you for the fix @victoralmau

bealdav:
Thanks a lot @victoralmau I note it's a good approach to systemically copy demo in test. Thanks

victoralmau:
Actually travis fail related to `sale_order_lot_generator` https://github.com/OCA/sale-workflow/pull/1456

pedrobaeza:
You must include both in the same PR, or we won't be able to merge this. But do it as separate commits.

victoralmau:
Now included 2 commits (1 for each addon)

pedrobaeza:
OK, and green with no conflicting changes, so

/ocabot merge patch

marcelsavegnago (approved):

PR #1468, opened 2021-02-14
[13.0][FIX] sale_order_type Existing SO can't be cancelled after install

jobiols:
**Fixing sale_order_type**

If this module is installed in a environment where there are sales orders not in draft state. The added field type_id is left null
then the order can't be cancelled.

A hook is provided to fill the type_id with data. It is a hook but not a migration becouse it must run also at first install.
It was tested in a multicompany environment.


jobiols:
Hi guys
I have fixed a bug in sale_order_type as stated above, then I solved some linter issues, 
but the las one was a redefined-builtin in another module, after fixing it **Travis stopped showing logs**.
I try even to close and reopen the PR getting the same...

Any help will be apreciated.
Thanks

jobiols:
Ok a after a day Travis shows a log, but I can't understand the error, i will try changing some thing.

jobiols:
Hi @rvalyi  I left this PR for a while, but I am going to return this weekend to resolve the conflict. We'll be in touch.

jobiols:
Hi @rvalyi, I have just fix all issues in this PR.

PR #1469, opened 2021-02-15
[12.0][IMP] sale_order_invoicing_finished_task: Avoid cancel SO with invoiced tasks

carlosdauden:
Avoid cancel SO with invoiced lines to don't have inconsistent lines if we want reconfirm _check_sale_line_state restrict raises error:
https://github.com/OCA/sale-workflow/blob/c38c050b81ac5f82a38c06ff90ebf7b65f5c9608/sale_order_invoicing_finished_task/models/project.py#L62-L66

@Tecnativa TT28217

### DIFF
--- a/sale_order_invoicing_finished_task/models/sale_order.py
+++ b/sale_order_invoicing_finished_task/models/sale_order.py
@@ -2,7 +2,8 @@
 # Copyright 2017 Carlos Dauden <carlos.dauden@tecnativa.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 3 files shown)


sale_order_invoicing_finished_task/models/sale_order.py
```diff
@@ -19,6 +20,17 @@ def _get_invoiced(self):
...
                 })
 
+    def action_cancel(self):
+        # Avoid cancel SO with invoiced lines to don't have inconsistent lines,
+        # if we want reconfirm _check_sale_line_state restrict raises error
+        lines = self.mapped('order_line').filtered(
+            lambda x: x.task_id and x.invoice_status == 'invoiced')
+        if lines:
+            raise ValidationError(_(
+                'You can not cancel because this lines have invoiced tasks:'
```
pedrobaeza:
```suggestion
                'You cannot cancel because these lines have invoiced tasks:'
```

pedrobaeza (commented):
Isn't this something that should be on core?

carlosdauden:
> Isn't this something that should be on core?

Surely yes, but Odoo core does not control anything about it and this module is the one that adds the following restriction:
https://github.com/OCA/sale-workflow/blob/162aa6ccf489ad63adae7224d11470adc6df4cc4/sale_order_invoicing_finished_task/models/project.py#L57-L66

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to v13

PR #1471, opened 2021-02-16
[13.0] sale_product_set: allow to pass custom lines to wizard

simahawk:
Includes some more improvements:

* views/actions: allow to add the set to an order from set view as well
* view: allow to ciew/select/unselect lines
* tests: cleanup test case and unify vars

### DIFF
--- a/sale_product_set/wizard/product_set_add.py
+++ b/sale_product_set/wizard/product_set_add.py
@@ -14,7 +14,9 @@ class ProductSetAdd(models.TransientModel):
         "sale.order",
         "Sale Order",
         required=True,
-        default=lambda self: self.env.context.get("active_id"),
+        default=lambda self: self.env.context.get("active_id")
+        if self.env.context.get("active_model") == "sale.order"
+        else None,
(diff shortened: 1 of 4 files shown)


sale_product_set/wizard/product_set_add.py
```diff
@@ -14,13 +14,23 @@ class ProductSetAdd(models.TransientModel):
         "sale.order",
         "Sale Order",
         required=True,
-        default=lambda self: self.env.context.get("active_id"),
+        default=lambda self: self.env.context.get("active_id")
+        if self.env.context.get("active_model") == "sale.order"
+        else None,
```
mmequignon, line 19: Isn't False the value to set for that?
```suggestion
        else False,
```

mmequignon (approved):
Only code review

TDu (approved):

sale_product_set/wizard/product_set_add.py
simahawk: it's the same

PR #1474, opened 2021-02-18
[13.0] Add sale_order_line_replacement module

mmequignon:
This module adds a two new fields on sale_order_line, in order to keep track of order lines replacements.


simahawk (changes requested):
the fields are not available in any view, there's no action/wizard/anything that allows to control this: either you add something like this or you mention in the description / usage that it's only a basic module and you should implement your own logic.

I'd mention at least that it could be integrate w/ product multi link and sale exception to handle some specific cases...

simahawk:
FTR we'll move this to a local module

PR #1482, opened 2021-02-23
[MIG] product_supplierinfo_for_customer_sale: Migration to 13.0

PabloPaezSheridan:


product_supplierinfo_for_customer_sale/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Product Supplierinfo for Customer Sale",
-    "version": "12.0.1.1.0",
+    "version": "13.0.1.1.0",
```
jjscarafia: small detail, but it should be 13.0.1.0.0 (the migration script didn't correct it automatically?)

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
```diff
@@ -43,19 +43,14 @@ def setUp(self):
...
-                "email": "example@yourcompany.com",
-                "customer": True,
-                "phone": 123456,
-            }
+            {"name": name, "email": "example@yourcompany.com", "phone": 123456}
         )
 
     def _create_partnerinfo(
-        self, supplierinfo_type, partner, product, empty_variant=False
+        self, supplierinfo_type, customer, product, empty_variant=False
```
jjscarafia: why are we renaming this argument?

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
```diff
@@ -43,19 +43,14 @@ def setUp(self):
...
     ):
         vals = {
-            "name": partner.id,
+            "name": self.customer.id,
```
jjscarafia: why not using customer.id? (actually if should be partner.id if the argument is not renamed)

jjscarafia (changes requested):

product_supplierinfo_for_customer_sale/__manifest__.py
PabloPaezSheridan: No, the changes were only suggeted but not maked by the pre-commit feature. Okey understood, then i commit that

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
PabloPaezSheridan: was only for better understanding across the debugging, nothing functional at all, i'll make the changes you suggest

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
PabloPaezSheridan: partner_id is a mandatory many2one field in the sale.order model

jjscarafia (approved):
LGTM

danimv5 (approved):
Functional review. LGTM!

rousseldenis:
@PabloPaezSheridan Could you squash your commits ?

PabloPaezSheridan:
> @PabloPaezSheridan Could you squash your commits ?

This all done

LoisRForgeFlow:
@PabloPaezSheridan I did not see this pr and created a new migration here: https://github.com/OCA/sale-workflow/pull/1667. However it also include an improvement to use a centralized method to get the customer code (check https://github.com/OCA/product-attribute/pull/891).

do you think we can supersede this pr with https://github.com/OCA/sale-workflow/pull/1667?

Thanks!

rafaelbn:
Already merged in https://github.com/OCA/sale-workflow/pull/1667#issuecomment-891609528

PR #1488, opened 2021-02-26
[13.0][FIX] sale_product_multi_add

acsonefho:
If the user doesn't have access rights to create products, he can't select products.

With this `create="1"`, he is able to select products but not creating (depending on access rights)


Cedric-Pigeon (approved):

Zar21 (approved):
LGTM

PR #1494, opened 2021-03-02
Support pricelist with excluded discounts

ddufresne:
Before this commit, discounts defined in pricelists for a products worked
if the price was included in the unit price.

It would not work if the pricelist is parametrized with prices and discounts shown separately.


ddufresne:
@rousseldenis @Cedric-Pigeon would you like to review this PR?

Thanks in advance.

Cedric-Pigeon (approved):
LGTM (code review only)

atchuthan (approved):

PR #1495, opened 2021-03-03
[12.0][REF] sale_triple_discount: subtotal computation

SimoRubi:
Depends on https://github.com/OCA/account-invoicing/pull/876.
Use the new mixin for triple discount computation

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -4,42 +4,12 @@
 # Copyright 2018 ~ 2021 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
-import odoo.addons.decimal_precision as dp
+from odoo import api, models
 
 
 class SaleOrderLine(models.Model):
-    _inherit = "sale.order.line"
-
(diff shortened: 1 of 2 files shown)


SimoRubi:
This module currently erases all the data in `discount2`, `discount3` and `discounting_type` due to the above issue in Odoo

rco-odoo:
> This module currently erases all the data in `discount2`, `discount3` and `discounting_type` due to the above issue in Odoo

I just had a look at the runbot build, and the logfile shows the error
```
TypeError: Model 'sale.order.line' inherits from non-existing model 'line.triple_discount.mixin'.
```
In other words, the code of the mixin model is not present in this branch.

Isn't it what is causing the data loss with this branch?
Make sure you combine both branches in order to test an upgrade.

SimoRubi:
> In other words, the code of the mixin model is not present in this branch.
> 
> Isn't it what is causing the data loss with this branch? Make sure you combine both branches in order to test an upgrade.

Thanks for having a look, the _erased data_ issue can't be reproduced in runbot because it requires to coordinate code changes and module updates, so the runbot's build results are not the problem right now, they will simply fail until the dependency is merged.

I have pointed out the steps to reproduce the _erased data_ issue in a more abstract fashion in https://github.com/odoo/odoo/issues/67168.

PR #1496, opened 2021-03-04
[12.0][FIX] sale_triple_discount: Cache management

SimoRubi:
Fix https://github.com/OCA/sale-workflow/issues/1391.

@ioans73 I had to propose a new PR instead of editing https://github.com/OCA/sale-workflow/pull/1410 because that approach is very different from this one, could you please check if this fixes your issue too?

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -1,7 +1,7 @@
 # Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
 # Copyright 2017 - 2019 Alex Comba - Agile Business Group
 # Copyright 2017 Tecnativa - David Vidal
-# Copyright 2018 Simone Rubino - Agile Business Group
+# Copyright 2018 ~ 2021 Simone Rubino - Agile Business Group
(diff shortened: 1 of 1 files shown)


ioans73:
Yes, the changes seem to be working correctly.
Thank you very much

ioans73 (approved):

SimoneVagile (approved):

rousseldenis (approved):
Code review.

Indeed, this change removes ugly things. But it can makes sense that lines triple discount preprocess should be done one by one.

rousseldenis:
@SimoRubi Is this ready ?

SimoRubi:
Thanks for the review!

> Code review.
> 
> Indeed, this change removes ugly things. But it can makes sense that lines triple discount preprocess should be done one by one.

As far as I have investigated, everything depends on the `super` method: 

- if it is `api.one` (deprecated), then we should pre/post-process lines one-by-one
- otherwise, we can pre/post-process in bulk

So in my opinion it makes sense that pre/post-process methods are able to manage recordsets so we can handle the non-deprecated case without adding everytime a `for` loop.

> @SimoRubi Is this ready ?

yes

PR #1499, opened 2021-03-05
[10.0][ADD] sale_exception: rewrite exception code to reflect name changes in eval context

hbrunn:
earlier versions of this used [order](https://github.com/OCA/sale-workflow/blob/9.0/sale_exception/models/sale.py#L192) and [line](https://github.com/OCA/sale-workflow/blob/9.0/sale_exception/models/sale.py#L201), but in v10 we call it [sale](https://github.com/OCA/sale-workflow/blob/10.0/sale_exception/models/sale.py#L29) and [sale_line](https://github.com/OCA/sale-workflow/blob/10.0/sale_exception/models/sale_order_line.py#L38).

I think it's prudent to not use those when rewriting, but `object` as that exists in all upper versions.

### DIFF
--- a/sale_exception/migrations/10.0.2.0.0/post-migration.py
+++ b/sale_exception/migrations/10.0.2.0.0/post-migration.py
@@ -36,7 +36,7 @@ def migrate(cr, version=None):
         return
     cr.execute(
         'select id, code from exception_rule where model in %s',
-        (tuple('sale.order', 'sale.order.line'),),
+        (('sale.order', 'sale.order.line'),),
     )
     for exception_id, code in cr.fetchall():
         new_code = astunparse.unparse(RewriteNames().visit(ast.parse(code)))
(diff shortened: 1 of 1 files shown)


StefanRijnhart (commented):
`astunparse` method works very well. The "sloppy" method leaves some of the original references in exceptions like https://github.com/OCA/connector-magento/blob/9.0/magentoerpconnect/magentoerpconnect_data.xml#L97:
```
if object.magento_bind_ids and abs(order.amount_total - object.magento_bind_ids[0].total_amount) >= 0.01:
```
Not sure that having this second code path available is very useful if installing the extra library gives you perfect results.

hbrunn:
my argument here is that the sloppy version works for this module's data. My main gripe is that we don't have something like migration_dependencies, and I have the idea it's nice to try at least

StefanRijnhart (approved):

PR #1020, opened 2019-12-17
[13.0][MIG]sale_order_line_input

Khalid-SerpentCS:
This module extends the functionality of sales to allow you to work with
sale lines directly.

It allows you to create sales orders from this sales lines view.

PR #1502, opened 2021-03-08
[13.0] [MIG] sale_order_line_input

Rad0van:
As #1020 seems to be abandoned and missing missing split about pre-commit stuff and the migration itself I have decided to try to push this forward.


pedrobaeza (approved):

sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -46,6 +46,7 @@
                     force_save="1"
                     context="{'partner_id':order_partner_id, 'quantity':product_uom_qty, 'pricelist':pricelist_id, 'uom':product_uom, 'company_id': company_id}"
                 />
+                <field name="product_uom_category_id" invisible="1" />
```
carlosdauden, line 66: This view should have at least the same fields and options as the original v13 view:
https://github.com/odoo/odoo/blob/13.0/addons/sale/views/sale_views.xml#L436-L553
We should compare and add the missing fields and options.
It would be nice if all the fields have the attribute `optional="show/hide"`. (Except key fields like partner or product)

carlosdauden (changes requested):

pedrobaeza:
@Rad0van can you take into account Carlos' comments?

Rad0van:
> @Rad0van can you take into account Carlos' comments?

Sure, I am planning to do that shortly.

Rad0van:
@carlosdauden I did the comparison and the only difference are these fields: `display_type, price_tax, product_template_id and qty_delivered_manual`. All of them have `invisible="1"` So I guess it's not really of any use to add these, is it? I will add the `optional="show"` to the shown fields.

Rad0van:
@carlosdauden is it OK this way? I have added optional="show". If so I will squash commits.

carlosdauden:
Sorry for taking so long to respond, I've been quite busy.

We should add the fields even though they are invisible because they can be taken into account to perform certain calculations and when we are in context "OnChange / NewId" or I don't know how to call it :relaxed: , the data is only accessible if the field is defined on the screen.

If Odoo adds them and makes them invisible, it will be because they are needed for something, that is why he commented that the ideal is that in the view that this module creates we always have, at least, the same fields that are defined in the original view.

pedrobaeza:
@Rad0van any news about this?

Rad0van:
> @Rad0van any news about this?

Will finish it tomorrow.

pedrobaeza:
Have you pushed anything? I'm not able to find the reported commit.

Rad0van:
> Have you pushed anything? I'm not able to find the reported commit.

I have. I have also squashed the commits together that's why probably...

pedrobaeza:
OK, GitHub is not showing such changes right now, but I see the optional part.

@carlosdauden is it OK for you now?

carlosdauden (approved):

PR #1506, opened 2021-03-10
[14.0][MIG] sale_order_type: Migration to 14.0

ferran-S73:
Standard migration to version 14.0


rousseldenis:
@ferran-73 Could you check travis ?

.coveragerc
```diff
@@ -0,0 +1,10 @@
+[run]
```
rousseldenis, line 1: Why are you modifying this ???

rousseldenis:
@ferran-73 It seems that you didn't look into https://github.com/OCA/sale-workflow/issues/1274

That module is already done there. Could you review it ?

ferran-S73:
> @ferran-73 It seems that you didn't look into #1274
> 
> That module is already done there. Could you review it ?

You're right, didn't find it. I'll close this pull request.

PR #1510, opened 2021-03-12
[14.0][ADD] sale_pricelist_from_commitment_date

jcoux:
When the sale order commitment date is set,
this date is used by pricelist to compute price unit instead of using order date.

### DIFF
--- a/sale_pricelist_from_commitment_date/models/product_pricelist.py
+++ b/sale_pricelist_from_commitment_date/models/product_pricelist.py
@@ -1,0 +1,43 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductPricelist(models.Model):
+
+    _inherit = "product.pricelist"
+
+    def _compute_price_rule_multi(self, products_qty_partner, date=False, uom_id=False):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super()._compute_price_rule_multi(products_qty_partner, date, uom_id)
+
+    def _compute_price_rule(self, products_qty_partner, date=False, uom_id=False):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super()._compute_price_rule(products_qty_partner, date, uom_id)
+
+    def get_products_price(
+        self, products, quantities, partners, date=False, uom_id=False
+    ):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super().get_products_price(products, quantities, partners, date, uom_id)
+
+    def get_product_price(self, product, quantity, partner, date=False, uom_id=False):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super().get_product_price(product, quantity, partner, date, uom_id)
+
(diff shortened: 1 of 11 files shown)


sale_pricelist_from_commitment_date/models/product_pricelist.py
```diff
@@ -0,0 +1,20 @@
...
+    def get_products_price(
+        self, products, quantities, partners, date=False, uom_id=False
+    ):
+        new_date = date
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            new_date = force_pricelist_date
+        return super().get_products_price(
+            products, quantities, partners, new_date, uom_id
+        )
```
damdam-s:
```suggestion
        force_pricelist_date = self.env.context.get("force_pricelist_date")
        if force_pricelist_date:
            date = force_pricelist_date
        return super().get_products_price(
            products, quantities, partners, date, uom_id
        )
```

damdam-s (approved):
just a suggestion but LGTM

sale_pricelist_from_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+            sale._apply_pricelist_from_commitment_date()
+        return sale
+
+    def write(self, vals):
+        if "commitment_date" not in vals and "pricelist_id" not in vals:
+            return super().write(vals)
+        for sale in self:
+            old_commitment_date = sale.commitment_date
+            old_pricelist = sale.pricelist_id
+            super(SaleOrder, sale).write(vals)
```
rousseldenis, line 37: Is this wanted as changing a multi recordset flow into single one?..

jcoux: Yes, it's to check for each sale if the commitment_date or the pricelist_id is really changed and not just "rewrite" with the same value.
Because in case of just passing the date in values without changing the date (or pricelist), I don't want to recompute the price of all lines.

sale_pricelist_from_commitment_date/models/product_pricelist.py
jcoux: Fixed!

sale_pricelist_from_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    def _apply_pricelist_from_commitment_date(self):
+        self.ensure_one()
+        for line in self.order_line:
+            # Price unit is still modifiable if not quantity invoiced
+            if not line.qty_invoiced:
+                # Call product_uom_change as it only update price_unit using pricelist
```
gurneyalex:
```suggestion
                # Call product_uom_change as it only updates price_unit using pricelist
```

gurneyalex (approved):

gurneyalex (commented):
small question: commitment_date is not required.  How does the module behave when there is no commitment date on the SO? Do we have meaningful error message?

jcoux:
> small question: commitment_date is not required. How does the module behave when there is no commitment date on the SO? Do we have meaningful error message?

If we don't have `commitment_date`, the module uses the core behavior, then the order date is used as date.

danimv5 (approved):
Functional review!! LGTM

brendapaniagua:
Functional review! LGTM

jcoux:
Added 2 new fixup to cover missing cases.

Waiting validation of mergeable before squash.

damdam-s (approved):

rousseldenis:
> Added 2 new fixup to cover missing cases.
> 
> Waiting validation of mergeable before squash.

@jcoux

jcoux:
@rousseldenis squashed! Ready to merge :)

PR #1397, opened 2020-12-27
[MIG] sale_automatic_workflow: Migration to 14.0

phuctranfxvn:

PR #1514, opened 2021-03-16
[FIX] sale_automatic_workflow

max3903:
Depends on #1397

### DIFF
--- a/sale_automatic_workflow/models/sale_order.py
+++ b/sale_automatic_workflow/models/sale_order.py
@@ -22,6 +22,16 @@ class SaleOrder(models.Model):
         store=True,
     )
 
+    @api.model
+    def default_get(self, fields):
+        res = super(SaleOrder, self).default_get(fields)
+        default_workflow = self.env["sale.workflow.process"].search(
+            [("default", "=", True)], limit=1
+        )
+        if default_workflow:
+            res["workflow_process_id"] = default_workflow.id
+        return res
+
     # TODO: v15 -> make this module dependent on sale_delivery_state and
     # use the code in sale_automatic_workflow_delivery_state to replace this function
     @api.depends("order_line.qty_delivered", "order_line.product_uom_qty")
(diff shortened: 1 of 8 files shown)


hailangvn:
Good day @max3903 . As https://github.com/OCA/sale-workflow/pull/1397 has been merged. Could you please help to rebase this PR on latest 14.0 branch? That will ease the review task.

atchuthan:
@hailangvn rebase done

PR contains additional functionalities to sale_automatic_workflow
1. Add default option and use this automatically on SO creation.
2. Number of records to process(Currently changed the search limit to 20)
3. Commit after each record is processed via savepoint, this is to facilitate if customer confirms order at the same time the scheduler is taking care of it.  If the record is committed after each process, then we can avoid this scenario.
4. Propagate the Workflow selected in Partner to SO

cc: @SodexisTeam

sale_automatic_workflow/models/sale_order.py
```diff
@@ -21,6 +21,16 @@ class SaleOrder(models.Model):
         store=True,
     )
 
+    @api.model
+    def default_get(self, fields):
+        res = super(SaleOrder, self).default_get(fields)
+        default_workflow = self.env["sale.workflow.process"].search(
```
hailangvn: I think we need to match field name to avoid extra processing for fields other than `workflow_process_id`.

hailangvn (changes requested):
Thanks for your contribution. Please help to check the suggestion. Thanks.

sale_automatic_workflow/models/sale_order.py
```diff
@@ -21,6 +21,16 @@ class SaleOrder(models.Model):
...
+            [("default", "=", True)], limit=1
+        )
+        if default_workflow:
+            res["workflow_process_id"] = default_workflow.id
```
grindtildeath: You need to check if `workflow_process_id` is in the fields list to get a default value. Otherwise, this would overwrite a specific value that could have been set explicitely in the values dict passed to create function.
```suggestion
        if "workflow_process_id" in fields:
            default_workflow = self.env["sale.workflow.process"].search(
                [("default", "=", True)], limit=1
            )
            if default_workflow:
                res["workflow_process_id"] = default_workflow.id
```

grindtildeath (changes requested):

PR #1507, opened 2021-03-10
[12.0][IMP]sale_order_type: update quotation name when user change or…

chandni299:
[12.0][IMP]sale_order_type: update quotation(having "draft" or "sent" as state only) name when user change order type

PR #1517, opened 2021-03-19
[12.0][FIX] sale_order_type, singleton error on write

GSLabIt:


GSLabIt:
@rousseldenis, @HaraldPanten quick review?

pedrobaeza:
@dreispt @chandni299 I think this has been introduced with your latest merge. Please watch this kind of things that compromises the stability of the module in such older branches.

The weird thing is that the PR is not with that bug:

https://github.com/OCA/sale-workflow/pull/1507/files#diff-d8265b9c3f7f83166d7d08725ca19b785b444aeb8f82a39d2b31138956531878R66

so maybe @dreispt type the command before latest change.

@GSLabIt thanks for the PR. Some comments later, for doing a correct and optimal implementation of this.

sale_order_type/models/sale_order.py
```diff
@@ -62,8 +62,8 @@ def create(self, vals):
     @api.multi
     def write(self, vals):
         res = super(SaleOrder, self).write(vals)
-        if vals.get("type_id") and (self.state == "draft" or self.state == "sent"):
-            for order in self:
+        for order in self:
+            if vals.get("type_id") and (order.state in ["draft", "sent"]):
```
pedrobaeza: Please evaluate this and if there's a sequence before calling super, and also checking if the record was previously with the same type, and modify dictionary for calling super for only write once. A quick code snippet would be:
```python
if vals.get("type_id"):
    if sale_type.sequence_id:
        for record in self:
            if record.state in {"draft", "sent"} and record.type_id != sale_type:
                new_vals = vals.copy()
                new_vals["name"] = sale_type.sequence_id.next_by_id()
                super(SaleOrder, record).write(new_vals)
            else:
                super(SaleOrder, record).write(vals)
```

sale_order_type/readme/CONTRIBUTORS.rst
```diff
@@ -29,4 +29,8 @@
   * Angel Moya Pardo <angel.moya@pesol.es>
   * Antonio J Rubio Lorente <antonio.rubio@pesol.es>
 
+* `GSLab.it <https://www.gslab.it>`_
+  * Giovanni Serra <giovanni@gslab.it>
```
pedrobaeza, line 34: Empty line needed for proper syntax

pedrobaeza (changes requested):

sale_order_type/models/sale_order.py
GSLabIt: @pedrobaeza thanks for the hint. just a question `{"draft", "sent"}` is correct? i'd expect `["draft", "sent"]`

pedrobaeza: Yes, it's correct. It's the Python definition of a set, that improves better with `in` operator, although the other works as well

GSLabIt: k, thanks!

sale_order_type/models/sale_order.py
```diff
@@ -53,21 +53,27 @@ def onchange_type_id(self):
...
+                for record in self:
+                    if (
+                        record.state in {"draft", "sent"}
+                        and record.type_id != sale_type
+                    ):
+                        new_vals = vals.copy()
+                        new_vals["name"] = sale_type.sequence_id.next_by_id()
+                        super(SaleOrder, record).write(new_vals)
+                    else:
+                        super(SaleOrder, record).write(vals)
```
pedrobaeza, line 76: Missing else part calling super as is. Also missing `return True for this part.

GSLabIt: Aarrrgghhh sorry. `return True` should be to all function, isn't it?

pedrobaeza: Well, it can be an option, the other is:
```python
if ...:
    return True
else:
   return super().write(vals)
```

GSLabIt: done, thx again

sale_order_type/models/sale_order.py
```diff
@@ -53,21 +53,30 @@ def onchange_type_id(self):
...
+            return True
+        else:
+            return super(SaleOrder, record).write(vals)
```
pedrobaeza: Sorry again, but I'm detecting a problem, and it's if there's a new type, but no sequence: no write will be done. Proper code is:
```suggestion
        if vals.get("type_id"):
            sale_type = self.env["sale.order.type"].browse(vals["type_id"])
            if sale_type.sequence_id:
                for record in self:
                    if (
                        record.state in {"draft", "sent"}
                        and record.type_id != sale_type
                    ):
                        new_vals = vals.copy()
                        new_vals["name"] = sale_type.sequence_id.next_by_id()
                        super(SaleOrder, record).write(new_vals)
                    else:
                        super(SaleOrder, record).write(vals)
                return True
        return super(SaleOrder, record).write(vals)
```

GSLabIt: got it ..

pedrobaeza (approved):

sale_order_type/models/sale_order.py
```diff
@@ -53,21 +53,30 @@ def onchange_type_id(self):
...
+                        and record.type_id != sale_type
+                    ):
+                        new_vals = vals.copy()
+                        new_vals["name"] = sale_type.sequence_id.next_by_id()
+                        super(SaleOrder, record).write(new_vals)
+                    else:
+                        super(SaleOrder, record).write(vals)
+            return True
+
+        return super(SaleOrder, record).write(vals)
```
pedrobaeza:
```suggestion
        return super().write(vals)
```

GSLabIt: Arg ..sorry ..my fault

PR #1520, opened 2021-03-22
[ADD] sale_product_classification: New module

chienandalu:
Classify products according to their sales and seasonality.

cc @Tecnativa TT28816


legalsylvain:
Hi @chienandalu. your module seems to implement A B C classification. Do you know the alternative implementation available here : https://github.com/OCA/product-attribute/pull/623

In a first sight, I think I prefer percentage implementation than absolute sale amount threshold. Don't you think ?

CC : @MiquelRForgeFlow

Thanks.

chienandalu:
> Hi @chienandalu. your module seems to implement A B C classification. Do you know the alternative implementation available here : [OCA/product-attribute#623](https://github.com/OCA/product-attribute/pull/623)

Thanks @legalsylvain I didn't know the module. I put a note in the roadmap to try to merge this functionality into that module when we migrate to v13 if it's already in the branch. I also saw that it already diverged from the original PR in another v10 backport... as we are a little bit tight on budget here, we'll go with this approach for now and join when the module is settled down.

> In a first sight, I think I prefer percentage implementation than absolute sale amount threshold. Don't you think ?

It was the first option. Anyway, it should not be very hard to extend the module to support that behavior

legalsylvain:
Hi @chienandalu. The roadmap section is great. thanks ! Makes senses to move forward with this version and then see how to converge. As you worked on that ABC features, if you have a bit of time, could you review the other PR to give your PoV regarding the alternative design ? 
regards.

sale_product_classification/tests/test_sale_product_classification.py
```diff
@@ -0,0 +1,64 @@
...
+        product_classification.update({
+            self.prod_1: "d",
+            self.prod_4: "c",
+            self.prod_3: "c",
+        })
+        self._test_product_classification(product_classification)
+        # Product 1 gets an A!
+        self._create_sale("2021-04-01", self.partner, self.prod_1, 20000)
+        cron_classify()
+        product_classification.update({self.prod_1: "a"})
```
victoralmau, line 64:
```suggestion
        product_classification.update({self.prod_1: "a"})
        self._test_product_classification(product_classification)
```

victoralmau (changes requested):

sale_product_classification/tests/test_sale_product_classification.py
chienandalu: Oh, yeah :+1:

CarlosRoca13 (approved):

victoralmau (approved):

PR #1524, opened 2021-03-25
[WIP][ADD] partner_default_route: Add default route to partner

MateuGForgeFlow:
Add default route to partner.
When generating a Sale Order (for instance) if a route is not selected and the partner has a default route set, the system will use it.


sale_partner_default_route/models/sale_order_line.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @api.multi
+    def _prepare_procurement_values(self, group_id=False):
+        """ Prepare specific key for moves or other components that will be created from
+        a stock rule comming from a sale order line. This method could be override in order
+        to add other custom key that could be used in move/po creation.
+        """
+        values = super(SaleOrderLine, self)._prepare_procurement_values(group_id)
+        self.ensure_one()
+        date_planned = self.order_id.confirmation_date \
```
rousseldenis, line 19: Why this part of code here (that update 'values')? IMHO, it should be limited only to default route behaviour.

sale_partner_default_route/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2021 ForgeFlow S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
```
rousseldenis, line 2: Can you sync all the files with the license of the module (mixed AGPL/LGPL) ?

rousseldenis:
@MateuGForgeFlow Could you attend comments?

MateuGForgeFlow:
Hi @rousseldenis, we have this on hold. We need to decide what to do with it. Thank you for the review!

PR #1531, opened 2021-03-26
[FIX] sale_order_type: Non reset the value when it's sitting by the user in the view.

nicomacr:
In this way we reset so that if I choose a value in the view this value is not stepped on by the compute method, before the change it always stepped on regardless that there are any value to replace about on the outcome of the logic.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -19,10 +19,12 @@ class AccountMove(models.Model):
 
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
-        self.sale_type_id = self.env["sale.order.type"]
-        for record in self.filtered(
-            lambda am: am.type in ["out_invoice", "out_refund"]
-        ):
+        for record in self:
+            if record.type not in ["out_invoice", "out_refund"]:
+                record.sale_type_id = self.env["sale.order.type"]
+                continue
+            else:
+                record.sale_type_id = record.sale_type_id
             if not record.partner_id:
                 record.sale_type_id = self.env["sale.order.type"].search(
                     [("company_id", "in", [self.env.company.id, False])], limit=1


bodedra (approved):
LG :+1:

jjscarafia (approved):

sale_order_type/models/account_move.py
```diff
@@ -19,7 +19,6 @@ class AccountMove(models.Model):
 
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
-        self.sale_type_id = self.env["sale.order.type"]
```
rousseldenis, line 22: Ok, for this.

Apart of that, I think the method does not compute the value for the whole recordset when type not out_invoice or out_refund (should be initialized with False)

nicomacr: Yes good point, but the field is computed and stored=True, for all value outside of the filtered is already taken the value in False.

rousseldenis (approved):

sale_order_type/models/account_move.py
pedrobaeza, line 22:
```suggestion
self.sale_type_id = self.sale_type_id or self.env["sale.order.type"]
```

pedrobaeza (changes requested):
There's no default value now in v13, so with this, you will get a CacheMiss record sooner or later. That's why it was put such line. Can you try this code if it works?

sale_order_type/models/account_move.py
pedrobaeza: Sorry, it should be record by record for avoiding singleton error, but the idea is this one.

nicomacr: @pedrobaeza ready this change.

pedrobaeza (approved):

PR #1534, opened 2021-03-30
[13.0][FIX]sale_procurement_group_by_line: fix fill previous_product_…

JordiMForgeFlow:
This fix prevents having duplicated procurements when having several order lines involving any operation that has a specific `<sale.order.line>._get_qty_procurement` relying on `previous_product_uom_qty` (Manufacturing kits, dropshipping..)

A fix was implemented in https://github.com/OCA/sale-workflow/commit/c9b483ce3f08f053f806d8ab14c225618121a3b3 but it was only filling the previous_product_uom_qty for the last line.


LoisRForgeFlow (approved):
Thanks for the fix! LGTM

rousseldenis (approved):

PR #1533, opened 2021-03-29
[13.0][MIG] sale_procurement_amendment

rousseldenis:
Depends on :

- [x] #1564

PR #1535, opened 2021-03-30
[13.0][ADD] sale_route_amendment

rousseldenis:
Depends on :
- [ ] #1533


sale_route_amendment/wizards/sale_order_line_route_amend.py
```diff
@@ -0,0 +1,48 @@
...
+            res["warning"] = _(
+                "Route updating is not possible as some logistics "
+                "actions have been processed or partially processed "
+                "for all lines."
+            )
+        return res
+
+    def update_route(self):
+        for wizard in self:
+            wizard.order_line_ids._amend_and_reprocure()
```
Cedric-Pigeon, line 54: it miss an essential piece where you have to set the route on the sale order line before triggering any stock rule.
Otherwise it does not work.

rousseldenis: > it miss an essential piece where you have to set the route on the sale order line before triggering any stock rule.
> Otherwise it does not work.

:sweat_smile:

rousseldenis:
@Cedric-Pigeon Could you update your review and pull some people to review in order to merge this ?

PR #1536, opened 2021-03-30
[12.0][IMP] add barcode to product set and report

sergiocorato:


legalsylvain (commented):
hi @sergiocorato. Thanks for adding that. 
At the first sight, I wonder new kind of nomenclature is not added in the same time. Otherwise, I dont think it will be possible to use this new field in Odoo.
Or did I missed something.

sergiocorato:
Hi @legalsylvain  and thanks for the review.
I'll use this field in another module, but I added some logic for search and a constraint (if I got the point).

legalsylvain:
Hi @sergiocorato. Thanks for your reply. 
naive question : don't you think it should be great to add a new nomenclature.rule like here https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/data/default_barcode_patterns.xml#L21. 
with a new type, ``product_set`` ?

sergiocorato:
@legalsylvain I wasn't aware of this object, thanks for the info!
I've updated the PR.

PR #1537, opened 2021-03-31
[13.0][FIX] sale_coupon_multi_use: Replace sudo by with_user

jcoux:
Replace sudo by with_user.

### DIFF
--- a/sale_coupon_multi_use/models/sale_order_line.py
+++ b/sale_coupon_multi_use/models/sale_order_line.py
@@ -1,7 +1,7 @@
 # Copyright 2021 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import SUPERUSER_ID, fields, models
 
 
 class SaleOrderLine(models.Model):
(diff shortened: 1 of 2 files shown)


sale_coupon_multi_use/models/sale_order_line.py
```diff
@@ -19,6 +19,6 @@ def unlink(self):
         consumption_lines = self.mapped("coupon_consumption_line_ids")
         if consumption_lines:
             # Remove related consumption lines
-            # Using sudo because no user can remove consumption lines
-            consumption_lines.sudo().unlink()
+            # Using super user because no user can remove consumption lines
+            consumption_lines.with_user(SUPERUSER_ID).unlink()
```
rousseldenis, line 23: Actually, sudo() still exists and apply automatically SUPERUSER_ID. But with_user() allows to 'do the operations with high privileges keeping the user details - logging for instance'.

sudo : https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5150

with_user: https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5161


So, you can let sudo() here :smiley:

jcoux: Now it's done: it will be more easy to migrate in future version ;)

ValentinVinagre (approved):
LGTM 👍

jcoux:
Seems this PR is useless: close it.

PR #1538, opened 2021-03-31
[13.0][ADD] sale_coupon_most_expensive_delivery

jcoux:
Bridge for Most Expensive program and shipping costs.

### DIFF
--- a/sale_coupon_most_expensive_delivery/models/sale_order.py
+++ b/sale_coupon_most_expensive_delivery/models/sale_order.py
@@ -1,0 +1,21 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+from odoo.tools import float_compare
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    def _get_most_expensive_line(self):
+        # Copied from sale_coupon_most_expensive module to ignore deliveries
+        price_reduce_digits = self.env["decimal.precision"].precision_get(
+            self.order_line._fields["price_reduce"]._digits
+        )
+        lines = self.order_line.filtered(
+            lambda r: not (r.is_reward_line or r.is_delivery)
+            and float_compare(r.price_reduce, 0, precision_digits=price_reduce_digits)
+            > 0
+        )
+        return max(lines, key=lambda r: r["price_reduce"])
--- a/sale_coupon_most_expensive_delivery/__init__.py
+++ b/sale_coupon_most_expensive_delivery/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/sale_coupon_most_expensive_delivery/__manifest__.py
+++ b/sale_coupon_most_expensive_delivery/__manifest__.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale Coupon Most Expensive Delivery",
+    "summary": "Bridge for Most Expensive program and shipping costs",
+    "version": "13.0.1.0.0",
+    "category": "Sale",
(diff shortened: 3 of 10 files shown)


damdam-s (approved):

sale_coupon_most_expensive_delivery/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+from odoo.tools import float_compare
+
+
+class SaleOrder(models.Model):
+
```
ivantodorovich, line 8:
```suggestion
```
not blocking

sale_coupon_most_expensive_delivery/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+        # Copied from sale_coupon_most_expensive module to ignore deliveries
+        price_reduce_digits = self.env["decimal.precision"].precision_get(
+            self.order_line._fields["price_reduce"]._digits
+        )
+        lines = self.order_line.filtered(
+            lambda r: not (r.is_reward_line or r.is_delivery)
+            and float_compare(r.price_reduce, 0, precision_digits=price_reduce_digits)
+            > 0
+        )
+        return max(lines, key=lambda r: r["price_reduce"])
```
ivantodorovich, line 21:
```suggestion
        return max(lines.mapped("price_reduce")) if lines else 0.0
```

on empty recordsets max would raise an error, also mapped is more readable I think :)

ivantodorovich, line 21:
```suggestion
    def _get_most_expensive_line(self):
        lines = self.mapped("order_line").filtered(lambda r: not (r.is_reward_line or r.is_delivery))
        return max(lines.mapped("price_reduce")) if lines else 0.00
```

isn't this virtually the same though?
not sure why float_compare is required

(not blocking though)

ivantodorovich (changes requested):

sale_coupon_most_expensive_delivery/models/sale_order.py
damdam-s: with your solution, you don't exclude the lines with price_reduce <= 0 so I think @jcoux code is ok (except for the max on empty recorset)

damdam-s: :+1:

ivantodorovich: yeah but then again whats the point :)

jcoux:
@damdam-s @ivantodorovich This code and similar of the core code for the cheapest product:
https://github.com/odoo/odoo/blob/13.0/addons/sale_coupon/models/sale_order.py#L134-L136

Just used a float compare to do the float comparison.

I prefer to have a code similar of the code core to have the same behavior (expect than we take the most expensive product instead of cheapest).

What's your opinion about this?

damdam-s:
OK for me

ivantodorovich:
> @damdam-s @ivantodorovich This code and similar of the core code for the cheapest product:
> https://github.com/odoo/odoo/blob/13.0/addons/sale_coupon/models/sale_order.py#L134-L136
> 
> Just used a float compare to do the float comparison.
> 
> I prefer to have a code similar of the code core to have the same behavior (expect than we take the most expensive product instead of cheapest).
> 
> What's your opinion about this?

not blocking :)

ivantodorovich (approved):

epanisset:
### Functionnal test done and validated. 

_Tested with 2 products (12$ and 40$) + shipping cost of 500$ 
Promotion giving 10% on the most expensive product
Result: 10% on the product at 40$ 
=> OK_

epanisset (approved):
### Functionnal test done and validated.

_Tested with 2 products (12$ and 40$) + shipping cost of 500$
Promotion giving 10% on the most expensive product
Result: 10% on the product at 40$
=> OK_

PR #1551, opened 2021-04-12
[13.0][FIX] sale_exception ignore after confirm

TDu:
After the sale order has been confirmed, some fields can still be
changed, those changes can raise some more excetptions again.
Unfortunatelly those changes are raised as ValidationError and they are
blocking.

If during the confirmation of the sale.order the ignore excetpiton has
been used. Those same ignored exception will block some changes made
when the sale order is in `sale` state.

This change ensure that if exceptions have been ignored they will not be
locking in the `sale` state.

### DIFF
--- a/sale_exception/models/sale.py
+++ b/sale_exception/models/sale.py
@@ -66,11 +66,6 @@ class SaleOrder(models.Model):
         if orders:
             orders._check_exception()
 
-    @api.onchange("order_line")
-    def onchange_ignore_exception(self):
-        if self.state == "sale":
-            self.ignore_exception = False
-
     def action_confirm(self):
         if self.detect_exceptions():
             return self._popup_exceptions()


rousseldenis (approved):
Makes sense.

@TDu Maybe a litlle test for that ?

TDu:
Looking at building a test for it, but I don't have the same behaviour than in the UI...
Which is weird, it is like this is not needed !?

HaraldPanten (approved):

sebalix (approved):

mmequignon (approved):

TDu:
I totally change the implementation of this PR, it needs to be reviewed again

nicomacr (approved):
Excellent i found the same problems with confirmed order with ignored exceptions

rousseldenis:
@TDu Could you confirm that this PR concerns only the deletion of the onchange?

TDu:
@rousseldenis The first implementation which you reviewed was more complex. But finally only the removal of that onchange is required. So yes.

rousseldenis (approved):
Makes sense.

As order lines can be added after sale confirmation, they can produce errors. Maybe something should be done to warn in case of exception, no ?

rousseldenis:
> Makes sense.
> 
> As order lines can be added after sale confirmation, they can produce errors. Maybe something should be done to warn in case of exception, no ?

@TDu

PR #1552, opened 2021-04-12
[FIX] sale_order_type: changes ignored

chienandalu:
f4e61607fae6f735452b735a4cd7f45c5914f71e added some changes that made
imposible to reach a proper write if the sale type didn't have the
proper conditions.

cc @Tecnativa TT29148

### DIFF
--- a/sale_order_type/models/sale_order.py
+++ b/sale_order_type/models/sale_order.py
@@ -61,21 +61,23 @@ class SaleOrder(models.Model):
 
     @api.multi
     def write(self, vals):
+        """A sale type could have a different order sequence, so we could
+        need to change it accordingly"""
         if vals.get("type_id"):
             sale_type = self.env["sale.order.type"].browse(vals["type_id"])
             if sale_type.sequence_id:
                 for record in self:
                     if (
                         record.state in {"draft", "sent"}
-                        and record.type_id != sale_type
+                        and record.type_id.sequence_id != sale_type.sequence_id
                     ):
                         new_vals = vals.copy()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
For reducing the diff, why not changing the sequence comparison and remove the `return True` part?

chienandalu:
> For reducing the diff, why not changing the sequence comparison and remove the `return True` part?

Ok, I wanted to simplify it to ease reading. Changed

sale_order_type/models/sale_order.py
```diff
@@ -61,21 +61,24 @@ def create(self, vals):
...
     def write(self, vals):
+        """A sale type could have a different order sequence, so we could
+        need to change it accordingly"""
         if vals.get("type_id"):
             sale_type = self.env["sale.order.type"].browse(vals["type_id"])
             if sale_type.sequence_id:
                 for record in self:
                     if (
                         record.state in {"draft", "sent"}
                         and record.type_id != sale_type
```
pedrobaeza: You can remove this condition.

chienandalu: :+1:  done

pedrobaeza (approved):

pedrobaeza:
As the bug is important, let's fast-track it:

/ocabot merge patch

pedrobaeza:
As the feature has been introduced, we should fw-port it to recent versions, or someone will miss it.

HaraldPanten (approved):

sergio-teruel:
How does this affect in v13?

pedrobaeza:
It's already and totally ported in #1553

PR #1555, opened 2021-04-13
[14.0] [FIX]  sale_order_lot_generator: allow to create lines in redrafted orders

PierrickBrun:


sale_order_lot_generator/models/sale_order.py
```diff
@@ -6,9 +6,23 @@
...
+        # when a new line is added to confirmed sale order
+        # get the max index_lot from the other lines
+        index_lot = 1
+        lot_ids = self.order_line.filtered(lambda l: l.lot_id).mapped("lot_id")
+        for lot in lot_ids:
+            lot_name = lot.name
+            index_str = lot_name.replace(self.name + "-", "")
+            last_index = int(index_str) if index_str.isdigit() else 0
+            index_lot = max(index_lot, last_index)
+            index_lot += 1
```
sebastienbeau: @PierrickBrun
I think there is something wrong here.
If I have a SO with two line with the lot FOO-01 and FOO-02 and the sequence of the line have changed

In the loop
   - I will first process the lot FOO-02 => index_str = 02 => index_lot = 3
   -  second time FOO-01 => index_str = 01 => index_lot = 4
   - 
I think we should init index_lot to 0 and then always add 1 after your loop, what do you think?

sebastienbeau: can you add a test with this scenario ?

PierrickBrun: I think this scenario is already tested but yes there is something weird in the code, I'll fix it

sebastienbeau (approved):
LGTM (code review)

florian-dacosta (approved):

PR #1557, opened 2021-04-14
[ADD] Sale Order Weight: New module to 12.0

manuelcalerosolis:
This module is used to calculate the weight in a sale order.

image


sale_order_weight/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+            total_weight = 0.0
+            for line in order.order_line:
+                total_weight += line.weight
+            order.total_weight = total_weight
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    weight = fields.Float(compute='_compute_weight', string='Weight', store=True)
```
legalsylvain: Hi. This field should not be a computed field, but a field populated by an ochange in my opinion. 
If you change the weight of a product, it should not change the weight of the older orders. (same mechanism as for the price).

don't you think ?

legalsylvain:
Hi. Thanks for the changes. 

Some ideas, to make the module more generic : 

- I think you could replace the ``weight`` field by a ``unit_weight`` field (in the same spirit as ``unit_price``), on the ``sale.order.line``, and so remove the ``product_uom_qty`` from the ``onchange``.
- Once done ``total_weight`` renamed to ``total_ordered_weight``, computed depends on ``order_line.unit_weight`` and ``order_line.product_uom_qty``
- Then, merge features present in the ``sale_order_delivery_weight``, adding in the current module (``sale_order_weight``) the field ``total_delivery_weight``, computed, depends on ``order_line.unit_weight`` and ``order_line.qty_delivered``

Finally one module, more complete, instead of two, and a homogeneous design to compute each ``total_xx_weight``.

What do you think

manuelcalerosolis:
> * unit_weight

I like your vision of the solution, I'm going to get on with it.
Thank you for your interest in this module.

legalsylvain (approved):
Thanks !

sale_order_weight/models/sale_order.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2021 Manuel Calero Solís (https://xtendoo.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    total_ordered_weight = fields.Float(
```
rousseldenis, line 10: You should use digits precision for every weight fields:

https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L88

sale_order_weight/models/sale_order.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @api.depends('order_line.total_delivered_weight')
+    def _compute_total_delivered_weight(self):
+        for order in self:
+            order.total_delivered_weight = sum(
+                order.mapped('order_line.total_delivered_weight')
+            )
+
+
+class SaleOrderLine(models.Model):
```
rousseldenis: Could you please put one model per file ?

rousseldenis (changes requested):
LGTM apart the digits precision

sale_order_weight/models/sale_order.py
manuelcalerosolis: done!

sale_order_weight/models/sale_order.py
manuelcalerosolis: done!

legalsylvain:
@rousseldenis : could you update your review ? (and merge if all is ok)

rousseldenis (approved):

HaraldPanten (approved):

PR #1564, opened 2021-04-21
[13.0][ADD] sale_order_line_chained_move

rousseldenis:
A technical module that adds sale order line related moves as standard ones are for the transfer > Customers

Improvement:
- [x] Add pre_init_hook for existing moves


sale_order_line_chained_move/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    related_sale_line_id = fields.Many2one(comodel_name="sale.order.line", index=True,)
```
legalsylvain, line 11: Hi @rousseldenis. the index will generate extra time when writing on ``stock.move``. (if I understand correctly indexes ...) Do you think it's necessary ?

rousseldenis: Do you have metrics to compare gain with/without indexes on such records that cannot be computed without recursive function otherwise?

The index value is inserted on the same query as normal flow (see https://github.com/OCA/sale-workflow/pull/1564/files#diff-35bc9f8bfb82e6c1af01cbf043eb5fc0b52a3397bb2f083ee0fabafc9664adb1R31). So, I think this addition should not influence write performances.

legalsylvain: > Do you have metrics to compare gain with/without indexes

No I don't. I just know that it is not allways a good idea to create indexes on fields, because it can make insertion / update slower. And so it should be reserved for very searched fields. (name, company_id, etc...)
But maybe I'm wrong, or it is totally obsolete on recent Odoo version !

rousseldenis:
@legalsylvain Could you update you review ?

rousseldenis:
@gurneyalex Could you do review here also as you did for v14 in order to merge these ?

sale_order_line_chained_move/hooks.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import SUPERUSER_ID, api
+
+
+def __fill_related(moves, line_id):
+    for move in moves:
```
acsonefho, line 15: This will trigger `X` sql update.
IMO the best way is to find and group every moves and then assign the related `line_id`.

Something like:
```
def _get_children_moves(moves):
    all_moves = moves
    if moves.move_orig_ids:
        all_moves |= _get_children_moves(moves.move_orig_ids)
    return all_moves
```

Then you can do
`moves.write({"related_sale_line_id": line_id.id})`

acsonefho (changes requested):
Just minor improvement on the `hook`.

sale_order_line_chained_move/hooks.py
rousseldenis: @acsonefho Done, with coverage improvement :-)

acsonefho (approved):
Code LGTM

PR #1565, opened 2021-04-21
[14.0][ADD] sale_product_rating_verified

david-popa:
Verify if a user has previously bought a product


sale_product_rating_verified/models/rating.py
```diff
@@ -0,0 +1,42 @@
...
+
+        for record in self:
+            record.purchase_verified = False
+            purchased_product = (
+                partner_product_count.get((record.partner_id.id, record.res_id), 0) > 0
+            )
+            if record.res_model == "product.product" and purchased_product:
+                record.purchase_verified = True
+
+    purchase_verified = fields.Boolean(
```
rousseldenis: Could you put fields definition on top ?

sale_product_rating_verified/models/rating.py
```diff
@@ -0,0 +1,42 @@
...
+        required=True, default=False, compute="_compute_verified", store=True
```
rousseldenis: I would have done _compute_purchase_verified()

rousseldenis (changes requested):
LGTM. Could you maybe add a USAGE fragment with a screenshot ?

rousseldenis (approved):

gurneyalex (approved):
looks a bit strange that the "verified" badge does not change when the invoice gets paid, or that the date of the invoice is before the advice. This could be mentionned in the readme, I think.

PR #1566, opened 2021-04-21
[13.0][MIG+REF] sale_manual_delivery

ivantodorovich:
Superseeds https://github.com/OCA/sale-workflow/pull/1319

I refactored the module to simplify it by making use of new methods introduced by odoo since it was originally developped in 10.0.
This should ease maintaining it, and more importantly avoid fully overwritting core methods.

It also removes the additional fields that were added, like the `carrier_id` in `stock.move` which didn't make much sense.


ivantodorovich:
Tests are failing because of module `sale_procurement_group_by_line`, as it's changing the way procurements are managed, affecting this module tests.

I think the module `sale_procurement_group_by_line` should provide a config setting to enable/disable this feature per-company or something like this, to avoid this kind of issues. Any ideas on how to solve this?

I tried adding EXCLUDE="sale_procurement_group_by_line" in travis.yaml, but it didn't really work.

sale_manual_delivery/README.rst
```diff
@@ -0,0 +1,86 @@
...
+    :target: https://translation.odoo-community.org/projects/sale-workflow-12-0/sale-workflow-12-0-sale_manual_delivery
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/167/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+The goal of this module is to allow the manual creation of delivery slips. If activated,
+a sale order won't directly impact directly the stock. It will not make a reservation on the stock.
```
ajaniszewska-dev, line 29:
```suggestion
a sale order won't directly impact  the stock. It will not make a reservation on the stock.
```

sale_manual_delivery/models/crm_team.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import fields, models
+
+
+class CrmTeam(models.Model):
+    _inherit = "crm.team"
+
+    manual_delivery = fields.Boolean(
+        string="Manual Delivery",
+        help="If Manual, the deliveries are not created at SO confirmation. "
```
ajaniszewska-dev:
```suggestion
        help="If True, the deliveries are not created at SO confirmation. "
```

sale_manual_delivery/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    manual_delivery = fields.Boolean(
+        string="Manual Delivery",
+        default=False,
+        help="If Manual, the deliveries are not created at SO confirmation. "
```
ajaniszewska-dev:
```suggestion
        help="If True, the deliveries are not created at SO confirmation. "
```
nitpicking: since boolean can have only True or False, Manual sounds more like a option for Selection field to me.

sale_manual_delivery/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+The goal of this module is to allow the manual creation of delivery slips. If activated,
+a sale order won't directly impact directly the stock. It will not make a reservation on the stock.
```
ajaniszewska-dev:
```suggestion
a sale order won't  impact directly the stock. It will not make a reservation on the stock.
```

ajaniszewska-dev (approved):

ivantodorovich:
Thanks @ajaniszewska-dev :)
Changes applied & also added a few more test cases to increase coverage

ivantodorovich:
ping @i-vyshnevska @rousseldenis @newtratip @phuctranerp

phuctranfxvn:
Hello, just made a quick review, LGTM, except for the readme file, you may need to regenerate it

ivantodorovich:
> Hello, just made a quick review, LGTM, except for the readme file, you may need to regenerate it

It gets regenerated by travis automatically :)

sale_manual_delivery/models/sale_order_line.py
```diff
@@ -0,0 +1,146 @@
...
+        "move_ids.location_dest_id",
+    )
+    def _compute_qty_procured(self):
+        """
+        Computes the already planned quantities for the given sale order lines,
+        based on the existing stock.moves
+        """
+        for line in self:
+            if line.qty_delivered_method == "stock_move":
+                line.qty_procured = line._get_qty_procurement()
```
i-vyshnevska, line 41: do we need to set 
`line.qty_procured = 0.0`
to cleaunp previously setup data?

sale_manual_delivery/wizard/manual_delivery.py
```diff
@@ -0,0 +1,88 @@
...
+                        "quantity": line.qty_to_procure,
+                    },
+                )
+                for line in sale_lines
+                if line.qty_to_procure and line.product_id.type != "service"
+            ]
+        return res
+
+    commercial_partner_id = fields.Many2one(
+        "res.partner", required=True, readonly=True, ondelete="cascade",
```
i-vyshnevska, line 53:
```suggestion
        "res.partner", required=True, readonly=True,
```
suppose not needed on the wizard, or even dangerous as the wizard will be removed soon or later, maybe it's disabled in core didn't checked

ivantodorovich: This is safe AFAIK, it's to automatically remove the wizard in case one of these foreign keys gets removed.
The wizard itself gets removed by `ir.autovacuum`, but I add `ondelete="cascade"` to avoid possible issues in case the garbage collector doesn't work or doesn't do it in time.


Possible issues be like: user unable to remove a partner just because the wizard record is still in the db

sale_manual_delivery/models/sale_order_line.py
ivantodorovich: Done! Thanks!

i-vyshnevska (approved):

PR #1570, opened 2021-04-29
[12.0] sale_coupon* from Odoo v13

eLBati:


rousseldenis:
@eLBati

eLBati:
@rousseldenis hi, `sale_promotion_rule` was merged into 12.0, so I think sale_coupon* modules are not interesting for OCA v12 and this PR should be closed

PR #1573, opened 2021-05-04
[13.0][ADD] sale_order_qty_change_no_recompute: Prevent recompute if only quantity has changed in sale order line-

victoralmau:
Prevent recompute if only quantity has changed in sale order line.

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT29404


youring:
Functional tested not work on [runbot.](http://3480769-1573-60ceb9.runbot1.odoo-community.org/?)
Change the unit price, then change the quantity will still trigger recomputing.

sale_order_qty_change_no_recompute/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.onchange(
+        "product_id", "price_unit", "product_uom", "product_uom_qty", "tax_id"
+    )
+    def _onchange_discount(self):
+        if (
+            self.product_uom_qty == self._origin.product_uom_qty
```
pedrobaeza: This condition is not correct

sale_order_qty_change_no_recompute/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.onchange(
+        "product_id", "price_unit", "product_uom", "product_uom_qty", "tax_id"
+    )
+    def _onchange_discount(self):
```
pedrobaeza: Are you sure this is the proper method to override?

pedrobaeza (changes requested):
Please generate README

HaraldPanten (changes requested):
I've been testing in runbot, but everytime the quantity is changed, prices are recomputed at sale order line level.

youring:
https://github.com/OCA/sale-workflow/commit/dcdef5c4616abd91270ef975cb3ca1b8361a5ad6 Works as expected now.

sale_order_qty_change_no_recompute/models/sale_order.py
victoralmau: Condition are changed now.

sale_order_qty_change_no_recompute/models/sale_order.py
victoralmau: Now addon override only `product_uom_change` method.

victoralmau:
> I've been testing in runbot, but everytime the quantity is changed, prices are recomputed at sale order line level.

Sorry, it's necessary save order before try to change quantity and prevent prices recomputation.

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -0,0 +1,51 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.product = cls.env["product.product"].create(
+            {"name": "Test Product", "list_price": 25.00}
+        )
+        cls.partner = cls.env.ref("base.res_partner_12")
+
+    def test_sale_line_without_change(self):
+        sale_form = Form(self.env["sale.order"])
```
pedrobaeza: Put this in `setUpClass`.

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_sale_line_without_change(self):
+        with self.sale_form.order_line.new() as line_form:
+            line_form.product_id = self.product
+            line_form.product_uom_qty = 1
+            self.assertEqual(line_form.price_unit, 25)
+            self.assertEqual(line_form.price_subtotal, 25)
+
+    def test_sale_line_change_price(self):
+        with self.sale_form.order_line.new() as line_form:
```
pedrobaeza: I think you can even put the line_form and the product set on the setUpClass

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -0,0 +1,47 @@
...
+        self.line_form.product_uom_qty = 1
+
+    def test_sale_line_without_change(self):
+        self._set_init_line()
+        self.assertEqual(self.line_form.price_unit, 25)
+        self.assertEqual(self.line_form.price_subtotal, 25)
+
+    def test_sale_line_change_price(self):
+        self._set_init_line()
+        self.line_form.price_unit = 10
```
pedrobaeza, line 25: I don't get the sense of this test. Changing the price unit is not what this module prevents.

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -0,0 +1,47 @@
...
+        super().setUpClass()
+        cls.product = cls.env["product.product"].create(
+            {"name": "Test Product", "list_price": 25.00}
+        )
+        sale_form = Form(cls.env["sale.order"])
+        sale_form.partner_id = cls.env.ref("base.res_partner_12")
+        with sale_form.order_line.new() as cls.line_form:
+            cls._set_init_line(cls)
+
+    def _set_init_line(self):
```
pedrobaeza: Why doing this sub-method? In fact the only thing you should test is:

- Fix a price on the line
- Change product > check the price changes
- Change quantity > check that the price doesn't change.

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
@HaraldPanten can you please update your review?

youring:
Functional tested on proper way. LG!
But I like the naming of this module by @StefanRijnhart , more accessible for functional user.
https://github.com/opener-am/various-odoo-addons/tree/8.0/sale_keep_unit_price_on_qty_change
Just FYI.

pedrobaeza:
That name is not accurate by several reasons:

- It doesn't include `order` for specifying in which part of `sale` this happens.
- The computation can affect also the discount, not only the price unit.

Anyway, there's no good name most of the times, but you have the README for explaining the exact goal ;) 

imagen

pedrobaeza:
Can you approve then the PR if you agree?

HaraldPanten (approved):
LGTM. Sorry @pedrobaeza , I didn't read you before.

incoluz:
Hello, this is not working for us on 14.0-20220606 core version. Only prevents when uom changes not on qty change.

PR #1575, opened 2021-05-06
[13.0][ADD] sale_not_cancel_with_validate_picking : New module to 13.0

manuelcalerosolis:
It does not allow you to cancel an order if a validated delivery.

image


sale_not_cancel_with_validate_picking/README.rst
```diff
@@ -0,0 +1,72 @@
+=====================================
+Sale Not Cancel with validate Picking
```
gurneyalex, line 2:
```suggestion
Sale forbid cancel if delivered
```

gurneyalex (commented):
ok generally speaking except for the name (but I won't fight over this). I suggest sale_forbid_cancel_delivered. At least s/validate/validated/

manuelcalerosolis:
This PR has been replaced by this other one, 

[https://github.com/OCA/sale-workflow/pull/1768](https://github.com/OCA/sale-workflow/pull/1768)

 already with the new name that he proposed @gurneyalex  to us, which I see as more accurate.

PR #1578, opened 2021-05-07
[13.0][IMP] sale_order_product_recommendation: Added option to filter sale_orders by delivery address

CarlosRoca13:
cc @Tecnativa TT29597

please @chienandalu @sergio-teruel review this

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -45,6 +45,7 @@ class SaleOrderRecommendation(models.TransientModel):
         string="Product price origin",
         default="pricelist",
     )
+    use_delivery_address = fields.Boolean(string="Use delivery address")
 
     @api.model
     def _default_order_id(self):
@@ -54,12 +55,20 @@ class SaleOrderRecommendation(models.TransientModel):
         """Domain to find recent SO lines."""
         start = datetime.now() - timedelta(days=self.months * 30)
         start = fields.Datetime.to_string(start)
+        partner = (
+            self.order_id.partner_shipping_id
+            if self.use_delivery_address
+            else self.order_id.partner_id.commercial_partner_id
+        )
+        sale_order_partner_field = (
+            "partner_shipping_id" if self.use_delivery_address else "partner_id"
+        )
         other_sales = self.env["sale.order"].search(
             [
                 (
-                    "partner_id",
+                    sale_order_partner_field,
                     "child_of",
-                    self.order_id.partner_id.commercial_partner_id.id,
+                    partner.commercial_partner_id.id,
                 ),
                 ("date_order", ">=", start),
             ]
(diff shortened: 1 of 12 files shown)


sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -54,16 +55,28 @@ def _recomendable_sale_order_lines_domain(self):
...
+            other_sales = self.env["sale.order"].search(
+                [
+                    (
+                        "partner_shipping_id",
+                        "child_of",
+                        self.order_id.partner_shipping_id.commercial_partner_id.id,
+                    ),
+                    ("date_order", ">=", start),
+                ]
+            )
```
victoralmau:
```suggestion
        partner = self.order_id.partner_shipping_id if self.use_delivery_address else self.order_id.partner_id.commercial_partner_id
        sale_order_partner_field = "partner_id if self.use_delivery_address else "partner_shipping_id"
        other_sales = self.env["sale.order"].search(
            [
                (
                    sale_order_partner_field,
                    "child_of",
                    partner_id.commercial_partner_id.id,
                ),
                ("date_order", ">=", start),
            ]
        )
```

victoralmau (changes requested):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
CarlosRoca13: It isn't what I want to achieve... because with this solution you will allways search the sale orders by the partner_id... and I want to search it by partner_shipping_id when the option is selected

victoralmau: Sorry, i have updated my previous suggestion.

CarlosRoca13: I think it's still incorrect... The domain is allways the same... When the option is selected it should search by partner_shipping_id and not by partner id... I don't know if I'm explaining myself well 😅

victoralmau: Yes, of course, but these code:
```
partner_id = self.order_id.partner_id
if self.use_delivery_address:
    partner_id = self.order_id.partner_shipping_id
```
set "correct" `partner_id` to search after.

chienandalu: I'd say it's still wrong @victoralmau

chienandalu: @victoralmau I think that to respect the logic it would be:
```python
partner = self.order_id.partner_shipping_id if self.use_delivery_address else self.order_id.partner_id.commercial_partner_id
```
And then you can have the common filter

victoralmau: Ok, thanks, i have update my initial suggestion about these change.

CarlosRoca13:
@victoralmau changes done

victoralmau (approved):

chienandalu (approved):

pedrobaeza:
@CarlosRoca13 please rebase and check the error in Travis

CarlosRoca13:
@pedrobaeza fixed tests, the problem was that the dates were hardcoded... Now it souldn't happen again

pedrobaeza:
I think all the dates must be fixed instead of changing the fixed dates to variable ones.

CarlosRoca13:
Ok, so changed

PR #1582, opened 2021-05-10
[REF] sale_order_price_recalculation: recompute name with the ability to force language

len-foss:
Refactoring useful to make so it so that the recomputed name can be identical with a normal SOL creation, while being able to force the language.
This is typically useful if there's an "anonymous partner", but you want to use the website's context language:
https://github.com/shopinvader/odoo-shopinvader/pull/958


lmignon (approved):
LGTM (Code review)

lmignon:
ping @rousseldenis

PR #1583, opened 2021-05-11
[14.0][MIG] sale promotion rule

xavier-bouquiaux:


rousseldenis:
@hparfr IMHO, either a Roadmap, either a modification before merge should integrate coupon module that has been split by Odoo from 13 => 14 as it could be interesting to base our flow on coupon.coupon model

https://github.com/odoo/odoo/blob/14.0/addons/coupon/models/coupon.py#L10

Not too much logic there, no mandatory fields. What do you think ?

@sebastienbeau @Cedric-Pigeon @qgroulard

sale_promotion_rule/views/sale_promotion_rule.xml
```diff
@@ -0,0 +1,96 @@
...
+                                colspan="3"
+                            />
+                            <label for="restrict_partner_ids" colspan="3" />
+                            <field
+                                name="restrict_partner_ids"
+                                nolabel="1"
+                                colspan="3"
+                            />
+                        </group>
+                    </group>
```
hailangvn:
```xml
                    </group>
                    <group name="condition" string="Condition" col="6">
                        <label for="date_from" string="Date début/fin"/>
                        <field name="date_from" nolabel="1"/>
                        <field name="date_to" nolabel="1"/>
                        <field name="minimal_amount" colspan="3"/>
                        <field name="is_minimal_amount_tax_incl" colspan="3"/>
                        <field name="only_newsletter" colspan="3"/>
                        <field name="restrict_pricelist_ids" widget="many2many_tags" colspan="3"/>
                        <label for="restrict_partner_ids" colspan="6"/>
                        <field name="restrict_partner_ids" nolabel="1" colspan="6"/>
                    </group>
```

sale_promotion_rule/demo/product.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="coupon" model="product.product">
+        <field name="name">Coupon</field>
+        <field name="type">service</field>
+    </record>
+</odoo>
```
hailangvn, line 7: Is this needed with current coupon logic?

hailangvn (approved):
Thanks for the PR. There are minor improvements not required to merge this.

hparfr:
Why not. I was not aware of this coupon module.

hailangvn:
My comment was about demo coupon product, which IMHO helps new user to understand this module. I was wondering what to do next after having a coupon product while we already added coupon code to promotions page. There is another demo file as well.

sale_promotion_rule/views/sale_promotion_rule.xml
```diff
@@ -0,0 +1,96 @@
...
+                            />
+                            <field
+                                name="discount_product_id"
+                                attrs="{'required':  [('promo_type', '=', 'discount'), ('discount_type', 'in', ('amount_tax_included', 'amount_tax_excluded'))],
+                                           'invisible': ['|', ('promo_type', '!=', 'discount'), ('discount_type', 'not in', ('amount_tax_included', 'amount_tax_excluded'))]}"
+                            />
+                            <field name="multi_rule_strategy" />
+                        </group>
+                        <group name="condition" string="Condition" col="3">
+                            <label for="date_from" string="Date début/fin" />
```
rousseldenis: Could you change this to English version ?

sale_promotion_rule/views/sale_promotion_rule.xml
rousseldenis: @hailangvn This is reformated through pre-commit.

sale_promotion_rule/views/sale_promotion_rule.xml
rousseldenis: @xavier-bouquiaux

sale_promotion_rule/views/sale_promotion_rule.xml
hailangvn: Thanks @rousseldenis. I was new at the time I commented. :)

francesco-ooops:
@xavier-bouquiaux @rousseldenis wouldn't it make more sense to move this module to sale-promotion repo?

let us know if you need us to work on it

francesco-ooops:
or rather: what are the differences between this module's promotions and odoo standard promotions? should they coexist?

hparfr:
>or rather: what are the differences between this module's promotions and odoo standard promotions? should they coexist?

After few months of implementation with coupons in v14, it's missing key features for our e-commerce case (like apply more than 1 promotion at time).  Coupon is not designed to allow behavior changes. So it's a bit hacky to add features on top of it. We spent a lot of effort to make it work. 

Sale promotion rule, is an old module date here from 2017 and I think it came from a time when we had to synchronize with Magento / Prestashop and cover main features. It's designed and implemented to allow behavior changes if needed. I think a good refactoring will not hurt.

Like always, it depends of your needs and your budget.

rousseldenis:
@Cedric-Pigeon

rousseldenis:
> After few months of implementation with coupons in v14, it's missing key features for our e-commerce case (like apply more than 1 promotion at time). Coupon is not designed to allow behavior changes. So it's a bit hacky to add features on top of it. We spent a lot of effort to make it work.

@hparfr  But from 13 version, it is possible to use coupon for its model only and build on top of it as we did here : https://github.com/acsone/acsone-addons/tree/14.0/account_wallet_coupon

sale_promotion_rule/views/sale_promotion_rule.xml
sebastienbeau: ping @xavier-bouquiaux

sebastienbeau (approved):
LGTM (code review), we should merge it even odoo have an alternative implementation (coupon do not solve all case). In a long terme we should see if we can use coupon or improve this module. 
For now some customer need such feature so let's merge it

xavier-bouquiaux:
> LGTM (code review), we should merge it even odoo have an alternative implementation (coupon do not solve all case). In a long terme we should see if we can use coupon or improve this module. For now some customer need such feature so let's merge it

Changed to english version and rebased

regards

xbo

PR #1587, opened 2021-05-13
[ADD] Sale Order Delivery Products: New module to 13.0

manuelcalerosolis:
Insert a button on the sales order to directly access the first of the deliveries.

image.


sale_order_delivery_products/models/sale_order.py
```diff
@@ -0,0 +1,29 @@
...
+        pick_ids = self.mapped("picking_ids").sorted("id")
+        action = self.env.ref("stock.action_picking_tree_all")
+        result = action.read()[0]
+        # override the context to get id of the default filtering on operation type
+        result["context"] = {
+            "default_partner_id": self.partner_id.id,
+            "default_origin": self.name,
+        }
+        result["domain"] = [("id", "in", pick_ids.ids)]
+        result["res_id"] = pick_ids[:1].id
```
gurneyalex, line 23: don't you want to filter on the state of the pickings here? I can easily think of cases where the first picking is done, and this would defeat the purpose of the module.

manuelcalerosolis: Are you proposing to open the first non-validated picking?

sale_order_delivery_products/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "sale.order"
+
+    def action_view_picking(self):
+        pick_ids = (
+            self.mapped("picking_ids")
+            .filtered(lambda x: x.state not in ("done", "cancel"))
+            .sorted("id")
+        )
+        action = self.env.ref("stock.action_picking_tree_all")
+        result = action.read()[0]
```
rousseldenis, line 16: Please use for_xml_id(). See: https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_actions.py#L209

sale_order_delivery_products/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+        # override the context to get id of the default filtering on operation type
+        result["context"] = {
+            "default_partner_id": self.partner_id.id,
+            "default_origin": self.name,
+        }
+        result["domain"] = [("id", "in", pick_ids.ids)]
```
rousseldenis, line 22:
```suggestion
        result["domain"] = [("picking_ids.state", "not in", ("done", "cancel"))]
```

To avoid a search

sale_order_delivery_products/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+    def action_view_picking(self):
+        pick_ids = (
+            self.mapped("picking_ids")
+            .filtered(lambda x: x.state not in ("done", "cancel"))
+            .sorted("id")
+        )
+        action = self.env.ref("stock.action_picking_tree_all")
+        result = action.read()[0]
+        # override the context to get id of the default filtering on operation type
+        result["context"] = {
```
rousseldenis, line 18: Do you really want to remove all possible inherited context values ?

sale_order_delivery_products/views/sale_order_views.xml
```diff
@@ -0,0 +1,21 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.delivery_product</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='action_quotation_send']" position="after">
+                <button
+                    string="Deliver Products"
```
rousseldenis, line 12: Is that really 'delivery' pickings ? As in several steps configuration, you can have Pick and Pack ones too

rousseldenis (changes requested):
Some changes and one question

PR #1589, opened 2021-05-14
sale_product_set: fix SO line vals get

simahawk:
Ensure all required onchanges are applied on new SO line values.

Not all the modules rely on 'product_id_change'
to change values related to products.

A typical case is 'sale_stock' whereas product lead time
is handled by '_onchange_product_id_set_customer_lead'.

In this example, without this change,
all lines added via product set wizard will have a wrong lead time.

### DIFF
--- a/sale_product_set/wizard/product_set_add.py
+++ b/sale_product_set/wizard/product_set_add.py
@@ -123,11 +123,9 @@ class ProductSetAdd(models.TransientModel):
 
     def prepare_sale_order_line_data(self, set_line, max_sequence=0):
         self.ensure_one()
-        sale_line = self.env["sale.order.line"].new(
-            set_line.prepare_sale_order_line_values(
-                self.order_id, self.quantity, max_sequence=max_sequence
-            )
+        line_values = set_line.prepare_sale_order_line_values(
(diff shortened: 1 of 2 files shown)


sale_product_set/wizard/product_set_add.py
```diff
@@ -123,11 +123,9 @@ def _get_lines(self):
...
-            set_line.prepare_sale_order_line_values(
-                self.order_id, self.quantity, max_sequence=max_sequence
-            )
+        line_values = set_line.prepare_sale_order_line_values(
+            self.order_id, self.quantity, max_sequence=max_sequence
         )
-        sale_line.product_id_change()
-        line_values = sale_line._convert_to_write(sale_line._cache)
+        sol_model = self.env["sale.order.line"]
+        line_values.update(sol_model.play_onchanges(line_values, line_values.keys()))
```
sebalix, line 130: Can't we use `odoo.tests.common.Form` instead of adding the dependency on `onchange_helper` here?

simahawk: It turned out it's much slower than onchange_helper.

mmequignon (approved):

PR #1591, opened 2021-05-17
[IMP] sale_order_line_price_history: turn into a widget

chienandalu:
This way, the user doesn't have to save the line first to see the
customer prices.

cc @Tecnativa TT29630

history-in-new-record

please review @Yajo @Tardo


pedrobaeza:
Can we let it in black color as previously?

chienandalu:
> Can we let it in black color as previously?

Oh, yes, sure. Done

MeetKD (approved):

sale_order_line_price_history/wizards/sale_order_line_price_history.py
```diff
@@ -12,12 +12,19 @@ def _default_partner_id(self):
         line_id = self.env.context.get("active_id")
         return self.env["sale.order.line"].browse(line_id).order_partner_id.id
 
+    @api.model
+    def _default_product_id(self):
+        line_id = self.env.context.get("active_id")
+        return self.env["sale.order.line"].browse(line_id).product_id.id
```
pedrobaeza, line 18: Why do you need this if you are already passing `default_product_id` in the context?

pedrobaeza (approved):
Tested in runbot

sale_order_line_price_history/wizards/sale_order_line_price_history.py
chienandalu: This was a related field. I preferred to keep the ability to call the wizard with just the line id in the context

PR #1592, opened 2021-05-19
[14.0][MIG] sale_order_qty_change_no_recompute: Migration to 14.0

victoralmau:
Migration to 14.0.

Locked by:
- [x] https://github.com/OCA/sale-workflow/pull/1593

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT29404

PR #1593, opened 2021-05-19
[13.0][IMP] sale_order_qty_change_no_recompute: Improve tests

victoralmau:
Improve tests.

It's need to apply to 14.0 (https://github.com/OCA/sale-workflow/pull/1592)

Please @joao-p-marques and @pedrobaeza can you review it?

@Tecnativa TT29404

### DIFF
--- a/sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
+++ b/sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
@@ -8,13 +8,15 @@ class TestSaleOrderQtyChange(SavepointCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.product_1 = cls.env["product.product"].create(
-            {"name": "Test Product 1", "list_price": 25.00}
+            {"name": "Test Product 1", "list_price": 25.00, "taxes_id": False}
         )
         cls.product_2 = cls.env["product.product"].create(
-            {"name": "Test Product 2", "list_price": 30.00}
+            {"name": "Test Product 2", "list_price": 30.00, "taxes_id": False}
         )
+        pricelist = cls.env["product.pricelist"].create({"name": "Test pricelist"})
         sale_form = Form(cls.env["sale.order"])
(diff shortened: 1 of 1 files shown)


sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -7,14 +7,32 @@ class TestSaleOrderQtyChange(SavepointCase):
...
+                "taxes_id": [(6, 0, cls.tax_0.ids)],
+            }
         )
         cls.product_2 = cls.env["product.product"].create(
-            {"name": "Test Product 2", "list_price": 30.00}
+            {
+                "name": "Test Product 2",
+                "list_price": 30.00,
+                "taxes_id": [(6, 0, cls.tax_0.ids)],
+            }
```
joao-p-marques: Can't you make `taxes_id` empty instead of creating a tax of 0%?

joao-p-marques (changes requested):

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
joao-p-marques: I don't know if removing `taxes_id` completely will do it, as it might be computing it somewhere else... Could you initialize it as `False`, for example?

victoralmau: `taxes_id` is not necessary.

joao-p-marques: I am not entirely sure about that... Check the Odoo code for the field: https://github.com/odoo/odoo/blob/13.0/addons/account/models/product.py#L25
If none is set, the default will be computed from the company's tax plan, so we can end up with problems in a CI environment with different modules... Maybe it's better to explicitly mark it as `False` IMHO

joao-p-marques (approved):

HaraldPanten (approved):

pedrobaeza (approved):

PR #1597, opened 2021-05-25
[14.0][ADD] - New module sale_order_change_analytic_account

sbejaoui:
This addon allow user to update analytic account on sale orders and related journal entries.

### DIFF
--- a/sale_order_change_analytic_account/models/sale_order.py
+++ b/sale_order_change_analytic_account/models/sale_order.py
@@ -1,0 +1,18 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_open_sale_order_change_analytic_wizard(self):
+        return {
+            "name": _("Update Analytic Account"),
+            "type": "ir.actions.act_window",
+            "res_model": "sale.order.change.analytic",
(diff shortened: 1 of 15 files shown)


sbejaoui:
cc/ @rousseldenis

sale_order_change_analytic_account/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_open_sale_order_change_analytic_wizard(self):
```
rousseldenis, line 10: @sbejaoui This function is not called anywhere.

rousseldenis (changes requested):
Could you check behaviour as no action link nor button is present on sale order ?

Could you add a readme fragment that explain a little bit more ('the analytic account is readonly on sale orders with invoices, so, this is a helper to update all already linked invoices' or something similar) ?

sale_order_change_analytic_account/models/sale_order.py
sbejaoui: fixed

rousseldenis (approved):
Code and functional review

rousseldenis:
The only thing is that analytic account is maybe undefined on Account Invoice level. It's only updated on Invoice lines.

Maybe something should be formalized on README. @Cedric-Pigeon What do you think ?

PR #1598, opened 2021-05-25
[FIX] for drop shipping module

vrenaville:
* Purpose
The stock_dropshipping module do not have a default value:
https://github.com/odoo/odoo/blob/14.0/addons/stock_dropshipping/models/sale.py#L22
We need to force the parameter


rousseldenis:
@vrenaville Could you check pre-commit and maybe add a test to highlight this change ?

gurneyalex (approved):

gurneyalex:
@vrenaville I just pushed on your branch to fix the CI

PR #1604, opened 2021-06-02
[13.0][ADD] sale_order_disable_user_autosubscribe

mmequignon:
This module removes the order's salesperson from default followers.


pedrobaeza:
I think this should go to OCA/social, and see if this can be more generic, not only for sales, but that's not a requirement.

salesperson_disable_mail_autosubscribe/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "sale.order"
+
+    def _message_auto_subscribe_followers(self, updated_values, default_subtype_ids):
+        # By default, _message_auto_subscribe_followers adds the user_id of
+        # any record inheriting mail.thread to the list of followers.
+        # This removes this follower from the auto-subscribed followers.
+        res = super()._message_auto_subscribe_followers(
+            updated_values, default_subtype_ids
+        )
+        user_id = updated_values.get("user_id")
```
sebalix: Explain why you do it as post operation instead of removing the `user_id` from `updated_values` before calling `super` maybe?

mmequignon: done

atchuthan (approved):

sebalix (approved):

simahawk:
I agree w/ Pedro that this could be done generic on mail.thread. We could have configuration on ir.model or via config param. Would you mind adding this to ROADMAP?

mmequignon:
> I agree w/ Pedro that this could be done generic on mail.thread. We could have configuration on ir.model or via config param. Would you mind adding this to ROADMAP?

All right, I'm agree with Pedro too. Thanks for your reviews!

PR #1607, opened 2021-06-04
[14.0] ci: migrate to GitHub actions

sbidoul:
As there are travis issues in https://github.com/OCA/product-variant/pull/217, I propose to use this repo as the next test for GitHub actions CI.

mis-builder (14.0) and pim (all branches) are already configured that way, as well as all shoinvader repos.

Things to note:
- compared to travis/mqt, these it should run the exact same tests, so no surprise expected, but
- pay particular attention to false positives (i.e. read the build logs to detect anomalies)
- there are additional checks for license and development status compatibility


sbidoul:
Hm, wrong repo :/

sbidoul:
Setting to draft as we need a runbot upgrade to work without .travis.yml.

PR #1610, opened 2021-06-11
[12.0][NEW] Partner Industry Pricelist

mileo:


partner_industry_pricelist/models/res_partner_industry.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResPartnerIndustry(models.Model):
+
+    _inherit = 'res.partner.industry'
+
+    pricelist_id = fields.Many2one(
+        string='Pricelist',
+        comodel_name='product.pricelist',
+    )
```
legalsylvain, line 14: we could have a button @button_apply_pricelist() to set the pricelist to all partner that have a given ``industry_id``, don't you think ?

legalsylvain (commented):
partial review.

PR #1611, opened 2021-06-13
[14.0] [MIG] sale_order_line_packaging_qty

estekadigital:


sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,104 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools import float_round
+
+
+class SaleOrderLine(models.Model):
+
+    _inherit = "sale.order.line"
+
+    product_packaging_qty = fields.Integer(
```
rousseldenis, line 12: This should be Float

estekadigital: In my opinion it shouldn't be float, actually there was a comment in the previous code suggesting to make it integer. As a seller if I configure packaging to sell my product I want to sell it that way, otherwise the functionality is further complicated

simahawk:
already in v14 https://github.com/OCA/sale-workflow/tree/14.0/sale_order_line_packaging_qty feel to propose fixes or improvements

PR #1615, opened 2021-06-15
[14.0][FIX] sale_product_seasonality: Fix using templates on sale order

jcoux:
By default, the field `product_template_id` is hidden on sale order
lines, but we can display it.

If it's displayed, we want to have product templates limites to allowed
products.

### DIFF
--- a/sale_product_seasonality/views/sale_order.xml
+++ b/sale_product_seasonality/views/sale_order.xml
@@ -28,6 +28,15 @@
                 >[('id', 'in', season_allowed_product_ids)]</attribute>
                 <attribute name="options">{"no_create": True}</attribute>
             </xpath>
+            <xpath
+                expr="//field[@name='order_line']//tree//field[@name='product_template_id']"
+                position="attributes"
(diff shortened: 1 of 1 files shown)


sale_product_seasonality/models/sale_order.py
```diff
@@ -24,6 +24,13 @@ class SaleOrder(models.Model):
         column2="product_id",
         compute="_compute_season_allowed_product_ids",
     )
+    season_allowed_product_template_ids = fields.Many2many(
```
simahawk: this makes the other field useless because we could use the template as domain for variants.

simahawk (changes requested):

sale_product_seasonality/models/sale_order.py
jcoux: You're right.
My way was totally bad.
Redo in a very more easy way.

damdam-s (approved):

simahawk (approved):

Issue #1621, opened 2021-06-20
[13.0] sale_partner_delivery_window: Error on tests


pedrobaeza:
This module is provoking an error on current CI:

https://travis-ci.com/github/OCA/sale-workflow/jobs/516775218#L2177

Can you please check @simahawk @sebalix @grindtildeath ?

simahawk:
It seems another error happens before on sale_by_packaging https://travis-ci.com/github/OCA/sale-workflow/jobs/516775218#L2129

pedrobaeza:
Thanks for the quick reaction.

grindtildeath:
Issue should be fixed with this PR: https://github.com/OCA/stock-logistics-workflow/pull/862

PR #1616, opened 2021-06-15
[IMP] sale_order_product_recommendation: Get sale orders with sudo to get orders from other salesperson

sergio-teruel:
cc @Tecnativa TT30453

ping @carlosdauden @chienandalu

PR #1622, opened 2021-06-21
[13.0] s_partner_delivery_window: improve err message

simahawk:
EDIT: Original issue fixed in https://github.com/OCA/stock-logistics-workflow/pull/862

Old description:

> Error reported here #1621 https://travis-ci.com/github/OCA/sale-workflow/jobs/516775218#L2129
> 
> Might be related to `sale_by_packaging` tests using NewId records that could make `sale_stock.models.sale_order._compute_qty_at_date` fail.
> 
> This method is overridden by `sale_partner_delivery_window` that's why is reported on the traceback but it does not change its behavior.
> 
> I'll get back to this.

### DIFF
--- a/sale_partner_delivery_window/models/res_partner.py
+++ b/sale_partner_delivery_window/models/res_partner.py
@@ -34,7 +34,10 @@ class ResPartner(models.Model):
         for dwin_start in datetime_windows:
             if dwin_start >= from_date:
                 return dwin_start
-        raise UserError(_("Something went wrong trying to find next delivery window"))
+        raise UserError(
+            _("Something went wrong trying to find next delivery window. Date: %s")
+            % str(from_date)
+        )
 
     def get_next_windows_start_datetime(self, from_datetime, to_datetime):
         """Get all delivery windows start time.


simahawk:
@pedrobaeza it's green here :thinking: 
Seems like a flappy test...  I'll try to reproduce it somehow.

pedrobaeza:
Uhm, I wasn't able to launch the merge in 3 occasions in #1616

simahawk:
cool, perfect conditions for debugging :sweat_smile:

simahawk:
and here it fails on another module `sale_order_carrier_auto_assign` https://travis-ci.com/github/OCA/sale-workflow/jobs/516789691#L2328 the common ground: all these tests rely on a test form and the computation of `virtual_available_at_date` (from `sale_stock`) seems to be broken because of cache miss.

pedrobaeza:
The common ground I see is the module `sale_partner_delivery_window`

simahawk:
@pedrobaeza that's the last link on the chain of inheritance. When it fails there it was already failing before.
See "During handling of the above exception, another exception occurred:" message.
And that module only alters the depends of that compute method.
But now I see that this depends is wrong because it adds a dependency on `expected_date` which is not stored and this might cause the cache miss error.

I couldn't reproduce the failure locally, but I've removed the depends override and improved a bit the error msg.

sale_partner_delivery_window/models/sale_order_line.py
```diff
@@ -80,8 +80,3 @@ def _prepare_procurement_values_time_windows(self, date_planned):
                 self.name,
             )
         return
-
```
simahawk: @grindtildeath do you remember why this was done?
Also, do you have a specific case where we want to keep the the UserError on `next_delivery_window_start_datetime`?
Can't we do anything better?

grindtildeath:
@simahawk @pedrobaeza As I understand, the issue is actually this change:
https://github.com/OCA/stock-logistics-workflow/commit/13e3613d50c0a68e62051768b914878b2e2ce4c7#diff-fbea30ea878379785066dc89fe393a00ee17bb0a5152638bd5bed547969c1d0dR26

A new default value is set on the fields, but it's not reflected in sale_partner_delivery_window.

We can merge this PR to avoid the issue: https://github.com/OCA/stock-logistics-workflow/pull/862

simahawk:
@grindtildeath kept only the err msg improvement

grindtildeath (approved):

PR #1490, opened 2021-02-26
[14.0][ADD] sale_cancel_confirm

kittiu:
Depends on,
- [x] base_cancel_confirm https://github.com/OCA/server-ux/pull/297

This module extend `base_cancel_confirm` to make cancel confirmation with cancel reason (none / optional / requried).

Selection_044

image

**Note:** We use `sale_cancel_reason` before, but we have we have requirement to make cancel confirm/reason for many documents (almost 10 kind of documents). Rather than to make xxx_cancel_reason one by one, this way, make it very easy to achieve that.

TODO:

- [ ] Unit test

PR #1627, opened 2021-06-22
[14.0][MIG] sale_cancel_reason

psugne:


theangryangel:
Installed and manually tested on 14.0.

Unit tests seem to fail without demo data - not sure if that's a concern or not?

atchuthan:
There is already an open PR: https://github.com/OCA/sale-workflow/pull/1490
Please test the open PR and review it such that we can merge it to 14.0

theangryangel:
@atchuthan I've taken a look, but your module makes a major change in that the cancel reason is free text, rather than selection of another model. Tell me if I'm incorrect?

That's a relatively significant difference when you take into account ease of reporting, which is a key use case for us.

i.e. "doesn't want it", "customer no longer interested", "does not want the product" are all basically the same reason, but would be a pain to collate into a single metric. Now multiply that across all the other variations.

atchuthan:
@theangryangel I have raised your doubt in https://github.com/OCA/server-ux/pull/297#issuecomment-872857207.  Lets wait for the reply.

sale_cancel_reason/wizard/cancel_reason.py
```diff
@@ -0,0 +1,25 @@
+# © 2013 Guewen Baconnier, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class SaleOrderCancel(models.TransientModel):
+
+    """ Ask a reason for the sale order cancellation."""
+
+    _name = "sale.order.cancel"
```
CasVissers-360ERP: This model is already defined in base odoo-14.0/addons/sale/wizard

kittiu:
> @theangryangel I have raised your doubt in [OCA/server-ux#297 (comment)](https://github.com/OCA/server-ux/pull/297#issuecomment-872857207). Lets wait for the reply.

These are 2 separated modules. This one `sale_cancel_reason` should still exists.
We did `sale_cancel_confirm` by extend `base_cancel_confirm` and it has simple function (free text) and easier to create.
**Note:** 
* another reason, in our project, we need xxx_cancel_confirm for almost all kinds of document, and the simple function will do the job.
* this is all the code required in xxx_cancel_confirm, https://github.com/OCA/sale-workflow/blob/14.0/sale_cancel_confirm/model/sale.py

CasVissers-360ERP:
To always show the wizard I think we need to add:
```
    def _show_cancel_wizard(self):
        for order in self:
            if order._context.get('disable_cancel_warning'):
                return False
        return True
```

ao-landoo:
@psugne Are you continuing with this PR?

rousseldenis:
@psugne Do you mind if we switch to #2158 ?

PR #1628, opened 2021-06-23
[FIX] use same version as other oca projects

llacroix:
This cause a conflict in some tools thinking they are different
versions.

    The conflict is caused by:
      The user requested openupgradelib 3.0.0 (from git+https://github.com/OCA/openupgradelib.git)
      The user requested openupgradelib 3.0.0 (from git+https://github.com/OCA/openupgradelib.git@master)


dreispt (approved):

llacroix:
thanks

PR #1631, opened 2021-06-23
[13.0][FIX]sale_procurement_group_by_line: use SO company as the procurement company

JordiMForgeFlow:
When creating procurements for sales order lines, use the company in the sales order to prevent selecting the current active company, as in multi-company environments this can lead to inconsistencies between the company in the stock moves created and the actual company of the order.


AaronHForgeFlow (approved):
Totally agree

LoisRForgeFlow (approved):
:+1:

PR #1636, opened 2021-06-28
[FIX] sale_order_action_invoice_create_hook

azucenatrey:
PROBLEM DETECTED:
    When the "sale_order_action_invoice_create_hook" module is installed, Odoo executes the internal function "new_action_invoice_create" instead of the generic "action_invoice_create", so the latter is not executed. So the functionality of the original function "action_invoice_create" and "new_action_invoice_create" cannot be inherited from another module because the latter function is private and is not known from outside.

PROPOSED SOLUTION:
    A new generic function "_modify_invoices" is added outside the hook and it is called from within "new_action_invoice_create" to be able to inherit it in other modules and to be able to make the necessary modifications.

### DIFF
--- a/sale_order_action_invoice_create_hook/__manifest__.py
+++ b/sale_order_action_invoice_create_hook/__manifest__.py
@@ -6,7 +6,7 @@
     "summary": "Add more flexibility in the grouping parameters "
                "for the creation of invoices",
     "author": "Eficent, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


rlizana:
LGTM :+1:

rlizana (approved):
LGTM :+1:

jctrey:
LGTM +1

MiguelPoyatos (approved):

PR #1643, opened 2021-07-03
12.0 mig sale pricelist triple discount

scigghia:
Migration to 12.0

sale pricelist triple discount

actually works with pricelist policy = without_discount


sale_pricelist_triple_discount/model/sale.py
```diff
@@ -0,0 +1,51 @@
...
+#  Copyright (c) 2021 Andrea Cometa - Apulia Software s.r.l.
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    """
```
rousseldenis: @scigghia Could you remove unused comments ?

rousseldenis (changes requested):

rousseldenis:
And could you check flake8 ?

sale_pricelist_triple_discount/model/sale.py
scigghia: sure
thanks

SimoRubi (changes requested):
Please clean commit history, right now there are:

- A commit `[MIG] sale_pricelist_triple_discount` that mainly migrates `sale_pricelist_discount`; this should migrate `sale_pricelist_triple_discount` instead
- A commit `[MIG] sale_pricelist_discount` that removes `sale_pricelist_discount` and also does things to `sale_pricelist_triple_discount`; this can be dropped.
  If `sale_pricelist_discount` is no more needed, just remove it from `sale_pricelist_triple_discount`'s manifest and maybe mention in the migration issue https://github.com/OCA/sale-workflow/issues/700 that the module is obsolete in `12.0`; there is no need to have its code in this PR.
- A commit `works with pricelist policy, rem unused code and comments` that edits `sale_pricelist_triple_discount`

There should be only one commit that migrates `sale_pricelist_triple_discount`, note the recommended format of commit message:
```
git commit -m "[MIG] $MODULE: Migration to 12.0"
```
 (from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#technical-method-to-migrate-a-module-from-110-to-120-branch).
As long as there is a discussion in the PR you can also keep the temporary commits like `works with...` because it can come in handy to the reviewer for checking the new changes, but they should be squashed/fixed up before merge.

PR #1647, opened 2021-07-07
[14.0] [NEW] sale_product_approval: New module added in the repo.

Chandresh-OSI:
CC @patrickrwilson @dreispt


patrickrwilson:
@Chandresh-OSI I was testing this on my local instance and couldn't get the 'sale_ok' field to properly set when changing stages. During troubleshooting, if I made the sale_ok not readonly it worked so that seemed to be preventing the field to change properly. I'm testing on enterprise, can you try this on enterprise and see if you see the same issue?

sale_product_approval/models/product_state.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductState(models.Model):
+    _inherit = "product.state"
+
+    approved_sale = fields.Boolean(
+        string="Approved to be Sold",
```
dreispt: To not break current behavior once installed 
```suggestion
        string="Approved to be Sold",
        default=True,
```

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    sale_ok = fields.Boolean("Can be Sold", default=False)
+    candidate_sale = fields.Boolean(
+        string="Candidate to be Sold",
+    )
```
dreispt, line 25: On install, we need a script to copy the `sale_ok` values to the `candidate_sale` column.

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.onchange("product_state_id")
+    def _onchange_product_state(self):
```
dreispt: This won't work for direct API writes, only for user form interaction.
I suggest instead adding a new-stype "onchange" computation to sale_ok.

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+        self.sale_ok = False
+        if self.product_state_id.approved_sale and self.candidate_sale:
+            self.sale_ok = True
```
dreispt:
```suggestion
        if self.product_state_id:
            self.sale_ok = self.candidate_sale and self.product_state_id.approved_sale
```

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.model
+    def _get_default_product_state_id(self):
```
dreispt, line 78: Is this missing from the product_state module?

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+        res = super(ProductProduct, self).write(vals)
+        if not vals.get("sale_ok"):
+            order_ids = (
+                self.env["sale.order.line"]
+                .search(
+                    [("product_id", "=", self.id), ("state", "in", ["draft", "sent"])]
+                )
+                .mapped("order_id")
+            )
+            self.env["sale.order"]._log_product_state(order_ids, self)
```
dreispt: Instead of "pushing" these updates to Sales Orders, we could have them "pulled" on Sales Orders Lines with a computed field, and then on Sales Order with a computed field depending on SO Lines.

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+        return res
+
+    @api.onchange("product_state_id")
+    def _onchange_product_state(self):
```
dreispt: Not sure we will need this, once we have a computed field.

sale_product_approval/models/product_state.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+    def write(self, vals):
+        res = super(ProductState, self).write(vals)
+        if vals.get("approved_sale"):
+            product_ids = self.env["product.product"].search(
+                [("product_state_id", "=", self.id), ("candidate_sale", "=", True)]
+            )
+            for product in product_ids:
+                product.sale_ok = True
```
dreispt: Make sale_ok computed, and this won't be needed.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,52 @@
...
+        string="Approved for Sale",
+        store=True,
+        default=True,
+    )
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    exceptions = fields.Boolean(
```
dreispt: Field name look too short prone to name collisions 
```suggestion
    exceptions_sale_approval = fields.Boolean(
```

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,52 @@
...
+        string="Exception",
+    )
+
+    def _compute_exceptions(self):
+        self.exceptions = False
+        if any(not line.approved_sale for line in self.order_line):
+            self.exceptions = True
+
+    def _log_exception_activity_sale(self, render_method, documents, product_id):
+        for order in documents:
```
dreispt: OOP antipattern
```suggestion
    def _log_exception_activity_sale(self, render_method, product_id):
        for order in self:
```

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,52 @@
...
+            note = render_method(order, product_id)
+            order.activity_schedule(
+                "mail.mail_activity_data_warning",
+                date.today(),
+                note=note,
+                user_id=order.user_id.id or SUPERUSER_ID,
+            )
+
+    def _log_product_state(self, documents, product_id, cancel=False):
```
dreispt: I now see why there we have the write() logic calling this.
Thinking that maybe we could do this with a cron, but let's not go that way for now.

patrickrwilson:
@Chandresh-OSI It mostly works, I can see the exception banner on the SO and the icon on the SO line but the exception activity doesn't seem to be created anymore. Is this on purpose or a bug?
image

patrickrwilson:
@Chandresh-OSI Tested and works great after the last commit, just need a bit more coverage on tests.

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+        string="Can be Sold",
+        default=False,
+        store=True,
+    )
+    candidate_sale = fields.Boolean(
+        string="Candidate to be Sold",
+        default=True,
+    )
+
+    @api.depends("product_state_id")
```
dreispt: Missing dependencies
```suggestion
    @api.depends("candidate_sale", "product_state_id.approved_sale")
```

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+        for product in self:
+            if product.product_state_id:
+                product.sale_ok = (
+                    product.candidate_sale and product.product_state_id.approved_sale
+                )
+                if not product.sale_ok:
+                    product_variant_ids = (
+                        self.env["product.product"]
+                        .search([("product_tmpl_id", "=", product.id)])
+                        .ids
```
dreispt: Tip: as a general rule, work with records, and extracts ids as late as possible

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+                    )
+                    order_ids = (
+                        self.env["sale.order.line"]
+                        .search(
+                            [
+                                ("product_id", "in", product_variant_ids),
```
dreispt: `product_variant_ids` is available as a One2many field
```suggestion
                                ("product_id", "in", product.product_variant_ids.ids),
```

sale_product_approval/models/product_state.py
```diff
@@ -0,0 +1,23 @@
...
+        if vals.get("approved_sale") or vals.get("approved_sale") is False:
+            product_ids = self.env["product.product"].search(
+                [("product_state_id", "=", self.id), ("candidate_sale", "=", True)]
+            )
+            for product in product_ids:
+                if vals.get("approved_sale"):
+                    product.sale_ok = True
+                else:
+                    product.sale_ok = False
+        return res
```
dreispt: `sale_ok` is already computed, so this code should not be needed.
```suggestion
```

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+        compute="_compute_exceptions",
+        string="Exception",
+    )
+
+    def _compute_exceptions(self):
+        self.exceptions_sale_approval = False
+        if any(not line.approved_sale for line in self.order_line):
+            self.exceptions_sale_approval = True
+
+    def _log_exception_activity_sale(self, render_method, documents, product_id):
```
dreispt: OOP smells: `documents` is a Sales Order recordset. Just use `self`!

sale_product_approval/models/sale_order.py
dreispt: Still not fixed.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    exceptions_sale_approval = fields.Boolean(
+        compute="_compute_exceptions",
+        string="Exception",
+    )
+
+    def _compute_exceptions(self):
```
dreispt, line 19: Missing dependencies
```suggestion
    @api.depends("order_line.approved_sale")
    def _compute_exceptions(self):
```

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+        self.exceptions_sale_approval = False
+        if any(not line.approved_sale for line in self.order_line):
+            self.exceptions_sale_approval = True
```
dreispt: Simplifying
```suggestion
        self.exceptions_sale_approval = any(not line.approved_sale for line in self.order_line)
```

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+                                ("state", "in", ["draft", "sent"]),
+                            ]
+                        )
+                        .mapped("order_id")
+                    )
+                    self.env["sale.order"]._log_product_state(order_ids, self)
```
dreispt: Can we trigger the exception directly on the the Sale order Line, and have that logic there propagate the exception status up to the Sales Order?
The _log_product_state method causes an exception to be presented so the method name is surprising, and that's not a good thing.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+            note = render_method(order, product_id)
+            order.activity_schedule(
+                "mail.mail_activity_data_warning",
+                date.today(),
+                note=note,
+                user_id=order.user_id.id or SUPERUSER_ID,
+            )
+
+    def _log_product_state(self, documents, product_id, cancel=False):
+        def _render_product_state_excep(order, product_id):
```
dreispt: Why a nested method. These are not extensible, do we want to prevent that?

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+            values = {"sale_order_ref": order, "product_ref": product_id}
+            return self.env.ref("sale_product_approval.exception_on_product")._render(
+                values=values
+            )
+
+        self.env["sale.order"]._log_exception_activity_sale(
```
dreispt: Why `self.env["sale.order"]`if self is already a a sales order?

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+            _render_product_state_excep, documents, product_id
```
dreispt: Why pass a method as an argument?
This adds complexity, and I don''t see the need or the benefit.

dreispt:
@Chandresh-OSI @patrickrwilson There are some coding issues here that need improvement, can you work on this please?

Also, there is already a feature for exception handling.
It should be leveraged, and improved if it doesn't do the Activities already.
See https://github.com/OCA/sale-workflow/tree/14.0/sale_exception

patrickrwilson:
@Chandresh-OSI Looks very close but can you check travis? A test may need changes resulting from your last update. Thanks.

patrickrwilson:
@Chandresh-OSI can you squash commits then this should be ready to move forward.

Chandresh-OSI:
supersede by #1713

dreispt:
Why a new PR was needed?

Chandresh-OSI:
Due to commit squash issue. I have some merge commits in this PR and it is hard to squash individual commits. So I created new PR for the same code.

dreispt:
That doesn't require a new PR that restarts the review process.
You squash commits with `git rebase -i` and then force push the same branch.
Do you want to reopen and try that?
Reach out to me if you need more details.

patrickrwilson (approved):
This works great, LGTM

dreispt:
...there seems to be an issue with `sale_order_carrier_auto_assign` that is causing tests to fail: https://app.travis-ci.com/github/OCA/sale-workflow/jobs/536164726#L2357

dreispt:
Trying merge anyway...

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @api.depends("candidate_sale", "product_state_id.approved_sale")
+    def _compute_sale_ok_product(self):
+        for product in self:
+            if product.product_state_id:
+                product.sale_ok = (
+                    product.candidate_sale and product.product_state_id.approved_sale
+                )
+                if not product.sale_ok:
+                    order_ids = (
```
rousseldenis: Try to move this search outside the loop for performances reasons

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+
+    approved_sale = fields.Boolean(
+        related="product_id.sale_ok",
+        string="Approved for Sale",
+        store=True,
+        default=True,
+    )
+
+
+class SaleOrder(models.Model):
```
rousseldenis, line 10: Please put one model per file.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+    _inherit = "sale.order"
+
+    exceptions_sale_approval = fields.Boolean(
+        compute="_compute_exceptions", string="Exception", default=False
+    )
+
+    @api.depends("order_line.approved_sale")
+    def _compute_exceptions(self):
+        self.exceptions_sale_approval = any(
```
rousseldenis: You should modify this in order to support multi recordset

sale_product_approval/tests/test_product_approval.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.exceptions import UserError
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderLineDates(TransactionCase):
```
rousseldenis: SavepointCase (until v14) is better in terms of performance.

You can look at this at 6:22:00 https://www.youtube.com/watch?v=lT___hG110Q

rousseldenis (changes requested):
Code review

rousseldenis:
@Chandresh-OSI Could you attend comments and solve Travis ?

As you break standard behaviour you need to find a mean to activate or not this behaviour.

rousseldenis:
@Chandresh-OSI @dreispt What's the status of this ?

rousseldenis:
@Chandresh-OSI @dreispt

dreispt:
All review comments addressed, merging.

/ocabot merge nobump

PR #1648, opened 2021-07-07
[13.0][IMP] sale_order_product_assortment: Use product_assortment domain and only one field to get allowed products result from whitelist and blacklist

sergio-teruel:
cc @Tecnativa TT30584
ping @carlosdauden @joao-p-marques 

This PR depends on:
* https://github.com/OCA/product-attribute/pull/878

### DIFF
--- a/sale_order_product_assortment/models/sale_order.py
+++ b/sale_order_product_assortment/models/sale_order.py
@@ -6,52 +6,34 @@ from odoo import api, fields, models
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    whitelist_product_ids = fields.Many2many(
+    allowed_product_ids = fields.Many2many(
         comodel_name="product.product",
-        string="Whitelisted Products",
+        string="Allowed Products",
         compute="_compute_product_assortment_ids",
     )
-    blacklist_product_ids = fields.Many2many(
-        comodel_name="product.product",
-        string="Blacklisted Products",
-        compute="_compute_product_assortment_ids",
-    )
-    has_whitelist = fields.Boolean(compute="_compute_product_assortment_ids")
-    has_blacklist = fields.Boolean(compute="_compute_product_assortment_ids")
+    has_allowed_products = fields.Boolean(compute="_compute_product_assortment_ids")
 
-    @api.depends("partner_id")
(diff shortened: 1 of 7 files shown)


sale_order_product_assortment/models/sale_order.py
```diff
@@ -6,52 +6,31 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    whitelist_product_ids = fields.Many2many(
+    allowed_product_ids = fields.Many2many(
```
joao-p-marques, line 9: Would a migration script be needed to rename this table and keep the data?

joao-p-marques (commented):
Code review

sale_order_product_assortment/models/sale_order.py
sergio-teruel: No because this field is computed no store.

joao-p-marques: Oh right, sorry :man_facepalming:

joao-p-marques (approved):
Code review :+1:

pedrobaeza:
Please check tests

sergio-teruel:
This PR has dependencies

sale_order_product_assortment/models/sale_order.py
```diff
@@ -6,52 +6,34 @@
...
+        if self[partner_field]:
+            filters = IrFilters.browse()
+            filters_partner_domain = self.env["ir.filters"].search(
+                [("is_assortment", "=", True)]
             )
+            for ir_filter in filters_partner_domain:
+                if self[partner_field] & ir_filter.all_partner_ids:
+                    filters |= ir_filter
+                    self.allowed_product_ids += self.env["product.product"].search(
+                        ir_filter._get_eval_domain()
```
carlosdauden, line 36: We can add sale_ok=True to keep odoo standar behaviour

pedrobaeza: I have thought about this, but if added, you will see different results when previewing the filter results, and when searching for it in the sales order, so for sure that will confuse user. Do you think that's correct?

pedrobaeza (approved):

pedrobaeza:
@sbidoul working!

sbidoul:
Cool. I gave it a new token.

PR #1651, opened 2021-07-14
sale_discount_display_amount: fix init hook

simahawk:
Prevent failure if new columns are already in DB.

### DIFF
--- a/sale_discount_display_amount/hooks.py
+++ b/sale_discount_display_amount/hooks.py
@@ -4,33 +4,23 @@ import logging
 
 from odoo import SUPERUSER_ID
 from odoo.api import Environment
+from odoo.tools.sql import column_exists, create_column
 
 _logger = logging.getLogger(__name__)
 
+COLUMNS = (
+    ("sale_order", "price_total_no_discount"),
+    ("sale_order", "discount_total"),
+    ("sale_order_line", "price_total_no_discount"),
+    ("sale_order_line", "discount_total"),
+)
+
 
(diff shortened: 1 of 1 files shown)


mmequignon (approved):

sale_discount_display_amount/hooks.py
```diff
@@ -2,35 +2,32 @@
...
-    """
-    )
+    for table, column in COLUMNS:
+        if not column_exists(cr, table, column):
+            _logger.info("Create discount column %s in database", column)
+            cr.execute(
+                """
+                ALTER TABLE %s ADD COLUMN %s numeric;
+            """,
+                (AsIs(table), AsIs(column)),
```
sebalix: We should use [`sql.Identifier`](https://www.psycopg.org/docs/sql.html#psycopg2.sql.Identifier) for that no?
Check this discussion on StackOverflow: https://stackoverflow.com/questions/42947885/psycopg2-difference-between-asis-and-sql-module

simahawk: hmm good point. I've never used it TBH. In this case it doesn't change anything because the values are controlled. There's nothing to escape here. Still, I'll change it because it will encourage its usage in such cases ;)

simahawk: done

simahawk: I'm using the odoo helper now, they will deal w/ future changes if any ;)

PR #1646, opened 2021-07-06
[14.0][MIG][IMP] sale_order_type

BT-cserra:
This PR builds on https://github.com/OCA/sale-workflow/pull/1425

This PR attends the comments made on https://github.com/OCA/sale-workflow/pull/1425, including adding the commits of the two following PRs: https://github.com/OCA/sale-workflow/pull/1531 and https://github.com/OCA/sale-workflow/pull/1553. It also deletes the README so that it is re-generated.

PR #1654, opened 2021-07-15
[14.0][FIX] sale_order_type: call to create from copy crashes

BT-cserra:
When a record is created in odoo, all its fields are sent to the dictionary received by `create`, even if empty. However, this is not the case when we copy a record, in which case `create` is called without those fields having `copy=False`.

The previous implementation asumed that we would always have the field `date_order` when calling the `create` of the sale.order, without taking into account that being non-copy, it wouldn't be sent. As a result, the code crashed.

To fix this, the method `get` has been used instead of the index operator. If the key `date_order` is not found in `vals`, then `get` will return `None`, that makes everything go well.

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -74,7 +74,7 @@ class SaleOrder(models.Model):
             sale_type = self.env["sale.order.type"].browse(vals["type_id"])
             if sale_type.sequence_id:
                 vals["name"] = sale_type.sequence_id.next_by_id(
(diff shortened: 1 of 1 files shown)


BT-cserra:
Fixes a bug introduced in https://github.com/OCA/sale-workflow/pull/1646.

newtratip (approved):

BT-nfrancisco (approved):

ioans73 (approved):

PR #1396, opened 2020-12-26
[MIG] sale_order_line_packaging_qty: Migration to 14.0

phuctranfxvn:

Issue #1116, opened 2020-04-08
Installation error when i try to install this module.


Vignesh000:
I already install sale_isolated_quotation for the sequence numbers. then I install this module for revisions sequence it can't install.

bodedra:
For which version? :smile:

Vignesh000:
odoo12 :(

bodedra:
I have re-read your question and cannot able to understand. Can you update your question and provide clear information? And what do you mean by "this module"?

Vignesh000:
sorry, "this module(sale_order_revisions)". 
**odoo server error**
 "duplicate key value violates unique constraint "sale_order_revision_unique"
DETAIL: Key (unrevisioned_name, revision_number, company_id)=(TTPL/QUO/075, 0, 1) already exists." this error came when i installed sale_order_revisions module.

PR #1657, opened 2021-07-19
[14.0][FIX] sale_stock_picking_blocking: tests

nilshamerlinck:
This simple fix of the tests will help for the merge of https://github.com/OCA/sale-workflow/pull/1396

### DIFF
--- a/sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
+++ b/sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
@@ -32,17 +32,24 @@ class TestSaleDeliveryBlock(common.TransactionCase):
         product = (
             self.env["product.product"].with_user(self.user_test).create(prod_dict)
         )
-        # Create Sales order and lines:
+        # Create Sale order:
+        # TODO/TMP:
+        # - we explicitely add a name to avoid
+        #   a weird issue occuring randomly during tests
+        # - seems related to sale_order_revision,
+        #   further investigations ongoing
         so_dict = {
             "partner_id": self.env.ref("base.res_partner_1").id,
+            "name": "Test Sale Delivery Block",
         }
         self.sale_order = self.so_model.with_user(self.user_test).create(so_dict)
+        # Create Sale order lines:
         sol_dict = {
             "order_id": self.sale_order.id,
             "product_id": product.id,
             "product_uom_qty": 1.0,
         }
-        self.sale_order_line = self.sol_model.create(sol_dict)
+        self.sale_order_line = self.sol_model.with_user(self.user_test).create(sol_dict)
 
     def test_check_auto_done(self):
         # Set active auto done configuration


sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
```diff
@@ -35,14 +35,15 @@ def setUp(self):
         # Create Sales order and lines:
         so_dict = {
             "partner_id": self.env.ref("base.res_partner_1").id,
+            "name": "Test Sale Delivery Block",
```
simahawk, line 43: can you add a comment explaining why this is needed?

BTW IMO it's wrong that we need this. `sale_order_revision` [should not write revision related data](https://github.com/OCA/server-ux/blob/14.0/base_revision/models/base_revision.py#L83) if this is not needed.

simahawk: @kittiu I see you've migrated the module and before [this migration](https://github.com/OCA/sale-workflow/commit/4a38fcf20fcf55f115b36ccdf764ff48c15e91ca#diff-2b7f646de13a325a8564bc11e174f489a505e97bb29a567bbdbf576e222cb19e) a sequence was used.

IMO when we create the order first we should not trigger any revision machinery. The `unrevisioned_name` should probably be a computed stored field and depend on `name`.

What do you think?

simahawk: FTR an issue that might need an answer https://github.com/OCA/sale-workflow/issues/1116

nilshamerlinck:
Hi @simahawk my short term goal was to get travis green; I faced what I believe to be a non-deterministic issue and choose this fix as the most innocuous way to bypass it; I will continue to dig into the root cause, but not sure how much time it will take

simahawk:
> Hi @simahawk my short term goal was to get travis green; I faced what I believe to be a non-deterministic issue and choose this fix as the most innocuous way to bypass it; I will continue to dig into the root cause, but not sure how much time it will take

@nilshamerlinck that's ok and thanks for this time.

I'm not blocking, I'm saying that you should add a comment to remember our future selves why it has been done like this :wink: 
When you add the comment we can fast-track this.
Then we'll continue discussing w/ @kittiu (and whoever wants to join).

sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
kittiu: Noted, let me try to understand this. You are saying that, this PR need to add this line,
https://github.com/nilshamerlinck/sale-workflow/blob/14.0-fix-sale_stock_picking_blocking-tests/sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py#L43

Because without it, test error result from module sale_order_revision.

Correct?

PR #1659, opened 2021-07-20
[12.0] Fix sale_automatic_workflow: on some cases you ship more products than planned. You want to consider them as fully shipped

leemannd:
Fix sale_automatic_workflow: on some cases you ship more products than planned.
You want to consider them as fully shipped.

EDIT: forward ports done for 13.0 & 14.0


guewen (approved):
Thanks!

atchuthan (approved):

PR #1660, opened 2021-07-20
[13.0] Fix sale_automatic_workflow: on some cases you ship more products than planned. You want to consider them as fully shipped

leemannd:
Fix sale_automatic_workflow: on some cases you ship more products than planned.
You want to consider them as fully shipped.

Forward-port of https://github.com/OCA/sale-workflow/pull/1659

### DIFF
--- a/sale_automatic_workflow/models/sale_order.py
+++ b/sale_automatic_workflow/models/sale_order.py
@@ -32,7 +32,7 @@ class SaleOrder(models.Model):
                 or float_compare(
                     line.qty_delivered, line.product_uom_qty, precision_digits=precision
                 )
-                == 0
+                >= 0
                 for line in order.order_line
             )
 


skeller1 (approved):

SilvioC2C (approved):

sale_automatic_workflow/models/sale_order.py
```diff
@@ -32,7 +32,7 @@ def _compute_all_qty_delivered(self):
                 or float_compare(
                     line.qty_delivered, line.product_uom_qty, precision_digits=precision
                 )
-                == 0
+                >= 0
```
simahawk, line 35: FTR this is ok on https://github.com/OCA/sale-workflow/blob/14.0/sale_delivery_state/models/sale_order.py

simahawk (approved):

PR #1662, opened 2021-07-21
[13.0] Add sale_product_set_sale_by_packaging

simahawk:
Glue module between `sale_by_packaging` and `sale_product_set_packaging_qty`.

### DIFF
--- a/sale_product_set_sale_by_packaging/models/product_set_line.py
+++ b/sale_product_set_sale_by_packaging/models/product_set_line.py
@@ -1,0 +1,26 @@
+# Copyright 2021 Camptocamp SA
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import _, api, exceptions, fields, models
+
+
+class ProductSetLine(models.Model):
+    _inherit = "product.set.line"
+
+    # Just for UI purpose
+    sell_only_by_packaging = fields.Boolean(related="product_id.sell_only_by_packaging")
+
+    @api.constrains("product_id.sell_only_by_packaging", "product_packaging_id")
+    def _check_sell_only_by_packaging(self):
+        errored = self.filtered(
+            lambda x: x.product_id.sell_only_by_packaging and not x.product_packaging_id
+        )
+        if errored:
+            raise exceptions.UserError(
+                self._check_sell_only_by_packaging_err_msg(errored)
+            )
+
+    def _check_sell_only_by_packaging_err_msg(self, lines):
+        return _(
+            "The following product(s) can be sold only by packaging: \n   %s"
+        ) % ", ".join(lines.mapped("product_id.display_name"))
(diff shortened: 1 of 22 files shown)


sale_product_set_sale_by_packaging/models/product_set_line.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+        if errored:
+            raise exceptions.UserError(
+                self._check_sell_only_by_packaging_err_msg(errored)
+            )
+
+    def _check_sell_only_by_packaging_err_msg(self, lines):
+        return _(
+            "The following product(s) can be sold only by packaging: \n   %s"
+        ) % ", ".join(lines.mapped("product_id.display_name"))
```
mmequignon, line 26:
```suggestion
    @api.constrains("product_id.sell_only_by_packaging", "product_packaging_id")
    def _check_sell_only_by_packaging(self):
        errored = self.filtered(
            lambda x: x.product_id.sell_only_by_packaging and not x.product_packaging_id
        )
        if errored:
            raise exceptions.UserError(
                self._get_sell_only_by_packaging_err_msg(errored)
            )

    def _get_sell_only_by_packaging_err_msg(self, lines):
        return _(
            "The following product(s) can be sold only by packaging: \n   %s"
        ) % ", ".join(lines.mapped("product_id.display_name"))
```

mmequignon (approved):

sale_product_set_sale_by_packaging/models/product_set_line.py
simahawk: I prefer to prefix the name w/ the main method name (kind of a namespace)

sale_product_set_sale_by_packaging/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+                value = set(rec.product_variant_ids.ids).intersection(ids_to_check)
+            rec.sell_only_by_packaging_prod_set_tocheck = value
+
+    def _product_ids_to_check(self):
+        """Retrieves product.product IDS to check on product sets.
+        """
+        ids_to_check = self.filtered("sell_only_by_packaging").product_variant_ids.ids
+        if not ids_to_check:
+            return []
+        query = """
```
rousseldenis: Is a search() with that limited domain (ids and product_packaging_id) has really poor performances to do it through query ?

simahawk: security is checked already by accessing the variants, IMO there's no need to rely on ORM if we can get this faster.
I could go for a search_read but it does not allow to pass `load="_classic_write"` to fetch only prod ids.

sebalix: About security, maybe a comment is welcome to explain this?

simahawk: I've removed the SQL to keep it clean.
Can you update your review?

sebalix (approved):

TDu (approved):

PR #1664, opened 2021-07-21
[14.0][FIX] incompatible modules

nilshamerlinck:
`sale_isolated_quotation` and `sale_quotation_number` are incompatible (both define a sequence with same code `sale.quotation`): they should not be installed nor tested together

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,18 +18,29 @@ stages:
 
 jobs:
   include:
+    # Test separately: sale_partner_version
     - stage: test
       env:
-        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT="1" EXCLUDE="sale_partner_version"
+        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT=1 INCLUDE="sale_partner_version"
     - stage: test
       env:
-        - TESTS=1 ODOO_REPO="OCA/OCB" EXCLUDE="sale_partner_version"
+        - TESTS=1 ODOO_REPO="OCA/OCB" INCLUDE="sale_partner_version"
+    # Test separately: sale_quotation_number
     - stage: test
       env:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -20,10 +20,10 @@ jobs:
   include:
     - stage: test
       env:
-        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT="1"
+        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT="1" EXCLUDE="sale_quotation_number"
```
simahawk: is this relevant? It's only for generating POT no?

.travis.yml
```diff
@@ -20,10 +20,10 @@ jobs:
...
     - stage: test
       env:
-        - TESTS=1 ODOO_REPO="OCA/OCB"
+        - TESTS=1 ODOO_REPO="OCA/OCB" EXCLUDE="sale_isolated_quotation"
```
simahawk: don't we need one that excludes `sale_quotation_number`?

simahawk: I would still change the name of the sequence and avoid running tests separately

.travis.yml
pedrobaeza: It makes both things: test with `odoo/odoo` and generate the POT, but anyways, you need to generate POT for all modules, so each EXCLUDE/INCLUDE should have both lines.

nilshamerlinck:
Hi @simahawk @pedrobaeza
- I updated the PR to properly use the _rebel_module_groups_ setting of the repo template
- we are being bitten by this on some PRs again, would you mind merging it?

.travis.yml
nilshamerlinck: right, fixed

.travis.yml
nilshamerlinck: but beyond the sequence, they are overlapping functionally so testing them together could lead to other issues in the future

pedrobaeza (approved):

simahawk (approved):

PR #1665, opened 2021-07-22
[13.0] sale_discount_display_amount: fix init hook

simahawk:
Prevent failure if new columns are already in DB.

Backport of https://github.com/OCA/sale-workflow/pull/1651
Not needed in v13 because of the clause "IF NOT EXISTS" but this version is more compact and it uses a core util which annotates columns too.


skeller1 (approved):
LGTM :+1: tested local

PR #1669, opened 2021-07-23
[14.0][ADD] sale_order_serial

patrickrwilson:
Work In Progress:

This adds serial order numbers to sale order lines.


patrickrwilson:
Thanks @Chanakya-OSI, can you please squash commits?

rousseldenis:
@patrickrwilson @Chanakya-OSI Could you explain a little bit more in description what this module does as I see mixed stock.production.lot and ir.sequence and it's a little bit confusing ? Thanks

Chanakya-OSI:
> @patrickrwilson @Chanakya-OSI Could you explain a little bit more in description what this module does as I see mixed stock.production.lot and ir.sequence and it's a little bit confusing ? Thanks

This module adds the ability to create serial numbers on sale order lines. On sale order line, the field 'serial_sequence_id' will display only those sequences with product_serial_ok=True. And it will generate the serial list based on the selected sequence.

patrickrwilson:
@rousseldenis 
Hi, to add to @Chanakya-OSI comment, we needed a way to specify serial numbers at the sales order level that can be used throughout the rest of the processes so this is the first module to do so. Not only does this add the ability to have 1 or more serial numbers on the sales order line but also to specify a specific sequence to generate those serial numbers. The sequence can be based on the customer or the product. 
We originally looked at the sale_order_lot_selection module to see if that was usable in this circumstance but decided we needed something more specific to serial numbers as it makes sense that a group of products could belong to a lot but serial numbers are more specific per unit. (example a sales order line can have qty 6 therefore we need six serial numbers for that same line).
So what this module does is provides the ability to generate serial numbers for the number of items on a sales order line using a specific sequence.
Does that answer your question?

rousseldenis:
> @rousseldenis Hi, to add to @Chanakya-OSI comment, we needed a way to specify serial numbers at the sales order level that can be used throughout the rest of the processes so this is the first module to do so. Not only does this add the ability to have 1 or more serial numbers on the sales order line but also to specify a specific sequence to generate those serial numbers. The sequence can be based on the customer or the product. We originally looked at the sale_order_lot_selection module to see if that was usable in this circumstance but decided we needed something more specific to serial numbers as it makes sense that a group of products could belong to a lot but serial numbers are more specific per unit. (example a sales order line can have qty 6 therefore we need six serial numbers for that same line). So what this module does is provides the ability to generate serial numbers for the number of items on a sales order line using a specific sequence. Does that answer your question?

Ok, understood.

Could you add the explanation to DESCRIPTION ?

rousseldenis:
> @rousseldenis Hi, to add to @Chanakya-OSI comment, we needed a way to specify serial numbers at the sales order level that can be used throughout the rest of the processes so this is the first module to do so. Not only does this add the ability to have 1 or more serial numbers on the sales order line but also to specify a specific sequence to generate those serial numbers. The sequence can be based on the customer or the product. We originally looked at the sale_order_lot_selection module to see if that was usable in this circumstance but decided we needed something more specific to serial numbers as it makes sense that a group of products could belong to a lot but serial numbers are more specific per unit. (example a sales order line can have qty 6 therefore we need six serial numbers for that same line). So what this module does is provides the ability to generate serial numbers for the number of items on a sales order line using a specific sequence. Does that answer your question?

Ok, understood.

Could you add the explanation to DESCRIPTION ?

mcassuto:
@patrickrwilson It looks like your module is improving the 2 other existing modules :
* sale_order_lot_generator by adding the generation by sequence on top of it
* sale_order_lot_selection by allowing to add multiple S/N on a order line (use case is a product tracked by serial, and we sell more than one)

Then, I guess you should improve the 2 existing module instead of creating a new module which contains these 2 features.

PR #1283, opened 2020-10-09
[13.0] [MIG] sale_automatic_workflow_payment_mode

ThomasBinsfeld:

PR #1183, opened 2020-07-07
[MIG] sale_automatic_workflow_payment_mode: Migration to 13.0

captivea-dcr:

PR #1080, opened 2020-02-27
[MIG][13.0] sale_automatic_workflow_payment_mode

SodexisTeam:

PR #1673, opened 2021-07-27
[13.0][MIG] sale_automatic_workflow_payment_mode: Migration to 13.0

victoralmau:
Migration to 13.0

Please @joao-p-marques and @pedrobaeza can you review it?

@Tecnativa TT31208


pedrobaeza:
There are already 3 opened PRs for this module:

#1283
#1183
#1080 

But checking them, they seems not finished, so closing them.

pedrobaeza:
Please check the reviews in #1283

bizzappdev:
> Please check the reviews in #1283

Not sure what is the reason, But I have already solved those review changes in #1633

pedrobaeza:
@bizzappdev as you don't follow the proper title scheme for the PR, they are not findable through general search. Please follow such scheme in all your PRs, and you will avoid surprises, but I'm OK to switch to review that one, giving it priority and see if we finally get this merged.

bizzappdev:
@pedrobaeza  Extremely sorry for now the following standard.

victoralmau:
Sorry, can you check it to mergue and unblock other PRs that depend on this one?

bizzappdev:
> Sorry, can you check it to mergue and unblock other PRs that depend on this one?

sorry. Do I have to do anything?

sale_automatic_workflow_payment_mode/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Sodexis <dev@sodexis.com>
```
joao-p-marques, line 2: You can add Tecnativa here with your name :)

Also, you can re-generate the README if you change it.

joao-p-marques (approved):
Code review :+1: 

Just a small suggestion, but non-blocking.

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
```diff
@@ -7,8 +7,10 @@
 <odoo noupdate="1">
     <record id="automatic_workflow_payment_filter" model="ir.filters">
         <field name="name">Automatic Workflow Payment Filter</field>
-        <field name="model_id">account.invoice</field>
-        <field name="domain">[('state', '=', 'open')]</field>
+        <field name="model_id">account.move</field>
```
pedrobaeza, line 10: This field is a many2one:

```suggestion
<field name="model_id" ref="account.model_account_move" />
```

pedrobaeza, line 10: And this requires migration script.

pedrobaeza (changes requested):
There are other references to `account.invoice`, which disappears in this version.

sale_automatic_workflow_payment_mode/data/automatic_workflow_data.xml
victoralmau: I am checking v13 and v14 in runbot and the `model_id` field in `ir.filters` model is a selection field.

pedrobaeza: Oh, sorry, I have mixed it with `ir.rule`.  The name of the field doesn't also help too much.

victoralmau:
> There are other references to `account.invoice`, which disappears in this version.

Thanks, changes done.

pedrobaeza:
I still see references to `account.invoice` model.

sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class AutomaticWorkflowJob(models.Model):
+    _inherit = "automatic.workflow.job"
+
+    @api.model
+    def run_with_workflow(self, sale_wkf):
+        workflow_domain = [("workflow_process_id", "=", sale_wkf.id)]
+        res = super(AutomaticWorkflowJob, self).run_with_workflow(sale_wkf)
+        if sale_wkf.register_payment:
```
pedrobaeza: I don't get how this is related with `payment_mode`. This is something standard...

victoralmau: IMO, it is not necessary to add anything extra here since for each payment mode a `sale.workflow.process` is defined, and therefore in the onchange of the `payment_mode_id` of the order, you have already defined the correct workflow; therefore, when this action is executed, it is no longer necessary to take into account the payment mode.

pedrobaeza: The question is that this is registering a payment, which is something standard, not linked with `payment.mode`. Why is this code here? I understand this should go in the main module if needed, and if it's modifying a bit such registering, then it should be done by a hook on main module + the extension method here.

victoralmau: Ok, I understand.
I have added the `_register_payments()` function to the `sale_automatic_workflow` and also, in `sale_automatic_workflow_payment_mode` I have added only the necessary related to the payment mode.

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -140,6 +140,35 @@ def _sale_done(self, sale_done_filter):
...
 
+    def _prepare_dict_account_payment(self, invoice):
+        partner_type = (
+            invoice.type in ("out_invoice", "out_refund") and "customer" or "supplier"
+        )
+        return {
+            "invoice_ids": [(6, 0, invoice.ids)],
+            "amount": invoice.residual,
+            "payment_date": fields.Date.context_today(self),
+            "communication": invoice.reference or invoice.number,
```
pedrobaeza: `invoice.number` doesn't exist anymore. It should be `invoice.name` for customer invoices, and `invoice.reference` for vendor bills.

sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
```diff
@@ -24,41 +22,22 @@ def run_with_workflow(self, sale_wkf):
...
+
+    def _register_payment_invoice(self, invoice):
+        if not invoice.payment_mode_id.fixed_journal_id:
+            _logger.debug(
+                "Unable to Register Payment for invoice %s: "
+                "Payment mode %s must have fixed journal",
+                invoice.id,
+                invoice.payment_mode_id.id,
+            )
+        else:
```
pedrobaeza: No need of the else, and put `return super...`

sale_automatic_workflow_payment_mode/views/sale_workflow_process_view.xml
```diff
@@ -25,15 +25,15 @@ The licence is in the file __openerp__.py -->
                         name="payment_filter_domain"
```
pedrobaeza: This should go also to the main module.

pedrobaeza (changes requested):
You have made a mix on the 2 commits, as you removing the code for registering payments in the migration commit. Better to do everything in the same migration commit, stating on the commit message what's going on.

sale_automatic_workflow_payment_mode/views/sale_workflow_process_view.xml
victoralmau: Ok, some fields have been moved to the main addon `sale_automatic_workflow`.

sale_automatic_workflow/models/sale_workflow_process.py
```diff
@@ -115,3 +115,17 @@ def _default_filter(self, xmlid):
             "sale_automatic_workflow.automatic_workflow_sale_done_filter"
         ),
     )
+
+    @api.model
```
pedrobaeza: Don't put this method in the middle of the field definition.

pedrobaeza (approved):

PR #1677, opened 2021-07-30
[13.0][IMP] sale_automatic_workflow: Send order confirmation mail option in Sale Workflow Process

victoralmau:
Send order confirmation mail option in Sale Workflow Process.

@Tecnativa TT30759

### DIFF
--- a/sale_automatic_workflow/data/automatic_workflow_data.xml
+++ b/sale_automatic_workflow/data/automatic_workflow_data.xml
@@ -44,6 +44,7 @@
         <field name="name">Automatic</field>
         <field name="picking_policy">one</field>
         <field name="validate_order" eval="1" />
+        <field name="send_order_confirmation_mail" eval="1" />
         <field name="order_filter_id" eval="automatic_workflow_order_filter" />
         <field name="create_invoice" eval="1" />
         <field
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -54,6 +54,20 @@ class AutomaticWorkflowJob(models.Model):
         sale.action_confirm()
         return "{} {} confirmed successfully".format(sale.display_name, sale)
 
+    def _do_send_order_confirmation_mail(self, sale):
+        """Send order confirmation mail, while filtering to make sure the order is
+        confirmed with _do_validate_sale_order() function"""
+        if not self.env["sale.order"].search_count(
+            [("id", "=", sale.id), ("state", "=", "sale")]
+        ):
+            return "{} {} job bypassed".format(sale.display_name, sale)
+        if sale.user_id:
+            sale = sale.with_user(sale.user_id)
+        sale._send_order_confirmation_mail()
+        return "{} {} send order confirmation mail successfully".format(
+            sale.display_name, sale
+        )
+
     @api.model
     def _validate_sale_orders(self, order_filter):
         sale_obj = self.env["sale.order"]
@@ -62,6 +76,8 @@ class AutomaticWorkflowJob(models.Model):
         for sale in sales:
             with savepoint(self.env.cr), force_company(self.env, sale.company_id):
                 self._do_validate_sale_order(sale, order_filter)
+                if self.env.context.get("send_order_confirmation_mail"):
+                    self._do_send_order_confirmation_mail(sale)
 
     def _do_create_invoice(self, sale, domain_filter):
         """Create an invoice for a sales order, filter ensure no duplication"""
@@ -178,7 +194,9 @@ class AutomaticWorkflowJob(models.Model):
     def run_with_workflow(self, sale_workflow):
         workflow_domain = [("workflow_process_id", "=", sale_workflow.id)]
         if sale_workflow.validate_order:
-            self._validate_sale_orders(
+            self.with_context(
+                send_order_confirmation_mail=sale_workflow.send_order_confirmation_mail
+            )._validate_sale_orders(
                 safe_eval(sale_workflow.order_filter_id.domain) + workflow_domain
             )
         if sale_workflow.validate_picking:
--- a/sale_automatic_workflow/views/sale_workflow_process_view.xml
+++ b/sale_automatic_workflow/views/sale_workflow_process_view.xml
@@ -45,6 +45,7 @@
                             />
                         </div>
                     </div>
+                    <field name="send_order_confirmation_mail" colspan="4" />
                     <field name="validate_picking" />
                     <label
                         for="picking_filter_id"
--- a/sale_automatic_workflow/README.rst
+++ b/sale_automatic_workflow/README.rst
@@ -41,6 +41,7 @@ A workflow can:
 - Apply automatic actions:
 
   * Validate the order (only if paid, always, never)
+  * Send order confirmation mail (only when order confirmed)
   * Create an invoice
   * Validate the invoice
   * Confirm the picking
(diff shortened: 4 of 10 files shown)


sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -144,7 +165,10 @@ def _sale_done(self, sale_done_filter):
     def run_with_workflow(self, sale_workflow):
         workflow_domain = [("workflow_process_id", "=", sale_workflow.id)]
         if sale_workflow.validate_order:
-            self._validate_sale_orders(
+            _self = self.with_context(
+                send_order_confirmation_mail=sale_workflow.send_order_confirmation_mail
+            )
+            _self._validate_sale_orders(
```
chienandalu: Why not doing it in a single statement? Isn't it a little bit confusing this way?

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -54,6 +54,25 @@ def _do_validate_sale_order(self, sale, domain_filter):
...
+        Prevent mail send dupplication according to odoo
+        /addons/website_sale/models/sale_order.py#L357"""
+        if not self.env["sale.order"].search_count(
+            [("id", "=", sale.id)] + domain_filter
+        ):
+            return "{} {} job bypassed".format(sale.display_name, sale)
+        if (
+            sale.state == "sale"
+            and not sale.transaction_ids
+            and not sale.amount_total
```
chienandalu: Not amount total?

chienandalu (commented):
A test would be nice as well

sale_automatic_workflow/models/automatic_workflow_job.py
victoralmau: Although it would be a good idea, I think it is not possible because there is no field similar to `send_confirmation_mail_date` to help us not to send email several times.
Can you think of a way to do it?

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -144,7 +165,10 @@ def _sale_done(self, sale_done_filter):
     def run_with_workflow(self, sale_workflow):
         workflow_domain = [("workflow_process_id", "=", sale_workflow.id)]
         if sale_workflow.validate_order:
-            self._validate_sale_orders(
+            _self = self.with_context(
+                send_order_confirmation_mail=sale_workflow.send_order_confirmation_mail
+            )
+            _self._validate_sale_orders(
                 safe_eval(sale_workflow.order_filter_id.domain) + workflow_domain
             )
```
chienandalu: Well I meant somthing like
```suggestion
            self.with_context(
                send_order_confirmation_mail=sale_workflow.send_order_confirmation_mail
            )._validate_sale_orders(
                safe_eval(sale_workflow.order_filter_id.domain) + workflow_domain
            )
```

victoralmau:
Thanks @chienandalu , changes done.

chienandalu (approved):

sale_automatic_workflow/data/automatic_workflow_data.xml
```diff
@@ -44,6 +44,7 @@
         <field name="name">Automatic</field>
         <field name="picking_policy">one</field>
         <field name="validate_order" eval="1" />
+        <field name="send_order_confirmation_mail" eval="1" />
```
joao-p-marques, line 47: As this is a new optional feature being introduced, wouldn't it be better to disable it by default?
```suggestion
        <field name="send_order_confirmation_mail" eval="0" />
```

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -54,6 +54,25 @@ def _do_validate_sale_order(self, sale, domain_filter):
         sale.action_confirm()
         return "{} {} confirmed successfully".format(sale.display_name, sale)
 
+    def _do_send_order_confirmation_mail(self, sale):
+        """Send order confirmation mail, filter ensure no duplication.
+        Prevent mail send dupplication according to odoo
+        /addons/website_sale/models/sale_order.py#L357"""
```
joao-p-marques:
```suggestion
        """Send order confirmation mail, while filtering to avoid duplication if the message was already sent in odoo/addons/website_sale/models/sale_order.py#L357"""
```

Also, the reference is not very resilient to change, so maybe you could reference the actual method, model and module instead...

joao-p-marques (commented):
Code review

sale_automatic_workflow/data/automatic_workflow_data.xml
victoralmau: IMO it is not necessary to leave it at 0 by default because this is noupdate=1

joao-p-marques: Ok, makes sense 👍 Not blocking this, then

sale_automatic_workflow/models/automatic_workflow_job.py
joao-p-marques: I still don’t see this change… Did you push it?

joao-p-marques (approved):

sale_automatic_workflow/views/sale_workflow_process_view.xml
```diff
@@ -19,6 +19,7 @@
                 </group>
                 <group name="workflow_options" string="Workflow Options" col="4">
                     <field name="validate_order" />
+                    <field name="send_order_confirmation_mail" />
```
manuelcalerosolis: If you place the field here, the screen is completely out of square.

This is how it looked before applying this PR. 

image

After applying the PR.

image

I think it is best to put it before field name = "validate_picking" and with colspan = "4".

'field name="send_order_confirmation_mail" colspan="4"'

image

manuelcalerosolis (changes requested):

manuelcalerosolis:
The emails that are sent do not take the correct format and look bad, this is because the sender is Odoobot and he is not logged in and some resources necessary for printing are not available without the session.sid 

image

victoralmau:
It's ready to merge, right?

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -54,6 +54,27 @@ def _do_validate_sale_order(self, sale, domain_filter):
...
 
+    def _do_send_order_confirmation_mail(self, sale):
+        """Send order confirmation mail, while filtering to avoid duplication if the
+        message was already sent in in action_confirm() function from sale.order model
+        in website_sale"""
+        if not self.env["sale.order"].search_count(
+            [("id", "=", sale.id), ("state", "=", "sale")]
+        ):
+            return "{} {} job bypassed".format(sale.display_name, sale)
+        if (
```
pedrobaeza: Why do you need this extra check?

victoralmau: You can probably reduce this condition a bit, but it is necessary to maintain it in order to avoid that an email is sent 2 times in the event that the order has been self-confirmed because the transaction has been done.

Thinking about it carefully, this function only executes when trying to validate an order according to a specified domain; Therefore, if you have reached this point, the email should always be sent because if it were a self-confirmed e-commerce order (transaction done) it would not have reached here so this check could probably be removed.

What do you think about it?

pedrobaeza: You tell me then...

victoralmau: Now it's ready.

pedrobaeza (approved):

PR #1679, opened 2021-08-02
14.0 mig sale invoice policy

kimirondon:


sale_invoice_policy/models/product_template.py
```diff
@@ -28,6 +27,14 @@ def _default_default_invoice_policy(self):
         inverse="_inverse_invoice_policy",
     )
 
+    @api.model
+    def default_get(self, fields_list):
```
rousseldenis, line 31: @kimirondon Why does the default lambda not work ?

rousseldenis:
Runbot is complaining about these : 

image


Could you check ?

sale_invoice_policy/models/product_template.py
kimirondon: Hello @rousseldenis, sorry for my lateness!, about the default, i had an issue with the method _reflect_selections,
but i believe #1762  will be merge soon.

rousseldenis: @kimirondon If you want to review that one and close this.

PR #1600, opened 2021-05-27
[14.0] mig product_supplierinfo_for_customer_sale

fredzamoabg:
**product_supplierinfo_for_customer_sale** module migration from 12.0
depends on product_supplierinfo_for_customer module migration https://github.com/OCA/product-attribute/pull/859

PR #1680, opened 2021-08-03
[14.0][MIG] product_supplierinfo_for_customer_sale

LoisRForgeFlow:
Migration to v14.

Supersedes https://github.com/OCA/sale-workflow/pull/1600 because this one is aligned with version 13.

@ForgeFlow


product_supplierinfo_for_customer_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @api.onchange("product_id")
+    def product_id_change(self):
+        result = super(SaleOrderLine, self).product_id_change()
+        for line in self.filtered(
+            lambda sol: sol.product_id.product_tmpl_id.customer_ids
+            and sol.order_id.pricelist_id.item_ids
+        ):
+            product = line.product_id
+            items = self.env["product.pricelist.item"].search(
```
rousseldenis, line 37: @LoisRForgeFlow Would it be possible to do the search outside the loop ?

```suggestion
            lines = self.filtered(...)
            items.search([("pricelist_id", "in", lines.mapped("order_id.princelist_id.id") (...) ])
            
            for line in ...
```

LoisRForgeFlow: It is an onchange method, 99% of the times will be called with one line. So, I would prefer to stick to a strict migration and do not include that change.

rousseldenis (approved):

RaulOForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
@rousseldenis Good to merge here?

PR #1687, opened 2021-08-06
14.0 mig sale_delivery_split_date

fredzamoabg:
sale_delivery_split_date module migration from 12.0.

depends on https://github.com/OCA/sale-workflow/pull/1689


stevech091 (approved):
functional review ok

sale_delivery_split_date/models/stock_picking.py
```diff
@@ -0,0 +1,21 @@
...
+    min_dt = fields.Date(
+        string="Scheduled Date (for filter purpose only)",
+        compute="_compute_min_dt",
+        store=True,
+    )
+
+    @api.depends("scheduled_date")
+    def _compute_min_dt(self):
+        for picking in self:
+            min_dt = fields.Date.from_string(picking.scheduled_date)
```
SimoRubi: Please remove `from_string` because it is deprecated: 
>  kept for backwards compatibility, but consider `from_string` as deprecated, will probably
    be removed after V12 

(from https://github.com/odoo/odoo/blob/de81cc386d3a7f85b192a60b19477fcec8097624/odoo/fields.py#L1811-L1813)

In the latest versions (after `12.0` I think) the value of Date/Datetime fields are already `date`/`datetime` objects respectively, is this conversion still really needed?

sale_delivery_split_date/models/stock_picking.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    min_dt = fields.Date(
```
SimoRubi: Is this field still needed? Why don't we directly use `scheduled_date`?
I see the definition of this field is almost unchanged from v8, but in that time the `date` field was not stored so this new `stored` field would have been needed _for filter purpose_.
Now `scheduled_date` is stored so it would be possible to use that for filters; what do you think?

SimoRubi (changes requested):

sale_delivery_split_date/models/sale_order_line.py
```diff
@@ -21,3 +19,9 @@ def _get_procurement_group_key(self):
                 # group by date instead of datetime
                 return (priority, self.commitment_date.date())
         return key
+
+    def _prepare_procurement_values(self, group_id=False):
+        vals = super(SaleOrderLine, self)._prepare_procurement_values(group_id)
```
SimoRubi: Please assign keyword arguments as such, here it should be `group_id=group_id`.
Otherwise if another argument `foo` is added before `group_id` in the method's signature like:
```
def _prepare_procurement_values(self, foo, group_id=False):
```
This code would assign the value `group_id` to `foo` method parameter in `super` instead of to `group_id`.

sale_delivery_split_date/models/sale_order_line.py
```diff
@@ -21,3 +19,9 @@ def _get_procurement_group_key(self):
...
+        if self.commitment_date:
+            vals.update({"date_planned": self.commitment_date})
```
SimoRubi, line 26: I see in the previous method that the priority is checked before using `commitment_date`, why aren't we checking priority here too?

Also, why is this new method needed in `14.0`?

SimoRubi (changes requested):

sale_delivery_split_date/models/sale_order_line.py
fredzamoabg: I did not received this from functional, @stevech091 what do you think about this?

SimoRubi: Investigating more, I found out that it makes sense not to manage priority here because it is only used to group procurements; when they are in the process of being created it does no more make sense to manage priority.

As for **why**: elaborating what you wrote to me through other channels, the motivation is to have the created pickings with a `scheduled_date` equal to the line's `commitment_date`.

This has to go through `date_planned` because:

1. The picking's `scheduled_date` is computed based on its moves' `date` in https://github.com/odoo/odoo/blob/c5de12310be023e1048f8a1174f87582c13a9160/addons/stock/models/stock_picking.py#L509-L516:
    ```
    @api.depends('move_lines.state', 'move_lines.date', 'move_type')
    def _compute_scheduled_date(self):
        for picking in self:
            moves_dates = picking.move_lines.filtered(lambda move: move.state not in ('done', 'cancel')).mapped('date')
            if picking.move_type == 'direct':
                picking.scheduled_date = min(moves_dates, default=picking.scheduled_date or fields.Datetime.now())
            else:
                picking.scheduled_date = max(moves_dates, default=picking.scheduled_date or fields.Datetime.now())
    ```
2. The `date` in the moves generated by a procurement is set based on the procurement's `date_planned` in https://github.com/odoo/odoo/blob/710f228732b79a75ded7912cbe91fc43dec7795b/addons/stock/models/stock_rule.py#L278-L313:
    ```
        date_scheduled = fields.Datetime.to_string(
            fields.Datetime.from_string(values['date_planned']) - relativedelta(days=self.delay or 0)
        )
    [...]
            'date': date_scheduled,
    ```

In the `super`'s computation of `date_planned` and the closely related `date_deadline`:
```
        date_deadline = self.order_id.commitment_date or (self.order_id.date_order + timedelta(days=self.customer_lead or 0.0))
        date_planned = date_deadline - timedelta(days=self.order_id.company_id.security_lead)
        values.update({
            'group_id': group_id,
            'sale_line_id': self.id,
            'date_planned': date_planned,
            'date_deadline': date_deadline,
```
(from https://github.com/odoo/odoo/blob/4b36dc25929b8aeac03b3a93c043b23b89141d94/addons/sale_stock/models/sale_order.py#L512-L518)
there are also lead times added/subtracted from the `date_planned` but looking at the tests of this module I assume they have to be ignored.

SimoRubi (approved):

dreispt:
Red build

fredzamoabg:
> Red build

dependency is still not merged https://github.com/OCA/sale-workflow/pull/1689

fredzamoabg:
> Red build

@dreispt rebased... can you check?

PR #1688, opened 2021-08-08
[14.0][IMP] sale_order_type: add sale order type to sales and invoice analysis reports

skeller1:
- add sale order type field to account invoice and sale order report
- add sale order type as group filter to account invoice report and sale order report view (pivot,graph)

grafik

grafik

### DIFF
--- a/sale_order_type/reports/sale_report.py
+++ b/sale_order_type/reports/sale_report.py
@@ -1,0 +1,19 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    type_id = fields.Many2one(
+        comodel_name="sale.order.type",
+        string="Type",
+    )
+
+    # flake8: noqa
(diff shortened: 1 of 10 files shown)


ivantodorovich (approved):

sale_order_type/reports/sale_report.py
```diff
@@ -0,0 +1,19 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    type_id = fields.Many2one(
```
bodedra, line 9: may we change field name from `type_id` to `sale_type_id`? Same as you implemented for Invoice analysis report.

ivantodorovich: IMO it's ok like this as it's the same name in the sale.order model 🤷‍♂️

bodedra: Ah. Got it. :+1:

bodedra (approved):
Functional review. :+1: 

Sale and Invoice analysis report.

PR #1691, opened 2021-08-09
[13.0][ADD] sale_exception_warning: Adds exceptions in sales orders when the product has defined warning or partner has sale warning

victoralmau:
Adds exceptions in sales orders when the product has defined warning or partner has sale warning.

@Tecnativa TT30756


pedrobaeza:
You should respect previous authorship, even including 8.0 commit.

PR #1692, opened 2021-08-09
[13.0][IMP] sale_exception: Add customer sale warning + product warning exceptions

victoralmau:
Add customer sale warning + product warning exceptions.

@Tecnativa TT30756

### DIFF
--- a/sale_exception/README.rst
+++ b/sale_exception/README.rst
@@ -37,6 +37,59 @@ found in Odoo, wrong line format etc.)
 .. contents::
    :local:
 
+Configuration
+=============
+
+If you are going to use Customer sale warning and Product warning,
+for setting corresponding information, you need to:
+
+#. Go to *Settings > User & Companies > Users*.
+#. Edit your user.
+#. Check "A warning can be set on a product or a customer (Sale)" group.
+#. Install sale_management addon.
+
+Usage
+=====
+
+Not Enough Virtual Stock:
+#. Go to *Sales > Products > Products*.
+#. Create new storable product without stock available.
+#. Go to *Sales > Orders > Quotations*
+#. Create new quotation.
+#. Add product without stock available.
+#. An exception will be displayed.
+
+No ZIP code on destination:
+#. Go to *Contacts*.
+#. Edit or create new contact.
+#. Set empty zip code.
+#. Go to *Sales > Orders > Quotations*
+#. Create new quotation.
+#. Set delivery address with no zip code.
+#. An exception will be displayed.
+
+Product warning:
+#. Go to *Sales > Products > Products*.
+#. Edit or create new product.
+#. Go to *Sales* tab.
+#. Set your desired warning option under the *Warning when Selling this Product* group.
+#. Set some warning message.
+#. Go to *Sales > Orders > Quotations*
+#. Create new quotation.
+#. Add product with warning message.
+#. An exception will be displayed.
+
+Partner warning:
+#. Go to *Contacts*.
+#. Edit or create new contact.
+#. Go to *Internal notes* tab.
+#. Set warning option according to *Warning on the Sales Order* group.
+#. Set some warning message.
+#. Go to *Sales > Orders > Quotations*
+#. Create new quotation.
+#. Set partner with warning message.
+#. An exception will be displayed.
+
 Bug Tracker
 ===========
 
--- a/sale_exception/data/sale_exception_data.xml
+++ b/sale_exception/data/sale_exception_data.xml
@@ -18,7 +18,7 @@
         <field name="description">No ZIP code on destination</field>
         <field name="sequence">50</field>
         <field name="model">sale.order</field>
-        <field name="code">if not obj.partner_shipping_id.zip: failed=True</field>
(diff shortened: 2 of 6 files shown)


sale_exception/README.rst
```diff
@@ -37,6 +37,42 @@ found in Odoo, wrong line format etc.)
...
    :local:
 
+Configuration
+=============
+
+To configure this module you need to:
+
+#. Go to *Settings > User & Companies > Users*.
+#. Edit your user.
+#. Check "A warning can be set on a product or a customer (Sale)" group.
```
pedrobaeza, line 48: This is only if you want to use such exception, but not for others, so you should specify it.

sale_exception/README.rst
```diff
@@ -37,6 +37,42 @@ found in Odoo, wrong line format etc.)
...
+
+Usage
+=====
+
+Product warning:
```
pedrobaeza, line 71: Document existing exceptions then if you put the new ones.

sale_exception/data/sale_exception_data.xml
```diff
@@ -31,4 +31,22 @@
         >if obj.product_id and sale_line.product_id.type == 'product' and obj.product_id.with_context(warehouse=obj.order_id.warehouse_id.id).virtual_available &lt; obj.product_uom_qty: failed=True</field>
```
pedrobaeza: The module doesn't depend on stock, but has this exception, so it's inconsistent

pedrobaeza (changes requested):

sale_exception/readme/USAGE.rst
```diff
@@ -0,0 +1,23 @@
+Product warning:
+#. Go to *Sales > Products > Products*.
+#. Edit or create new product.
+#. Go to *Sales* tab.
+#. Set warning option according to *Warning when Selling this Product* group.
```
joao-p-marques:
```suggestion
#. Set your desired warning option under the *Warning when Selling this Product* group.
```

sale_exception/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+To configure this module you need to:
+
+#. Go to *Settings > User & Companies > Users*.
+#. Edit your user.
+#. Check "A warning can be set on a product or a customer (Sale)" group.
```
joao-p-marques, line 6: You could maybe include here that the user should install `sale_management` to use the module, although it is not strictly a dependency...

sale_exception/readme/USAGE.rst
```diff
@@ -0,0 +1,23 @@
...
+#. Set some warning message.
+#. Go to *Sales > Orders > Quotations*
+#. Create new quotation.
+#. Add product with warning message.
+#. Confirm order.
+#. An exception will be displayed.
```
joao-p-marques, line 27: From my testing, the exception is displayed when saving the quotation, not when confirming the order, so maybe change the order here to make it clearer.

joao-p-marques (commented):
Code and functional review :+1: 

Apart from @pedrobaeza's comments, just a few suggestions to improve the docs, otherwise LGTM. 

As a note, I tested locally with only this module, as I was getting some weird behavior in the OCA runbot. Is this module integrating well with other modules?

sale_exception/data/sale_exception_data.xml
victoralmau: Exactly, this exception is incorrect because it depends on stock. 
Maybe you should create a new addon called: `sale_exception_stock` with the dependency on `sale_stock` and move that exception there. What do you think?

sale_exception/data/sale_exception_data.xml
```diff
@@ -31,4 +31,22 @@
...
         <field name="active" eval="False" />
     </record>
+    <record id="exception_partner_sale_warning" model="exception.rule">
+        <field name="name">Customer sale warning</field>
+        <field name="description">The customer has a sale warning in his form</field>
+        <field name="sequence">40</field>
+        <field name="model">sale.order</field>
+        <field name="code">if obj.partner_id.sale_warn == "warning": failed=True</field>
```
pedrobaeza: Can't this be simplified to (and the same for the rest)?:
```suggestion
        <field name="code">failed=obj.partner_id.sale_warn == "warning"</field>
```

sale_exception/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,6 @@
+Only according to Customer sale warning and Product warning you need to:
```
pedrobaeza:
```suggestion
If you are going to use Customer sale warning and Product warning, 
for setting corresponding information, you need to:
```

victoralmau:
Thanks @pedrobaeza , changes done.

pedrobaeza (approved):

pedrobaeza:
As @joao-p-marques' requests are honored, I'm merging:

/ocabot merge minor

pedrobaeza:
Not sure why it's failing. Once more to try and if not, please rebase to see if it appears:

/ocabot merge minor

victoralmau:
@pedrobaeza can you try to merge again?

PR #1693, opened 2021-08-09
[14.0][ADD] sale_mail_autosubscribe

ivantodorovich:


sale_mail_autosubscribe/data/mail_autosubscribe.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
simahawk, line 1: do we really need a module only for this? :thinking: 

Looks like proper documentation in README of the base module is enough.

sale_mail_autosubscribe/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+
+{
+    "name": "Sale Mail Autosubscribe",
+    "summary": "Automatically subscribe partners to their company's sale orders",
+    "version": "14.0.1.0.0",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["ivantodorovich"],
+    "license": "AGPL-3",
+    "category": "Sale",
+    "depends": ["mail_autosubscribe", "sale"],
```
simahawk, line 13: auto_install?

simahawk (changes requested):

sale_mail_autosubscribe/data/mail_autosubscribe.xml
ivantodorovich: I asked myself the same question but then I decided to add the module because this way it's easier for non-functional people.

Imagine people installing the module from the odoo appstore, for example

sale_mail_autosubscribe/__manifest__.py
ivantodorovich: Indeed! :+1:  Done

rousseldenis (approved):

sale_mail_autosubscribe/data/mail_autosubscribe.xml
simahawk: fair enough :)

simahawk:
base PR is getting merged

ivantodorovich:
ready :tada:

simahawk (approved):

PR #1700, opened 2021-08-23
[FIX][12.0] sale_order_line_date - removing xpath preventing module installation

baimont:
@rousseldenis thanks for the solution


baimont:
> @rousseldenis thanks for the solution

in fact this is not good : we should not remove these fields from this view otherwise the module is useless

baimont:
> > @rousseldenis thanks for the solution
> 
> in fact this is not good : we should not remove these fields from this view otherwise the module is useless

With [this](https://github.com/OCA/sale-workflow/commit/0214cb59b44a79099dc428dc315f211e45839f15) it is better.

PR #1704, opened 2021-08-30
[14.0][IMP] sale_discount_display_amount: Adds hook for depends

rousseldenis:
Allows to override the different amount computations.

@chafique-delli @simahawk

### DIFF
--- a/sale_discount_display_amount/models/sale_order.py
+++ b/sale_discount_display_amount/models/sale_order.py
@@ -21,7 +21,11 @@ class SaleOrder(models.Model):
         store=True,
     )
 
-    @api.depends("order_line.discount_total", "order_line.price_total_no_discount")
+    @api.model
+    def _get_compute_discount_total_depends(self):
+        return ["order_line.discount_total", "order_line.price_total_no_discount"]
+
+    @api.depends(lambda self: self._get_compute_discount_total_depends())
     def _compute_discount_total(self):
         for order in self:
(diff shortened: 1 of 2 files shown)


sale_discount_display_amount/models/sale_order.py
```diff
@@ -21,7 +21,11 @@ class SaleOrder(models.Model):
         store=True,
     )
 
-    @api.depends("order_line.discount_total", "order_line.price_total_no_discount")
+    @api.model
+    def _get_compute_discount_total_domain(self):
```
simahawk:
```suggestion
    def _compute_discount_total_depends(self):
```

sale_discount_display_amount/models/sale_order_line.py
```diff
@@ -41,8 +41,12 @@ def _update_discount_display_fields(self):
                 }
             )
 
-    @api.depends("product_uom_qty", "discount", "price_unit", "tax_id")
+    @api.model
+    def _get_compute_amount_domain(self):
```
simahawk:
```suggestion
    def _compute_amount_depends(self):
```

simahawk (changes requested):

simahawk (approved):
not blocking indeed. Pre-approving anyway ;)

rousseldenis:
@simahawk Done

ivantodorovich (approved):

PR #1708, opened 2021-09-01
[14.0] [NEW] sale_product_approval_mrp: New module to handle product approval workflow for MRP.

Chandresh-OSI:
Depends on #1647


patrickrwilson:
@Chandresh-OSI I went to test this but get the following error when installing, please fix:
Element '<div name="options">' cannot be located in parent view

View name: product_state_form_view_inherit_mrp
Error context:
 view: ir.ui.view(1594,)
 xmlid: product_state_form_view_inherit_mrp
 view.model: product.state
 view.parent: ir.ui.view(1230,)
 file: c:\odoo\odoo 14.0e\server\odoo\addons\sale_product_approval_mrp\views\product_state.xml

dreispt:
@Chandresh-OSI Notice that all checks are red, even pre-commit. So this needs fixing. Thanks!

patrickrwilson:
@Chandresh-OSI I don't think this is functioning correctly. I created a simple product that can be manufactured and if the state doesn't allow any manufacturing I am still able to create an new MO and add the product. It also allows me to confirm, mark as done and perform all the workflows even though the exception shows and the override exception is false.

dreispt:
@Nikul-OSI pre-commit is failing!

dreispt:
@patrickrwilson The new commit removes `mrp_order_state` to fix a traceback error `precision_rounding must be positive` when adding a Component to a confirmed MO.
I have no idea why that field is causing that, but I could reproduce the error as soon as the field was added to the view 🙅🏼

PR #1715, opened 2021-09-10
[IMP] 13.0: Extract delivery date logic to specific methods

mmequignon:
Add a few hooks to ease overrides.

### DIFF
--- a/sale_cutoff_time_delivery/models/sale_order.py
+++ b/sale_cutoff_time_delivery/models/sale_order.py
@@ -58,6 +58,9 @@ class SaleOrderLine(models.Model):
     def _prepare_procurement_values(self, group_id=False):
         """Postpone delivery according to cutoff time"""
         res = super()._prepare_procurement_values(group_id=group_id)
+        return self._sale_cutoff_time_delivery_prepare_procurement_values(res)
+
+    def _sale_cutoff_time_delivery_prepare_procurement_values(self, res):
         date_planned = res.get("date_planned")
         if not date_planned:
             return res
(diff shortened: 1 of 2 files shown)


simahawk (approved):

simahawk:
note: the version in the commit msg is not relevant IMO, it woudl be more interesting to see module names instead (shorted)

sebalix (approved):

sebalix:
@mmequignon I agree with @simahawk , commit messages should not mention a version (when the commit is ported it is not relevant)

sale_cutoff_time_delivery/models/sale_order.py
```diff
@@ -58,6 +58,9 @@ class SaleOrderLine(models.Model):
     def _prepare_procurement_values(self, group_id=False):
         """Postpone delivery according to cutoff time"""
         res = super()._prepare_procurement_values(group_id=group_id)
+        return self._sale_cutoff_time_delivery_prepare_procurement_values(res)
```
TDu, line 61: Looking at the function below nothing will be returned from `_prepare_procurement_values` if there is a  `date_planned` in res.
This could lead to issues, maybe ?

mmequignon: There's actually a trick here, you can click on this button, which will unfold the code that hadn't changed :smile: 
image
Then, you'll see what happens when `date_planned` is set ^o^

TDu: :rofl: I need to find my :eyeglasses:

PR #1718, opened 2021-09-23
[14.0] [ADD] sale_product_approval_stock: Added new module to handle delivery order product workflow.

Chandresh-OSI:
Depends on #1647


patrickrwilson:
@Chandresh-OSI during testing I noticed two things on the delivery order.
1. Can you change the wording on the exception activity to say 'Deliverable' instead of 'Saleable' where is says the product is marked as not saleable.
2. After the delivery order is created and you move the product to a stage that doesn't allow shipping, it seems to generate two exceptions on the DO. Can you make it only create one.
image

sale_product_approval_stock/models/product_template.py
```diff
@@ -0,0 +1,45 @@
...
+    @api.depends("candidate_ship", "product_state_id.approved_ship")
+    def _compute_ship_ok_product(self):
+        for product in self:
+            if product.product_state_id:
+                product.ship_ok = (
+                    product.candidate_ship and product.product_state_id.approved_ship
+                )
+
+
+class ProductProduct(models.Model):
```
rousseldenis: Please split one model per file

sale_product_approval_stock/models/product_template.py
```diff
@@ -0,0 +1,45 @@
...
+    _inherit = "product.product"
+
+    def write(self, vals):
+        res = super().write(vals)
+        for product in self:
+            if not product.ship_ok:
+                pick_ids = self.env["stock.picking"].search(
```
rousseldenis: Please do the search before loop.

sale_product_approval_stock/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+from datetime import date
+
+from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    do_exceptions = fields.Boolean(
```
rousseldenis, line 13: IMHO, things have to be split as following the name of this, it should be moved in a more generic module.

sale_product_approval_stock/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+        self.check_do_exception()
+        return res
+
+    def action_confirm(self):
+        res = super().action_confirm()
+        self.check_do_exception()
+        return res
+
+    def check_do_exception(self):
+        for do in self:
```
rousseldenis, line 52:
```suggestion
        for picking in self:
```
is more explicit for reviewer

sale_product_approval_stock/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+                raise UserError(
+                    _(
+                        "You can not do Validate/Mark as Done because some products are not "
+                        "allowed in this delivery order."
+                    )
+                )
+        return True
+
+
+class StockMove(models.Model):
```
rousseldenis, line 67: Please split one model per file

sale_product_approval_stock/views/delivery_order_exception.xml
```diff
@@ -0,0 +1,32 @@
+<odoo>
+    <data>
```
rousseldenis, line 2:
```suggestion
```

rousseldenis (changes requested):
Code review

PR #1719, opened 2021-09-23
[14.0][IMP] sale_automatic_workflow: Use with_company instead of user…

rousseldenis:
….company_id


rousseldenis:
@Cedric-Pigeon @grindtildeath @acsonefho

acsonefho (approved):
Code LGTM!
But you have to be sure that into every following code (and more), `self.env.company` is used instead of `self.env.user.company_id`.

rousseldenis:
> Code LGTM!
> But you have to be sure that into every following code (and more), `self.env.company` is used instead of `self.env.user.company_id`.

Of course, but that code is in Odoo...

grindtildeath:
Looks legit :+1:

grindtildeath (approved):

PR #1733, opened 2021-10-05
[13.0][ADD] sale_order_line_delivery_state

JordiMForgeFlow:
Module based on sale_order_delivery_state, aimed at being able to have delivery state at sales order line level.


sale_order_line_delivery_state/models/sale_order_line.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+            >= 0
+        )
+
+    def _partially_delivered(self):
+        """
+        Returns True if line has qty_delivered != to 0 and < ordered
+        quantities.
+
+        If `delivery` module is installed, ignores the lines with delivery costs
```
AdriaGForgeFlow: Nitpicking: I'm not sure this comment goes here or rather in the _compute_sale_line_delivery_state function where you actually filter out delivery costs

AdriaGForgeFlow (approved):
Functional and code review :+1:

sale_order_line_delivery_state/models/sale_order_line.py
JordiMForgeFlow: Yes! I modified how the delivery lines were treated at one point and forgot to fix the comments. Thank you for spotting it! :)

AaronHForgeFlow (approved):
Functional + code :+1: 

I have a question. For services the status is not delivered, I guess this is intended right?
image

JordiMForgeFlow:
> Functional + code +1
> 
> I have a question. For services the status is not delivered, I guess this is intended right? image

Exactly! As there is no delivery expected IMO it makes more sense to have this value rather than unprocessed, which denotes that a delivery needs to be processed at some time.

AaronHForgeFlow:
Ok! I agree.

JordiMForgeFlow:
@JordiBForgeFlow @pedrobaeza could I get some reviews here? :)

pedrobaeza:
I don't use this module, but I trust in the reviewers, so

/ocabot merge nobump

PR #1734, opened 2021-10-06
[12.0][IMP]  sale_force_whole_invoiceability: allow forcing in done state too; fix invisibility domain

hbrunn:
given we can configure Odoo to move sale orders to locked (`done`) [immediately](https://github.com/OCA/OCB/blob/14.0/addons/sale/models/res_config_settings.py#L10), it makes sense to show the button also in this state and allow the function to be called - we're still protected from messing up things by the check for `invoice_count`.

### DIFF
--- a/sale_force_whole_invoiceability/models/sale_order.py
+++ b/sale_force_whole_invoiceability/models/sale_order.py
@@ -13,7 +13,7 @@ class SaleOrder(models.Model):
             raise UserError(_(
                 "Only Sales Managers are allowed to force the lines to invoice"
             ))
-        if self.state != "sale":
+        if self.state not in ("sale", "done"):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You always forget to put the module in the message commit :wink:

hbrunn:
@pedrobaeza yes, this has become so bad I was already considering a pre-commit check for that ;-)

pedrobaeza (approved):

hbrunn:
...and fixed, thanks

HaraldPanten (approved):

pankk (approved):

StefanRijnhart (approved):

chienandalu (approved):

PR #1735, opened 2021-10-07
s_prod_set_sale_by_packaging: add cron for fixing lines

simahawk:
When

* packaging cannot be sold anymore
* lines are tied to a product that can be sold only by packaging
  and have no packaging

change the packaging on the line to the 1st saleable packaging.

### DIFF
--- a/sale_product_set_sale_by_packaging/models/product_set_line.py
+++ b/sale_product_set_sale_by_packaging/models/product_set_line.py
@@ -1,7 +1,13 @@
 # Copyright 2021 Camptocamp SA
 # @author Simone Orsi <simone.orsi@camptocamp.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import logging
+from collections import defaultdict
+
 from odoo import _, api, exceptions, fields, models
+from odoo.osv import expression
+
+_logger = logging.getLogger(__name__)
 
 
 class ProductSetLine(models.Model):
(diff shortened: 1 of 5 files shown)


sebalix (approved):

sale_product_set_sale_by_packaging/models/product_set_line.py
```diff
@@ -24,3 +30,53 @@ def _check_sell_only_by_packaging_err_msg(self, lines):
...
+    def cron_check_packaging(self):
+        """Ensure lines are tied to the right packaging.
+
+        When
+
+        * packaging cannot be sold anymore
+        * lines are tied to a product that can be sold only by packaging
+          and have no packaging
+
+        change the packaging on the line to the 1st saleable packaging.
```
rousseldenis, line 44: @simahawk Isn't this behaviour too intrusive related to first behaviour described here :

https://github.com/OCA/sale-workflow/blob/13.0/sale_product_set_sale_by_packaging/README.rst

Shouldn't be optional (even if the cron configuration is an option for instance), maybe on sale order level ('Auto-updateable packagings' or sth like that)

simahawk: If you want to make it optional you can turn off the cron. To me it's more important to keep data consistent rather than rely on user action. Moreover, this is not changing a line on a sale order. The user will be able to change it on product set or on sale order before confirming anyway.

rousseldenis (approved):

PR #1696, opened 2021-08-13
[WIP][DO NOT MERGE][13.0] Add sale_warehouse_calendar

mmequignon:
:warning: 
Work in progress
Needs fixing
Do not merge

Computes sale order lead times with respect to to Warehouse calendar

PR #1414, opened 2021-01-12
[MIG] sale_partner_cutoff_delivery_window: Migration to 14.0

dzungtran89:

PR #1413, opened 2021-01-12
[MIG] sale_partner_delivery_window: Migration to 14.0

dzungtran89:

PR #1403, opened 2020-12-29
[MIG] sale_cutoff_time_delivery: Migration to 14.0

phuctranfxvn:

PR #1736, opened 2021-10-07
[14.0][ADD] sale_delivery_date

mmequignon:
Description
===========

Cutoff
------

Potpone order preparation by 1 day if it has been validated after
the partner or the warehouse cutoff.

This can be bypassed if commitment date is set.

Warehouse Calendar
------------------

Postpone delivery according to the warehouse calendar, if any.

Partner delivery window
-----------------------

Allows to define scheduling preference for delivery orders on customers,
in order to select possible delivery windows to postpone deliveries to.

Roadmap
=======

The following modules (available in 13.0) has been merged into this module:
- sale_warehouse_calendar
- sale_cutoff_time_delivery
- sale_partner_delivery_window
- sale_partner_cutoff_delivery_window

The reason for that is that `sale_warehouse_calendar` overrides have to be
executed between `sale_cutoff_time_delivery` and `sale_partner_delivery_window`,
and there's no other clean way to do that.

However, there's still a few things to deal with:
- Clean tests
- Clean code
- Use TZ where it is not the case
- Use calendar instead of time windows
  when delivery preference is `working days`

supersedes:
- https://github.com/OCA/sale-workflow/pull/1414
- https://github.com/OCA/sale-workflow/pull/1403
- https://github.com/OCA/sale-workflow/pull/1413

Also includes https://github.com/OCA/sale-workflow/pull/1696 and https://github.com/OCA/sale-workflow/pull/1720

### DIFF
--- a/sale_delivery_date/__manifest__.py
+++ b/sale_delivery_date/__manifest__.py
@@ -1,0 +1,37 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Delivery Date",
+    "summary": (
+        "Postpones delivery dates based on customer preferences, "
+        "and/or warehouse configuration."
+    ),
+    "version": "14.0.1.0.0",
+    "category": "Sales",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": True,
+    "depends": [
+        "delivery",
+        "partner_tz",
+        "sale_stock",
+        "stock_partner_delivery_window",
+        "stock_warehouse_calendar",
+    ],
+    "data": [
+        # reports
+        "reports/sale_order.xml",
+        "reports/stock_picking.xml",
+        # views
+        "views/res_partner.xml",
+        "views/stock_picking.xml",
+        "views/stock_warehouse.xml",
+    ],
+    "external_dependencies": {
+        "python": ["openupgradelib"],
+    },
+    "pre_init_hook": "pre_init_hook",
+}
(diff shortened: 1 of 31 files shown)


sebalix (commented):
Migration scripts are missing

sale_delivery_date/hooks.py
```diff
@@ -0,0 +1,31 @@
...
+    """Since 4 modules (sale_warehouse_calendar, sale_cutoff_time_delivery,
+    sale_partner_delivery_window, sale_partner_cutoff_delivery_window)
+    are merged into this one, we have to set the previous modules as
+    uninstalled, as well as we need to update the module names in ir_model_data
+    """
+    modules_to_merge = [
+        (merged_module_name, "sale_delivery_date")
+        for merged_module_name in MERGED_MODULE_NAMES
+    ]
+    openupgrade.update_module_names(cr, modules_to_merge, merge_modules=True)
```
sebalix: You have to check first if the addons were installed in the database before calling `openupgrade`. This way Travis (which hasn't previous 13.0 installed) will installed your addon without issue.
For those having previous 13.0 addons installed in their DB, it will crash if `openupgradelib` is not installed, but that's fine in a migration process (we do not want to add `openupgradelib` in the external Python dependencies I think).

sale_delivery_date/__manifest__.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Delivery Date",
+    "summary": "Brief Summary",
```
TDu: This need to be filled in

rousseldenis (approved):
Code review. Looks great

sale_delivery_date/__manifest__.py
mmequignon: oops, nice catch

TDu (approved):
Thank you

sebalix:
@mmequignon do not mention the Odoo version in the commit (except for migration commits) when adding or improving addons

simahawk:
@mmequignon pls rebase here and check the build

sebalix:
Do not merge, the migration script still needs to be fixed (it'll crash if someone install the module for the first time and hasn't `openupgradelib` installed, which is an expected use case).

sebalix:
@simahawk I suppose Runboat doesn't take into account incompatible addons when testing, is it fine to merge or should we wait?

sebalix (approved):

PR #1739, opened 2021-10-15
[13.0][FIX] sale_product_set: fix sequence during set_apply on SO.

acsonefho:
If the user plays too much with sequences on the product set, it's possible to have some negative sequences.
But when the set is applied on the SO negative sequences are added to max_sequence. As the sequence is < 0 it adds the product 'in the middle' of the SO instead of let grouped every product's set.


rousseldenis (approved):
Code reveiw

PR #1744, opened 2021-10-21
[13.0][FIX] sale_order_line_input: The field customer does not exist

CarlosRoca13:
Since v13 the `customer` field on `res.partner` model has been disappeared, so when we try to add a value on the field order_partner_id we get an error.

ping @pedrobaeza @carlosdauden 

@Tecnativa TT31586

### DIFF
--- a/sale_order_line_input/views/sale_order_line_view.xml
+++ b/sale_order_line_input/views/sale_order_line_view.xml
@@ -36,8 +36,7 @@
                 />
                 <field
                     name="order_partner_id"
-                    domain="[('customer','=',True)]"
-                    context="{'search_default_customer':1, 'show_address': 1}"
+                    context="{'res_partner_search_mode': 'customer', 'default_is_company': True, 'default_customer_rank': 1}"
                     options='{"always_reload": True}'
                     required="1"
                     attrs="{'readonly': [('order_partner_id', '!=', False), ('order_id', '!=', False)]}"


sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -36,8 +36,7 @@
                 />
                 <field
                     name="order_partner_id"
-                    domain="[('customer','=',True)]"
-                    context="{'search_default_customer':1, 'show_address': 1}"
+                    context="{'show_address': 1}"
```
pedrobaeza: Please pass also the same context as in v13 regular sale order view that makes the many2one to search first partners with customer rank.

CarlosRoca13:
@pedrobaeza change done

pedrobaeza:
It's not that one. Going to the sales order form view, you can see it:
```
context="{'res_partner_search_mode': 'customer', 'show_address': 1, 'show_vat': True}"
```

CarlosRoca13:
Done

pedrobaeza (approved):

sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -36,8 +36,7 @@
                 />
                 <field
                     name="order_partner_id"
-                    domain="[('customer','=',True)]"
-                    context="{'search_default_customer':1, 'show_address': 1}"
+                    context="{'res_partner_search_mode': 'customer', 'show_address': 1}"
```
carlosdauden:
```suggestion
                    context="{'search_default_customer': 1,'res_partner_search_mode': 'customer', 'default_is_company': True, 'default_customer_rank': 1}"
```
We should use the same context that the customers action uses:
https://github.com/odoo/odoo/blob/4b1a9b2868b024ac3dc3a416f490a68ae09b7379/addons/account/views/partner_view.xml#L233

carlosdauden (changes requested):

sale_order_line_input/views/sale_order_line_view.xml
pedrobaeza: Note that sales order view doesn't do that, which may be consider a bug:

https://github.com/odoo/odoo/blob/4b1a9b2868b024ac3dc3a416f490a68ae09b7379/addons/sale/views/sale_views.xml#L310

carlosdauden: Required changes:

- Add `'default_customer_rank': 1` Without this parameter you can create a customer from an order and then it will not be seen from the orders menu option. Sale order view is wrong.
- Remove `'show_address': 1` This is only to show address in form view, It makes no sense here.

Optional changes:

- `'default_is_company': True` optional, but I preffer that new partner has this by default
- `'search_default_customer': 1'` Applies Customer filter in "Search more..." action. (Optional)

CarlosRoca13:
@carlosdauden done

carlosdauden (approved):
I see that you did not like the option 'search_default_customer' :pensive: :stuck_out_tongue_winking_eye:

PR #1746, opened 2021-10-21
[TO MOVE ON OTHER OCA REPO][14.0][ADD] sale_order_html_sale_conditions

jcoux:
With this module, you can define company sale conditions and display them on sale order reports.

### DIFF
--- a/sale_order_html_sale_conditions/models/sale_order.py
+++ b/sale_order_html_sale_conditions/models/sale_order.py
@@ -1,0 +1,18 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def _default_sale_conditions(self):
+        return self.env.company.default_sale_conditions or ""
+
+    sale_conditions = fields.Html(
+        readonly=True,
+        states={"draft": [("readonly", False)]},
+        default=_default_sale_conditions,
+    )
--- a/sale_order_html_sale_conditions/__init__.py
+++ b/sale_order_html_sale_conditions/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/sale_order_html_sale_conditions/__manifest__.py
+++ b/sale_order_html_sale_conditions/__manifest__.py
@@ -1,0 +1,21 @@
+# © 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale order HTML conditions",
+    "version": "14.0.1.0.0",
+    "category": "Sales",
+    "license": "AGPL-3",
+    "summary": "Add HTML sale conditions into sale order",
+    "depends": ["sale"],
+    "author": "Camptocamp, Odoo Community Association (OCA)",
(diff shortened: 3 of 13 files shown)


pedrobaeza:
Isn't this similar to `sale_comment_template`?

jcoux:
> Isn't this similar to `sale_comment_template`?

If you talk about https://github.com/OCA/sale-workflow/tree/14.0/sale_order_note_template not, because `sale_order_note_template` allow to define the existing `note` field of the sale order from an html template.
Generally this field is used as invoice terms on the invoice.

This new module allow to define sale conditions, but only for the sale order, and allow to continue to define invoice terms.
Sale conditions are generally used on quotation to indicate the conditions to the customer before he confirm the quotation.

SilvioC2C (approved):

omarie-ctc (approved):
Ok for the functionnal part

sale_order_html_sale_conditions/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def _default_sale_conditions(self):
+        return self.env.company.default_sale_conditions or ""
+
+    sale_conditions = fields.Html(
```
damdam-s, line 14: We may want a translation option there

damdam-s (approved):

sale_order_html_sale_conditions/models/sale_order.py
jcoux: Hum, I think it's good like this, the behavior is the same than the core field `note`.
I want to have the same behavior to be consistent.

damdam-s: fine with me ;)

pedrobaeza:
> If you talk about https://github.com/OCA/sale-workflow/tree/14.0/sale_order_note_template not, because `sale_order_note_template` allow to define the existing `note` field of the sale order from an html template. Generally this field is used as invoice terms on the invoice.
> 
> This new module allow to define sale conditions, but only for the sale order, and allow to continue to define invoice terms. Sale conditions are generally used on quotation to indicate the conditions to the customer before he confirm the quotation.

I mean https://github.com/OCA/sale-reporting/tree/14.0/sale_comment_template, and at least I think you should move it to OCA/sale-reporting if you don't consider that the feature can be achieved with this module.

omarie-ctc:
Hello pedrobaeza https://github.com/OCA/sale-reporting/tree/14.0/sale_comment_template doesn't fit with what we expect. We want the sales conditons below the sale order (on an another page) and not below the sale order line. And with this module we don't need Quotation Template.

pedrobaeza:
There's an option to put such comments below the content (but not below the report header, that's true), and you can include `page-break: always` in the HTML to do a page break.

Anyways, pleas move this module to OCA/sale-reporting, as this repository is very crowded, and this feature belongs more to that repository.

jcoux:
Thanks @pedrobaeza 
I will move it (This afternoon if I can or next week).

pedrobaeza:
Another idea: why not adding such option (Page below) to the existing module? It would be very easy to implement, and you can benefit from the features already done (multiple templates, partner assignation...).

jcoux:
Thanks @pedrobaeza for your comments.
I close this PR, and we will see later if we add the option in the module `sale_comment_template`, but you're right: we don't need 2 modules for a similar thing.

PR #1747, opened 2021-10-22
[15.0][MIG] sale_commercial_partner

santostelmo:
Migration of sale_commercial_partner to 15.0

PR #1748, opened 2021-10-22
[15.0][Add] product_sale_manufactured_for

santostelmo:
Add a new module that allows to indicate in products that they were made
(i.e. manufactured) specifically for some customers.

In a sale order, allow to only select products from
the UI that were manufactured for the chosen customer.

Depends of sale_commercial_partner: https://github.com/OCA/sale-workflow/pull/1747


ajaniszewska-dev (approved):
LG

product_sale_manufactured_for/README.rst
```diff
@@ -0,0 +1,87 @@
+=======================
+Sale Commercial Partner
```
ajaniszewska-dev: are you sure it is correct module name here?

santostelmo: No is wrong. The name should be "Product Manufactured for Customer"

rousseldenis:
@santostelmo Could you fix pre-commit and squash your commits ?

PR #1750, opened 2021-10-23
[14.0] Add pricelist_cache

hailangvn:
The aim of the module is to be able to access pricelists' prices very fast without having to wait for the whole calculation.
This is very handy when exposing prices to third parties via API.


pricelist_cache/readme/CREDITS.rst
```diff
@@ -0,0 +1 @@
+The development of this module was financially supported by Camptocamp
```
simahawk:
```suggestion
**Financial support**

* Cosanum
* Camptocamp R&D
```

pricelist_cache/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+Provide a new model to cache price lists and update it,
```
simahawk: would be better to have a longer explanation here. @mmequignon could you help here?

pricelist_cache/data/base_automation.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" ?>
+<odoo>
+    <record id="automation_update_product_pricelist_cache" model="base.automation">
```
simahawk, line 5: We could probably drop this dependency and make the trigger configurable via an ir_config_param.
Can you add a TODO?

pricelist_cache/tests/test_pricelist_cache.py
```diff
@@ -0,0 +1,504 @@
...
+            expected_product_ids,
+        ) in expected_groups.items():
+            self.assertEqual(expected_product_ids, groups[expected_pricelist_id])
+
+    @check_duplicates
+    @freeze_time("2021-04-15")
+    def test_cache_at_product_create(self):
+        """Ensures that cache prices are created at product creation on each global
+        pricelist."""
+        # TODO/FIXME : As all modules are installed by travis, some fields
```
simahawk: anything to do here?

simahawk (changes requested):

pricelist_cache/tests/test_pricelist_cache.py
hailangvn: Thanks for the comments. I am going to work on them.

pricelist_cache/models/product_pricelist_cache.py
```diff
@@ -0,0 +1,247 @@
...
+            - product_prices : The new prices to apply
+        """
+        # Write everything in single transaction
+        values = ",".join(
+            [
+                "({}, {})".format(record.id, product_prices[record.product_id.id])
+                for record in self
+            ]
+        )
+        # pylint: disable=sql-injection
```
TDu: I am not sure this is acceptable.
Same goes for the following ones

sebalix: Indeed, we should use `psycopg2.sql` module to build SQL queries (https://www.psycopg.org/docs/sql.html).
It is used in other addons like https://github.com/OCA/server-tools/blob/14.0/sql_request_abstract/models/sql_request_mixin.py#L13

hailangvn: Thanks @TDu and @sebalix for the comments. The two sql queries have been updated to use `psycopg2.sql` and will be pushed to this PR with other changes.

pricelist_cache/readme/DESCRIPTION.rst
mmequignon: Provides a cron task who caches prices for all products and all pricelists.
The goal is to be able to generate a whole catalog of prices and products for a given customer in a decent time.

Everyday, the cron task will trash the previous day's cache, and rebuild it from scratch.
It means that at any moment, the prices stored in the cache are those of the current day, and will not be recomputed before the next day.

However, new prices will be cached in the following cases:
- new product is created
- new pricelist is created
- new pricelist item is created

mmequignon: @simahawk is it enough, or should I be more precise about this?

pricelist_cache/models/product_product.py
```diff
@@ -0,0 +1,70 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+
+    _inherit = "product.product"
+
+    def write(self, vals):
```
mmequignon: To comment. We will implement a button to do this manually in another PR instead.

pricelist_cache/tests/test_pricelist_cache.py
```diff
@@ -0,0 +1,504 @@
...
+            [
+                ("product_id", "=", self.p8.id),
+                ("pricelist_id", "=", self.list0.id),
+            ]
+        )
+        self.assertTrue(p8_list0_cache)
+        self.assertEqual(p8_list0_cache.price, 100.0)
+
+    @check_duplicates
+    def test_update_pricelist_item(self):
```
mmequignon: Since we do not handle pricelist item updates anymore, please comment this test method as well.
It will be adapted and re-enabled afterwards.

pricelist_cache/tests/test_pricelist_cache.py
```diff
@@ -0,0 +1,504 @@
...
+        p6_cache = cache_model.search(
+            [
+                ("product_id", "=", self.p6.id),
+                ("pricelist_id", "=", self.list3.id),
+            ]
+        )
+        self.assertEqual(p6_cache.price, 25.0)
+
+    @check_duplicates
+    def test_update_product_price(self):
```
mmequignon: same for this one, to comment

pricelist_cache/tests/test_pricelist_cache.py
```diff
@@ -0,0 +1,504 @@
...
+        )
+        # item has been skipped, since parent item (item7) is based on dates
+        self.assertEqual(item9_cache.price, 25.0)
+        item9_cache2 = self.cache_model.get_cached_prices_for_pricelist(
+            item9.pricelist_id, item9.product_id
+        )
+        # Since cache update was previously skipped, get_cache_prices_for_pricelist
+        # should have updated it "on the fly"
+        self.assertEqual(item9_cache2.price, 32.0)
+        self.assertFalse(item9.pricelist_cache_update_skipped)
```
mmequignon: this one as well, I think

pricelist_cache/tests/test_pricelist_cache.py
mmequignon: to fix, this one is important

pricelist_cache/data/base_automation.xml
```diff
@@ -0,0 +1,37 @@
...
+        <field name="name">Update Product Pricelist Cache</field>
+        <field name="model_id" ref="product.model_product_pricelist_item" />
+        <field name="state">code</field>
+        <field name="active" eval="True" />
+        <field name="code">record.update_product_pricelist_cache()</field>
+        <field name="trigger">on_create_or_write</field>
```
mmequignon:
```suggestion
        <field name="trigger">on_create</field>
```

mmequignon (changes requested):

hailangvn:
Thanks @simahawk, @TDu, @mmequignon for your review. This PR has been updated. Please let us know if anything else is needed. Thank you.

TDu:
@hailangvn Thanks for your changes, unfortunately NOT putting them into a separated commit or fixup, makes reviewing them quite hard, could you split them ?

hailangvn:
Thanks @TDu for your comment. The PR has been updated accordingly.

mmequignon (approved):

sebalix (approved):

simahawk (approved):

PR #951, opened 2019-10-03
[10.0] Add sale_quotation_product_sync

grindtildeath:
This module adds a sale quotation synchronization wizard. The wizard aims to
ease the update of existing quotations and templates according to changes (e.g.
unit price) in the product definition.

PR #1756, opened 2021-11-01
14.0 mig sale_quotation_product_sync

leemannd:
Migration to 14.0 of https://github.com/OCA/sale-workflow/pull/951
The prices are not anymore on the template lines

### DIFF
--- a/sale_quotation_product_sync/tests/test_quotation_synchronizer.py
+++ b/sale_quotation_product_sync/tests/test_quotation_synchronizer.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+from odoo.tests import common
+
+
+class TestQuotationSynchronizer(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestQuotationSynchronizer, cls).setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.any_customer = cls.env['res.partner'].search([
+            ('customer', '=', True),
+        ], limit=1)
+        cls.template_default = cls.env.ref(
+            'website_quote.website_quote_template_default')
+        cls.product_2 = cls.env.ref('product.product_product_2')
+        cls.product_3 = cls.env.ref('product.product_product_3')
+        cls.template_default.write({
+            'quote_line': [
+                (0, 0, {
+                    'product_id': cls.product_3.id,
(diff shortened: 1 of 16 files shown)


sale_quotation_product_sync/tests/test_quotation_synchronizer.py
```diff
@@ -0,0 +1,83 @@
...
+                "sale_order_template_line_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": cls.product_3.id,
+                            "name": "Sample Text",
+                            "product_uom_qty": 2.0,
+                            "product_uom_id": cls.product_3.uom_id.id,
+                            # "tested_sample": "Sample3",
```
leemannd, line 26: TODO: check if this field has landed or not on the OCA

sale_quotation_product_sync/tests/test_quotation_synchronizer.py
```diff
@@ -79,12 +72,12 @@ def test_price_propagation(self):
         )
         wizard.execute()
         # The price change should propagate to quotation template line
-        self.assertEqual(self.template_default.quote_line.price_unit, 500.0)
+        # self.assertEqual(self.template_default.quote_line.price_unit, 500.0)
```
leemannd, line 75: There are not anymore price_unit on templates

sale_quotation_product_sync/tests/test_quotation_synchronizer.py
```diff
@@ -68,9 +61,9 @@ def test_price_propagation(self):
         order_copy.action_confirm()
 
         self.assertEqual(order_line_p3.price_unit, 450.0)
-        self.assertEqual(order.amount_total, 900.0)
+        self.assertEqual(order.amount_total, 24400.0)
```
grindtildeath, line 64: I have a hard time understanding this change, probably because products on the template changed? Then wouldn't it be easier to define a new template and keep the totals in line with the products defined in the setup?

grindtildeath (commented):
AFAICS there is no more price defined on sale quote template and options, and the call to onchange updates uom and description, so IMO this must be stated in the wizard if it's what we want to do.

I'm also wondering if it really makes sense to keep this module considering its goal was to update the prices :question:

leemannd:
Closing. This module seems not to be usefull anymore with odoo latest changes

PR #1761, opened 2021-11-04
[14.0] [ADD] sale_exception_product_sale_manufactured_for

sonhd91:
TMP commit will be removed from PR when [PR#2198](https://github.com/OCA/server-tools/pull/2198) merged


sale_exception_product_sale_manufactured_for/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def exception_partner_can_order(self):
```
TDu:
```suggestion
    def exception_partner_can_order_manufactured_for(self):
```
Just to avoid a too generic name, which could be conflicting in the future.

sonhd91: Thanks

mmequignon (approved):

rousseldenis:
/ocabot migration sale_exception_product_sale_manufactured_for

rousseldenis (approved):

PR #1767, opened 2021-11-09
[FIX] sale_order_lot_selection: Avoid missing warning in the product selection.

nicomacr:
This call of super without a return is cause that alert of warning for product wasn't working.
After this change we ensure that return of warning or domain will apply finally


nicomacr:
@rousseldenis could you check this out?, please

rousseldenis (approved):

rousseldenis:
Fast-tracking this

/ocabot merge patch

nicomacr:
@rousseldenis could you check this https://github.com/OCA/sale-workflow/pull/1771 that is a FW to v14.

PR #1771, opened 2021-11-10
[14.0][FIX] sale_order_lot_selection: Avoid missing warning in the product selection.

nicomacr:
This call of super without a return is cause that alert of warning for product wasn't working.
After this change we ensure that return of warning or domain will apply finally

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -14,8 +14,9 @@ class SaleOrderLine(models.Model):
 
     @api.onchange("product_id")
     def product_id_change(self):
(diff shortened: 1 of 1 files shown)


rousseldenis:
@sbidoul Could you check pre-commit error ?

sbidoul:
A rebase should fix the pre-commit error.

rousseldenis:
@nicomacr So, could you rebase ?

nicomacr:
@rousseldenis ready the rebase, we wait for the result of the tests.

rousseldenis (approved):

rousseldenis:
Fast-tracking this

/ocabot merge patch

PR #1777, opened 2021-11-18
[IMP] sale_exception: remove "ignore" field, add a simpler confirmation action that ignores exceptions instead of a redundant confirm button

kevinkhao:


sebastienbeau (approved):
LGTM (code review)

Can you rebase?

sebastienbeau:
@guewen what do you think about this ?

guewen (commented):
I tend to think it is not a good change. It doesn't work as well with other add-ons.
It is easier to update the value in the field at some point rather than to adapt the callers to call this method. Maybe the case were we want to ignore the exception in an "automatic" workflow is rare or non-existent, but if anything in the tests it's much easier to deal with an "ignore" field.

rousseldenis:
@kevinkhao Is this still applicable?

As there were some improvements in 16.0 version and particularly this one : https://github.com/OCA/sale-workflow/pull/2378/files

PR #1782, opened 2021-11-22
[FIX] sale_automatic_workflow: incorrect fields

paulius-sladkevicius:
TODO: I suppose it would be nice to cover and tests as looks it doesn't cover.


paulius-sladkevicius:
@pedrobaeza could we go "Fast-tracking" with this one too?

pedrobaeza:
Yes, please add tests if possible, as this is not so obvious, and if not, please add more explanation to the commit/PR message.

paulius-sladkevicius:
@pedrobaeza pity, but I'm able to do only quick fixes now. I've updated the commit message.

pedrobaeza (approved):
OK, let's go for this time

PR #1574, opened 2021-05-05
[14.0][MIG] sale_procurement_amendment

rousseldenis:
Depends on:

- [x] #1572

PR #1786, opened 2021-11-25
[14.0][ADD] sale_stock_propagate_quantity_diminution module

chafique-delli:
This module allows the cascade of the decrease of sale order line in the related pickings.


sale_stock_propagate_quantity_diminution/tests/test_sale_stock_propagate_quantity_diminution.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertEqual(stock_move_out2.product_uom_qty, 7)
+        # decrease of quantities
+        line.write({"product_uom_qty": 10})
+        self.assertEqual(line.product_uom_qty, 10)
+        self.assertEqual(stock_move_pick1.product_uom_qty, 10)
+        self.assertEqual(stock_move_pack1.product_uom_qty, 8)
+        self.assertEqual(stock_move_out1.product_uom_qty, 6)
+        self.assertEqual(stock_move_pick2.product_uom_qty, 0)
+        self.assertEqual(stock_move_pack2.product_uom_qty, 2)
+        self.assertEqual(stock_move_out2.product_uom_qty, 4)
```
sebastienbeau, line 113: Your test are really really verbose, too much thing (code repetition, useless assert).
Normally a test or a python function should have 20 lines of code, (you can have exception), but the idea is that when is really really long is just unreadable, and we write the code one time and we read it 1000 times, so please make them shorted

sale_stock_propagate_quantity_diminution/tests/test_sale_stock_propagate_quantity_diminution.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertEqual(len(line.move_ids), 1)
+        stock_move_out1 = line.move_ids[0]
+        self.assertEqual(len(stock_move_out1.move_orig_ids), 1)
+        stock_move_pack1 = stock_move_out1.move_orig_ids[0]
+        self.assertEqual(len(stock_move_pack1.move_orig_ids), 1)
+        stock_move_pick1 = stock_move_pack1.move_orig_ids[0]
+        # increase of quantities
+        self.assertEqual(line.product_uom_qty, 10)
+        line.write({"product_uom_qty": 13})
+        self.assertEqual(line.product_uom_qty, 13)
```
sebastienbeau: useless "self.assertEqual(line.product_uom_qty, 13)"

sale_stock_propagate_quantity_diminution/tests/test_sale_stock_propagate_quantity_diminution.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertEqual(stock_move_pick1.product_uom_qty, 13)
+        self.assertEqual(stock_move_pack1.product_uom_qty, 13)
+        self.assertEqual(stock_move_out1.product_uom_qty, 13)
+        # partial delivery of the pick
+        pick1 = self.order4.picking_ids.filtered(
+            lambda r: r.location_id.id == 8 and r.state not in ["cancel", "done"]
+        )
+        stock_move_pick1.write({"quantity_done": 10})
+        wiz_act = pick1.button_validate()
```
sebastienbeau: avoid duplicated code it's hard to read

sebastienbeau (changes requested):

sale_stock_propagate_quantity_diminution/tests/test_sale_stock_propagate_quantity_diminution.py
chafique-delli: @sebastienbeau , I have simplified the tests.

sebastienbeau:
@bguillot

sale_stock_propagate_quantity_diminution/models/sale.py
```diff
@@ -0,0 +1,47 @@
...
+from odoo import _, api, models
+from odoo.tools import float_compare
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _log_decrease_ordered_quantity(self, documents, cancel=False):
+        # no log message as we support the feature
+        pass
```
enriquemartin, line 13: Why remove this functionality? In my opinion, the best option is to modify the `exception_on_so` template and remove the part of the message where it says that a manual action is needed, isn't it?

enriquemartin (commented):
Review done. IMHO the name could be more like its purchase version, something like: sale_stock_propagate_qty

sebastienbeau:
> Review done. IMHO the name could be more like its purchase version, something like: sale_stock_propagate_qty
you right this name will be better, I will do it

rousseldenis:
@chafique-delli @sebastienbeau  There is already a module doing the same. https://github.com/OCA/sale-workflow/pull/1574

Could you review this instead ?

sebastienbeau:
@rousseldenis sorry for responding so late.
This module do not have exactly the same behaviour (we allow to reduce the quantity on partial processed qty).
For example you have to pick 3 product, because of some stock issue you pick 2, then you call your customer and reduce the ordered qty to 2, in that case we want to update the pick and the pack to 2.

I do not think it's a good idea to have 2 modules that do quite the same thing in the repo. So I propose to keep this PR Open and not merge it (or we close it and I move this module in ak repo). Choose what you prefere
Then in V16 I will see if I can propose a PR directly in odoo core, we should have something in native odoo, it's pity that we do not have such feature

PR #1787, opened 2021-11-25
[14.0][MIG] sale_invoice_policy

enriquemartin:
Migration + some changes:
- Simplified field management in `res.config.settings`: removed duplicate fields to use the ones added by sale (`default_invoice_policy`).
- Moved `post-init-hook` to `pre-init-hook` in order to keep the default method of the `default_invoice_policy` field.
- Same literals for `invoice_policy` field as in sale
- Updated pot file + Spanish translation


damdam-s:
travis error is not related to this module.

However, can you please create 1 commit containing the "normal" migration to v14 and another one in which you do the evolutions/changes please?

Changes will be easier to review w/ this process.
thanks a lot in advance

enriquemartin:
That was my initial idea: do a standard migration and then propose changes. But in the process of reviewing the code,  I encountered some issues that did not work without changes. E.g. the default value of `default_invoice_policy` field is not working on a first installation, because of that the `post_init_hook` has to be changed to `pre_init_hook`. 
I will separate the rest of the changes in a separate commit.
Thanks!

enriquemartin:
Done @damdam-s two separate commits to facilitate review!

enriquemartin:
any news on this @damdam-s? thanks in advance!

jonerikceberio (commented):
LGTM

jonerikceberio (approved):

sale_invoice_policy/models/product_template.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        Apply the invoice_policy given by context (if exist) otherwise use the
+        default invoice policy given by the field with this same name.
+        If the product is type = 'service', we don't have to apply the invoice
+        policy given by the context.
+        :return:
+        """
+        invoice_policy = self.env.context.get("invoice_policy")
+        for tmpl in self:
+            if tmpl.type != "service" and invoice_policy:
```
flotho, line 48: Could you please precise why you make a difference with service products.
either 'service' and 'product' could be invoiced based on ordered or delivered quantity.

sale_invoice_policy/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+            ("order", "Invoice what is ordered"),
+            ("delivery", "Invoice what is delivered"),
+        ],
+        string="Invoicing Policy",
+        readonly=True,
+        states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
+        help="Ordered Quantity: Invoice quantities ordered by the customer.\n"
+        "Delivered Quantity: Invoice quantities delivered to the customer.",
+    )
+    invoice_policy_required = fields.Boolean(default=False, readonly=True)
```
flotho: sorry if I didn't read correctly but could you precise where this field is used, I can see a view but no python part using this variable.

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+        "qty_invoiced",
+        "qty_delivered",
+        "product_uom_qty",
+        "order_id.state",
+        "order_id.invoice_policy",
+    )
+    def _get_to_invoice_qty(self):
+        invoice_policies = set(self.mapped("order_id.invoice_policy"))
+        line_by_id = {sol.id: sol for sol in self}
+        done_lines = self.env["sale.order.line"].browse()
```
flotho: won't it return all the order.line of the database?

enriquemartin:
@i-vyshnevska any chance to review this? Thanks in advance!

sale_invoice_policy/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["sale", "sale_stock"],
+    "data": [
+        "data/sale_data.xml",
+        "wizards/res_config_view.xml",
+        "views/product_template_view.xml",
+        "views/sale_view.xml",
+    ],
+    "pre_init_hook": "pre_init_hook",
```
i-vyshnevska, line 18: should stay in post_init_hook as in pre-state "invoice_policy" field will not de recomputed with code presented in this addon

sale_invoice_policy/models/sale_order_line.py
i-vyshnevska: no that will be empty model this is equal to 
`done_lines = self.env["sale.order.line"]`

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+        # were not in so_lines
+        super(SaleOrderLine, self - done_lines)._get_to_invoice_qty()
+        return True
+
+    @api.depends(
+        "state",
+        "product_uom_qty",
+        "qty_delivered",
+        "qty_to_invoice",
+        "qty_invoiced",
```
i-vyshnevska:
```suggestion
```
dependency on these fields injected in parent method so we don't need to have it here

sale_invoice_policy/models/product_template.py
enriquemartin: You are right, we no longer make a distinction.

sale_invoice_policy/models/sale_order.py
enriquemartin: Field removed. Indeed it was not used after code simplification.

sale_invoice_policy/models/sale_order_line.py
enriquemartin: Modified to simplify understanding

sale_invoice_policy/models/sale_order_line.py
enriquemartin: Removed!

sale_invoice_policy/__manifest__.py
enriquemartin: I do not understand the problem. The _pre_init_hook()_ just creates in SQL the new _default_invoice_policy_ column and initializes it with the same value as _invoice_policy_.
It does not need any python code at all.

i-vyshnevska: nevermind, you are right

i-vyshnevska (approved):

damdam-s (approved):
thanks

sale_invoice_policy/models/product_template.py
enriquemartin: I was not right. We need to make a distinction, for example, on orders with storable product and shipping charges: if we mark the billing policy as delivered quantities on the order, shipping charges will never be billed (unless the delivered quantity of this service is forced).
I reverse this change, and leave a comment in the code.

flotho (approved):
thanks for your explanations

flotho (approved):

enriquemartin:
In order to be compatible with other addons as `sale_automatic_workflow`, the module cannot set the default value for _invoice_policy_ on `sale.order`, so I have deleted the `default_get` method.
An empty value on field _invoice_policy_ on sale order means that the used value will be the one in each product.

sale_invoice_policy/models/sale_order.py
```diff
@@ -1,43 +1,22 @@
...
         readonly=True,
         states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
-        help="Ordered Quantity: Invoice based on the quantity the customer "
-        "ordered.\n"
-        "Delivered Quantity: Invoiced based on the quantity the vendor "
-        "delivered (time or deliveries).",
+        help="Ordered Quantity: Invoice quantities ordered by the customer.\n"
+        "Delivered Quantity: Invoice quantities delivered to the customer.",
     )
-    invoice_policy_required = fields.Boolean(
```
rousseldenis: This module changes the standard behaviour. It's better to keep it optional.

enriquemartin: I don't understand what you are talking about: are you meaning the deletion of the `invoice_policy_required` field is not "standard"?
As I said the option to make required `invoice_policy` on sale.order makes this module not compatible with others like `sale_automatic_workflow`.

bunhin: It seems the comment from @rousseldenis has made this PR review process not active, while asked question not yet replied

elvise:
Hi @enriquemartin and thanks for your porting ! 
Can you rebase ?
Thanks in advance :)

ivantodorovich (approved):

enriquemartin:
@dreispt why the needs fixing tag?

dreispt:
@enriquemartin That looks like a mistake. I guess this is ready to merge!

PR #1797, opened 2021-11-29
[15.0][MIG]sale_validity: Migration to 15.0

GuillemCForgeFlow:
Standard migration
@ForgeFlow


HaraldPanten (changes requested):
Hi @GuillemCForgeFlow @AaronHForgeFlow @MiquelRForgeFlow is this module really needed?

What's the difference between this one and the standard functionality?

THX!

GuillemCForgeFlow:
> Hi @GuillemCForgeFlow @AaronHForgeFlow @MiquelRForgeFlow is this module really needed?
> 
> What's the difference between this one and the standard functionality?
> 
> THX!

The only difference I see is that with the standard functionality, if I change the quotation date, the expiration date doesn't update accordingly. That doesn't happen with the sale_validity module. @HaraldPanten

HaraldPanten:
> > Hi @GuillemCForgeFlow @AaronHForgeFlow @MiquelRForgeFlow is this module really needed?
> > What's the difference between this one and the standard functionality?
> > THX!
> 
> The only difference I see is that with the standard functionality, if I change the quotation date, the expiration date doesn't update accordingly. That doesn't happen with the sale_validity module. @HaraldPanten

@rousseldenis @pedrobaeza  I think we shouldn't keep this module, as for standard we have the 90% of the functionality. On the other hand, If we keep this one it has to be refactored because right now, users have 2 settings options for doing the same.

At the most, maybe we could make an improvement for the standard functionality, but IMO I think it is not needed...

WDYT?

Look at my screenshot:

Screenshot

GuillemCForgeFlow:
> > > Hi @GuillemCForgeFlow @AaronHForgeFlow @MiquelRForgeFlow is this module really needed?
> > > What's the difference between this one and the standard functionality?
> > > THX!
> > 
> > 
> > The only difference I see is that with the standard functionality, if I change the quotation date, the expiration date doesn't update accordingly. That doesn't happen with the sale_validity module. @HaraldPanten
> 
> @rousseldenis @pedrobaeza I think we shouldn't keep this module, as for standard we have the 90% of the functionality. On the other hand, If we keep this one it has to be refactored because right now, users have 2 settings options for doing the same.
> 
> At the most, maybe we could make an improvement for the standard functionality, but IMO I think it is not needed...
> 
> WDYT?
> 
> Look at my screenshot:
> 
> Screenshot

I agree with you that it makes no sense to show two options making almost the same. I don't know if it's really important to add the part of the funcionality that it's missing with the standard one.

So, as you choose. @HaraldPanten @pedrobaeza @rousseldenis

pedrobaeza:
I don't think it worths to only keep such onchange. What is interesting for the migration path is to merge the module in OpenUpgrade and convert corresponding data for keeping the same behavior without extra configuration.

rousseldenis:
> I don't think it worths to only keep such onchange. What is interesting for the migration path is to merge the module in OpenUpgrade and convert corresponding data for keeping the same behavior without extra configuration.

@GuillemCForgeFlow Is that already done ? Could we close this ?

GuillemCForgeFlow:
Hi @rousseldenis I'm not sure if this has been done on the OpenUpgrade. What I assume then, is that this migration is not needed as the module doesn't make sense anymore with the standard functionality.

The only thing left here is to merge the data from this module on the OpenUpgrade, perhaps @pedrobaeza knows that?

pedrobaeza:
No, not done yet in OpenUpgrade.

PR #1801, opened 2021-11-30
[12,0] [FIX] Fixed delivery_state when the sale_order is in the state "sent"

mattiolato98:
When the sale_order switches to the state "sent", the delivery_state becomes "unprocessed". 
This is not correct, because it's still a price quotation.

Now, by adding "sent" to the list of the states in which the delivery_state should be 'no', only when the sale_order is confirmed the delivery_state assumes value "unprocessed".


rousseldenis:
@mattiolato98 Could you add this case to tests ?

PR #1802, opened 2021-12-01
[FIX] sale_automatic_workflow_payment_mode: not register payment when no fixed journal

sonhd91:
This PR fix bug when `fixed_journal_id `is null


sonhd91:
Hello @rousseldenis , can you merged ?

rousseldenis:
> Hello @rousseldenis , can you merged ?

Before merging, the process is to have reviews. I'll look at it

rousseldenis:
@sonhd91 Could you maybe add a test for this case ?

rousseldenis:
@sonhd91

PR #1805, opened 2021-12-02
[13.0][IMP] sale_partner_selectable_option: Change position field.

victoralmau:
Change position field.

Please @ernestotejeda and @pedrobaeza can you review it?

@Tecnativa TT33223

### DIFF
--- a/sale_partner_selectable_option/views/res_partner_view.xml
+++ b/sale_partner_selectable_option/views/res_partner_view.xml
@@ -2,16 +2,13 @@
 <!-- License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
     <record id="view_partner_form" model="ir.ui.view">
-        <field name="name">Add selectable_in_sales_orders to partner form</field>
+        <field name="name">Add sale_selectable to partner form</field>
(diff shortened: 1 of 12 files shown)


ernestotejeda (approved):

sale_partner_selectable_option/views/res_partner_view.xml
```diff
@@ -6,12 +6,12 @@
...
         <field name="arch" type="xml">
-            <xpath
-                expr="//page[@name='sales_purchases']//field[@name='user_id']"
-                position="after"
-            >
-                <field name="selectable_in_sales_orders" />
-            </xpath>
+            <group name="sale" position="inside">
+                <field
+                    name="selectable_in_sales_orders"
```
pedrobaeza: Sorry for not checking before, but the field name is a bit convoluted. It can be `sale_selectable`.

pedrobaeza (approved):
Thanks for the changes and the migration script.

pedrobaeza:
I'm tired of that ugly modules failing each time...

victoralmau:
I have restarted Travis's jobs and they are already green. Can you try again?

PR #1820, opened 2021-12-12
[14.0][IMP] sale_delivery_date: update translations

i-vyshnevska:


sebalix:
I think translations should be done on Weblate for existing addons (we are allow to update/add PO files only when adding a new addon in the repository, otherwise updating them on existing addons will generate conflicts with contributions done on Weblate).
https://odoo-community.org/page/translate

So this PR cannot be merged unfortunately.

pedrobaeza:
Yes, that's it. You can still import .po files into weblate, so barely no extra work is needed to be done. Just to request access to such languages.

PR #1822, opened 2021-12-13
[13.0][ADD] sale_stock_cancel_restriction: new module.

ernestotejeda:
Cc @Tecnativa TT33358

This module extends the functionality of Sales to avoid canceling sales orders with 'Done' pickings.
Derived from: https://github.com/odoo/odoo/commit/abc9ee8e3b33079ddedcaed39fea9116fb0aebff


carlosdauden (approved):

sale_stock_cancel_restriction/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Tecnativa - Ernesto Tejeda
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
```
pedrobaeza: It should be AGPL

sale_stock_cancel_restriction/models/sale_order.py
```diff
@@ -0,0 +1,12 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_cancel(self):
+        self.mapped("picking_ids").filtered(lambda r: r.state == "done").action_cancel()
```
pedrobaeza, line 15: Put a docstring explaining the whys:
```suggestion
        """Force to call the cancel method on done picking for having the expected error,
        as Odoo has now filter out such pickings from the cancel operation.
        """
        self.mapped("picking_ids").filtered(lambda r: r.state == "done").action_cancel()
```

pedrobaeza (changes requested):

ernestotejeda:
@pedrobaeza done

pedrobaeza (approved):

pedrobaeza:
Let's try to merge to see if the random error doesn't happen:

/ocabot merge nobump

PR #1826, opened 2021-12-15
[13.0][ADD] sale_order_partner_restrict

OriolVForgeFlow:
@forgeflow @JordiMForgeFlow


sale_order_partner_restrict/models/sale.py
```diff
@@ -0,0 +1,57 @@
...
+                        ("company_id", "=", False),
+                        ("company_id", "=", order.company_id.id),
+                        "|",
+                        ("parent_id", "=", False),
+                        ("type", "=", "contact"),
+                    ]
+                )
+
+    @api.constrains("partner_id")
+    def _check_order_line_company_id(self):
```
JordiMForgeFlow:
```suggestion
    def _check_order_partner_restrict(self):
```

sale_order_partner_restrict/models/sale.py
```diff
@@ -0,0 +1,57 @@
...
+    sale_order_partner_restrict = fields.Selection(
+        related="company_id.sale_order_partner_restrict",
+        string="Partner Restriction on Sale Orders",
+    )
+    available_partner_ids = fields.Many2many(
+        "res.partner", compute="_compute_available_partner_ids"
+    )
+
+    @api.depends("partner_id", "sale_order_partner_restrict")
+    def _compute_available_partner_ids(self):
```
JordiMForgeFlow: A fallback is needed in every compute method, otherwise an error is raised if you remove the company from the order.

JordiMForgeFlow (approved):

JoanSForgeFlow (approved):

mariadforgeflow (approved):
LGTM! Functional and code review.

HaraldPanten:
Hi Oriol, could you check Travis?

AaronHForgeFlow:
LGTM, but I am wondering  if it is possible to filter the partners instead of showing a validation error. Perhaps a dynamic domain? WDYT @JordiMForgeFlow ?

OriolVForgeFlow:
> LGTM, but I am wondering if it is possible to filter the partners instead of showing a validation error. Perhaps a dynamic domain? WDYT @JordiMForgeFlow ?

Hi Aaron, I'm already filtering the partners with a domain, the validation error is just to double-check. Do you think I should remove that validation error and leave only the filter?

AaronHForgeFlow (commented):
Hi @OriolVForgeFlow I see now the domain in the view. But is that working? 

It seems to me that is not working because the field `available_partner_ids` is computed, and it is not ready when the view is loaded.
In order to solve that you can use a onchange method that changes the domain dinamically, like this one: https://github.com/OCA/account-fiscal-rule/blob/13.0/account_multi_vat/models/res_partner_id_number.py#L63 I would make the onchange method to depends on the sale_order_partner_restrict field. But that is in case it is not working right now only.

AaronHForgeFlow (approved):
Functional + code :+1:

HaraldPanten:
@OriolVForgeFlow @AaronHForgeFlow do you know why is Travis failing?

OriolVForgeFlow:
> @OriolVForgeFlow @AaronHForgeFlow do you know why is Travis failing?

It's not always failing but when it does is for the test of another module:
" duplicate key value violates unique constraint "sale_order_revision_unique" "
I'm not sure how to solve this or if the problem is on the other module, let me know if you have any thoughts!

AaronHForgeFlow:
I do not know  the details but it seems the sale_order_revision modules adds a constraint called "sale_order_revision_unique". I guess a revision number is assigned to every sales order. I need to investigate if the sale order created in the test can be created in a way that constraint does not raise. If that is not possible, then I guess this is something that needs to be solved in the other module.

sale_order_partner_restrict/tests/test_sale_order_partner_restrict.py
```diff
@@ -0,0 +1,127 @@
...
+        cls.partner_delivery = cls.partner_model.create(
+            {
+                "name": "Partner Delivery",
+                "type": "other",
+                "parent_id": cls.partner_parent.id,
+            }
+        )
+
+    def _create_sale_order(self):
+        so = self.env["sale.order"].create({"partner_id": self.partner_parent.id})
```
AaronHForgeFlow: I think here you can create the sales order like this

```suggestion
        so = self.env["sale.order"].create({"partner_id": self.partner_parent.id, "name": "/"})
```
The reason:

https://github.com/OCA/sale-workflow/blob/13.0/sale_order_revision/models/sale_order.py#L78

JordiMForgeFlow (approved):
Code + functional review 👍🏼 Lets see if travis does not complain now :)

HaraldPanten (approved):

PR #1829, opened 2021-12-20
14.0 ADD sale_line_partner

bealdav:
Quick jump to sale order lines linked to current partner.
It jumps to `sale_order_line_input` sale line view and then allow to add lines directly.

cc @carlosdauden @pedrobaeza , thanks for your review


pedrobaeza:
Why not including it in the base module? Why the dependency to `sale_commercial_partner`?

bealdav:
Good question, dependency because of https://github.com/OCA/sale-workflow/pull/1829/files#diff-c1362ead3cd6555c7a1e0d3090e2017998bda2355d0d614651c1775d20244ea6R14

bealdav:
That's why I didn't submitted to the base module

pedrobaeza:
OK then, but seems weird to access from a sales order to other sale lines of the commercial partner. I would see logical to access from the partner to their lines.

bealdav:
Hi Pedro

Imagine you buy products to a company. This company record a customer/contact `Tecnativa, Pedro`. All products are delivered to Tecnativa Corporation
Your colleague Carlos buy some products too. The company record a new contact for this sale `Tecnativa, Carlos` also delivered to Tecnativa.

When the company analyze data, they want to have global view by whole customer not by contact. If they want contact info they can add an extra filter.

I think this is way Odoo intends to do things.

And that's the behavior my customer want. Do you have this needs ?

pedrobaeza:
Yeah, but that's done with the `child_of` operator of `commercial_partner_id` from the partner, not on the sales order children. That's the weird thing. It's only one click away to get to the partner, and then such "statistics".

PR #1830, opened 2021-12-21
[13.0][ADD]sale_advance_payment: Backport from 14.0

GuillemCForgeFlow:
@ForgeFlow


JordiBForgeFlow:
@JasminSForgeFlow can you help @GuillemCForgeFlow to figure out why travis fails here?

JordiBForgeFlow:
The error in Travis seems to be unrelated to this module, but then I can see other pull requests in this repo targeting 13.0 that are green.

sale_advance_payment/tests/test_sale_advance_payment.py
```diff
@@ -0,0 +1,247 @@
...
+            {"name": "Repair Services", "type": "service", "invoice_policy": "order"}
+        )
+
+        cls.tax = cls.env["account.tax"].create(
+            {"name": "Tax 15", "type_tax_use": "sale", "amount": 20}
+        )
+
+        # Sale Order
+        cls.sale_order_1 = cls.env["sale.order"].create(
+            {"partner_id": cls.res_partner_1.id}
```
AaronHForgeFlow:
```suggestion
            {"name": "/", partner_id": cls.res_partner_1.id}
```

I think CI fails because the module sale_order_revision works if you create a sales order as a user, but it does not if you create a sales order using create method. If the name is "/" it will generate a new number for the sales order, so the constraint will not raise.

Can you try that?

JordiMForgeFlow: It still fails...it is weird because it is only failing in the OCB tests :/

AaronHForgeFlow: The same error?

PR #1834, opened 2021-12-22
[IMP] sale_order_line_price_history : add color indication

v-garcies:
This commit adds a color indication to the clock icon in sales order lines :

- black if the customer has never purchased the item
- blue if the customer has already purchased the item

Capture d’écran 2021-12-22 à 11 18 09


sale_order_line_price_history/models/sale_order_line.py
```diff
@@ -9,3 +10,19 @@ class SaleOrderLine(models.Model):
...
     order_partner_id = fields.Many2one(depends=["product_id"])
+
+    def _compute_already_sold(self):
+        for line in self:
+            domain = [
+                ("product_id", "=", line.product_id.id),
+                ("state", "in", ["sale", "done"]),
+                ("order_partner_id", "child_of", line.order_id.partner_id.commercial_partner_id.ids),
+            ]
+            order_lines = line.env["sale.order.line"].search(domain, limit=1)
```
rousseldenis: I think this could be optimized through a read_group before the loop

rousseldenis (changes requested):
Code review.

I like this kind of addition.

Could you add some functional information about this in the /readme section and improve this PR title (as 'color indication' yes, but related to ?) ?

Thanks

pedrobaeza:
Isn't better to disable the icon if no purchases? Clicking on it will lead to an empty list, so let's not allow to click it at all.

rousseldenis:
> Isn't better to disable the icon if no purchases? Clicking on it will lead to an empty list, so let's not allow to click it at all.

I would say that having an icon or not mixed on lines could be UX deficient, no ?

pedrobaeza:
Disabling it is not hiding it

pedrobaeza:
Anyways, hiding it is another approach of UX, not UX deficient. Odoo does it continuously across modules.

fdesmottes:
Thanks @v-garcies it's working well for me 👍

ajaniszewska-dev (approved):
LG

fdesmottes (approved):
Thanks @v-garcies 👍

sale_order_line_price_history/models/sale_order_line.py
v-garcies: Done 👍

pedrobaeza (changes requested):
Please check my comments and don't add yourself as author, as this is a minimum improvement

PR #1842, opened 2021-12-27
[ADD] Sale Order Line Hide Create Product: New module to 14.0

manuelcalerosolis:
This module change the views of the sales lines so that it does not allow the creation of products on the fly.


pedrobaeza:
Better to use IMO `web_m2x_options` with the system parameter `create=False`.

manuelcalerosolis:
@pedrobaeza thanks for the info.

web_m2x_options changes the behavior in the entire application, in this case we only seek to hide this possibility in the sales operations, and only with the products.

pedrobaeza:
The quick create option is very harmful in any screen (I'm not talking about the "Create and edit..." one, that lets you cancel), so I disable it entirely for whole Odoo.

manuelcalerosolis:
> The quick create option is very harmful in any screen (I'm not talking about the "Create and edit..." one, that lets you cancel), so I disable it entirely for whole Odoo.

I totally agree with you, it only manages to create incorrect and incomplete customer and product records, but the customer in this case rules.

pedrobaeza:
OK, as you prefer :)

PR #1843, opened 2021-12-27
[10.0][IMP] sale_stock_picking_note: update picking notes if changes in the order

AaronHForgeFlow:
It could happen the customer demands some special instructions after the order is confirmed, those instructions need to be communicated to the warehouse.

If this PR is accepted I will forward-port this to newer versions.

@ForgeFlow


rousseldenis:
It seems that 10.0 branch is broken.

PR #1851, opened 2022-01-05
[15.0] [MIG] sale_delivery_state

celm1990:


celm1990:
somebody can help me review this PR?
ping @OCA/oca-members  @OCA/crm-sales-marketing-maintainers

sale_delivery_state/__manifest__.py
```diff
@@ -5,13 +5,14 @@
...
     "summary": "Show the delivery state on the sale order",
-    "version": "14.0.1.1.0",
+    "version": "15.0.1.0.0",
     "category": "Product",
     "website": "https://github.com/OCA/sale-workflow",
     "author": "Akretion, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
     "depends": ["sale"],
+    "external_dependencies": {"python": ["mock"]},
```
simahawk: please remove this, it's not a functional dependency

simahawk (changes requested):

sale_delivery_state/__manifest__.py
simahawk: also `mock` is a built-in lib, no need to declare its dependency :wink:

celm1990: Thank for you review, i updated code according your suggestions

simahawk (approved):

sale_delivery_state/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    @api.depends("order_line.qty_delivered", "state", "force_delivery_state")
+    def _compute_sale_delivery_state(self):
+        for order in self:
+            if order.state in ("draft", "cancel"):
+                order.delivery_state = "no"
+            elif order.force_delivery_state or order._all_qty_delivered():
+                order.delivery_state = "done"
+            elif order._partially_delivered():
```
carlosdauden, line 80: Checking in the same loop would be more efficient than filtering, getting precision and going through the lines for both the _all_qty_delivered method and the _partially_delivered method, but I suppose it will be because it is necessary to extend these methods in some case ...

carlosdauden (approved):

Issue #1817, opened 2021-12-10
[BUG][14.0] sale_product_set: Removing products in "Add Set" wizard doesn't do anything


gaikaz:
## Module
sale_product_set

## Describe the bug
Removing products in "Add Set" wizard doesn't do anything.
Even though intuitively it seems that this table is sort of a "staging" place before adding the products to the Sale Order. 

## To Reproduce
**Affected versions**: 14.0 (probably earlier too)

Steps to reproduce the behavior (doable on OCA Runbot):
1. On DB with demo data
2. Start creating new Sale Order
3. Click "Add Set" (opens the wizard)
4. Select demo set "i5 computer offer" (3 products show up)
5. Remove one product from the wizard window, so that 2 remain
6. Submit the wizard so it adds the products to the Sale Order
7. !! All 3 initial set products get added, so the removed one was re-added

**Expected behavior**
Set should have added only 2 products to the Sale Order Lines instead of the whole set.

**Additional context**

https://user-images.githubusercontent.com/17097810/145581157-b4067fbb-d4c8-490b-8540-0515425be851.mp4

legalsylvain:
Hi. this seems the expected behaviour for "set", don't you think. i better see the unpossiblity to remove product in the wizard, don't you think ?

gaikaz:
Hi. Our project managers and couple of clients that use this module really thought that my described behavior is what should happen. It just seems this way intuitively.
But I do agree, that this is sort of a UI/UX issue if that was never the intention. Those x'es shouldn't be there if they don't even do anything.
Although, I'm pretty sure they are added explicitly, so I would guess that some versions ago, they did work?

pedrobaeza:
For improving the UX, such list can be readonly.

rousseldenis:
> For improving the UX, such list can be readonly.

Yes, but in a way, maybe with a configuration option/group to let it flexible.

gaikaz:
We probably could make the X'es work instead of removing them. I honestly would prefer that as it would come in handy.

Mantux11:
I'm working on this bug. I'll make a PR soon.

PR #1862, opened 2022-01-12
[FIX] sale_product_set: fixed issue with wizard lines delete

Mantux11:
Issue: [#1817](https://github.com/OCA/sale-workflow/issues/1817)

Previously deleted line in wizard was not capable because of `compute` method. `Compute` method is generating from new one after you press `Add set` button in wizard. And all your changes in lines, isn't affected. So added `store=True` and it fix for now one.

### DIFF
--- a/sale_product_set/tests/test_product_set.py
+++ b/sale_product_set/tests/test_product_set.py
@@ -166,3 +166,12 @@ class TestProductSet(common.SavepointCase):
         )
         order_line.ensure_one()
         self.assertEqual(order_line.discount, set_line.discount)
+
+    def test_product_set_delete_in_wizard(self):
+        products = self.so.order_line.product_id
+        wizard = self._get_wiz()
+        wizard.write({"product_set_line_ids": [(3, wizard.product_set_line_ids[0].id)]})
+        products |= wizard.product_set_line_ids.product_id
+        wizard.add_set()
+        self.assertEqual(products, self.so.order_line.product_id)
+        self.assertEqual(len(self.so.order_line), 3)
--- a/sale_product_set/README.rst
+++ b/sale_product_set/README.rst
@@ -98,6 +98,7 @@ Contributors
 * Souheil Bejaoui <souheil.bejaoui@acsone.eu>
 * Adria Gil Sorribes <adria.gil@forgeflow.com>
 * Phuc (Tran Thanh) <phuc@trobz.com>
+* Mantas Šniukas <mantas@vialaurea.lt>
 
 Other credits
 ~~~~~~~~~~~~~
--- a/sale_product_set/wizard/product_set_add.py
+++ b/sale_product_set/wizard/product_set_add.py
@@ -25,8 +25,9 @@ class ProductSetAdd(models.TransientModel):
     )
     product_set_line_ids = fields.Many2many(
         "product.set.line",
-        "Product set lines",
+        string="Product set lines",
         required=True,
+        store=True,
         ondelete="cascade",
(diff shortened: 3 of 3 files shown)


gaikaz (approved):
Functional and Code review. LGTM

Mantux11:
Can someone review this PR and changes? @OCA/oca-members @OCA/crm-sales-marketing-maintainers  
It seems that this module was updated lately.

simahawk (changes requested):
Thanks for your contrib.
Please not that you are breaking a feature - passing lines via ctx - that was added for a reason.

Can you elaborate a bit more on what are you trying to fix? Is not clear to me...

simahawk:
> Thanks for your contrib. Please not that you are breaking a feature - passing lines via ctx - that was added for a reason.
> 
> Can you elaborate a bit more on what are you trying to fix? Is not clear to me...

ah never mind, I have to read https://github.com/OCA/sale-workflow/issues/1817  :sweat_smile:

simahawk:
ok, the motivation is clear. The solution is not. 
Have you tried setting the field as `store=True`?

Mantux11:
@simahawk Okey, I haven't thought with that solution. You're right. I will change this.

Mantux11:
@simahawk Could you review now?

petrus-v:
@Mantux11 It's shame you've refactored test using Form object in previous version that I suppose reproduce error describe in #1817 ? could be nice to have them if you kept them ?

simahawk:
LG to me except the commit msg: can you rewrite it like this?

```
sale_product_set: fix custom lines edit

Was not possible to change lines manually.
Closes https://github.com/OCA/sale-workflow/issues/1817
```
Keep in mind that commit msgs must be kept short and their description goes to... the description :wink: 
The are also some Odoo guidelines here https://www.odoo.com/documentation/15.0/developer/misc/other/guidelines.html#commit-message-structure.
Personally I think the `[TAG]` is just consuming chars for nothing IF the commit msg explains clear what you are doing (eg: "fix this or that").

@petrus-v I would add a specific test for this. Also, is not needed to use the Form class: you can set the field and invalidate the cache to see that happening. OTOH if he wants to use Form class... fine :wink:

sale_product_set/tests/test_product_set.py
```diff
@@ -166,3 +166,13 @@ def test_discount(self):
         )
         order_line.ensure_one()
         self.assertEqual(order_line.discount, set_line.discount)
+
+    def test_product_set_delete_in_wizard(self):
+        self.wizard = self._get_wiz()
```
simahawk:
```suggestion
        wizard = self._get_wiz()
```
there's no need to set this on the test class instance

sale_product_set/tests/test_product_set.py
```diff
@@ -166,3 +166,13 @@ def test_discount(self):
...
+        self.wizard.product_set_line_ids -= self.wizard.product_set_line_ids[1]
```
simahawk: I bet this test pass even w/out your change because you set the value of the cache here, isn't it? :wink:

sale_product_set/tests/test_product_set.py
Mantux11: Yes, I just forgot to take out from self.

sale_product_set/tests/test_product_set.py
Mantux11: @simahawk Thanks. I fixed that now and agree with you. Could you look now, maybe you have more suggestions?

simahawk (approved):
LG

PR #1867, opened 2022-01-14
[12.0][FIX] get sequence from date order

sergiocorato:
Currently an order created on 2022 with order date 2021 will get a sequence from 2022, even if sequence is set as multi-year.
This PR fix it.


rousseldenis:
@sergiocorato Thanks for this. Could you provide a test to avoid regression ?

PR #1855, opened 2022-01-09
[14.0] [MIG] sale_automatic_workflow_job

saxomanu:

PR #1868, opened 2022-01-17
[14.0][MIG] sale_automatic_workflow_job: Migrate to version 14.0

hailangvn:


leemannd (approved):

simahawk (approved):

rousseldenis:
@hailangvn @simahawk Please check before proposing a PR if another one is not present.

https://github.com/OCA/sale-workflow/pull/1855

simahawk:
ops, sorry, I missed it :disappointed:

hailangvn:
> @hailangvn @simahawk Please check before proposing a PR if another one is not present.
> 
> #1855

Agree that we should avoid duplicated work. But we need to consider the how. We have three options.
1. Everyone should follow process: update [Migration Issue](https://github.com/OCA/sale-workflow/issues/1274) before doing migration;
2. Someone follows 1. and someone does not, extra two tasks are a) to search for PR before doing migration; + b) to search for PR before  creating PR;
3. Drop process to update [Migration Issue](https://github.com/OCA/sale-workflow/issues/1274) and do a) + b) only;

IMHO Number 1. is the ideal one and PR not registered in [Migration Issue] will be considered as redundant and for reference only. Number 2. has most work and number 3 still cause duplicated work. Is there other way?

Good day @saxomanu, @yostashiro, may I know about your thought?

PR #1869, opened 2022-01-17
[14.0][ADD] sale_automatic_workflow_ignore_exception

jcoux:
With this module,
you can ignore sale exceptions when validate sale with sale automatic workflow.

### DIFF
--- a/sale_automatic_workflow_ignore_exception/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow_ignore_exception/models/automatic_workflow_job.py
@@ -1,0 +1,14 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AutomaticWorkflowJob(models.Model):
+
+    _inherit = "automatic.workflow.job"
+
+    def _do_validate_sale_order(self, sale, domain_filter):
+        if sale.workflow_process_id.ignore_exception_when_confirm:
+            sale = sale.with_context(ignore_exception_when_confirm=True)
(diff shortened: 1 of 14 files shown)


tarbib (approved):
Functional tests ok

damdam-s (approved):

ivantodorovich (approved):

sale_automatic_workflow_ignore_exception/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import models
+
+
+class AutomaticWorkflowJob(models.Model):
+
+    _inherit = "automatic.workflow.job"
+
+    def _do_validate_sale_order(self, sale, domain_filter):
+        if sale.workflow_process_id.ignore_exception_when_confirm:
+            sale = sale.with_context(ignore_exception_when_confirm=True)
```
rousseldenis, line 13: @jcoux I'm wondering if:
```suggestion
            sale.write({"ignore_exception": True})
```
is not better ?

jcoux: It can be more easy.

But in the original module, we have a checked to bypass the sale order validation:
https://github.com/OCA/sale-workflow/blob/50675a1ffb8ac8f60993c22a7246ec321bb0c522/sale_automatic_workflow/models/automatic_workflow_job.py#L43

I don't want to ignore exception in this case.
Then the context option seems to be the best way for me.

dreispt:
@rousseldenis Are there objections for a merge?

rousseldenis (approved):

jcoux:
@dreispt Now approved: you can merge :)

PR #1876, opened 2022-01-19
[14.0][FWD][FIX] sale_product_set: fix sequence during set_apply on SO.

rousseldenis:
If the user plays too much with sequences on the product set, it's possible to have some negative sequences.
But when the set is applied on the SO negative sequences are added to max_sequence. As the sequence is < 0 it adds the product 'in the middle' of the SO instead of let grouped every product's set.

Forwardport from #1739


manuelregidor (approved):
We have added this commit to the v15 migration and it is working fine.

pedrobaeza (approved):

PR #1880, opened 2022-01-20
[FIX] duplicate sale.order need to have 'New' as default name

njeudy:
When copy a sale order, new order should have new number, this is done by passing 'New' value à default name for a sale.order.
This is the case in sale_order_revision too.

This Merge request should fix the red status of travis for the sale-workflow repo

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -4,7 +4,7 @@
 # © 2020 Manuel Regidor  <manuel.regidor@sygel.es>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import _, api, models
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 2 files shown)


sale_quotation_number/models/sale_order.py
```diff
@@ -25,7 +25,7 @@ def copy(self, default=None):
         self.ensure_one()
         if default is None:
             default = {}
-        default["name"] = "/"
+        default["name"] = "New"
```
legalsylvain: Shoule not be _("new") ?
Just a question, i'm not sure.

rousseldenis (approved):

njeudy:
Think there is a mistake with sale_quotation_number and sale_order_revision .. will investigate this in this PR to clean travis for the repos ..

njeudy:
@pedrobaeza @legalsylvain I fix the same behaviour in `sale_order_revision` and `sale_quotation_number` because travis need this two module to be green. I wait for the ci to finish :)

njeudy:
@legalsylvain @pedrobaeza I don't understand why 1 of 2 travis test is ok and the other failed  ? do you ?

pedrobaeza:
My colleague has prepared a workaround for the tests in #1884, so I have pushed it directly in 02891714e79a841aa484818d9b3b366060101e51 and we can close this one. Thanks for trying to solve it. It's a nasty patch though, as there's no good way to avoid it.

PR #1882, opened 2022-01-20
[13.0][FIX] sale_order_line_input: company_id field is empty and multi-company...

carlosdauden:
domains don't work. Taxes empty

Before this PR only partners and products without company could be selected and the tax field was empty.

I try use several methods to set order_id (NewId) related info from sale.order.line, but when the flow returns to user browser this info is lost.

Until we find a better solution or the ORM allows it, the record will be created in an onchange, something that, as I say, is not ideal.
The only downside is that if after selecting a partner you discard the introduction of the order line, you have to delete it manually.

@Tecnativa


carlosdauden:
Travis error in sale_quotation_number module

Tardo (approved):
Code review

sergio-teruel (approved):

PR #1890, opened 2022-01-29
[14.0][FIX] Fix access right issues when using actions

sebalix:


pedrobaeza:
Should we add this change as a line in the migration guide?

sebalix:
@pedrobaeza Yes indeed

simahawk (approved):

PR #1891, opened 2022-01-30
[12.0][ADD] sale_product_email module

fcayre:


sale_product_email/tests/test_project.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020 Commown SCIC SAS (https://commown.fr)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+
+
+class TestProjectRatingNPS(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
```
rousseldenis: This part should not be here (not related)

sale_product_email/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def send_sold_products_mail(self):
+        for order in self:
+            for line in order.order_line:
```
rousseldenis, line 18: @fcayre Is this really wanted by line ? Shouldn't it be grouped by sale order ? The amount of outbound emails can be great with this.

rousseldenis (changes requested):
Code review

sale_product_email/models/sale_order.py
fcayre: @rousseldenis Sorry I haven't seen your review. Thanks very much for the review.
Yes we want this for each product _which was configured with an email template_, so for each sale order line. An example use case is a manual to use the product or an activation procedure that is specific to a product.

sale_product_email/tests/test_project.py
fcayre: Indeed, gonna remove this!

sale_product_email/models/sale_order.py
rousseldenis: @fcayre Yes, but shouldn't be better to group in one email with all lines ? If customer has plenty lines in his sale order, he will receive the same amount of **different** emails. It can be counter-productive, don't you think ?

fcayre: Hummm, OK @rousseldenis, understood you suggest grouping all potential emails into a single one.

Although this suggestion obviously responds to a superset of the use cases of the present implementation, I am afraid this would make things impracticable:
- for the backoffice odoo user: how to define email parts? how to get a preview of the resulting email?
- for the developer: the extra-complexity may not be worth it regarding maintenance; I would suggest keeping it simple in this module and let another modules reimplement the `send_sold_products_mail` method to extend to more "massive" use cases if someone needs it. Maybe people would never use this extra concatenation feature and let their customers access the documents online instead of sending the by email.

If you have a simple implementation in mind, however, tell me and I'd be happy to try to implement it.

PR #1893, opened 2022-01-31
[15.0][FIX] sale_order_line_date: Refactor code

BernatPForgeFlow:
Sale order line date refactor code. 

Previous behavior:
User can't change Delivery Date from a SO. Moreover, connection between order lines Delivery Dates and their sales order Delivery Date is not coherent.

Current behavior:
- User can select different delivery dates for each order line and SO Delivery Date will be calculated taking into account shipping policy. This works in the different states of the SO.¡
- If there's no delivery dates in sales order lines, the operation will be as the standard.

### DIFF
--- a/sale_order_line_date/models/sale_order.py
+++ b/sale_order_line_date/models/sale_order.py
@@ -6,25 +6,36 @@
 # Copyright 2018 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    @api.onchange("expected_date")
+    def _onchange_expected_date(self):
+        """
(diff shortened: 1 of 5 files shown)


BernatPForgeFlow:
Still have to update tests

mariadforgeflow (approved):
LGTM! Functional and code review.

sale_order_line_date/__manifest__.py
```diff
@@ -8,7 +8,7 @@
 
 {
     "name": "Sale Order Line Date",
-    "version": "15.0.1.0.1",
+    "version": "15.0.1.0.2",
```
rousseldenis: You don't need to update this unless you don't have migration scripts. It will be done by the merger with ocabot

rousseldenis (approved):

sale_order_line_date/__manifest__.py
BernatPForgeFlow: Ok, I'll change it then

PR #1884, opened 2022-01-21
[13.0][FIX] sale_order_type: Inconsistent data in multi-company environments

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @chienandalu

PR #1894, opened 2022-01-31
[FIX] sale_order_type: Default value check company

hildickethan:
This could pick an incorrect type for the order's company when creating as Superuser

Only started happening after the recent changes in https://github.com/OCA/sale-workflow/pull/1884, I assume it's due to the `check_company` added to the field
After those changes a cron job that created sale orders in multiple companies (as superuser) would end up always picking a `sale_order_type` for the 1st company by default and would error when trying to create any order for other companies

With this change in default it returns to picking an appropriate type

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -25,7 +25,9 @@ class SaleOrder(models.Model):
 
     @api.model
     def _default_type_id(self):
-        return self.env["sale.order.type"].search([], limit=1)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as obvious patch:

/ocabot merge patch

pedrobaeza:
Backported to v13 in 23c5e59f4dfb51296230fffd388ec1b3ee18c29c

PR #1896, opened 2022-02-02
[IMP][PERF] sale_coupon_multi_currency

gurneyalex:
On a loaded database, the computed field currency_id created by this
module is having a measurable performance impact when the
_compute_program_amount method is called.

This PR makes the computed field stored, in order to avoid recomputing
it too often. On the customer instance I'm testing this, I'm saving 60%
of the execution time of _compute_program_amount with this patch
applied.


jcoux (approved):

gurneyalex:
note: this should be merged with a minor version update

rousseldenis (approved):

PR #1904, opened 2022-02-08
[13.0][IMP] change development_status

acsonefho:
Upgrade `development_status` of these modules:
- `sale_order_line_packaging_qty`;
- `sale_by_packaging`.

Because this status wasn't updated for 2 years and we use it in production. IMO these could be upgraded to `"Production/Stable"`


rvalyi (approved):

flotho (approved):
easy change, LGTM

PR #1906, opened 2022-02-08
[12.0][ADD] New module sale_invoiced_state

HekkiMelody:
Hello!

Suggesting a new module `sale_invoiced_state`. It's meant to be roughly equivalent to `sale_delivery_state` (which I also used as a loose template) but for invoices.

It indicates whether a sales order has been invoiced fully, partially or not at all. It's different from the field "Invoice Status" from the base Odoo module because "Invoice Status" mostly shows whether there is anything one can invoice at the moment, and doesn't show if anything has been invoiced already.

This is my first time contributing an entire new module to OCA, as opposed to a fix here and there, so let me know if I need to change anything.

## Screenshots

SO with no invoice
image

SO with downpayment invoice
image

SO Fully Invoiced
image

SO Partially Invoiced
image

The same information is visible in the SO tree view
image

## Main doubts:

- The field's name `invoiced_state` (which is also the module's name) as well as its String attribute "Invoiced". I couldn't find a better alternative that would help clearly distinguish it from the existing "invoice status".
- Technically it only depends on `sale`, but in practice it requires `sale_management`. The same applies to `sale_delivery_state`, so I left it as is. Open to changing the `depends` of course.

If this is eventually accepted, I'd happily provide the equivalent module for purchase-workflow.

Both me and the company I work for sent a signed CLA to OCA.


HekkiMelody:
Not really sure why Runbot is red...

rafaelbn:
> Not really sure why Runbot is red...

https://odoo-community.org/blog/the-oca-blog-1/post/goodbye-runbot-welcome-runboat-116

https://odoo-community.org/groups/contributors-15/contributors-236081?mode=thread&date_begin=&date_end=

rafaelbn:
Usually as easy to review as more reviewers, example: screenshots, short videos, ... 😄 You are welcome!

HekkiMelody:
Thank you for the explanation and guidance @rafaelbn . I've updated the opening message with more information and a couple of screenshots.

sale_invoiced_state/models/sale_order.py
```diff
@@ -0,0 +1,90 @@
...
+from odoo import api, fields, models
+from odoo.tools import float_is_zero, float_compare
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoiced_state = fields.Selection(
+        [
+            ("no", "No"),
```
victor-champonnois: Wouldn't it be better to have more consistent options ? Either "Yes" (instead of "invoiced") and "No", or "Not invoiced" (instead of "No") and "invoiced".

HekkiMelody: My main reasoning was that "no" is pretty unambiguous: There have been no invoices. Whereas "yes" is more ambiguous, since it can be invoiced but only partially. I don't mind changing it, though.

victor-champonnois: Great thank you :)

victor-champonnois (approved):
Looked at the code and tested it locally, LGTM :)

PR #1907, opened 2022-02-10
[ADD][15.0] sale_date_order_editable

emagdalenaC2i:


sale_date_order_editable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 C2i Change 2 improve <soporte@c2i.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale date order editable",
+    "version": "15.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Uncategorized",
```
rafaelbn: Dear @emagdalenaC2i , could you please add the categoty please? this is really useful when searching in odoo-community.org/shop, and other places and I think easy to solve :-)

Example: https://github.com/odoo/odoo/blob/15.0/addons/sale_margin/__manifest__.py#L7

rafaelbn (commented):
I have to test it as I haven't found a problem with da sale order date 😄

sale_date_order_editable/__manifest__.py
emagdalenaC2i: done

emagdalenaC2i:
@rafaelbn The problem is simply that it is read-only and sometimes it is useful to change the date

rafaelbn (approved):
👍

emagdalenaC2i:
@OCA/community-maintainers please, could you add the tag ready to review to this PR?

Edit: Sory, it is already tagged

sale_date_order_editable/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
...
+    date_order = fields.Datetime(
+        string="Order Date",
+        required=True,
+        readonly=False,
+        index=True,
+        states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
+        copy=False,
+        default=fields.Datetime.now,
+        help="Creation date of draft/sent orders,\nConfirmation date of confirmed orders.",
+    )
```
dreispt: Can you strictly redefine only the field attributes that need change?
That makes code smaller and prevents clash with other modules.

emagdalenaC2i: @dreispt done: But I don't know why precommit is red now

dreispt (approved):

rousseldenis:
@emagdalenaC2i Could you explain the use case where you want to edit the order date ? As it is editable in 'draft' and 'sent' states.

https://github.com/odoo/odoo/blob/ebf3d8c9a20515baab2873e4cc529ae8175bb950/addons/sale/models/sale_order.py#L182

emagdalenaC2i:
Hi @rousseldenis 

In some cases, it is useful to edit the order date, for example, after a data migration. However, given the lack of interest in this module, I will not continue with its migration to the newer versions of Odoo.

PR #1925, opened 2022-02-18
14.0 add sale order line last price

dani-xtendoo:
Added sale_line_last_price, which extends the functionality of sale_order_line_price_history, removing filtering by customers


flotho (changes requested):
functionnaly tested and it rocks.
I would just recommend to put your icon next to the unit price so that user will be able to understand what is this behaviour.

sale_order_line_last_price/README.rst
```diff
@@ -0,0 +1,106 @@
...
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales -> Quotations* and select a Quotation.
+#. Click on the new clock button in one of the sale order lines.
+#. A pop-up will open and you will see the *price history* for the product of
+   the sale order line and for the customer of the sale order.
+#. You can select other customer or leave it empty to see the price history for
+   all customers.
```
angelmoya, line 47: This comment is not acording with the code... in code you hide partner field... that also is not ritght

sale_order_line_last_price/wizards/sale_order_line_last_price.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="name">sale.order.line.last.price.view.form</field>
+        <field name="model">sale.order.line.price.history</field>
+        <field name="priority" eval="100" />
+        <field
+            name="inherit_id"
+            ref="sale_order_line_price_history.sale_order_line_price_history_view_form"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//group/field[@name='partner_id']" position="replace">
+            </xpath>
```
angelmoya, line 16: Why do you use replace? Why not to show it and be able to set or unset partner to show related history

angelmoya (changes requested):
Can you explain the behavior for this development?

I checked that you disable customer filter, but this filter is useful.

Also you talk about sale_line_last_price, but I can only see that you disable the original partner filter.

rousseldenis:
@dani-xtendoo What's the status of this ?

PR #1932, opened 2022-02-24
[14.0][ADD] sale_partner_approval: only allow approved customers in Sales Orders

dreispt:


sale_partner_approval/data/exception_rule.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="excep_partner_approved" model="exception.rule">
+        <field name="name">Customer is Approved</field>
+        <field name="description">Customer must be a customer approved for Sales</field>
+        <field name="model">sale.order</field>
+        <field name="exception_type">by_domain</field>
+        <field name="domain">[('partner_id.sale_ok', '=', False)]</field>
+        <field name="is_blocking">True</field>
```
patrickrwilson, line 9: I am not able to install the module due to this line which is causing the following error: 
ValueError: Invalid field 'is_blocking' on model 'exception.rule'
Is this missing a dependancy?

patrickrwilson: my base_exception module was out of date, updated that and will test.

dreispt:
@patrickrwilson 

image

image

patrickrwilson:
@dreispt after testing some more, it works if the contact isn't a child of another, so if you move a child contact to a different stage, the field doesn't update. Is this by design?

dreispt:
Child contacts should inherit the parent's status:
If I can't sell to "ACME Inc", I can't sell to "ACME Inc, John Doe" either.

Does this make sense?
I realize it is not clear in the docs...

dreispt:
@patrickrwilson  -> https://github.com/OCA/sale-workflow/pull/1932/files#diff-be9fb179cf4a00800944f28d42c85f0be1d42e423f96a9d713dc6d9806ee3962R24-R31

patrickrwilson:
@dreispt if that is by design then that is fine, I do think customers may want more flexibility though, because what if you can sell to acme (Parent) and acme_purchaser (Child of Acme) but can't sell to acme_techsupport (Also child of Acme).

dreispt:
@patrickrwilson In that case it would be fine to customize other behaviors by overriding `_compute_sale_ok_product`.

patrickrwilson:
> @patrickrwilson In that case it would be fine to customize other behaviors by overriding `_compute_sale_ok_product`.

I'm fine with that, reviewing now.

patrickrwilson (approved):
LGTM

PR #1935, opened 2022-03-01
[12.0][FIX] sale_product_set: Add invisible field order_id to wizard view

sergio-teruel:
cc @Tecnativa TT34880

If order_if field is not in wizard view you can not select a set with a partner assigned


carlosdauden (approved):

CarlosRoca13 (approved):

sergio-teruel:
not needed in v>=13

PR #1938, opened 2022-03-08
[IMP] sales_team_security: salesman and channel inherited

chienandalu:
We want to propagate any salesman or sales channel change to the entity
children.

cc @Tecnativa TT34055

please check @pedrobaeza @victoralmau

### DIFF
--- a/sales_team_security/models/res_partner.py
+++ b/sales_team_security/models/res_partner.py
@@ -91,3 +91,7 @@ class ResPartner(models.Model):
         if "user_id" in vals or vals.get("type") in {"invoice", "delivery"}:
             self._add_followers_from_salesmans()
         return result
+
+    @api.model
+    def _commercial_fields(self):
+        return super()._commercial_fields() + ["user_id", "team_id"]


pedrobaeza:
Check Travis. And not sure if the method should be that one. There's already code for the propagation AFAIK.

sales_team_security/models/res_partner.py
```diff
@@ -91,3 +91,7 @@ def write(self, vals):
         if "user_id" in vals or vals.get("type") in {"invoice", "delivery"}:
             self._add_followers_from_salesmans()
         return result
+
+    @api.model
+    def _commercial_fields(self):
+        return super()._commercial_fields() + ["user_id", "team_id"]
```
rousseldenis, line 97: This has nothing common with module purpose.

This should be in another module IMHO

rousseldenis (changes requested):

pedrobaeza:
This one is not needed at all, at least not this way.

PR #1940, opened 2022-03-10
[13.0] [FIX] sale_order_lot_selection: Restrict create and edit lots from the lot field in sale order line.

nicomacr:
Before this change any user could crate or edit lots, and for this case in sales the only reason is to select a lot of banch of existing lots, no create one.

### DIFF
--- a/sale_order_lot_selection/view/sale_view.xml
+++ b/sale_order_lot_selection/view/sale_view.xml
@@ -13,6 +13,7 @@
                     domain="[('product_id','=', product_id)]"
                     context="{'default_product_id': product_id}"
                     groups="stock.group_production_lot"
+                    options="{'no_create_edit': True}"
                 />
             </xpath>
             <xpath
@@ -24,6 +25,7 @@
                     domain="[('product_id','=', product_id)]"
                     context="{'default_product_id': product_id}"
                     groups="stock.group_production_lot"
+                    options="{'no_create_edit': True}"
                 />
             </xpath>
         </field>


nicomacr:
@rafaelbn  what do you think about this?

rafaelbn:
Dear @florian-dacosta @bodedra @rousseldenis what do you think? I'm not using this module and I saw you  have recently migrated to v13 or v14. Thank you! 😄

rousseldenis (approved):
Seems legit.

florian-dacosta (changes requested):
I disagree, in my use case, I always create a lot in the sale order.
We even have a module in this repo with this use case : sale_order_lot_generator, this module create the lot automatically.
IMO, this can't be in this module. 
The creation of the lot may make sense in case of manufacturing process where you want to produce a specific lot for the sale order line

rousseldenis:
> I disagree, in my use case, I always create a lot in the sale order. We even have a module in this repo with this use case : sale_order_lot_generator, this module create the lot automatically. IMO, this can't be in this module. The creation of the lot may make sense in case of manufacturing process where you want to produce a specific lot for the sale order line

@nicomacr Maybe through a parameter ?

bodedra (commented):
With the same reason as @florian-dacosta , I disagree with proposed changes.

rousseldenis:
@nicomacr Reading comments, I changed my mind as agree with @florian-dacosta and @bodedra. You should be able to create lots on sale order level.

If you want to disallow it, IMHO, it should be on your side or at least with a parameter here but not deactivating it completely

nicomacr:
Ok  I understood the reasons, let's cancel this PR and i take this on my side.
Thanks.

@rousseldenis @florian-dacosta

PR #1942, opened 2022-03-10
[14.0] [FIX] sale_blanket_order

jado95:
When sale order line is created from blanket order, product's name is not complete. Before, only the product name was taken, now, display_name.

### DIFF
--- a/sale_blanket_order/wizard/create_sale_orders.py
+++ b/sale_blanket_order/wizard/create_sale_orders.py
@@ -100,7 +100,7 @@ class BlanketOrderWizard(models.TransientModel):
     def _prepare_so_line_vals(self, line):
         return {
             "product_id": line.product_id.id,
-            "name": line.product_id.name,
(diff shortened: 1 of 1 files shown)


sale_blanket_order/wizard/create_sale_orders.py
```diff
@@ -100,7 +100,7 @@ def _default_lines(self):
     def _prepare_so_line_vals(self, line):
         return {
             "product_id": line.product_id.id,
-            "name": line.product_id.name,
+            "name": line.product_id.display_name,
```
rousseldenis, line 103: @jado95 Thanks for your contribution.

In order to be more compliant with standard Odoo behaviour, could you adapt it with this : https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py#L1714 ?

Thanks

rousseldenis (changes requested):

Issue #1901, opened 2022-02-04
[14.0] sale_commitment_date_mandatory issue


damdam-s:
Hello,

After some tests, there is an issue with the module sale_commitment_date_mandatory.
It's working pretty well when you create and confirm the sale order the same day or when the commitment date is enough far away.

This case is not covered and can lead to issues:

- security_lead is 1 day
- a SO is created today, as of 2021-10-14, commitment date is 2021-10-15
- the SO is confirmed 1 week later : 2021-10-21
- the picking will still have 2021-10-14 as a scheduled date

2 propositions here to discuss:

1. remove this "alpha" module from the repo
2. keep it but edit the "known bugs" section to mention the failing case

What do you think?

mmequignon:
Could you please create a draft pr with a test that reproduces this error, and link it in the description ?
Thanks for your feedback ! :heart:

jcoux:
@damdam-s @mmequignon 
PR to remove this module: https://github.com/OCA/sale-workflow/pull/1950

sebalix:
I agree, making the `commitment_date` required is breaking tests of other addons in this repository (like here: https://app.travis-ci.com/github/OCA/sale-workflow/jobs/565374862). We should have be an option to enable in the sale configuration panel so we can enable it only for its own tests without disturbing other addons.

sebalix:
I mitigate the issue this addon bring regarding Travis in this PR: https://github.com/OCA/sale-workflow/pull/1969

PR #1950, opened 2022-03-16
[14.0][REMOVE] sale_commitment_date_mandatory

jcoux:
This "alpha" module imply bug:
See: https://github.com/OCA/sale-workflow/issues/1901

Best way is to remove it!


mmequignon:
> 

Better add a note in the readme, saying that there's a bug. There might be someone here that needs it, that could fix it.
What do you think?

jcoux:
> > 
> 
> Better add a note in the readme, saying that there's a bug. There might be someone here that needs it, that could fix it. What do you think?

I will do that!

jcoux:
@mmequignon 
https://github.com/OCA/sale-workflow/pull/1951

PR #1960, opened 2022-03-25
[14.0] [FIX] sale_partner_approval: Fixed Can Sale is lazy updated

Nikul-OSI:
@dreispt Fixed Can Sell is lazy updated.

### DIFF
--- a/sale_partner_approval/data/init_sale_ok.xml
+++ b/sale_partner_approval/data/init_sale_ok.xml
@@ -1,0 +1,9 @@
+<odoo noupdate="1">
+    <!--
+    To not cause disruption upon install, set pre-existing Partners as Can Sale
+    -->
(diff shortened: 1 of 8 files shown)


dreispt (approved):
Reviewed, tested, and fixed tests

sale_partner_approval/data/init_sale_ok.xml
```diff
@@ -0,0 +1,9 @@
+<odoo noupdate="1">
```
flotho, line 1: FMI, wont it be better to use pre init hook? is it still recommended ?

dreispt: This should be equivalent to an init hook, afaics.

flotho (approved):

PR #1964, opened 2022-03-28
[13.0][UPD] product_form_sale_link: New fileds include

manuelcalerosolis:
The columns that are displayed in the product_form_sale_link module are not at all similar to those of the product_form_purchase_link module, which are much more accurate when it comes to displaying the purchase information.

Therefore we have created a view similar to the purchase one.


flotho (approved):

product_form_sale_link/views/sale_order_line.xml
```diff
@@ -2,13 +2,32 @@
 <!-- Copyright 2019 ACSONE SA/NV
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
+    <record id="sale_order_line_tree" model="ir.ui.view">
```
rousseldenis, line 5: The problem with this is that if one want to change the tree view fields, you must depends on this module.

FrancoMaxime (commented):
I agree with Denis

rousseldenis:
@manuelcalerosolis What's the status of this ?

PR #1967, opened 2022-03-29
[14.0][WIP][FIX] fix delivery date

mmequignon:
The last 5 commits are ported from https://github.com/OCA/sale-workflow/pull/1968 (postpone delivery dates for backorders, or when ordered qty is modified)

### DIFF
--- a/sale_delivery_date/models/res_partner.py
+++ b/sale_delivery_date/models/res_partner.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta
 
 from pytz import timezone, utc
 
-from odoo import _, fields, models
+from odoo import _, fields, models, tools
 from odoo.exceptions import UserError
 from odoo.tools.date_utils import date_range
 
@@ -91,6 +91,12 @@ class ResPartner(models.Model):
             if date.weekday() < 5
         ]
 
+    @tools.ormcache("weekday_number")
+    def _get_weekday(self, weekday_number):
+        return self.env["time.weekday"].search(
+            [("name", "=", weekday_number)], limit=1
+        )
+
     def get_next_windows_start_datetime(self, from_datetime, to_datetime):
         """Get all delivery windows start time.
 
@@ -117,9 +123,7 @@ class ResPartner(models.Model):
             from_datetime_tz, to_datetime_tz, timedelta(days=1)
         ):
             this_weekday_number = this_datetime.weekday()
-            this_weekday = self.env["time.weekday"].search(
-                [("name", "=", this_weekday_number)], limit=1
-            )
+            this_weekday = self._get_weekday(this_weekday_number)
             # Sort by start time to ensure the window we'll find will be the first
             # one for the weekday
             this_weekday_windows = self.delivery_time_window_ids.filtered(
--- a/sale_delivery_date/models/res_partner.py
+++ b/sale_delivery_date/models/res_partner.py
@@ -2,9 +2,13 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from datetime import datetime, timedelta
 
+from pytz import timezone, utc
+
 from odoo import _, fields, models
 from odoo.exceptions import UserError
 from odoo.tools.date_utils import date_range
+
+from odoo.addons.partner_tz.tools import tz_utils
 
 
 class ResPartner(models.Model):
@@ -47,6 +51,7 @@ class ResPartner(models.Model):
         if timedelta_days is None:
             timedelta_days = 7
         if self.delivery_time_preference == "workdays":
+            # TODO tz
             datetime_windows = self.get_next_workdays_datetime(
                 from_date, from_date + timedelta(days=timedelta_days)
             )
@@ -69,8 +74,21 @@ class ResPartner(models.Model):
         :param to_datetime: Datetime object
         :return: List of Datetime objects
         """
-        dates = date_range(from_datetime, to_datetime, timedelta(days=1))
-        return [date for date in dates if date.weekday() < 5]
+        # Converting input dates as customer's tz here, as weekday might change
+        # during the conversion
+        tz_string = self.tz or self.env.company.partner_id.tz or "UTC"
+        tz = timezone(tz_string)
+        from_datetime_utc = utc.localize(from_datetime)
+        from_datetime_tz = from_datetime_utc.astimezone(tz)
+        to_datetime_utc = utc.localize(to_datetime)
+        to_datetime_tz = to_datetime_utc.astimezone(tz)
+        dates = date_range(from_datetime_tz, to_datetime_tz, timedelta(days=1))
+        return [
+            # returns naive utc datetimes
+            date.astimezone(utc).replace(tzinfo=None)
+            for date in dates
+            if date.weekday() < 5
+        ]
 
     def get_next_windows_start_datetime(self, from_datetime, to_datetime):
         """Get all delivery windows start time.
@@ -79,13 +97,24 @@ class ResPartner(models.Model):
 
         Note result can include a start datetime that is before from_datetime
         on the from_datetime weekday
+        Note 2 time windows are in the context of customer's tz.
+        We have to localize from_date, to_datetime first, and return a naive
+        UTC datetime.
 
         :param from_datetime: Datetime object
         :param to_datetime: Datetime object
         :return: List of Datetime objects
         """
         res = list()
-        for this_datetime in date_range(from_datetime, to_datetime, timedelta(days=1)):
+        tz_string = self.tz or self.env.company.partner_id.tz or "UTC"
+        tz = timezone(tz_string)
+        from_datetime_utc = utc.localize(from_datetime)
+        from_datetime_tz = from_datetime_utc.astimezone(tz)
+        to_datetime_utc = utc.localize(to_datetime)
+        to_datetime_tz = to_datetime_utc.astimezone(tz)
+        for this_datetime in date_range(
+            from_datetime_tz, to_datetime_tz, timedelta(days=1)
+        ):
             this_weekday_number = this_datetime.weekday()
             this_weekday = self.env["time.weekday"].search(
                 [("name", "=", this_weekday_number)], limit=1
@@ -99,5 +128,11 @@ class ResPartner(models.Model):
                 this_weekday_start_datetime = datetime.combine(
                     this_datetime, win.get_time_window_start_time()
                 )
-                res.append(this_weekday_start_datetime)
+                # this_weekday_start_datetime is a naive datetime in the context
+                # of customer's tz.
+                # Convert is to UTC, then make it naive
+                weekday_start_utc_naive = tz_utils.tz_to_utc_naive_datetime(
+                    tz, this_weekday_start_datetime
+                )
+                res.append(weekday_start_utc_naive)
         return res
--- a/sale_delivery_date/models/res_partner.py
+++ b/sale_delivery_date/models/res_partner.py
@@ -43,9 +43,10 @@ class ResPartner(models.Model):
                                (Leave empty to use 7 days or 1 week)
         :return: Datetime object
         """
-        self.ensure_one()
         if not from_date:
             from_date = datetime.now()
+        if not self:
+            return from_date
         if self.is_in_delivery_window(from_date):
             return from_date
         if timedelta_days is None:
--- a/sale_delivery_date/models/sale_order_line.py
+++ b/sale_delivery_date/models/sale_order_line.py
@@ -7,7 +7,7 @@ from datetime import datetime, time, timedelta
 import pytz
 from pytz import UTC, timezone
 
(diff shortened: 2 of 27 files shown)


sale_delivery_date/models/stock_picking.py
```diff
@@ -148,3 +148,29 @@ def _onchange_scheduled_date(self):
...
+        res = super()._create_backorder()
+        now = fields.Datetime.now()
+        for picking in res:
+            # If the scheduled_date is before the current datetime, then date_deadline
+            # cannot be satisfied. Therefore, we need to recompute move's dates
+            if picking.scheduled_date < now:
+                for line in picking.move_lines:
+                    dates = line._get_delivery_dates(from_date=now)
+                    line.write(dates)
+        return res
```
jbaudoux: This is wrong. You cannot reschedule at backorder creation.
A picking is scheduled based customer/transporter/marketing constrains and if you create a backorder at operational level, the scheduled date has no reason to change.

The use case of rescheduling the open deliveries that could not be processed on the planned day must be tackled differently. For instance, at release channel level.

Can you move this commit out of this PR?

jbaudoux (changes requested):

sale_delivery_date/models/stock_picking.py
jbaudoux: This relates to my remark here that is still valid https://github.com/OCA/sale-workflow/pull/1968#pullrequestreview-1087544016

jbaudoux:
What's the difference between this PR and https://github.com/OCA/sale-workflow/pull/1968 ?

mmequignon:
> What's the difference between this PR and #1968 ?

this refactoring commit, mostly https://github.com/OCA/sale-workflow/pull/1967/commits/88476b53a3a58af2beabe2e0012dff14c3cbda84
And the one moving "all" `sale.order.line` methods to `@api.model`, in order to be used from elsewhere.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -5,176 +5,341 @@
...
+        # date_planned on which we apply the customer's time windows
         res = super()._prepare_procurement_values(group_id=group_id)
-        res = self._cutoff_time_delivery_prepare_procurement_values(res)
-        res = self._warehouse_calendar_prepare_procurement_values(res)
-        res = self._delivery_window_prepare_procurement_values(res)
+        # There's 2 cases here:
+        # 1) commitment date is set, compute date_planned from date_deadline
+        # 2) commitment date isn't set, update date_deadline according to
+        #    cutoff / calendar / delivery window, and compute date_planned
+        if self.order_id.commitment_date:
```
jbaudoux: As you now depend on sale_order_line_date
```suggestion
        if self.commitment_date or self.order_id.commitment_date:
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -5,176 +5,341 @@
...
+        )
+        cutoff = self.order_id.get_cutoff_time()
+        # TODO We should stop here, as the date_planned as defined by odoo represents
+        # the expedition date (date_deadline - security_lead).
+        # In a next refactor, extract the following code in a glue module
+        # between stock_available_to_promise_release and this module.
+        # stock_available_to_promise_release uses the date_planned as the
+        # preparation date before the picking is release, and as the expedition date
+        # after it has been released.
+        # /TODO Put this in the roadmap
```
jbaudoux, line 127: TODO

sale_delivery_date/models/sale_order_line.py
```diff
@@ -5,176 +5,341 @@
...
+        __, security_lead, __ = delays
+        # From there, we know the best delivery date, but we don't want to send
+        # goods to early, or start working too early.
+        # The idea is that, customer might accept deliveries on friday, but the
+        # earliest_work_end might be on monday, which means that we could have
+        # sent goods on Thursday instead (with delivery lead of 1 day).
+        # To avoid this, from the earliest delivery date, we need to retrieve
+        # the various delays applied above in order to find the latest
+        # possible expedition datetime and from it, the latest preparation datetime.
+        # N
```
jbaudoux, line 326:
```suggestion
        #
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -5,176 +5,341 @@
...
+        # would be after latest_expedition_datetime.
+        # expedition_date cannot be after delivery_date
+        # TODO float compare
+        if security_lead > 0:
+            latest_expedition_datetime = self._deduct_delay(
+                delivery_date, security_lead
             )
-        return
+            latest_expedition_date = datetime.combine(
+                latest_expedition_datetime, time.max
```
jbaudoux, line 336: but `time.max` is not in the warehouse timezone. When you call `_get_latest_work_end_from_date_range` below, this is converted to local timezone, and could end-up on a different date

sale_delivery_date/models/sale_order_line.py
```diff
@@ -184,57 +349,72 @@ def _prepare_procurement_values_cutoff_time(
...
+        """Postpone a delivery date according to customer's delivery preferences"""
+        if partner.delivery_time_preference == "anytime":
+            return delivery_date
+        return partner.next_delivery_window_start_datetime(from_date=delivery_date)
+
+    @api.model
+    def _get_latest_work_end_from_date_range(
+        self, earliest_work_end, latest_expedition_date, calendar=False
+    ):
+        """Returns the nearest date in a calendar's attendance within a date range"""
```
jbaudoux, line 425: `_get_latest_work_end_from_date_range` `Returns the nearest date` `_get_closest_work_time`
latest / nearest / closest ?
The docstring should be improved as I found it hard to understand what's happenning
```suggestion
    def _get_latest_work_end_from_date_range(
        self, earliest_work_end, latest_expedition_date, calendar=False
    ):
        """Returns the last open date of the range according to the given calendar of open days"""
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -184,57 +349,72 @@ def _prepare_procurement_values_cutoff_time(
...
+        if calendar:
```
jbaudoux, line 426: A bit easier to read:
```suggestion
        if not calendar:
           return latest_expedition_date
        ...
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -184,57 +349,72 @@ def _prepare_procurement_values_cutoff_time(
...
+            # If there's no working time within the date range, then it returns nothing.
+            timeframe = (earliest_work_end_tz, latest_expedition_date_tz)
+            latest_date_end = calendar._get_closest_work_time(
+                latest_expedition_date_tz, match_end=True, search_range=timeframe
             )
-        )
-        return new_date_planned
+            if latest_date_end:
+                return latest_date_end.astimezone(UTC).replace(tzinfo=None)
+            return earliest_work_end  # TODO latest_expedition_date?
```
jbaudoux, line 440: earliest make sense

jbaudoux (changes requested):
To be continued...

sale_delivery_date/models/res_partner.py
```diff
@@ -47,6 +51,7 @@ def next_delivery_window_start_datetime(self, from_date=None, timedelta_days=Non
         if timedelta_days is None:
             timedelta_days = 7
```
jbaudoux, line 53: 7 is not mutable. Why don't you put this in the signature of the method?

jbaudoux, line 53: 7 is not mutable. Why don't you put this in the signature of the method?

sale_delivery_date/models/res_partner.py
```diff
@@ -47,6 +51,7 @@ def next_delivery_window_start_datetime(self, from_date=None, timedelta_days=Non
...
         if self.delivery_time_preference == "workdays":
+            # TODO tz
```
jbaudoux, line 55: I don't get why we compute all the possibilities while we just need the first correct one.
I would have dropped the call to `is_in_delivery_window` (because this is computing the first element you already compute in each method computing all possibilities in the range) and made this method cover all `delivery_time_preference` possibilities:
```
if self.delivery_time_preference == "anytime":
    return from_date
elif self.delivery_time_preference == "workdays":
    return self._get_first_workday(...)
elif self.delivery_time_preference == "time_windows":
    return self._get_first_timewindow(...)
else:
    raise UserError("Invalid delivery_time_preference")

File: sale_delivery_date/models/sale_order_line.py [hunk 11]
```diff
@@ -184,57 +349,72 @@ def _prepare_procurement_values_cutoff_time(
...
+            # at attendance start
+            return calendar.plan_hours(0, date_start, compute_leaves=True)
+        return date_start
+
+    @api.model
+    def _apply_customer_window(self, delivery_date, partner):
+        """Postpone a delivery date according to customer's delivery preferences"""
+        if partner.delivery_time_preference == "anytime":
+            return delivery_date
+        return partner.next_delivery_window_start_datetime(from_date=delivery_date)
```
jbaudoux, line 418: This should be renamed into `partner._get_first_date_in_delivery_time_preference(from_date=delivery_date)

And the 2 lines above for the anytime case should be moved inside that method

sale_delivery_date/models/res_partner.py
```diff
@@ -79,13 +97,24 @@ def get_next_windows_start_datetime(self, from_datetime, to_datetime):
...
-        for this_datetime in date_range(from_datetime, to_datetime, timedelta(days=1)):
+        tz_string = self.tz or self.env.company.partner_id.tz or "UTC"
+        tz = timezone(tz_string)
+        from_datetime_utc = utc.localize(from_datetime)
+        from_datetime_tz = from_datetime_utc.astimezone(tz)
+        to_datetime_utc = utc.localize(to_datetime)
+        to_datetime_tz = to_datetime_utc.astimezone(tz)
+        for this_datetime in date_range(
+            from_datetime_tz, to_datetime_tz, timedelta(days=1)
+        ):
```
jbaudoux, line 124: This method could be simplified as we just need to find the first day for which a time window is defined. 
We don't care on the time.

sale_delivery_date/models/sale_order_line.py
mmequignon: <s>Mah, we shouldn't need that, as order's commitment date is computed based on its line's commitment_dates.</s>

I was wrong this is not the case...
I think we should drop the dependency anyway.
@sebalix why did we added this `sale_order_line_date` dep? I'm not sure to understand why, according to this commit. Could we drop it?
https://github.com/OCA/sale-workflow/pull/1967/commits/7b68a2e7d69de05fc12e4d345b74ea25121fc55a

sale_delivery_date/models/stock_picking.py
simahawk: solved?

sale_delivery_date/models/res_partner.py
grindtildeath: Actually it's there already.

rousseldenis:
@mmequignon What's the status of this ?

jbaudoux:
> @mmequignon What's the status of this ?

@rousseldenis We are redoing it differently in v16

mmequignon:
> > @mmequignon What's the status of this ?
> 
> @rousseldenis We are redoing it differently in v16

@jbaudoux Not quite. Having another customer using release channels for this doesn't invalidate this module.

@rousseldenis This is currently used a lot by one of our customers, they did lots of tests before this version has been deployed in product. IMO, this is good to go. I have to find time to fix those tests.

rousseldenis:
> > > @mmequignon What's the status of this ?
> > 
> > 
> > @rousseldenis We are redoing it differently in v16
> 
> @jbaudoux Not quite. Having another customer using release channels for this doesn't invalidate this module.
> 
> @rousseldenis This is currently used a lot by one of our customers, they did lots of tests before this version has been deployed in product. IMO, this is good to go. I have to find time to fix those tests.

Should be great in order to merge this.

sale_delivery_date/models/sale_order_line.py
sebalix: Because of this (copy/paste):

> **[FIX] sale_delivery_date: add dependency]**
> 
> Add `sale_order_line_date` as dependency because this addon is overriding the same methods (`_onchange_commitment_dat`e and `_prepare_procurement_values`) and we want to ensure than `sale_delivery_date` takes the precedence over `sale_order_line_date`.
> 
> Also, the other feature of `sale_order_line_date` which sync the commitment_date between SO and its lines is welcome in `sale_delivery_date`.

https://github.com/OCA/sale-workflow/pull/1968

If you think it is not needed anymore, feel free to remove it.

rousseldenis:
@mmequignon

jbaudoux:
Rewritten with a new modular approach on v16.0
https://github.com/OCA/wms/pull/1022

PR #1968, opened 2022-03-30
[WIP][14.0][FIX] sale_order_line_date & sale_delivery_date: fixes

sebalix:
Depends on this PR to get Travis happy regarding the addons installed:
- https://github.com/OCA/sale-workflow/pull/1969

## [FIX] sale_order_line_date: fix commitment_date onchange]

In 14.0 there are two methods named `_onchange_commitment_date`:

1) one which set the `commitment_date` with the `expected_date` value
   (triggered by `expected_date`)
2) the second which triggers a warning (triggered by `commitment_date`
   only)

The first one is never executed because the second has the precedence,
and it is a good thing because we do not want to automatically set the
`commitment_date` by default!
Odoo has fixed this code in 15.0 by removing the first method and added
the `expected_date` field as trigger on the second one
(see https://github.com/odoo/odoo/commit/4031e86f2564378d18f7a930ad45f178c71b2939).

This commit is adding `expected_date` as well on this overridden method
so we are in line with the expected behavior.

## [FIX] sale_delivery_date: add dependency]

Add `sale_order_line_date` as dependency because this addon is
overriding the same methods (`_onchange_commitment_date` and
`_prepare_procurement_values`) and we want to ensure than
`sale_delivery_date` takes the precedence over `sale_order_line_date`.

Also, the other feature of `sale_order_line_date` which sync the
`commitment_date` between SO and its lines is welcome in
`sale_delivery_date`.

## [FIX] sale_delivery_date: take into account 'date_deadline' field

TODO
- [x] compute the `date_planned` and `date_deadline` if `commitment_date` is not provided
- [ ] compute the `date_planned` from the `date_deadline/commitment_date` if any (reverse)

### DIFF
--- a/sale_delivery_date/models/sale_order_line.py
+++ b/sale_delivery_date/models/sale_order_line.py
@@ -185,10 +185,13 @@ class SaleOrderLine(models.Model):
             )
             next_working_day = self._next_working_day(next_preferred_date, calendar)
             count += 1
-        if date_done != next_preferred_date:
-            # TODO : Add logs ?
-            return next_preferred_date
-        return next_preferred_date
+        # Now that next_working_day.date() equals next_preferred_date.date(),
+        # goods can't leave the warehouse outside of working hours
+        preferred_date = max(next_preferred_date, next_working_day)
+        # If scheduled date and deadline are the same day, we promise the latest date
+        if date_planned.date() == preferred_date.date():
+            preferred_date = max(date_planned, preferred_date)
+        return preferred_date
 
     @api.model
     def _next_working_day(self, date_, calendar=None):
--- a/sale_delivery_date/models/sale_order_line.py
+++ b/sale_delivery_date/models/sale_order_line.py
@@ -18,6 +18,15 @@ FIND_WORKING_DAY_COUNT = 10  # To avoid infinite loops, could be increased
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    def _action_launch_stock_rule(self, previous_product_uom_qty=False):
+        # Add a context key, so we know in `_prepare_procurement_values` that we
+        # might have to update the date based on today instead of the order date
+        if previous_product_uom_qty:
+            self = self.with_context(updated_line_qty=True)
+        return super(SaleOrderLine, self)._action_launch_stock_rule(
+            previous_product_uom_qty
+        )
+
     def _prepare_procurement_values(self, group_id=False):
         # Update 'date_planned' and 'date_deadline' with respect to:
         #   - the warehouse or partner cutoff time
@@ -34,6 +43,12 @@ class SaleOrderLine(models.Model):
             res = self._prepare_procurement_values_commitment_date(res)
         else:
             res = self._prepare_procurement_values_no_commitment_date(res)
+        # If `updated_line_qty` context key is set, and now is after the computed
+        # date planned, then we're late, and should postpone the delivery dates
+        now = datetime.now()
+        late = res["date_planned"] <= now
+        if self.env.context.get("updated_line_qty") and late:
+            res = self._prepare_procurement_values_no_commitment_date(res, now)
         return res
 
     def _prepare_procurement_values_commitment_date(self, res):
@@ -55,12 +70,18 @@ class SaleOrderLine(models.Model):
         )
         return res
 
-    def _prepare_procurement_values_no_commitment_date(self, res):
-        """Set 'date_planned' and 'date_deadline' if no 'commitment_date'."""
+    def _prepare_procurement_values_no_commitment_date(self, res, date_from=None):
+        """Set 'date_planned' and 'date_deadline' if no 'commitment_date'.
+
+        If date_from is set, use it as the starting date to compute delivery dates,
+        otherwise, use order's date_order.
+        """
         customer_lead, security_lead, workload = self._get_delays()
         workload_days = self._delay_to_days(workload)
+        if not date_from:
+            date_from = self.order_id.date_order
         date_planned = (
-            self.order_id.date_order
+            date_from
             + timedelta(days=customer_lead or 0.0)
             - timedelta(days=security_lead)
         )
--- a/sale_delivery_date/models/sale_order_line.py
+++ b/sale_delivery_date/models/sale_order_line.py
@@ -12,6 +12,8 @@ from odoo.addons.partner_tz.tools import tz_utils
 
 _logger = logging.getLogger(__name__)
 
+FIND_WORKING_DAY_COUNT = 10  # To avoid infinite loops, could be increased
+
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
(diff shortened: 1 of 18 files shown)


mmequignon (approved):

TDu (approved):

sale_delivery_date/models/sale_order_line.py
```diff
@@ -17,13 +17,53 @@ class SaleOrderLine(models.Model):
...
+        # 4) apply cutoff (with `keep_same_day` param)
+        return res
+
+    def _prepare_procurement_values_no_commitment_date(self, res):
+        """Set 'date_planned' and 'date_deadline' if no 'commitment_date'."""
+        res["date_planned"] = (
+            self.order_id.date_order
+            + timedelta(days=self.customer_lead or 0.0)
+            - timedelta(days=self.order_id.company_id.security_lead)
+        )
```
mmequignon: `res["date_planned"]` should already be the delivery date without the security lead, right?

sale_delivery_date/models/sale_order_line.py
```diff
@@ -55,66 +96,40 @@ def _warehouse_calendar_prepare_procurement_values(self, res):
...
-        # If no commitment date is set, we must consider next preferred delivery
-        #  window to postpone date_planned
+        """Return 'date_deadline' according to customer's time windows.
 
-        # Remove security lead time to ensure the delivery date (and not the
-        # date planned of the picking) will match delivery windows
-        date_planned_without_sec_lead = date_planned + timedelta(
-            days=self.order_id.company_id.security_lead
-        )
+        This computation is called only if no commitment_date is set.
```
mmequignon: and if customer's delivery time preference is `time_windows`

sale_delivery_date/models/sale_order_line.py
sebalix: Yes, indeed. In 14.0 the computations are the same than for 13.0 actually (done by the call to `super` right before), but now Odoo can store the effective delivery date in the `date_deadline` field, and the `date_planned` stays what it is (the date when operators could start working on the transfers):
```
# v13
date_planned = date_order + self.customer_lead - security_lead
if commitment_date:
    date_planned = commitment_date - security_lead

# v14:
date_deadline = commitment_date or (date_order + self.customer_lead)
date_planned = date_deadline - security_lead
```
I could even avoid these lines to compute the `date_planned`, it's already done by `super()`... just wanted to be clear about the intent.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -41,13 +41,21 @@ def _prepare_procurement_values_commitment_date(self, res):
...
+        date_transfer_done = res["date_deadline"] - timedelta(
+            days=self.order_id.company_id.security_lead
+        )
+        res["date_planned"] = date_transfer_done
+        # 2) Find the first date in the WH calendar (by going back in the past)
+        calendar = self.order_id.warehouse_id.calendar_id
+        if calendar:
+            res["date_planned"] = calendar.plan_days(
+                -1, res["date_planned"], compute_leaves=True
+            )
```
mmequignon: is this really working? :open_mouth:

sebalix: I didn't take time to test it in a real use case or UT, but after checking the implementation of this method and a quick test in a shell, seems to work as expected yes:
```python
>>> from odoo import fields
>>> calendar = env["stock.warehouse"].browse(1).calendar_id  # Weekend are days off in this calendar
>>> monday = fields.Datetime.from_string("2022-04-04")
>>> calendar.plan_days(-1, monday, compute_leaves=True)
datetime.datetime(2022, 4, 1, 11, 0)  # Friday 1st April
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -43,129 +43,79 @@ def _prepare_procurement_values_commitment_date(self, res):
...
-    def _get_date_planned_with_cutoff_time(self, date_planned, keep_same_day=False):
-        """Apply the cut-off time on a planned date
-
-        The cut-off configuration is taken on the partner if set, otherwise
-        on the warehouse.
+    @api.model
+    def _deduce_workload_and_security_lead(
+        self, date, partner, warehouse, days, security_lead
+    ):
+        """Return the 'date_planned' from the 'date_deadline'.
```
sebalix: Could we rename the `date` parameter to `date_deadline`? The docstring says we expected a date deadline as parameter but better to name this parameter accordingly as the name of the method doesn't tell from what it deduce the workload and security lead.
Or if you want to be agnostic about the kind of date we could process with this method, update the docstring.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -43,129 +43,79 @@ def _prepare_procurement_values_commitment_date(self, res):
...
-        # As the 'date_planned' is the date when the work can start on the
-        # transfer, we have to add the security lead of the company to get
-        # the date indicating when the transfer will be ready to be shipped.
-        # From this date we will be able to compute the preferred delivery date
-        # of the customer.
-        date_transfer_done = date_planned + timedelta(
+    @api.model
+    def _apply_delivery_window(self, date_planned, partner, calendar=None):
+        date_done = date_planned + timedelta(
             days=self.order_id.company_id.security_lead
```
sebalix: /!\ You are making use of `self.order_id` in an `@api.model` method, we should also send the security lead or the order as parameter then

sale_delivery_date/models/sale_order_line.py
```diff
@@ -43,129 +43,79 @@ def _prepare_procurement_values_commitment_date(self, res):
...
+        # Remove the security lead from the date_deadline
+        date_planned = date - timedelta(days=security_lead)
+        # Deduce the workload
+        date_planned = self._apply_workload(date_planned, -days, warehouse.calendar_id)
+        # the correct day has already been computed, only change the cut-off time
+        return self._apply_cutoff(date_planned, partner, warehouse, keep_same_day=True)
+
+    @api.model
+    def _apply_cutoff(self, date, partner, warehouse, keep_same_day=False):
+        cutoff = self.order_id.get_cutoff_time(partner, warehouse)
```
sebalix: Same: `self.order_id` with `@api.model`, send the `order` as parameter?

mmequignon: No, this should be independent from the `sale.order*` models since, as you will soon see, those methods will be called from `stock.{move,picking}` soon

mmequignon: Also, this `get_cutoff_time` is an `api.model` method now.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -87,11 +89,38 @@ def _deduce_workload_and_security_lead(
...
+                if count > FIND_WORKING_DAY_COUNT:  # To avoid infinite loop
+                    raise exceptions.UserError(
+                        _(
+                            "Unable to find a working day matching "
+                            "customer's delivery time window."
+                        )
+                    )
+                date_planned -= timedelta(days=1)
+                next_working_day = self._next_working_day(date_planned, calendar)
+                count += 1
```
sebalix, line 131: This part is almost duplicated from `_apply_delivery_window`, could be cool to put the common part in a method, not blocking.

mmequignon: My thought exactly.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -160,13 +160,13 @@ def _apply_workload(self, date, days, calendar=None):
     def _apply_delivery_window(
         self, date_planned, partner, security_lead, calendar=None
     ):
-        date_done = date_planned + timedelta(days=security_lead)
+        datetime_done = date_planned + timedelta(days=security_lead)
+        # Only the date here is relevant, both to find out if this is a working day
+        # and to apply the delivery window on top of it.
+        date_done = datetime_done.replace(hour=0, minute=0, second=0, microsecond=0)
```
sebalix, line 166: Could be written
```python
date_done = datetime.combine(datetime_done, time.min)
```
With `time` coming with `from datetime import time`.

sebalix:
Need to merge this soon as it's blocking other PRs like https://github.com/OCA/wms/pull/355

lmignon (approved):
LGTM (Code review only)

rousseldenis:
@sebalix Still WIP ?

sebalix:
@dreispt we need to do some cleanup in the commit history first, and fix the tests, but this is already used in production.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -136,97 +213,33 @@ def _get_delays(self):
...
+        # Overwritten to compute the expected date with respect to:
+        #   - the warehouse or partner cutoff time
+        #   - the warehouse calendar
+        #   - the delivery time window of the customer
+        customer_lead, security_lead, workload = self._get_delays()
+        date_planned = (
+            (self.order_id.date_order or fields.Datetime.now())
+            + timedelta(days=customer_lead or 0.0)
+            - timedelta(days=security_lead)
+        )
```
jbaudoux, line 225: According to me, the leads should be removed
```suggestion
        date_planned = self.order_id.date_order or fields.Datetime.now()
```
workload = customer_lead - security_lead
So you apply twice the workload, once here and once below taking care of the calendar.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+        """
+        customer_lead, security_lead, workload = self._get_delays()
+        workload_days = self._delay_to_days(workload)
+        if not date_from:
+            date_from = self.order_id.date_order
+        date_planned = (
+            date_from
+            + timedelta(days=customer_lead or 0.0)
+            - timedelta(days=security_lead)
         )
```
jbaudoux, line 87: According to me, the lead should be removed
```suggestion
        date_planned = date_from
```
workload = customer_lead - security_lead
So you apply twice the workload, once here and once below taking care of the calendar.

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+        # internally and 'date_deadline' is when the goods will be delivered to
+        # the customer.
+        # There's 2 cases here:
+        #   1) commitment_date is set, compute date_planned from date_deadline
+        #   2) commitment_date isn't set, compute date_planned and date_deadline
         res = super()._prepare_procurement_values(group_id=group_id)
-        res = self._cutoff_time_delivery_prepare_procurement_values(res)
-        res = self._warehouse_calendar_prepare_procurement_values(res)
-        res = self._delivery_window_prepare_procurement_values(res)
+        if self.order_id.commitment_date:
```
jbaudoux, line 42: As you now depend on `sale_order_line_date`
```suggestion
        if self.commitment_date or self.order_id.commitment_date:
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
-            # if we have a commitment date, even if we are too late, respect
-            # the original planned date (but change the time), the transfer
-            # will be considered as "late"
-            keep_same_day=bool(self.order_id.commitment_date),
+    def _prepare_procurement_values_commitment_date(self, res):
+        """Set 'date_planned' according to 'commitment_date'.
+
+        We want to find the date when we could start working on the transfers
+        so we will be able to ship the goods to the customer in respect to the
+        `commitment_date`.
```
jbaudoux, line 59:
```suggestion
        We want to find the date when we have to ship the goods to the customer
        in respect to the `commitment_date`.
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+        The 'date_deadline' is already set to the 'commitment_date' (that should
+        preferably fit the partner's time window if any).
+        """
+        warehouse = self.order_id.warehouse_id
+        partner = self.order_id.partner_id
+        __, security_lead, workload = self._get_delays()
+        worload_days = self._delay_to_days(workload)
+        # Find the first date in the WH calendar by going back in the past, if any.
+        res["date_planned"] = self._deduce_workload_and_security_lead(
+            res["date_deadline"], partner, warehouse, worload_days, security_lead
```
jbaudoux, line 69:
```suggestion
        calendar = warehouse.calendar2_id
        date_planned = res["date_deadline"] - timedelta(days=security_lead)
        if calendar:
            # The commitment time does not matter when looking at the calendar open days
            # TODO: Ideally, we should set the time to end of day
            date_planned = self._apply_cutoff(date_planned, partner, warehouse, keep_same_day=True)
            # Looking backward for the first open day to deliver the parcel
            date_planned = calendar.plan_hours(-1, date_planned, compute_leaves=True)
        # Set the time to the cutoff time
        date_planned = self._apply_cutoff(
            date_planned, partner, warehouse, keep_same_day=True
        )
        res["date_planned"] = date_planned
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+                if count > FIND_WORKING_DAY_COUNT:  # To avoid infinite loop
+                    raise exceptions.UserError(
+                        _(
+                            "Unable to find a working day matching "
+                            "customer's delivery time window."
+                        )
+                    )
+                date_planned -= timedelta(days=1)
+                next_working_day = self._next_working_day(date_planned, calendar)
+                count += 1
```
jbaudoux, line 131: If the time is not a working time, your implementation won't work.
This should do what you want and the time must be carefully set before
```suggestion
            date_planned = calendar.plan_hours(-1, date_planned, compute_leaves=True)
```

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
-        # Remove security lead time to ensure the delivery date (and not the
-        # date planned of the picking) will match delivery windows
-        date_planned_without_sec_lead = date_planned + timedelta(
-            days=self.order_id.company_id.security_lead
+        partner = self.order_id.partner_shipping_id
+        warehouse = self.order_id.warehouse_id
+        calendar = warehouse.calendar_id
+        date_planned = self._apply_cutoff(date_planned, partner, warehouse)
+        date_planned = self._apply_workload(date_planned, workload_days, calendar)
+        date_deadline = self._apply_delivery_window(
```
jbaudoux, line 93: For readability, I would add the security here and not in delivery_window as it is a separate step and not relevant to the method name

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+            date_planned, partner, security_lead, calendar
         )
-        ops = self.order_id.partner_shipping_id
-        next_preferred_date = ops.next_delivery_window_start_datetime(
-            from_date=date_planned_without_sec_lead
+        date_planned = self._deduce_workload_and_security_lead(
```
jbaudoux, line 96: Now that you know when it can reach the customer, you only have to remove the security and set the cutoff time (a separate method for this would be nice. That method just needs to call _apply_cutoff(same_day))

sale_delivery_date/models/sale_order_line.py
```diff
@@ -6,124 +6,201 @@
...
+            date_deadline, partner, warehouse, workload_days, security_lead
         )
-        # Add back security lead time
-        next_preferred_date_with_sec_lead = next_preferred_date - timedelta(
-            days=self.order_id.company_id.security_lead
+        res.update({"date_deadline": date_deadline, "date_planned": date_planned})
+        return res
+
+    @api.model
+    def _deduce_workload_and_security_lead(
```
jbaudoux, line 103: I would drop this method, I don't see a reason for removing the workload

sale_delivery_date/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+        self.ensure_one()
+        if not from_date:
+            from_date = fields.Datetime.now()
+        partner = self.picking_id.partner_id
+        warehouse = self.picking_id.picking_type_id.warehouse_id
+        calendar = warehouse.calendar_id
+        customer_lead, security_lead, workload = self._get_delays()
+        sale_line_model = self.env["sale.order.line"]
+        workload_days = sale_line_model._delay_to_days(workload)
+        # Those steps are the same as in sale_order_line.py,
```
jbaudoux, line 30: Then probably also wrong

jbaudoux (changes requested):
The procurement created from the sale order express a need to ship goods at a given date.
There is a confusion in the implementation like this is when the pick/ship process must start, but it is in fact when it ends.

sale_delivery_date/models/stock_picking.py
```diff
@@ -22,6 +22,18 @@ class StockPicking(models.Model):
     )
     expected_delivery_date = fields.Datetime(compute="_compute_expected_delivery_date")
 
+    def _create_backorder(self):
+        res = super()._create_backorder()
+        now = fields.Datetime.now()
+        for picking in res:
+            # If the scheduled_date is before the current datetime, then date_deadline
+            # cannot be satisfied. Therefore, we need to recompute move's dates
+            if picking.scheduled_date < now:
```
jbaudoux, line 31: * This should be only for outgoing picking
* Only moves where date < now should be updated as you can have different dates on each moves
* Why computing each move separately? They all belong to the same warehouse and for the same partner, so the result should be the same for all the moves
* most importantly, I don't understand why the date should change on the backorder creation. It could still be processed the same day

sale_delivery_date/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+        # in the `_prepare_procurement_values_no_commitment_date` method,
+        # which means that we compute both date_planned and date_deadline based
+        # on `from_date`:
+        #   1) apply cutoff (partner or warehouse cutoff, depending on the config)
+        #   2) apply the workload
+        #   3) apply partner's time window to get the date deadline
+        #   4) deduce security_lead and the workload to get the date planned
+        date_planned = sale_line_model._apply_cutoff(from_date, partner, warehouse)
```
jbaudoux, line 38: Doesn't make sense. The customer delivery date should not depend on when the backorder has been created, before or after the cutoff. The cutoff is applied based on when the order is placed.

sale_delivery_date/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+        date_planned = sale_line_model._apply_workload(
```
jbaudoux, line 39: Only make sense if the from_date is when the picking process start. So valid when you pass now()
Add a comment explaining what from_date represent to clarify this.

sale_delivery_date/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+            date_planned, workload_days, calendar
+        )
+        date_deadline = sale_line_model._apply_delivery_window(
+            date_planned, partner, security_lead, calendar
+        )
+        date_planned = sale_line_model._deduce_workload_and_security_lead(
+            date_deadline, partner, warehouse, workload_days, security_lead
+        )
```
jbaudoux, line 47: Add security, apply delivery window, remove security
Don't remove the workload as you compute the outgoing date

jbaudoux (changes requested):
A cron calling `_get_delivery_dates` to update what was planned for previous days and not processed would make sense but not when a backorder is created.

sale_delivery_date/models/sale_order_line.py
jbaudoux: In fact, the order date & time may matter in regards of the cutoff time if not enough in the future

rousseldenis:
@sebalix @i-vyshnevska What's the status of this ?

jbaudoux:
> @sebalix @i-vyshnevska What's the status of this ?

@rousseldenis It has been replaced by https://github.com/OCA/sale-workflow/pull/1967 which in turn is being approached differently in v16

PR #1969, opened 2022-03-31
[14.0][FIX] sale_commitment_date_mandatory: remove NOT NULL constraint

sebalix:
This commit aims to make the commitment date required through an opt-in option on the `Sale` configuration panel:

- remove the NOT NULL constraint on `commitment_date` field
- add an option 'Delivery commitment date mandatory' on 'Sale'  configuration panel
- make the field required on SO form only if the option is enabled

The tests of other addons can fail on Travis (which install all addons) currently because as this field is required and is getting a default value, this is messing up the results of tests.

ping @damdam-s @jcoux @mmequignon


mmequignon (approved):

damdam-s (approved):
thanks

sale_commitment_date_mandatory/__manifest__.py
```diff
@@ -3,16 +3,17 @@
 
 {
     "name": "Sale commitment date mandatory",
-    "version": "14.0.1.0.1",
+    "version": "14.0.1.1.0",
```
dreispt, line 6: Please don't edit version numbers manually.

sebalix: When there is a related migration script I think it's better to edit this one manually. If a conflict appears, we would have to rename the migration script accordingly to match the version it fixes.

sebalix:
Thanks!

PR #1979, opened 2022-04-12
[IMP] sale_invoice_plan: remove dependency

gurneyalex:
remove the dependency on sale_stock which does not seem to be required.
this allows to use the module in service companies which don't need
the stock application


kittiu (approved):

rousseldenis (approved):

ivantodorovich (approved):

kittiu (approved):

grindtildeath:
There is still an XMLID referencing a record from stock modules which needs to be removed:
https://github.com/OCA/sale-workflow/blob/15.0/sale_invoice_plan/security/ir.model.access.csv#L5

yankinmax:
Opened PR to add separate module `sale_stock_invoice_plan` depending on `sale_invoice_plan`
https://github.com/OCA/sale-workflow/pull/2009

PR #1980, opened 2022-04-12
[13.0][IMP] sale_order_product_recommendation: Exclude delivery carrier products from recommendations

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @CarlosRoca13 @LudLaf 

TT35965

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -75,12 +75,18 @@ class SaleOrderRecommendation(models.TransientModel):
                 ]
             )
         )
-        return [
+        domain = [
             ("order_id", "in", (other_sales - self.order_id).ids),
             ("product_id.active", "=", True),
             ("product_id.sale_ok", "=", True),
             ("qty_delivered", "!=", 0.0),
         ]
+        # Exclude delivery products
+        # We can not use the method _is_delivery() from sale module because we are
+        # doing a domain for a readgroup query
+        if "is_delivery" in self.env["sale.order.line"]._fields:
+            domain.append(("is_delivery", "=", False))
+        return domain
 
     def _prepare_recommendation_line_vals(self, group_line, so_line=False):
         """Return the vals dictionary for creating a new recommendation line.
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

pedrobaeza (commented):
Why not directly exclude SO lines with `is_delivery` field set to True?

sergio-teruel:
@pedrobaeza Changes done!!

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -75,12 +75,16 @@ def _recomendable_sale_order_lines_domain(self):
...
         )
-        return [
+        domain = [
             ("order_id", "in", (other_sales - self.order_id).ids),
             ("product_id.active", "=", True),
             ("product_id.sale_ok", "=", True),
             ("qty_delivered", "!=", 0.0),
         ]
+        # Exclude delivery products
+        if "delivery.carrier" in self.env.registry:
```
pedrobaeza:
```suggestion
        if "is_delivery" in self.env["sale.order.line"]._fields:
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -128,8 +132,9 @@ def _generate_recommendations(self):
         found_dict = {l["product_id"][0]: l for l in found_lines}
         recommendation_lines = self.env["sale.order.recommendation.line"]
         existing_product_ids = set()
-        # Always recommend all products already present in the linked SO
-        for line in self.order_id.order_line:
+        # Always recommend all products already present in the linked SO except delivery
+        # carrier products
+        for line in self.order_id.order_line.filtered(lambda ln: not ln.is_delivery):
```
pedrobaeza:
```suggestion
        lines = self.order_id.order_line
        if "is_delivery" in self.env["sale.order.line"]._fields:
            lines = lines.filtered(lambda ln: not ln.is_delivery)
        for line in lines:
```

pedrobaeza (changes requested):

rousseldenis:
Mmmh, don't like code based on unloaded things. The best is to do a glue module...

pedrobaeza:
Well, it's a good trade-off. We know that it's not the perfect one, but have a glue module `*_delivery` for every module we have... Did you contribute in Odoo a hook method for excluding SO lines that is hooked by `delivery` module? I remember something like that. That can be a good alternative.

sergio-teruel:
Hi
We can not use the method _is_delivery() from sale module because we are doing a domain for a readgroup query.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: This one is missing

sergio-teruel: Here we can use method _is_delivery() from sale module

sergio-teruel:
Changes done!

pedrobaeza (approved):

PR #1983, opened 2022-04-14
[14.0][MIG] sale_manual_delivery

ChrisOForgeFlow:
Standard Migration
@ForgeFlow


francesco-ooops:
@ChrisOForgeFlow can you rebase?

sale_manual_delivery/hook.py
```diff
@@ -0,0 +1,64 @@
+def pre_init_hook(cr):
+    cr.execute(
+        """
+        ALTER TABLE sale_order_line ADD COLUMN IF NOT EXISTS qty_procured numeric;
+        COMMENT ON COLUMN sale_order_line.qty_procured IS 'Quantity Procured';
```
legalsylvain, line 5: Naïve question why to add comment on postgres field ?

legalsylvain (commented):
Functional question : how to enable the feature if we dont use crm.team ?

sale_manual_delivery/hook.py
ChrisOForgeFlow: In this case I do it like this, to reproduce the native behavior of odoo when creating a new column

ChrisOForgeFlow:
> Functional question : how to enable the feature if we dont use crm.team ?

I think it would be better to put a default true when installing this module, but when migrating it I kept that behavior

sale_manual_delivery/wizard/manual_delivery_line.py
```diff
@@ -0,0 +1,56 @@
...
+        if any(
+            float_compare(
+                line.quantity,
+                line.qty_ordered - line.qty_procured,
+                precision_rounding=line.product_id.uom_id.rounding,
+            )
+            > 0.00
+            for line in self
+        ):
+            raise UserError(
```
rousseldenis: You should raise ValidationError in constraints

sale_manual_delivery/wizard/manual_delivery.xml
```diff
@@ -0,0 +1,63 @@
...
+                        name="confirm"
+                        string="Confirm"
+                        class="btn-primary"
+                    />
+                    <button string="Cancel" class="btn-default" special="cancel" />
+                </footer>
+            </form>
+        </field>
+    </record>
+    <act_window
```
rousseldenis: This is deprecated

sale_manual_delivery/wizard/manual_delivery.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    date_planned = fields.Datetime(string="Date Planned")
+
+    def confirm(self):
+        """Creates the manual procurements"""
+        self.ensure_one()
+        sale_order_lines = self.line_ids.mapped("order_line_id")
+        sale_order_lines.with_context(
+            sale_manual_delivery=self
+        )._action_launch_stock_rule_manual()
```
youring, line 95: From a functional user perspective: Here can we return to the picking form like in purchase_manual_delivery? https://github.com/OCA/purchase-workflow/blob/874c7d7acbac075481a1a41578cb37beb0c7eaa0/purchase_manual_delivery/wizard/create_manual_stock_picking.py#L158

sale_manual_delivery/wizard/manual_delivery_line.py
ChrisOForgeFlow: done!

sale_manual_delivery/wizard/manual_delivery.xml
ChrisOForgeFlow: done!

sale_manual_delivery/wizard/manual_delivery.py
ChrisOForgeFlow: It seems to me that this feature is more difficult to implement here, because the picking is not built in this context, the stock rules will take care of it and the result can be very varied, several pickings, it seems to me that it would be better to do an independent implementation for that case, in the other module it is more precise because the picking is created by the same assistant

youring: Thanks for your explanation. IMHO maybe it more ux friendly to return something, notify user that draft delivery has been created, or simply to the picking list.

youring:
In runboat, I created a sale order with 5 pieces Large Cabinet, set Manual Delivery to True, confirm order.  Then create manual delivery of 3 pieces, partial delivery. The "Create Delivery" button is gone, not able to create delivery of the other 2 later.

nguyenhk99:
> In runboat, I created a sale order with 5 pieces Large Cabinet, set Manual Delivery to True, confirm order. Then create manual delivery of 3 pieces, partial delivery. The "Create Delivery" button is gone, not able to create delivery of the other 2 later.

Hello, I created a PR to address this issue (button disappearing after partial delivery) https://github.com/ForgeFlow/sale-workflow/pull/5. @ChrisOForgeFlow  can you check it?

leemannd (approved):

PR #1992, opened 2022-04-22
[15.0][mig] sale_product_seasonality

damdam-s:
set installable=False because of pending refactoring https://github.com/OCA/product-attribute/pull/1063


damdam-s:
@simahawk something like this one? and now I create a new PR with the refactoring?

sale_product_seasonality/__manifest__.py
```diff
@@ -1,9 +1,9 @@
-# Copyright 2021 Camptocamp SA
+# Copyright 2021 - 2022 Camptocamp SA
```
simahawk: no need to add current year

sale_product_seasonality/__manifest__.py
```diff
@@ -1,9 +1,9 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Sale Product seasonality",
-    "summary": "Integrates rules for products' seasonal availability with sales",
-    "version": "14.0.1.1.0",
+    "summary": "Integrates rules for products' date availability with sales",
+    "version": "15.0.1.0.0",
```
simahawk: if we don't keep the name is pointless to change this number

simahawk (changes requested):
yes, no need for any migration process tho: just port the module AS IS from v14 and set it as not installable (like when we used to have pre-initialized repos)

rousseldenis:
@damdam-s What's the status of this ?

rousseldenis:
@damdam-s Could we close this as the depending PR is closed too ?

PR #1899, opened 2022-02-03
[FIX] sale_order_type: When creating a sale order, the "Type of order" changes.

alfredoavanzosc:
When creating the sale order, the selected sale order type changes.
1.- When selecting a customer in the sale order, it brings me the type of sale order defined in the customer, to the sales order. 
2.- I change the sales order type, and save the new sales order.
3.- The error is that AFTER EXECUTING the CREATE function of the sales order, the "_compute_sale_type_id" function is EXECUTED, in the pyhton file "sale_order.py", of the "sale_order_type" module, and the sales order is given the sales order type defined in the customer.

@anajuaristi was talking about this bug here https://github.com/OCA/sale-workflow/issues/1852

PR #1993, opened 2022-04-22
[FIX] sale_order_type: When creating a sale order, the "Type of order" changes

matiasperalta1:
Backport to 13.0 of https://github.com/OCA/sale-workflow/pull/1899

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -50,24 +50,23 @@ class SaleOrder(models.Model):
 
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
-        for record in self:
-            if not record.partner_id:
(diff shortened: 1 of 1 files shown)


sale_order_type/models/sale.py
```diff
@@ -50,24 +50,23 @@ def _default_sequence_id(self):
...
-                    ).sale_type
-                    or record.partner_id.commercial_partner_id.with_context(
-                        force_company=record.company_id.id
-                    ).sale_type
-                )
-                if sale_type:
-                    record.type_id = sale_type
-                else:  # HACK: Avoid CacheMiss when no sale_type is set
-                    record.type_id = record.type_id
+        if "compute_sale_type_id" not in self.env.context:
```
rousseldenis, line 53: @matiasperalta1 I don't like using context keys with stored computed fields.

PR #1995, opened 2022-04-25
[14.0] [FIX] sale_automatic_workflow_payment_mode: avoid overwritting/removing SO's workflow if payment.mode hasn't one set

ivantodorovich:
It seems this feature was added during the 14.0 migration PR, here: 111bb97c7
But it wasn't like this before, and IMO it was ok.

ping @hailangvn


hailangvn (approved):
Thanks @ivantodorovich for the fix. I realized the different now.

jcoux (approved):

PR #1948, opened 2022-03-15
[15.0][MIG] sale_sourced_by_line

JoanMForgeFlow:
Standard migration 15.0
Red because: https://github.com/OCA/sale-workflow/pull/1937

PR #2006, opened 2022-05-09
[FIX][15.0] Fixed tests sale_invoice_plan,

JoanMForgeFlow:
Fixed the module adding post_install, for CoA.

https://github.com/OCA/sale-workflow/pull/1948


mariadforgeflow (approved):
LGTM! But only last commit is necessary:)

PR #2014, opened 2022-05-12
12.0 add sale_technical_pricelist and account_invoice_technical_pricelist

legalsylvain:
propose a new module ``sale_technical_pricelist``. 
- it adds a simple boolean field ``is_technical`` on pricelist. 
- If checked, the pricelist is available to compute another pricelist, but is not available in the rest of the odoo Interface, (customer form, sale order form).

Note : 
- I'm not sure about the name. I was thinking also to ``sale_hidden_pricelist`` and for the boolean ``is_hidden``. What do you think ? (@pedrobaeza ;-) ?)
- Add also a trivial glue module ``account_invoice_technical_pricelist`` that make the feature available on customer invoice form, if OCA ``account_invoice_pricelist`` is installed. 
- another glue module should be done, for point of sale. See discussion https://github.com/OCA/pos/pull/788. CC @
francesco-ooops, @isserver1, @OCA/pos-maintainers 

GRAP Task#477 : CC : @quentinDupont, @mounasb


pedrobaeza:
Module naming convention I use is:

```
<main_area>_<subarea>_<feature>
```

With this in mind, it should be `product_pricelist_technical` or `sale_pricelist_technical`, depending on the scope, as I don't have clear the implementation. One possibility is to have the first name, and don't depend on any other module. You inherit `search` method, and add the domain `is_technical = False` by default, unless a specific context is passed. Then, on the pricelist view, you add such context for the manipulation you want to do. Something similar to this:

https://github.com/OCA/delivery-carrier/blob/12.0/delivery_multi_destination/models/delivery_carrier.py#L27

With this, no extra glue modules should be needed.

legalsylvain:
> Module naming convention I use is:

> <main_area>_<subarea>_<feature>

> With this in mind, it should be product_pricelist_technical or sale_pricelist_technical, depending on the scope, as I don't have clear the implementation

Thanks a lot, I'll update the name of the module. ``sale_pricelist_technical`` sounds a better OCA module name.

 > One possibility is to have the first name, and don't depend on any other module. You inherit search method, and add the domain is_technical = False by default, unless a specific context is passed. Then, on the pricelist view, you add such context for the manipulation you want to do. Something similar to this:

Yes, I considered that option but : 
- It could generate undesired side effect. (if search function is called legitimately).
- anyway, it will force to develop glue modules if we want to display all the pricelists in a module that implement another ``pricelist_id`` field)
- in a global way only 4 models requires to have pricelist hidden : ``res.partner``, ``sale.order`` (same module) ``account.invoice`` and ``pos.order`` (so 1 main module and 2 glue modules)
- ``pos_pricelist_technical`` should exist anyway, because PoS should load all the pricelists but hide "technical" ones. (otherwise, the pricelist that depends on the technical ones will not be computed correctly and generate javascript bugs.

ivantodorovich:
I'm wondering if we can avoid all these glue modules by playing with an `active` field (or even, make your new field act behave like an `active` field by overriding search)

Basically what you want to achieve is, I think, technical pricelists shouldn't be searched on unless there's a specific context key, which of course you'd add to the pricelists `act_window`, since that's the only place we really want to see them, right?

rousseldenis:
@legalsylvain IMHO, this has a lack of functional requirements.

'technical' could mean a lot of things... Do you have a more accurate description ?

legalsylvain:
> 'technical' could mean a lot of things... Do you have a more accurate description ?

Well : 3 pricelists : 
- a pricelist A
- a pricelist B with some items based on A
- a pricelist C with come items based on A

Usage :
- set B / C pricelists on customers / orders / invoices makes sense.
- set A pricelist doesn't make sense. 

I want to prevent sellers to use the A pricelist. 
So for me it's a "technical" pricelist, but another name is OK for me.

rousseldenis:
> > 'technical' could mean a lot of things... Do you have a more accurate description ?
> 
> Well : 3 pricelists :
> 
> * a pricelist A
> * a pricelist B with some items based on A
> * a pricelist C with come items based on A
> 
> Usage :
> 
> * set B / C pricelists on customers / orders / invoices makes sense.
> * set A pricelist doesn't make sense.
> 
> I want to prevent sellers to use the A pricelist. So for me it's a "technical" pricelist, but another name is OK for me.

Ok, I see. So, this is more a 'main' pricelist.

legalsylvain:
> Ok, I see. So, this is more a 'main' pricelist.

Do you mean that pricelist A is the "main" pricelist ? or the pricelist B & C. (in my exemple).

quentinDupont:
Scope seems quite clear for me and well explained in DESCRIPTION, so I'm wondering if it is an issue of translation ?
In french, `technical` is quite clear, but maybe not in english ?

It's not a `main` pricelist. The `technical` pricelist will not be used by users, so it can't be a `main` pricelist.
It's only used for technical purpose by admin / manager.

account_invoice_pricelist_technical/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2022 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Technical Pricelists For account Invoice",
+    "summary": "Prevent to select some pricelists on invoices form",
```
carmenbianca:
```suggestion
    "summary": "Prevent technical pricelists from being selected on invoice forms",
```

account_invoice_pricelist_technical/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2022 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Technical Pricelists For account Invoice",
```
carmenbianca:
```suggestion
    "name": "Technical Pricelists For Account Invoice",
```

sale_pricelist_technical/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2015 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Technical Pricelists for Sales",
+    "summary": "Prevent to select some pricelists on orders and customer form",
```
carmenbianca:
```suggestion
    "summary": "Prevent some pricelists from being selected on order and customer forms",
```

sale_pricelist_technical/models/product_pricelist.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    is_technical = fields.Boolean(
+        string="Is Technical",
+        help="Check this box if you want to prevent user to use"
+        " this pricelist on the orders, or in the customer form."
+        " This is usefull for pricelist created only to compute"
+        " other pricelists.",
```
carmenbianca:
```suggestion
        help="Check this box if you want to prevent users from using"
        " this pricelist on orders or in customer forms."
        " This is useful for pricelists that are created only to compute"
        " other pricelists.",
```

sale_pricelist_technical/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module extends the functionality of Pricelist Module to let
+the possibility to have "technical pricelists" that are available
+only to compute another pricelists.
+Those technical pricelists are hidden in the rest of the odoo Interface
+avoiding bad selection by user on sale order form, customer form, etc...
```
carmenbianca:
```suggestion
This module extends the functionality of the Pricelist module to make it
possible to create "technical pricelists" that can only be used
to compute another pricelists.

These technical pricelists are hidden in the rest of the Odoo interface,
preventing users from using them on sale order forms, customer forms, etc..
```

account_invoice_pricelist_technical/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module extends the functionality of the Technical pricelist Module (same repository) to prevent user to select undesired pricelist on account invoice form.
```
carmenbianca:
```suggestion
This module extends the functionality of the Technical Pricelist module (same repository) to prevent users from selecting technical pricelists on account invoice forms.
```

sale_pricelist_technical/i18n/fr.po
```diff
@@ -0,0 +1,61 @@
...
+#: model:product.pricelist,name:sale_pricelist_technical.technical_pricelist
+msgid "Technical Pricelist"
+msgstr "Liste de prix technique"
+
+#. module: sale_pricelist_technical
+#: model:ir.model.fields,help:sale_pricelist_technical.field_res_partner__property_product_pricelist
+#: model:ir.model.fields,help:sale_pricelist_technical.field_res_users__property_product_pricelist
+msgid "This pricelist will be used, instead of the default one, for sales to the current partner"
+msgstr "Cette liste de prix sera utilisée, à la place de la liste par défaut, pour les ventes faites au partenaire actuel."
+
```
carmenbianca: I couldn't find the source text in the code.

carmenbianca (approved):
LGTM, just some improvements to strings.

I think 'technical' as a descriptor is _fine_, but not great. I don't know any better suggestions, though. 'Parent pricelist'? Seems inaccurate.

sale_pricelist_technical/i18n/fr.po
legalsylvain: Hi. It's a text that is autogenerated by Odoo when generating new translation.
I don't know why exactly. The module overload the field `property_product_pricelist` but doesn't define help value...
I removed it, but maybe weblate will add it again !

quentinDupont (approved):
Using in production since 6 months !

legalsylvain:
Hi @OCA/crm-sales-marketing-maintainers could you merged this accepted PR ? thanks !

PR #2016, opened 2022-05-13
[14.0][FIX] product_form_sale_link: assure compatibility with sale_stock

MiquelRForgeFlow:
The button 'action_view_sales' is replaced in sale_stock, so it cannot be referenced.


LoisRForgeFlow (approved):
:+1:

rousseldenis (approved):

PR #2017, opened 2022-05-13
[14.0][IMP] sale_blanket_order: Make open Blanket Sale Orders more editable

BernatPForgeFlow:


sale_blanket_order/models/blanket_orders.py
```diff
@@ -89,8 +89,6 @@ def _compute_amount_all(self):
     payment_term_id = fields.Many2one(
         "account.payment.term",
         string="Payment Terms",
-        readonly=True,
-        states={"draft": [("readonly", False)]},
```
rousseldenis, line 93: Should payment terms change after confirmation ? Strange

rousseldenis:
@Cedric-Pigeon IMHO those changes give too much edit rights after confirmation.

sale_blanket_order/models/blanket_orders.py
BernatPForgeFlow: @rousseldenis  Edit Payment Terms while a Blanket Order is open should not be a problem because the sale orders you create will have this Payment Term. 

You can agree with your client to change the Payment Terms and once you change it in the Blanket Order will affect the next Sale Orders.

BernatPForgeFlow:
> @Cedric-Pigeon IMHO those changes give too much edit rights after confirmation.

What do you think now? 

I added a checkbox in the Settings to choose if the user wants open Blanket Orders editable.

ChrisOForgeFlow (approved):
Functional and code review

PR #2020, opened 2022-05-16
[IMP] sale_order_mass_action: Add missing target states

rousseldenis:


sale_order_mass_action/wizards/sale_order_mass_action_wizard.py
```diff
@@ -4,22 +4,50 @@
...
+        domain = expression.AND([domain, self._get_sale_order_state_domain()])
+        return domain
+
+    def _launch_action(self, sale_orders):
+        self.ensure_one()
+        action = "action_" + self.action
+        if hasattr(sale_orders, action):
+            getattr(sale_orders, action)()
+            return True
+        return False
```
legalsylvain, line 50: minor question (not blocking) : if the action is not found in the hasattr, I expect the function raise an error. don't you think ?

legalsylvain (approved):
LGTM. Thanks !

sale_order_mass_action/wizards/sale_order_mass_action_wizard.py
rousseldenis: Maybe. In this case, the user won't have the result notification with the updated sale orders.

But we could improve it and display an error notify in that case

rousseldenis: @legalsylvain What do you think ?

Garibaldy:
why this pull request is not merge yet?
 i'm considering to migrate it to v15  ASAP

sale_order_mass_action/wizards/sale_order_mass_action_wizard.py
legalsylvain: But in which case the action doesn't exist ? Sorry but i dont understand that test.

PR #2027, opened 2022-05-17
[15.0][MIG] sale_invoice_policy

yankinmax:


ivantodorovich (approved):
LG

yankinmax:
@rousseldenis can you pls take a look and probably approve the workflow?

ajaniszewska-dev (approved):

sale_invoice_policy/models/product_template.py
```diff
@@ -7,17 +7,16 @@
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
-    def _default_default_invoice_policy(self):
```
rousseldenis: Mmmh maybe this had been done at purpose to easy override

yankinmax: yep, I got the idea of what it was used for.
But the reason I made it another way is the fact that at v15 we don't have this default.
I checked it through debugging. We have `None` value returned in this method.
At the module installation in test db I've always met:
```
  File "/odoo/src/odoo/service/server.py", line 1246, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/odoo/src/odoo/modules/registry.py", line 87, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/odoo/src/odoo/modules/loading.py", line 474, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/odoo/src/odoo/modules/loading.py", line 363, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/odoo/src/odoo/modules/loading.py", line 199, in load_module_graph
    registry.init_models(cr, model_names, {'module': package.name}, new_install)
  File "/odoo/src/odoo/modules/registry.py", line 432, in init_models
    env['ir.model.fields.selection']._reflect_selections(model_names)
  File "/odoo/src/odoo/addons/base/models/ir_model.py", line 1216, in _reflect_selections
    expected = {
  File "/odoo/src/odoo/addons/base/models/ir_model.py", line 1219, in <dictcomp>
    for field_id in [IMF._get_ids(field.model_name)[field.name]]
KeyError: 'default_invoice_policy'
```
So, not to break logic, I propose to use another default which should return the same value => `order`.
If you have suggestion how to resolve this error another way I would highly appreciate ideas.
Thanks for review @rousseldenis

PR #1569, opened 2021-04-26
[14.0][MIG] sale_force_invoiced

MateuGForgeFlow:
Standard migration to v14.
cc @JordiBForgeFlow

PR #2031, opened 2022-05-23
[13.0][FIX] sale_force_invoiced: Editable also for confirmed SOs

pedrobaeza:
The check should be editable as well on non-locked SOs, because the invoicing status is independent from the lock status, and moreover, you may not have enabled the "Lock confirmed sales" option in settings, and thus, the locked status won't be never reachable.

@Tecnativa TT36813

### DIFF
--- a/sale_force_invoiced/README.rst
+++ b/sale_force_invoiced/README.rst
@@ -51,8 +51,8 @@ Usage
 #. Deliver the products/services.
 #. Create an invoice and reduce the invoiced quantity. The sales order
    invoicing status is 'To Invoice'.
-#. Lock the Sale Order, to change the status of it to 'Done'.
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow (approved):
You could have backported https://github.com/OCA/sale-workflow/pull/1569/commits/2994ece7e22c007fdbf85748bca6b15b6005b3ae :)

pedrobaeza:
I didn't know it exists such one, but the README requires also an update. I will cherry-pick directly the README part when merged.

pedrobaeza:
And please, put proper commit messages on why you do such changes like I do.

pedrobaeza:
As this is already merged in upper versions, let's merge it:

/ocabot merge minor

PR #2034, opened 2022-05-27
[14.0][FIX] sale_partner_incoterm: Makes the incoterm field visible

antoniospneto:
This PR makes the default incoterm field for sales orders visible, even without having registered sales orders.
Without this PR it is only possible to inform the incoterm after having a sales order issued.


felipemotter (approved):

flotho (approved):
thanks,
easy fix 
LGTM

PR #2035, opened 2022-05-30
[ADD] MODULE: sale_team_payment_term

Kev-Roche:
This module allows to set payment terms by sales team, if no one exists on partner.

### DIFF
--- a/sale_team_payment_term/models/sale_order.py
+++ b/sale_team_payment_term/models/sale_order.py
@@ -1,0 +1,39 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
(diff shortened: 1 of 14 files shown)


sale_team_payment_term/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+    _inherit = "sale.order"
+
+    payment_term_id = fields.Many2one(
+        compute="_compute_payment_term_id",
+        inverse="_inverse_payment_term_id",
+        readonly=False,
+        store=True,
+    )
+
+    @api.depends("partner_id.property_payment_term_id", "team_id.team_payment_term_id")
```
rousseldenis, line 18: @Kev-Roche This is not correct. As if you change the payment term on partner, every sale orders will be recomputed...

You should depends on partner_id only. Moreover, I think the onchange function could be removed too.

Could you check the behavior in that case ?

PR #2036, opened 2022-05-31
[13.0][FIX] sale_order_price_recalculation: Buttons visibility

pedrobaeza:
With current inheritance, the buttons are only visible if you have the security group "Lock Confirmed Sales" in your user.

As we need to be seen in all cases, we modify the inheritance.

TT37124

### DIFF
--- a/sale_order_price_recalculation/views/sale_order_view.xml
+++ b/sale_order_price_recalculation/views/sale_order_view.xml
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <record id="view_sales_order_auto_done_setting" model="ir.ui.view">
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form - Add recompute buttons</field>
         <field name="model">sale.order</field>
-        <field name="inherit_id" ref="sale.view_sales_order_auto_done_setting" />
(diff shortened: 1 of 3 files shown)


legalsylvain (approved):
thanks.

sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -1,10 +1,11 @@
...
 <odoo>
-    <record id="view_sales_order_auto_done_setting" model="ir.ui.view">
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form - Add recompute buttons</field>
         <field name="model">sale.order</field>
-        <field name="inherit_id" ref="sale.view_sales_order_auto_done_setting" />
+        <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
-            <button name="action_unlock" position="after">
+            <field name="state" position="before">
```
legalsylvain, line 8: nitpicking (not blocking) : positioning with other button has more sense IMO. something like : 
```<button name="action_draft" position="after">```

pedrobaeza: I want to always be the last buttons, so we can't assure that positioning according other buttons.

victoralmau (approved):

PR #2043, opened 2022-06-02
[15.0][MIG] sale_stock_cancel_restriction: migrate sale_stock_cancel_restriction 

cesar-tecnativa:
@Tecnativa TT36542


pedrobaeza:
@cesar-tecnativa Please avoid the merge commit.

cesar-tecnativa:
> @cesar-tecnativa Please avoid the merge commit.
Done

pedrobaeza (approved):
You may complete the test coverage adding a test for a successful order cancellation (sometimes there are surprises in this way).

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -0,0 +1,30 @@
...
+        cls.sale_order = so_form.save()
+        cls.sale_order.action_confirm()
+        cls.picking = cls.sale_order.picking_ids
+        cls.picking.move_lines.quantity_done = 2
+        cls.picking.button_validate()
+
+    def test_cancel_sale_order(self):
+        with self.assertRaises(UserError):
+            self.sale_order.action_cancel()
+            self.assertEqual(self.sale_order.state, "cancel")
```
victoralmau: IMO, this can confuse that the behavior is wrong, I think it would be better to remove it.

Code review and test in runbot OK.

pedrobaeza: Yeah, this line will never be reached

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -25,5 +24,45 @@ def setUpClass(cls):
...
 
     def test_cancel_sale_order(self):
-        with self.assertRaises(UserError):
-            self.sale_order.action_cancel()
+        """
+        Test a SO with a product on delivery with a 5 quantity.
+        Create two invoices: one for 3 quantity and one for 2 quantity
+        Then cancel Sale order, it won't raise any warning, it should be cancelled.
+        """
+        partner = self.partner
```
pedrobaeza: Why are you creating another sales order if the setUp already contains one and its picking? Previous test code was correct.

pedrobaeza (changes requested):

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -22,8 +21,19 @@ def setUpClass(cls):
         cls.sale_order.action_confirm()
         cls.picking = cls.sale_order.picking_ids
         cls.picking.move_lines.quantity_done = 2
-        cls.picking.button_validate()
```
rousseldenis: Isn't this changing the test behaviour ?

The picking should be done before entering tests.

cesar-tecnativa: Previously there was only one setup function in the [test file](https://github.com/OCA/sale-workflow/blob/1c8b405dc7d135c13fa00999ff777937b7390d34/sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py#L8).

I added two new tests: one to validate and one to cancel. 

It does not change the test behavior.

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -22,8 +21,19 @@ def setUpClass(cls):
...
 
     def test_cancel_sale_order(self):
-        with self.assertRaises(UserError):
-            self.sale_order.action_cancel()
+        """
+        Cancel Sale order, it won't raise any warning, it should be cancelled.
+        """
+        # check the status of invoices after cancelling the order
+        self.sale_order.action_cancel()
+        for picking in self.sale_order.picking_ids:
```
pedrobaeza: No need to make a for. There's only one.

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -22,8 +21,19 @@ def setUpClass(cls):
         cls.sale_order.action_confirm()
         cls.picking = cls.sale_order.picking_ids
         cls.picking.move_lines.quantity_done = 2
-        cls.picking.button_validate()
+
+    def test_validate_sale_order(self):
+        self.picking.button_validate()
 
     def test_cancel_sale_order(self):
-        with self.assertRaises(UserError):
```
pedrobaeza: Don't remove this test. The tests should be:

- `test_cancel_sale_order_restrict`: validates the picking and do the `assertRaises` cancelling the order for checking that it's forbidden (`self.picking.button_validate()` + these 2 lines)
- `test_cancel_sale_order_ok`: don't validate the picking and cancel the order, being completed (the next test you have written).

pedrobaeza (changes requested):

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
cesar-tecnativa: ok done

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -0,0 +1,46 @@
...
+        cls.sale_order.action_confirm()
+        cls.picking = cls.sale_order.picking_ids
+        cls.picking.move_lines.quantity_done = 2
+
+    def test_cancel_sale_order_restrict(self):
+        """
+        Validates the picking and do the assertRaises cancelling the
+        order for checking that it's forbidden
+        :return:
+        """
```
pedrobaeza: Proper docstring according PEP257 is:

```suggestion
        """Validates the picking and do the assertRaises cancelling the
        order for checking that it's forbidden
        """
```

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -0,0 +1,46 @@
...
+        self.picking.button_validate()
+        with self.assertRaises(UserError):
+            self.sale_order.action_cancel()
+
+    def test_cancel_sale_order_ok(self):
+        """
+        Don't validate the picking and cancel the order, being completed.
+        """
```
pedrobaeza: If it fits in one line, the docstring is:

```suggestion
        """Don't validate the picking and cancel the order, being completed."""
```

but if not, the first line goes with the opening quotes.

sale_stock_cancel_restriction/tests/test_sale_stock_cancel_restriction.py
```diff
@@ -0,0 +1,46 @@
...
+        # check the status of invoices after cancelling the order
+        self.sale_order.action_cancel()
+        self.assertEqual(
+            self.sale_order.picking_ids.state,
+            "cancel",
+            "After cancelling a picking the state should be cancelled",
```
pedrobaeza:
```suggestion
            "After cancelling a picking, the state should be cancelled",
```

pedrobaeza (approved):

cesar-tecnativa:
@victoralmau Could you please review it?

victoralmau (approved):
Code review OK.

Could you squash administrative commits?

PR #2055, opened 2022-06-07
[15.0] [ADD] sale_automatic_workflow_advance

Nikul-OSI:


sale_automatic_workflow_advance/models/account_move.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import json
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def action_post(self):
```
dreispt: You can remove this, as the `sale_advance_payment` module will already implement it.

dreispt (changes requested):

dreispt:
@Nikul-OSI This looks good, we might need to add more flexibility on the moment the Payment is recorded, but let me test drive this and come back here.

sale_automatic_workflow_advance/models/sale_workflow_job.py
```diff
@@ -0,0 +1,38 @@
...
+                )
+                .id,
+            )
+            self.env["account.voucher.wizard"].with_context(**ctx).create(
+                {
+                    "journal_id": bank_journal,
+                    "amount_advance": sale.amount_residual,
+                }
+            ).make_advance_payment()
+        return res
```
dreispt: @Nikul-OSI We want to be able to trigger this for non-confirmed Quotations.
Make this a separate method, and have it called from `run_with_workflow()`, before calling `super()`.
Thanks

Nikul-OSI: @dreispt Code updated base on comments, Can you please review it?

sale_automatic_workflow_advance/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+* Sale Automatic Workflow Advance
```
rousseldenis: Could you improve a little bit this and add a USAGE. Thanks

rousseldenis (changes requested):
Code review. Just a little enhancement of README but looks great !

sale_automatic_workflow_advance/readme/DESCRIPTION.rst
Nikul-OSI: @rousseldenis code updated base your comments.

sale_automatic_workflow_advance/models/sale_workflow_job.py
```diff
@@ -0,0 +1,62 @@
...
+        ):
+            return "{} {} job bypassed".format(sale.display_name, sale)
+        else:
+            ctx = {
+                "active_ids": sale.ids,
+                "default_order_id": sale.id,
+                "default_amount_total": sale.amount_residual,
+                "default_currency_id": sale.pricelist_id.currency_id.id,
+            }
+            bank_journal = (
```
Cedric-Pigeon: IMO this code is not optimal. You cannot take the first one arbitrarily.
The best method would be to use https://github.com/OCA/bank-payment/tree/15.0/account_payment_mode features to get a proper journl

Cedric-Pigeon (changes requested):

sale_automatic_workflow_advance/models/sale_workflow_job.py
dreispt: @Cedric-Pigeon 
Odoo standard uses this approach in several occasions, so it is line with that.
We don't want to introduce the account_payment_mode dependency.
What I suggest is to have this in a `_prepare_...` method, that can then be extended to support that or other logic.

rousseldenis: @dreispt Yes, at least that should be welcome (even if I agree with @Cedric-Pigeon and taking Odoo example is not necessary a good thing - if there is no good reason :smile: )

dreispt:
Hello, review comments have been addressed.
Can we finalize reviews? Thanks!

dreispt (approved):

dreispt:
Reviewers, con this be merged?

CasVissers-360ERP (approved):
Functional review, great add-on!

dreispt:
I see improvement suggestions, but not blocking comments.
Moving this forward.

/ocabot merge nobump

Fatimahairup:
@dreispt , is there anything I can do to help with this. The logs on this seems to be outdated for the test

dreispt:
Rebased and commits squashed.

rousseldenis:
@Nikul-OSI @dreispt You should fix the tests

PR #2065, opened 2022-06-10
[13.0] [IMP] sale_order_product_recommendation: Add unit of measure at wizard

willytal:
cc @Tecnativa TT37389

Please review this @sergio-teruel @CarlosRoca13

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -210,7 +210,9 @@ class SaleOrderRecommendationLine(models.TransientModel):
         readonly=True,
     )
     sale_line_id = fields.Many2one(comodel_name="sale.order.line")
-    sale_uom_id = fields.Many2one(related="sale_line_id.product_uom")
+    sale_uom_id = fields.Many2one(
+        compute="_compute_sale_uom_id", comodel_name="uom.uom", store=True
+    )
 
     @api.depends(
         "partner_id",
@@ -279,3 +281,11 @@ class SaleOrderRecommendationLine(models.TransientModel):
             .with_context(prefetch_fields=False)
         )
         return so_line.price_unit or 0.0
+
+    @api.depends("sale_line_id", "product_id")
+    def _compute_sale_uom_id(self):
+        for record in self:
+            if record.sale_line_id:
+                record.sale_uom_id = record.sale_line_id.product_uom
+            else:
+                record.sale_uom_id = record.product_id.uom_id
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
@@ -26,6 +26,7 @@
                                 <field name="times_delivered" />
                                 <field name="units_delivered" />
                                 <field name="units_included" widget="numeric_step" />
+                                <field name="sale_uom_id" optional="show" />
                             </tree>
                             <kanban
                                 class="o_kanban_mobile"


CarlosRoca13 (approved):

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -279,3 +281,11 @@ def _get_last_sale_price_product(self):
             .with_context(prefetch_fields=False)
         )
         return so_line.price_unit or 0.0
+
+    @api.depends("sale_line_id", "product_id")
```
cesar-tecnativa, line 285: @api.depends("sale_line_id", "product_id", "sale_line_id.product_uom", "product_id.uom_id", )

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -279,3 +281,11 @@ def _get_last_sale_price_product(self):
...
+    def _compute_sale_uom_id(self):
+        for record in self:
+            if record.sale_line_id:
+                record.sale_uom_id = record.sale_line_id.product_uom
```
cesar-tecnativa, line 289: record.sale_uom_id = record.sale_line_id and record.sale_line_id.product_uom and record.sale_line_id.product_uom.id or False

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -279,3 +281,11 @@ def _get_last_sale_price_product(self):
...
+            else:
+                record.sale_uom_id = record.product_id.uom_id
```
cesar-tecnativa, line 291: record.sale_uom_id = record.product_id and record.product_id.uom_id and record.product_id.uom_id.id or False

sale_order_product_recommendation/wizards/sale_order_recommendation.py
CarlosRoca13: I think this change is not necessary as it is for a TranscientModel and the change of uom on the line or the product, will not be updated asynchronously. So, detecting just the change of the product or the sale_line_id will be enough

CarlosRoca13:
@cesar-tecnativa I don't understand why we need to make this assignations when we are doing the check with the if

rousseldenis:
> @cesar-tecnativa I don't understand why we need to make this assignations when we are doing the check with the if

That's more pythonic way of doing it.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
rousseldenis: IIMHO the good one is only : ("sale_line_id.product_uom", "product_id.uom_id")

sale_order_product_recommendation/wizards/sale_order_recommendation.py
sergio-teruel: I think that the better way is 

record.sale_uom_id = record.sale_uom_id or record.product_id.uom_id

cesar-tecnativa: you are right

sale_order_product_recommendation/wizards/sale_order_recommendation.py
cesar-tecnativa: you are right

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -210,7 +210,9 @@ class SaleOrderRecommendationLine(models.TransientModel):
         readonly=True,
     )
     sale_line_id = fields.Many2one(comodel_name="sale.order.line")
-    sale_uom_id = fields.Many2one(related="sale_line_id.product_uom")
+    sale_uom_id = fields.Many2one(
+        compute="_compute_sale_uom_id", comodel_name="uom.uom", store=True
```
pedrobaeza, line 214: This shouldn't be stored.

pedrobaeza (changes requested):
The problem is that the related fields initialized in the wizard doesn't get populated on first launch, but when saving it. But the solution is easier: just assign the value you want in such initialization.

Here:

https://github.com/OCA/sale-workflow/blob/afc6927100cee72aea0ad40d3175bf32c1fa9681/sale_order_product_recommendation/wizards/sale_order_recommendation.py#L103

just add:

```python
            vals["sale_uom_id"] = so_line.product_uom.id
```

and also cover the case when no sale line using `group_line["product_id"][0]` for getting the UoM of the product.

You can optionally remove the related and you will have both things covered.

pedrobaeza:
@sergio-teruel did you see my comment?

rousseldenis:
@cesar-tecnativa What's the status of this ?

rousseldenis:
@sergio-teruel What's the status of this ?

PR #2067, opened 2022-06-13
[14.0][MIG] sale_order_product_assortment: Migration to 14.0

ilyasProgrammer:


francesco-ooops (approved):
functional review ok, @ilyasProgrammer you can mark as ready for review

---

one great thing about product assortments is that they can be applied to multiple models through their various extension modules (sales, purchases, pickings, website sale, etc), but the downside is that user has no control on which models they should be applied to, eg: 

assortment A restricts customer 1 from buying certain products on ecommerce, but the same restriction is not applied to SO and picking

we would like each module extending product_assortment to feature a flag "apply to" that can be activated for each model, and we could start with this one since the others have not been migrated to 14 yet (apart from base). Should we add the feature to base?

pinging @CarlosRoca13 @rousseldenis @pedrobaeza for opinion as main contributors/maintainers

CarlosRoca13 (approved):
Code review :+1:

ilyasProgrammer:
@francesco-ooops it makes sence as for me. But perhaps need to implement it in different repo.

pedrobaeza:
It might be a mixin that is applied by default to field `product_id` and only needed a small layer on top if the field is called different in the applied model. That can be an interesting refactoring, but that doesn't prevent to add little glue modules. And try to make this configurable to other level will provoke to overhead whole system reading if the model is configured to be "assorted".

PR #2070, opened 2022-06-14
[15.0][FIX] sale_product_multi_add: Fixed the issue with Pricelist discount policy configuration

st-anagpal:
[15.0][FIX] sale_product_multi_add: The module now works with the discount policy set on the Pricelists as Show public price & discount to the customer.

Earlier when the 'Discount Policy' on Pricelist when set to 'Show public price & discount to customer',
the addition of multiple products to the sale order line didn't consider pricelist at all.
Now this feature works with the above mentioned pricelist configuration as well.

### DIFF
--- a/sale_product_multi_add/__manifest__.py
+++ b/sale_product_multi_add/__manifest__.py
@@ -9,7 +9,7 @@
     "category": "Sale Management",
     "version": "15.0.1.0.0",
     "license": "AGPL-3",
-    "depends": ["sale_stock"],
+    "depends": ["sale"],
     "data": [
         "security/ir.model.access.csv",
         "wizards/sale_import_products_view.xml",
--- a/sale_product_multi_add/__manifest__.py
+++ b/sale_product_multi_add/__manifest__.py
@@ -9,7 +9,7 @@
     "category": "Sale Management",
     "version": "15.0.1.0.0",
     "license": "AGPL-3",
-    "depends": ["sale"],
+    "depends": ["sale_stock"],
     "data": [
         "security/ir.model.access.csv",
         "wizards/sale_import_products_view.xml",
--- a/sale_product_multi_add/wizards/sale_import_products.py
+++ b/sale_product_multi_add/wizards/sale_import_products.py
@@ -57,8 +57,8 @@ class SaleImportProducts(models.TransientModel):
                 for item in wizard.items:
                     vals = self._get_line_values(sale, item)
                     if vals:
-                        self.env["sale.order.line"].create(vals)
-
+                        line = self.env["sale.order.line"].create(vals)
+                        line._onchange_discount()
         return {"type": "ir.actions.act_window_close"}
 
 
--- a/sale_product_multi_add/tests/test_sale.py
+++ b/sale_product_multi_add/tests/test_sale.py
@@ -17,7 +17,10 @@ class TestSale(common.TransactionCase):
         """
 
         so = self.env["sale.order"].create(
-            {"partner_id": self.env.ref("base.res_partner_2").id}
+            {
+                "partner_id": self.env.ref("base.res_partner_2").id,
+                "picking_policy": "direct",
+            }
         )
 
         wiz_obj = self.env["sale.import.products"]
--- a/sale_product_multi_add/tests/test_sale.py
+++ b/sale_product_multi_add/tests/test_sale.py
@@ -17,10 +17,7 @@ class TestSale(common.TransactionCase):
         """
 
         so = self.env["sale.order"].create(
-            {
-                "partner_id": self.env.ref("base.res_partner_2").id,
-                "picking_policy": "direct",
-            }
+            {"partner_id": self.env.ref("base.res_partner_2").id}
         )
 
         wiz_obj = self.env["sale.import.products"]


sale_product_multi_add/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "category": "Sale Management",
     "version": "15.0.1.0.0",
     "license": "AGPL-3",
-    "depends": ["sale"],
+    "depends": ["sale_stock"],
```
rousseldenis: @st-anagpal Could you explain why you need this dependency?

st-anagpal: @rousseldenis  Yes, a field 'picking_policy' is used in test cases which is defined in the module sale_stock. The test case was failing for this module earlier and in order to fix it I added picking_policy and this fixed the issue.

rousseldenis: > @rousseldenis Yes, a field 'picking_policy' is used in test cases which is defined in the module sale_stock. The test case was failing for this module earlier and in order to fix it I added picking_policy and this fixed the issue.

Test failing is not a good reason to introduce a new dependency. You should be able to install sale_product_mutli_add module without stock one!

st-anagpal: @rousseldenis Ok noted. I'll remove this dependency but in that case the test cases would fail. What is the best way out here?

sale_product_multi_add/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "category": "Sale Management",
     "version": "15.0.1.0.0",
```
n17: Should we also bump the version number to `15.0.1.0.1`?

st-anagpal: @n17  I think yes we should update the version number to `15.0.1.0.1`

sale_product_multi_add/__manifest__.py
st-anagpal: @n17 What are your thoughts here? Any suggestions?

rousseldenis: > @rousseldenis Ok noted. I'll remove this dependency but in that case the test cases would fail. What is the best way out here?

If you speak about your tests code (as here there is no error), that depends on how you build them. Maybe you pull a dependency there that requires the picking policy. So, you need to adapt your tests

sale_product_multi_add/__manifest__.py
rousseldenis: @st-anagpal @n17 There is no need of updating module version as it will be done through ocabot.

The only case where you need it is when you have migration scripts

sale_product_multi_add/__manifest__.py
st-anagpal: @rousseldenis Ok thank you. It is done.

sale_product_multi_add/__manifest__.py
st-anagpal: @rousseldenis Ok noted.

st-anagpal:
@rousseldenis The necessary corrections have been made. You can review it.

sale_product_multi_add/wizards/sale_import_products.py
```diff
@@ -57,8 +57,8 @@ def select_products(self):
                 for item in wizard.items:
                     vals = self._get_line_values(sale, item)
                     if vals:
-                        self.env["sale.order.line"].create(vals)
-
+                        line = self.env["sale.order.line"].create(vals)
+                        line._onchange_discount()
```
rousseldenis, line 61: @st-anagpal You should call it in _get_line_values() after product_id_change().

IMHO, this process should be improved in order to be compliant with further modules on top of this (and that are not aware of).

A solution could be onchange_helper https://github.com/OCA/server-tools/tree/14.0/onchange_helper

TumbaoJu:
@rousseldenis Hello!  Could it be fixed also for V14?
We found the same issue using the module in V14.

Thanks in advance.

@st-anagpal : For your information

TumbaoJu:
@Cedric-Pigeon : Hello, could you or any other contributor to the module sale_product_multi_add see my note above and tell me if it is possible to fix it in V14?
Thank you in advance

TumbaoJu:
@simahawk
Hello!  Can you help with our question? What is the best way to have a follow-up on this?
Thank you in advance ;)

sale_product_multi_add/wizards/sale_import_products.py
simahawk: In theory we should never call onchanges explicitly. A better approach is to use a tool like `onchange_helper` as Denis reported. 

However, as this is a patch and the method `_get_line_values` already calls an onchange, to me it's fair if add this call. @st-anagpal please move it to `_get_line_values`.

It would also be nice to add a specific check for this issue in the tests so that we reduce the chances to have the same issue in the future.

simahawk (changes requested):
@st-anagpal thanks for your contrib. Please check our reviews above.

Once you've done w/ those changes, please rewrite your commits as such:

1. squash the last commit as is not relevant
2. rewrite the 1st one like

```
 [FIX] sale_product_multi_add: fix SO line discount

Earlier when the 'Discount Policy' on Pricelist when set to 'Show public price & discount to customer', the addition of multiple products to the sale order line didn't consider pricelist at all. 
Now this feature works with the above mentioned pricelist configuration as well.
```

Note that the odoo version has been removed and the 1st line contains only a very short statement.

simahawk:
> @simahawk Hello! Can you help with our question? What is the best way to have a follow-up on this? Thank you in advance ;)

Thanks for the ping! :+1:

TumbaoJu:
@majouda  Maybe you could propose the fix for V14 based on this one and comments?

PR #2071, opened 2022-06-14
[14.0][IMP]sale_discount_display_amount: Show total discount on sale report

manuelregidor:
This PR adds the total discount value to the sale report, right before the sale total.


rousseldenis (approved):
Seems legit

rousseldenis:
@Cedric-Pigeon FYI

Cedric-Pigeon (approved):

HaraldPanten (changes requested):
I think something is missing/not correct:

Captura de Pantalla 2022-06-15 a las 11 19 19

HaraldPanten (approved):
I can reply myself. It's correct

PR #2077, opened 2022-06-16
13.0 mig sale order action invoice create hook

dantiller2:


rousseldenis:
@dantiller2 Thanks for this.

As invoicing method implementation has changed, is this still applicable ?

@rven

dantiller2:
I agree it may not be applicable. Looks like my own need for it in upgrading has gone away.

> As invoicing method implementation has changed, is this still applicable ?
> @rven

PR #2081, opened 2022-06-18
[14.0[FIX] CI after switching to GH actions

pedrobaeza:


rousseldenis:
You can promote sale_order_line_chained_move to Stable, used in 13 and 14 for months without problems

sale_stock_delivery_address/__manifest__.py
```diff
@@ -1,14 +1,14 @@
...
 
 {
     "name": "Sale Stock Delivery Address",
     "version": "14.0.1.0.1",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "category": "Sales Management",
-    "license": "LGPL-3",
-    "development_status": "Production/Stable",
+    "license": "AGPL-3",
```
rousseldenis, line 10: Before modifying this, I would have asked author

pedrobaeza: This is not something that the author has to decide: they are not complying with the licenses. They can later think about relicensing the whole chain. For now, I'm only fixing what is not correct.

rousseldenis: I didn't see licensing error in tests

pedrobaeza: https://github.com/OCA/sale-workflow/runs/6948099807?check_suite_focus=true#step:5:5

rousseldenis (approved):

sale_stock_delivery_address/__manifest__.py
pedrobaeza: @LoisRForgeFlow be aware of this license change for complying with dependency chain to proceed with the actions you prefer.

rousseldenis:
@pedrobaeza I've created this: https://github.com/OCA/sale-workflow/issues/2082

pedrobaeza:
OK, thanks.

PR #1957, opened 2022-03-23
[15.0]mig] sale_exception

damdam-s:
migration of `sale_exception` to v15

PR #2091, opened 2022-06-23
[ADD] sale_quotation_delivery_mandatory

xavier-bouquiaux:
a small module that force you to set a delivery method on your quotation to be able to confirm it

### DIFF
--- a/sale_quotation_delivery_mandatory/__init__.py
+++ b/sale_quotation_delivery_mandatory/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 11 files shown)


acsonefho (approved):

leemannd:
For this kind of needs I would use the module `sale_exception` => https://github.com/OCA/sale-workflow/pull/1957
I don't think we want to add one module for each kind of mandatory field we want to manage into quotations/sale orders

xavier-bouquiaux:
> For this kind of needs I would use the module `sale_exception` => #1957 I don't think we want to add one module for each kind of mandatory field we want to manage into quotations/sale orders

Thanks a lot for the advice
I close this PR

leemannd:
You're welcome

PR #2093, opened 2022-06-24
[13.0][ADD] sale_attached_product: New module

chienandalu:
cc @Tecnativa TT37588


sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,92 @@
...
+            attached_line_qtys = set(line.attached_line_ids.mapped("product_uom_qty"))
+            if any(q != line.product_uom_qty for q in attached_line_qtys):
+                line.attached_line_ids.update({"product_uom_qty": line.product_uom_qty})
+
+    def recompute_attached_products(self):
+        """Recurrent method for recomputing attached lines"""
+        for order in self.filtered(lambda x: x.state not in {"done", "cancel"}):
+            order._remove_attached_lines()
+            order._create_attached_lines()
+            order._update_attached_lines()
```
legalsylvain: I'm not sure of that design. 
If I understand correctly, on each write on a sale order (adding / changing a line for exemple), it will drop + create + update all the attached lines.
With the current design, If 
1 I had a main product that adds an attached product, 
2 I change any information on the attached line. (description / price / discount / etc...)
3 I update the sale order
-> the changed I did will be dropped don't you think ?

What about to have a lighter algorithm like 
```_check_attached_lines```.
-> if lines with main product doesn't have attached lines -> create it
-> if lines with attaches lines are updated check if value (product_id / product_uom_qty) are the same. If not update the attached lines.
-> regarding the deletion of main lines, maybe adding ondelete cascade on attached_from_line_id could do the work. Don't you think ?

We can though keep a button "``button_reset_attached_lines`` to recreate all the lines.

File: sale_attached_product/models/sale_order.py [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        orders = super().create(vals_list)
+        orders.recompute_attached_products()
+        return orders
+
+    def write(self, vals):
+        super().write(vals)
+        self.recompute_attached_products()
+
```
legalsylvain, line 158: I think you should return result of super() after.

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,92 @@
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_attached_line = fields.Boolean(
+        help="Falg products that are attached to their main counterpart"
+    )
+    attached_from_line_id = fields.Many2one(comodel_name="sale.order.line",)
+    attached_line_ids = fields.One2many(
+        comodel_name="sale.order.line", inverse_name="attached_from_line_id",
+    )
```
legalsylvain, line 170: what about if there is a direct write on a line. (via sale.order.line views or via code) 
The update of related lines will be raised ? 
(naive question : don't know the answer).

legalsylvain (commented):
Hi @chienandalu. 

Thanks a lot for sharing this module ! I have in my roadmap this feature. (in case of returnable product like beers with returnable bottles).

however I was never very clear on how to implement it.

some remarks / question inline.

General questions : 
- do you think it will work out of the box in a website_sale context ? 
- did you planned to work on a ``pos_attached_product`` module ?

kind regards.

sale_attached_product/models/sale_order.py
chienandalu: Hi, @legalsylvain thanks for your comments :)

In general, the design is following that of the sale coupon updates with the OCA auto refresh functionallity. So in that design:

1. We cleanup those lines that shouldn't be on the order.
2. We create those which for lines with line links.
3. And finally update those already existing that could have change.

This is more or less what you pointed. Now is improved to trigger the computation only when certain fields are changed.

> -> regarding the deletion of main lines, maybe adding ondelete cascade on attached_from_line_id could do the work. Don't you think ?

I tried, but it didn't seem to work. One2many self relationships tend to be tricky I'm afraid.

> -> regarding the deletion of main lines, maybe adding ondelete cascade on attached_from_line_id could do the work. Don't you think ?

Our goal is full automation. Although it could be easily extended to make it manual.

sale_attached_product/models/sale_order.py
chienandalu: Yes, done :) It was a very preliminary version.

sale_attached_product/models/sale_order.py
chienandalu: Also done the way it is done in `sale_coupon_auto_refresh` with configurable triggers if someone needed too.

chienandalu:
> * do you think it will work out of the box in a website_sale context ?

With my last update it does! I wonder anyway if we should disable the line editing so the eCommerce customer could understand that those products go altogether. It'd be a work for an extension though.

> * did you planned to work on a `pos_attached_product` module ?

Not in our plans I'm afraid.

legalsylvain:
> > did you planned to work on a pos_attached_product module ?

> Not in our plans I'm afraid.

OK. I'll work on that topic in the monthes.

thanks for all the answers regarding the other points.

sale_attached_product/README.rst
```diff
@@ -0,0 +1,114 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure free product rewards from domains:
```
pedrobaeza: This text is not correct

sale_attached_product/README.rst
```diff
@@ -0,0 +1,114 @@
...
+  - `sale_attached_product.auto_update_attached_lines`
+
+Otherwise, the lines will be added, but they can be modified, deleted, etc.
+
+Usage
+=====
+
+Now that you have your product configured:
+
+- Place a new sale order and then add that product in a new line.
```
pedrobaeza: Put numbered list with `#.`

sale_attached_product/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Attached products",
```
pedrobaeza:
```suggestion
    "name": "Attached products in sales",
```

sale_attached_product/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "summary": "Define products that will be added automatically when adding another",
```
pedrobaeza:
```suggestion
    "summary": "Define products that will be added automatically when adding another in a sales order",
```

sale_attached_product/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    attached_product_ids = fields.Many2many(
+        comodel_name="product.product",
+        relation="product_attached_rel",
+        string="Attached Products",
+        help="Similar to optional products although they're added automatically to the"
+        "sale order and they're removed when the main product goes away.",
```
pedrobaeza:
```suggestion
        help="Similar to optional products, although they're added automatically to the"
        "sale order and optionally removed when the main product goes away.",
```

sale_attached_product/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+        check_company=True,
```
pedrobaeza: I think we should free this to allow that one company-free product may have company-specific attached products

sale_attached_product/models/sale_attached_product_mixin.py
```diff
@@ -0,0 +1,83 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleAttachedProductMixin(models.AbstractModel):
+    _name = "sale.attached.product.mixin"
+    _description = "Mixin class for sale attached product features"
+
+    auto_refresh_attached_product_triggers_data = fields.Binary(
```
pedrobaeza: Binary? It seems very hackish if you want to store something. You should at least convert it to JSON, and better use a Char.

sale_attached_product/models/sale_attached_product_mixin.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2021 Tecnativa - David Vidal
+# Copyright 2021 Camptocamp - Silvio Gregorini
```
pedrobaeza: Why this copyright?

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+        triggers.update({"order_line.auto_refresh_attached_product_triggers_data"})
+        return triggers
+
+    def _get_attached_line_values_product(self, line, product):
+        """Prepare the values for the attached line. This is used for creating or
+        updating."""
+
+        def _execute_onchanges(records, field_name):
+            """Helper methods that executes all onchanges associated to a field."""
+            for onchange in records._onchange_methods.get(field_name, []):
```
pedrobaeza, line 28: Is this the same as `sale_coupon`? This is missing chained onchange moves.

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+            self.recompute_attached_products()
+        return res
+
+
+class SaleOrderLine(models.Model):
+    _name = "sale.order.line"
+    _inherit = ["sale.order.line", "sale.attached.product.mixin"]
+
+    is_attached_line = fields.Boolean(
+        help="Falg products that are attached to their main counterpart"
```
pedrobaeza:
```suggestion
        help="Flag products that are attached to their main counterpart"
```

pedrobaeza: And why do you need this one having the many2one?

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+    attached_from_line_id = fields.Many2one(comodel_name="sale.order.line")
+    attached_line_ids = fields.One2many(
+        comodel_name="sale.order.line", inverse_name="attached_from_line_id",
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self._check_skip_attached_product_refresh():
+            return super().create(vals_list)
+
```
pedrobaeza: Avoid empty lines inside a method.

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+        self_ctx = self.with_context(skip_auto_refresh_attached_product=True)
+        lines = super(SaleOrderLine, self_ctx).create(vals_list)
+        lines.mapped("order_id").recompute_attached_products()
+        return lines
+
+    def write(self, vals):
+        if self._check_skip_attached_product_refresh():
+            return super().write(vals)
+
```
pedrobaeza: Avoid empty lines inside a method.

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+        # the method will return an empty set.
+        new_triggers = self._new_trigger()
+        if old_data != new_data or any(x in new_triggers for x in vals):
+            (old_orders | new_orders).recompute_attached_products()
+        return res
+
+    def unlink(self):
+        if self._check_skip_attached_product_refresh():
+            return super().unlink()
+
```
pedrobaeza: Avoid empty lines inside a method.

sale_attached_product/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+                lines_to_remove += line.attached_line_ids
+                continue
+            # For every unit of the main line there will another of the attached one
+            attached_line_qtys = set(line.attached_line_ids.mapped("product_uom_qty"))
+            if any(q != line.product_uom_qty for q in attached_line_qtys):
+                line.attached_line_ids.update({"product_uom_qty": line.product_uom_qty})
+                # Trigger possible pricelist changes
+                for attached_line in line.attached_line_ids:
+                    attached_line.product_uom_change()
+            attached_products = line.product_id.product_tmpl_id.attached_product_ids
```
pedrobaeza: Filtered by company according the new specification

pedrobaeza (changes requested):

sale_attached_product/models/sale_order.py
chienandalu: Mainly to delete them when the relation is gone as ondelete cascade doesn't do the trick

sale_attached_product/models/sale_order.py
chienandalu: The alternative would be to use ` onchange_helper` to play the onchanges performantly as we saw `Form` isn't a very good alternative for these cases.

sale_attached_product/models/sale_attached_product_mixin.py
chienandalu: Or maybe we could simplify it quite much as the checks aren't very heavy anyway and we can deal with some uneeded checks

sale_attached_product/README.rst
chienandalu: Fixed

sale_attached_product/README.rst
chienandalu: Fixed

sale_attached_product/models/product_template.py
chienandalu: Ok, fixed to behave that way. I set it up like that because optional_product_ids had that definition

sale_attached_product/models/sale_order.py
chienandalu: I added a common method to do so

sale_attached_product/models/sale_attached_product_mixin.py
chienandalu: I finally simplified it quite a lot keeping the performant approach but without hacky stuff

pedrobaeza (approved):

PR #2094, opened 2022-06-28
[15.0][MIG] sale_order_priority

JuanyDForgeflow:
This module adds the field *Priority* in sale order lines and sale orders:
priority of the sale order is computed as the maximum of the priorities of its
lines, setting the priority in the order sets the priority of all its lines
accordingly.

When a picking is created as a result of sale order confirmation,
the created procurement inherits the priority of the order,
then the stock moves and the picking inherit the procurement's priority.


MiquelRForgeFlow (approved):

sale_order_priority/models/stock_picking.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models
+
+
+class Picking(models.Model):
+    _inherit = "stock.picking"
+
+    def action_assign(self):
+        res = super(Picking, self).action_assign()
```
cesar-tecnativa: super().action_assign()

cesar-tecnativa (changes requested):

sale_order_priority/models/stock_picking.py
JuanyDForgeflow: pre-commit fails if I don't return the super

MiquelRForgeFlow: the suggestion was not to delete the `res=`  + `return`. please, put this again. the suggestion was to delete the content inside super()

JuanyDForgeflow: @MiquelRForgeFlow @cesar-tecnativa Done! :)

sale_order_priority/models/stock_picking.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models
+
+
+class Picking(models.Model):
+    _inherit = "stock.picking"
+
+    def action_assign(self):
+        res = super().action_assign()
+        picking_priority = self.env.context.get("sale_priority")
+        self.priority = picking_priority if picking_priority else 0
```
GuillemCForgeFlow:
```suggestion
    def action_confirm(self):
        res = super().action_confirm()
        picking_priority = self.env.context.get("sale_priority", False)
        self.write({"priority": picking_priority or 0})
```

GuillemCForgeFlow: From what I've tested, the action_assign method doesn't seem to do the trick anymore. With the action_assing the module was not working, the priority of the order was not passed to the picking. Applying the changes,  it should work again! :+1:

GuillemCForgeFlow (changes requested):

GuillemCForgeFlow (approved):
LGTM! :+1:

sale_order_priority/models/stock_picking.py
JuanyDForgeflow: @GuillemCForgeFlow Thanks! it's working now

AaronHForgeFlow:
@cesar-tecnativa the comments are attended. Can you please update the review? Thanks

AaronHForgeFlow (approved):
Code LG!

manuelregidor (approved):
Functional and technical review. LGTM

ValentinVinagre (approved):
LGTM 👍🏻

manuelregidor:
@HaraldPanten

ValentinVinagre:
@HaraldPanten

HaraldPanten (approved):

PR #2099, opened 2022-06-29
[15.0][FIX] sale_invoice_plan: adding fully-qualified external id

cesar-tecnativa:
@Tecnativa TT36525

### DIFF
--- a/sale_invoice_plan/views/sale_view.xml
+++ b/sale_invoice_plan/views/sale_view.xml
@@ -84,7 +84,7 @@
                     />
                     <field
                         name="invoice_plan_ids"
-                        context="{'tree_view_ref': 'view_sale_invoice_plan_tree'}"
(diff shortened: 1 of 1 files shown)


sale_invoice_plan/views/sale_view.xml
```diff
@@ -160,7 +160,10 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">sale.invoice.plan</field>
         <field name="view_mode">tree</field>
-        <field name="view_id" ref="view_sale_invoice_plan_tree_readonly" />
+        <field
+            name="view_id"
+            ref="sale_invoice_plan.view_sale_invoice_plan_tree_readonly"
```
pedrobaeza: There's no need here of this. Only the first one. In fact, here it's advisable to use relative XML-IDs.

pedrobaeza (approved):

pedrobaeza:
Let's fast-track this for avoiding the warning in the log:

/ocabot merge minor

PR #2102, opened 2022-06-30
[FIX] sale_manual_delivery: value must be computed for all records.

adrienpeiffer:


sale_manual_delivery/models/sale_order_line.py
```diff
@@ -36,15 +36,19 @@ def _compute_qty_procured(self):
         Computes the already planned quantities for the given sale order lines,
         based on the existing stock.moves
         """
+        no_procurement_records = self.browse([])
```
rousseldenis, line 39: I would have done:

```
records = self.filtered(lambda line: line.qty_delivered_method == "stock_move")
for line in records:
    line.qty_procured = line._get_qty_procurement(previous_product_uom_qty=False)
(self - records).update({"qty_procured": 0.0})
```

As '|' is not very performant, it's better to do like here under or to build a set of ids, then browse, then update.

rousseldenis (approved):
LGTM.

I would have optimized the compute

aliciagaarzo (approved):

PR #2106, opened 2022-06-30
[13.0][FIX] sale_order_type: default sequence

chienandalu:
When the sale type has no sequence, the order gets the default one. When we change the type and the sequence is the same that that default, we should change the name, as we'd get a new reference of the same sequence.

cc @Tecnativa TT37832

please review @pedrobaeza @victoralmau

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -33,6 +33,19 @@ class SaleOrder(models.Model):
             [("company_id", "in", [company_id, False])], limit=1
         )
 
+    @api.model
+    def _default_sequence_id(self):
+        """We get the sequence in same way the core next_by_code method does so we can
+        get the proper default sequence"""
+        force_company = self.env.context.get("force_company")
+        if not force_company:
+            force_company = self.company_id.id or self.env.company.id
+        return self.env["ir.sequence"].search(
+            [("code", "=", "sale.order"), ("company_id", "in", [force_company, False])],
+            order="company_id",
+            limit=1,
+        )
+
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
         for record in self:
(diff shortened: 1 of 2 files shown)


sale_order_type/models/sale.py
```diff
@@ -33,6 +33,19 @@ def _default_type_id(self):
...
+    @api.model
+    def _default_sequence_id(self):
+        """We get the sequence in same way the core next_by_code method does so we can
+        get the proper default sequence"""
+        force_company = self.env.context.get("force_company")
+        if not force_company:
+            force_company = self.company_id.id or self.env.company.id
+        return self.env["ir.sequence"].search(
+            [("code", "=", "sale.order"), ("company_id", "in", [force_company, False])],
+            order="company_id",
```
cesar-tecnativa, line 45: why order="company_id" ?

sale_order_type/models/sale.py
```diff
@@ -101,13 +114,25 @@ def create(self, vals):
     def write(self, vals):
         """A sale type could have a different order sequence, so we could
         need to change it accordingly"""
+        force_company = self.env.context.get("force_company")
+        if not force_company:
+            force_company = self.env.company.id
+        default_sequence = self._default_sequence_id()
```
cesar-tecnativa, line 117: Lines 117-119 force_company are included in function

sale_order_type/models/sale.py
```diff
@@ -101,13 +114,25 @@ def create(self, vals):
...
         if vals.get("type_id"):
             sale_type = self.env["sale.order.type"].browse(vals["type_id"])
             if sale_type.sequence_id:
                 for record in self:
+                    # An order with a type without sequence would get the default one.
+                    # We want to avoid changing the order reference when the new
+                    # sequence has the same default sequence.
+                    ignore_default_sequence = (
+                        not record.type_id.sequence_id
+                        and sale_type.sequence_id == default_sequence
```
cesar-tecnativa, line 127: should be sale_type.sequence_id.id == default_sequence.id?

cesar-tecnativa (changes requested):

victoralmau (approved):

sale_order_type/models/sale.py
chienandalu: We don't need to. We're comparing at object level.

sale_order_type/models/sale.py
chienandalu: Mmm, yes. You're right

sale_order_type/models/sale.py
chienandalu: Is the same order `next_by_code()` does it: https://github.com/OCA/OCB/blob/13.0/odoo/addons/base/models/ir_sequence.py#L280

cesar-tecnativa (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to upper versions

PR #2111, opened 2022-07-05
[14.0][IMP] sale_automatic_workflow_job: Refactor and fix

grindtildeath:
The extra filtering and return strings added in sale_automatic_workflow
are only useful if sale_automatic_workflow_job is used, as:

- the returned values are not used in sale_automatic_workflow
- the extra check sonly makes sense when a job is running these functions
asynchronously.

Check if the recordset being passed as parameter to the job function
still exists when the job is being run to avoid issues on deleted
records.

Said issues could be a CacheMiss or worse case, the job being stuck in
started state.

### DIFF
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -36,27 +36,14 @@ class AutomaticWorkflowJob(models.Model):
     )
 
     def _do_validate_sale_order(self, sale, domain_filter):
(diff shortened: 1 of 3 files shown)


grindtildeath:
ping @kittiu

mmequignon (approved):

leemannd (approved):

ivantodorovich (approved):

cesar-tecnativa (approved):

PR #2113, opened 2022-07-05
[15.0][IMP] sale_automatic_workflow: make datetime test more reliable.

cesar-tecnativa:
Previous version sometime fails depending on when the test was executed.


cesar-tecnativa:
@pedrobaeza please check

cesar-tecnativa:
@CarlosRoca13 please review

CarlosRoca13 (approved):

pedrobaeza (changes requested):
If the rest of the test is using relative deltas to today, as it's the case in https://github.com/OCA/sale-workflow/blob/8be2de4806f61a0e83bfc1ab70a579a73e814b07/sale_automatic_workflow/tests/common.py#L44 , this can't be fixed. You have to use instead a timezone aware `date reference`, or fix all the dates across the tests.

cesar-tecnativa:
It is not necessary, I close it

PR #2120, opened 2022-07-13
[14.0][FIX] product_form_sale_link: replace sales_count by a coherent one

MiquelRForgeFlow:
The sales_count field is already used in odoo for another button, which calculates the sum of the "product_uom_qty" of that product for each sale line instead of the number of sale lines.

Now, the number in the button show the correct number of sale lines.

### DIFF
--- a/product_form_sale_link/__init__.py
+++ b/product_form_sale_link/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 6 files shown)


mariadforgeflow (approved):
LGTM!

MiquelRForgeFlow:
@rousseldenis

LoisRForgeFlow (approved):
:+1:

PR #2121, opened 2022-07-13
[14.0][MIG] sale_pricelist_triple_discount

are-agilebg:
Adapted from https://github.com/OCA/sale-workflow/pull/1643


sale_pricelist_triple_discount/model/pricelist.py
```diff
@@ -1,22 +1,41 @@
...
         help="Third discount applied on a sale order line.",
         default=0.0,
     )
+
+    def _compute_price(self, price, price_uom, product, quantity=1.0, partner=False):
+        """Compute the unit price of a product in the context of a pricelist
+        application. The unused parameters are there to make the full
+        context available for overrides.
+        """
+        price = super(ProductPricelistItem, self)._compute_price(
```
mmequignon:
```
price = super()._compute_price(
```
It is possible to call `super().method()` since odoo 12.0 :blush: 
no-blocker

mmequignon (commented):
Could you please add a few tests, please?
Thanks !

are-agilebg:
@mmequignon I've done some improvements as you've requested: looking forward for your feedback!

sale_pricelist_triple_discount/view/pricelist_view.xml
```diff
@@ -0,0 +1,14 @@
...
+<odoo>
+        <record model="ir.ui.view" id="product_pricelist_view_form_discount">
+            <field name="name">product.pricelist.item.form.discount</field>
+            <field name="model">product.pricelist.item</field>
+            <field name="inherit_id" ref="product.product_pricelist_item_form_view" />
+            <field name="arch" type="xml">
+                <field name="price_discount" position="after">
+                    <field name="discount2" />
+                    <field name="discount3" />
+                </field>
```
v-garcies, line 11: I don't see the discount2 and discount3 fields appear anywhere...

I suggest replace this part by this :

```suggestion
            <xpath expr="//field[@name='percent_price']/.." position="after">
                <label for="discount2" attrs="{'invisible':[('compute_price', '!=', 'percentage')]}"/>
                <div attrs="{'invisible':[('compute_price', '!=', 'percentage')]}">
                    <field name="discount2" class="oe_inline" attrs="{'invisible':[('compute_price', '!=', 'percentage')]}"/>
                    %
                </div>
                <label for="discount3" attrs="{'invisible':[('compute_price', '!=', 'percentage')]}"/>
                <div attrs="{'invisible':[('compute_price', '!=', 'percentage')]}">
                    <field name="discount3" class="oe_inline" attrs="{'invisible':[('compute_price', '!=', 'percentage')]}"/>
                    %
                </div>
            </xpath>
```
For this result :
Capture d’écran 2022-09-11 à 18 12 25

v-garcies (changes requested):
I don't see the discount2 and discount3 fields in pricelist form, so I suggest you a change in view definition 👍

sale_pricelist_triple_discount/view/pricelist_view.xml
are-agilebg: The fields discount2 and discount3 should appear when you select the 'formula' strategy for price calculation. In this regards the module behave exactly like the version is been migrated from. I've started from this stale PR (https://github.com/OCA/sale-workflow/pull/1643) migrating the module to 12.0, and adapted to the 14.0 which we're interested in targeting.

sale_pricelist_triple_discount/model/pricelist.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if self.compute_price != "formula":
+            return super()._compute_price(price, price_uom, product, quantity, partner)
+        else:
+            # Here we save price_round, then we temporarily unset it
+            # to ensure super() calculates the price without applying
+            # any rounding. We will apply the price_round only to the
+            # subtotal after all discounts have been applied.
+            store_price_round = self.price_round
+            self.price_round = None
```
lorenzoallegrucci, line 40: Hi, I'm using this module and in case of a logged in portal user there is a permission error on `self.price_round = None` and `self.price_round = store_price_round`.

matteoopenf:
@are-agilebg  can you do the fix? thanks

sale_pricelist_triple_discount/model/pricelist.py
```diff
@@ -1,22 +1,48 @@
...
+            price = super()._compute_price(price, price_uom, product, quantity, partner)
+            self.price_round = store_price_round
+
+            if self.pricelist_id.discount_policy == "with_discount":
```
SirAionTech, line 44: `_compute_price` does not care about the discount policy of the pricelist, it just computes the price according to the current price rule so this price recomputation should be always performed.

sale_pricelist_triple_discount/model/pricelist.py
```diff
@@ -1,22 +1,48 @@
...
+
+    def _compute_price(self, price, price_uom, product, quantity=1.0, partner=False):
+        """Compute the unit price of a product in the context of a pricelist
+        application. The unused parameters are there to make the full
+        context available for overrides.
+        """
+        self.ensure_one()
+
+        if self.compute_price != "formula":
+            return super()._compute_price(price, price_uom, product, quantity, partner)
```
SirAionTech, line 33: Here and below
```suggestion
            return super()._compute_price(price, price_uom, product, quantity=quantity, partner=partner)
```
otherwise if the kwargs order is changed, this leads to bugs hard to detect.

sale_pricelist_triple_discount/model/pricelist.py
```diff
@@ -1,22 +1,48 @@
...
+        else:
+            # Here we save price_round, then we temporarily unset it
```
SirAionTech, line 35: In the **formula** computation there are also margins and surcharge that are not considered here.
This leads to an unexpected behavior: if the user sets that he wants at least a 100 margin and applies the additional discounts, he would end up having less than a 100 margin.

As a end user I would expect that the discounts are applied at the same time as the formula's discount (`price_discount`).

Instead of reversing most of the formula, it is probably better to simplify the module behavior and only apply the additional discounts in the **percentage** case, what do you think?
See the proposed approach in https://github.com/OCA/sale-workflow/pull/2773.

sale_pricelist_triple_discount/model/sale.py
```diff
@@ -1,55 +1,192 @@
...
+from functools import partial
+
+from odoo import api, models, tools
+from odoo.tools.misc import formatLang
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _amount_by_group(self):
```
SirAionTech, line 15: Could you please clarify why is this needed? I don't see the new fields **discount*** used in this override

sale_pricelist_triple_discount/tests/test_sale_pricelist_triple_discount.py
```diff
@@ -0,0 +1,147 @@
...
+from odoo.tests.common import SavepointCase
+
+
+class PricelistTripleDiscount(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create({"name": "Mr. Odoo"})
+
+        # A pricelist showing discounts to user and with three discounts set
```
SirAionTech, line 13: Could you please add a **with_discount** test case?

sale_pricelist_triple_discount/model/sale.py
```diff
@@ -1,55 +1,192 @@
...
-        if product and pricelist:
-            value = res["value"]
-            current_pricelist = self.env["product.pricelist"].browse(pricelist)
+        if rule_id:
+            rule = self.env["product.pricelist.item"].browse(rule_id)
+            read_rule = rule.read(["discount2", "discount3"])[0]
+            self.discount2 = read_rule["discount2"] or 0.00
+            self.discount3 = read_rule["discount3"] or 0.00
+
+    @api.depends(
```
SirAionTech, line 120: It is safer to only add the new fields you need to depend on, it works because `extend` is used in https://github.com/odoo/odoo/blob/4fd3b7a062e9f8107587ceed98ebb2f51627a4d6/odoo/fields.py#L452

SirAionTech (changes requested):
Thanks for the PR!
Pricelists have changed a lot since `8.0` so I think this module can be simplified, see the comments below

rousseldenis:
@are-agilebg What's the status of this ?

rousseldenis:
@tafaRU

PR #2124, opened 2022-07-15
[13.0][FIX] sale_order_type: Analytic account on copy

cesar-tecnativa:
@Tecnativa TT38012

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -160,6 +160,13 @@ class SaleOrder(models.Model):
             )
         return sequence
 
+    def copy(self, default=None):
+        res = super().copy(default)
+        sale_type = res.type_id or self.type_id
+        if sale_type.analytic_account_id:
+            res["analytic_account_id"] = sale_type.analytic_account_id.id
+        return res
+
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -1,6 +1,7 @@
 # Copyright 2015 Oihane Crucelaegui - AvanzOSC
 # Copyright 2017 Pierre Faniel - Niboo SPRL (<https://www.niboo.be/>)
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2022 Tecnativa - César A. Sánchez
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 from freezegun import freeze_time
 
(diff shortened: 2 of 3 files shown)


cesar-tecnativa:
@sergio-teruel please review

pedrobaeza:
Incorrect commit and PR message. Follow the guidelines (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message)

- Short on main line
- Module name
- TT on the content, not on the main line

```
[FIX] sale_order_type: Analytic account on copy

When a sale order is duplicated, we set the default account analytic
account that is configured in the order type.

TT38012
```

cesar-tecnativa:
Done

sale_order_type/models/sale.py
```diff
@@ -125,6 +125,13 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    def copy(self, default=None):
+        if self.type_id and self.type_id.analytic_account_id:
+            default = dict(
+                default or {}, analytic_account_id=self.type_id.analytic_account_id.id,
+            )
+        return super().copy(default)
```
pedrobaeza: You should read the type on the defaults, as you don't know if there's any override or default value set on the DB (and respect kw args):

```suggestion
        res = super().copy(default=default)
        sale_type = self.env["sale.order.type"].browse(res.get("type_id", False)) or self.type_id
        if sale_type.analytic_account_id:
            res["analytic_account_id"] = sale_type.analytic_account_id.id
        return res
```

pedrobaeza (changes requested):

sergio-teruel:
@cesar-tecnativa Can you add some test to coverage this.?

cesar-tecnativa:
> @cesar-tecnativa Can you add some test to coverage this.?

done

sale_order_type/models/sale.py
```diff
@@ -125,6 +125,13 @@ def _prepare_invoice(self):
             res["sale_type_id"] = self.type_id.id
         return res
 
+    def copy(self, default=None):
```
pedrobaeza, line 163: Doing this in the method `copy_data` you avoid an extra write, although you would need to do a browse over ` type_id` ID result if exists.

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -218,3 +231,15 @@ def test_sale_order_quotation_validity(self):
         order = self.create_sale_order()
         order.onchange_type_id()
         self.assertEqual(fields.Date.to_string(order.validity_date), "2022-01-11")
+
+    def test_sale_copy_function(self):
+        """
+        Test when duplicating the sale order the account analytic account is set.
+        """
```
pedrobaeza, line 314: PEP257 indicates that the docstring should be put this way:

```suggestion
        """Test when duplicating the sale order the account analytic account is set.
        """
```

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -50,6 +50,14 @@ def setUp(self):
         self.immediate_payment = self.env.ref("account.account_payment_term_immediate")
         self.sale_pricelist = self.env.ref("product.list0")
         self.free_carrier = self.env.ref("account.incoterm_FCA")
+        self.analytic_account = self.env["account.analytic.account"].create(
```
pedrobaeza, line 54: You should put these changes in a separate commit with the message:

```
[IMP] sale_order_type: Don't depend on demo data for tests
```

for not mixing unrelated things in the diff.

pedrobaeza:
@carlosdauden please attend my comments.

pedrobaeza:
@sergio-teruel can you please do the requested changes?

rousseldenis:
@sergio-teruel What's the status of this ?

PR #2128, opened 2022-07-17
[15.0][MIG] sale_order_line_price_history: Migration to 15.0

ernestotejeda:
Cc @Tecnativa TT36522

Supersede https://github.com/OCA/sale-workflow/pull/1923:

- I kept the changes made by @bizzappdev and but I have done what i proposed here https://github.com/OCA/sale-workflow/pull/1923#pullrequestreview-1016981475 since @bizzappdev doesn't answer to my comment
- I added a line to roadmap readme section


ernestotejeda:
I just added a commit as a workaround for one of the roadmap specifications, could you check it @chienandalu @pedrobaeza ?

sale_order_line_price_history/wizards/sale_order_line_price_history.xml
```diff
@@ -0,0 +1,84 @@
...
+                </group>
+                <group string="Sale history">
+                    <field name="line_ids" nolabel="1" readonly="1">
+                        <form>
+                            <div
+                                class="oe_button_box"
+                                name="button_box"
+                                attrs="{'invisible': [('history_sale_order_line_id', '=', False)]}"
+                            >
+                                <button
```
pedrobaeza, line 24: Put this button as well in the tree view

chienandalu: This one belongs to the tree inside the sale form.

chienandalu (approved):

sale_order_line_price_history/wizards/sale_order_line_price_history.xml
pedrobaeza: I know, but that's why I want it as well in the tree view.

ernestotejeda: When the popup opens, the wizard record hasn't been created yet, so the button on those lines will be disabled (normal odoo behavior).

pedrobaeza: I'm not saying that. I just want to have the button "Set price" on the tree view of the wizard with an icon, only when the related sale line already exists.

ernestotejeda: I wrote the limitation in the roadmap in the migration commit, please, check it out.

pedrobaeza (approved):

PR #2130, opened 2022-07-18
[13.0][FIX] sale_blanket_order Use display_name instead of name

sergioeix:
There is an issue with variants if you use name instead of display_name when you print picking reports, the variant info is not shown in name (only template description)
exp -> template name= iphone and attribute memory=16gb,32gb
with name on the picking reports only iphone is shown with display_name iphone 16gb or iphone 32gb is shown


sergioeix:
Hi guys can you review this small change:
@AdriaGForgeFlow 
@JordiBForgeFlow 
@tafaRU

rousseldenis:
@sergioeix Thanks for this.

IMHO, the good fix would use _get_sale_order_line_multiline_description_sale() function from sale_order_line as this this the one used by Odoo for that purpose.

See https://github.com/odoo/odoo/blob/13.0/addons/sale/models/sale.py#L1784

PR #2150, opened 2022-08-05
[13.0][FIX] sale_order_type: sale type is reset to partner value when journal changes

carlosdauden:
If a sale_type has journal_id, the journal change causes the company to be established, which causes that sale_type to be computed and the sale_type assigned to the customer to be established. This means that, in practice, it is not possible to select a type of order other than that of the partner if it is associated with a journal.

@Tecnativa TT38438

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -1,7 +1,9 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2020 Tecnativa - Carlos Dauden
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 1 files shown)


sale_order_type/models/account_move.py
```diff
@@ -1,7 +1,9 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2020 Tecnativa - Carlos Dauden
```
pedrobaeza, line 3:
```suggestion
# Copyright 2022 Tecnativa - Carlos Dauden
```

sale_order_type/models/account_move.py
```diff
@@ -17,7 +19,7 @@ class AccountMove(models.Model):
         copy=True,
     )
 
-    @api.depends("partner_id", "company_id")
+    @api.depends("partner_id")
```
pedrobaeza, line 22: Although I understand the reason to remove that on the use case, this is limiting the correct logic. Why the company is always being assigned no matter if it's the same? Isn't that what we have to tackle?

rousseldenis:
@carlosdauden What's the status of this ?

PR #2152, opened 2022-08-07
[12.0][UPD] Repo template

alexey-pelykh:


sbidoul:
It's red

alexey-pelykh:
@sbidoul true, tests are truly failing while the shouldn't be

alexey-pelykh:
@MiquelRForgeFlow @serpentcs-dev1 @AaronHForgeFlow any ideas what may be causing `sale_procurement_group_by_commitment_date` tests to fail?

PR #2154, opened 2022-08-10
[15.0][FIX] sale_procurement_group_by_line: group negative moves when reducing qty in a sales order

AaronHForgeFlow:
Reducing the qty in a sales order in v15 will update the stock moves by creating a negative stock moves and then merging it with the existing stock moves. The thing is, with this module installed Odoo will not merge the negative move. That will cause a return picking to be created, weird. 

Steps to reproduce:

- Create a sales order with one sale order line and confirm it
- Add another line for the same product and save
- The customer disagrees, so you set the quantity in this second line to zero
- Odoo creates and incoming shipment from customers to stock, instead of reducing the existing stock move to have zero quantity.

Useful links:
A procurement is launch when saving the sale order line with zero qty
https://github.com/odoo/odoo/blob/15.0/addons/sale_stock/models/sale_order.py#L434

The procurement has negative qty
https://github.com/odoo/odoo/blob/15.0/addons/sale_stock/models/sale_order.py#L583

A negative stock move is created:
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_rule.py#L268

Odoo tries to merge within the same picking
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1237

For some reason, with the sale_procurement_group_by_line installed, this is filter returns nothing and the negative move is not merged:
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L876

When it does not merge it converts the move to a return
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1254

The solution is to include the stock moves of the procurement group within the stock moves to do the merge. Those are not included by default.

Test added that checks this use case.

cc @ForgeFlow

### DIFF
--- a/sale_procurement_group_by_line/model/__init__.py
+++ b/sale_procurement_group_by_line/model/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import sale
+from . import stock_move
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -115,9 +115,9 @@ class SaleOrderLine(models.Model):
                 errors.append(error.args[0])
         if errors:
             raise UserError("\n".join(errors))
-        return super()._action_launch_stock_rule(
-            previous_product_uom_qty=previous_product_uom_qty
-        )
+        return super(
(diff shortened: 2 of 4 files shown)


MiquelRForgeFlow (approved):
LGTM

rousseldenis (approved):

PR #1453, opened 2021-02-09
[14.0][ADD]Added Module sale_order_line_menu

Freni-OSI:

PR #2155, opened 2022-08-11
[MIG][12.0] Backport of sale_order_line_menu to 12.0

theangryangel:
Standard backport.


sale_order_line_menu/views/sale_order_line_views.xml
```diff
@@ -1,9 +1,19 @@
 <odoo>
+    <record model="ir.ui.view" id="view_order_line_tree">
+        <field name="name">sale.order.line.tree (in sale_order_line_menu)</field>
+        <field name="model">sale.order.line</field>
+        <field name="inherit_id" ref="sale.view_order_line_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='order_id']" position="before">
+                <field name="currency_id" invisible="1" />
```
rousseldenis, line 8: @theangryangel Is there a reason for this ?

rousseldenis:
@theangryangel

rafaelbn:
Hello friends! This module is the same (IMHO) than this one from 11:

- https://github.com/OCA/sale-workflow/tree/11.0/sale_order_line_input (available in all versions from v11)

sale_order_line_menu was created in v14 and after backported or forwardported.

Just to know, just in case to not have duplications  of a mature module

My best regards and wishes!

rafaelbn:
@dreispt did you know about this https://github.com/OCA/sale-workflow/pull/2155#issuecomment-1377775793 ?

I think you create the new one in 14 because the name of the module wasn't explicitr https://github.com/OCA/sale-workflow/pull/1453 😄 

Best regards! And happy new year!

dreispt:
Thanks for pointing that out. Reading the module description it server a different purpose from this.
The goal here is to make order line list available for querying and reporting.
We don't want users to create new lines there.

So I think they are both valid modules, and I see the case for them to coexist.
For that they should probably use more specific menu names.

PR #2158, opened 2022-08-12
[MIG] sale_cancel_reason: Migration to 14.0

ao-landoo:
`sale.order.cancel` transient model is defined on sale module, so the module is changed so that is inherits core `sale.order.cancel` model


ao-landoo:
@CasVissers-360ERP @atchuthan @theangryangel Can you review the PR? The old one had no activity (#2158 1627)

CasVissers-360ERP (approved):
Functional review. All changes missed in previous PR are fixed.

rousseldenis (approved):
LGTM

ao-landoo:
@pedrobaeza Can it be merged if you see it OK?

CasVissers-360ERP:
@ao-landoo Forgot one thing we might need to improve shouldn't cancel_reason_id be copy=False?

PR #2063, opened 2022-06-09
[13.0][backport] sale_automatic_workflow_job

damdam-s:
backport of #2054 to v13

PR #2170, opened 2022-08-26
13.0 mig back sale automatic workflow job

jcoux:
**Work in progress...**
**TEMP**

To try to fix unit on https://github.com/OCA/sale-workflow/pull/2063 (Will update the original PR once time I will be success)


jcoux:
Pushed on https://github.com/OCA/sale-workflow/pull/2063

PR #2173, opened 2022-08-26
[IMP] Add section note to sale blanket order

santostelmo:
### Context
Improve the usability of the OCA Blanket order module by adding the section/note system available as standard in sale orders and customer invoices.

The improvement does NOT include the following:
- Sale orders should include sections and notes coming from the blanket order. That feature requires more analyses because we need a mechanism to decide if a section is added or not depending on the existence of section child order lines.


leemannd (approved):

ajaniszewska-dev (approved):

cyrilmanuel (commented):
seems to be good but i'm not an expert

PR #2175, opened 2022-08-26
[15.0][MIG] sale require po doc

Hardik-OSI:


Chanakya-OSI (approved):

Murtaza-OSI (approved):

Murtaza-OSI:
@Hardik-OSI Fix per-commit

hitrosol:
I think we can solve the use case by using the sale_exception.

PR #2181, opened 2022-09-02
[12.0][FIX] sale_invoice_plan: rounds the value to the rounding precision of the product's unit of measure

marcelsavegnago:
Testing product invoicing to 0.01 rounding accuracy generates a difference between the amount invoiced according to invoices VS the amount invoiced on the sales order.

Unit of messure: hour
Quantity: 1.0
Price: 100.00
Installment quantity: 3


marcelsavegnago:
ping @rvalyi

rvalyi (approved):

kittiu (approved):

PR #2186, opened 2022-09-05
[14.0] [IMP] `sale_invoice_blocking`: Make field readonly for cancelled orders

ivantodorovich:
To be aligned with core behaviour where most fields are readonly when the order is cancelled.


rousseldenis (approved):
LGTM

ivantodorovich:
Could we merge it? It's pretty straightforward

yankinmax (approved):

PR #2197, opened 2022-09-14
[14.0][IMP] sale_global_discount: improve README with reference to the one of base_global_discount

tafaRU:
Without this *Global Discounts* doesn't show up!

### DIFF
--- a/sale_global_discount/readme/CONFIGURE.rst
+++ b/sale_global_discount/readme/CONFIGURE.rst
@@ -1,14 +1,1 @@
-To configure this module, you need to:
-
-#. Go to *Settings > Technical > Parameters > Global Discounts*
-#. Add a new discount.
-#. Choose the discount scope (sales or purchases).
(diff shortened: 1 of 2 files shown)


pedrobaeza:
But this is already stated in the base module: https://github.com/OCA/server-backend/tree/14.0/base_global_discount#configuration

Maybe a note about seeing also the other README?

tafaRU:
@pedrobaeza thanks for the review!
Do you think now can go well? I remai awaiting your confirmation before squashing.

pedrobaeza:
Putting an absolute path will require a maintenance on each version. I think It's better to just say to see the configure section of the `base_global_discount` module.

tafaRU:
@pedrobaeza done thanks

pedrobaeza (approved):

houssine78 (approved):

houssine78:
Hi @pedrobaeza I think it ready to be merge. thanks

PR #2198, opened 2022-09-15
[15.0][ADD]: new module to block delivery if sale order is blocked

imlopes:
This module allow to block a delivery if the corresponding sale order has checked Block Delivery field.

To unblock the delivery, on sale we need to click on release button and if needed, we can fill a reason.

### DIFF
--- a/sale_stock_picking_delivery_block/__init__.py
+++ b/sale_stock_picking_delivery_block/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import wizard
--- a/sale_stock_picking_delivery_block/__manifest__.py
+++ b/sale_stock_picking_delivery_block/__manifest__.py
@@ -1,0 +1,20 @@
+# Copyright 2022 Camptocamp (https://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale Stock Picking Delivery Block",
+    "summary": "Block delivery orders when sale order is blocked",
+    "version": "15.0.1.0.0",
+    "category": "Sales",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Camptocamp, Italo LOPES, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["sale_stock"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/sale_order_view.xml",
+        "views/stock_picking_view.xml",
+        "wizard/release_delivery_wizard.xml",
+    ],
+    "installable": True,
+}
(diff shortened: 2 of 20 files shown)


ivantodorovich (changes requested):
Could you add some unit tests? 🙏🏻

rousseldenis:
@imlopes There is already https://github.com/OCA/sale-workflow/tree/15.0/sale_stock_picking_blocking

epanisset:
> @imlopes There is already https://github.com/OCA/sale-workflow/tree/15.0/sale_stock_picking_blocking

Hello @rousseldenis, this one is little bit different. 
The one you point block the creation of the stock.picking from the sale order confirmation (meaning that if you have other operation depending of the creation of the stock.picking (needs in stock forecast, manufacturing order in MTO or other point), you won't be able to have them. 

This one will block the confirmation of stock.picking, not its creation. 
For me, it's not a duplicate.

rousseldenis:
IMHO, we should try to converge and make those two modules complementary.

The common is the delivery block from the sale order.

It should implements a 'blocking strategy' :

- Do not create stock moves
- Create stock moves and don't display validate button.

What do you think ?

@LoisRForgeFlow @AaronHForgeFlow

LoisRForgeFlow:
Hi @imlopes 

I agree with @rousseldenis, what you are proposing here is similar to the existing module that has been pointed, I think it would be better if you add "blockin strategy" as suggested that can be configured in the sales settings for example. In this module you also need to provide a reason, but it seems a bit different, it would also be great if you can make it converge with the existing module.

My 2 cents.

epanisset:
@rousseldenis @LoisRForgeFlow 
Hello, 
I am ok with the solution to add this Improve and a "sale configuration blocking option" on sale_stock_picking_blocking
But I propose to keep the way this module worked, they are a little bit different. Complicated to make them converged

epanisset:
@rousseldenis @LoisRForgeFlow 

Hello, 
I thought about this module and the one you mentionned [sale_stock_picking_blocking]:

- The one you mention avoid the creation of a stock.picking from a sale.order. 
- This one avoid the confirmation of expedition of a stock.picking (hidding the validate button). 

If we make them merge to a new module with a "blocking strategy", that means we will be able to use only one of the 2 strategies and not both. 

It could be possible to have a client wanting both modules in their sale strategy: 
- First, confirming a sale.order, blocking the stock.picking creation until a first validation is done (invoice payement for example). 
- Second, having a stock.picking to organise the planification, but still blocking the real expedition of the stock.picking until a last validation (a manager checking the expedition address or some issues with the client....) 

For high values product, I can totally picture in my mind a company wanting both modules. 
If I make them merge, then it won't be possible.  

I think we need to let them independant, maybe we can make them merge with a closed name.

Camille0907 (approved):

PR #1341, opened 2020-11-18
14.0 [MIG] sale_elaboration

JSeguiN97:
Migration of sale_elaboration module.

PR #2203, opened 2022-09-28
[14.0][MIG] sale_elaboration: Migration to v14 

CarlosRoca13:
cc @Tecnativa TT37689

please @carlosdauden @sergio-teruel review this


CarlosRoca13:
This PR superseds this other: https://github.com/OCA/sale-workflow/pull/1341

sale_elaboration/models/sale_order.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+def _execute_onchanges(records, field_name):
```
rousseldenis, line 7: I know you don't like onchange_helper, but you could avoid code duplicates like this...

I encourage you to reconsider

sale_elaboration/models/sale_order.py
```diff
@@ -0,0 +1,77 @@
...
+            line.elaboration_note = line.elaboration_id.name
+
+    def _prepare_invoice_line(self, **optional_values):
+        vals = super()._prepare_invoice_line(**optional_values)
+        if self.is_elaboration:
+            vals["name"] = "{} - {}".format(self.order_id.name, self.name)
+        return vals
+
+    @api.depends("product_id")
+    def _compute_is_elaboration(self):
```
rousseldenis, line 75: Wouldn't it be a related stored instead?

pedrobaeza: It misses a comment of the why, but this is because related fields are not initialized with their value on one2many which related field is the inverse_name, so with this we get immediately the value on NewIds.

rousseldenis: Ok a little comment should be great

pedrobaeza: Indeed. That's always my fight :disappointed:

CarlosRoca13:
Added comment and added changes suggested by @victoralmau in other PR. We prefer to keep mantaing `_execute_onchanges` method @rousseldenis

victoralmau (approved):
Code review OK.

sale_elaboration/models/product_elaboration.py
```diff
@@ -0,0 +1,46 @@
...
+        ("name_uniq", "unique(name)", "Name must be unique!"),
+        ("code_uniq", "unique(code)", "Code must be unique!"),
+    ]
+
+    @api.model
+    def name_search(self, name, args=None, operator="ilike", limit=100):
+        """Give preference to codes on name search, appending
+        the rest of the results after.
+        """
+        args = args if args else []
```
carlosdauden: Since it is modified, this is simpler:
```suggestion
        args = args or []
```

carlosdauden (approved):
non-blocking suggestion

CarlosRoca13:
Changed @carlosdauden :smiley:

PR #2207, opened 2022-09-29
[MIG] sale_order_lot_selection: Migration to 15.0

nicomacr:


sale_order_lot_selection/__manifest__.py
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Sale Order Lot Selection",
-    "version": "14.0.1.1.0",
+    "version": "15.0.1.1.0",
```
rousseldenis: Could you reset to 15.0.1.0.0?

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -17,7 +17,3 @@ def product_id_change(self):
         res = super().product_id_change()
         self.lot_id = False
         return res
-
-    @api.onchange("product_id")
```
rousseldenis: Have you replaced by other mechanism ?

rousseldenis (changes requested):

sale_order_lot_selection/models/sale_order_line.py
nicomacr: yes, I've updated the commit with this part.

rousseldenis (approved):
Code review

bodedra (approved):
Code Review :+1:

bodedra:
@pedrobaeza @rousseldenis is it possible to restart runboat?

pedrobaeza:
A new push should be done to trigger it.

bodedra:
@OCA/crm-sales-marketing-maintainers would you please review and help to merge it?

sale_order_lot_selection/tests/test_sale_order_lot_selection.py
```diff
@@ -1,5 +1,6 @@
 # © 2015 Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 import odoo.tests.common as test_common
```
mmequignon, line 6: I think now is a good time for a refactoring here.
The setUp steps are executed for each test, I would change an use setUpClass instead, which is executed once for all tests. If you use TransactionCase, the checkpoint is restored after each test.

There's basically a single test: I would split it. By the way, It looks like to me that
It's ok to test a whole process in 1 test, but it's also good to test individual parts of the code.

Also, I have the feeling that this would fix your issues regarding naming of tests methods.
Their order shouldn't matter. I would drop the numbering.
It's more a matter of finding out what are the common needs for setting up the environment, and throw it up in the `setUpClass` method.

diggy128:
Anyone to merge this?

sale_order_lot_selection/tests/test_sale_order_lot_selection.py
rousseldenis: @nicomacr What do you think of this ? @mmequignon Maybe in another PR ?

rousseldenis:
/ocabot merge patch

I think @nicomacr is no more active following his github profile.

I'll try to improve tests

dreispt:
Superseded by #2932

rousseldenis:
@nicomacr Could you remove en.po ?

rousseldenis:
Closing as supersedes by #3123. User is no more active on github

PR #2219, opened 2022-10-10
[14.0][ADD] sale_outstanding: create add-on

stephansainleger:
Sale_outstanding add-on adds sum_outstanding field to sale.order model, and displays it in the sale order form view.
The field content is the sum of the invoiced amounts linked to the sale order.

**Test scenario:**
- basic test:
  - create a sale order with several products
  - invoice a part of the global amount, open it (invoice_lines must match with order_lines)
  => sum_outstanding field content is the amount invoiced
  - create a new invoice, open it
  => sum_outstanding field content is the sum of the two previous invoices


stephansainleger:
Functionality already covered by add-on `sale_order_invoice_amount`.
I close the PR.

PR #2224, opened 2022-10-11
[15.0][MIG] sale_order_revision: Migration to 15.0

CRogos:


CRogos:
Has anyone an idea why tests from "sale_order_partner_restrict" are executed and fail?

PR #2225, opened 2022-10-11
[15.0][IMP] Update deprecated warnings

CRogos:
There are a lot of warnings in the unit tests.

### DIFF
--- a/sale_order_invoicing_finished_task/tests/test_sale_order_invoicing_finished_task.py
+++ b/sale_order_invoicing_finished_task/tests/test_sale_order_invoicing_finished_task.py
@@ -5,7 +5,7 @@ from odoo.exceptions import ValidationError
 from odoo.tests import common
 
 
-class TestInvoicefinishedTask(common.SavepointCase):
+class TestInvoicefinishedTask(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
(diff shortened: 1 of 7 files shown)


CRogos:
Am I understading the warning correctly that it means remove the return value?

2022-09-30 12:20:45,491 249 WARNING odoo odoo.models: onchange method SaleOrderLine.product_uom_change returned a domain, this is deprecated 
image

like this:
image

MiquelRForgeFlow (approved):

ivantodorovich (approved):

ivantodorovich:
Could you set the PR as "ready to review" ? (It's currently on draft)

CRogos:
@ivantodorovich there are a lot more warnings to cleanup. Do you have an answer on my question above. Can the onchange methods be changed as suggested?

ivantodorovich:
I take this is the faulty line: https://github.com/OCA/sale-workflow/blob/3983568e105f6c281be4a474b92fc4a647f04d4e/sale_blanket_order/models/sale_orders.py#L105

Indeed returning a domain from an onchange method has been deprecated. But the fix is not just about removing that line nor having the onchange return nothing.

The code needs to be refactored to find another alternative to filter the available `blanket_order_line`, that doesn't require returing a domain from an onchange. Possibly using a computed field for the `eligible_bo_lines`, or maybe using the module https://github.com/OCA/web/tree/15.0/web_domain_field

But since I'm not familiar with `sale_blanket_order` module, maybe it should be discussed with its maintainers

MiquelRForgeFlow:
My suggestion is merging this PR as it is now, and then open a new one when someone has a clear idea how to solve the pending warnings.

ivantodorovich:
I agree

sergio-teruel (approved):

PR #2227, opened 2022-10-17
[14.0][ADD] sale_single_attribute_value_display

ilyasProgrammer:
By default attributes with single value are excluded from display name of a product.
This module will include it only for sale order line description.


francesco-ooops (approved):
functional ok!

---
A product with this setup 

image

appears as follows in SO line description: 

image

instead of 

image

francesco-ooops:
@pedrobaeza got an opinion on this one?

sale_single_attribute_value_display/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+Configuration not required.
```
pedrobaeza: Don't add this file if no configuration.

sale_single_attribute_value_display/models/models.py
```diff
@@ -0,0 +1,22 @@
+from odoo import models
```
pedrobaeza: Call the file `product.py` instead.

sale_single_attribute_value_display/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+Install module. Create product with single value attribute. Add it to Sale Order Line.
```
pedrobaeza: Put it as numbered bullet list:

```suggestion
#. Create a product with any single value attribute.
#. Add it to a sales order.
#. ...
```

pedrobaeza (changes requested):
For the module, I would choose `sale_order_line_description_single_attribute`

And a test with a `Form` of the sales order and checking the contents of the description would be very desirable.

ilyasProgrammer:
updated

pedrobaeza (approved):

PR #2229, opened 2022-10-18
[15.0][MIG] sale_credit_point: Migrate to version 15.0

hailangvn:
* From V11.0 https://github.com/OCA/sale-workflow/pull/579

Excerpt from README:

> Sale based on partners' credit point.
> ### Use case
>
> You have employees that have some credit (up to you where this come from)
> and you allow employees to buy products based on this credit.
>
> Products are sold by points rather than money.

### DIFF
--- a/sale_credit_point/models/partner.py
+++ b/sale_credit_point/models/partner.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models, api, exceptions, _
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    credit_point = fields.Monetary(
+        string='Points Attribution',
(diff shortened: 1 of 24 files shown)


mmequignon (approved):
I'm not sure comments from @pedrobaeza on the v11.0 PR are still relevant.
LGTM

sale_credit_point/models/partner.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2018-2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, exceptions, fields, models
+
+POINT_OPERATIONS = [
```
cyrilmanuel: not an expert but to avoid one import, this can be added to the pointHistory model ? 
No blocking comment

cyrilmanuel (approved):
not an expert but LGTM

sale_credit_point/models/partner.py
hailangvn: Thanks for spotting it. The PR has been updated.

PR #2239, opened 2022-10-24
[FIX] sale_order_line_packaging_qty: precision is mandatory for pkg qty compute

i-vyshnevska:
default values in case of removed product

### DIFF
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -35,16 +35,22 @@ class SaleOrderLine(models.Model):
                 )
             else:
                 product_qty = sol.product_uom_qty
+            # Provide default values in case of missing product.
+            # Odoo does not prevent you to modify product packaging
+            # that are already linked to a sale order.
+            # If by chance the product is removed (yes, you can do it)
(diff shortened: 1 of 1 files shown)


sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -35,16 +35,18 @@ def _compute_product_packaging_qty(self):
                 )
             else:
                 product_qty = sol.product_uom_qty
+            # provide default values in case of missing product
```
simahawk:
```suggestion
            # Provide default values in case of missing product.
            # Odoo does not prevent you to modify product packaging
            # that are already linked to a sale order.
            # If by chance the product is removed (yes, you can do it)
            # this computation will be broken w/out this default.
```

simahawk (changes requested):
@i-vyshnevska thanks for the fix. Small remark on the comment and big remark on the commit mgs: please keep it short and meaningful.
Can you rewrite it as: `[fix] s_order_line_pkg_qty: safe default for pkg qty precision`.

simahawk (approved):

dreispt (approved):

PR #1776, opened 2021-11-17
[15.0][MIG] portal_sale_personal_data_only: Migration to 15.0

moaadbourhim:
Migration of portal_sale_personal_data_only to 15.0

Please @pedrobaeza can you review it?

PR #2243, opened 2022-10-26
[15.0][MIG+REF] portal_sale_personal_data_only: Migration to 15.0 and split in two

### LINKED ISSUE #2236: [portal_sale_personal_data_only] Split for sale and invoice
Hello,

Module [portal_sale_personal_data_only](https://github.com/OCA/sale-workflow/tree/14.0/portal_sale_personal_data_only) includes security rules for sales and invoices and it forces to depend in [sales](https://github.com/OCA/sale-workflow/blob/14.0/portal_sale_personal_data_only/__manifest__.py#L10)

Sales: https://github.com/OCA/sale-workflow/blob/14.0/portal_sale_personal_data_only/security/security.xml#L3

Invoices: https://github.com/OCA/sale-workflow/blob/14.0/portal_sale_personal_data_only/security/security.xml#L19

We have the use case with the same problem (IMHO a Odoo BUG...) but client don't use sales, just invoicing.

I guess it's not a problem and in fact make sense to split it.

Do you agree?

cc @pedrobaeza @chienandalu @victoralmau @kirca  (last PRs I saw)

yajo:
Pretty basic standard migration. Includes relevant commits from #1776.

Fix https://github.com/OCA/sale-workflow/issues/2236 by extracting these features into the new module `portal_account_personal_data_only`.

The module now depends on it and is auto-installed if `sale` is also installed.

@moduon MT-1451

Keeping in draft until https://github.com/OCA/account-invoicing/pull/1275 is merged.


portal_sale_personal_data_only/__manifest__.py
```diff
@@ -7,7 +7,8 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "license": "AGPL-3",
-    "depends": ["sale"],
+    "depends": ["sale", "portal_account_personal_data_only"],
```
pedrobaeza, line 10: Is the dependency over `portal_account_personal_data_only` strictly needed?

portal_sale_personal_data_only/readme/DESCRIPTION.rst
```diff
@@ -1,7 +1,7 @@
-By default, portal users are allowed to see all the orders and invoices in
+By default, portal users are allowed to see all the sale orders in
 which a member of their organization are followers. That could cause a leaking
 of  documents between members and departments and of the organization that
 should stay private.
 
-This module restricts that behaviour so the portal users only see their own
+This module restricts that behavior so the portal users only see their own
```
pedrobaeza: I usually prefer British English over American:

https://grammarist.com/spelling/behavior-behaviour/

but I think the main reason to let it as it was is for reducing the diff.

portal_sale_personal_data_only/__manifest__.py
yajo: Yes. Due to the split, these rules had to be split in two, one on each of the modules:

https://github.com/OCA/sale-workflow/blob/53cb00f656baf4ce642d683202d7769e49719d0f/portal_sale_personal_data_only/security/security.xml#L19-L34

Without the dependency, usability would be broken. That's also why this module is marked now as auto install.

portal_sale_personal_data_only/readme/DESCRIPTION.rst
yajo: It was just the code spell checker underlining this on red and I simply changed it. Let me undo the change because it is indeed not so important.

OTOH, AFAIK the base lang for odoo is `en_US`. I've never read a guideline on the subject for OCA, but since we follow Odoo's I always understood that's our standard here too. Does it make sense? We need a standard "behavior" in this subject. 🥁 🙃

portal_sale_personal_data_only/__manifest__.py
pedrobaeza: Yes, but then it's not needed technically if I'm not interpreting it badly. I think the `auto_install` is enough, and now yo can put not only `True` / `False`, but the names of the modules to be installed for this one being auto-installed, and you can include there `portal_account_personal_data_only` instead of a hard dependency.

rafaelbn:
@fcvalgar please could you review? Thank you!

portal_sale_personal_data_only/__manifest__.py
yajo: I see what you mean. I also thought about that. However I'm not sure that removing that dependency is the best idea.

On one side, this is a migration. By adding the dependency, we avoid the need of migration scripts, as this module will behave just as it was before. This is good IMHO as it avoids surprises.

OTOH, according to [the docs](https://www.odoo.com/documentation/15.0/developer/reference/backend/module.html?highlight=auto_install#manifest), I can't do it as you said:

> **If it is a list, it must contain a subset of the dependencies**. This module will automatically be installed as soon as all the dependencies in the subset are installed. The remaining dependencies will be automatically installed as well. If the list is empty, this module will always be automatically installed regardless of its dependencies and these will be installed as well.

If I use a module in `auto_install` that's not among its dependencies, it's not guaranteed to work AFAICS.

So, if I have to choose among removing the dependency or removing the `auto_install` feature, IMHO `auto_install` is more valuable, as it makes the whole experience more coherent. After all, it makes sense that, if you want to restrict sale orders visibility, you want the same on invoices.

In the worst case, this is how it was working before the split, so this is not a regression. If you want I can add the idea to the roadmap.

pedrobaeza: OK, let's keep it as it is.

portal_sale_personal_data_only/readme/DESCRIPTION.rst
pedrobaeza: Yes, this has never been discussed, but indeed the base module is American English. We have similar question with for example colour/color. Not sure if Odoo has gone one way or the other. On the other hand, the British one is supposed to be the "template" (not sure if we can compare it with `Spanish (Spain)`). But as said, this one was already put, and it's on the benefit of the diff, but on homogeneity terms, if we see that better everything in American English, let's change it.

I let it in your hands.

Shide (approved):
Functional review LGTM

fcvalgar (approved):
Functional review LGTM

rafaelbn (commented):
Please @yajo check test! 😄

portal_sale_personal_data_only/readme/DESCRIPTION.rst
yajo: I opened https://github.com/moduon/odoo-community.org/pull/2 as I'm in favor of standardization. However I'll leave it as it is as I'm also in favor of MVP. 😄

yajo:
Tests fixed. Please somebody merge? Thanks!

chienandalu (approved):
Thanks! :slightly_smiling_face: :+1:

PR #2246, opened 2022-10-27
[14.0][IMP] sale_restricted_qty: show sale_min sale_max optionally

ilyasProgrammer:
… in SOL

### DIFF
--- a/sale_restricted_qty/views/sale_views.xml
+++ b/sale_restricted_qty/views/sale_views.xml
@@ -62,10 +62,10 @@
                 expr="//field[@name='order_line']/tree/field[@name='product_uom_qty']"
                 position="after"
             >
-               <field name="sale_min_qty" />
+               <field name="sale_min_qty" optional="hide" />
                <field name="is_qty_less_min_qty" invisible="1" />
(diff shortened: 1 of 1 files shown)


SicurSam (approved):
Functional ok!

francesco-ooops (approved):
minor change

francesco-ooops:
@OCA/crm-sales-marketing-maintainers good change?

francesco-ooops:

simahawk:
@ilyasProgrammer can you fix the commit msg which is cut? (See how I rephrased it in the title)
Wondering: the initial intent was to show it by default: are we sure we shouldn't use `show` by default instead?

ilyasProgrammer:
> @ilyasProgrammer can you fix the commit msg which is cut? (See how I rephrased it in the title) Wondering: the initial intent was to show it by default: are we sure we shouldn't use `show` by default instead?
As for me there is no difference.

ilyasProgrammer:
updated

francesco-ooops:
@simahawk is it ok now?

simahawk (approved):
If nobody minds about the default visibility... ok :)

PR #2251, opened 2022-10-29
[IMP] sale_automatic_workflow: pay with same currency as invoice

hoangtrann:
There is an instance where the invoice is using a currency but payment is registered as another currency, due to undetermined journal. This leads to invoice being partially paid due to exchance rate while converting from a currency to another.


hoangtrann:
@simahawk could you have a look?

rousseldenis (approved):

rousseldenis:
@simahawk

PR #2256, opened 2022-11-09
[15.0][FIX] tests run twice

CRogos:


pedrobaeza:
This is done on purpose, for testing things both on OCB and on vanilla Odoo.

CRogos:
yes, but it should not be run twice each, or?
image

rousseldenis:
@CRogos You have an excluded module. So, that's normal to have two flows (one with module, one without) and for each flow, we test Odoo and OCB versions.

CRogos:
ah ok... now I see it... thank you for the explanation.

PR #2261, opened 2022-11-14
New OCA module: allow different account for discount

imlopes:
This module allows configuring a specific account for the accounting of
discounts, which can be required in some localization (or by some accountants,
especialy in the e-commerce sector).


sale_order_discount_invoicing/models/account_move.py
```diff
@@ -0,0 +1,154 @@
...
+        return discount_line
+
+    def _process_discount_line_vals(self, invoice_line_vals):
+        """if the invoice line has a slashed price, split it in two, one for the full
+        price and one for the discount. Otherwise keep it"""
+        if "origin_price_unit" in invoice_line_vals:
+            origin_price_unit = invoice_line_vals.pop("origin_price_unit")
+            if not self._can_process_discount_line_vals(invoice_line_vals):
+                yield invoice_line_vals
+            else:
```
Camille0907: From what I understand, the test `_can_process_discount_line_vals` could be at the beginning of the function : 
```suggestion
            if not self._can_process_discount_line_vals(invoice_line_vals):
                yield invoice_line_vals
            else:
                if "origin_price_unit" in invoice_line_vals:
                    origin_price_unit = invoice_line_vals.pop("origin_price_unit")
```

And if you do that, you may not need to add the function `_can_process_discount_line_vals` here for the specific later, you could directly remove the test here : 
```python
         if "origin_price_unit" in invoice_line_vals:
                    origin_price_unit = invoice_line_vals.pop("origin_price_unit")
                    price_unit = invoice_line_vals["price_unit"]
                    ...
```

And then in your specific, make : 
```python
def _process_discount_line_vals(self, invoice_line_vals):
        if not self._can_process_discount_line_vals(invoice_line_vals):
                yield invoice_line_vals
        else:
                return super()._process_discount_line_vals
```

sale_order_discount_invoicing/tests/test_sale_discount.py
```diff
@@ -0,0 +1,405 @@
...
+            }
+        )
+        cls.sale_order2 = cls.env["sale.order"].create(
+            {
+                "partner_id": cls.partner1.id,
+                "partner_invoice_id": cls.partner1.id,
+                "partner_shipping_id": cls.partner1.id,
+            }
+        )
+        # cls.product_order_tax_excluded.taxes_id = cls.tax_price_exclude.id
```
Camille0907: Why is this commented ?

Camille0907 (approved):
Two non-blocking suggestions
But LGTM from what I understand, pre-approved : check red tests

Camille0907 (approved):

gurneyalex (approved):

PR #2262, opened 2022-11-14
[14.0][ADD] sale_stock_line_customer_ref

sebalix:
This module allows you to add a customer reference on a sale order line.
Lines sharing the same product but different customer references won't be merged together in stock transfers.

TODO:
- [ ] package.customer_ref tests
- [x] package.customer_ref add to the view


sale_stock_line_customer_ref/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Stock Line Customer Reference",
+    "summary": (
+        "Allow you to add a customer reference on order lines "
+        "propataged to move operations."
```
simahawk:
```suggestion
        "propagated to move operations."
```

sale_stock_line_customer_ref/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    customer_ref = fields.Char(
+        compute="_compute_customer_ref", readonly=True, store=True
+    )
+
+    @api.depends("sale_line_id.customer_ref", "move_dest_ids")
```
simahawk:
```suggestion
    @api.depends("sale_line_id.customer_ref", "move_dest_ids.sale_line_id")
```

sale_stock_line_customer_ref/tests/test_customer_ref.py
```diff
@@ -0,0 +1,45 @@
...
+                with form.order_line.new() as line:
+                    line.product_id = cls.env.ref("product.consu_delivery_01")
+                    line.customer_ref = f"TEST_{i}"
+            order = form.save()
+            return order
+
+    def test_customer_ref(self):
+        # Ensure we have two moves that didn't get merged despite the identical
+        # product as their customer reference is different.
+        self.assertEqual(len(self.order.order_line.move_ids), 2)
```
simahawk: I would add more than 1 record per case

simahawk (changes requested):
tests are :red_circle: 

```
 2022-11-14 17:34:35,718 249 ERROR odoo unittest.suite: ERROR: setUpClass (odoo.addons.sale_stock_line_customer_ref.tests.test_customer_ref.TestSaleLineCustomerRef)
  File "/opt/odoo/odoo/tests/common.py", line 173, in _handleClassSetUp
    setUpClass()
  File "/__w/sale-workflow/sale-workflow/sale_stock_line_customer_ref/tests/test_customer_ref.py", line 13, in setUpClass
    cls.order = cls._create_order()
  File "/__w/sale-workflow/sale-workflow/sale_stock_line_customer_ref/tests/test_customer_ref.py", line 25, in _create_order
    line.customer_ref = f"TEST_{i}"
  File "/opt/odoo/odoo/tests/common.py", line 1857, in __setattr__
    assert descr is not None, "%s was not found in the view" % field
AssertionError: customer_ref was not found in the view
```

sale_stock_line_customer_ref/tests/test_customer_ref.py
sebalix: You mean, for the same product, put some SO lines with the same customer ref, and others with a different one? I'll do that.

sale_stock_line_customer_ref/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    customer_ref = fields.Char(
```
simahawk: Wondering: shall we add related non-stored field on the move line?
IMO it could help in many cases. No?

sebalix: I was wondering too, put a related on `stock.move.line` to know what we are referring to.

sebalix:
I can reproduce the test issue locally, there is incompatibility with `sale_rental` module, checking why.
EDIT: The `customer_ref` field was simply missing on the form view of SO line.

sale_stock_line_customer_ref/tests/test_customer_ref.py
sebalix: I pushed a fixup that extend the tests and add some comments in the code as I wrote this initially without considering some obvious behavior of moves merging.

sale_stock_line_customer_ref/models/stock_package_level.py
```diff
@@ -0,0 +1,20 @@
...
+
+    has_customer_ref = fields.Boolean(
+        compute="_compute_has_customer_ref",
+        help="Technical field to display 'Customer Ref' column in a package level.",
+    )
+
+    @api.depends("move_ids.customer_ref")
+    def _compute_has_customer_ref(self):
+        for pl in self:
+            pl.has_customer_ref = any(pl.move_ids.mapped("customer_ref")) or any(
```
simahawk: This is going to loop on all records every time.
In such cases I normally prefer a loop + break on the 1st one matching.
Here's a shortcut:
```
next((x for x in pl.move_ids if x.customer_ref), False)
```
See https://stackoverflow.com/questions/2361426/get-the-first-item-from-an-iterable-that-matches-a-condition

sale_stock_line_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+
+    has_customer_ref = fields.Boolean(
+        compute="_compute_has_customer_ref",
+        help="Technical field to display 'Customer Ref' column on moves.",
+    )
+
+    @api.depends("move_lines.customer_ref")
+    def _compute_has_customer_ref(self):
+        for picking in self:
+            picking.has_customer_ref = any(
```
simahawk: same

simahawk (approved):
LG, just a nit-picking perf remark

sale_stock_line_customer_ref/models/stock_package_level.py
sebalix: Indeed, knowing that we are just saving CPU time here, we still fetch the same amount of data (IO) even if we break on the first occurrence, but it's better than nothing.

sebalix: Small test added regarding `picking.has_customer_ref`, a bit lazy to add one for the package level... :)

sebalix:
Squashed

sale_stock_line_customer_ref/models/stock_move.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    sale_line_with_customer_ref_id = fields.Many2one(
```
simahawk:
```suggestion
    customer_ref_sale_line_id = fields.Many2one(
```

simahawk (changes requested):
no need for a post init hook to initialize default column + existing records?

sale_stock_line_customer_ref/models/stock_move.py
sebalix: Done, and:
- rewrite the recursive function to not be recursive anymore using a while loop.
- add a `pre_init_hook` script to avoid the recomputation of all moves (useless, and avoid a MemoryError on big DB)

sebalix:
Squashed

simahawk:
Added TODO to PR description

sale_stock_line_customer_ref/tests/test_customer_ref.py
```diff
@@ -10,31 +10,59 @@ def setUpClass(cls):
...
+            lambda x: not x.move_lines.move_orig_ids
+        )
+
+    @classmethod
+    def _get_pack(cls, order):
+        return cls._get_pick(order).move_lines.move_dest_ids.picking_id
+
     def test_customer_ref(self):
-        # Ship moves can't be merged anyway in std Odoo as they belong to
+        # Ship moves can"t be merged anyway in std Odoo as they belong to
```
sebalix: Typo to revert

sale_stock_line_customer_ref/models/sale_order_line.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    customer_ref = fields.Char(string="Customer Ref.")
```
ivantodorovich, line 10: For [consistency with the core field in `sale.order`](https://github.com/odoo/odoo/blob/2cd813c5911c4eb55473e39fa071370c2148143c/addons/sale/models/sale.py#L161), maybe consider naming it `client_order_ref`

ivantodorovich (approved):

sale_stock_line_customer_ref/models/sale_order_line.py
sebalix: I prefer to keep `customer_ref` :thinking: and not overlapping the `client_order_ref` std field.

ivantodorovich: Well it wouldn't overlap as one is for sale.order and the new one is for the lines. As the purpose seems to be the same, maybe the name could be the same.
But anyway, not blocking 😉

simahawk (approved):

PR #2266, opened 2022-11-17
[MIG] sale_exception: Migration to 16.0

matiasperalta1:


sale_exception/models/sale.py
```diff
@@ -0,0 +1,89 @@
...
+        return ["ignore_exception", "order_line", "state"]
+
+    def _check_sale_check_exception(self, vals):
+        check_exceptions = any(
+            field in vals for field in self._fields_trigger_check_exception()
+        )
+        if check_exceptions:
+            self.sale_check_exception()
+
+    @api.model
```
ghost:
```suggestion
    @api.model_create_multi
```

Keep in mind that you should also adapt the method _check_sale_check_exception. It will receive a list of dict from the create method and only an dict from the write method.

ghost (changes requested):

aliciagaarzo (approved):

aliciagaarzo (commented):
After testing the migrated module, I’ve found some improvements:
- Description field is noy properly displayed; user needs to scroll to the right to see it.
- Allow to ignore exceptions even if the “is blocking” option is checked

sale_exception/models/sale.py
```diff
@@ -53,11 +53,12 @@ def _check_sale_check_exception(self, vals):
         if check_exceptions:
             self.sale_check_exception()
 
-    @api.model
```
sebastienbeau: Hi thanks for the migration.

It's not related to your change but after think twice
I think that the check on create is useless !

Indeed if you look at the method sale_check_exception, it explicitly skip all sale order that are not in the state "sale", when we create a sale order the state is always "draft".

So indeed of adding some extra complexity to the code, maybe it's a good opportunity to simplify it and just to remove the inherit on "def create".

I should have seen it before but I miss to review this PR : https://github.com/OCA/sale-workflow/pull/916

@simahawk @rousseldenis @sbejaoui what do you think ?

sebastienbeau: @florian-dacosta

sebastienbeau:
also, can you rebase as base_exception have been merged ?

sebastienbeau:
Note: I also realize that the migration in V9 have drop the commit historic (it was correct in 8 see here: https://github.com/OCA/sale-workflow/commits/8.0?after=605d0ae2da6d37e9acd8466eb77d2cc023d2f289+69&branch=8.0&path%5B%5D=sale_exceptions&qualified_name=refs%2Fheads%2F8.0)

When this PR will be really, I will, before merging the PR, restore the commit history, so please ping me when it's ready, do not waist your time with this point, I will do it.

sale_exception/models/sale.py
florian-dacosta: I agree the create override can be removed as it seems useless.

rousseldenis:
> Note: I also realize that the migration in V9 have drop the commit historic (it was correct in 8 see here: https://github.com/OCA/sale-workflow/commits/8.0?after=605d0ae2da6d37e9acd8466eb77d2cc023d2f289+69&branch=8.0&path%5B%5D=sale_exceptions&qualified_name=refs%2Fheads%2F8.0)


@sebastienbeau That's normal as module has been renamed at that time.

sale_exception/models/sale.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2011 Akretion, Sodexis
+# Copyright 2018 Akretion
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ExceptionRule(models.Model):
```
rousseldenis, line 9: Please put models in separate files

rousseldenis (approved):
LGTM. Minor change

sale_exception/tests/test_multi_records.py
```diff
@@ -0,0 +1,125 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+import mock
```
rousseldenis, line 3:
```suggestion
from unittest import mock
```

lmignon (changes requested):
LGTM once the requested changes will be implemented

rousseldenis:
@matiasperalta1

lmignon:
@matiasperalta1 Do you plan to work on this PR soon or do you prefer we finalize your work? We need to move forward with this one....

sbejaoui:
I pushed a new PR (#2344) to finalize this one because we need it for one of our projects.

rousseldenis:
@matiasperalta1 Could you close this in favor of #2344 ?

PR #2267, opened 2022-11-17
[FIX] Fix use of next_by_code only when name is Draft

hurrinico:
If you bring back a order to Draft state, when is confirmed, it changes is sequence
PR in order to fix https://github.com/OCA/sale-workflow/issues/2268

### DIFF
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -301,7 +301,8 @@ class BlanketOrder(models.Model):
             sequence_obj = self.env["ir.sequence"]
             if order.company_id:
                 sequence_obj = sequence_obj.with_company(order.company_id.id)
-            name = sequence_obj.next_by_code("sale.blanket.order")
+            if order.name == "Draft":
+                name = sequence_obj.next_by_code("sale.blanket.order")
             order.write({"confirmed": True, "name": name})
         return True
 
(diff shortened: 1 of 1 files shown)


sale_blanket_order/models/blanket_orders.py
```diff
@@ -301,7 +301,10 @@ def action_confirm(self):
             sequence_obj = self.env["ir.sequence"]
             if order.company_id:
                 sequence_obj = sequence_obj.with_company(order.company_id.id)
-            name = sequence_obj.next_by_code("sale.blanket.order")
+            if order.name == "Draft":
+                name = sequence_obj.next_by_code("sale.blanket.order")
+            else:
+                name = order.name
             order.write({"confirmed": True, "name": name})
```
CRogos, line 308: It is only a small issue, but wouldn't it be better not to write the name at all, if it does not change?

vals = {"confirmed": True}
if order.name == "Draft":
    vals["name"] = sequence_obj.next_by_code("sale.blanket.order")

order.write(vals)

rousseldenis:
@hurrinico Could you confirm the current code is working ?

https://github.com/OCA/sale-workflow/blob/64bc64cfc9706e0ae477caa67602db04c0189084/sale_blanket_order/models/blanket_orders.py#L325C9-L325C25

hurrinico:
> @hurrinico Could you confirm the current code is working ?
> 
> https://github.com/OCA/sale-workflow/blob/64bc64cfc9706e0ae477caa67602db04c0189084/sale_blanket_order/models/blanket_orders.py#L325C9-L325C25

This code is from 2 years ago, I've changed company from that date but looking at the code i guess that it works

rousseldenis:
Ok, closing

rousseldenis:
@hurrinico I do a little bit household :sweat_smile:

PR #2086, opened 2022-06-20
[15.0][MIG] sale_fixed_discount

xhnilic3:
PR in order to migrate sale_fixed_discount module to version 15.0

PR #1601, opened 2021-05-28
[14.0] [MIG] sale_fixed_discount

khushbu-3:
Migration sale_fixed_discount module from 13.0 to 14.0 version

PR #2269, opened 2022-11-20
[15.0][MIG] sale_fixed_discount

Rad0van:
As #1601 and #2086 are abandoned and non-functional, here's another attempt. Would be fairly standard but when using in real life it was found that the new json-style summary shows incorrect values. Hence the ugly hack at the end of the file trying to make it work. Nicer way would be to override the inner `compute_taxes` method but I am not sure how to do that. Any advice is welcome.


Rad0van:
@LoisRForgeFlow @JordiBForgeFlow @MiquelRForgeFlow @HviorForgeFlow would you please care to have a look at the inner method thing? I am sure there is a better way...

ao-landoo:
@Rad0van check https://github.com/odoo/odoo/commit/174d52a8283d306169c72fef8390cc397675d5d0 
You need to modify xpaths as they're no longer valid

Rad0van:
Closing as #2485 moved forward better than me. Thanx @ps-tubtim

PR #2274, opened 2022-11-22
update repo from copier

AnizR:


adrienpeiffer (approved):

AnizR:
I don't know if "product_set_line" was used but it wasn't in the data of  sale_product_set_sale_by_packaging so, I removed it.

PR #2277, opened 2022-11-23
[13.0][ADD] sale_fully_delivered_invoiced

AaronHForgeFlow:
The standard field invoice_status shows information about the status invoices related
as closed or not. Because of this, this module introduce two new fields that  will show
account.

cc @ForgeFlow


AaronHForgeFlow:
In v15.0 this module is divided into two now. Closing, but please do not delete branch

PR #2279, opened 2022-11-24
[15.0][MIG] sale_order_general_discount: Migration to version 15.0

stefan-tecnativa:
CC @Tecnativa TT40605

@pedrobaeza @chienandalu Please review


sale_order_general_discount/views/sale_order_view.xml
```diff
@@ -0,0 +1,26 @@
...
+                        <tr>
+                            <td class="o_td_label">
+                                <label for="general_discount" />
+                            </td>
+                            <td>
+                                <field name="general_discount" nolabel="1" />
+                            </td>
+                        </tr>
+                    </tbody>
+                </table>
```
chienandalu: Testing in runbut I think this would be enough:

```suggestion
                <field name="general_discount" class="oe_subtotal_footer" />
```

Result:

image

sale_order_general_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+                vals["discount"] = sale_order.general_discount
+        return super().create(vals)
+
+    @api.depends("order_id", "order_id.general_discount")
+    def _compute_discount(self):
+        if hasattr(super(), "_compute_discount"):
+            super()._compute_discount()
+        for line in self:
+            line.discount = line.order_id.general_discount
+        return True
```
chienandalu: You don't really need to return anything
```suggestion
        return
```

sale_order_general_discount/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+The discount is not aligned with the totalization fields because they are added with a widget.
+For this to be aligned, it would be necessary to investigate how the widget works, which would
+complicate the migration of the module more than it should.
```
chienandalu: With my suggestion I'd say this warning can go away

sale_order_general_discount/readme/CONTRIBUTORS.rst
```diff
@@ -1,3 +1,4 @@
 * Sergio Teruel <sergio.teruel@tecnativa.com>
 * Raf Ven <raf.ven@dynapps.be>
 * Sudhir Arya <sudhir@erpharbor.com>
+* Stefan Ungureanu <stefan.ungureanu@tecnativa.com>
```
chienandalu: Can you change our attributions in our current format? Example:

https://github.com/OCA/rma/blame/14.0/rma/readme/CONTRIBUTORS.rst#L1-L5

chienandalu (changes requested):
Thanks @stefan-tecnativa ! Here you have some suggestions:

sale_order_general_discount/models/sale_order_line.py
stefan-tecnativa: Thanks for the observation.

sale_order_general_discount/views/sale_order_view.xml
stefan-tecnativa: It looks fine, thanks.

sale_order_general_discount/readme/ROADMAP.rst
stefan-tecnativa: I am agree with you.

sale_order_general_discount/readme/CONTRIBUTORS.rst
stefan-tecnativa: Let me know if I've done it right

chienandalu (approved):
Great! Thanks!

sale_order_general_discount/views/sale_order_view.xml
```diff
@@ -7,8 +7,8 @@
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='amount_tax']" position="after">
-                <field name="general_discount" />
+            <xpath expr="//field[@name='tax_totals_json']" position="before">
+                <field name="general_discount" class="oe_subtotal_footer" />
```
pedrobaeza, line 11: Why is this put on the footer instead on the header part?

stefan-tecnativa: From version 13 to 15, this field has changed the way it stores and processes information. Because of this, the field we want to enter is not quite aligned, and as studying the behavior of the widget would be too much work for the migration of the module, we decided that this was the best way to display the field.

chienandalu: > Why is this put on the footer instead on the header part?

It's the original design and the migration doesn't change that.

pedrobaeza: OK, it seems a bit weird that something changed on the footer changes the lines, but it seems this way since at least 13.0.

pedrobaeza (approved):
/ocabot merge nobump

PR #2280, opened 2022-11-24
[16.0][MIG] sale_delivery_state

atchuthan:
cc: @SodexisTeam


atchuthan:
@OCA/community-maintainers Test fails due to mock python package missing at https://github.com/OCA/sale-workflow/actions/runs/3540792580/jobs/5944271297

```
  File "/__w/sale-workflow/sale-workflow/sale_delivery_state/tests/__init__.py", line 1, in <module>
    from . import test_delivery_state
  File "/__w/sale-workflow/sale-workflow/sale_delivery_state/tests/test_delivery_state.py", line 6, in <module>
    import mock
ModuleNotFoundError: No module named 'mock'
```

sale_delivery_state/tests/test_delivery_state.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2018 Akretion (http://www.akretion.com).
+# @author Benoît GUILLOT <benoit.guillot@akretion.com>
+# Copyright 2018 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import mock
```
rousseldenis:
```suggestion
from unittest import mock
```
@atchuthan

atchuthan: @rousseldenis done changes

sale_delivery_state/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+            ("done", "Done"),
+        ],
+        # Compute method have a different name then the field because
+        # the method _compute_delivery_state already exist to compute
+        # the field delivery_set
+        compute="_compute_sale_delivery_state",
+        store=True,
+    )
+
+    force_delivery_state = fields.Boolean(
```
rousseldenis, line 27: @PierrickBrun @bguillot @atchuthan Shouldn't `delivery_state' converted in computed readonly False field in order to remove force_delivery_state field ?

atchuthan: @rousseldenis force_delivery_state is used by Backend user to force the delivery_state to "Done" by button. And, if the user wants to revert the change, then he uses the unforce button such that it will have the info based on the Calculated field.

If we remove this field(force_delivery_state), the computation might be re-triggered for the delivery_state field based on list of dependent.

rousseldenis: @atchuthan Nope.

Making the delivery_state field readonly False, will allow user to change manually the delivery state. The compute won't trigger after that only if one field in the depends is changed.

But maybe this is something unwanted as we really want to force it (even if qty_delivered on lines change)

atchuthan: @rousseldenis We removed force_delivery_state field as suggested.  Please review.

dreispt (approved):

sale_delivery_state/models/sale_order.py
RodrigoBM: @rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state? 

I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed. 

But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable.

RodrigoBM:
@rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state?

I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed.

But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable. And the name of the function action_unforce_delivery_state should be changed to for example action_recalculate_delivery_state

rousseldenis:
> @rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state?
> 
> I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed.
> 
> But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable. And the name of the function action_unforce_delivery_state should be changed to for example action_recalculate_delivery_state

As I said:

> But maybe this is something unwanted as we really want to force it (even if qty_delivered on lines change)

So, maybe it's a good idea to keep the force field with current changes too.

RodrigoBM:
I think that both solutions can be correct

> > @rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state?
> > I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed.
> > But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable. And the name of the function action_unforce_delivery_state should be changed to for example action_recalculate_delivery_state
> 
> As I said:
> 
> > But maybe this is something unwanted as we really want to force it (even if qty_delivered on lines change)
> 
> So, maybe it's a good idea to keep the force field with current changes too.

But considering that this is a migration issue, we should keep the development logic. What do you think? @atchuthan @dreispt

atchuthan:
> I think that both solutions can be correct
> 
> > > @rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state?
> > > I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed.
> > > But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable. And the name of the function action_unforce_delivery_state should be changed to for example action_recalculate_delivery_state
> > 
> > 
> > As I said:
> > > But maybe this is something unwanted as we really want to force it (even if qty_delivered on lines change)
> > 
> > 
> > So, maybe it's a good idea to keep the force field with current changes too.
> 
> But considering that this is a migration issue, we should keep the development logic. What do you think? @atchuthan @dreispt

so, do we revert the code to use force field as it was in v15? @rousseldenis @RodrigoBM

RodrigoBM:
> > I think that both solutions can be correct
> > > > @rousseldenis you think it is a good idea to remove the logic associated with force_delivery_state?
> > > > I think that if I have decided to set the status of a sale to delivered as done manually, this status should not be recalculated if the quantity of any line is changed.
> > > > But if this change is correct, the action_unforce_delivery_state method is not needed, as the delivery_state field is now selectable. And the name of the function action_unforce_delivery_state should be changed to for example action_recalculate_delivery_state
> > > 
> > > 
> > > As I said:
> > > > But maybe this is something unwanted as we really want to force it (even if qty_delivered on lines change)
> > > 
> > > 
> > > So, maybe it's a good idea to keep the force field with current changes too.
> > 
> > 
> > But considering that this is a migration issue, we should keep the development logic. What do you think? @atchuthan @dreispt
> 
> so, do we revert the code to use force field as it was in v15? @rousseldenis @RodrigoBM

In my opinion, I think so.

RodrigoBM:
@atchuthan, can you revert the last change?

atchuthan:
> @atchuthan, can you revert the last change?

@RodrigoBM reverted.  Please review

RodrigoBM (approved):

RodrigoBM (commented):
Checking version 16, in the base module sale_stock it adds the [delivery_status](https://github.com/OCA/OCB/blob/16.0/addons/sale_stock/models/sale_order.py#L34) field which is the same logic as the delivery_state field in this module.
I think this module is no more necessary. And create a new module called sale_force_delivery for example, to add the force status logic for the new delivery_status base field.
cc: @atchuthan @rousseldenis @dreispt

RodrigoBM (changes requested):
https://github.com/OCA/sale-workflow/pull/2280#pullrequestreview-1240533900

atchuthan:
> Checking version 16, in the base module sale_stock it adds the [delivery_status](https://github.com/OCA/OCB/blob/16.0/addons/sale_stock/models/sale_order.py#L34) field which is the same logic as the delivery_state field in this module. I think this module is no more necessary. And create a new module called sale_force_delivery for example, to add the force status logic for the new delivery_status base field. cc: @atchuthan @rousseldenis @dreispt

I am okay with removing this module in favor of odoo's already field, and we will remove the dependency to sale_automatic_workflow too.  Please confirm whether I can proceed or not?

atchuthan:
Closing this PR as this module is not needed for v16

PR #2284, opened 2022-11-25
[IMP] sale_by_packaging: restrict packaging removing

i-vyshnevska:


sale_by_packaging/models/product_packaging.py
```diff
@@ -9,6 +9,7 @@ class ProductPackaging(models.Model):
     can_be_sold = fields.Boolean(
         string="Can be sold", compute="_compute_can_be_sold", readonly=False, store=True
     )
+    active = fields.Boolean(default=True)
```
simahawk, line 12: why do we need the active flag?

i-vyshnevska: because the archive option is not available for packaging by default, and for some cases customer floated with obsolete packages and the only way to rid of them is to remove these packages, and ofcourse they were used in SO and missing data later, IMO this should be in core

simahawk: ok, then pls split this change in its own commit and add the ribbon on the view

i-vyshnevska: done

simahawk (approved):

dreispt (approved):

PR #2286, opened 2022-11-29
[15.0][IMP] sale_order_product_recommendation: Change context value to avoid warning logs

victoralmau:
Change context value to avoid warning logs

`unsupported operand type(s) for "==": 'res.partner()' == '230'`

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -233,7 +233,7 @@ class SaleOrderRecommendationLine(models.TransientModel):
         for one in self:
             if price_origin == "pricelist":
                 one.price_unit = one.product_id.with_context(
-                    partner=one.partner_id.id,
+                    partner=one.partner_id,
                     pricelist=one.pricelist_id.id,
                     quantity=one.units_included,
                 ).price


pedrobaeza:
Can you point to the code that emits such error?

victoralmau:
If we run tests, we get this log from https://github.com/OCA/sale-workflow/blob/15.0/sale_order_product_recommendation/tests/test_recommendation.py#L121

pedrobaeza:
Sorry, I wasn't clear. I mean the Odoo code computing the price that raises this problem to check if `partner` is expected to be a recordset in all places.

victoralmau:
Ok sorry.
An example of use is in https://github.com/odoo/odoo/blob/15.0/addons/sale_management/models/sale_order.py#L227 which calls https://github.com/odoo/odoo/blob/15.0/addons/product/models/product.py#L259.

pedrobaeza (approved):
Wow, there's a total fuzz in the successive calls, contexts, etc for getting where `partner` is used, but it seems OK judging the sales order example.

victoralmau:
Ping @CarlosRoca13

CarlosRoca13 (approved):

PR #2287, opened 2022-11-29
[15.0][IMP] sale_payment_sheet: Remove redundant default on related field.

victoralmau:
Remove redundant default on related field.

`Redundant default on sale.payment.sheet.company_id `

@Tecnativa


pedrobaeza (approved):
Fast-tracking this straight-forward patch:

/ocabot merge patch

PR #2289, opened 2022-11-29
[IMP] sale_order_line_date: update existing moves date_deadline

ferran-S73:
Update existing moves `date_deadline` when writing on sol's `commitment_date`. This is useful when the `sale.order` is already confirmed.

### DIFF
--- a/sale_order_line_date/models/sale_order_line.py
+++ b/sale_order_line_date/models/sale_order_line.py
@@ -28,3 +28,16 @@ class SaleOrderLine(models.Model):
                 }
             )
         return vals
+
+    def write(self, vals):
+        res = super().write(vals)
+        moves_to_upd = set()
+        if "commitment_date" in vals:
+            for move in self.move_ids:
+                if move.state not in ["cancel", "done"]:
+                    moves_to_upd.add(move.id)
+        if moves_to_upd:
+            self.env["stock.move"].browse(moves_to_upd).write(
+                {"date_deadline": vals.get("commitment_date")}
+            )
+        return res
--- a/sale_order_line_date/tests/test_sale_order_line_date.py
+++ b/sale_order_line_date/tests/test_sale_order_line_date.py
@@ -154,3 +154,8 @@ class TestSaleOrderLineDates(TransactionCase):
         self._assert_equal_dates(
             self.sale_line3.commitment_date, self.sale_line3.move_ids.date_deadline
         )
+        # Test line date change after confirmation
+        self.sale_line1.write({"commitment_date": self.dt2})
+        self._assert_equal_dates(
+            self.sale_line1.commitment_date, self.sale_line1.move_ids.date_deadline
+        )


sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,15 @@ def _prepare_procurement_values(self, group_id=False):
...
         return vals
+
+    def write(self, vals):
+        res = super(SaleOrderLine, self).write(vals)
+        moves_to_upd = self.env["stock.move"]
+        if "commitment_date" in vals:
+            for line in self:
+                for move in line.move_ids:
+                    if move.state not in ["cancel", "done"]:
+                        moves_to_upd += move
```
hildickethan:
```suggestion
            for move in self.move_ids:
                if move.state not in ["cancel", "done"]:
                    moves_to_upd += move
```

ferran-S73: done @hildickethan-S73

hildickethan (approved):

sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,14 @@ def _prepare_procurement_values(self, group_id=False):
                 }
             )
         return vals
+
+    def write(self, vals):
+        res = super(SaleOrderLine, self).write(vals)
```
rousseldenis:
```suggestion
        res = super().write(vals)
```

sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,14 @@ def _prepare_procurement_values(self, group_id=False):
...
+        moves_to_upd = self.env["stock.move"]
```
rousseldenis:
```suggestion
        moves_to_upd = set()
```

sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,14 @@ def _prepare_procurement_values(self, group_id=False):
...
+        if "commitment_date" in vals:
+            for move in self.move_ids:
+                if move.state not in ["cancel", "done"]:
+                    moves_to_upd += move
```
rousseldenis:
```suggestion
                    moves_to_upd.append(move.id)
```

sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,14 @@ def _prepare_procurement_values(self, group_id=False):
...
+        if moves_to_upd:
+            moves_to_upd.write({"date_deadline": vals.get("commitment_date")})
```
rousseldenis:
```suggestion
            self.env["stock.move"].browse(moves_to_upd).write({"date_deadline": vals.get("commitment_date")})
```

rousseldenis (changes requested):
Code review.

Could you add a test ?

ferran-S73:
Hi @rousseldenis changes made and test added

rousseldenis (approved):
Code review

@AaronHForgeFlow @nikul-serpentcs @SerpentCS Want to have your functional advice here

sale_order_line_date/models/sale_order_line.py
```diff
@@ -28,3 +28,16 @@ def _prepare_procurement_values(self, group_id=False):
...
+        moves_to_upd = set()
+        if "commitment_date" in vals:
+            for move in self.move_ids:
+                if move.state not in ["cancel", "done"]:
+                    moves_to_upd.add(move.id)
+        if moves_to_upd:
+            self.env["stock.move"].browse(moves_to_upd).write(
+                {"date_deadline": vals.get("commitment_date")}
+            )
+        return res
```
nikul-serpentcs, line 43: Here we improve code like this.
```
    def write(self, vals):
        res = super().write(vals)
        if "commitment_date" in vals and vals.get('commitment_date', False):
            moves_to_upd = self.move_ids.filtered(lambda m: m.state not in ["cancel", "done"])
            if moves_to_upd:
                moves_to_upd.write(
                    {"date_deadline": vals.get("commitment_date")}
                )
        return res
```

ferran-S73: Hi, thanks for the suggestion. We've actually tested code like this and the use of `self.move_ids.filtered()` iterates twice the moves whereas my original code does it once. So I think it's better for performance to leave it as it is.

AaronHForgeFlow (approved):
LGTM

ferran-S73:
@rousseldenis could you please merge this pr

sale_order_line_date/models/sale_order_line.py
rousseldenis: I +1 on this

PR #2290, opened 2022-12-01
[14.0][IMP] sale_order_type: Misc changes in tests

victoralmau:
Changes done:
- [x] Change `with_context(force_company=...)` to `with_company(..)`
- [x] Add post_install tag in tests to prevent test errors

@Tecnativa

### DIFF
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -6,9 +6,10 @@ from freezegun import freeze_time
 
 import odoo.tests.common as common
 from odoo import fields
-from odoo.tests import Form
+from odoo.tests import Form, tagged
 
 
+@tagged("post_install", "-at_install")
 class TestSaleOrderType(common.TransactionCase):
     def setUp(self):
         super(TestSaleOrderType, self).setUp()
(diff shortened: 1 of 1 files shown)


sale_order_type/tests/test_sale_order_type.py
```diff
@@ -19,9 +19,22 @@ def setUp(self):
         )
         self.account_model = self.env["account.account"]
         self.user_type_id = self.env.ref("account.data_account_type_revenue")
-        self.account = self.account_model.create(
-            {"code": "410000", "name": "Income", "user_type_id": self.user_type_id.id}
+        self.account = self.account_model.search(
```
pedrobaeza: This is happening because tests are not post-install. Change them, and don't create ever an account. Just look for an account of the indicated type (no restricting to the code).

victoralmau:
Changes done.

pedrobaeza:
CI is red.

victoralmau:
> CI is red.

Test error is related to https://github.com/OCA/sale-workflow/blob/14.0/sale_quotation_number/models/sale_order.py#L16 and https://github.com/OCA/sale-workflow/blob/14.0/sale_quotation_number/models/res_company.py#L13.

What do you think would be the best correction in this case? IMO `keep_name_so` should be false by default.

pedrobaeza:
But default=True is in theory the conservative one.

victoralmau:
CI is already green.

pedrobaeza (approved):

PR #2291, opened 2022-12-01
[14.0][IMP] sale_restricted_qty: Change dp.get_precision digits

victoralmau:
Change dp.get_precision digits

@Tecnativa


pedrobaeza (approved):
Straight-forward patch, so fast-tracking:

/ocabot merge patch

PR #2295, opened 2022-12-07
[FIX] sale_automatic_workflow: Currency from Invoice

SodexisTeam:
When payment is created, the currency_id is not obtained from invoice instead the value is taken from company default.

### DIFF
--- a/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
@@ -21,7 +21,7 @@ class AutomaticWorkflowJob(models.Model):
         return vals
 
     def _register_payment_invoice(self, invoice):
-        if not invoice.payment_mode_id.fixed_journal_id:
+        if invoice.payment_mode_id and not invoice.payment_mode_id.fixed_journal_id:
             _logger.debug(
                 "Unable to Register Payment for invoice %s: "
                 "Payment mode %s must have fixed journal",
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -129,6 +129,7 @@ class AutomaticWorkflowJob(models.Model):
             "partner_id": invoice.partner_id.id,
             "partner_type": partner_type,
             "date": fields.Date.context_today(self),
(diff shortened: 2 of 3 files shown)


rousseldenis:
@SodexisTeam Thanks for this. Could you add a test for that change ?

atchuthan:
> @SodexisTeam Thanks for this. Could you add a test for that change ?

Sure, we will add it to test file

atchuthan:
> @SodexisTeam Thanks for this. Could you add a test for that change ?

@rousseldenis Testing added.  Ready for review.

atchuthan:
> > @SodexisTeam Thanks for this. Could you add a test for that change ?
> 
> @rousseldenis Testing added. Ready for review.

@rousseldenis waiting for review, thanks

sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
```diff
@@ -21,7 +21,7 @@ def _prepare_dict_account_payment(self, invoice):
         return vals
 
     def _register_payment_invoice(self, invoice):
-        if not invoice.payment_mode_id.fixed_journal_id:
+        if invoice.payment_mode_id and not invoice.payment_mode_id.fixed_journal_id:
```
rousseldenis, line 24: @SodexisTeam Apart this is correct, note that if payment_mode_id is False, ORM will compute result of invoice.payment_mode_id.fixed_journal_id as False too.

So, this change is not necessary

atchuthan: For the condition `not invoice.payment_mode_id.fixed_journal_id`, if there is no payment_mode_id, this condition returns True.  Thus, the payment will not be created for invoices that do not have payment_mode_id selected.

atchuthan: @rousseldenis please review my last comment

atchuthan:
@OCA/crm-sales-marketing-maintainers This PR is ready for review.  Thanks

rousseldenis (approved):
Code review

atchuthan (approved):
LGTM :+1:

amkarthik (approved):

PR #2300, opened 2022-12-13
14.0 imp sale order pricelist tax report

Kev-Roche:
Change t-esc in t-field in invoice report to keep the monetary format (comma, not dot on prices)


Kev-Roche:
cc @bguillot , @sebastienbeau

PR #2301, opened 2022-12-13
[13.0][IMP] product_form_sale_link: fix compatibility with sale_stock and not filter out quotations

JordiMForgeFlow:
Standard module sale_stock replaces the button used in the inherited product template view, which gives an error when trying to install the module (https://github.com/odoo/odoo/blob/1fc4b241ab4cb4753228127a1829033deb16602e/addons/sale_stock/views/stock_views.xml#L78).

This PR also proposes to remove the filtering of the SO Lines based on the state as users may want to also see quotations. To not modify the current behavior a default filter is added.

Similar proposal as in https://github.com/OCA/purchase-workflow/pull/1682

### DIFF
--- a/product_form_sale_link/views/sale_order_line.xml
+++ b/product_form_sale_link/views/sale_order_line.xml
@@ -2,20 +2,36 @@
 <!-- Copyright 2019 ACSONE SA/NV
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
+    <record id="view_sales_order_line_filter" model="ir.ui.view">
+        <field name="model">sale.order.line</field>
+        <field name="name">sale.order.line.select (in product_form_sale_link)</field>
+        <field name="inherit_id" ref="sale.view_sales_order_line_filter" />
+        <field name="arch" type="xml">
+            <filter name="to_invoice" position="before">
+                <filter
+                    string="Quotations"
+                    name="draft"
+                    domain="[('state','in',('draft', 'sent'))]"
+                />
+                <filter
+                    string="Sales Orders"
+                    name="sales"
+                    domain="[('state','in',('sale','done'))]"
+                />
+                <separator />
+            </filter>
+        </field>
+    </record>
     <record id="action_product_template_sale_list" model="ir.actions.act_window">
         <field name="name">Sale Order Lines</field>
         <field name="res_model">sale.order.line</field>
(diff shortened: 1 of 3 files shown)


product_form_sale_link/views/sale_order_line.xml
```diff
@@ -6,16 +6,12 @@
...
+        <field name="domain">[('product_id.product_tmpl_id', 'in', active_ids)]</field>
     </record>
     <record id="action_product_product_sale_list" model="ir.actions.act_window">
         <field name="name">Sale Order Lines</field>
         <field name="res_model">sale.order.line</field>
         <field name="context">{}</field>
-        <field
-            name="domain"
-        >[('state', 'in', ['sale', 'done']), ('product_id', 'in', active_ids)]</field>
+        <field name="domain">[('product_id', 'in', active_ids)]</field>
```
LoisRForgeFlow, line 35: Then, I propose to apply a default filter for status sale

JordiMForgeFlow: Good point, I have added a default filter for it, thanks! :)

LoisRForgeFlow (approved):

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

mariadforgeflow (approved):
LGTM

JoanSForgeFlow (approved):
Technical review, LGTM!

JordiMForgeFlow:
Hi @rousseldenis 

Could I get your input on this one? :)

sale_advance_payment/tests/test_sale_advance_payment.py
```diff
@@ -3,6 +3,7 @@
 
 import json
 
+from odoo import fields
```
rousseldenis, line 6: @JordiMForgeFlow Is it normal to change another module ?

rousseldenis: Ok, for this : https://github.com/OCA/sale-workflow/actions/runs/5189763968/jobs/9355209642#step:8:100

MiquelRForgeFlow:
To forward port.

rousseldenis:
> To forward port.

@MiquelRForgeFlow Done https://github.com/OCA/sale-workflow/pull/2566

PR #2306, opened 2022-12-15
[15.0][MIG]sale_fully_delivered_invoiced

AaronHForgeFlow:
The standard field invoice_status shows information about the status invoices related
as closed or not. Because of this, this module introduce two new fields that will show
account.

cc @ForgeFlow


legalsylvain (commented):
Hi. Thanks for this features. Seems usefull !

Two remarks :
 - minor : could you rename ptython files ? Sale_ordre_line and sale_order
 - refactoring : i think we should have two modules. One for the full delivery status depending on sale stock and one for the invoice non draft status, depending on sale. What do you think ?

AaronHForgeFlow:
I will propose two separate modules. Thanks for reviewing!

JuanyDForgeflow (approved):
Code + Functional review LGTM

MiquelRForgeFlow (approved):

AaronHForgeFlow:
Superseded https://github.com/OCA/sale-workflow/pull/2552and https://github.com/OCA/sale-workflow/pull/2553

PR #2307, opened 2022-12-16
[15.0][FIX] Update copier template

stefan-tecnativa:
To avoid the isort problem (PyCQA/isort#2077)

@Tecnativa


pedrobaeza:
In this case it's not the Flake8 Gitlab problem. Maybe it was happening in your local because you didn't have the latest definition on the hook. This is worth anyway removing the second line on the commit message. For you local issue, try to do `pre-commit install -f`, and then `pre-commit run -a` to see if the problem persists.

PR #2334, opened 2023-01-12
[16.0][MIG] sale order price recalculation

aliciagaarzo:
Migration v16- module sale_order_price_recalculation


rousseldenis:
@aliciagaarzo In fact, this behavior is fully supported in core:

https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order.py#L936

Could you close this ?

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+            line2._compute_discount()
+            line.write({field: line2[field] for field in fields})
+
+    def recalculate_prices(self):
+        self._recompute_form_line_vals(["price_unit", "discount", "pricelist_item_id"])
+        return True
+
+    def recalculate_names(self):
+        self._recompute_form_line_vals(["name"])
+        return True
```
lmignon, line 38: @aliciagaarzo Thank you for this work. Unfortunately, this module should not be migrated since this functionality is now provided natively by odoo. You can check the method `action_update_prices` from the `sale.order` model.

lmignon (changes requested):

PR #2342, opened 2023-01-18
[15.0][IMP] sale_commercial_partner: promote to stable

chienandalu:
Promoting to stable as otherwise it isn't possible to set dependecies so. I set @alexis-via as contributor as he's the original contributor but I can set other.

cc @Tecnativa TT38310

ping @pedrobaeza


pedrobaeza (approved):
Seems reasonable

PR #2351, opened 2023-01-31
[15.0][ADD] sale_invoice_plan_retention

Saran440:
Add feature retention in sale order view. it help user to auto create retention from sale order

Depend On:
- [ ] https://github.com/OCA/account-invoicing/pull/1288
- [x] sale_invoice_plan


rousseldenis:
@Saran440 Shouldn't it be `sale_invoice_plan_payment_retention`?

rousseldenis:
@Saran440

PR #2352, opened 2023-01-31
[IMP] sale_stock_picking_note: Set default note on partner

SirTakobi:
Allow to set a default value in the partner that propagates to the sale order on partner change.

### DIFF
--- a/sale_stock_picking_note/models/sale_stock.py
+++ b/sale_stock_picking_note/models/sale_stock.py
@@ -1,7 +1,8 @@
 # Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# Copyright 2023 Simone Rubino - TAKOBI
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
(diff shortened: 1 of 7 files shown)


sale_stock_picking_note/models/sale_stock.py
```diff
@@ -11,6 +12,15 @@ class SaleOrder(models.Model):
         string="Picking Note",
     )
 
+    @api.onchange(
+        'partner_id',
+    )
+    def onchange_partner_picking_note(self):
+        partner = self.partner_id
+        if partner:
+            partner_picking_note = partner.default_picking_note
```
rousseldenis, line 21: @SirTakobi Is the real desired behavior of deleting the existing note if there is already one defined (and maybe lose data)?

SirTakobi: I think it is, but it seems to me that you are implying it's a bad thing :smile:
@eLBati can you please check if what's happening is the desired behavior?

rousseldenis: @eLBati

rousseldenis:
@SirTakobi @eLBati No progress on this ?

PR #2356, opened 2023-02-02
[MIG] 16.0: sale_order_product_assortment

Robrechtc:
This module allow to use the product assortment related to partners on sale orders.


sale_order_product_assortment/models/sale_order.py
```diff
@@ -0,0 +1,48 @@
...
+                    filters |= ir_filter
+                    allowed_products += self.env["product.product"].search(
+                        ir_filter._get_eval_domain()
+                    )
+                    if ir_filter.apply_black_list_product_domain:
+                        not_allowed_products += self.env["product.product"].search(
+                            ir_filter._get_eval_black_list_domain()
+                        )
+                    partner_has_filter = True
+            self.allowed_product_ids = allowed_products - not_allowed_products
```
lmignon: @Robrechtc Not related to the migration but IMO, we should avoid this approach where we compute a list of ids that could be very huge to filter product on the sale order line form. A most efficient way could be to use the `web_domain_field` addon https://pypi.org/project/odoo-addon-web-domain-field/ and replace the field `allowed_product_ids` by `allowed_product_domain`

The code could looks like:

in sale_order.py
```python
from odoo.osv.expression import AND

allowed_product_domain = fields.Char(
    compute="_compute_allowed_product_domain",
)

@api.depends("partner_id", "partner_shipping_id", "partner_invoice_id")
def _compute_allowed_product_domain(self):
    self.allowed_product_domain = json.dumps([(1, "=", 1)])
    self.has_allowed_products = False
    partner_field = (
        self.env["ir.config_parameter"]
        .sudo()
        .get_param("sale_order_product_assortment.partner_field", "partner_id")
    )
    filters_partner_domain = self.env["ir.filters"].search(
        [("is_assortment", "=", True)]
    )
    for rec in self:
        domains = []
        for filter in filters_partner_domain.filtered(lambda f, partner=rec[partner_field]: partner.id in f.all_partner_ids.ids):
            domain.append(ir_filter._get_eval_domain())
        if domains:
            rec.has_allowed_products = True
            rec.allowed_product_domain = json.dumps(AND(domains))
         
```

in sale_order_view.xml

```xml

            <field name="partner_id" position="after">
                <field name="allowed_product_domain" invisible="1" />
            </field>
            <!-- Form -->
            <xpath
                expr="//field[@name='order_line']/form//field[@name='product_id']"
                position="attributes"
            >
                <attribute name="domain">parent.allowed_product_domain</attribute>
            </xpath>
        </field>
```

With this code, you no more need to extend the sale order line and the extend of the view is easier (and the performance too on databases with a lot of products)

sale_order_product_assortment/models/sale_order.py
```diff
@@ -0,0 +1,48 @@
...
+        not_allowed_products = self.env["product.product"]
+        self.has_allowed_products = False
+        partner_has_filter = False
+        partner_field = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("sale_order_product_assortment.partner_field", "partner_id")
+        )
+        if self[partner_field]:
+            filters = IrFilters.browse()
```
lmignon: filters is never used and should be removed

sale_order_product_assortment/models/sale_order.py
lmignon:
```suggestion
            for ir_filter in filters_partner_domain:
                if self[partner_field] & ir_filter.all_partner_ids:
                    allowed_products += self.env["product.product"].search(
                        ir_filter._get_eval_domain()
                    )
                    partner_has_filter = True
            self.allowed_product_ids = allowed_products
```
The cal to `get_eval_domain` already take into account black listed products sse https://github.com/OCA/product-attribute/blob/16.0/product_assortment/models/ir_filters.py#L91

lmignon (changes requested):
@Robrechtc Thank you for your work. Some fix todo and proposals to improve the approach...

sale_order_product_assortment/models/sale_order.py
Robrechtc: Thanks for your constructive feedback.

I agree with the web_product_domain-approach but am unsure if I can just add a new external dependency. Is there a way of communicating this to OCA users that are looking to migrate to v16?

rousseldenis (approved):
Code review. Even if it adds a supplementary dependency, we all see the whole enhancement in terms of performances this allows.

Simplified view too. Great! Thanks

rousseldenis:
@lmignon Could you update your review ?

lmignon (commented):
Thank you for the migration @Robrechtc LGTM (Code review) but I'm not able to functionally validate your change. An addon installed into the database override the get_view method in a wrong way....
image

I fixed this kind of error last week maybe a rebase is required to get it...

rousseldenis:
@Robrechtc Could you rebase ?

Robrechtc:
Sorry, I missed the comment by @lmignon. I rebased

pedrobaeza:
Please rebase to solve the conflict.

carolinafernandez-tecnativa:
@Robrechtc could you please rebase for getting a fresh runboat? Thanks!

carolinafernandez-tecnativa:
@Robrechtc how are you? thanks for rebasing!

I tested but is not working.

1- If i add system parameter, when I go to Sales, I got this error message:

image

2- I create this product assortment only for Wood Corner:

image

image

When I go to Sales, and create a sales order for Wood Corner, is not filtering products.

image

image

Could you please check this errors? Thanks!

Robrechtc:
@carolinafernandez-tecnativa I do not currently have the time to review and refactor this further, my sincere apologies

pedrobaeza:
Superseded by #2712

PR #2357, opened 2023-02-02
[16.0][ADD] - sale_order_line_cancel

sbejaoui:
This module allows you to cancel the remaining quantity on sale order by adding
a dedicated action to sale lines. It also add two new fields to track canceled
and remaining to deliver quantities.


sbejaoui:
cc/ @lmignon

lmignon (changes requested):
@sbejaoui I tested the odoo native behavior and it seems that this addon could be simplified and only provide the functionality in a specific context. On a sale order line, you can edit the quantity even if the SO is validated... IMO this behavior is dangerous and should be forbidden. Hopefully, an option is available into the settings to prevent the modification of quantities when the sale order is confirmed... It's only in the context where the modification is no more possible that this functionality makes sens and should be activated. Moreover, I've the feeling that all the custom code written to cancel the moves could be removed (and the dependency on https://github.com/OCA/stock-logistics-workflow/pull/1202). When you change the qty on a confirmed line, Odoo takes the necessary steps to reflect this change in the picking chain. The only thing Odoo doesn't do is to prevent the cancel of quantities that are partially processed. Instead, it generate the return picking to revert this outgoing moves... I prefer to prevent the cancellation of qties already partially processed than adding more work to do by opertors. This behaviour could be a configurable option..

sbejaoui:
> @sbejaoui I tested the odoo native behavior and it seems that this addon ..

Thanks @lmignon for your input,

I understand your PoV, It's true that the change of the ordered quantity trigger the change/cancellation of the related moves.

This feature is meant to simplify the process and make it done with a simple click. It also keep the information of the initial ordered quantity and store the canceled quantity. The difference between the two fields may be helpful for sale reporting.

The dependency to https://github.com/OCA/stock-logistics-workflow/pull/1202 is not a must. It's, IMO, a simpler way to get all sale line related moves instead of recursive loop.

lmignon:
> > @sbejaoui I tested the odoo native behavior and it seems that this addon ..
> 
> Thanks @lmignon for your input,
> 
> I understand your PoV, It's true that the change of the ordered quantity trigger the change/cancellation of the related moves.
> 
> This feature is meant to simplify the process and make it done with a simple click. It also keep the information of the initial ordered quantity and store the canceled quantity. The difference between the two fields may be helpful for sale reporting.
> 
> The dependency to [OCA/stock-logistics-workflow#1202](https://github.com/OCA/stock-logistics-workflow/pull/1202) is not a must. It's, IMO, a simpler way to get all sale line related moves instead of recursive loop.

IMO, the functionality should only be available if the change of qty on confirmed sale order is activated..

sale_cancel_remaining/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows you to cancel the remaining quantity on sale order by adding
+a dedicated action to sale lines. It also add two new fields to track canceled
+and remaining to deliver quantities.
```
lmignon: You should add:

```
This module differs from the original odoo behavior in the following way:

* In odoo, if the update of the qty ordered is allowed on the sale order even 
  when the order is confirmed, odoo will recompute the required stock operations 
  according to the changes done to the qty ordered. This change is possible
  even the stock operations are started for this sale order line.
* In this module, the qty ordered is not updated on the sale order line to keep
  the original qty ordered. At the same time, we cancel only the stock moves
  for the remaining qty to deliver. This is only possible if no operation is
  started for this sale order line.


.. warning::
    
    It's not recommended to use this module at the same time that allowing
    the update of the qty ordered on the sale order line after the order is
    confirmed. This could lead to unpredictable behavior.

```

sale_cancel_remaining/wizards/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2018 Okia SPRL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
lmignon: copyright is useless here

lmignon (approved):
Some comment to add to the desciption

sale_cancel_remaining/views/sale_order_line.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+  <record id="sale_order_line_tree_view" model="ir.ui.view">
+    <field name="model">sale.order.line</field>
+    <field name="inherit_id" ref="sale.view_order_line_tree" />
+    <field name="arch" type="xml">
+      <xpath expr="//field[@name='product_uom_qty']" position="after">
+        <field name="product_qty_canceled" />
```
jbaudoux: As the other fields also have a string to display Order | Delivered | Invoiced
```suggestion
        <field name="product_qty_canceled" string="Canceled"/>
```

sale_cancel_remaining/views/sale_order_line.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="product_qty_remains_to_deliver" />
```
jbaudoux:
```suggestion
        <field name="product_qty_remains_to_deliver" string="To Deliver"/>
```

sale_cancel_remaining/wizards/cancel_remaining_wizard.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @api.model
+    def _check_pickings_to_cancel(self, line):
+        pickings_to_cancel = self._get_pickings_to_cancel(line)
+        if not pickings_to_cancel:
+            raise UserError(_("No picking can be canceled"))
+
+    @api.model
+    def _get_moves_to_cancel(self, line):
+        moves_to_cancel = line.mapped("move_ids").filtered(
```
jbaudoux:
```suggestion
        moves_to_cancel = line.move_ids.filtered(
```

sale_cancel_remaining/wizards/cancel_remaining_wizard.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2018 Okia SPRL
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class CancelRemainingWizard(models.TransientModel):
+    _name = "cancel.remaining.wizard"
```
jbaudoux: Maybe safer to bind the wizard to the model it applies to
```suggestion
    _name = "sale.order.line.cancel"
```

jbaudoux (commented):
I did many tests and it works quite well. It restricts the possibility of changing the quantity on the sale order line but at least the cancelation is working properly while in standard odoo you can end-up with a very weird reverse picking as decreasing the sale order line quantity generates a negative move that is supposed to be merged with the original move but if it can't be assigned to the same picking, it will create a reverse picking.

A few remarks that would make this module more complete:
* when you set the sales order back to quotation, `product_qty_canceled` on the lines should be reset to 0
* I would display the cancel button on a locked sales order as you expect auto-locking to be enabled and you allow canceling the lines
* When canceling the sales order, I would also write `product_qty_remains_to_deliver` in `product_qty_canceled`
* If you don't generate the backorder of the outgoing picking, the `product_qty_canceled` on the lines should also increase

jbaudoux:
> The dependency to [OCA/stock-logistics-workflow#1202](https://github.com/OCA/stock-logistics-workflow/pull/1202) is not a must. It's, IMO, a simpler way to get all sale line related moves instead of recursive loop.

Just a side note that this way of finding the right moves of the chain is not safe as moves can be merged. For example, a single pick move could serve multiple delivery moves. This specific use case is covered by the stock_available_to_promise_release module which, in case you cancel the outgoing move, will unrelease the outgoing move and while unreleasing it, will go through the picking chain up to the source picking and split any merged move.

A solution would be that this module only cancel the outgoing move and have another module using the dependency to [OCA/stock-logistics-workflow#1202](https://github.com/OCA/stock-logistics-workflow/pull/1202) to propagate the cancelation.

You can maybe add this remark in a ROADMAP file.

sale_cancel_remaining/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# © 2016 Sylvain Van Hoof
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale Cancel Remaining",
```
jbaudoux: I would rename this module to sale_order_line_cancel

jbaudoux (changes requested):

jbaudoux (approved):
Thanks for the improvements
LGTM

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    can_cancel_remaining_qty = fields.Boolean(
+        compute="_compute_can_cancel_remaining_qty"
+    )
+
+    @api.depends("product_qty_remains_to_deliver", "state")
+    def _compute_can_cancel_remaining_qty(self):
+        for rec in self:
+            rec.can_cancel_remaining_qty = (
+                rec.product_qty_remains_to_deliver > 0
```
rousseldenis: You should use float_compare

rousseldenis (changes requested):
Code review

sale_order_line_cancel/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# © 2016 Sylvain Van Hoof
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale Order Line Cancel",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, ACSONE SA/NV, Odoo Community Association (OCA)",
```
jbaudoux: Let's keep original authors
```suggestion
    "author": "Okia, BCIM, Camptocamp, ACSONE SA/NV, Odoo Community Association (OCA)",
```

jbaudoux (changes requested):

cyrilmanuel (approved):

cyrilmanuel:
ping @sbejaoui :)

sale_order_line_cancel/models/sale_order_line.py
sbejaoui: done

cyrilmanuel (approved):

cyrilmanuel:
thank you @sbejaoui  :+1:  @rousseldenis  can yo check the PR ?

sale_order_line_cancel/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+* Camptocamp
+* ACSONE SA/NV
+
+Contributors
+~~~~~~~~~~~~
+
+* Sylvain Van Hoof <sylvain@okia.be>
+* Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+* Souheil Bejaoui <souheil.bejaoui@acsone.eu.com>
```
jbaudoux:
```suggestion
* Souheil Bejaoui <souheil.bejaoui@acsone.eu>
```

sale_order_line_cancel/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def _action_cancel(self):
+        sale_moves = self.filtered(
+            lambda m: m.sale_line_id and m.state not in ("done", "cancel")
+        )
+        res = super()._action_cancel()
+        for rec in sale_moves:
+            if rec.state != "cancel":
+                continue
+            rec.sale_line_id.product_qty_canceled = rec.product_uom_qty
```
jbaudoux: This is not correct for the kits (mrp)

sale_order_line_cancel/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def action_draft(self):
+        res = super().action_draft()
+        orders = self.filtered(lambda s: s.state == "draft")
+        orders.order_line.write({"product_qty_canceled": 0})
+        return res
+
+    def _action_cancel(self):
+        orders = self.filtered(lambda s: s.state != "cancel")
+        orders_with_picking = orders.filtered("picking_ids")
```
jbaudoux: Maybe this is better as in `sale_stock` they use `qty_delivered_method`:
```suggestion
        orders_with_picking = orders.filtered(lambda o: any(line.qty_delivered_method == 'stock_move' for line in o.order_line)):
```

sale_order_line_cancel/models/stock_move.py
jbaudoux: we should also ensure on sale.order.line that `product_qty_canceled <= product_uom_qty - qty_delivered`

jbaudoux (commented):
We should also support dropship & kits
in regards of https://github.com/odoo/odoo/blob/16.0/addons/sale_mrp/models/sale_order_line.py#L33

cyrilmanuel:
Hi @sbejaoui little ping, could you check about the comment of @jbaudoux  ?

jbaudoux:
See https://github.com/acsone/sale-workflow/pull/4

sale_order_line_cancel/security/sale_order_line_cancel.xml
```diff
@@ -0,0 +1,14 @@
...
+<odoo>
+    <record model="ir.model.access" id="sale_order_line_cancel_access">
+        <field name="name">sale.order.line.cancel access</field>
+        <field name="model_id" ref="model_sale_order_line_cancel" />
+        <field name="group_id" ref="sales_team.group_sale_manager" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="1" />
+        <field name="perm_write" eval="1" />
+        <field name="perm_unlink" eval="1" />
+    </record>
```
cyrilmanuel, line 13: is it possible to add  entry for users in groups  sales_team.group_sale_salesman ?

<record model="ir.model.access" id="sale_order_line_cancel_access_own_document">
        <field name="name">sale.order.line.cancel access own document</field>
        <field name="model_id" ref="model_sale_order_line_cancel" />
        <field name="group_id" ref="sales_team.group_sale_salesman" />
        <field name="perm_read" eval="1" />
        <field name="perm_create" eval="1" />
        <field name="perm_write" eval="1" />
        <field name="perm_unlink" eval="1" />
    </record>

cyrilmanuel (approved):

sale_order_line_cancel/wizards/sale_order_line_cancel.py
```diff
@@ -0,0 +1,27 @@
...
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class SaleOrderLineCancel(models.TransientModel):
+    _name = "sale.order.line.cancel"
+    _description = "Cancel Remaining Wizard"
+
+    def _get_sale_order_line(self):
+        active_id = self._context.get("active_id")
```
rousseldenis: `_context` is deprecated

```suggestion
        active_id = self.env.context.get("active_id")
```

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -0,0 +1,77 @@
...
+    def _check_moves_to_cancel(self, moves):
+        """Override this method to add checks before cancel"""
+        self.ensure_one()
+
+    def _update_qty_canceled(self):
+        """Update SO line qty canceled only when all remaining moves are canceled"""
+        for line in self:
+            if line._get_moves_to_cancel():
+                continue
+            line.product_qty_canceled = line.qty_to_deliver
```
sbejaoui, line 64: @jbaudoux , 

The canceled quantity is not the same as the quantity to deliver because the quantity to deliver includes returned quantities. 
For example, in an order of 10 units, if I deliver 8 and then cancel the remaining 2, and later my customer returns 2 units, I would expect the following values:

- Quantity to deliver: 4
- Quantity canceled: 2

jbaudoux (approved):

jbaudoux:
@rousseldenis Is this now also good for you? Can you update your review?

rousseldenis (approved):

PR #2318, opened 2022-12-28
[MIG] sale_order_general_discount: migration to 16.0

ferran-S73:

PR #2358, opened 2023-02-02
[16.0][MIG] sale_order_general_discount_triple

RabbitJon-S73:
Standard migration to 16.0

Depends on https://github.com/OCA/sale-workflow/pull/2318


rousseldenis:
@RabbitJon-S73 Don't forget to write an understandable PR's title (not the default one provided by github based on your branch name)

rousseldenis:
@RabbitJon-S73 Could you also add in your PR description the depending PR(s) ?

You could also follow this guide to include it here to help reviewers testing on runboat (and make tests passing).

https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

RabbitJon-S73:
> @RabbitJon-S73 Could you also add in your PR description the depending PR(s) ?
> 
> You could also follow this guide to include it here to help reviewers testing on runboat (and make tests passing).
> 
> https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

Updated!

RabbitJon-S73:
@pedrobaeza could you review?

RabbitJon-S73:
@rousseldenis  @pedrobaeza

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -0,0 +1,54 @@
...
+        digits="Discount",
+        compute=False,
+        store=True,
+        readonly=False,
+        default=0.0,
+    )
+
+    @api.model
+    def default_get(self, fields):
+        vals = super(SaleOrderLine, self).default_get(fields)
```
rousseldenis:
```suggestion
        vals = super().default_get(fields)
```

rousseldenis (approved):
Code review. Minor comment

sale_order_general_discount_triple/models/sale_order_line.py
RabbitJon-S73: @rousseldenis Done!

ioans73 (approved):
LGTM 👍🏻

rousseldenis (approved):
Code review

RabbitJon-S73:
Hi @rousseldenis could we merge this PR?

pedrobaeza:
Changing the `discount` field to a non computed one is provoking all sort of errors when the module is installed, as the standard flows have been totally screwed up. I don't know the reason for that, but you should find another technique for such goal. I have directly disabled it in 9aadac128a383c3883605a381edf34577e806f72 while you find the approach for not provoking errors on CIs. Splitting the CIs isolating the module is not a solution, as anyone using the module will experiment the problem, so we will only trick the system, not solve the problem.

RabbitJon-S73:
> Changing the `discount` field to a non computed one is provoking all sort of errors when the module is installed, as the standard flows have been totally screwed up. I don't know the reason for that, but you should find another technique for such goal. I have directly disabled it in [9aadac1](https://github.com/OCA/sale-workflow/commit/9aadac128a383c3883605a381edf34577e806f72) while you find the approach for not provoking errors on CIs. Splitting the CIs isolating the module is not a solution, as anyone using the module will experiment the problem, so we will only trick the system, not solve the problem.

https://github.com/OCA/sale-workflow/pull/2392 here I revert the non-compute, but there are others changes done. Should I split it in two PR?

pedrobaeza:
Well, a comprehensive separation of commits with the explanation of the changes are the way to go. If all fit in the same commit, but with a good explanation why they are not split, it's OK as well.

PR #2361, opened 2023-02-06
[16.0][MIG] - sale_delivery_state

Hardik-OSI:


rousseldenis:
@Hardik-OSI It seems this is not needed anymore. See https://github.com/OCA/sale-workflow/pull/2280#issuecomment-1377131947

rousseldenis:
See https://github.com/OCA/OCB/blob/16.0/addons/sale_stock/models/sale_order.py#L34

rousseldenis:
@Hardik-OSI

flotho:
Hi any chance to close this PR in favor of the standard Odoo ?

yostashiro:
I made this comment https://github.com/OCA/purchase-workflow/pull/1831#issuecomment-1548924543 on the purchase counterpart, and I think it holds true on the sales side as well. IMO, there is a certain point in keeping this module.

flotho:
Hi @yostashiro ,

Thanks for your answer , so obviously there is more in this module than in the core, so it's normal to keep it.
Yet I would recommend to inherit from the native field delivery_status instead of creating a new one, otherwise once installed you'll have 2 modules providing 2 fields with mostly the same definition and potentialy not the same logic. It could be a mess to interpret and maintain.
What do you think of this proposal ?
Regards

sebastienbeau:
Hi after looking at odoo code. We will still need this module for two reason
- support of service
- support of several delivery step. (if you activate the "2 step" delivery on your warehouse) then native odoo code will set the state to partial when you validate a pick (even if you didn't deliver anything)

Regarding the fact of using or not the same name of the field. Like @yostashiro, I am prefer to add a new fields so we do not add a mandatory dependency on "sale_stock".

@Hardik-OSI are you still interested on this PR or I can rebased it and adapt it to be able to merge it ?

Thanks (and sorry for being so long for the review)

sebastienbeau:
I have done the rebased here: https://github.com/OCA/sale-workflow/pull/2770

sebastienbeau:
@flotho after thinking twice, it's not a good idea to add a new field (user will never know which field to used in advanced filter and group by)

I have refactored the module so we can re-use the native odoo field. see here : https://github.com/OCA/sale-workflow/pull/2770

PR #2339, opened 2023-01-16
[15.0][MIG] sale_missing_cart_tracking > sale_missing_tracking: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT39009

@carlosdauden @sergio-teruel please review

ping @chienandalu

PR #2363, opened 2023-02-07
[15.0][MIG] sale_missing_cart_tracking_tier_validation: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT40150

Pending from:
- [ ] #2339 

@sergio-teruel @chienandalu please review :)


pilarvargas-tecnativa:
@sergio-teruel This module should also be renamed, right?  Without "cart"

sergio-teruel:
Yes, to be consistent

PR #2365, opened 2023-02-07
[16.0][MIG] sale_order_qty_change_no_recompute: Migration to 16.0

ignissak:
https://github.com/OCA/sale-workflow/issues/2215


ghost (approved):

amarcosg (approved):
I like this approach

pedrobaeza (changes requested):
Adding that field is an overhead that shouldn't be for something transient like this

rousseldenis:
@ignissak

pedrobaeza:
A proper migration has been implemented in #2473, as this one only covers changes in the UI, not by code, and it uses an extra field with possible side effects, so closing.

PR #2367, opened 2023-02-07
[13.0][FIX] sale_order_line_price_history: Field order_partner_id is not updated when change partner_id 

carlosdauden:
2 commits with distinct solutions

- Commit 1:
When manipulating the dependencies of related fields, we may need to replicate the relationship itself when the field is on the screen and not readonly.
Because if we force the depends for when they act as Onchange/NewId and the field is on the screen (as of v13 it is saved even if it is invisible without the need for force_save)
So in this case only the related was computed in "onchange mode" when the product was changed and since the field was defined in the line, the old value was saved, so the related field was not obtained from relation when save document.
Replace:
`    order_partner_id = fields.Many2one(depends=["product_id"])`
by:
`    order_partner_id = fields.Many2one(depends=["product_id", "order_id.partner_id"])`

- Commit 2:
Use JS to get partner_id from form view.

@Tecnativa TT41519


carlosdauden:
ping @chienandalu

pedrobaeza (approved):
Squash the 2 existing commits

sale_order_line_price_history/migrations/13.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line sol
+        SET order_partner_id = so.partner_id
+        FROM sale_order so
+        WHERE so.id = sol.order_id AND sol.order_partner_id <> so.partner_id
```
pedrobaeza:
```suggestion
        WHERE so.id = sol.order_id AND sol.order_partner_id != so.partner_id
```

pedrobaeza (approved):

sale_order_line_price_history/migrations/13.0.2.0.0/post-migration.py
carlosdauden: Transact-SQL ISO standard not equal operator is `<>`
Why do you want us to use an alias?
Is it because it looks more like python?

pedrobaeza: Yes, for homogeneity, but honestly I didn't know about the standard being `<>` in SQL...

carlosdauden:
Change and squash done

sergio-teruel (approved):

pedrobaeza:
Can you please fw-port it to upper versions?

PR #2370, opened 2023-02-13
[14.0][FIX] sale_order_type: Disable compute when exist default_journal_id

igallart:
When the "default_journal_id" context value exists, the compute field should not be activated. This occurs for example when an invoice is created from the invoice kanban board, which ignores the context variable and inserts an incorrect journal.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -19,11 +19,17 @@ class AccountMove(models.Model):
     )
 
     @api.depends("partner_id", "company_id")
+    @api.depends_context(
+        "default_move_type", "default_journal_id", "active_model", "company"
+    )
     def _compute_sale_type_id(self):
         # If create invoice from sale order, sale type will not computed.
-        if not self.env.context.get("default_move_type", False) or self.env.context.get(
-            "active_model", False
-        ) in ["sale.order", "sale.advance.payment.inv"]:
+        if (
+            not self.env.context.get("default_move_type", False)
+            or self.env.context.get("default_journal_id", False)
+            or self.env.context.get("active_model", False)
+            in ["sale.order", "sale.advance.payment.inv"]
+        ):
             return
         self.sale_type_id = self.env["sale.order.type"]
         for record in self:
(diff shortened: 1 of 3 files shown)


sale_order_type/models/account_move.py
```diff
@@ -21,9 +21,12 @@ class AccountMove(models.Model):
     @api.depends("partner_id", "company_id")
     def _compute_sale_type_id(self):
```
rousseldenis, line 25: @igallart In fact, you should add an  @api.depends_context() decorator with all keys in the function.

You should add also a key 'company'.

rousseldenis (changes requested):
Code review. At least add the decorator.

A test case would be welcome too

sale_order_type/models/account_move.py
igallart: @rousseldenis , thank you very much for your comment, but what I have done is to correct a particular error that was currently failing, my intention is not to refactor the module, I have adapted the correction to what was there, I think that being a version 12 is not necessary to invest so much time, if it were for a newer version or in the process of migration if it should be done. If someone wants to make the improvements will be welcome.

igallart: Sorry, is version 14 not 12, but my objective of this pull request is to correct a specific error, not to refactor. The original module also does not make use of api_depends_context...

igallart: @rousseldenis , add api_depends_context.

sale_order_type/models/account_move.py
```diff
@@ -19,11 +19,17 @@ class AccountMove(models.Model):
     )
 
     @api.depends("partner_id", "company_id")
+    @api.depends_context(
+        "default_move_type", "default_journal_id", "active_model", "company_id"
```
rousseldenis:
```suggestion
        "default_move_type", "default_journal_id", "active_model", "company"
```

rousseldenis:
@igallart Do you plan to add test on this ?

rafamarpe (approved):
Correct functional test

rousseldenis (approved):

PR #2377, opened 2023-02-15
[14.0][FIX] sale_automatic_workflow: Show warning only when Validate Order

SodexisTeam:


sale_automatic_workflow/models/sale_order.py
```diff
@@ -62,7 +62,7 @@ def _onchange_workflow_process_id(self):
             self.picking_policy = workflow.picking_policy
         if workflow.team_id:
             self.team_id = workflow.team_id.id
-        if workflow.warning:
+        if workflow.warning and workflow.validate_order:
```
rousseldenis, line 65: @SodexisTeam I think this is done at purpose following help message:

image

So, this change is not wanted.

PR #2381, opened 2023-02-17
[15.0][MIG] sale_order_revision

CRogos:
Followup of https://github.com/OCA/sale-workflow/pull/2275
- Renamed test_requirements
- removed oca_requirements
- squashed commits

@simahawk @grindtildeath @cyrilmanuel @leemannd

PR #2382, opened 2023-02-17
[16.0][MIG] sale_order_revision

CRogos:
based on v15 https://github.com/OCA/sale-workflow/pull/2381
required: https://github.com/OCA/server-ux/pull/599


sale_order_revision/view/sale_order.xml
```diff
@@ -0,0 +1,51 @@
...
+            </field>
+        </field>
+    </record>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="priority">15</field>
+        <field name="arch" type="xml">
+            <xpath expr="//header" postion='inside'>
```
rven:
```suggestion
            <xpath expr="//header" position='inside'>
```

CRogos: I've added this in a separate commit for easier backport to v15

MohamedOsman7 (approved):
code + functional review LGTM

rven (approved):

sale_order_revision/view/sale_order.xml
pedrobaeza: Although there's a typo, the default is position="inside", so this is not adding a real fix.

rven: @pedrobaeza I don't think it is working in v16 by default, we are already using this in our project, and the button was not showing at all, this change fixed the issue we had.

pedrobaeza: OK, maybe OWL is stricter in this way.

PR #2384, opened 2023-02-20
[14][ADD] sale_automatic_workflow_with_delay

TDu:


.pre-commit-config.yaml
```diff
@@ -113,7 +113,7 @@ repos:
           - requirements.txt
           - --header
           - "# generated from manifests external_dependencies"
-  - repo: https://gitlab.com/PyCQA/flake8
+  - repo: https://github.com/PyCQA/flake8
```
simahawk: are you sure you are based on latest 14.0?

sale_automatic_workflow_with_delay/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,47 @@
...
+
+class AutomaticWorkflowJob(models.Model):
+    _inherit = "automatic.workflow.job"
+
+    def run_with_workflow(self, sale_workflow):
+        if sale_workflow.validate_order_with_delay:
+            self.env = self.env(
+                context=dict(self.env.context, sale_workflow_validate_with_delay=True)
+            )
+        return super(AutomaticWorkflowJob, self).run_with_workflow(sale_workflow)
```
simahawk:
```suggestion
        super_self = self.with_context(sale_workflow_validate_with_delay=sale_workflow.validate_order_with_delay)
        return super(AutomaticWorkflowJob, super_self).run_with_workflow(sale_workflow)
```

sale_automatic_workflow_with_delay/data/ir_cron.xml
```diff
@@ -0,0 +1,20 @@
...
+        forcecreate="True"
+        id="ir_cron_automatic_workflow_job_delayed"
+        model="ir.cron"
+    >
+        <field name="name">Automatic Workflow Job Delayed</field>
+        <field ref="model_automatic_workflow_job" name="model_id" />
+        <field name="state">code</field>
+        <field
+            name="code"
+        >model.with_context(sale_workflow_job_delayed=True).run()</field>
```
simahawk: IMO we should have a specific method to call

sale_automatic_workflow_with_delay/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _validate_sale_orders(self, order_filter):
+        if self.env.context.get("sale_workflow_validate_with_delay"):
```
simahawk, line 28: Do we really need all these overrides?
Can't we have a specific method to call (eg: run_with_workflow_with_delay() which runs only the allowed ones? 
Only validate orders in this case.

sale_automatic_workflow_with_delay/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _validate_pickings(self, picking_filter):
+        if self.env.context.get("sale_workflow_job_delayed"):
+            return
+        return super()._validate_pickings(picking_filter)
+
+    def _create_invoices(self, create_filter):
+        if self.env.context.get("sale_workflow_job_delayed"):
+            return
+        return super()._validate_invoices(create_filter)
```
simahawk: wrong call to super?

simahawk (changes requested):
a PR description would be nice

sale_automatic_workflow_with_delay/models/automatic_workflow_job.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def run_with_workflow(self, sale_workflow):
+        super_self = self.with_context(
+            sale_workflow_validate_with_delay=sale_workflow.validate_order_with_delay
+        )
+        return super(AutomaticWorkflowJob, super_self).run_with_workflow(sale_workflow)
+
+    def _validate_sale_orders(self, order_filter):
+        if self.env.context.get("sale_workflow_validate_with_delay"):
+            if not self.env.context.get("sale_workflow_job_delayed"):
```
simahawk: this key is not set anywhere, am I wrong?

TDu: indeed, you are right.

sale_automatic_workflow_with_delay/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Thierry Ducrest <thierry.ducrest@camptocamp.com>
```
simahawk, line 1:
```suggestion
* Thierry Ducrest <thierry.ducrest@camptocamp.com>
* Simone Orsi <simone.orsi@camptocamp.com>
```

simahawk:
Unrelated test failing 
```
2023-02-27 07:38:18,060 253 ERROR odoo odoo.addons.sale_by_packaging.tests.test_sale_only_by_packaging: FAIL: TestSaleProductByPackagingOnly.test_create_auto_fill_packaging
  File "/__w/sale-workflow/sale-workflow/sale_by_packaging/tests/test_sale_only_by_packaging.py", line 41, in test_create_auto_fill_packaging
    self.assertFalse(so_line.product_packaging)
AssertionError: product.packaging(12,) is not false
``

**Review by simahawk — COMMENTED** [review]
Sounds good as first version :+1: 
Not approving due to :red_circle: CI

File: sale_automatic_workflow_with_delay/readme/CONTRIBUTORS.rst (hunk 7 above)
Comment (TDu) [reply sale_automatic_workflow_with_delay/readme/CONTRIBUTORS.rst:1]: Added

File: sale_automatic_workflow_with_delay/models/automatic_workflow_job.py [hunk 8]
```diff
@@ -0,0 +1,43 @@
...
+            return
+        return super()._validate_sale_orders(order_filter)
+
+    @api.model
+    def _run_with_delay_workflow_domain(self):
+        return [("validate_order_with_delay", "=", True)]
+
+    @api.model
+    def run_with_delay(self):
+        """Called from ir.cron"""
```
simahawk, line 38: not really a docstring explaining what it does ;)

simahawk (approved):
can you squash?

TDu:
Squashed

sale_automatic_workflow_with_delay/data/ir_cron.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
```
grindtildeath:
```suggestion
<odoo noupdate="1">
```

grindtildeath (commented):
@TDu Could you please explain a bit more how this module differs from https://github.com/OCA/sale-workflow/tree/14.0/sale_automatic_workflow_job and what's the benefit? 
Also since the term `with_delay` is used in queue job module, maybe we could find a better name for this module?

sale_automatic_workflow_with_delay/data/ir_cron.xml
TDu: Changed

TDu:
> @TDu Could you please explain a bit more how this module differs from https://github.com/OCA/sale-workflow/tree/14.0/sale_automatic_workflow_job and what's the benefit? Also since the term `with_delay` is used in queue job module, maybe we could find a better name for this module?

I tried to improve the description of the module.
For the name I understand your point about the similarity with the queue job terminology. And it is unfortunate.
Any suggestion ?

sale_automatic_workflow_with_delay/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,9 @@
+This module allows to execute workflows with a different frequence than
```
rousseldenis, line 1: > with a different frequence

In what manner ?


@TDu I'm extremely doubtful on the usage of this module.

IMHO as @grindtildeath said, the functional needs should be explained in a better way. Technically I'm also doubtful about the way it's done especially this : 

image

Thanks to clarify

simahawk: Thierry is off so I'll try to answer your questions :)

In short: the validation of the same orders by the std cron (which runs every 1m) should be blocked by the ctx key  `sale_workflow_validate_with_delay=sale_workflow.validate_order_with_delay`. 

The main use case is: some orders from shop or from EDI must be confirmed automatically but some of them might require a delay to leave sales people a bit of time to check and do modification or block their automatic approval.
Usually we combine this w/ an automatic assignation of the workflow (there's another module that we would like to share).

This should be explained in the README indeed.

Regarding *_job: it's not the same thing and I think the explanation above clarifies this (correct me if I'm wrong).

Regarding the implementation: it's a bit tricky I must admit and we can probably make it a bit smoother, maybe by reworking the base module a little bit but I think Thierry tried to not impact on that one.

grindtildeath: @simahawk Thanks for the explanation, I definitely got confused by the "with delay" wording used in queue job terminology.

IIUC then the goal is mostly to have a second cron running less often than the standard one and that will validate orders that would not be validated by the first one?

Cannot this be achieved by using an extra filter for order validation (e.g. on the creation date) in a  different workflow record without the need of extra modules? Like a workflow without filter on create date for direct validation and another worflow with filter on create date for what is meant to be a "delayed" validation?

Anyway I'd still suggest to find a better name for the module if it is to be kept in the OCA.

TDu:
About changing the name ?
`sale_automatic_workflow_2nd_cron`
`sale_automatic_workflow_delayed`
`sale_automatic_workflow_postponed`
`sale_automatic_workflow_defered`
Other ideas ?

grindtildeath:
IMO the actual implementation is quite limited in the sense that it doesn't allow to have more than two different cron frequencies, and code would have to be redesigned if there were at least three.

I would suggest to use a filter on create_date if the goal is to introduce a "minimum delay" before a specific object has to be processed (e.g. don't validate sale order if it's been posted in the last two hours).

However, I could see a use case where having different frequencies could make sense (e.g process sale order for vegetables every hours, dairy products every three hours and those for other goods only once a day). For this IMO we would could have fields to define frequencies on the sale.workflow.process and a stored field that would define the "next run" (a bit similar to the Execute only after field on queue jobs) and would be updated on each run (similar to next execution date on crons). This way we could keep a single cron executed every x minutes and filter out records from workflows where the next run date is not past.

TDu:
To implement the solution proposed by @grindtildeath I opened this #2504

PR #2387, opened 2023-02-22
[13.0][IMP] new function for inheritance in sale_manual_delivery

AnizR:
This is a small change to be able to modify which lines are taken in the default get of the manual delivery wizard. (I just extracted the condition into a function)


leemannd (approved):

rousseldenis (approved):

rousseldenis:
@AnizR Next time, please use more conventional `[IMP] sale_manual_delivery: (...)` in commit message.

PR #2388, opened 2023-02-26
[16.0][fix][sale_blanket] the packagee is moved to v16.0

mostafabarmshory:
sale_blanket_order was not available for v16. Here is a patch to support odoo 16.0. needs more work to update views for better performance.

I test it on a local instance. Needs some configuration:

- Enabling product unit (UOM)
- Enabling chart of accounts and sales


mostafabarmshory:
Dear all, tests are failed due to others!?

PR #2392, opened 2023-02-28
[16.0][IMP] sale_order_general_discount_triple

RabbitJon-S73:
Pricelist discount is override by global discount, With this improve pricelist discount can be applied in any discount like global discount does


RabbitJon-S73:
@rousseldenis could you review this contribution?

RabbitJon-S73:
@AshishHirapara could you review? Thanks!

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -52,3 +52,31 @@ def create(self, vals):
         if general_discount:
             vals[general_discount] = sale_order.general_discount
         return super().create(vals)
+
+    @api.onchange("product_id", "product_uom_qty", "product_uom")
+    def _onchange_product_id_discount(self):
```
rousseldenis: @RabbitJon-S73 I would avoid onchanges by changing fields properties to compute, stored=True, readonly=False

sale_order_general_discount_triple/tests/test_module.py
```diff
@@ -2,26 +2,42 @@
...
-
-        product_form = Form(product_obj)
-        product_form.name = "Test Product"
-        product = product_form.save()
-
-        partner_form = Form(partner_obj)
-        partner_form.name = "Test Partner"
-        partner = partner_form.save()
-        setting_form = Form(setting_obj)
+    def setUp(self):
```
rousseldenis: Change this to classmethod

RabbitJon-S73:
@rousseldenis changed

hildickethan (approved):

sale_order_general_discount_triple/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Sale Order General Discount Triple",
     "summary": "General discount per sale order with triple",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
```
rousseldenis: Unless you have migration scripts that depends on this, this is not necessary, done by ocabot

RabbitJon-S73: I didn't know. Its done

RabbitJon-S73:
@rousseldenis if everything its okey now could we merge this PR?

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -4,51 +4,81 @@
...
-        readonly=False,
-        default=0.0,
-    )
-
-    @api.model
-    def default_get(self, fields):
-        vals = super().default_get(fields)
-        general_discount = (
+    discount2 = fields.Float(compute="_compute_discount2", store=True, readonly=False)
+    discount3 = fields.Float(compute="_compute_discount3", store=True, readonly=False)
```
hildickethan, line 8: these need to be `precompute=True` to avoid these warnings
 `Field sale.order.line.price_reduce_taxinc cannot be precomputed as it depends on non-precomputed field sale.order.line.price_total`
`Field sale.order.line.price_total cannot be precomputed as it depends on non-precomputed field sale.order.line.discount2`
and a few more caused by the same root
I will push the change

RabbitJon-S73:
@rousseldenis I splited the changes on this PR in two commits, one for fixing the issue commented on https://github.com/OCA/sale-workflow/pull/2358 and the other one for the improvements.


@pedrobaeza

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -4,51 +4,44 @@
...
-        vals.update(
-            {
-                "discount": 0.0,
-                "discount2": 0.0,
-                "discount3": 0.0,
-            }
-        )
+    @api.depends("product_id", "product_uom", "product_uom_qty")
+    def _compute_discount(self):
+        if hasattr(super(), "_compute_discount"):
```
pedrobaeza: I don't think you need this if, as the compute is on the module that declares the field, isn't it?

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -4,51 +4,44 @@
...
+            if general_discount == "discount2":
+                vals.update({"discount2": line.order_id.partner_id.sale_discount})
+            line.update(vals)
+
+    @api.depends("product_id", "product_uom", "product_uom_qty")
+    def _compute_discount3(self):
+        general_discount = self._get_general_discount_field()
+        for line in self:
+            vals = {"discount3": 0.0}
+            if general_discount == "discount3":
```
pedrobaeza: Exit early if the discount field is not the one we are looking for, and you can pre-assign `self.discount3 = 0`

RabbitJon-S73: Done!

RabbitJon-S73:
@pedrobaeza @rousseldenis PR ready to merge

pedrobaeza (approved):

hildickethan:
Since this fixes the compute issues you can change the `installable` in the manifest back to `True`
And also would be cleaner to cut off the `<` typo from this commit 
image

sale_order_general_discount_triple/README.rst
```diff
@@ -38,8 +38,10 @@ Usage
 To use features of this module, you need to:
 
 #. Go to Settings > General Settings > Pricing
-#. Under pricing section, select your prefered Discount type from 'General Discount' field.
-#. The selected discount type among the 3 discounts will be automatically parsed on sale order line while using 'Discount %' text field.
+#. Under pricing section, select your prefered Discount type from 'General Discount' field and
+   the same for 'Pricelist Discount'.
+#. The selected global discount type among the 3 discounts will be automatically parsed on sale order line while using 'Discount %' text field.
+#. If pricelist has a discount, it will be applied on the selected pricelist discount.
```
francesco-ooops: @RabbitJon-S73 can you add this change in the USAGE file?

RabbitJon-S73: Done! I also add some info that can be usefull

francesco-ooops:
@RabbitJon-S73 question: what is the purpose/behavior of the option "no apply" for "pricelist discount" setting? I think it can be clarified in usage and/or description

francesco-ooops:
@RabbitJon-S73 on a side note, you may want to take a look at this behavior: 

- set "general discount" in "discount 2" (settings)
- in SO, add a product and a general discount
- now add shipping expenses to SO

CURRENT BEHAVIOR

Discount for shipping product is set on Discount 1 and not Discount 2

image

FYI we're trying to fix this in #2666

RabbitJon-S73:
> @RabbitJon-S73 question: what is the purpose/behavior of the option "no apply" for "pricelist discount" setting? I think it can be clarified in usage and/or description

I did that in order to let the user choose the "old behaviour" with no pricelist discounts. What do you think about doing the same with general discount to allow disable it?

RabbitJon-S73:
> @RabbitJon-S73 on a side note, you may want to take a look at this behavior:
> 
>     * set "general discount" in "discount 2" (settings)
> 
>     * in SO, add a product and a general discount
> 
>     * now add shipping expenses to SO
> 
> 
> CURRENT BEHAVIOR
> 
> Discount for shipping product is set on Discount 1 and not Discount 2
> 
> image
> 
> FYI we're trying to fix this in #2666

I see your example in runboat but i can't reproduce it anymore, could you give me more details? The only I noticed is that general discount is not applied on discount2 of delivery line because the compute is not triggered. I'm trying to fix it

francesco-ooops:
> > @RabbitJon-S73 question: what is the purpose/behavior of the option "no apply" for "pricelist discount" setting? I think it can be clarified in usage and/or description
> 
> I did that in order to let the user choose the "old behaviour" with no pricelist discounts. What do you think about doing the same with general discount to allow disable it?

Can you explain what you mean by "old behaviour with no pricelist discounts"?

francesco-ooops:
> I see your example in runboat but i can't reproduce it anymore, could you give me more details? The only I noticed is that general discount is not applied on discount2 of delivery line because the compute is not triggered. I'm trying to fix it

You can check this video: https://recordit.co/R6S2YKT2Id (you can see there are also other behaviors that don't seem right...)

RabbitJon-S73:
> > > @RabbitJon-S73 question: what is the purpose/behavior of the option "no apply" for "pricelist discount" setting? I think it can be clarified in usage and/or description
> > 
> > 
> > I did that in order to let the user choose the "old behaviour" with no pricelist discounts. What do you think about doing the same with general discount to allow disable it?
> 
> Can you explain what you mean by "old behaviour with no pricelist discounts"?

Before this PR, general discount overwrite the pricelist discount, i suppose that is not desired but I think that could be useful for some users to disable pricelist discount or general discount by choosing the option to "No apply"

RabbitJon-S73:
> > I see your example in runboat but i can't reproduce it anymore, could you give me more details? The only I noticed is that general discount is not applied on discount2 of delivery line because the compute is not triggered. I'm trying to fix it
> 
> You can check this video: https://recordit.co/R6S2YKT2Id (you can see there are also other behaviors that don't seem right...)

I think I fixed that, but I'm not happy with the solution since I had to call compute functions from the method that creates the delivery line. If someone finds a better solution it's welcome

francesco-ooops:
> > > > @RabbitJon-S73 question: what is the purpose/behavior of the option "no apply" for "pricelist discount" setting? I think it can be clarified in usage and/or description
> > > 
> > > 
> > > I did that in order to let the user choose the "old behaviour" with no pricelist discounts. What do you think about doing the same with general discount to allow disable it?
> > 
> > 
> > Can you explain what you mean by "old behaviour with no pricelist discounts"?
> 
> Before this PR, general discount overwrite the pricelist discount, i suppose that is not desired but I think that could be useful for some users to disable pricelist discount or general discount by choosing the option to "No apply"

I don't think it's good ux to allow users to set discounts in pricelist and disabling them through a general setting. 

I think having a setting (disc 1, 2 or 3) for each type of discount (general and pricelist) is sufficient, but I would leave both fields in settings empty and required to force user to set them up correctly.

RabbitJon-S73:
@francesco-ooops @pedrobaeza can be merged?

celm1990:
@RabbitJon-S73  delete default value but fields are `required` cause issue on other modules, please can review other alternative(remove attribute `required`?)
https://github.com/OCA/sale-workflow/pull/2392/files#diff-06b9aec82dbbd6bf4122a463642bbeac0e486e2e760e5cf336b3a18c153fe43aL14

See https://github.com/OCA/sale-workflow/actions/runs/6808350353/job/18512649384
image

PR #2394, opened 2023-03-02
[13.0][IMP] sale_order_secondary_unit: Set default units in product onchagemethod before. commit not needed in higher versions.

sergio-teruel:
cc @Tecnativa  TT41926


CarlosRoca13 (approved):
LGTM

carlosdauden (approved):

PR #2403, opened 2023-03-07
[16.0][MIG] sale_procurement_group_by_line: Migration to 16.0

kimirondon:

PR #2341, opened 2023-01-17
[16.0] [MIG] sale_order_line_date

Kev-Roche:

PR #2404, opened 2023-03-07
[16.0][MIG] sale_delivery_split_date: Migration to 16.0

kimirondon:


kimirondon:
depending on #2341 and #2403

ajaniszewska-dev:
to include dependencies and fix build you could add  test-requirements.txt and in it something like:
`
odoo-addon-<dependency_addon_name> @ git+https://github.com/OCA/<addon_repo>.git@refs/pull/<PR number>/head#subdirectory=setup/<addon_name>`

2 entries since you depends on 2 modules.
just in separate commit like "[DO NOT MERGE] test-requirements.txt"

kimirondon:
hello @ajaniszewska-dev ! thanks a lot for the help.

sale_delivery_split_date/models/sale_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+        """Return a key with priority to be used to regroup lines in multiple
+        procurement groups
+        """
+        priority = 24
+        key = super()._get_procurement_group_key()
+        # Check priority
+        if key[0] < priority:
+            if self.commitment_date:
+                # group by date instead of datetime
+                return (priority, self.commitment_date.date())
```
nguyenminhchien, line 28: @kimirondon 
Hello, my timezone is GMT+7
i create an order having 2 lines:
- 1st line: Delivery date at 07/13/2023 05:00:00
- 2nd line: Delivery date at 07/13/2023 10:00:00
They are in the same date (07/13/2023) but 2 delivery orders are created after confirming the order.
this line `return (priority, self.commitment_date.date())` should take into account the user timezone
image

nguyenminhchien: we should add an unittest for this case as well.

nguyenminhchien (changes requested):

nguyenminhchien (approved):
look good as in previous version

sale_delivery_split_date/models/sale_order_line.py
nguyenminhchien: we can add another commit to improve this later.

kimirondon:
hello @nguyenminhchien !! yes like you said is the same in previous versions, it could be tricky to split by time, if the difference is seconds it will create different pickings.

gurneyalex (approved):

PR #2405, opened 2023-03-08
[15.0][FIX] sale_order_secondary_unit: Make test resilient

pedrobaeza:
As the test should only test that no change on the UoM quantity is done when modifying the secondary unit, we remember the existing quantity and compare it later with the final value, instead of testing against a fixed value, so any interaction with other modules don't interfere.

@Tecnativa


pedrobaeza:
This puts the branch again in green, and I have also cleared other warnings found on the log, so merging:

/ocabot merge patch

PR #2408, opened 2023-03-08
[15.0][FIX] sale_invoice_plan: hide create invoice button

ps-tubtim:
This pull request is used to hide **Create Invoice** button when **Use Invoice Plan** because an invoice that was created by **Create Invoice** button does not link with the invoice plan.


dreispt (approved):

Saran440 (approved):
LGTM :+1:

PR #2409, opened 2023-03-09
[15.0][ADD]: sale_downpayment_product

imlopes:


sale_downpayment_product/models/product_template.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    is_down_payment_product = fields.Boolean(string="Down-payment product")
```
ivantodorovich, line 10: 💡 Rather than a flag, this would be a good candidate for a new `detailed_type` IMO

```suggestion
    detailed_type = fields.Selection(
        selection_add=[("downpayment", "Down-payment")],
        ondelete={"downpayment": "set service"}
    )

    def _detailed_type_mapping(self):
        res = super()._detailed_type_mapping()
        res["downpayment"] = "service"
        return res
```

This would simplify the module because you'd no longer need this view: https://github.com/OCA/sale-workflow/blob/06af12f95de2ced1e4fa2055a190b62557f70fd3/sale_downpayment_product/views/product_template.xml
And the `product_id` field domain would just be `[("detailed_type", "=", "downpayment")]`

ivantodorovich (approved):

sale_downpayment_product/models/product_template.py
rousseldenis: Agree.

rousseldenis:
@imlopes

PR #2418, opened 2023-03-17
[16.0][MIG] sale_manual_delivery: Migration to 16.0

nguyenhk99:
Based upon:
* #1983


StephaneMangin (approved):

leemannd (approved):

sale_manual_delivery/tests/test_manual_delivery.py
```diff
@@ -289,7 +289,7 @@ def test_03_sale_selected_lines(self):
             'Picking should be created after "manual delivery" wizard call',
         )
         self.assertEqual(
-            len(order3.picking_ids.move_lines),
+            len(order3.picking_ids.move_line_ids),
```
grindtildeath: :warning: `move_lines` refers to `stock.move` whereas `move_line_ids` refers to `stock.move.line`.
I guess, here you should use `move_ids`: https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L346

sale_manual_delivery/tests/test_manual_delivery.py
```diff
@@ -412,7 +412,7 @@ def test_03_sale_multi_delivery(self):
             ' after "manual delivery" wizard call with same date',
         )
         self.assertEqual(
-            sum(first_picking.mapped("move_lines.product_uom_qty")),
+            sum(first_picking.mapped("move_line_ids.reserved_uom_qty")),
```
grindtildeath: Same as above, AFAIU we want to test the quantity of each move that is created, not how much was reserved.

nguyenhk99: Hello @grindtildeath , Thank for your review. The PR has been updated.

PR #648, opened 2018-05-11
[10.0][ADD] sale_procurement_amendment

rousseldenis:
Waiting for this change has been merged in 10.0 :
- [x] https://github.com/odoo/odoo/pull/20874. Done in https://github.com/odoo/odoo/commit/58e85bdec08f6f07afe7b8cdfaea6f7cef2776d1

PR #2423, opened 2023-03-21
[16.0][ADD] sale_route_amendment

lmignon:
Once this addon is installed, when a route is modified on a sale order line, the related existing moves are cancelled and the procurement rules are replayed. This is only possible if nothing has been already begun in stock for the given line. Otherwise, a user error is raised.


lmignon:
ping @rousseldenis

rousseldenis (approved):

rousseldenis:
@sbejaoui

rousseldenis:
@jbaudoux

jbaudoux (commented):
I would start the module name with sale_stock_xxxx
and what about propagate_change instead of amendment ?
`sale_stock_route_propagate_change`

*Edit: I changed my mind*
I would start the module name with sale_order_line_xxxx
and what about propagate_change instead of amendment ?
`sale_order_line_route_propagate_change`

rousseldenis:
> I would start the module name with sale_stock_xxxx and what about propagate_change instead of amendment ? `sale_stock_route_propagate_change`
> 
> _Edit: I changed my mind_ I would start the module name with sale_order_line_xxxx and what about propagate_change instead of amendment ? `sale_order_line_route_propagate_change`

@jbaudoux This was already existing. See: https://github.com/OCA/sale-workflow/pull/648

That's why we kept `amendment`

rousseldenis:
@jbaudoux

jbaudoux:
> @jbaudoux

I'm not fan of the module name.
But it's just my opinion...

PR #2436, opened 2023-03-24
[MIG][16.0]sale_stock_picking_note

Vandan-OSI:


dalonsod:
This PR should be closed, because addon was already migrated

PR #2438, opened 2023-03-28
[16.0][MIG] sale_invoice_plan: Migrate to 16.0

MS-OSI:
Sales Invoice Plan
==============

By standard feature, user can gradually create partial invoices, one by one.
This module add ability to create invoices based on the predefined invoice plan,
either all at once, or one by one.
The plan support both advance invoice and installment invoices.


imlopes (approved):
LGTM

DigitalAutomations:
Any update about this?

dreispt (approved):

dreispt:
Added forward port of https://github.com/OCA/sale-workflow/pull/2408

dreispt:
@kittiu I see you are the maintainer. 
With just the basic migration to 16.0 this worked terribly wrong.
I added this [fix](https://github.com/OCA/sale-workflow/pull/2438/commits/8f3ec06fa7ce814837384027384b7a5a91bc7352) to try make it work.

This migration might be worth deeper review and testing from you, if possible. Thanks!

Borruso:
can you see this PR?
https://github.com/ursais/sale-workflow/pull/10

kittiu:
@dreispt let me check.

> @kittiu I see you are the maintainer. With just the basic migration to 16.0 this worked terribly wrong. I added this [fix](https://github.com/OCA/sale-workflow/pull/2438/commits/8f3ec06fa7ce814837384027384b7a5a91bc7352) to try make it work.
> 
> This migration might be worth deeper review and testing from you, if possible. Thanks!

Sorry for delayed response, I will review this shortly. What is in the "fix" link, clicked and see nothing there.

dreispt:
@kittiu https://github.com/OCA/sale-workflow/pull/2438/commits/a52b26077c7f76a6b8ebf4da73b2b1c23e5a0166

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,185 @@
...
+    invoice_plan_total_percent = fields.Float(
+        compute="_compute_invoice_plan_total",
+        string="Percent",
+    )
+    invoice_plan_total_amount = fields.Monetary(
+        compute="_compute_invoice_plan_total",
+        string="Total Amount",
+    )
+    next_installment_date = fields.Date(
+        compute="_compute_next_installment_date",
```
kittiu: This method is missing, the module still can't get installed.

sale_invoice_plan/models/sale.py
```diff
@@ -0,0 +1,185 @@
...
+        compute="_compute_invoiced_installment_amount",
+        store=True,
+    )
+    pending_installment_amount = fields.Monetary(
+        compute="_compute_pending_installment_amount",
+        store=True,
+    )
+
+    @api.depends("invoice_plan_ids")
+    def _compute_next_installment_date_amount(self):
```
kittiu:
```suggestion
    def _compute_next_installment_date(self):
```

kittiu: But somehow, these new methods is not triggered and recomputed. So, currently all the 3 new fields, next_installment_date, invoiced_installment_amount, pending_installment_amount is not computed correctly.

kittiu:
@dreispt By the way, shouldn't new features be in separated PR? So this can be plain MIG first.

kittiu:
I think I found something. Let me explain.

We only use this module until version 14 without problem. I think bug about Advance was introduced in **version 15**, when following new fields was introduced.

https://github.com/OCA/sale-workflow/blob/15.0/sale_invoice_plan/models/sale.py#L28-L35

`invoice_plan_total_percent` and `invoice_plan_total_amount` -> trigger `[sale.order]._compute_invoice_plan_total` which also trigger `[sale.order.plan]._compute_amount` that mess up the percent.

My test is,
* Case Advance 10%, and 4 installments 25% each.
* Create the Advance -> OK.
* Then create 1st Installment and it mess up the percent in installment table.

kittiu:
> I think I found something. Let me explain.
> 
> We only use this module until version 14 without problem. I think bug about Advance was introduced in **version 15**, when following new fields was introduced.
> 
> https://github.com/OCA/sale-workflow/blob/15.0/sale_invoice_plan/models/sale.py#L28-L35
> 
> `invoice_plan_total_percent` and `invoice_plan_total_amount` -> trigger `[sale.order]._compute_invoice_plan_total` which also trigger `[sale.order.plan]._compute_amount` that mess up the percent.

OK the bug mentioned above was fixed here https://github.com/OCA/sale-workflow/pull/2543

Let's merge it and cherry pick it here.

dreispt:
@kittiu Thank you for looking into this.
You are right, I moved the changes to https://github.com/ursais/sale-workflow/pull/12
I also forward ported the 15.0 fix you mentioned.

kittiu:
So, I think this one is ready after porting the fix @Saran440

kittiu:
Hello,

As I come back to check this module, sale_invoice_plan. I tested that everything including Advance works fine on version 14 and 15.
But when I migrate 15 to 16, the Advance is not working correctly. So, I think there must be changes on Odoo core.

I want to propose to close this PR first, and I will do the fresh upgrade from 15 to 16 make sure everything works according to its normality. Then, if you have any other features to add, you can do in separated PR (so it can be backport too)

cc @dreispt @MS-OSI

kittiu:
Here the new PR with minimum fix to make it works as in previous version https://github.com/OCA/sale-workflow/pull/2988

kittiu:
@MS-OSI can you help closing this one? Thanks!

kittiu:
Can anyone help closing this PR please. Thanks!

DigitalAutomations:
Any udpdate?
What is the correct pr to test ?

Saran440:
@DigitalAutomations you can test PR https://github.com/OCA/sale-workflow/pull/2988

rousseldenis:
As suggested

PR #2416, opened 2023-03-16
[FIX] sale_blanket_order: do not use supplier_taxes_id

maciej-wichowski:
Backporting a fix from 14.0 migration. 
This PR is made according to [closed PR](https://github.com/OCA/sale-workflow/pull/2230) and related with issue #2228 , since it appears to be abandoned.

PR #2439, opened 2023-03-29
[FIX] sale_blanket_order - proper taxes computation

### LINKED ISSUE #2440: [13.0] sale_blanket_order - taxes are not properly computed according to fiscal position
## Module
`sale_blanket_order`

## Describe the bug
There are few issues in the code that prevent proper computation of taxes according to fiscal position. Right now fiscal position does not any effect on the taxes.

## To Reproduce
**Affected versions**:
`13.0`

1. Create blanket order, add partner with custom fiscal position
2. Add a product to the order

**Expected behavior**
Taxes are recomputed according to fiscal position.

## FIX
I propose to fix it with PR #2439

maciej-wichowski:
Correct taxes computation according to fiscal position:
* Add `fiscal_position_id` into form view
* Add `onchange` method to properly recompute taxes after fiscal position change

Somehow associated PR #2416 
Fixes #2440

### DIFF
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -136,7 +136,10 @@ class BlanketOrder(models.Model):
     sale_count = fields.Integer(compute="_compute_sale_count")
 
     fiscal_position_id = fields.Many2one(
-        "account.fiscal.position", string="Fiscal Position"
+        "account.fiscal.position",
+        string="Fiscal Position",
+        domain="[('company_id', '=', company_id)]",
+        check_company=True,
     )
 
     amount_untaxed = fields.Monetary(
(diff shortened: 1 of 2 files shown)


steeluis:
This haven't been merged yet, the problem still exists in at least 13.0 to 16.0. What is the problem?

StefanRijnhart:
@steeluis The problem is that 13.0 is already quite old. It would be good to propose on 16 or 17 to get reviewers' interest, then backport.

maciej-wichowski:
@StefanRijnhart I don't see any point in fishing for reviewers interest. I created an issue (it's a major problem in my opinion), proposed a fix and had no response. The fact that no one seems to bother, means the module is not widely used. So no point in spending more time on it.

StefanRijnhart:
@maciej-wichowski Fair enough!

rousseldenis:
> @StefanRijnhart I don't see any point in fishing for reviewers interest. I created an issue (it's a major problem in my opinion), proposed a fix and had no response. The fact that no one seems to bother, means the module is not widely used. So no point in spending more time on it.

Thanks for the fix.

What @StefanRijnhart wanted to say is an answer to @steeluis comment. But, indeed, OCA is based on contributors good will. The only way to have this merged is to attract people.

maciej-wichowski:
@StefanRijnhart @rousseldenis I'm sorry, I didn't want to sound unpleasant. Just wanted to point out why I'm not doing anything further to fix that matter.

StefanRijnhart:
@maciej-wichowski That's perfectly understandable! No offence taken, or intended on my side.

rousseldenis:
> @maciej-wichowski That's perfectly understandable! No offence taken, or intended on my side.

Nor mine :-)

PR #2442, opened 2023-03-30
[15.0][FIX] sale_credit_point: Adjusting  partner credit_point constrain

stefan-tecnativa:
cc @Tecnativa

The `_check_credit_point method` has a contrain on  `credit_point`, but it is not evaluating how many partners are being passed. So when the method tries to evaluate a recordset with several partners, it will throw a value error complaining that it expects a singleton. With this modification we ensure that only one value is passed at the same time. 

ping @ernestotejeda


ernestotejeda (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2451, opened 2023-03-30
[16.0][MIG] sale_order_line_input

SoporteOperu:


dreispt (approved):

dreispt (commented):
Local machine testing showed error, regarding a missing "parent" used on views.

dreispt (changes requested):
Local machine testing showed error, regarding a missing "parent" used on views.
I feel this needs fixing.

PR #2460, opened 2023-03-31
[14.0][FIX] Stop deleting lot on sale order cancelation

bguillot:
When you cancel a sale order the lot could have been already bought or produce so we want to avoid deleting it automaticaly.


florian-dacosta (approved):
I agree, if someone would want to cancel the lot, it should be done in a submodule or with an parameter because it is too violent. (you may already have some stock for these lot for instance...

bealdav (approved):

PR #2471, opened 2023-04-11
[16.0][mig] sale_last_price_info

aktiv-heli-kantawala:


sale_last_price_info/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+{
+    "name": "Product Last Price Info - Sale",
+    "version": "16.0.1.0.0",
+    "author": "AvanzOSC, Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "category": "Sales",
+    "license": "AGPL-3",
+    "contributors": [
```
rousseldenis: This is not a valid entry. Should be in CONTRIBUTORS.rst only.

If some of them (or you) want to adopt as maintainer, you should add a "maintainers":` [<github login>, ...] `instead

sale_last_price_info/models/product.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _compute_last_sale(self):
+        """Get last sale price, last sale date and last customer"""
+        so_line_obj = self.env["sale.order.line"]
+        for product in self:
+            line = so_line_obj.search(
```
rousseldenis: Something has to be done here as this will generate as many queries as products in recordset.

sale_last_price_info/models/product.py
```diff
@@ -0,0 +1,28 @@
...
+                [("product_id", "=", product.id), ("state", "in", ["sale", "done"])],
+                limit=1,
+                order="date_order_sale_last_price_info desc",
+            )
+            product.last_sale_date = fields.Datetime.to_string(
```
rousseldenis:
```suggestion
            product.update({
                "last_sale_date" : fields.Datetime.to_string(line.date_order_sale_last_price_info),
                "last_sale_price": line.price_unit,
                "last_customer_id": line.order_id.partner_id,
            })
```

sale_last_price_info/models/product.py
```diff
@@ -0,0 +1,28 @@
...
+                line.date_order_sale_last_price_info
+            )
+            product.last_sale_price = line.price_unit
+            product.last_customer_id = line.order_id.partner_id
+
+    last_sale_price = fields.Float(compute="_compute_last_sale")
```
rousseldenis: Fields definitions should be on top

sale_last_price_info/tests/test_sale_last_price_info.py
```diff
@@ -0,0 +1,28 @@
...
+        self.sale_order_model = self.env["sale.order"]
+        self.sale_line_model = self.env["sale.order.line"]
+        self.sale_order = self.env.ref("sale.sale_order_4")
+        self.sale_line = self.env.ref("sale.sale_order_line_9")
+        self.partner = self.env.ref("base.res_partner_3")
+        self.product = self.env.ref("product.product_delivery_02")
+        self.price_unit = 100.0
+
+    def test_sale_last_price_info_demo(self):
+        sale_line = self.sale_line_model.search(
```
rousseldenis, line 18: It's better to rely on new records created in test. Then, don't search but rely on the record values you've just created.

rousseldenis (changes requested):
Code review

rousseldenis (approved):
Code and functional review

dreispt (approved):

PR #2473, opened 2023-04-11
[16.0][MIG] sale_order_qty_change_no_recompute: Migration to version 16.0

Yadier-Tecnativa:
@Tecnativa TT42279


sale_order_qty_change_no_recompute/models/sale_order.py
```diff
@@ -2,28 +2,34 @@
...
-        ):
-            cls = type(self)
-            for method in self._onchange_methods.get(field_name, ()):
-                if method == cls.product_uom_change:
-                    self._onchange_methods[field_name].remove(method)
-                    break
-        return super()._onchange_eval(field_name, onchange, result)
+    @api.depends("product_id", "product_uom", "product_uom_qty")
+    def _compute_price_unit(self):
+        for line in self:
```
pedrobaeza: You don't have to copy the upstream method. This should be restricted to just store the current price and discount before calling super, and then restore it after the call.

pedrobaeza (changes requested):
I understand that the previous code is not working. Isn't it?

sale_order_qty_change_no_recompute/models/sale_order.py
Yadier-Tecnativa: The computed method had to be overwritten for when the dependent fields change and obtain a referential value to the old product, in order to compare, since the calculation of the new price depends on the product

pedrobaeza: But we don't want that... this makes this module incompatible with any change in the price unit computations by other modules, as it's totally overwritten the way it's computed. We just want to not perform the computations on quantity changed. You have to find another technique.

pedrobaeza (approved):
Changed to a new approach to get the feature covered.

sergio-teruel (approved):
Code and functional review.
:+1:

sergio-teruel:
ping @juangomezzurita
Done!

PR #2477, opened 2023-04-12
[16.0][mig] sale_product_set

aktiv-heli-kantawala:


sale_product_set/readme/USAGE.rst
```diff
@@ -7,18 +7,11 @@ To use this module, you need to:
     - Sort *set* lines, this order will be the default when added into the
       quotation
 
-.. image:: /sale_product_set/static/description/product_set.png
```
HugoCordobaLeal: I agree with removing images from previous versions, but it would be correct to add those updated images.

HugoCordobaLeal:
The rest of the changes LGTM.

sale_product_set/readme/USAGE.rst
rousseldenis: @heliaktiv Instead of removing the indication images, it should be better to do new ones if visual changes has occurred. Thanks

sale_product_set/tests/test_product_set.py
```diff
@@ -0,0 +1,202 @@
...
+            self.env.ref("sale_product_set.product_set_line_computer_4").product_id.id
+        )
+        seq_line2 = sequence.pop(
+            self.env.ref("sale_product_set.product_set_line_computer_1").product_id.id
+        )
+        seq_line3 = sequence.pop(
+            self.env.ref("sale_product_set.product_set_line_computer_3").product_id.id
+        )
+        self.assertTrue(
+            max(v for k, v in sequence.items()) < seq_line1 < seq_line2 < seq_line3
```
CRogos, line 79: here

sale_product_set/tests/test_product_set.py
```diff
@@ -0,0 +1,202 @@
...
+            self.env.ref("sale_product_set.product_set_line_computer_4").product_id
+        )
+        seq_line2 = sequence.pop(
+            self.env.ref("sale_product_set.product_set_line_computer_1").product_id
+        )
+        seq_line3 = sequence.pop(
+            self.env.ref("sale_product_set.product_set_line_computer_3").product_id
+        )
+        self.assertTrue(
+            max(v for k, v in sequence.items()) < seq_line1 < seq_line2 < seq_line3
```
CRogos, line 103: here

sale_product_set/wizard/product_set_add.py
```diff
@@ -107,7 +107,7 @@ def _prepare_order_lines(self):
     def _get_max_sequence(self):
         max_sequence = 0
         if self.order_id.order_line:
-            max_sequence = max([line.sequence for line in self.order_id.order_line])
+            max_sequence = max(line.sequence for line in self.order_id.order_line)
```
CRogos, line 110: here

CRogos (approved):
Code review LGTM, beside of one concern:

Is there a problem with the max function in case sequence has only one element?
image

Anyway, I could not find any problem while testing.

Yadier-Tecnativa:
Please you should check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests
 for a procedure for reducing commits from "OCA Transbot...".  in order to contains the minimum necessary commit messages.

sale_product_set/wizard/product_set_add.py
Yadier-Tecnativa, line 1: The use of the @api.onchange decorator is recommended instead of the one computed with the @api.depends decorator since in this case it is only necessary to capture the change of a field and also to refresh the list of elements in the form when changing and for the user to have a better experience.

Yadier-Tecnativa (changes requested):

Yadier-Tecnativa:
ping @pedrobaeza

sale_product_set/wizard/product_set_add.py
pedrobaeza: Better to put the comment in the affected line instead of the whole file. But if you are saying that is better an onchange, I would say no, it's preferable a filled computed writable

sbejaoui (approved):
Code review + functional test

rousseldenis:
@heliaktiv

pedrobaeza (changes requested):
Previous images are still deleted with no replacements, so the README has lost in quality. Also squash migration commits.

rousseldenis:
@heliaktiv

pedrobaeza:
Superseded by #2582

Issue #2443, opened 2023-03-30
12.0 <-> 15.0 sale_order_invoicing_finished_task : Unable to confirm a Quotation if a sale order line product_uom_qty value is < 0


maisim:
## Module
sale_order_invoicing_finished_task

## Describe the bug
Unable to confirm a Quotation if a sale order have a sale order line with product qty is < 0

![image](https://user-images.githubusercontent.com/6648433/228797684-92de68db-8736-4030-8cfe-be213640b859.png)

In odoo sale addons, order line invoice_status is set  has invoiced if  line.qty_invoiced > line.product_uom_qty

In this case negative value (product_uom_qty) is < 0 (qty_invoiced)
Witch if not the truth (The order had not been invoiced) 
So the bug is more about odoo sale addon behavior but ...

https://github.com/odoo/odoo/blob/5bbbec5a80c2d16219c500aa27d26b7166fb6f7c/addons/sale/models/sale_order_line.py#L821-L822

This part of code is the same from 12.0 to 16.0

## To Reproduce
12.0 is affected, 
It look like to be the same until 15.0

1. create a quotation
2. add a line with product qty < 0
3. Try to confirm the quotation

**Expected behavior**
- Does not consider order line invoiced if it is not true (odoo upstream) wich could have side effect on many addons :/
or
- Add a test to not block confirmation is sale_line.product_uom_qty  <= 0, has we can consider we will not invoice nothing 

The code is actually:
https://github.com/OCA/sale-workflow/blob/34a8045cf3de465ec7c20f40621c7d3c0859f290/sale_order_invoicing_finished_task/models/project.py#L57-L65

We can do it:
```python
    def _check_sale_line_state(self, sale_line_id=False):
        sale_lines = self.mapped('sale_line_id')
        if sale_line_id:
            sale_lines |= self.env['sale.order.line'].browse(sale_line_id)
        for sale_line in sale_lines:
            if (sale_line.state in ('done', 'cancel')
                    or (sale_line.invoice_status == 'invoiced' and sale_line.product_uom_qty > 0)):
                raise ValidationError(
                    _('You cannot create/modify a task related with a '
                      'invoiced, done or cancel sale order line '))
```

an opinion ?

gfcapalbo:
your fix will work well , but the root of the problem seems to be in the behaviour of  sale order, as you have already found:
[...]

PR #2478, opened 2023-04-12
[14.0][FIX] sale_order_invoicing_finished_task: SOL with qty =< 0

gfcapalbo:
[FIX] manage correctly the case of SOL with qty<0,  even if falsely reported as "invoiced"

this error is a fix for the problem sale module has when computing the invoice_status of a sale order line with negative quantities, first found in this issue.

       https://github.com/OCA/sale-workflow/issues/2443

the fix allows the module to behave normally even with SOL's with qty<0, but the root of the problem is how odoo sales computes the invoice status of a SOL , without considering a possible qty<0.

      https://github.com/odoo/odoo/issues/118365

@maisim we have made this 14.0 branch for our internal use, but i bewlieve this should be solved in odoo/addons/sales.
same problem in v16.

### DIFF
--- a/sale_order_invoicing_finished_task/models/project.py
+++ b/sale_order_invoicing_finished_task/models/project.py
@@ -58,9 +58,8 @@ class ProjectTask(models.Model):
         if sale_line_id:
             sale_lines |= self.env["sale.order.line"].browse(sale_line_id)
         for sale_line in sale_lines:
-            if (
-                sale_line.state in ("done", "cancel")
-                or sale_line.invoice_status == "invoiced"
+            if sale_line.state in ("done", "cancel") or (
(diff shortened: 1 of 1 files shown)


gjotten (approved):
#2443 should be fixed by this indeed.

Nasty problem this, since when for example creating a sale order from a project, selecting your 'Timesheets on tasks' product for the project's `sale_line_id` led to the error mentioned because of the 0 qty being written on the SOL.

gjotten:
@NL66278 can we get a review here to see whether or not you're in agreement about this fix, if not how to handle the upstream problem?

NL66278 (approved):
:+1:  LGTM. There still is the possibility to create negative quantity invoice lines, that should IMHO be solved by Odoo, but al least those, and the more general case of quantity 0 invoice lines, no longer will lead to errors.

PR #2490, opened 2023-04-20
[16.0][IMP]sale_cancel_reason: set reason to false when status is != cancel

hugo-cordoba:
I have find the following problem:

When a cancelled order, is moved to draft, it keeps the cancellation reason and this causes that if we group orders by cancellation reasons:

Captura de pantalla de 2023-04-20 17-32-16

we will get orders that are not cancelled, but were once cancelled.

With that change on action_draft, when we move the order to draft, again, the reason for cancellation will be reset.

### DIFF
--- a/sale_cancel_reason/model/sale.py
+++ b/sale_cancel_reason/model/sale.py
@@ -22,6 +22,11 @@ class SaleOrder(models.Model):
                 return True
         return False
 
+    def action_draft(self):
+        res = super().action_draft()
+        self.write({"cancel_reason_id": False})
+        return res
+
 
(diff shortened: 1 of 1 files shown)


aliciagaarzo (approved):

flalexg (approved):
Good change! LGTM!

hugo-cordoba:
Good afternoon @pedrobaeza, I saw this change at the last minute that I had missed, what do you think of the improvement?

pedrobaeza (approved):
/ocabot merge patch

hugo-cordoba:
Thanks @pedrobaeza!

PR #2492, opened 2023-04-26
[14.0][IMP] sale_global_discount, added init hook to avoid memory issues on big database installation

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/sale_global_discount/hooks.py
+++ b/sale_global_discount/hooks.py
@@ -1,0 +1,40 @@
+from odoo.tools.sql import column_exists
+
+
+def _pre_init_global_discount_fields(cr):
+    if not column_exists(cr, "sale_order", "amount_global_discount"):
+        cr.execute(
+            """
+                ALTER TABLE "sale_order"
+                ADD COLUMN "amount_global_discount" double precision DEFAULT 0
+            """
+        )
+        cr.execute(
+            """
+        ALTER TABLE "sale_order" ALTER COLUMN "amount_global_discount" DROP DEFAULT
+        """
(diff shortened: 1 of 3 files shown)


rousseldenis (approved):

hugo-cordoba (approved):
LGTM (Code review)

sale_global_discount/hooks.py
```diff
@@ -0,0 +1,35 @@
+from odoo.tools.sql import column_exists
+
+
+def _pre_init_global_discount_fields(cr):
+    if not column_exists(cr, "sale_order", "amount_global_discount"):
+        cr.execute(
+            """
+                ALTER TABLE "sale_order"
+                ADD COLUMN "amount_global_discount" double precision DEFAULT 0
```
pedrobaeza, line 9: Remove this default after the creation, or it will apply always when no value is set, having potentially side effects:

https://github.com/OCA/openupgradelib/blob/f45dcfe2a3e9661a4104298e9358dedb8aa65b2f/openupgradelib/openupgrade.py#L2811-L2822

pedrobaeza:
It's OK and correct to have `DEFAULT 0` in the add column query. Just drop it afterwards, the same as the link I have put.

ChrisOForgeFlow:
> It's OK and correct to have `DEFAULT 0` in the add column query. Just drop it afterwards, the same as the link I have put.

Sorry, I added default and drop after, thank you

pedrobaeza (approved):
/ocabot merge patch

PR #2496, opened 2023-04-27
[16.0][MIG] sale_order_line_price_history: Migration to version 16.0

Yadier-Tecnativa:
@tecnativa TT42430


Yadier-Tecnativa:
ping @sergio-teruel  , @pedrobaeza

pedrobaeza (changes requested):
JS code should be converted to new OWL

Yadier-Tecnativa:
> JS code should be converted to new OWL

Done in sale_order_line_price_history/static/src/js/sale_line_price_history_widget.js

sale_order_line_price_history/views/sale_views.xml
```diff
@@ -9,15 +9,18 @@
...
                 <field name="order_partner_id" invisible="1" />
-                <widget
-                    name="sale_line_price_history_widget"
-                    width="20px"
+                <field
+                    name="id"
+                    string="Price Widget"
+                    context="{'active_id': id, 'active_ids': [id]}"
                     attrs="{'invisible': [('id', '!=', False)]}"
+                    widget="sale_line_price_history_widget"
```
pedrobaeza, line 16: There's not too much sense to have this column, and more having the button next to it. Right now is taking a lot of space for nothing and confusing user.

sale_order_line_price_history/views/sale_views.xml
```diff
@@ -9,15 +9,18 @@
...
                 />
                 <button
                     name="%(sale_order_line_price_history_action)d"
                     type="action"
-                    icon="fa-clock-o"
+                    icon="fa-history"
```
pedrobaeza: Why changing the icon?

Yadier-Tecnativa: Because this icon represent better the main purpose of this button, show historical prices

pedrobaeza: OK, but the button should go out and it should be inside the widget.

sale_order_line_price_history/views/sale_views.xml
Yadier-Tecnativa: The reason for the widget is because the buttons before saving are disabled. This way it makes more sense to only use the widget for the two options, before saving and after saving the record.

pedrobaeza: Yeah, the problem is that the widget is not doing its work. Talked by chat, so you know what to do.

sergio-teruel (approved):
Functional review ok

pedrobaeza (approved):
/ocabot merge nobump

sergio-teruel:
ping @juangomezzurita

PR #2497, opened 2023-04-28
[16.0][FIX] sale_exception, fixes decoration-danger cases

jok-adhoc:
It happens that exception_ids is always being evaluated as True, so the condition 'not ignore_exception and exception_ids' is not behaving as excepting

### DIFF
--- a/sale_exception/models/sale_order_line.py
+++ b/sale_exception/models/sale_order_line.py
@@ -11,8 +11,9 @@ class SaleOrderLine(models.Model):
     _name = "sale.order.line"
 
     exception_ids = fields.Many2many(
-        "exception.rule", string="Exceptions", copy=False, readonly=True
+        related="order_id.exception_ids", string="Exceptions", copy=False, readonly=True
     )
+    has_exception = fields.Boolean(compute="_compute_has_exceptions", default=False)
     exceptions_summary = fields.Html(
         readonly=True, compute="_compute_exceptions_summary"
     )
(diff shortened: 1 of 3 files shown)


glitchov:
I do apologize for having missed you PR before tackling the same issue.

sale_exception/models/sale_order_line.py
```diff
@@ -11,8 +11,9 @@ class SaleOrderLine(models.Model):
     _name = "sale.order.line"
 
     exception_ids = fields.Many2many(
-        "exception.rule", string="Exceptions", copy=False, readonly=True
+        related="order_id.exception_ids", string="Exceptions", copy=False, readonly=True
```
glitchov, line 14: Why doing this ?
exception_ids should be the exceptions related to the line not to the order. In your case the decoration condition will be True for all SO lines since there is at least one exception on the SO or on a SO line.

glitchov: By the way can you agree with https://github.com/OCA/sale-workflow/pull/2499 and accept this solution ?

glitchov (changes requested):

sale_exception/models/sale_order_line.py
jok-adhoc: Hi @glitchov . That's correct, I made a mistake thinking about it. I agree with #2499 . It's more complete and clean than mine :)

PR #2499, opened 2023-05-03
[FIX] sale_exception : so line decoration-danger condition

glitchov:
The decoration-danger condition of the so line was using exception_ids as True or False depending on the fact there was exception or not. But having an empty Many2many is not evaluated as False in this case and any so line is red right after creation. So decoration-danger depends now on a computed boolean taking the length of exception_ids into account.

### DIFF
--- a/sale_exception/models/sale_order_line.py
+++ b/sale_exception/models/sale_order_line.py
@@ -19,6 +19,14 @@ class SaleOrderLine(models.Model):
     ignore_exception = fields.Boolean(
         related="order_id.ignore_exception", store=True, string="Ignore Exceptions"
     )
+    is_exception_danger = fields.Boolean(compute="_compute_is_exception_danger")
+
+    @api.depends("exception_ids", "ignore_exception")
+    def _compute_is_exception_danger(self):
+        for rec in self:
+            rec.is_exception_danger = (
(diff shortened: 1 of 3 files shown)


rousseldenis:
@jok-adhoc @glitchov You are trying to adress the same problem.

IMHO, this solution is better as the danger visibility is depending on one field only.

That apart, @glitchov I would have added the 'exception' term in field in order to avoid name collisions.

Could you both sync ?

rousseldenis:
@aliciagaarzo Could you check this ?

aliciagaarzo (approved):

jok-adhoc:
Hello @rousseldenis , yes, this solution is better than mine!

rousseldenis:
> Hello @rousseldenis , yes, this solution is better than mine!

@jok-adhoc  So, could you review it ? Thanks

john-herholz-dt (commented):
Had the same problem with a similar solution. Can confirm the code is correct.

john-herholz-dt (approved):

john-herholz-dt:
@guewen are you able to merge? Thanks a lot!

guewen (approved):
Thanks

PR #2501, opened 2023-05-05
[14.0][FIX] sale workflow job email send before validation

TDu:
When using the `sale_automatic_workflow_job` module the validation of
the order is done by a job. But the sending of the confirmation is done
in line. Causing the email to be sent before the validation of the
order.

### DIFF
--- a/sale_automatic_workflow_job/tests/test_auto_workflow_job.py
+++ b/sale_automatic_workflow_job/tests/test_auto_workflow_job.py
@@ -45,6 +45,7 @@ class TestAutoWorkflowJob(TestCommon, TestAutomaticWorkflowMixin):
                     ("state", "=", "draft"),
                     ("workflow_process_id", "=", self.sale.workflow_process_id.id),
                 ],
+                False,
             )
             self.assert_job_delayed(
                 delayable_cls, delayable, "_do_validate_sale_order", args
@@ -150,6 +151,7 @@ class TestAutoWorkflowJob(TestCommon, TestAutomaticWorkflowMixin):
                 args=(
                     self.sale,
                     safe_eval(workflow.order_filter_id.domain) + workflow_domain,
+                    False,
                 ),
             )
             job = trap.enqueued_jobs[0]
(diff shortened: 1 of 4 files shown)


simahawk (approved):

sale_automatic_workflow_job/tests/test_auto_workflow_job.py
```diff
@@ -150,6 +151,7 @@ def test_deleted_sale_order(self):
                 args=(
                     self.sale,
                     safe_eval(workflow.order_filter_id.domain) + workflow_domain,
+                    False,
```
sebalix, line 154:
```suggestion
                    send_order_confirmation=False,
```

sale_automatic_workflow_job/tests/test_auto_workflow_job.py
```diff
@@ -45,6 +45,7 @@ def test_validate_sale_order(self):
                     ("state", "=", "draft"),
                     ("workflow_process_id", "=", self.sale.workflow_process_id.id),
                 ],
+                False,
```
sebalix, line 48:
```suggestion
                send_order_confirmation=False,
```

TDu: I tried but it does not work.

sale_automatic_workflow_job/tests/test_auto_workflow_job.py
simahawk: I didn't ask for this as I feared you would point at me as "nitpicking" :rofl:

sale_automatic_workflow_job/tests/test_auto_workflow_job.py
simahawk: @TDu what do you mean?

TDu: https://github.com/OCA/sale-workflow/blob/4b91bb664d6ca58f7543801bebb8bf753673f1e2/sale_automatic_workflow_job/tests/test_auto_workflow_job.py#L21

sebalix: Indeed, `args` is a tuple/list here, no way to pass kwargs.

sebalix (approved):

hildickethan (approved):

PR #2399, opened 2023-03-06
[16.0][MIG] sale_partner_incoterm

BT-apadron:

PR #2506, opened 2023-05-10
[16.0][MIG] sale_partner_incoterm: Migration to 16.0

hailangvn2023:
* Supersedes https://github.com/OCA/sale-workflow/pull/2399


sale_partner_incoterm/models/sale_order.py
```diff
@@ -14,11 +14,9 @@ class SaleOrder(models.Model):
...
     @api.onchange("partner_id")
     def onchange_partner_id(self):
-        res = super().onchange_partner_id()
         self.update(
             {
                 "incoterm": self.partner_id.sale_incoterm_id.id,
                 "incoterm_address_id": self.partner_id.sale_incoterm_address_id.id,
             }
         )
-        return res
```
ivantodorovich: Why are you no longer calling super?

AFAICS, then this won't be executed anymore, and that'd be wrong: https://github.com/odoo/odoo//blob/b72dc83b6050c6b662d6d85ebb94cdb9709aa433/addons/purchase/models/purchase.py#L294

sale_partner_incoterm/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
+        self.update(
+            {
+                "incoterm": self.partner_id.sale_incoterm_id.id,
+                "incoterm_address_id": self.partner_id.sale_incoterm_address_id.id,
+            }
+        )
```
ivantodorovich:
```suggestion
    @api.onchange("partner_id")
    def _onchange_partner_id_set_incoterm(self):
        self.update(
            {
                "incoterm": self.partner_id.sale_incoterm_id.id,
                "incoterm_address_id": self.partner_id.sale_incoterm_address_id.id,
            }
        )
```

Alternatively, to not call super and to avoid possible conflicts with other modules, I find this approach more "secure".
This way it's pretty much impossible to collide.

Otherwise, `onchange_partner_id` is too generic making the solution fragile: any other module (or even odoo core) adding a method named like this would create a silent incompatibility

hailangvn2023: Thanks @ivantodorovich for the review. Your suggestion has been applied.

ivantodorovich (approved):

mmequignon (approved):
Needs to be squashed.
Pre-approved

hailangvn2023:
Thanks @ivantodorovich and @mmequignon for the approval. This PR has been squashed. Its workflow need approval also.

hailangvn2023:
Thanks @ivantodorovich and @mmequignon for the approvals. This PR has been squashed. Its workflows need approval also.

hailangvn2023:
Hi @mmequignon, the failed test has been fixed with latest commit.

sebalix (approved):

PR #2512, opened 2023-05-12
[FIX] sale_planner_calendar: Filter button closes wizard if is opened from RedirectWarning

carlosdauden:
Filter button closes wizard if is opened from RedirectWarning popup.
Display wizard inline is not possible because v15 try save partner record.
Use of "ir.action.do_nothing" can cause concurrency problems in registry writing.

@Tecnativa TT43060

### DIFF
--- a/sale_planner_calendar/models/res_partner.py
+++ b/sale_planner_calendar/models/res_partner.py
@@ -3,7 +3,7 @@
 from datetime import timedelta
 
 from odoo import _, fields, models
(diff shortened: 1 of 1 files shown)


carlosdauden:
ping @sergio-teruel @CarlosRoca13

CarlosRoca13 (approved):

sergio-teruel (approved):

PR #2518, opened 2023-05-17
[15.0][FIX] sale_order_type: Default value not set for compute fields

sergio-teruel:
"default" key in compute fields (store, writeable) has no set when a record is created with default values in context because the onchange not are triggered.
This is a behavior to increase performance...

cc @Tecnativa TT43405 

ping @carlosdauden @chienandalu


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

Please do the fw-port

PR #1812, opened 2021-12-08
[14.0][IMP] sale_order_general_discount: apply discount product optin

petrus-v:
My customer has hundred lines in sale order, he wants apply discount only on some kind of products according his business rules. As today he set general discoung and move back to zero each lines manually.

The purpose of this PR is to improve `sale_order_general_discount` module in order to add an option on product to define if the general discount should be applied on sale order lines according linked product.

PR #2524, opened 2023-05-23
[16.0][IMP] sale_order_general_discount: apply discount only on gener…

ferran-S73:
…al discount apply product option

Forward port from 
- https://github.com/OCA/sale-workflow/pull/1812

### DIFF
--- a/sale_order_general_discount/README.rst
+++ b/sale_order_general_discount/README.rst
@@ -31,7 +31,9 @@ Sale Order General Discount
 This module allows to set a general discount in a sales order. This general
 discount is set to each line order in the standard `discount` field.
 
-You can also set a default general discount on customers.
+You can configure:
+ * a default general discount on customers
+ * On each product define if general discount is applied
 
 **Table of contents**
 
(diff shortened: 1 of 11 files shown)


NICO-SOLUTIONS (commented):
Funtional review. LG!

petrus-v:
Thanks to have port this work, unfortunately this PR https://github.com/OCA/sale-workflow/pull/2685 has been merged before and do a similar job, so just to tell you that I've closed the initial PR (https://github.com/OCA/sale-workflow/pull/1812) on version 14.0, It would be probably better to port the merged PR (https://github.com/OCA/sale-workflow/pull/2685) to make easier maintenance over versions !

rousseldenis:
Closing this as another PR should be required. see comments

ferran-S73:
Hi @rousseldenis could you please reopen this PR at least while we make another one/update this one? We already have this PR aggregated and are using it in some clients.

Moreover, I'd like to know which approach should be followed to implement this functionality since in v17 it's already merged similarly as we developed here but it's different from v14

PR #2533, opened 2023-05-30
[MIG] sale_order_carrier_auto_assign: Migration to 16.0

matiasperalta1:


florenciafrigieri2 (approved):
Functional review ok

mmequignon (approved):

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+from odoo import models
+from odoo.tools import config
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    def action_confirm(self):
+        if config["test_enable"] and not self.env.context.get(
```
rousseldenis, line 12: @matiasperalta1 @grindtildeath I don't like such code. Better to add a setting to enable the behaviour that breaks standard one. But not blocking

rousseldenis: @matiasperalta1 Could you put at least in ROADMAP we need to add a configuration setting ?

santostelmo: @rousseldenis I mplemented that in another PR:
https://github.com/OCA/sale-workflow/pull/2837

StephaneMangin: Why adding this in an other PR ? Would you not prefer to include it here ?

santostelmo: I would prefer. But probably @matiasperalta1 did not had opportunity to implement that.

Issue #2480, opened 2023-04-16
sale_advance_payment: Negative "zero" residual amount


gaikaz:
Sometimes, the residual amount becomes a "negative zero" when the Sale Order if fully paid.
To be fair, I'm not really sure how to reproduce it, but it's clearly a rounding and/or logic bug with how residual amount is computed.

Module: **sale_advance_payment**
Affected versions: **15.0** (at least)

**Expected behavior**
When the Sale Order get's marked as "PAID", might as well set the residual amount as just zero without actually trying to compute it.

![image](https://user-images.githubusercontent.com/17097810/232308524-4295d13c-a78a-46d0-8c14-dd0af9d3aa4b.png)

Mantux11:
I made a fix here, simple solution, but I think this could prevent `amount_residual` form being with negative zero:

https://github.com/OCA/sale-workflow/pull/2541

yaazkal:
Since this has been merged, I think the issue should be closed.

Regards.

gaikaz:
Thanks @yaazkal 
Sure thing.

PR #2541, opened 2023-06-02
[15.0][FIX] sale_advance_payment: fixed negative amount_residual

Mantux11:
Issue: https://github.com/OCA/sale-workflow/issues/2480

Just tried to make SO with negative zero amount residual, but with no luck. So, seeing this issue, I made some changes:
If SO is in `paid` state, `amount_residual` would be setted to zero, just to be sure, that it not change somehow to minus zero.

### DIFF
--- a/sale_advance_payment/__manifest__.py
+++ b/sale_advance_payment/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Sale Advance Payment",
-    "version": "15.0.1.0.2",
+    "version": "15.0.1.0.3",
     "author": "Comunitea, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "category": "Sales",
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -3,6 +3,7 @@
 
 from odoo import api, fields, models
 from odoo.tools import float_compare
+from odoo.tools.float_utils import float_round
 
 
 class SaleOrder(models.Model):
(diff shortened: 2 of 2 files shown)


mmequignon:
I think we should use float_round on the value before assigning it to `record.amount_residual`
Here: https://github.com/OCA/sale-workflow/blob/15.0/sale_advance_payment/models/sale.py#L97
```python3
amount_residual = float_round(amount_residual, precision_rounding=order.currency_id.rounding)
```
if the value is like `-0.00000000003`, it will round it up to `0.00`.
I think this is enough.
On the other hand, your fix might change `-42.0` to `0.0`, which is not good, IMO.

Mantux11:
@mmequignon I had to add to your suggested line `or 0.01` because if order has currency rounding that equals to `0.0`, it raises an error, that `precision_rounding must be greater than 0.0`. So, to prevent that, if there is no currency rounding on order(it happens on tests, i don't think this is possible in database), it should round to `0.01`. I think this should be okay.

mmequignon:
> @mmequignon I had to add to your suggested line `or 0.01` because if order has currency rounding that equals to `0.0`, it raises an error, that `precision_rounding must be greater than 0.0`. So, to prevent that, if there is no currency rounding on order(it happens on tests, i don't think this is possible in database), it should round to `0.01`. I think this should be okay.

No, there's sql constraints to ensure rounding is > 0.
https://github.com/odoo/odoo/blob/15.0/addons/uom/models/uom_uom.py#L83-L87
I think the way of doing this is not to hardcode values, but to fallback to another currency, maybe?

Mantux11:
@mmequignon Yeah, there is `sql_constraints` for that, bus also in `_float_check_precision` function it checks, if it's greater or not and raises an assertError. It's a good question, which currency would be the best option and truly, I'm not also a fan of hardcoding values. I'll think about to which currency would be the best option to change.

Mantux11:
@mmequignon After more digging into this, just found that Pricelist `default currency` is from company, so I added a Company Currency if there is no currency on the order. And why test was failing(I just keeped my eye on one module `sale_order_qty_change_no_recompute`), just found that when the form is in creation, all compute functions is called, and no matter if after form creation pricelist is set, error raises even before that. Adding `or self.env.company.currency_id` should get currency always in tests and no errors would be raised. Also, if order has his own currency, it would pick up order's currency.

Mantux11:
@gaikaz Could you take a look here?

gaikaz (approved):
Code change looks good.

JordiMForgeFlow:
@Mantux11 would it be possible to squash the commits?

Mantux11:
@JordiMForgeFlow Sure, no problem. Done that.

JordiMForgeFlow (approved):
Code review LGTM!

Mantux11:
@OCA/crm-sales-marketing-maintainers Could anyone take a look to this one? Thanks 👍🏼

PR #2547, opened 2023-06-05
[15.0][MIG] sale_order_note_template

ChristianSantamaria:
Standard migration to v15


ChristianSantamaria:
Already exist in Odoo v15

PR #2549, opened 2023-06-06
[15.0][FIX] Sales order states in tier validation

BT-cserra:
The states for the sales order were outdated for V15.

### DIFF
--- a/sale_tier_validation/models/sale_order.py
+++ b/sale_tier_validation/models/sale_order.py
@@ -7,8 +7,8 @@ from odoo import models
 class SaleOrder(models.Model):
     _name = "sale.order"
     _inherit = ["sale.order", "tier.validation"]
-    _state_from = ["draft", "sent", "to approve"]
-    _state_to = ["sale", "approved"]
+    _state_from = ["draft", "sent"]
+    _state_to = ["sale", "done"]
 
     _tier_validation_manual_config = False
 


BT-cserra:
Hello @sergio-teruel (you did the last commit touching those lines) and @JasminSForgeFlow (you did the migration of `sale_tier_validation` to V15). I've done this PR to adapt the states of a sales order to those of V15. We have been using these states for eight months and it has worked fine for us, but I'd like to know your opinion about it. Could you have a look at it? Many thanks :)

sergio-teruel (approved):

BT-dmoreno (approved):

BT-rmartin (approved):

pedrobaeza (changes requested):
Afaik this is changing the behavior, requiring an approval when going to send the mail. Doing this in a stable version is not acceptable.

Am I right?

If so, you need to add this as a configuration and preserve the old behavior as default

BT-cserra:
> Afaik this is changing the behavior, requiring an approval when going to send the mail. Doing this in a stable version is not acceptable.
> 
> Am I right?
> 
> If so, you need to add this as a configuration and preserve the old behavior as default

Hello @pedrobaeza  The old behaviour didn't work for us. When we investigated it was found that it was because the states listed in the migrated code were no longer valid as states for a sales order. This PR is to adapt the states to those in V15. If the old behaviour works for you, then it means we missconfigured something and that this PR is not needed. So: did/does the old behaviour work for you?

sale_tier_validation/models/sale_order.py
```diff
@@ -7,8 +7,8 @@
 class SaleOrder(models.Model):
     _name = "sale.order"
     _inherit = ["sale.order", "tier.validation"]
-    _state_from = ["draft", "sent", "to approve"]
-    _state_to = ["sale", "approved"]
+    _state_from = ["draft", "sent"]
+    _state_to = ["sent", "sale", "done"]
```
pedrobaeza: For not changing the behavior, I think this is what you have to put:

```suggestion
    _state_to = ["sale", "done"]
```

BT-cserra: > For not changing the behavior, I think this is what you have to put:

Hello @pedrobaeza thank you for your analysis. I was talking about it with other people involved in our project and we think your suggestion is right. I rebased and amended the commit.

pedrobaeza (approved):
OK, thanks for the contribution and the analysis.

There's still a gap, as it can be considered incorrect to send a quotation by mail if the prices are not validated by the corresponding tier, but that's something that you are not preventing adding to `_state_to` the `sent` value, as the mail will be sent anyway.

You can do an extra PR adding that point to the roadmap, or take it into account for future references, and if you want to add some code on that, it would be a great addition to the module.

Meanwhile, let's merge this for fixing the existing problem:

/ocabot merge patch

Can you forward-port it to v16 as well?

BT-cserra:
> OK, thanks for the contribution and the analysis.
> 
> There's still a gap, as it can be considered incorrect to send a quotation by mail if the prices are not validated by the corresponding tier, but that's something that you are not preventing adding to `_state_to` the `sent` value, as the mail will be sent anyway.
> 
> You can do an extra PR adding that point to the roadmap, or take it into account for future references, and if you want to add some code on that, it would be a great addition to the module.
> 
> Meanwhile, let's merge this for fixing the existing problem:
> 
> /ocabot merge patch
> 
> Can you forward-port it to v16 as well?

Hello @pedrobaeza. Of course:
- The roadmap has been added in https://github.com/OCA/sale-workflow/pull/2590.
- The forward port is in https://github.com/OCA/sale-workflow/pull/2589 (I can add there also https://github.com/OCA/sale-workflow/pull/2590 if approved, or create a different forward port).

PR #2553, opened 2023-06-06
[15.0][ADD] sale_fully_invoiced

AaronHForgeFlow:
Split from https://github.com/OCA/sale-workflow/pull/2306 according to this comment: https://github.com/OCA/sale-workflow/pull/2306#pullrequestreview-1218920523

The standard field invoice_status show information about the status invoices related
as closed or not. Because of this, this module introduce two new fields that  will show
account.

cc @ForgeFlow


JuanyDForgeflow (approved):
Functional Review LGTM

sale_fully_invoiced/models/sale.py
```diff
@@ -0,0 +1,35 @@
...
+        ],
+        string="Posted Invoice Status",
+        compute="_compute_invoice_status_validated",
+        store=True,
+        readonly=True,
+        copy=False,
+        default="no",
+    )
+
+    @api.depends("invoice_status")
```
em230418: Missing "invoice_ids.state" in "depends".

AaronHForgeFlow: Good catch. Fixed.

MiquelRForgeFlow (approved):
LGTM

sale_fully_invoiced/tests/test_sale_fully_invoiced.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2022 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.tests import common, tagged
+
+
+@tagged("-at_install", "post_install")
+class TestSaleFullyDelivered(common.TransactionCase):
+    def setUp(self):
```
rousseldenis: @AaronHForgeFlow Could you change this to classmethod ?

rousseldenis (approved):
LGtTM. Minor comment

sale_fully_invoiced/tests/test_sale_fully_invoiced.py
AaronHForgeFlow: @rousseldenis did that :) sorry for the delay

PR #2554, opened 2023-06-06
[16.0] [MIG] sale_payment_sheet

angelmoya:


sale_payment_sheet/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Sale payment sheet",
     "summary": "Allow to create invoice payments to commercial users without "
     "accounting permissions",
-    "version": "15.0.1.1.2",
+    "version": "16.0.1.1.2",
```
FrancoMaxime: Hi, you should restart the version from 16.0.1.0.0
```suggestion
    "version": "16.0.1.0.0",
```

sale_payment_sheet/tests/test_sale_payment_sheet.py
```diff
@@ -76,7 +91,7 @@ def _create_payment_sheet(self):
                 with sheet_form.line_ids.new() as line_sheet:
                     line_sheet.partner_id = self.partner
                     line_sheet.invoice_id = invoice
-                    line_sheet.ref = "REF{}".format(line_sheet.id)
+                    # line_sheet.ref = "REF{}".format(line_sheet.id)
```
FrancoMaxime: Why this line is commented?

FrancoMaxime (changes requested):
Hi, code LGTM, minors comment

sale_payment_sheet/tests/test_sale_payment_sheet.py
angelmoya: Ref in view is not accesible, with group="base.group_no_one", in this version we can't get this field, for other fields I add full accounting features to the user, but I don't know why for this field and group thid doesn't work

sale_payment_sheet/__manifest__.py
angelmoya: ok fixed

sale_payment_sheet/tests/test_sale_payment_sheet.py
angelmoya: I just delete instead of coment

aliciagaarzo (approved):

FrancoMaxime (approved):
LGTM, thanks

sale_payment_sheet/tests/test_sale_payment_sheet.py
```diff
@@ -60,18 +61,30 @@ def _create_invoice(self):
                 line_form.price_unit = 100.00
         return invoice_form.save()
 
+    def _create_refund(self):
+        with Form(
```
pedrobaeza, line 65: Don't do `with` here, as you are saving 2 times with it (one exiting the context and the other on the explicit call).

pedrobaeza (commented):
Please squash your last commit with the migration one, indicating that it also improves tests, or put the version change in the migration commit if you want to preserve both commits.

sale_payment_sheet/tests/test_sale_payment_sheet.py
rousseldenis: @angelmoya

pedrobaeza:
@carolinafernandez-tecnativa please supersede this PR fixing my last commit to proceed with the merge.

pedrobaeza:
Superseded by #2699

Issue #2551, opened 2023-06-06
[16] sale_order_type - Error when posting an expense


elarapuntsistemes:
**## Module**
sale_order_type

**## Describe the bug**
Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 984, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 17

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1160, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 991, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.move(17,).company_currency_id'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 984, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 44

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1160, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 991, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.move.line(44,).compute_all_tax'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 984, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 44

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1160, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 991, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.move.line(44,).currency_rate'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1809, in dispatch
[...]

PR #2557, opened 2023-06-07
[FIX] sale_order_type: Error when hr-expense is installed

igallart:
https://github.com/OCA/sale-workflow/issues/2551

### DIFF
--- a/sale_order_type/models/res_currency.py
+++ b/sale_order_type/models/res_currency.py
@@ -1,0 +1,10 @@
+from odoo import api, models
+
+
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    @api.model
+    def _get_conversion_rate(self, from_currency, to_currency, company, date):
+        from_currency = from_currency or company.currency_id
+        return super()._get_conversion_rate(from_currency, to_currency, company, date)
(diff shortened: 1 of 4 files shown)


sale_order_type/models/res_currency.py
```diff
@@ -0,0 +1,10 @@
+from odoo import api, models
+
+
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    @api.model
+    def _get_conversion_rate(self, from_currency, to_currency, company, date):
+        from_currency = from_currency or company.currency_id
```
rousseldenis, line 9: @igallart Mmmh, this seems to be fixed elsewhere than here

rousseldenis: I would have searched first why account.move.line.company_currency_id is not filled in there: 

https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move_line.py#L636

igallart: @rousseldenis I did not find the cause of the problem, but if you do exactly what it says in [#2551](https://github.com/OCA/sale-workflow/issues/2551) the problem occurs. For some strange reason, the field line.company_currency_id (related) is empty and fails to do the division.

rousseldenis: I think that's because the field on account.move.line is a related stored field on account.move one which is not stored.

antoniocanovas (approved):
Functional review Ok.

elarapuntsistemes (approved):
Funtional test ok

PR #2558, opened 2023-06-07
[16.0][FIX] sale_order_type: Default value not set for compute fields

sbiosca-s73:
When create a sales order by accessing from the magic button with Sales in **res_partner**, it correctly sets the **sale_type** value of **res_partner** to the **type_id** of the sales order.

I changed the default method (**_default_type_id**) of field **type_id** in sale order.


ferran-S73 (approved):

ferran-S73:
You should consider this approach instead https://github.com/OCA/sale-workflow/commit/ee47bbc89c8f1075199d97f23a5d87b33cc04c2a

sbiosca-s73:
@sergio-teruel ping

pedrobaeza:
Let's fast-track this as already merged in previous versions:

/ocabot merge patch

PR #2561, opened 2023-06-07
[16.0][MIG] sale_stock_cancel_restriction

aliciagaarzo:
Migration of **sale_stock_cancel_restriction** module to version 16


aliciagaarzo:
In v16, a wizard is used to cancel orders and the tests fail because I do not include a dependency on the season_id field of the sale_cancel_reason module
```
        wizz = self.sale_order.action_cancel()
        wizz = self.env["sale.order.cancel"].create({"order_id": self.sale_order.id})
        wizz.action_cancel()
        self.assertEqual(
           self.sale_order.picking_ids.state,
           "cancel",
            "After cancelling a picking, the state should be cancelled",
        )
```

dsolanki-initos (approved):
Functional LGTM.

hugo-cordoba (approved):
LGTM! (Code Review)

PR #2485, opened 2023-04-18
[15.0][MIG] sale_fixed_discount

ps-tubtim:
From https://github.com/OCA/sale-workflow/pull/2269

PR #2574, opened 2023-06-19
[15.0][MIG] sale_fixed_discount

CRogos:
Followup of https://github.com/OCA/sale-workflow/pull/2485
- migration based on v14
- merged commits


ao-landoo:
When creating the invoice from the sale order, invoice's total tax value is not considering the fixed discount. You need to set it's value on the line to 0 and set it's value again

https://watch.screencastify.com/v/Q9Q3IkXpvKHAoyiizsTG

CRogos:
I totally agree with this issue but I assume this has do be solved in account_invoice_fixed_discount
There is also a ticket regarding this issue: https://github.com/OCA/account-invoicing/issues/1107

ao-landoo (approved):
You're right

sale_fixed_discount/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+        def compute_taxes(order_line):
+            real_price = order_line.price_unit * (
+                1 - (order_line.discount or 0.0) / 100.0
+            ) - (order_line.discount_fixed or 0.0)
+            twicked_price = real_price / (1 - (order_line.discount or 0.0) / 100.0)
+            price = (
+                order_line.price_unit * (1 - (order_line.discount or 0.0) / 100.0)
+                if not order_line.discount_fixed
+                else twicked_price
+            )
```
CRogos: Does anyone understand this code?

I would say this line would be fine:
price = order_line.price_unit * (1 - (order_line.discount or 0.0) / 100.0) - (order_line.discount_fixed or 0.0)

CRogos: @pedrobaeza do you have an opinion on this? 
Using the UI you can only insert discount or discount_fixed. So my simplification should be sufficient.

Via the API discount and discount_fixed might be possible and the current implementation of twicked_price is like this:
twicked_price = (100€ * 0,9 - 20€) / 0.9 = 77,78€ 
While my suggested calculation would be:
price = 100€ * 0,9 - 20€ = 70€ which would be an addition of the discounts. 

I also have my concerns that the tax is calculated correctly when discount and discount_fixed is provided.
https://github.com/OCA/account-invoicing/issues/1107

pedrobaeza: Add then a constraint (SQL or Python) for not being able to save both fields with value.

CRogos: Now you are saying it, there is already such a constraint.

pedrobaeza: Then no problem, isn't it? :)

CRogos: @pedrobaeza everything is done from my side. Would you like to review / merge this PR?

pedrobaeza:
Merging blindly according what have been talked:

/ocabot migration sale_fixed_discount
/ocabot merge nobump

pedrobaeza:
Cancelling...

What happens with https://github.com/OCA/sale-workflow/pull/2485 ?

CRogos:
There was no reaction and no progress. #2485 is also migrated based on v13 while this PR is based von the v14 branch. 
This PR supersede #2485

PR #2582, opened 2023-06-23
[16.0][MIG] sale_product_set

CRogos:
follow up of https://github.com/OCA/sale-workflow/pull/2477
- used oca-port to migrate based on v15
- squashed commits


CRogos:
Could you review this PR because there is no progress in https://github.com/OCA/sale-workflow/pull/2477 
@sbejaoui , @Yadier-Tecnativa , @heliaktiv

Yadier-Tecnativa (approved):
Code reviewed and tested in runboat TT43246

pedrobaeza:
Thanks for resuming this. This doesn't seem to have new screenshots to reflect a more modern UI. Maybe you can do such screenshots. Another pending thing that can be good is to split the module into 2: `product_set`, for defining the base models and menus in neutral places (like settings part), that can be hosted in OCA/product-attribute or OCA/product-pack, and `sale_product_set`, adding the menu in sales part, and the wizard. That way, we don't have to install `sale` if for example we only want the sets in projects.

CRogos:
I've update the screenshots. 

I also agree on the other changes, but shall we do this in a separate PR and clarify thigs like "OCA/product-attribute or OCA/product-pack" in an issue first?

pedrobaeza:
What you prefer, but I think they are very clear and won't generate debates.

lmignon (approved):
LGTM (Code review + functional)

CRogos:
@pedrobaeza I've added your comment to the roadmap. Could you merge the PR?

pedrobaeza:
On hold as you have made already the split. Thanks!

CRogos:
@lmignon , @Yadier-Tecnativa could you also review the split PR linked above?

CRogos:
> ding the menu in sales part, and the wizard. That way, we don't have

@pedrobaeza As always there is more to change than expected ;) . Every change is it worth, but I would prefer to commit this migration as is and merge the other two PR when proper reviewed/tested.

simahawk:
@CRogos shall we merge this one first then?

CRogos:
> @CRogos shall we merge this one first then?

Yes, if you don't mind. The other PR, is based on this branch, anyway.

simahawk:
As long as you handle the transition to the new version in those PR I'm fine with merging this.
I'll wait some more hours to see if anybody has objections than I'll merge :wink:

pedrobaeza:
There's one last comment on the split PR, and we can merge it, so no I think there's no need to start merging this one, but as you prefer.

simahawk:
The 1st split PR is getting merged, we can close this one. Thanks @CRogos

lmignon:
The initial pr has been submitted on Apr 12. It has been installed by a lot of people.  I didn't see any migration script to manage the transition....

CRogos:
> The initial pr has been submitted on Apr 12. It has been installed by a lot of people. I didn't see any migration script to manage the transition....

Can you be more detailed what transition you have in mind?
sale_product_set = sale_product_set + product_set 
product_set is installed as a dependency listed in the manifest file.

pedrobaeza:
Being in PR, there shouldn't be any commitment to support a migration path if someone has installed on its own the content of the non merged PR, don't you think? It's good to have it if someone wants to contribute it, but this can't be demanded to the migration + split contributor. But as already said by @CRogos, this one should be smooth, as it's only a new dependency.

lmignon:
Not fair.... I've to make a first PR on product_set since all the translations are lost... (No test possible into the UI before the merge). When table definition are moved from one addon to another one, you should instruct odoo by writing a migration script....  Now I've extra work also on my side since you requested to make the split...

CRogos:
> Not fair.... I've to make a first PR on product_set since all the translations are lost... (No test possible into the UI before the merge). When table definition are moved from one addon to another one, you should instruct odoo by writing a migration script.... Now I've extra work also on my side since you requested to make the split...

I agree that without migration script, customized translation might get lost. But still unclear to me, how a migration script should look like on a split scenario. I did a lot of merge scripts, which is easy ;-) .

To recover the translations shouldn't it be suffient to replace sale_product_set by product_set in the PO files?

e.g.: replace
```
#. module: sale_product_set
#: model:ir.model.fields,field_description:sale_product_set.field_product_set__active
#: model:ir.model.fields,field_description:sale_product_set.field_product_set_line__active
msgid "Active"
msgstr ""
```

with

```
#. module: sale_product_set
#: model:ir.model.fields,field_description:product_set.field_product_set__active
#: model:ir.model.fields,field_description:product_set.field_product_set_line__active
msgid "Active"
msgstr ""
```

I am so happy that you all give feedback so quickly, and this issue is not stretched over months like it is sometimes the case :) .
I think together get a good result on this.

simahawk:
@CRogos look around for `pre-migrate` scripts and openupgradelib.

You'll find many examples like these (just random here):

https://github.com/OCA/community-data-files/blob/c7712984d978a28c97d4b2afa43441a380a8be84/l10n_eu_product_adr/migrations/14.0.1.0.0/pre-migration.py#L59

https://github.com/OCA/helpdesk/blob/42b775ebfef0286879cd6d6150d32b9a90775fb4/helpdesk_mgmt/migrations/15.0.1.3.1/pre-migrate.py#L16

Docs: https://oca.github.io/openupgradelib/API.html#openupgradelib.openupgrade.update_module_moved_fields

lmignon:
Indeed, you should at least update the translation files.

For the models previously declared in sale_product_set and now in
product_set you can use the method
https://oca.github.io/openupgradelib/API.html#openupgradelib.openupgrade.update_module_moved_models
This call should should be done in a pre_init_hook method in the
product_set addon. (
https://www.odoo.com/documentation/16.0/developer/reference/backend/module.html?highlight=pre_init_hook,
example:
https://github.com/OCA/wms/blob/16.0/stock_available_to_promise_release/hooks.py
,
https://github.com/OCA/wms/blob/16.0/stock_available_to_promise_release/__manifest__.py
)
Thank you for the follow up!

pedrobaeza:
No migration script is needed for moved fields since several versions (not sure if since v12 or v13). The po files will be auto-updated (through pomerge or msgmerge - not remember the exact tool name) on ocabot merge, so no need to do anything on that part, just speed up the merge.

PR #2584, opened 2023-06-27
[15.0][ADD] sale_subscription_oca

carlos-domatix:
This module allows creating subscriptions that generate recurring invoices or orders. It also enables the sale of products that generate subscriptions.


pedrobaeza:
Hi, Carlos.

If this is generating invoices, it's a bit awkward that is named `sale_subscription`. And a better repository should be `account-invoicing` or `contract` for this.

carlos-domatix:
Hi, Pedro.
This generates invoices and orders, as the user prefers. It also allows selling products that generate subscriptions.
Can you tell me the name you would give it? I can change it to another repository, but it's also a sales flow.

pedrobaeza:
I would just call it then `subscription_oca`, and OCA/contract can be the host.

carlos-domatix:
Thanks, Pedro! Done!

https://github.com/OCA/contract/pull/971

PR #2590, opened 2023-06-28
[15.0][IMP] ROADMAP for sale_tier_validation

BT-cserra:
As suggested by PR https://github.com/OCA/sale-workflow/pull/2549

### DIFF
--- a/sale_tier_validation/readme/ROADMAP.rst
+++ b/sale_tier_validation/readme/ROADMAP.rst
@@ -1,0 +1,4 @@
+* The sales order, when moved to the state `sent`, will still
+  send the email even if the validation is not approved by the
+  corresponding tier. Code to consider this particular case is
+  not developed.


pedrobaeza:
Please put this roadmap in 16.0 directly, and another thing that should be homogenized is the commit message:

- It shouldn't contain the Odoo version (but the PR title should).
- It should contain the module name.

so, the commit message for this one should be:

```
[IMP] sale_tier_validation: Add in ROADMAP sent validation problem
```

pedrobaeza:
The ROADMAP text is good by the way.

BT-cserra:
> Please put this roadmap in 16.0 directly, and another thing that should be homogenized is the commit message:
> 
> * It shouldn't contain the Odoo version (but the PR title should).
> * It should contain the module name.
> 
> so, the commit message for this one should be:
> 
> ```
> [IMP] sale_tier_validation: Add in ROADMAP sent validation problem
> ```

Hello @pedrobaeza Thank you for your instructions. I followed them in https://github.com/OCA/sale-workflow/pull/2593.

I guess I can close this PR, right?

PR #2594, opened 2023-07-06
[15.0][FIX] sale_order_type: Fix in _compute_sale_type_id method

mem-adhoc:
Ticket 63072

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -10,7 +10,6 @@ class AccountMove(models.Model):
     sale_type_id = fields.Many2one(
         comodel_name="sale.order.type",
         string="Sale Type",
-        compute="_compute_sale_type_id",
         store=True,
         readonly=True,
         states={"draft": [("readonly", False)]},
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please indicate steps to reproduce the problem in a standard installation and a regression test would be interesting.

rousseldenis:
@mem-adhoc Could you answer @pedrobaeza comments and fix tests ? Thanks

mem-adhoc:
Hello @pedrobaeza, i'm still working in the pr, is not finished yet !

pedrobaeza (changes requested):
Changing the computed again to old onchange is not correct. The fix should be done other way.

rousseldenis:
@mem-adhoc What's the status of this ?

mem-adhoc:
Hello @pedrobaeza, how are you? I hope you doing fine! 

The change is because the field was causing a conflict when creating an invoice. The type of sale was not set correctly while the invoice did not have a partner or even having one and selecting the different types of sales, it never changed. Even when having a type of sale set for a specific partner, it also caused conflicts if it was added to an invoice with another type of sale configured. 

The behavior of this field should not be a compute, because it should not be constantly calculating. It should have a default value, which would change depending on whether the partner selected has a set type of sale, or is in favor of what the client decides! 

Greetings.

pedrobaeza:
That's not correct. Computed writable fields are replacing onchanges. You just need to put the corresponding `if` that you consider, but without writing specific steps to reproduce the problem and fixing the tests, this change won't be accepted.

mem-adhoc:
Hello @pedrobaeza, how are you?

This is the video to show you how the code works normally without any changes. As you will see, not having any partner set works wrong. Having a partner set also works wrong since the type of sale is never shown on the screen. https://drive.google.com/file/d/1zFlI99AGK4lCdeBnPSYzY8kA06kKfbYT/view

If I put the if so that the type of sale is not overwritten if it is already set by default, it still works wrong when there is no partner since it finds the first type of sale that there is https://drive.google.com/file/d/1xbT3_bdvr9ur1OhwmrBmkrLx0Wn_O9za/view

pedrobaeza:
Well, please start first getting a green CI, as the fail means that not everything is done correctly:

```
 2023-09-04 17:00:13,049 259 ERROR odoo odoo.addons.sale_order_type.tests.test_sale_order_type: FAIL: TestSaleOrderType.test_invoice_without_partner
  File "/__w/sale-workflow/sale-workflow/sale_order_type/tests/test_sale_order_type.py", line 188, in test_invoice_without_partner
    self.assertEqual(invoice.sale_type_id, self.default_sale_type_id)
AssertionError: sale.order.type() != sale.order.type(1,)
```

mem-adhoc:
Hello @pedrobaeza, how are you? 

I already have the green CI from the tests. I corrected a unit test according to the new logic of this field.

Greetings.

pedrobaeza:
Sorry, your changes are not correct. You should keep the computed nature of the field. If there's a corner case, model it inside the compute method.

rousseldenis:
Closing this as very old and user is not still active. Please feel free to reopen it if needed.

Issue #1859, opened 2022-01-10
sale_product_set: Split sale/product_set


rousseldenis:
@simahawk I think I saw a comment somewhere asking the split of "product_set" feature in its own module to avoid pulling the sale if we use it elsewhere.

simahawk:
done on https://github.com/OCA/sale-workflow/pull/2604

PR #2604, opened 2023-07-13
[16.0][UPD] sale_product_set: Extracted product_set

CRogos:
Splitting sale_product_set into product_set and sale_product_set as suggested by @pedrobaeza in https://github.com/OCA/sale-workflow/pull/2582

I've left the translations in both modules. Is there a easy way to clean these up what belongs to product_set and what to sale_product_set?

Addresses https://github.com/OCA/sale-workflow/issues/1859
Depends on  https://github.com/OCA/product-attribute/pull/1384


simahawk:
@CRogos build is :red_circle: 
I guess you forgot to add the pending dependency in test-requirements.txt

lmignon (changes requested):
Migration scripts are missing

CRogos:
> Migration scripts are missing

I need more info what should be in the migration script. 
product_set should be installed by the manifest dependencies. 
Is there a problem with the data when migrating?

sale_product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,128 @@
...
+                continue
+            yield set_line
+
+    def prepare_sale_order_line_data(self, set_line, max_sequence=0):
+        self.ensure_one()
+        line_values = set_line.prepare_sale_order_line_values(
+            self.order_id, self.quantity, max_sequence=max_sequence
+        )
+        sol_model = self.env["sale.order.line"]
+        line_values.update(sol_model.play_onchanges(line_values, line_values.keys()))
```
CRogos: Is the onchange_helper part still required or is there a better way to handle this in odoo 16?

pedrobaeza: `price_unit` and `discount` fields are now computed writable fields, so there's no need for onchange helper nor any other trick. They should be recomputed on the fly. Please try and check if the price is put according the pricelist, but I bet it will do.

pedrobaeza:
> Migration scripts are missing

Check my comment at https://github.com/OCA/sale-workflow/pull/2582#issuecomment-1639954987

sale_product_set/wizard/product_set_add.py
CRogos: Looks good in my opinion. Removed "onchange_helper" dependency.

CRogos:
@lmignon can you check your translation concerns with this database? I have had a look on the German translation and everything looked good to me. 
As Petro said, the PO files were fixed during the merge. Therefore a test with the code of my branch might lead to an false test result. 

Currently I see no need for a migration script. Maybe as mentioned before some custom translations might not be migrated correctly for textkeys which moved from sale_product_set to product_set. But as the most UI element stay in sale_product_set and only very few people will actually change these text keys, I think it is fine without script.

CRogos:
@Yadier-Tecnativa could you review this as well?

CRogos:
@lmignon and @pedrobaeza can we merge and finish this PR?

pedrobaeza (approved):
Thanks!

lmignon (approved):

simahawk:
Cool! Thank you all for your collaboration :)
#ILoveOpenSource :heart: 

/ocabot merge nobump

PR #2606, opened 2023-07-13
[15.0][IMP] sale_invoice_plan: add select type down payment of invoice plan

Saran440:
This PR implement module `sale_invoice_plan` for user can select type of down payment

Normally, when user use invoice plan and has advance. it will use down payment with type percentage only.

Example Case
- User create sale order 1 line with product A 100us and there is tax 7%
- Use invoice plan and advance 10%
- When create invoice by plan. Down payment will compute by percent and get value = 10.7 (include tax)

Some case user need down payment exclude tax. So, I add new field advance_type for user selected down payment with Fixed amount or percent

Selection_003

### DIFF
--- a/sale_invoice_plan/wizard/sale_make_planned_invoice.py
+++ b/sale_invoice_plan/wizard/sale_make_planned_invoice.py
@@ -19,8 +19,12 @@ class SaleAdvancePaymentInv(models.TransientModel):
         for plan in invoice_plans.sorted("installment"):
             makeinv_wizard = {"advance_payment_method": "delivered"}
             if plan.invoice_type == "advance":
-                makeinv_wizard["advance_payment_method"] = "percentage"
-                makeinv_wizard["amount"] = plan.percent
+                if sale.advance_type == "percentage":
+                    makeinv_wizard["advance_payment_method"] = "percentage"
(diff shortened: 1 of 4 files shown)


sale_invoice_plan/wizard/sale_make_planned_invoice.py
```diff
@@ -19,8 +19,12 @@ def create_invoices_by_plan(self):
         for plan in invoice_plans.sorted("installment"):
             makeinv_wizard = {"advance_payment_method": "delivered"}
             if plan.invoice_type == "advance":
-                makeinv_wizard["advance_payment_method"] = "percentage"
-                makeinv_wizard["amount"] = plan.percent
+                if sale.advance_type == "percentage":
+                    makeinv_wizard["advance_payment_method"] = "percentage"
+                    makeinv_wizard["amount"] = plan.percent
+                else:
+                    makeinv_wizard["advance_payment_method"] = "fixed"
```
rousseldenis, line 26: @Saran440 Could you add the test case for fixed ?

Saran440:
@rousseldenis Fixed. Add test script for this PR. THank you for your review

PR #2603, opened 2023-07-13
[14.0][IMP] sale_order_line_sequence: module refactor, new sequence fields for orders and invoices

OriolMForgeFlow:
Module refactor

PR #2607, opened 2023-07-14
[14.0][MIG] sale_stock_line_sequence

OriolMForgeFlow:
Standard migration
Taking into account the refactor done in [!2603](https://github.com/OCA/sale-workflow/pull/2603)


AaronHForgeFlow (commented):
Edit: nevermind, is the refactor. Can't do functional test in runboat

AaronHForgeFlow (commented):
Code review LG

rousseldenis:
@OriolMForgeFlow Thanks for this. FYI, you can embed your dependency PR here in order to get tests running and runboat for functional testing.

https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

rousseldenis:
@OriolMForgeFlow @AaronHForgeFlow 

I've tested on runboat and if the order of lines change should change the order of invoice lines, that does not work:

Sale Order after change:

image

Invoice after sale order change:

image

rousseldenis:
> @OriolMForgeFlow @AaronHForgeFlow
> 
> I've tested on runboat and if the order of lines change should change the order of invoice lines, that does not work:
> 
> Sale Order after change:
> 
> image
> 
> Invoice after sale order change:
> 
> image

Oops wrong PR. @OriolMForgeFlow This is for #2603

OriolMForgeFlow:
Thank you for your feedback, @rousseldenis!

The lines in the invoice remain unchanged as we are not utilizing the sequence field, which was causing some issues. Instead, we have introduced a new field called "related_so_sequence," designed as a Char field (SO12345/1 - so_name/sequence). This choice is based on the consideration that an invoice can be generated from multiple Sales Orders. As a result, we cannot use this field to sort the order lines.

Here's an example:

image

AaronHForgeFlow (approved):
Functional review, Works as expected.

LoisRForgeFlow (approved):
LGTM!

rousseldenis (approved):
Code review

PR #2608, opened 2023-07-14
15.0 stock picking on hold

alihashaam:


alihashaam:
CI is failing because of another module "sale_order_discount_invoicing", for which tests are failing

rousseldenis:
@alihashaam Thanks for this.

IMHO, this has same functionality as https://github.com/OCA/sale-workflow/tree/15.0/sale_stock_picking_blocking

As done in v16, the blocking reason is now computed stored. So, your module could depends on invoice_status field, then, assign the blocking reason.

https://github.com/OCA/sale-workflow/blob/16.0/sale_stock_picking_blocking/models/sale_order.py#L33

What do you think ?

ortlam:
@rousseldenis Thank you for the suggestion, it makes sense to base this module on `sale_stock_picking_blocking`.

jans23:
I don't think that test failure is related to this module stock_picking_on_hold.

rousseldenis:
@alihashaam What's the status of this ?

jans23:
@rousseldenis alihashaam isn't working on this anymore. I added missing tests. We use this module in production since more than a year and it's working fine. IMO it can be merged.

stock_picking_on_hold/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "Sale Payment Method - Hold Pickings (until payment)",
+    "version": "15.0.1.0.0",
+    "website": "https://github.com/OCA/sale-workflow",
+    "category": "Warehouse",
+    "summary": "This module allows to hold the picking until the invoice is paid",
+    "depends": [
+        "account_payment_mode",
+        "sale_stock_picking_blocking",
+        "website_sale",
```
rousseldenis: IMHO, this is a too restrictive dependency as people do not want necessarily install website to get this functionality.

That part of code should be externalized in another module.

rousseldenis (changes requested):

rousseldenis:
@alihashaam @jans23 Could you fix tests ?

jans23:
@rousseldenis Tests are fixed now.

PR #2609, opened 2023-07-17
[14.0] [ADD] New module sale_delivery_date_by_zone

HekkiMelody:
The goal of this module is to create delivery zones and then automatically have those settings on partners when their delivery zone is set.


rousseldenis (approved):
Code review

sale_delivery_date_by_zone/models/partner_delivery_zone.py
```diff
@@ -0,0 +1,63 @@
...
+        " will be postponed to the next warehouse cutoff time\n"
+        "* Use partner's cutoff time: Delivery order for sale orders"
+        " will be postponed to the next partner's cutoff time\n\n"
+        "Example: If cutoff is set to 09:00, any sale order confirmed before "
+        "09:00 will have its delivery order postponed to 09:00, and any sale "
+        "order confirmed after 09:00 will have its delivery order postponed "
+        "to 09:00 on the following day.",
+    )
+
+    delivery_time_preference = fields.Selection(
```
jbaudoux, line 29: This is to define the partner opening hours.
What's the purpose on a delivery zone?

HekkiMelody: This is the default to apply on partners when the delivery zone is selected for them.

In the use case of our customer, it's rare that partners have specific time needs, and they're much more constrained by the courier for that specific zone; and it's tiresome and time consuming to have to fill the info manually on each partner when it's almost always the same within the same delivery zone.

jbaudoux: Do you use that to schedule the work in the warehouse? Because this was the purpose of defining those values.

Can you elaborate how this is further used in your use case?

Why don't you set on the partner a new selection on order_delivery_cutoff_preference like partner_zone_cutoff and on delivery_time_preference like delivery_zone_window? Those would then read the values from the delivery zone instead of the partner.

HekkiMelody: > Do you use that to schedule the work in the warehouse? Because this was the purpose of defining those values.

I use it to schedule sales orders and their deliveries. There are very few instances of a partner having very specific opening hours that are outside the norm, in our use case. On the other hand, my customer wants to group deliveries by area so that the courier serves that area only once or twice a week.

All the modules work fine for that purpose, with the only limitation of having to set the information manually for every single partner.

> Why don't you set on the partner a new selection on order_delivery_cutoff_preference like partner_zone_cutoff and on delivery_time_preference like delivery_zone_window? Those would then read the values from the delivery zone instead of the partner.

I tried a proof of concept of that; unfortunately it requires rewriting at least 3 modules so they support taking the information from one source or another. As they currently are, `sale_delivery_date` and its dependencies are all hardcoded to look at the partner.

I would have to create a sort of "delivery.preferences.mixin" that implements an interface with fields such as `delivery_time_preferences` and `delivery_time_window_ids`, and related methods, so that the source of delivery preferences could be chosen based on a selection field or something of the sort. It's not impossible but it's a lot of work and I don't even know if it would get approved and how long it'd take for it to be merged.

This is a fairly simple solution that just pre-populates some default fields, and addresses a good 80/90% of my use case.

rousseldenis: @jbaudoux What's the conclusion of this ?

jbaudoux: I don't like that a partner constraint (time window) is used to express a delivery method constraint. But there is no clean solution for this in v14, so if this workarounds the problem in your use case, why not.
I have a working POC in v16 combining all possible constraints (partner opening time, warehouse opening time & working days, cutoff, delivery zones covered days, bank holidays by country/states). It's on my todo list for the coming months to finish it.

rousseldenis:
@jbaudoux Could you approve this ?

PR #2614, opened 2023-07-21
[16.0][MIG] sale_fixed_discount: Migration to 16.0

PieterPaulussen:


sale_fixed_discount/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Sale Fixed Discount",
+    "summary": "Allows to apply fixed amount discounts in sales orders.",
+    "version": "16.0.1.0.0",
+    "category": "Sales",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["sale", "account_invoice_fixed_discount"],
```
FrancoMaxime, line 13: Hi, this migration is linked to another pull request, you should make a link to this request by adding it in the requirement.txt file.  You can take a look at this documentation https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29 to have more information about the procedure

FrancoMaxime (changes requested):

sale_fixed_discount/__manifest__.py
PieterPaulussen: @FrancoMaxime Thanks for the feedback. Didn't know about the `test-requirements.txt`. Appreciated and added the requirement. I'm first going to fix up the dependency module though prior to completing this one. Going to mark it as WIP.

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo-addon-sale_fixed_discount @ git+https://github.com/OCA/account-invoicing.git@refs/pull/1488/head#subdirectory=setup/sale_fixed_discount
```
rousseldenis:
```suggestion
odoo-addon-account-invoice-fixed-discount @ git+https://github.com/OCA/account-invoicing.git@refs/pull/1488/head#subdirectory=setup/account_invoice_fixed_discount
```

@PieterPaulussen This line is intended to install the module that is not yet merged. So, account_invoice_fixed_discount.

CRogos: @PieterPaulussen you still need to adapt the module name at the end of the string. I think you only changed the beginning. 

Good work so far.. looking forward to do the testing/review.

PieterPaulussen: ~~I've refactored the flow of this module to align it with the new implementation of `account_invoice_fixed_discount` and the option arises to remove the requirement of the module entirely as both features can be used as stand-alone. The only problem is that the `discount_fixed` value does not get updated when invoicing the order.
Perhaps you know of a better way like an optional dependency or a new module entirely?~~

sale_fixed_discount/models/sale_order_line.py
```diff
@@ -14,45 +15,43 @@ class SaleOrderLine(models.Model):
...
+                    raise ValidationError(
+                        _(
+                            "The fixed discount %(fixed)s does not match the calculated "
+                            "discount %(discount)s %%. Please correct one of the discounts."
+                        )
+                        % {
+                            "fixed": line.discount_fixed,
+                            "discount": line.discount,
+                        }
+                    )
```
CRogos: In v15, only discount or discount_fixed was allowed. In the calculation they where implemented additive (even though you could not enter both, and also the further tax calculation wouldn't work properly.)

In v16 they are treated as one and will be both be set. This leads also to a different visualization in the PDF documents, because now both (fixed and %) is added to the PDF:
image
I think the change is not perfect because of the PDF output. 

I am not sure if we should allow fixed + % in future, when we get the tax calculation to handle this correctly. In the v15 tax calculation this wasn't easy to implement, but with the v16 changes this might be possible.
But when we allow both, we need to decide if we use:
price = (unit_price - fixed) * discount
or
price = (unit_price * discount) - fixed

I prefer the first one.

CRogos (changes requested):
There is a bug when creating the new SO with discount. First the fixed discount is removed, but when saving the form, it is not correct.
[video (9).webm](https://github.com/OCA/sale-workflow/assets/1799080/87e9c533-79e4-4369-87da-23dfc6d45af5)

sale_fixed_discount/models/sale_order_line.py
PieterPaulussen: @CRogos Actually, I personally disagree. The fixed discount should only serve as a _convenience_ implementation. All actual discounts are applied through the `discount` field. This also greatly simplifies the modules code.
So the workflow will always be `price = unit_price * discount` and nothing more.
I don't see the added benefit of splitting those.

I agree that the discount might be misleading on the PDF though. Perhaps we should not add a separate column for this and just do something like:
`$ 100.00 (10.0%)` Or the other way around. We could add the currency to it as clarification?

The initial bug you mentioned in your video also needs to be addressed. I'll follow up on the issue this Friday.

Thanks for testing!

CRogos: @PieterPaulussen I also had a quick view in the new tax calculation and you are right that even in the new implementation we should not change `price = unit_price * discount` . I like your proposal regarding the PDF `$100.00 (10%)`.

I have one more issue, when we use discount_fixed and discount as synonym. When you enter the discount_fixed, the discount is updated. But when you enter a discount, the dicount_fixed field stays on 0.00. 
Shouldn't we make dicount_fixed as computed column which directly updates discount?

sale_fixed_discount/__manifest__.py
FrancoMaxime: @PieterPaulussen You're welcome :)

sale_fixed_discount/models/sale_order_line.py
PieterPaulussen: @CRogos I've added some additional changes to the module. Particularly one to stabilise the rounding differences for the calculated subtotals.

As for your specific issue where saving the sale order, I believe it is affected on the test environment by a different module also available in the sale-workflow repo. I've checked it locally and was unable to reproduce the error.

sale_fixed_discount/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,59 @@
...
+        >
+            <th
+                t-if="display_discount_fixed"
+                t-attf-class="text-end {{ 'd-none d-sm-table-cell' if report_type == 'html' else '' }}"
+            >
+                <span>Disc. Fixed Amount</span>
+            </th>
+        </xpath>
+        <xpath
+            expr="//section[@id='details']//tbody[hasclass('sale_tbody')]//t[@t-foreach='lines_to_report']/tr//t[@t-if='not line.display_type']//td[3]/div"
```
CRogos, line 29: I get the following error when I install the module from your branch.

```
  File "/src/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/src/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/src/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/src/odoo/odoo/http.py", line 1815, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/src/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/src/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/dataset.py", line 46, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
  File "/src/odoo/odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/src/odoo/odoo/api.py", line 448, in _call_kw_multi
  File "<decorator-gen-74>", line 2, in button_immediate_install
  File "/src/odoo/odoo/addons/base/models/ir_module.py", line 74, in check_and_log
    return method(self, *args, **kwargs)
  File "/src/odoo/odoo/addons/base/models/ir_module.py", line 456, in button_immediate_install
    return self._button_immediate_function(type(self).button_install)
  File "/src/odoo/odoo/addons/base/models/ir_module.py", line 580, in _button_immediate_function
    registry = modules.registry.Registry.new(self._cr.dbname, update_module=True)
  File "<decorator-gen-14>", line 2, in new
  File "/src/odoo/odoo/tools/func.py", line 87, in locked
    return func(inst, *args, **kwargs)
  File "/src/odoo/odoo/modules/registry.py", line 90, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/src/odoo/odoo/modules/loading.py", line 488, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/src/odoo/odoo/modules/loading.py", line 372, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/src/odoo/odoo/modules/loading.py", line 231, in load_module_graph
    load_data(cr, idref, mode, kind='data', package=package)
  File "/src/odoo/odoo/modules/loading.py", line 71, in load_data
    tools.convert_file(cr, package.name, filename, idref, mode, noupdate, kind)
  File "/src/odoo/odoo/tools/convert.py", line 763, in convert_file
    convert_xml_import(cr, module, fp, idref, mode, noupdate)
  File "/src/odoo/odoo/tools/convert.py", line 829, in convert_xml_import
    obj.parse(doc.getroot())
  File "/src/odoo/odoo/tools/convert.py", line 749, in parse
    self._tag_root(de)
  File "/src/odoo/odoo/tools/convert.py", line 709, in _tag_root
    raise ParseError(msg) from None  # Restart with "--log-handler odoo.tools.convert:DEBUG" for complete traceback
odoo.tools.convert.ParseError: while parsing None:7
Error while validating view:

Element '<xpath expr="//section[@id=&#39;details&#39;]//table/tbody//t[@t-foreach=&#39;sale_order.order_line&#39;]/tr//t[@t-if=&#39;not line.display_type&#39;]//td[3]/div">' cannot be located in parent view

View error context:
{'file': '/src/user/modules/oca/sale-workflow/sale_fixed_discount/views/sale_portal_templates.xml',
 'line': 2,
 'name': 'sale_order_portal_content_fixed_discount',
 'view': ir.ui.view(870,),
 'view.model': False,
 'view.parent': ir.ui.view(861,),
 'xmlid': 'sale_order_portal_content_fixed_discount'}
```

can't we change it to:

```
        <xpath
            expr="//div[@t-field='line.price_unit']"
            position="replace"
        >
            <div
                t-if="line.discount &gt;= 0 or line.fixed_discount &gt;= 0"
                t-field="line.price_unit"
                t-att-style="(line.discount or line.discount_fixed) and 'text-decoration: line-through' or None"
                t-att-class="((line.discount or line.discount_fixed) and 'text-danger' or '') + ' text-right'"
            />
            <div t-if="line.discount_fixed">
                <t
                    t-esc="line.price_unit - line.discount_fixed"
                    t-options='{"widget": "float", "decimal_precision": "Product Price"}'
                />
            </div>
        </xpath>
```

sale_fixed_discount/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,59 @@
...
+            />
+            <div t-if="line.discount_fixed">
+                <t
+                    t-esc="line.price_unit - line.discount_fixed"
+                    t-options='{"widget": "float", "decimal_precision": "Product Price"}'
+                />
+            </div>
+        </xpath>
+        <xpath
+            expr="//section[@id='details']//table/tbody//t[@t-foreach='lines_to_report']/tr//t[@t-if='not line.display_type']//td[@t-if='display_discount']"
```
CRogos, line 46: I also get here an error and replaced it with:
`expr="//section[@id='details']//t[@t-if='not line.display_type']//td[@t-if='display_discount']"`

CRogos (changes requested):
@PieterPaulussen are you still working on this?
I've installed the code from your branch, and I ran into several errors.
Either it is not working or I did something wrong.
image

sale_fixed_discount/views/sale_portal_templates.xml
PieterPaulussen: ~~There's a warning from pylint-odoo when you use `replace`. I specifically had to refactor this in order to not get the warning.~~

CRogos: I wasn't happy using replace either. Something which should be avoided so modules do not overwrite each other. But I did not have a good idea in this case. 

To bypass this warning you can set a priority="100" in the template node.

I'll have a look on your solution.

PieterPaulussen: Please disregard my previous comment. My assumption was wrong. I did however test the module again and did not come across your installation errors. 
I ran it on a new demo database and only installed `sale_fixed_discount` and manually added `sale_management`. Nothing more. 
Are you experiencing a side effect from another module in your installation?

PieterPaulussen:
> @PieterPaulussen are you still working on this? I've installed the code from your branch, and I ran into several errors. Either it is not working or I did something wrong. image

I was unable to reproduce the issue. Are you experiencing a side effect from another module in your installation?

CRogos:
Yes you are right... I proberbly had an installation issue on the sale-workflow part.

But there is the same issue you've fixed in account-invoicing part, but the behavior is a little different, because the (sub-)total is not chaging when you change the discount, after setting the fixed-discount.
image

And there is an error message on save.
image

PieterPaulussen:
@CRogos I've run into an issue with the framework. Adding onchanges to both `discount` and `discount_fixed` has an adverse effect on the workflow.
I've added context values in the forms to trigger the proper workflow when either field is changed. Personally I don't really like the approach, but its a decent tradeoff to restore the proper onchange functionality.

CRogos:
@PieterPaulussen are you still working on this? On my last test, setting the fixed discount to a value and than resetting it to 0 wasn't working.

CRogos (approved):
Now it looks good to me! Thank you for all your work @PieterPaulussen

CRogos:
@rousseldenis could you update your review so we can merge this?

MohamedOsman7 (approved):
Code review LGTM

CRogos:
@PieterPaulussen I think you can remove the last commit chaning the test-requiremnts.txt

PieterPaulussen:
Rebased and updated with new pre-commit changes

CRogos:
@pedrobaeza can you merge this PR?

pedrobaeza:
I let @rousseldenis to give their blessing

CRogos:
@rousseldenis will you add some feedback on this PR?

CRogos:
@pedrobaeza I don't think there will be further feedback... can we continue without @rousseldenis ?

PR #2617, opened 2023-07-26
[FIX] sale_credit_point: Use correct expression for translation

rousseldenis:


rousseldenis:
@Ivorra78 Please pay attention that those kind of strings should not be translated. Thanks

Ivorra78:
Ok ok, sorry

 

De: Denis Roussel (ACSONE) ***@***.***> 
Enviado el: miércoles, 26 de julio de 2023 10:30
Para: OCA/sale-workflow ***@***.***>
CC: Ivorra78 ***@***.***>; Mention ***@***.***>
Asunto: Re: [OCA/sale-workflow] [FIX] sale_credit_point: Use correct expression for translation (PR #2617)

 

@Ivorra78 <https://github.com/Ivorra78>  Please pay attention that those kind of strings should not be translated. Thanks

PR #2625, opened 2023-07-28
[14.0] sale_quotation_number: fix company sequence lookup

simoneversienti:
This PR is to fix a not completely correct behavior of this module in a multicompany context. 

Example: we have three companies active in the drop-down menu on the right (A, B and C) and the active one is A. So,  `self.env.company` is A. 
Sales --> Quotations --> Create. 

I can create a quotation with company_id = B. 
When i save, the module calculates the sequence calling `next_by_code` but, if not explicitly stated, the company considered is `self.env.company`, so A. 

In this case, i have a Quotation with company B but with the sequence of company A, which is wrong. 

The same happens when I confirm a quotation of company B.

### DIFF
--- a/sale_quotation_number/tests/test_sale_order.py
+++ b/sale_quotation_number/tests/test_sale_order.py
@@ -2,15 +2,28 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo.exceptions import UserError
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import SavepointCase
 
 
-class TestSaleOrder(TransactionCase):
-    def setUp(self, *args, **kwargs):
-        super(TestSaleOrder, self).setUp()
-        self.sale_order_model = self.env["sale.order"]
-        company = self.env.company
+class TestSaleOrder(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.sale_order_model = cls.env["sale.order"]
+        company = cls.env.company
         company.keep_name_so = False
+        cls.company1 = cls.env["res.company"].create(
+            {"name": "Test Company 1", "keep_name_so": False}
+        )
+        cls.partner = cls.env["res.partner"].create({"name": "Test Partner"})
+        cls.order_company1 = cls.env["sale.order"].create(
+            {
+                "name": "SQ/2023/001",
+                "partner_id": cls.partner.id,
+                "company_id": cls.company1.id,
+            }
+        )
 
     def test_enumeration(self):
         order1 = self.sale_order_model.create(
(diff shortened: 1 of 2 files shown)


dariodelzozzo (approved):

valeriobelcastro (approved):

HviorForgeFlow:
ping @OCA/crm-sales-marketing-maintainers

rafaelbn:
Please @Vanneri , @bizzappdev , @dsolanki-initos , could you review this FIX/PR? I guess you are using this module in production for a long time as you migrate it in different versions.

Thank you! 😄 🙏🏼 ❤️

simahawk:
@simoneversienti thanks for your fix: could you please add a test? It will make sure this is not broken in future migrations or changes.
Also, when you are at it, could yo please rewrite the commit msg? Odoo version number and square brakets are not needed and some easy shortening can be applied. Eg: `sale_quotation_number: fix multi-company` or `[fix] sale_quotation_number: fix multi-company`. Grazie :pray: 
Ref: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

simoneversienti:
> @simoneversienti thanks for your fix: could you please add a test? It will make sure this is not broken in future migrations or changes. Also, when you are at it, could yo please rewrite the commit msg? Odoo version number and square brakets are not needed and some easy shortening can be applied. Eg: `sale_quotation_number: fix multi-company` or `[fix] sale_quotation_number: fix multi-company`. Grazie pray Ref: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Done!
Thanks for the suggestions

sale_quotation_number/tests/test_sale_order.py
```diff
@@ -11,6 +11,17 @@ def setUp(self, *args, **kwargs):
         self.sale_order_model = self.env["sale.order"]
         company = self.env.company
         company.keep_name_so = False
+        self.company1 = self.env["res.company"].create(
```
simahawk: Thanks for adding tests!
Could you please move this setup to setUpClass? Reason: this code will create 6 companies + 6 partners + 6 orders as it runs for each test. This will make CI run slooooower :wink: 

```

    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        ....replace all `self` with `cls`....
```
The ctx trick is to save some more speed by disabling tracking.

simoneversienti: > 

Thanks @simahawk 
If i try to replace `setUp` with `setUpClass` (as in your suggestion), i get following error: 

`AttributeError: type object 'TestSaleOrder' has no attribute 'env'`

what am I doing wrong? what other changes should I make?

rousseldenis: @simoneversienti Because you should also change to SavepointCase instead of TransactionCase

simoneversienti: > @simoneversienti Because you should also change to SavepointCase instead of TransactionCase

thanks. Done!

simoneversienti:
@rousseldenis @simahawk ping

simahawk (approved):

simoneversienti:
what is the problem?

mmequignon:
> what is the problem?

Looks like there's a pre-commit issue
https://github.com/OCA/sale-workflow/actions/runs/6314386186/job/17144536970

PR #2629, opened 2023-08-03
[16.0][ADD] sale_pricelist_display_surcharge

DavidJForgeFlow:
In standard Odoo only display the discount if it benefices the customer.
In the case of wanting to always show the discount (for example a surcharge), this module allows to choose if show or not the customer the discount.

@ForgeFlow


rousseldenis:
@DavidJForgeFlow What do you think adding 'pricelist' term to module name? As this is more explicit. So, sale_display_pricelist_surcharge'

DavidJForgeFlow:
> @DavidJForgeFlow What do you think adding 'pricelist' term to module name? As this is more explicit. So, sale_display_pricelist_surcharge'

Changed!

rousseldenis (approved):

rousseldenis:
@DavidJForgeFlow Could you attract people for more reviews ?

sale_pricelist_display_surcharge/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+In standard Odoo only display the discount if it benefices the customer.
+In the case of wanting to always show the discount (for example a surcharge),
+this module allows to choose if show or not the customer the discount.
```
JordiMForgeFlow:
```suggestion
In standard Odoo the discount is only displayed if it benefits the customer.
In the case of wanting to always show the discount (for example a surcharge),
this module allows to choose whether to show or not.
```

DavidJForgeFlow: Done!

AaronHForgeFlow (approved):
Code LGTM

JordiMForgeFlow:
@DavidJForgeFlow would it be possible to add some test?

DavidJForgeFlow:
> @DavidJForgeFlow would it be possible to add some test?

Makes sense to add tests here, I will note it and add them when I have time!

rousseldenis:
@DavidJForgeFlow

DavidJForgeFlow:
@JordiMForgeFlow @rousseldenis Tests added!

JordiMForgeFlow:
@DavidJForgeFlow test seem to be failing, could you check?

DavidJForgeFlow:
> @DavidJForgeFlow test seem to be failing, could you check?

Seems an error common in all 16.0 PRs. If monday keeps happening I will investigate further!

DavidJForgeFlow:
@JordiMForgeFlow there is a conflict between two modules that cannot be installed at the same time: account_invoice_triple_discount and account_invoice_fixed_discount. The test including them fails, the other one works. Should we exclude them?

JordiMForgeFlow:
@DavidJForgeFlow we should do the same as it was done in the account-invoicing repo where those base modules are (https://github.com/OCA/account-invoicing/commit/c6144b73438920915a64cf76cb2ac09f23d1bc2f).

Running a copier update you can specify the rebel modules (here we should separate sale_fixed_discount and sale_triple_discount). Maybe better if you do it in a separate PR

DavidJForgeFlow:
There is a PR open trying to fix this error: https://github.com/OCA/sale-workflow/pull/2885. As the module sale_triple_discount is also failing by adding this modules as rebel: https://github.com/OCA/sale-workflow/pull/3629. Let's wait until @grindtildeath fix this error!

DavidJForgeFlow:
@JordiMForgeFlow @rousseldenis This is now ready to merge!

PR #2631, opened 2023-08-07
[ADD] Ahora los tipo de venta tiene un flujo mas logico

Brian-M-Lopez:
Feature para el bug del ticket 
https://www.aitic.com.ar/web#id=7266&view_type=form&model=helpdesk.ticket&action=543&menu_id=406


Brian-M-Lopez:
It was a mistake, i taget the wrong repository

pedrobaeza:
Well, keeping forks with different features/hotfixes is not an ideal thing for both you and the community. If you think it's worth to change to that "more logical flow", try to propose the change, explaining the reasons why it's more logical.

PR #2649, opened 2023-08-21
[16.0][IMP] sale_triple_discount: Compute amount with decimal precision of 16

hildickethan:
- Force the calculation to be with full decimal precision, this is equivalent to the current computation in invoices
https://github.com/OCA/account-invoicing/blob/d2e12c03b50011bb788cf482346710f0fefe0de3/account_invoice_triple_discount/models/account_move_line.py#L73-L74

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -107,6 +107,8 @@ class SaleOrderLine(models.Model):
         # to invalidate the cache to avoid to flush the records to the database.
         # This is safe because we are going to restore the original value at the end
         # of the method.
+        digits = discount_field._digits
+        self.env["sale.order.line"]._fields["discount"]._digits = (16, 16)
         with self.env.protecting([discount_field], self):
             old_values = {}
             for line in self:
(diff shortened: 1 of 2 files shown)


johnny-longneck (approved):

rousseldenis (commented):
Could you add a test case for this ?

hildickethan:
> Could you add a test case for this ?

Added the exact use case I had issues with, if you multiply `138*2*0.55*0.95*0.9` you get 129.789 which without these changes Odoo cuts off at 129.78 because the "final discount" gets cut off at 2 decimal precision and then in turn on the invoice it calculates 129.79 because it calculates with 16 digits, causing the mismatch

sale_triple_discount/models/sale_order_line.py
```diff
@@ -97,6 +97,9 @@ def triple_discount_preprocess(self):
         this method is called multiple times."""
 
         prev_values = dict()
+        digits = self._fields["discount"]._digits
+        self.env["sale.order.line"]._fields["discount"]._digits = (16, 16)
```
rousseldenis, line 111: I'm not fan of this. I think this could lead to unwanted side effects.

hildickethan: Its not very pretty but I think it hasn't caused any issues for us so far, its mainly copying the process in the account equivalent module https://github.com/OCA/account-invoicing/blob/d2e12c03b50011bb788cf482346710f0fefe0de3/account_invoice_triple_discount/models/account_move_line.py#L73-L74 so it calculates the same discount, otherwise it can lead to different rounding

PR #2657, opened 2023-08-28
[ADD] sale_packaging_default: default packaging for sales

yajo:
This module simplifies and emphasizes the usage of packaging in sales orders.

- Packaging fields in sale order lines appear before product quantity fields.
- When selecting a product to sell, its default packaging is added automatically.

Why is this module useful? Packaging usage in Odoo is designed to be computed automatically when you choose the final amount of products to sell. Although you can still do that, with this module the natural usage is inverted: it's easier for you to select the packaging and quantity first, and then the final amount of products is computed automatically.

@moduon MT-3695


fcvalgar:
Would it be interesting if only one default field could be selected for the package? @yajo 

image

rafaelbn (commented):
The default vaule is set

imagen

yajo:
> Would it be interesting if only one default field could be selected for the package?

It would be problematic because I'd need to depend on records that aren't directly related (the value on one packaging depends on the others). Currently only the 1st one is considered, and that's documented. If it's enough, it's better like this (simpler to implement).

fcvalgar (approved):
LGTM, Thank you @yajo

yajo:
FWIW before designing this module, we took a look at https://github.com/OCA/sale-workflow/tree/14.0/sale_by_packaging but it's following a very different purpose. That module establishes a lot of constraints on how to do the sale process. This one just adds features that allow salespersons to work faster with Odoo, but doesn't alter dramatically the essential Odoo behaviors nor imposes any new constraints.

rafaelbn (approved):
Tested un Staging as runboat doesn't works!

👍🏼

PR #2659, opened 2023-08-29
[15.0][FIX] sale_planner_calendar: Recurrent events not removed when changing the recurrence from the planner

CarlosRoca13:
cc @Tecnativa TT44306

### DIFF
--- a/sale_planner_calendar/models/calendar.py
+++ b/sale_planner_calendar/models/calendar.py
@@ -113,6 +113,18 @@ class CalendarEvent(models.Model):
                 }
             )
 
+    def _rewrite_recurrence(self, values, time_values, recurrence_values):
+        """Remove key from context when this method is executed to avoid loose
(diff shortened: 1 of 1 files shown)


sale_planner_calendar/models/calendar.py
```diff
@@ -113,6 +113,18 @@ def _inverse_hour(self):
                 }
             )
 
+    def _rewrite_recurrence(self, values, time_values, recurrence_values):
+        """Remove key from context when this method is executed to avoid loose
+        information when accesing calendar_event_ids of the recurrence_id.
+        """
```
sergio-teruel, line 119: Set ctx before condition to avoid "local variable 'ctx' referenced before assignment" error

CarlosRoca13: Done

sergio-teruel (approved):

carlosdauden (approved):

PR #2660, opened 2023-08-30
[14.0][ADD]sale_order_general_discount_payment_term: Add sale order general discount on payment term

odooNextev:
This modulo allow to add sale_order general discount based on payment term discount field.
Also hide discount from partner to avoid conflicts


francesco-ooops (approved):
Functional ok!

sale_order_general_discount_payment_term/models/account_payment_term.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class AccountPaymentTerm(models.Model):
+    _inherit = "account.payment.term"
+
+    sale_discount = fields.Float(
+        store=True,
+        readonly=False,
+    )
```
HekkiMelody: No point restating obvious defaults IMO. (This is a new field, not an override, correct?)

Instead, I think it'd be good to add the same modifiers that the corresponding field has in `sale_order_general_discount`

```py
    sale_discount = fields.Float(
        digits=dp.get_precision('Discount'),
        string='Discount (%)',
        company_dependent=True,
        track_visibility="onchange",
    )
```

sale_order_general_discount_payment_term/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
...
+                position="attributes"
+            >
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath
+                expr="//page[@name='sales_purchases']//div[field[@name='sale_discount']]"
+                position="attributes"
+            >
+                <attribute name="invisible">1</attribute>
+            </xpath>
```
HekkiMelody:
```suggestion
            <field name="sale_discount" position="attributes">
                <attribute name="invisible">1</attribute>
            </field>
```

is this not sufficient, or does it not work for some reason?

sale_order_general_discount_payment_term/models/sale_order.py
```diff
@@ -0,0 +1,11 @@
...
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends("payment_term_id")
+    def _compute_general_discount(self):
+        for so in self:
+            so.general_discount = so.payment_term_id.sale_discount
```
HekkiMelody: `api.depends` and `_compute` functions are generally used if they're computing the value of a field.

I think this should be an `api.onchange`

Cf. the module `sale_order_general_discount`:

```py
    @api.onchange('partner_id')
        super().onchange_partner_id()
        self.general_discount = self.partner_id.sale_discount
        return
```

HekkiMelody (changes requested):
You hide the original field, but for existing installations the value of `sale_discount` on `res.partner`s will still exist and it will keep being applied to `sale.order`s, with no way for the user to change it.

There are a few options to deactivate it entirely, One of them could be a `post_init_hook` that just writes `0` on every partner's `sale_discount` field. I don't know if people would find it too aggressive, though.

Another option is to override `onchange_partner_id`, save the existing discount, call `super()` and then restore the discount that existed before the `onchange` was called, so as to nullify the effect of the change in `sale_order_general_discount` (every line's discount would be recalculated twice for nothing, but alas, I've seen worse in Odoo :D ).

Alternatively, since `sale_order_general_discount` is so small and you want to deactivate half of it, you could just replicate its functionality and not depend on it. (and maybe add it to the `excludes` key in the manifest so people don't accidentally install both).

I think option number 2 (overriding `onchange_partner_id` is by far the most preferable, though)

sale_order_general_discount_payment_term/views/res_partner.xml
odooNextev: Hi @aleuffre, thanks for your review. Set "invisible" on field sale_discount it doesn't works properly because we need to hide a div with sale_discount and a string inside.

sale_order_general_discount_payment_term/models/sale_order.py
odooNextev: Hi @aleuffre, I think that is 12.0 version. In v14 this is the code in sale_order_general_discount:

@api.depends("partner_id")
    def _compute_general_discount(self):
        for so in self:
            so.general_discount = so.partner_id.sale_discount

HekkiMelody: Ah, my bad. If you're overriding an existing compute and it works, then that's fine by me.

HekkiMelody:
Sorry for my oversights during my first review. Trying to do too many things at once will do that to you.

Adding the field parameters to `sale_discount` seems to be the only thing left. After that I'll approve.

sale_order_general_discount_payment_term/models/account_payment_term.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class AccountPaymentTerm(models.Model):
+    _inherit = "account.payment.term"
+
+    sale_discount = fields.Float(
+        digits="Discount",
+        string="Discount (%)",
+        company_dependent=True,
+        track_visibility="onchange",
```
HekkiMelody: as you correctly spotted in the other comment, I was looking at v12 code during my previous review.

`track_visibility` was a v12 option, in v14 it's simply `tracking=True`

HekkiMelody:
Code review looks good aside from that one small thing, please squash everything into a single commit.

HekkiMelody (approved):

francesco-ooops:

francesco-ooops:
@pedrobaeza could this be merged please?

sale_order_general_discount_payment_term/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
...
+<odoo>
+    <record id="view_partner_form_hide_discount" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="sale_order_general_discount.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//page[@name='sales_purchases']//label[@for='sale_discount']"
+                position="attributes"
+            >
+                <attribute name="invisible">1</attribute>
```
pedrobaeza: This is not correct. This is out of scope of the module. And there's already a group for hiding this for those you don't want to see it.

pedrobaeza (changes requested):

sale_order_general_discount_payment_term/views/res_partner.xml
francesco-ooops: @pedrobaeza the point of this change is that if there is a general discount set in partner and another general discount set in payment term they will be in conflict for setting the general discount on SO. 

How else do you suggest we avoid this scenario?

pedrobaeza: OK, then hide it only if the payment term is set and it contains a discount.

francesco-ooops: Sorry, the flow is unclear to me. What happens if partner 1 has a default general discount set and in SO user selects a payment term with a general discount? Which general discount should be applied?

With this module installed, it's very likely that the only general discount that should be applied is the payment term one. 

Shall we make it easier with a flag in settings allowing user to decide whether to hide the general discount in partner field?

pedrobaeza: You should clear that general discount on payment term assignment IMO.

pedrobaeza: The question is that you shouldn't remove the possibility of assigning directly a general discount for all cases. Just for those covered by this module (with payment term). If not, you will force to assign a payment term always if you want a general discount.

francesco-ooops: User has the possibility of assigning general discount in SO in 3 ways: 

1) default from partner
2) manually from field "discount" in SO
3) default from payment term

base module "sale_order_general_discount" implements the first two options; this module implements the third.

I think if we add a "hide default general discount in partner" flag in settings to this module we will have the cleaner implementation as the user can decide:

- whether to keep all 3 options,
- to keep only 2 and 3 (flag enabled), or 
- to keep 1 and 2 by not installing this module. 

Could that work?

pedrobaeza: I still think that you don't need another extra config. Just don't hide the field if you don't have a payment term selected.

francesco-ooops: Alright, @odooNextev can you implement that?

francesco-ooops: @pedrobaeza with the latest change by @odooNextev , if payment term discount = 0 and partner default discount is set, partner default is used in SO; otherwise payment term discount is used. Could this work?

pedrobaeza (approved):
/ocabot merge nobump

PR #2661, opened 2023-08-30
[16.0][MIG] sale_resource_booking: Migration to 16.0

norlinhenrik:


rrebollo:
@norlinhenrik me over here to. In my doodba local environment no tests are failling this same branch. Is there any way to trigger checks again???

rrebollo:
@jguenat @stefan-tecnativa @norlinhenrik  I can´t see failling tests. I even didn´t find any when running in my local doodba. Could anyone trigger tests again?
image

norlinhenrik:
I think OCA cannot successfully test `sale_resource_booking` until `resource_booking` is merged. But you can do it on your local computer with these parameters:
 `--test-enable --test-tags=/sale_resource_booking --stop-after-init` (optional)

rrebollo:
@norlinhenrik I'm considering to make an small improvement: trigger an onchange to auto select first product when booking type is selected. Why? I was testing to create a quote directly from booking, an error raises, looking at the trace I figure out it's because I forgot to select product. What do U think? Is there any convention against "onchanges"? Anyway we need to prevent quote action from calling if no product is selected? Which path would U suggest??
image

norlinhenrik:
"onchange to auto select first product" is good if there is only one product available. If there are many products available, I think it is better to select the product manually.

"prevent quote action from calling if no product is selected" is also good to implement.

rrebollo:
> "onchange to auto select first product" is good if there is only one product available. If there are many products available, I think it is better to select the product manually.
> 
> "prevent quote action from calling if no product is selected" is also good to implement.

Last two commits here to implement mentioned improvements https://github.com/BinhexSystemss/sale-workflow/tree/16.0-dev. Could U merge it???

rrebollo:
Sale order isn't been "reload" after booking confirmation wizard closes. U need to manually reload browser tab to see the state changed to confirmed. It doesn't matter which button U click on wizard.
sale order doesn´t reload after booking confirmation
I did some digging so I think the root cause it's maybe in [web_ir_actions_act_multi](https://github.com/OCA/web/tree/16.0/web_ir_actions_act_multi).
By our part the problem lies here:

https://github.com/ows-cloud/sale-workflow/blob/d68949ea1212ed5c458a1adb61930fce34c9adb3/sale_resource_booking/models/sale_order.py#L48-L66
```
    def action_confirm(self):
        """Ask to fill booking values, if necessary."""
        result = super().action_confirm()
        try:
            # Only open wizard if doing this from a single record with bookings
            if len(self) > 1 or not self.resource_booking_ids:
                return result
        except exceptions.AccessError:
            # User without access to resource.booking; no confirm wizard
            return result
        # Support chained actions, like when event_sale is installed
        action = self.env["ir.actions.act_window"]._for_xml_id(
            "sale_resource_booking.sale_order_booking_confirm_action"
        )
        action["context"] = {"default_order_id": self.id}
        actions = [action]
        if isinstance(result, dict):
            actions.insert(0, result)
        return {"type": "ir.actions.act_multi", "actions": actions}
```
Following [web_ir_actions_act_multi](https://github.com/OCA/web/tree/16.0/web_ir_actions_act_multi) README instructions I tried to just add reload action client to actions list but then tab is reloaded and wizard got never called. As a temporal (and awful) workaround I'm just directly returning reload action client from both wizard 's buttons actions. Any thougths @norlinhenrik @pedrobaeza?

norlinhenrik:
@rrebollo I made a commit with a workaround similar to what you described. This is how it will be then after CONFIRM.
Confirm order with resource booking

sale_resource_booking/views/resource_booking_views.xml
```diff
@@ -0,0 +1,57 @@
...
+                </group>
+            </group>
+        </field>
+    </record>
+    <record id="resource_booking_view_tree" model="ir.ui.view">
+        <field name="name">resource.booking.tree.inherit</field>
+        <field name="model">resource.booking</field>
+        <field name="inherit_id" ref="resource_booking.resource_booking_view_tree" />
+        <field name="arch" type="xml">
+            <field name="partner_id" position="after">
```
augustodinizl: i think that the correct field here, is partner_ids

<field name="partner_ids" position="after">

augustodinizl (changes requested):
typo error

sale_resource_booking/views/resource_booking_views.xml
norlinhenrik: Done

norlinhenrik:
@rrebollo @augustodinizl Are you ready to approve, or do you see more things to do first?

augustodinizl:
> @rrebollo @augustodinizl Are you ready to approve, or do you see more things to do first?

LGTM ... it works in my test base

augustodinizl (approved):
LGTM

norlinhenrik:
I have squahed my commits now, and I think it is ready to merge.

pedrobaeza:
CI seems to be red. Can you check?

rrebollo:
> CI seems to be red. Can you check?

Hi, it looks none of the failing tests are linked to sale_resource_booking, maybe re triggering?

pedrobaeza:
Try to rebase to see

norlinhenrik:
I have rebased now.

First failing check: `sale_resource_booking` is installed in lines 515-559.
The test fails at line 924. Bottom: `psycopg2.errors.SerializationFailure: could not serialize access due to concurrent update`

`sale.order.line`
In the last push, resource_booking_id is renamed to "Resource Booking" (singular).
Still the log complains about 2 fields named "Resource Bookings" (plural)...

christian-ramos-tecnativa:
IMHO, this PR includes too many changes for a migration, making it harder to review. I appreciate the work done here, but wouldn't it be better to handle the migration while preserving the existing behavior and introduce the product variant functionality in a separate PR?

norlinhenrik:
The product variant functionality is a core change and deserves to be discussed.

Odoo introduces core changes in main versions. I thought it might be best to introduce this change in the migration PR, since an upgrade from 15.0.x.x.x to 16.0.1.0.0 is usually more carefully tested than an upgrade from 16.0.1.x.x to 16.0.2.0.0.

Do you still want me to take the product variant functionality into a separate PR?

pedrobaeza:
Blocking the migration for both reviewing the migration itself and the improvements may be too much, so the suggestion is just for speeding up the migration merge. And this also avoids possible conflicts if the improvements are not accepted.

norlinhenrik:
This PR is ready for review.

christian-ramos-tecnativa:
It seems like some tests are failing, can you check it @norlinhenrik? Thanks in advance

norlinhenrik:
The tests are passing.

When ALL the sale-workflow addons are installed, canceling the sale order will not cancel the resource booking. I have no clue which addon interferes with sale_resource_booking. If anyone would like to install and debug locally, feel free to do so.

rrebollo (approved):
**Code and Functional Review**  
LGTM! :+1: But you need to deal with failing tests.

@norlinhenrik I can confirm that whatever addon was previously interfering with the cancellation of bookings linked to a canceled sale order is no longer causing issues. :thumbsup:

pedrobaeza:
/ocabot migration sale_resource_booking

Please check CI

ByteMeAsap:
The same issue as mentioned at https://github.com/OCA/sale-workflow/pull/3870#issuecomment-3298632747  for v18 is happening for v16 as well.

The proposed solution for the same is modifying the sale.order.booking.confirm by adding the following:

```
    @api.model
    def _default_resource_booking_ids(self):
        order = self.env["sale.order"].browse(self.env.context.get("default_order_id"))
        return order.order_line.mapped("resource_booking_ids")

    order_id = fields.Many2one(
        "sale.order",
        string="Order",
        index=True,
        required=True,
        readonly=True,
        ondelete="cascade",
    )
    resource_booking_ids = fields.Many2many(
        "resource.booking",
        "sale_order_booking_confirm_resource_booking_rel",
        "sale_order_booking_confirm_id",
        "resource_booking_id",
        string="Bookings",
        default=_default_resource_booking_ids,
    )
```

And updating the view as follows:
```
<field name="resource_booking_ids" options="{'unlink': False,'link':False}">
<tree string="Resource bookings" editable="bottom">
```

rousseldenis:
@norlinhenrik

rrebollo:
@norlinhenrik Could I take it from here?

rrebollo:
@pedrobaeza please close this in favor of #4072

PR #2666, opened 2023-08-31
[FIX][14.0]sale_order_general_discount_triple: add flag in context avoid adding multiple general discount

odooNextev:
…ltiple discounts

### DIFF
--- a/sale_order_general_discount/models/sale_order_line.py
+++ b/sale_order_general_discount/models/sale_order_line.py
@@ -1,5 +1,6 @@
 # Copyright 2018 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
 from odoo import api, fields, models
 
 
@@ -17,7 +18,11 @@ class SaleOrderLine(models.Model):
         """Apply general discount for sale order lines which are not created
         from sale order form view.
         """
-        if "discount" not in vals and "order_id" in vals:
+        if (
+            not self.env.context.get("no_propagate_general_discount")
+            and "discount" not in vals
+            and "order_id" in vals
+        ):
             sale_order = self.env["sale.order"].browse(vals["order_id"])
             if sale_order.general_discount:
                 vals["discount"] = sale_order.general_discount
--- a/sale_order_general_discount_triple/models/sale_order_line.py
+++ b/sale_order_general_discount_triple/models/sale_order_line.py
@@ -51,4 +51,5 @@ class SaleOrderLine(models.Model):
         )
         if general_discount:
             vals[general_discount] = sale_order.general_discount
+            self = self.with_context(no_propagate_general_discount=True)
         return super().create(vals)


HekkiMelody (changes requested):
On a technical level it works, but I fear that this is a bug waiting to happen. It will catch any field with `discount` in the name, which is potentially very broad and may interact in unpredictable ways with other future modules, public or private.

I assume you're trying to fix the interaction with `sale_order_general_discount_triple`, so you want to catch `discount2` and `discount3`.

I would recommend another option. Two ideas

1. Set some flag in the context, and skip this check if the flag is set. Then, from `sale_order_general_discount_triple` set this flag with `with_context` before calling `super()`

2. Let the `create` of `sale_order_general_discount` happen. then in `sale_order_general_discount_triple`, after calling `super()`, move the discount from the field `discount` (where it will always go) to `discount2` or `discount3` as necessary. It's computationally not ideal but should be a lot cleaner and have fewer side effects.

I personally prefer the latter.

odooNextev:
Hi @aleuffre, Thanks for your review. I'll proceed with first solution that you suggested me

odooNextev:
I opened another pr to add flag in context, do you think is the right solution?

sale_order_general_discount/models/sale_order_line.py
```diff
@@ -17,7 +18,11 @@ def create(self, vals):
         """Apply general discount for sale order lines which are not created
         from sale order form view.
         """
-        if "discount" not in vals and "order_id" in vals:
+        if (
+            not ("flag" in self.env.context and self.env.context["flag"])
```
HekkiMelody:
```suggestion
            not self.env.context.get("flag")
```

HekkiMelody (changes requested):
See also other review here: https://github.com/OCA/sale-workflow/pull/2668#pullrequestreview-1607045199

Please edit the first message in this PR to clearly indicate that it depends on the prior approval of https://github.com/OCA/sale-workflow/pull/2668

sale_order_general_discount/models/sale_order_line.py
odooNextev: Thanks @aleuffre, I added changes that you suggested me. It doesn't depends anymore from https://github.com/OCA/sale-workflow/pull/2668 because all commits are here. I think I need to close that Pr

HekkiMelody (approved):
Code review, LGTM, please squash your commits so only one commit is left.

francesco-ooops:
@aleuffre we're leaving 1 commit per module (I think that's standard?), @odooNextev can you change the message according to https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id48 ?

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@RabbitJon-S73 check if this solution can work for you :)

HekkiMelody:
> @aleuffre we're leaving 1 commit per module (I think that's standard?), @odooNextev can you change the message according to https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id48 ?

This is not what it currently is. Currently there's one commit with the old changes with the regex. And then there's one commit that reverts the first change and adds the change to the two modules. So the first commit could be simply removed.

And then yes, ideally as you say the changes to the two modules would be split in two different commits.

odooNextev:
@aleuffre I founded this discussion where they were talking about differents commits for differents module. https://github.com/OCA/l10n-italy/pull/3500#issuecomment-1646071938
What's the best solution?

HekkiMelody:
> @aleuffre I founded this discussion where they were talking about differents commits for differents module. [OCA/l10n-italy#3500 (comment)](https://github.com/OCA/l10n-italy/pull/3500#issuecomment-1646071938) What's the best solution?

Either way is fine. IMO. I think you can leave it as is unless someone with more experience than me says otherwise.

odooNextev:
I divided changes in two different commits, one for module,as you told me. I think that will be fine, otherwise I will rebase in only one commit.

francesco-ooops:

before this PR, if general discount was set on "discount 3" and on SO field "general discount" was populated, when adding shipping expenses to SO the discount would appear both on field "discount" and "discount 3".

RabbitJon-S73:
> @RabbitJon-S73 check if this solution can work for you :)

That won't work since I remove the `create` function and apply compute instead. Here is the explanation why this was done https://github.com/OCA/sale-workflow/pull/2358#issuecomment-1669151779.  Watch https://github.com/OCA/sale-workflow/pull/2392 because @francesco-ooops alredy report me that problems and I tried to fix them

francesco-ooops:
@pedrobaeza can we merge please?

PR #2668, opened 2023-09-01
[IMP]sale_order_general_discount_triple:add flag in context to check …

odooNextev:
…if discount already exist on sale_order_general_discount

### DIFF
--- a/sale_order_general_discount_triple/models/sale_order_line.py
+++ b/sale_order_general_discount_triple/models/sale_order_line.py
@@ -49,6 +49,8 @@ class SaleOrderLine(models.Model):
                 "sale_order_general_discount_triple.general_discount", "discount"
             )
         )
+        flag = False
         if general_discount:
             vals[general_discount] = sale_order.general_discount
-        return super().create(vals)
+            flag = True
+        return super(SaleOrderLine, self.with_context(flag=flag)).create(vals)


odooNextev:
I added flag in context to check if a discount already exist and avoid to add another general discount on sale_order_general_discount create method. 
https://github.com/OCA/sale-workflow/pull/2666

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -49,6 +49,8 @@ def create(self, vals):
                 "sale_order_general_discount_triple.general_discount", "discount"
             )
         )
+        flag = False
         if general_discount:
             vals[general_discount] = sale_order.general_discount
-        return super().create(vals)
+            flag = True
+        return super(SaleOrderLine, self.with_context(flag=flag)).create(vals)
```
HekkiMelody, line 56:
```suggestion
            self = self.with_context(no_propagate_general_discount=True)
        return super(SaleOrderLine, self).create(vals)
```

sale_order_general_discount_triple/models/sale_order_line.py
```diff
@@ -49,6 +49,8 @@ def create(self, vals):
                 "sale_order_general_discount_triple.general_discount", "discount"
             )
         )
+        flag = False
```
HekkiMelody, line 52:
```suggestion

```

HekkiMelody (changes requested):
It's more or less okay but a name that is more descriptive than "flag" would really help

odooNextev:
superseeded https://github.com/OCA/sale-workflow/pull/2666

PR #2674, opened 2023-09-06
[ADD] sale_order_line_tag

SMaciasOSI:


dreispt (approved):

dalonsod (commented):
Hello @SMaciasOSI thanks for the contribution!!!

Would you consider adding some typical features related to tags management, like `color` field (with color picker widget), `active` and `company_id` features, like https://github.com/OCA/helpdesk/blob/16.0/helpdesk_mgmt/models/helpdesk_ticket_tag.py ?

dreispt:
@dalonsod That is a good suggestion, I think it is a great opportunity to contribute and improve the module.

PR #2683, opened 2023-09-12
[FIX] sale_order_type: duplicate payment

jjscarafia:
Steps to reproduce the bug:
1. create an internal transfer between cash and bank, post it
2. duplicate the internal transfer, try to post it.

ERROR: ValueError: Expected singleton: res.currency()

VIDEO: https://drive.google.com/file/d/1JsGIhO8VhwVmhhotPGicr4P5xUPPRYDV/view

Of course this solution is not ideal, suggestions welcome. On my tests, the behaviour is broken due to the "@api.depends("sale_type_id")" and not to the inheritance of the method itself.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -56,9 +56,11 @@ class AccountMove(models.Model):
             move.invoice_payment_term_id = move.sale_type_id.payment_term_id
         return res
 
-    @api.depends("sale_type_id")
-    def _compute_journal_id(self):
-        res = super()._compute_journal_id()
+    @api.onchange("sale_type_id")
+    def _onchange_sale_type(self):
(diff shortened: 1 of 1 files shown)


rousseldenis:
@jjscarafia This is invalid as the bug is present without sale_order_type installed.

IMHO this is an Odoo bug

rousseldenis:
FYI, this could be solved by storing ccompany_currency_id field on account.move:

https://github.com/odoo/odoo/blob/324e1387bdc786175072fff6492ed361f5f6fb93/addons/account/models/account_move.py#L376

As this is not correct to store a related on another non stored related...

rousseldenis:
Created that one : https://github.com/odoo/odoo/pull/135604

jjscarafia:
closing for inactivity
also, as far as I've tested on 17 is no needed

PR #2685, opened 2023-09-12
[14.0][IMP] sale_order_general_discount: allow to include specific products

HekkiMelody:
This commit introduces a new setting that allows to set a domain on the products that are meant to be included in the general discount. Sale order lines with products that are excluded do not follow the order's global discount rate.

If the filter is empty, all products are included, which preserves the current behaviour of the module.

The second commit introduces slight changes to `sale_order_general_discount_triple` for compatibility with the first set of changes.


francesco-ooops (approved):
Functional ok!

This PR addresses the use case where user doesn't want general discount to be applied to shipping expenses (or any other specific product).

sale_order_general_discount/models/sale_order_line.py
```diff
@@ -13,24 +14,44 @@ class SaleOrderLine(models.Model):
...
+    @api.model
+    def get_discount_vals_for_product(self, product_id, order_id):
+        if isinstance(order_id, int):
+            order_id = self.env["sale.order"].browse(order_id)
+        if not isinstance(product_id, int):
+            product_id = product_id.id
+        domain = order_id.company_id.sudo()._get_general_discount_eval_domain()
+        if domain:
+            domain = expression.AND([[("id", "=", product_id)], domain])
+            product_to_discount = self.env["product.product"].sudo().search(domain).ids
```
geomer198: Just do like this:
```
if not self.env["product.product"].sudo().search(domain):
    return {}
```

sale_order_general_discount/models/sale_order_line.py
```diff
@@ -13,24 +14,44 @@ class SaleOrderLine(models.Model):
...
+            # hook for other modules such as sale_order_general_discount_triple
             not self.env.context.get("no_propagate_general_discount")
             and "discount" not in vals
-            and "order_id" in vals
         ):
-            sale_order = self.env["sale.order"].browse(vals["order_id"])
-            if sale_order.general_discount:
-                vals["discount"] = sale_order.general_discount
+            discount_vals = self.get_discount_vals_for_product(
+                vals.get("product_id", 0), vals["order_id"]
```
geomer198: `vals["order_id"]` maybe raise Exception. Please check this argument in condition.
`if .... and vals.get("order_id"):`

geomer198 (commented):
Please check comments.

sale_order_general_discount/models/sale_order_line.py
HekkiMelody: I don't know if this would ever happen since `order_id` is a mandatory field, but I don't mind adding it.

sale_order_general_discount/models/sale_order_line.py
HekkiMelody: Aaaah, it was a leftover from a previous attempt (it made sense to have the ids, at some point) thanks for pointing it out!

geomer198 (approved):
Code Review Ok!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers could this be merged? thanks!

rafaelbn:
I would like yo have the revision of the maintainer and original author @sergio-teruel , if possible. thank you!

sergio-teruel:
Sorry @aleuffre. You are changing the behavior of this module entirely... this module applies a discount for all lines.. if you wat apply a discount only for a set of products you can set it in pricelists

francesco-ooops:
> Sorry @aleuffre. You are changing the behavior of this module entirely... this module applies a discount for all lines.. if you wat apply a discount only for a set of products you can set it in pricelists

@sergio-teruel the base behavior of the module is untouched, while providing user the chance to employ all the features of general_discount and exclude products where discount is unwanted. 

This is the reason why we proposed an IMP rather than an extending module (also take into consideration new modules depending on this one have been merged and that would imply creating bridge modules which is far from ideal).

I struggle to figure out how the same behavior guaranteed by this IMP can be replicated in a sustainable fashion with pricelist, as this module allows for a quick discount in SO with a value set on the moment, and this is the use case we're addressing: 

> user doesn't want general discount to be applied to shipping expenses (or any other specific product).

This is especially evident taking into consideration the recent changes to sale_order_general_discount_triple and the addition of sale_order_general_discount_payment_term: pricelist discount can be preserved, and general discount added to another discount field. 

I hope I've been convincing about the solution proposed :)

francesco-ooops:
@sergio-teruel would it be ok for us to proceed with this improvement?

thanks :)

francesco-ooops:
@pedrobaeza could you help break the deadlock?

pedrobaeza:
Well, it's my colleague who should give his final call.

francesco-ooops:
@sergio-teruel gentle reminder then

sergio-teruel:
ok for me

pedrobaeza (changes requested):
What is not correct is to delay the discount assignation up to the record creation

francesco-ooops:
@pedrobaeza go?

HekkiMelody:
> What is not correct is to delay the discount assignation up to the record creation

I'm sorry, I don't quite understand what you mean. Could you please elaborate?

francesco-ooops:
> > What is not correct is to delay the discount assignation up to the record creation
> 
> I'm sorry, I don't quite understand what you mean. Could you please elaborate?

@pedrobaeza

pedrobaeza:
I mean the override on the `create` method, which is delaying the discount computation till the DB registration, which is not correct IMO.

HekkiMelody:
> I mean the override on the `create` method, which is delaying the discount computation till the DB registration, which is not correct IMO.

@pedrobaeza I always compute the discount before calling `super()` in create methods, so the discount computation is not being delayed.

What would you suggest? Removing the `compute` attribute from `discount2` and `discount3`?

https://github.com/OCA/sale-workflow/pull/2685/files#diff-82d47b9f2dd828758e2be65c5d754fd6349657b7758b996d66b6aef8d41a517fR7-R13

pedrobaeza:
What is the problem with delaying it? You have anyway the attribute `precompute=True` for that, although I think it's not available in 14.0.

HekkiMelody:
@pedrobaeza You were correct, the override of the `create` method on `sale_order_general_discount_triple` was unnecessary. I've removed it and the behaviour doesn't seem to have changed, and all tests still pass.

francesco-ooops (approved):
Functional ok!

@pedrobaeza merge?

pedrobaeza:
I still see the override of the create in the PR...

HekkiMelody:
@pedrobaeza Sorry but try as I might, I can't make the module work without that one override of the create, as the `@depends` doesn't seem to reliably trigger on creation of the order line.

If I remove the override, some tests suddenly stop passing because the discount is not set at all (even if I try to manually `flush()` or `recompute()`) and the behaviour during manual testing is inconsistent as well..

The override of the create was added here:

https://github.com/OCA/sale-workflow/commit/0beed8a0cc186c627de62d8950a059fd745811dc

With the latest commit I've made the call to the `_compute` function more explicit at least.

I'm open to suggestions by anyone about what changes to make, because logically I agree that it should work without manually calling the `_compute` function again, but it doesn't. (I've also tried setting the `discount` field's `default` to `False` but that didn't make any difference.

pedrobaeza:
That test is logical to fail, as you are creating a sales order line directly from a dictionary, and in that case, no `discounts` are changed. Is that a real flow we should cover?

HekkiMelody:
I'm not at all convinced that this is the correct approach, but I've removed the override of the `create` in `sale_order_general_discount` and I've also removed the tests that checked if the discount was immediately set after calling the `create()`

@pedrobaeza

pedrobaeza (approved):
/ocabot merge major

PR #2687, opened 2023-09-13
[16.0][MIG] sale_order_line_input: Migration to version 16.0

carolinafernandez-tecnativa:
- Standard migration to v16

@Tecnativa

TT44354

@pedrobaeza @carlosdauden


carlosdauden (changes requested):
Check that all the fields that are available in the "normal" view are present and check their attributes for similar behaviour.
https://github.com/odoo/odoo/blob/c77a8b4a35fb6ae62bb1824695563857b9e9116f/addons/sale/views/sale_views.xml#L458-L580

carolinafernandez-tecnativa:
> Check that all the fields that are available in the "normal" view are present and check their attributes for similar behaviour. https://github.com/odoo/odoo/blob/c77a8b4a35fb6ae62bb1824695563857b9e9116f/addons/sale/views/sale_views.xml#L458-L580

@carlosdauden fixed it! coudl you please check it now? Thanks!

sale_order_line_input/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    pricelist_id = fields.Many2one(
+        related="order_id.pricelist_id",
+        readonly=True,
```
rousseldenis: This is the default

sale_order_line_input/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    @api.depends("order_id")
+    def _compute_force_company_id(self):
+        """Related company is not computed already when we click create new line"""
+        for line in self:
+            line.force_company_id = (
+                line.order_id.company_id
+                # Is not necessary use browse here
+                or self.env.context.get("force_company")
```
rousseldenis: force_company is deprecated. Moreover I'm not sure about the behavior

carolinafernandez-tecnativa: @pedrobaeza

pedrobaeza: `force_company` indeed it's deprecated. Now you look for the current company at `self.env.company`. What I don't know exactly is the goal of the field `force_company_id`. I suppose that is to transfer it to the `sale.order` header. @carlosdauden can you please clarify?

sale_order_line_input/models/sale_order.py
carolinafernandez-tecnativa: @rousseldenis Odoo only adds pricelist_item_id on model sale.order.line, we should add pricelist_id
@pedrobaeza

rousseldenis: @carolinafernandez-tecnativa I know. What I wanted to say, readonly attribute is True by default on related fields. So, no need to declare it.

carolinafernandez-tecnativa: @rousseldenis done!

sale_order_line_input/models/sale_order.py
carolinafernandez-tecnativa: Hi @carlosdauden did you check this? Thanks!

carolinafernandez-tecnativa:
ping @carlosdauden

sale_order_line_input/models/sale_order.py
carlosdauden: The idea is that the user can enter orders from different companies in the same way as in the normal sales order view.

As the company_id field is a related of order_id.company_id and when starting to enter the order_id line does not exist, the force_company_id field allows the user to select the company so that the customers, products, taxes, etc. are those related to the selected company.

If force_company is never going to come in the context you could remove that line, but maybe some development is using this mechanism to inform the company.

carlosdauden: The idea is to enable the user to input orders from different companies in the same way they do from the regular sales order view.

Since the field company_id is a related field of order_id.company_id, and when starting to input the order_id line, it doesn't exist yet, the force_company_id field allows the user to select the company so that clients, products, taxes, etc., are associated with the chosen company.

If the force_company attribute is never going to be present in the context, you could consider removing that line. However, it's possible that some development might utilize this mechanism to specify the company.

carolinafernandez-tecnativa: @carlosdauden thanks! @pedrobaeza should we keep this field?

pedrobaeza: `force_company`, being deprecated now, shouldn't be kept, and you are already defaulting to `self.env.company`, so remove this line.

carolinafernandez-tecnativa: @pedrobaeza Removed force_company_id

pedrobaeza (approved):

carlosdauden (changes requested):
Why has the force_company_id field functionality been removed?
https://github.com/OCA/sale-workflow/pull/2687#discussion_r1346883401
The only thing that needed to be removed was the line that took the force_company from the context:
`or self.env.context.get("force_company")`

carolinafernandez-tecnativa:
> Why has the force_company_id field functionality been removed? [#2687 (comment)](https://github.com/OCA/sale-workflow/pull/2687#discussion_r1346883401) The only thing that needed to be removed was the line that took the force_company from the context: `or self.env.context.get("force_company")`

Sorry I did not understand, could you please check it now?

carolinafernandez-tecnativa:
ping @carlosdauden

pedrobaeza:
@carolinafernandez-tecnativa please correctly put the changes on the corresponding commits.

pedrobaeza:
ping @carolinafernandez-tecnativa

patel1008b (approved):
LGTM

patel1008b (approved):

PR #2689, opened 2023-09-15
[MIG] sale_order_secondary_unit: Migration to 16.0

matiasperalta1:


rousseldenis:

sale_order_line_menu/README.rst
```diff
@@ -7,7 +7,7 @@ Sale Order Line Menu
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:c42689ec894a9beeba3c383cefbcc056e9807f74c0f0d9ea40af9d741cc37d83
```
SirAionTech: Please do not edit other modules than the one you are migrating if not needed.

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -18,26 +18,42 @@ class SaleOrderLine(models.Model):
     )
 
     product_uom_qty = fields.Float(
-        store=True, readonly=False, compute="_compute_product_uom_qty", copy=True
+        compute="_compute_product_uom_qty",
+        precompute=True,
+        store=True,
```
SirAionTech: Most of these attributes are redundant because already in https://github.com/odoo/odoo/blob/e78f558b2c09c8616b50b92778a06bc9cbb5ae66/addons/sale/models/sale_order_line.py#L114-L118, can we only override the field attributes that need to be changed?

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -18,26 +18,42 @@ class SaleOrderLine(models.Model):
...
     )
 
-    @api.depends("secondary_uom_qty", "secondary_uom_id", "product_uom_qty")
+    secondary_uom_qty = fields.Float(
+        compute="_compute_secondary_uom_qty",
+        precompute=True,
+        store=True,
+    )
```
SirAionTech, line 30: Could you please clarify why this is needed? This is just a subset of the attributes already defined in the mixin (https://github.com/OCA/product-attribute/blob/3d4ee16b8bacfbe9a598b4a35e41e5e2a095f5d2/product_secondary_unit/models/product_secondary_unit_mixin.py#L40-L47).

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -18,26 +18,42 @@ class SaleOrderLine(models.Model):
...
+
+    @api.depends(
+        "display_type",
+        "product_id",
+        "product_packaging_qty",
+        "secondary_uom_qty",
+        "secondary_uom_id",
+        "product_uom_qty",
+    )
```
SirAionTech, line 39: Fields in `api.depends` are incremental (see https://github.com/odoo/odoo/blob/68c8d5012700d04ef253d22be3440b9922f1d8aa/odoo/fields.py#L577), this means that you only need to add the new fields your override needs to depend on, so for instance you don't need to add `product_id` here.

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -18,26 +18,42 @@ class SaleOrderLine(models.Model):
...
-    def product_id_change(self):
+    def _onchange_product_id_warning(self):
         """
         If default sales secondary unit set on product, put on secondary
         quantity 1 for being the default quantity. We override this method,
         that is the one that sets by default 1 on the other quantity with that
         purpose.
         """
-        res = super().product_id_change()
+        res = super()._onchange_product_id_warning()
```
SirAionTech, line 56: In `16.0` this method is not
> the one that sets by default 1 on the other quantity

as stated in the method docstring.
The _other quantity_ (I believe they are talking about `product_uom_qty`) was set by `product_id_change` in `15.0` (https://github.com/odoo/odoo/blob/0f90b8148f76d9aff906bedc93026909bab029ac/addons/sale/models/sale_order_line.py#L624), but in `16.0` it is a computed field (https://github.com/odoo/odoo/blob/e78f558b2c09c8616b50b92778a06bc9cbb5ae66/addons/sale/models/sale_order_line.py#L116) so probably this code goes somewhere else.
Looks like its purpose it to set `secondary_uom_qty` so its place might be in an override of `_compute_secondary_uom_qty`.

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
```diff
@@ -95,16 +95,59 @@ def test_independent_type(self):
...
         self.assertEqual(self.order.order_line.secondary_uom_qty, 4)
-        self.assertEqual(self.order.order_line.secondary_uom_unit_price, 500)
+        self.assertEqual(self.order.order_line.secondary_uom_unit_price, 0.5)
 
         self.order.order_line.write({"product_uom_qty": 8})
         self.assertEqual(self.order.order_line.secondary_uom_qty, 16)
-        self.assertEqual(self.order.order_line.secondary_uom_unit_price, 500)
-        self.assertEqual(self.order.order_line.price_subtotal, 8000)
+        self.assertEqual(self.order.order_line.secondary_uom_unit_price, 0.5)
+        self.assertEqual(self.order.order_line.price_subtotal, 8)
```
SirAionTech: Why is this change needed for the migration?
If the test was wrong in `15.0` already, could you please do this change in a separate commit so it can easily backported? Or at least create an issue so it is known that there is this bug in `15.0`.

sale_order_secondary_unit/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+To configure this module, you need to:
+
+#. Go to *Sales > Configuration > Products > Secondary Units of Measure*
```
SirAionTech, line 3: This menu is activated only when `sale_management` is installed (https://github.com/odoo/odoo/blob/e78f558b2c09c8616b50b92778a06bc9cbb5ae66/addons/sale_management/views/sale_management_menus.xml#L5), maybe we should depend on that one?
Or maybe just add a note here about that.

sale_order_secondary_unit/tests/test_sale_order_secondary_unit.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2018-2020 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import Form, TransactionCase, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestSaleOrderSecondaryUnit(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
SirAionTech, line 10: We are missing https://github.com/OCA/sale-workflow/pull/2742, could you please include it?

SirAionTech (changes requested):
Thanks for the PR!
Please check why the tests are :x: and let me know what you think of the comments below.

We are missing the commit of https://github.com/OCA/sale-workflow/pull/2732, could you please include it?
The added migration script should then be deleted in the migration commit.

Could you please also squash the bots commits as explained in https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate?

alexey-pelykh:
I propose to continue in #3022

pedrobaeza:
Superseded by #3022

PR #2693, opened 2023-09-22
[IMP] sale_stock_picking_blocking: auto-done use case with delivery blocking

sbejaoui:
Instead of forbidding the auto-done process with delivery blocking, this workaround allows orders that should be blocked to remain in the confirmed state and be set to 'done' when the block is removed by the user.


rousseldenis (approved):
Code review

gbrito (approved):

rousseldenis:
@sbejaoui There are changes in 16.0 branch: https://github.com/OCA/sale-workflow/pull/2884/files

sbejaoui:
> @sbejaoui There are changes in 16.0 branch: https://github.com/OCA/sale-workflow/pull/2884/files

rebased

PR #2694, opened 2023-09-22
[14.0][IMP] sale_order_line_chained_move: do not merge propagated moves

HekkiMelody:
This small change means that even in stock moves that are propagated from the initial one, moves for the same product are not merged if they refer to different sale order lines.

Without this change, the moves are merged and the field `related_sale_line_id` is overwritten with the last move processed, losing the link to the other sale order lines.


HekkiMelody:
I am isolating the module with incompatible tests `sale_stock_line_customer_ref` into its own set of tests. 

(I could have isolated this module as well, but this module has dependencies, whereas `sale_stock_line_customer_ref` does not)

francesco-ooops:
@rousseldenis what do you think?

elvise:
I think we need to IMP with an option like this:
IMG_0959

HekkiMelody:
> I think we need to IMP with an option like this: IMG_0959

Implemented this. The default for all rules is that they are kept separate.

francesco-ooops:
@rousseldenis ok, I think we've made good progress. 

Would this solution work for you? Do you have suggestions? thanks :)

francesco-ooops (approved):
Functional ok!

ilyasProgrammer (approved):
LGTM

geomer198 (approved):
LGTM!
Could you add message for each assertion in tests?

HekkiMelody:
done!

francesco-ooops:
@rousseldenis do you think it's good for merge? thanks!

francesco-ooops:
@rousseldenis gentle reminder :)

elvise:
@rousseldenis gentle reminder ;)

francesco-ooops:
@OCA/crm-sales-marketing-maintainers would it be possible to merge? approved since 3 weeks. Thanks!

francesco-ooops:
@pedrobaeza can we go?

.github/workflows/test.yml
```diff
@@ -50,11 +50,18 @@ jobs:
             include: "sale_procurement_group_by_commitment_date"
             name: test with OCB
           - container: ghcr.io/oca/oca-ci/py3.6-odoo14.0:latest
-            exclude: "sale_order_line_sequence,sale_procurement_group_by_commitment_date"
+            include: "sale_stock_line_customer_ref"
```
pedrobaeza: You mustn't touch this file manually, but through copier.

pedrobaeza (changes requested):
Anyway, adding yet another builds to separate modules doesn't seem to me as a good practice (other think on contrary). Instead, the code should inhibit its conflicting behavior if testing other modules. That can be done through a context and checking `config["test_enable"]`.

.github/workflows/test.yml
HekkiMelody: I didn't know, thanks for pointing it out.

simahawk:
> Anyway, adding yet another builds to separate modules doesn't seem to me as a good practice (other think on contrary). Instead, the code should inhibit its conflicting behavior if testing other modules. That can be done through a context and checking `config["test_enable"]`.

A better option, from my POV, is to keep new features OFF by default, provide a global conf for enabling them  and enable them for testing only.

This way you gain 2 things:
1. the feature can be turned on/off fast w/o uninstalling the module (which might cause other issues)
2. no need for test specific code

.github/workflows/test.yml
simahawk: to be rolled back if possible, see comments above

simahawk (changes requested):

francesco-ooops:
> > Anyway, adding yet another builds to separate modules doesn't seem to me as a good practice (other think on contrary). Instead, the code should inhibit its conflicting behavior if testing other modules. That can be done through a context and checking `config["test_enable"]`.
> 
> A better option, from my POV, is to keep new features OFF by default, provide a global conf for enabling them and enable them for testing only.
> 
> This way you gain 2 things:
> 
> 1. the feature can be turned on/off fast w/o uninstalling the module (which might cause other issues)
> 2. no need for test specific code

I also like this approach from a functional POV in order to maintain established flow on databases where module is already installed

@aleuffre we should probably change this field to "Merge propagated moves"

image

HekkiMelody:
> > Anyway, adding yet another builds to separate modules doesn't seem to me as a good practice (other think on contrary). Instead, the code should inhibit its conflicting behavior if testing other modules. That can be done through a context and checking `config["test_enable"]`.
> 
> A better option, from my POV, is to keep new features OFF by default, provide a global conf for enabling them and enable them for testing only.
> 
> This way you gain 2 things:
> 
>     1. the feature can be turned on/off fast w/o uninstalling the module (which might cause other issues)
> 
>     2. no need for test specific code

I don't mind implementing this solution, but the reason why I initially activated the option by default is that, in my opinion, without it the module works in a way that is really misleading. As moves with the same product are merged, the field `related_sale_line_id` only points to one of the possible sale order lines the move originated from, while losing the link to the others.

francesco-ooops:
@rousseldenis what do you think?

@pedrobaeza @simahawk I agree this is an IMP bordering on a FIX

pedrobaeza:
But why do you need to separate CIs with this IMP/FIX?

HekkiMelody:
I've reverted the changes to the CI, and turned the option of not merging moves off by default.

HekkiMelody:
I don't think the pre-commit failure is due to my code. Possibly due to the update to python 3.12 in the CI.

francesco-ooops:
@aleuffre rebase pls

francesco-ooops:
@simahawk @pedrobaeza ping :)

pedrobaeza (approved):

simahawk (approved):

PR #2696, opened 2023-09-25
[ADD] sale_probability_amount: New module

matthieusaison:
This module adds possibility to evaluate the probability to win a quotation.

You can show the total amount of sale in the currency of your company in the bottom of quotation tree view.

### DIFF
--- a/sale_probability_amount/models/sale_order.py
+++ b/sale_probability_amount/models/sale_order.py
@@ -27,16 +27,19 @@ class SaleOrder(models.Model):
         )
 
     def action_confirm(self):
+        res = super().action_confirm()
         self.probability = 100
-        return super().action_confirm()
+        return res
 
     def action_cancel(self):
+        res = super().action_cancel()
         self.probability = 0
-        return super().action_confirm()
+        return res
 
     def action_draft(self):
+        res = super().action_draft()
         self.probability = self._get_default_value()
-        return super().action_confirm()
+        return res
 
     @api.depends("amount_total_curr", "probability")
     def _compute_expected_amount_cur(self):
--- a/sale_probability_amount/models/sale_order.py
+++ b/sale_probability_amount/models/sale_order.py
@@ -1,0 +1,38 @@
+# Copyright 2023 Akretion (https://www.akretion.com).
+# @author Matthieu SAISON <matthieu.saison@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    percentage = fields.Integer("Winning sale %", default=50)
+    expected_amount_cur = fields.Monetary(
+        "Expected Amount in currency",
+        compute="_compute_expected_amount_cur",
+        currency_field="company_currency_id",
+        store=True,
+    )
+
+    def action_confirm(self):
+        res = super().action_confirm()
+        if res:
+            self.percentage = 100
+        else:
+            return res
+
+    def action_cancel(self):
+        res = super().action_cancel()
+        if res:
+            self.percentage = 0
+        else:
+            return res
+
+    @api.depends("amount_total_curr", "percentage")
+    def _compute_expected_amount_cur(self):
+        for record in self:
+            record.expected_amount_cur = (
+                record.amount_total_curr * record.percentage / 100
+            )
--- a/sale_probability_amount/models/sale_order.py
+++ b/sale_probability_amount/models/sale_order.py
@@ -9,11 +9,7 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     probability = fields.Integer(
-        "Winning sale %",
-        default=lambda self: self.env.ref(
-            "sale_probability_amount.quotation_default_probability",
-            raise_if_not_found=False,
-        ).value,
+        "Winning sale %", default=lambda self: self._get_default_value()
     )
 
     expected_amount_cur = fields.Monetary(
@@ -22,6 +18,13 @@ class SaleOrder(models.Model):
         currency_field="company_currency_id",
         store=True,
     )
+
+    def _get_default_value(self):
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("quotation_default_probability", 50)
+        )
 
     def action_confirm(self):
         self.probability = 100
@@ -32,7 +35,7 @@ class SaleOrder(models.Model):
         return super().action_confirm()
 
     def action_draft(self):
-        self.probability = 50
+        self.probability = self._get_default_value()
         return super().action_confirm()
 
     @api.depends("amount_total_curr", "probability")
--- a/sale_probability_amount/models/sale_order.py
+++ b/sale_probability_amount/models/sale_order.py
@@ -8,7 +8,14 @@ from odoo import api, fields, models
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    percentage = fields.Integer("Winning sale %", default=50)
+    probability = fields.Integer(
+        "Winning sale %",
+        default=lambda self: self.env.ref(
+            "sale_probability_amount.quotation_default_probability",
+            raise_if_not_found=False,
+        ).value,
+    )
+
     expected_amount_cur = fields.Monetary(
         "Expected Amount in currency",
         compute="_compute_expected_amount_cur",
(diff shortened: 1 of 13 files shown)


sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+# @author Matthieu SAISON <matthieu.saison@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    percentage = fields.Integer("Winning sale %", default=50)
```
yostashiro: I think `probability` would be better as the field name, to be consistent with the field in crm.lead.

It would also be nice to make the default configurable.

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+        currency_field="company_currency_id",
+        store=True,
+    )
+
+    def action_confirm(self):
+        res = super().action_confirm()
+        if res:
+            self.percentage = 100
+        else:
+            return res
```
yostashiro:
```suggestion
        self.percentage = 100
        return super().action_confirm()
```

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def action_cancel(self):
+        res = super().action_cancel()
+        if res:
+            self.percentage = 0
+        else:
+            return res
```
yostashiro:
```suggestion
        self.percentage = 0
        return super().action_cancel()
```

How `about action_draft()`?  Is it a design choice not to extend it?

sale_probability_amount/views/sale_views.xml
```diff
@@ -0,0 +1,28 @@
...
+
+    <record id="view_order_form" model="ir.ui.view">
+            <field name="model">sale.order</field>
+            <field name="inherit_id" ref="sale.view_order_form" />
+            <field name="arch" type="xml">
+                <field name="order_line" position="after">
+                    <group
+                    colspan="4"
+                    attrs="{'invisible':[('state', 'not in', ['draft', 'sent'])]}"
+                >
```
yostashiro, line 22:
```suggestion
                    <group
                        colspan="4"
                        attrs="{'invisible':[('state', 'not in', ['draft', 'sent'])]}"
                    >
```

sale_probability_amount/models/sale_order.py
matthieusaison: Thank for reviewing, I'll made this changes

sale_probability_amount/models/sale_order.py
matthieusaison: I forgot this one, I add it in next commit

sale_probability_amount/views/sale_views.xml
matthieusaison: The pre-commit prettier hook seems to force this indentation.

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    probability = fields.Integer(
+        "Winning sale %",
+        default=lambda self: self.env.ref(
```
sebastienbeau: self.env['ir.config_parameter'].sudo().get_param('quotation_default_probability', 50)

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+            "sale_probability_amount.quotation_default_probability",
+            raise_if_not_found=False,
+        ).value,
```
yostashiro: This is throwing an error in runboat build.  I suppose we should have a get-default method to handle the situation where the record has yet to exist?  We can't do `.value` anyway as it breaks the system if you delete the parameter record.

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+    def action_confirm(self):
+        self.probability = 100
+        return super().action_confirm()
+
+    def action_cancel(self):
+        self.probability = 0
+        return super().action_confirm()
+
+    def action_draft(self):
+        self.probability = 50
```
sebastienbeau: TODO reset to default value

sale_probability_amount/views/sale_views.xml
```diff
@@ -0,0 +1,28 @@
...
```
yostashiro, line 22:
```suggestion
                    <group
                        colspan="4"
                        attrs="{'invisible':[('state', 'not in', ['draft', 'sent'])]}"
                    >
```

sale_probability_amount/models/sale_order.py
yostashiro: Shouldn't we get the value from the config parameter?

sale_probability_amount/views/sale_views.xml
matthieusaison: I'm ok with this notation but the pre-commit hook named:
prettier (with plugin-xml) force this indentation.

sale_probability_amount/models/sale_order.py
matthieusaison: yes, done

sale_probability_amount/views/sale_views.xml
yostashiro: @matthieusaison Can you try manually adjusting the indentation before running pre-commit?  I suppose pre-commit won't force the current indentation as long as you have it right upfront.

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+            .get_param("quotation_default_probability", 50)
+        )
+
+    def action_confirm(self):
+        self.probability = 100
+        return super().action_confirm()
+
+    def action_cancel(self):
+        self.probability = 0
+        return super().action_confirm()
```
Kev-Roche:
```suggestion
        return super().action_cancel()
```

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def action_draft(self):
+        self.probability = self._get_default_value()
+        return super().action_confirm()
```
Kev-Roche:
```suggestion
        return super().action_draft()
```

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+    def _get_default_value(self):
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("quotation_default_probability", 50)
+        )
+
+    def action_confirm(self):
+        self.probability = 100
+        return super().action_confirm()
```
Kev-Roche: Here and for action_cancel and action_draft, we need to call the super() before changing the probability.

Kev-Roche (changes requested):

sale_probability_amount/views/sale_views.xml
matthieusaison: I pushed it, and it works. The pre-commit issue is only on my laptop.

Kev-Roche (approved):

sale_probability_amount/views/sale_views.xml
```diff
@@ -0,0 +1,28 @@
...
+
+    <record id="view_order_form" model="ir.ui.view">
+            <field name="model">sale.order</field>
+            <field name="inherit_id" ref="sale.view_order_form" />
+            <field name="arch" type="xml">
+                <field name="order_line" position="after">
+                <group
+                    colspan="4"
+                    attrs="{'invisible':[('state', 'not in', ['draft', 'sent'])]}"
+                >
```
yostashiro: Sorry to be pedantic but let's get this right as we are almost there ;)

```suggestion
                    <group
                        colspan="4"
                        attrs="{'invisible':[('state', 'not in', ['draft', 'sent'])]}"
                    >
```

sale_probability_amount/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("quotation_default_probability", 50)
+        )
+
+    def action_confirm(self):
+        res = super().action_confirm()
+        self.probability = 100
+        return res
```
yostashiro, line 32: Just curious.  Was there a reason not to do:

```python
        self.probability = 100
```

matthieusaison: sale_tier_validation don't pass the test if we set the probability before action_confirm. We can't change a sale under validation.

sebastienbeau (approved):
LGTM, @matthieusaison can you fix the pre-commit ?

sale_probability_amount/views/sale_views.xml
matthieusaison: https://github.com/OCA/sale-workflow/actions/runs/6390259999/job/17343158634?pr=2696#step:7:57

PR #2697, opened 2023-09-25
[14.0][ADD] sale_invoice_delivery_state

chafique-delli:
This module allows:
- to indicate on a partner, if you want to invoice it after it is partially delivered or fully delivered.
- to verify that the invoice status of the order is correct based on the customer's invoice policy and the delivery status of the order.


chafique-delli:
@Kev-Roche

chafique-delli:
@sebastienbeau

sale_invoice_delivery_state/models/partner.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    invoice_policy = fields.Selection(
+        [("partially", "Partially Delivered"), ("fully", "Fully Delivered")],
+        string="Invoice Policy",
```
gbrito: You can remove the string from the field definition, odoo by default will use the field name as label

sale_invoice_delivery_state/models/sale.py
```diff
@@ -0,0 +1,18 @@
...
+
+    @api.depends("delivery_state")
+    def _get_invoice_status(self):
+        super()._get_invoice_status()
+        for sale in self.filtered(lambda s: s.invoice_status == "to invoice"):
+            if (
+                sale.partner_id.invoice_policy == "fully"
+                and sale.delivery_state != "done"
+            ):
+                sale.invoice_status = "no"
```
gbrito: you could write in all the orders at once, no need to loop through all
```suggestion
        open_orders = self.filtered(
            lambda s: s.invoice_status == "to invoice"
            and s.partner_id.invoice_policy == "fully"
            and s.delivery_state != "done"
        )
        open_orders.write({"invoice_status": "no"})
```

sale_invoice_delivery_state/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+
+{
+    "name": "Sale Invoice Delivery State",
+    "version": "14.0.1.0.0",
+    "category": "Sales Management",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "sale_delivery_state",
```
gbrito, line 12:
```suggestion
        "account",
        "sale_delivery_state",
```

gbrito (changes requested):

sale_invoice_delivery_state/models/partner.py
chafique-delli: @gbrito , it's done.

sale_invoice_delivery_state/models/sale.py
chafique-delli: @gbrito , it's done.

sale_invoice_delivery_state/__manifest__.py
chafique-delli: @gbrito , sale_delivery_state already indirectly depends on account

sale_invoice_delivery_state/views/partner.xml
```diff
@@ -0,0 +1,11 @@
+<odoo>
+    <record id="view_partner_property_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="account.view_partner_property_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='property_payment_term_id']" position="after">
+                <field name="invoice_policy" />
+            </xpath>
```
Kev-Roche: To be consistent with other modules like account_invoice_base_invoicing_mode, I'll advise to put the new field in the accounting page.

chafique-delli: @Kev-Roche , it's done.

Kev-Roche (approved):

sale_invoice_delivery_state/models/sale.py
```diff
@@ -0,0 +1,18 @@
...
+
+    @api.depends("delivery_state")
+    def _get_invoice_status(self):
+        super()._get_invoice_status()
+        open_orders = self.filtered(
+            lambda s: s.invoice_status == "to invoice"
+            and s.partner_id.invoice_policy == "fully"
+            and s.delivery_state != "done"
+        )
+        open_orders.write({"invoice_status": "no"})
```
sebastienbeau: Do not use write in computed field, this can generate weird behavior.
You can use a "update", (the update will loop on each record to assign the value)

As the filtered is already doing a loop, I will recommends to  just do a loop with an if

```suggestion
        for record in self:
            if (record.invoice_status == "to invoice"
                 and record.partner_id.invoice_policy == "fully"
                 and record.delivery_state != "done"
                 ):
            record.invoice_status = "no"
```

chafique-delli: @sebastienbeau , it's fixed.

sebastienbeau (approved):
LGTM (code review)

PR #2700, opened 2023-09-27
[14.0][IMP] sale_order_type, added init hook to handle big db install

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/sale_order_type/hooks.py
+++ b/sale_order_type/hooks.py
@@ -1,0 +1,20 @@
+import logging
+
+from odoo.api import SUPERUSER_ID, Environment
+
+logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    Environment(cr, SUPERUSER_ID, {})
+    logger.info("Add type_id field in sale_order table")
+    cr.execute("ALTER TABLE sale_order ADD COLUMN IF NOT EXISTS type_id INTEGER")
+
+
+def post_init_hook(cr, registry):
+    env = Environment(cr, SUPERUSER_ID, {})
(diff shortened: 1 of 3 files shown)


JordiMForgeFlow (approved):
LGTM 👍🏼

LoisRForgeFlow (approved):

sale_order_type/hooks.py
```diff
@@ -0,0 +1,20 @@
...
+    logger.info("Add type_id field in sale_order table")
+    cr.execute("ALTER TABLE sale_order ADD COLUMN IF NOT EXISTS type_id INTEGER")
+
+
+def post_init_hook(cr, registry):
+    env = Environment(cr, SUPERUSER_ID, {})
+    order_type = env["sale.order"]._default_type_id()
+    if order_type:
+        cr.execute(
+            "UPDATE sale_order set type_id = %s where type_id is null", (order_type.id,)
```
pedrobaeza, line 19: Maybe is better to not fill the field at all?

AaronHForgeFlow: If the column is created without a value, will the compute method  be triggered?

pedrobaeza: No, as you have already prevented it in the pre hook

ChrisOForgeFlow: Because this we can't left empty field, is required on view, and empty can struggle current record, can't print, etc

https://github.com/OCA/sale-workflow/blob/14.0/sale_order_type/views/sale_order_view.xml#L8

pedrobaeza: OK then.

PR #2717, opened 2023-10-10
[14.0][ADD] sale_line_service_qty_delivered: Set qty_delivered of service lines

mt-software-de:
As soon as a sale.order.line is delivered
the qty_delivered of the siblings service lines
is set to its max qty


sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def _is_delivered_method_delivery(self):
+        self.ensure_one()
+        return (
+            self.product_id.type == "service"
+            and self.product_id.invoice_policy == "delivery"
+        )
+
+    def _compute_qty_delivered_method(self):
+        super(SaleOrderLine, self)._compute_qty_delivered_method()
```
vuwnevska:
```suggestion
        super()._compute_qty_delivered_method()
```

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,53 @@
...
+            else:
+                delivered = any(
+                    not float_is_zero(line.qty_delivered, precision)
+                    for line in lines["stock_move"]
+                )
+            for line in lines["delivery"]:
+                line.qty_delivered = line._get_delivery_qty_delivered(delivered)
+
+    def _compute_qty_delivered(self):
+        super(SaleOrderLine, self)._compute_qty_delivered()
```
vuwnevska:
```suggestion
        super()._compute_qty_delivered()
```

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,53 @@
...
+            return self.product_uom_qty
+        return 0.0
+
+    def _compute_qty_delivered_delivery_method(self):
+        precision = self.env["decimal.precision"].precision_get(
+            "Product Unit of Measure"
+        )
+        lines_grouped = groupby(self, lambda l: l.order_id)
+        for order, lines in lines_grouped:
+            lines = dict(groupby(order.order_line, lambda l: l.qty_delivered_method))
```
vuwnevska:
```suggestion
            lines = dict(groupby(lines, lambda l: l.qty_delivered_method))
```
also before grouping better to sort records

vuwnevska (commented):
AFAIS 
https://github.com/OCA/delivery-carrier/blob/14.0/delivery_auto_refresh/models/stock_picking.py
solves the similar issue of the delivery update on outgoing picking  done

mt-software-de:
> AFAIS https://github.com/OCA/delivery-carrier/blob/14.0/delivery_auto_refresh/models/stock_picking.py solves the similar issue of the delivery update on outgoing picking done

Not really.

rousseldenis:
@mt-software-de What's the status of this ?

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: lines can not be used, order.order_line needs to be used, 
to ensure all of the lines of a order are checked. 

And sorting is not needed when using default odoo groupby

mt-software-de:
> @mt-software-de What's the status of this ?

I checked the failing tests, they were coming from `sale_order_invoicing_finished_task`, because they interfere with each other. 

Also check the code of the new addon. 
I did like that, if there is a stock_move sale.order.line and a delivery sale.order.line the delivered qty of the delivery line was set to its max when the stock.move sale.order.line was delivered.
But when there is a order with just a delivery sale.order.line this line was always set marked as delivered,
with the commit https://github.com/OCA/sale-workflow/pull/2717/commits/867cbc94b9c6dcb87c5981217c20191fc2be16f2 i changed it.
So now if there is just a delivery sale.order.line the delivered qty is 0. 

@vuwnevska @jbaudoux @rousseldenis 
What do you think is better?

sale_line_service_qty_delivered/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+            if order.state not in ("sale", "done"):
+                continue
+            if order.invoice_status != "no":
+                continue
+            if any(
+                line._is_delivered_method_delivery()
+                and line.invoice_status == to_invoice_state
+                for line in order.order_line
+            ):
+                order.invoice_status = to_invoice_state
```
jbaudoux: This has nothing to do with the module purpose ??

In standard, a SO is not `to invoice` if only the delivery line is to invoice. This should not be changed. You need at least another line to invoice to be able to invoice the SO

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,60 @@
...
+
+    qty_delivered_method = fields.Selection(
+        selection_add=[(QTY_DELIVERED_METHOD_DELIVERY, "Delivery")]
+    )
+
+    def _is_delivered_method_delivery(self):
+        self.ensure_one()
+        return (
+            self.product_id.type == "service"
+            and self.product_id.invoice_policy == QTY_DELIVERED_METHOD_DELIVERY
```
jbaudoux: It's different fields with different selections, don't use the same variable. Just put the value here
```suggestion
            and self.product_id.invoice_policy == "delivery"
```

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,60 @@
...
+    def _compute_qty_delivered_method(self):
+        super()._compute_qty_delivered_method()
+        for line in self:
+            if line._is_delivered_method_delivery():
+                line.qty_delivered_method = QTY_DELIVERED_METHOD_DELIVERY
+
+    def _get_delivery_qty_delivered(self, delivered):
+        self.ensure_one()
+        if delivered:
+            return self.product_uom_qty
```
jbaudoux: Here you made a choice that the service is delivered as soon as there is a stock line delivered. But it could be when all is delivered. Or proportionally.
There should not be any argument to the method. The computation of delivered it part of what your qty computation is doing here.

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,60 @@
...
+    _inherit = "sale.order.line"
+
+    qty_delivered_method = fields.Selection(
+        selection_add=[(QTY_DELIVERED_METHOD_DELIVERY, "Delivery")]
+    )
+
+    def _is_delivered_method_delivery(self):
+        self.ensure_one()
+        return (
+            self.product_id.type == "service"
```
jbaudoux, line 17: All service lines?

jbaudoux (changes requested):
It looks to me that you do what is in standard (at least in v16).
If you configure the delivery service as invoiced at order, then it will be invoiced as soon as there is a stock line delivered.

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: no not all, only if the `invoice_policy` of the product is `delivery`

sale_line_service_qty_delivered/models/sale_order.py
mt-software-de: For lines which are is_delivery and invoice_policy == 'delivery' and type == 'service' they should only be invoice when something is shipped.
But for Service Products which are not added via deliver costs so is_delivery == False and invoice_policy == 'delivery' and type == 'service' it should be possible to invoice if there are no other sale.order.lines.

I will extend it.

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: Yes i will change it.

sale_line_service_qty_delivered/models/sale_order.py
mt-software-de: This is needed, because if there is no stock_move sale.order.line which is `to_invoice` the sale.order invoice_status will be `no`. 
In the case when you have two sale.order.lines one with `delivered_method` `stock_move` one with `delivery` and the `invoice_status` of the order is `to invoice`. If you the create the invoice and remove the invoice line for `delivery` line. The sale.order `invoice_status` will be `no` and that is not correct. 
https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/delivery/models/sale_order.py#L156

jbaudoux: Add a note about that in the readme as this is not like the default behavior for delivery cost lines

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,57 @@
...
+            contains_stock_move_sale_line = True
+            if not float_is_zero(line.qty_delivered, precision):
+                is_delivered = True
+                break
+        # Only return a delivered qty if there something delivered
+        # or if the current line is not a is_delivery sale.order line
+        # and the sale.order not contains a sale.order.line with a delivered_method = stock_move
+        # Because we will should change the behavior for is_delivery lines
+        if is_delivered or not contains_stock_move_sale_line and not self.is_delivery:
+            qty_delivered = self.product_uom_qty
```
jbaudoux, line 50: would have been nice to configure on the product if the delivered quantity is 
- ordered as soon as there is delivery
- prorata of the delivered quantity
- ordered when all is delivered

jbaudoux (approved):

sale_line_service_qty_delivered/models/sale_order.py
mt-software-de: I changed the readme.
I didn't explicitly point this out because that's the reason for the add-on.

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: Yes, maybe for the future. :-)

mt-software-de:
@jbaudoux please check https://github.com/OCA/sale-workflow/pull/2717/commits/fca445c7199ce43889af6c93fc1e207b5d13be13
 If a line is delivered_method = 'delivery' (the new selection value). It will always be treated as in the purpose of the module. It will not make a difference if it is `is_delivery` or not. 
 
 If you can't think of anything against it, i would do a squash and ask for a merge.

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+        return (
+            self.product_id.type == "service"
+            and self.product_id.invoice_policy == "delivery"
+        )
+
+    def _compute_qty_delivered_method(self):
+        super()._compute_qty_delivered_method()
+        for line in self:
+            if line._is_delivered_method_delivery():
+                line.qty_delivered_method = "delivery"
```
jbaudoux, line 26: That will break module like sale_timesheet no? referring to my comment https://github.com/OCA/sale-workflow/pull/2717/files#r1944366893
SO line on timesheet that are invoiced as the time is booked are product type service and invoice policy delivery I think

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+        is_delivered = False
+        for line in self.order_id.order_line:
+            if line.qty_delivered_method != "stock_move":
+                continue
+            contains_stock_move_sale_line = True
+            if not float_is_zero(line.qty_delivered, precision):
+                is_delivered = True
+                break
+        # Only return a delivered qty if there something delivered
+        # or the sale.order not contains a sale.order.line with a delivered_method = stock_move
```
jbaudoux:
```suggestion
        # Only return a delivered qty if there is something delivered
        # or if the sale.order does not contains a sale.order.line with a delivered_method = stock_move
```

sale_line_service_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+            "Product Unit of Measure"
+        )
+        qty_delivered = 0
+        contains_stock_move_sale_line = False
+        is_delivered = False
+        for line in self.order_id.order_line:
+            if line.qty_delivered_method != "stock_move":
+                continue
+            contains_stock_move_sale_line = True
+            if not float_is_zero(line.qty_delivered, precision):
```
jbaudoux, line 43: what if the product invoice policy is on order ?

sale_line_service_qty_delivered/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This addon automatically increases or decreases the delivered quantity
+for sale.order.lines where a service product with a invoice policy delivery is set.
+A order line's delivered quantity is increase to its max, if an other line is shipped
+or if there is no order line to ship.
+A order line's delivered quantity is decreased to zero, if everthing is returned.
```
jbaudoux:
```suggestion
An order line's delivered quantity is increased to its max as soon as another line with a stock product is delivered or if there is no line with stock product.
An order line's delivered quantity is decreased to zero when all lines with stock product are entirely returned.
```

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: Thx. It will break sale_timesheet if the `invoice_policy` is `delivery` and the `service_type` is `timesheet`.
I will add a check for `service_type`

sale_line_service_qty_delivered/models/sale_order_line.py
mt-software-de: For which line? For the service line? Then this will not be triggered.

jbaudoux: no for the stock line. If it's invoicing is on order, do you consider it is delivered even if the qty delivered is 0? It would make sense

mt-software-de: but then you would simple, configure the service product with a `invoice_policy` on order.

jbaudoux: Your service could be on delivery and then based on the other products you put in the order, that service is delivered or not.
If the purpose is to track products on delivery, you maybe also want to exclude the products on order. Or you want to track the delivery of all stock products, no matter their invoice policy is. It's a question of definition.
I guess it doesn't really matters as you don't have such use case....

mt-software-de: For me there is not such a use case. 
I could check not only for other lines `qty_delivered_method != "stock_move"` it then also check the `invoice_policy` and if it is not `delivery` just ignore those lines.

mt-software-de: You mean like that. https://github.com/OCA/sale-workflow/pull/2717/commits/7964ca40d625c63ef3f00a36618a9ba9c26092c2

jbaudoux (approved):

mt-software-de:
@OCA/crm-sales-marketing-maintainers could someone please merge this PR? Thx

rousseldenis (approved):

PR #2719, opened 2023-10-11
[15.0][IMP]sale_quotation_number: Boolean to determine whether quotation sequence has been used

manuelregidor:
With the current module, the sale order number is only changed from the quotation sequence to the sale order sequence if the first two first characters of the sale order number in draft state are "SQ", which is the default prefix. If this sequence is edited, the sale order number in quotation state will be kept when the sale order is validated and it will not use the sequence for validated sale orders.

Solution: a boolean field has been added to sale orders to indicate whether the quotation sequence has been used. This field is set to True when a new sale order is created and will return to False when the sale order is validated. When the sale order is validated, it will be checked whether the used sequence is the quotation sequence using this new field. If so, the sale order number will be changed. With this solution, the change of the number will be applied no matter what is the sequence configuration. It would also be interesting to add a field to res.company to set the sale quotation sequence that has to be used.

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -4,17 +4,21 @@
 # © 2020 Manuel Regidor  <manuel.regidor@sygel.es>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    quotation_seq_used = fields.Boolean(
+        string="Quotation Sequence Used", default=False, copy=False, readonly=True
+    )
+
     @api.model
     def create(self, vals):
         if self.is_using_quotation_number(vals):
             sequence = self.env["ir.sequence"].next_by_code("sale.quotation")
-            vals["name"] = sequence or "/"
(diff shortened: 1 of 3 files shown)


manuelregidor:

HaraldPanten (approved):
Functional review

ValentinVinagre (changes requested):
Migration script please 😄

manuelregidor:
> Migration script please 😄

done

sale_quotation_number/models/sale_order.py
```diff
@@ -4,17 +4,22 @@
...
+    quotation_seq_used = fields.Boolean(
+        string="Quotation Sequence Used", default=False, copy=False, readonly=True
+    )
+
     @api.model
     def create(self, vals):
         if self.is_using_quotation_number(vals):
             sequence = self.env["ir.sequence"].next_by_code("sale.quotation")
             vals["name"] = sequence or "/"
+            vals["quotation_seq_used"] = True
```
ValentinVinagre: Better to do it in a write.

manuelregidor: Why should we do create+write when we can do just create and set the value to True when necessary?

ValentinVinagre: ups, I meant update()

ValentinVinagre (approved):
LGTM 👍🏻

PR #2720, opened 2023-10-11
[14.0] [IMP] sale_order_product_assortment: Check products on write and create

renda-dev:
This PR aims to fix the following scenario:

- create assortment 1, allowed product: Product 1, allowed partner: Partner 1
- create assortment 2, allowed product: Product 2, allowed partner: Partner 2

Create SO, select partner 1, add product 1 to SOL > change partner to Partner 2 > save

With these changes if there are products that are not allowed by the assortment, it will raise an user error.

### DIFF
--- a/sale_order_product_assortment/models/sale_order.py
+++ b/sale_order_product_assortment/models/sale_order.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Tecnativa - Carlos Roca
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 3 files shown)


sale_order_product_assortment/models/sale_order.py
```diff
@@ -37,3 +38,39 @@ def _compute_product_assortment_ids(self):
...
+    def write(self, vals):
+        res = super().write(vals)
+        self._validate_allowed_products()
+        return res
+
+    def _validate_allowed_products(self):
+        """
+        Check if all the products in the order lines
+        contains products that are allowed for the partner
+        """
```
CarlosRoca13: Is it possible to make this using @api.constrains instead of overwriting create and write?

renda-dev: Good point, thank you.

CarlosRoca13 (approved):
LGTM Thanks :smile:

francesco-ooops (approved):
Functional ok!

@CarlosRoca13 would you mind merging? :)

pedrobaeza:
This is impacting on the general performance, which is not very good with assortments, and only applicable if you create the line via code, so I'm not sure this should be included. @carlosdauden your thoughts?

renda-dev:
Tried to optimize it the more that I could, the other way around this could be an `@api.onchange` on `allowed_product_ids` to raise an error while editing the sale order.

Please let me know what you think

sale_order_product_assortment/models/sale_order.py
```diff
@@ -37,3 +38,25 @@ def _compute_product_assortment_ids(self):
                     )
             if self.allowed_product_ids:
                 self.has_allowed_products = True
+
+    @api.constrains("allowed_product_ids", "order_line.product_id")
```
carlosdauden:
```suggestion
    @api.onchange("partner_id", "partner_shipping_id", "partner_invoice_id")
```
I would prefer this check to be carried out when changing the partner, as this is really the source of the problem.

carlosdauden (changes requested):

francesco-ooops (approved):
Functional still good!

carlosdauden (approved):

CarlosRoca13:
/ocabot merge patch 

Please set this feature to upper versions :) Thanks

PR #2721, opened 2023-10-11
[14.0] [FIX] sale_quick: multi company

renda-dev:
Simple PR to set a domain in the "quick add" view to filter out products that do not belong to the SO's company and products not for sale.

### DIFF
--- a/sale_quick/models/sale_order.py
+++ b/sale_quick/models/sale_order.py
@@ -4,6 +4,7 @@
 
 from odoo import _, models
 from odoo.exceptions import ValidationError
+from odoo.tools.safe_eval import safe_eval
 
 
 class SaleOrder(models.Model):
@@ -26,7 +27,8 @@ class SaleOrder(models.Model):
         return res
 
     def _get_domain_add_products(self):
-        return []
+        field = self.env["sale.order.line"]._fields.get("product_id")
+        return safe_eval(field.domain, {"company_id": self.company_id.id})
 
     def add_product(self):
         self.ensure_one()


francesco-ooops (approved):
Functional ok!

@renda-dev let's mark it as a FIX

legalsylvain (changes requested):
I dont think it's a generic change.
Some people could want to sale product from other company, or without company.

I think you should change a rule in your instance.

francesco-ooops:
Hi @legalsylvain , I think we're just aligning behavior of sale_quick to standard odoo sale order line domain, as you can't add a product belonging exclusively to Company 1 to a sale order belonging to Company 2. 

Indeed, this is the error we're fixing with current version:

image

Products with no company assigned can still be added to any sale order line.

legalsylvain:
Oh ! I see ! So you're right, but i propose a différent implémentation : 
- get the domain defined in the field product_id https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py#L1396
- return it.

That way : 
- it avoid to duplicate code
- if domain of sale.order.line. product_id changes, the domain in sale quick will change automatically.

legalsylvain (approved):
Thanks ! 

CC : @bealdav.

francesco-ooops:
@pedrobaeza merge please?

sale_quick/models/sale_order.py
```diff
@@ -26,7 +27,8 @@ def _get_context_add_products(self):
         return res
 
     def _get_domain_add_products(self):
-        return []
+        field = self.env["sale.order.line"]._fields.get("product_id")
+        return safe_eval(field.domain, {"company_id": self.company_id.id})
```
pedrobaeza, line 31: This is missing possible fields values, as domains may be "dynamic". Example: `"[('type_id', '=', type_id)]"`

pedrobaeza (dismissed):

sale_quick/models/sale_order.py
renda-dev: Thanks for the feedback, however, I'm not sure how to go about fixing it... What would you suggest?

legalsylvain: > This is missing possible fields values, as domains may be "dynamic". Example: "[('type_id', '=', type_id)]"

You're right.
however, I think that this case is quite theoritical and will not occures. and if it occures, it should be fixed by a custom glue module.
I mean, if a custom module add a domain like "[('type_id', '=', type_id)]" and if sale_quick is installed, the custom module (or another glue module) should update the function. _get_domain_add_products.
(with or without this patch).

For me, this implementation is OK.

I see it also robust because if the domain changes in the sale module, we have nothing to do in the OCA module.

pedrobaeza: OK, I dismiss my request changes if you want to go ahead, but I don't think this solution is good enough. Maybe this can be handled at `fields_view_get` level.

francesco-ooops: @pedrobaeza would you like for us to add a line in roadmap?

pedrobaeza: As you prefer. I just refrain for this.

francesco-ooops:
@OCA/crm-sales-marketing-maintainers anyone available for merge? 

Or if you have ideas about another implementation, we're open to find a solution that makes everyone happy

francesco-ooops:
@damdam-s as module creator, do you have any input on how to implement this in the best way? thank you!

francesco-ooops:
@yajo do you have any suggestion on how to approach this the best way?

francesco-ooops:
@simahawk do you have any suggestions on how to get out of this deadlock? appreciate it ;)

francesco-ooops:
@pedrobaeza seems your comment was blocking indeed 😓

francesco-ooops:
@legalsylvain can you help us get unstuck? appreciate it

legalsylvain:
Hi. i'm not a PSC of this repo. I have no merge right. should wait for PSC / maintainer of the module.
regards.

francesco-ooops:
@legalsylvain thank you, I was not asking for a merge but rather on whether you could think of a solution that would please @pedrobaeza since you suggested this approach in the early steps of implementation 

unfortunately I can't seem to get @OCA/crm-sales-marketing-maintainers to merge since a month even if Pedro has marked his comment as not blocking :/

francesco-ooops:
@pedrobaeza can you merge this anyway please?

pedrobaeza:
OK, I think a crash will appear while evaluing the domain that way, but let's see:

/ocabot merge patch

PR #2722, opened 2023-10-13
[14.0][IMP] sale_stock_return_request: optional sales order filter / shortcut from sales order

dalonsod:
Sometimes seleting some orders to be returned for a certain customer could be useful, specially when we want return a sales order with a lot of pickings. Adding a filter and/or having a shortcut from sale order it could be performed.

Optional sales orders filter preview:

imagen

Available action from an order ready to be returned:

imagen

From this button, request return is created with initial data (partner, returnable products) filled:

imagen

### DIFF
--- a/sale_stock_return_request/models/stock_return_request.py
+++ b/sale_stock_return_request/models/stock_return_request.py
@@ -1,17 +1,31 @@
 # Copyright 2019 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.osv import expression
 
 
 class StockReturnRequest(models.Model):
     _inherit = "stock.return.request"
 
+    filter_sale_order_ids = fields.Many2many(
+        comodel_name="sale.order",
+        relation="stock_return_request_so_filter",
+        string="Selected order(s)",
+        copy=False,
+        domain=[("state", "in", ["sale", "done"])],
+    )
     sale_order_ids = fields.Many2many(
         comodel_name="sale.order",
         string="Involved Sales",
         readonly=True,
         copy=False,
     )
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id_sale_stock_return_request(self):
+        self.filter_sale_order_ids = self.filter_sale_order_ids.filtered(
+            lambda x: x.partner_id == self.partner_id
+        )
 
     def _prepare_move_default_values(self, line, qty, move):
         """Extend this method to add values to return move"""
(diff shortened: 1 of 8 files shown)


dalonsod:
@cesar-tecnativa @chienandalu could you review? Thanks!

dalonsod:
@JoanSForgeFlow could you review? Thanks!

sale_stock_return_request/models/stock_return_request.py
```diff
@@ -38,3 +51,19 @@ def action_view_sales(self):
...
         return action
+
+
+class StockReturnRequestLine(models.Model):
+    _inherit = "stock.return.request.line"
+
+    def _get_moves_domain(self):
+        domain = super()._get_moves_domain()
+        if self.request_id.filter_sale_order_ids:
+            domain += [
```
chienandalu: **nitpick** : use `expression.AND`

chienandalu (commented):
Hi @dalonsod Thanks for the contribution! I like the idea. Just a minor nitpick :)

sale_stock_return_request/models/stock_return_request.py
dalonsod: Done!

chienandalu (approved):
:smile: :+1:

dalonsod:
@manuelregidor could you review? Thanks!

dalonsod:
@ValentinVinagre @HaraldPanten could you review? Thanks!

ValentinVinagre:
> @ValentinVinagre @HaraldPanten could you review? Thanks!

I'll check it out next week if I can.

ValentinVinagre:
Hey @dalonsod ,
Functional review (we don't use this module). Some comments and questions:

- The field "Selected order(s)" from "stock.return.request", why is editable? Is there any way to fill in the returns lines if you put the sales order in the field? I haven't seen the way like that above.
- How I can return several sales orders in the same time?
- If I return sales lines, it would be convenient to indicate on those lines what quantities have been returned or a link to the different sales returns.

Sorry for the delay

dalonsod:
Hello @ValentinVinagre thanks for the review!! I'll try to answer your questions:

- "Selected Order(s)" field is editable because is simply another filter when selecting available moves to be returned. In current implementation you can select e.g. the partner or limit date for potential moves. With this improvement, "_and only moves that come from the selected order(s)_" could be added. This is because for a single partner you can obtain a lot of pickings from a lot of orders, and you want to focus on a set of them.
- Autofill lines option is only available when return request form is accessed from a sales order form, in this version. A sales order action in order to autofill several orders could be added later, it's a good improvement IMO, thanks for the suggestion! I could add it to ROADMAP.
- Regarding to "_If I return sales lines, it would be convenient to indicate on those lines what quantities have been returned or a link to the different sales returns._", this could be also a good improvement for this addon, but actually independent of this PR IMO, maybe @chienandalu could provide us his point of view.

ValentinVinagre (approved):
👍🏻 
Technically I see it correct, I honestly see that it still needs evolution since there are functionalities that are not fully finalized. But since they are functionalities external to the PR and we do not use it, I don't see any problem.

PR #2727, opened 2023-10-18
[13.0][ADD] sale_order_line_remove: allows the removal of sale order lines

JoanSForgeFlow:
Adding new module Sale Order Line Remove

This PR introduces a new module that allows for the removal of sale order lines in certain conditions. Users can now delete lines in confirmed sale orders if they haven't been invoiced or received. Additionally, associated stock moves in specific states will be removed.

### DIFF
--- a/sale_order_line_remove/__manifest__.py
+++ b/sale_order_line_remove/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2023 ForgeFlow, S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sale Order Line Remove",
+    "version": "13.0.1.0.0",
+    "category": "Sales",
+    "summary": "Allows removal of sale order lines from confirmed "
+    "orders if not invoiced or received",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["sale_stock"],
+    "data": [],
+    "installable": True,
+    "application": False,
+}
--- a/sale_order_line_remove/readme/CONTRIBUTORS.rst
+++ b/sale_order_line_remove/readme/CONTRIBUTORS.rst
@@ -1,0 +1,1 @@
+* Joan Sisquella <joan.sisquella@forgeflow.com>
--- a/sale_order_line_remove/__init__.py
+++ b/sale_order_line_remove/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import tests
--- a/sale_order_line_remove/models/__init__.py
+++ b/sale_order_line_remove/models/__init__.py
@@ -1,0 +1,1 @@
+from . import sale_order
--- a/sale_order_line_remove/models/sale_order.py
+++ b/sale_order_line_remove/models/sale_order.py
@@ -15,14 +15,14 @@ class SaleOrderLine(models.Model):
             and not line.invoice_lines
             and not line.move_ids.filtered(lambda move: move.state == "done")
         )
-        invoiced_lines = self.filtered(
+        invoiced_lines = self.sudo().filtered(
             lambda line: line.state in ("sale", "done") and line.invoice_lines
         )
         if invoiced_lines:
             raise UserError(
                 _("You can not remove an order line that has been invoiced")
             )
-        delivered_lines = self.filtered(
+        delivered_lines = self.sudo().filtered(
             lambda line: line.state in ("sale", "done")
             and line.move_ids.filtered(lambda move: move.state == "done")
         )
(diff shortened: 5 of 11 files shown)


sale_order_line_remove/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
```
LoisRForgeFlow, line 3: Remember to add copyrights on non-init files.

JoanSForgeFlow: done! :+1:  thank you

mariadforgeflow (approved):
Code review. LGTM!

AaronHForgeFlow (commented):
I worked in another module that instead of removing the sales order lines it cancelled the lines, this is the main code:


    _inherit = "stock.move"
```
    _inherit = "stock.move"

    def cancel_delivery_order_line(self):
        for move in self:
            if len(move.move_line_ids):
                for move_line in move.move_line_ids:
                    move_line.product_uom_qty = 0
            move._do_unreserve()
            move.state = "cancel"

        return True

from odoo import models, api, fields


    _inherit = "sale.order.line"

    state = fields.Selection(
        [
            ("draft", "Quotation"),
            ("sent", "Quotation Sent"),
            ("sale", "Sales Order"),
            ("done", "Done"),
            ("cancel", "Cancelled"),
        ],
        compute="_compute_line_state",
        string="Line Status",
        readonly=True,
        copy=False,
        store=True,
        default="draft",
        related=False,
    )

    @api.depends(
        "order_id.state",
        "move_ids",
        "move_ids.state",
    )
    def _compute_line_state(self):
        for line in self:
            if line.move_ids and line.move_ids.mapped("state") == ["cancel"]:
                line.state = "cancel"
            else:
                line.state = line.order_id.state

    @api.multi
    def cancel_delivery_order_line(self):
        self.ensure_one()
        moves = self.mapped("move_ids")
        moves.cancel_delivery_order_line()
        self.state = "cancel"
        return True

```

Do you think that would be useful? Perhaps the 2 modules can coexist. The idea of keeping the lines is for traceability, in MTO cases specially.

JoanSForgeFlow:
Hi @AaronHForgeFlow,

Thanks for sharing. Your approach is good for detailed traceability, especially in MTO scenarios. However, it might require adjustments in other parts of Odoo, like reports. The unlink method is simpler and aligns better with Odoo's standard features.

I think both modules can coexist for different user needs.

AaronHForgeFlow (commented):
Functional review LG!

Is it possible to customize the error message  you get when you try to delete a sale order line that is delivered?
image

I would like to see a different error message according to this situation.

AaronHForgeFlow (approved):
Thank you! Good to go.

sale_order_line_remove/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 ForgeFlow, S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sale Order Line Remove",
+    "version": "13.0.1.0.0",
+    "category": "Sales",
+    "summary": "Allows removal of sale order lines from confirmed "
+    "orders uif not invoiced or received",
```
dalonsod: Typo:
```suggestion
    "orders if not invoiced or received",
```

sale_order_line_remove/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Joan Sisquella <joan.siksquella@forgeflow.com>
```
dalonsod: Is your email correct? Sorry it it's right 😇

dalonsod (changes requested):
See comments, And tested,

- When a line is removed and related picking only had linked line, picking remains in draft stated without any line. Should be better cancelling or removing it in this case?
- Made this test: created a SO and billed (draft) a line that contains a service product => for admin, an error is raised trying to delete SO line (OK). But logged at a user without Invoicing permissions, the line could be removed, and that's incorrect. Could you reproduce it?

sale_order_line_remove/readme/CONTRIBUTORS.rst
JoanSForgeFlow: Fixed! Thank you :smile:

JoanSForgeFlow:
Hi @dalonsod,

Thanks for your feedback.

- For the first point, I agree with you. I've updated the module to delete the related picking if it no longer has any lines. This should fix the issue you pointed out.

- Regarding the second point, I've tried to replicate the issue you described but couldn't. In my tests, a user without invoicing permissions was unable to delete a line of a service product in a Sales Order with an invoice in draft state. I've made some changes to the code after your comment, so it's possible the issue has been resolved. Could you please confirm if this is still happening and how I should go about reproducing it?

Thank you for bringing these issues to my attention.

dalonsod:
sure! I'll test it again ASAP

dalonsod:
I've still reproduced that error, in a very slightly way:

- admin creates SO, with a service product, confirm, set delivered qty and create invoice with that line.
- set demo user to "only own documents" sales user.
- Move SO salesman to demo.
- demo is able to delete SO line.

This is probably due to https://github.com/odoo/odoo/blob/13.0/addons/sale/security/sale_security.xml#L162 rule, because demo doesn't see that invoice lines (still "belong" to admin as salesman), that rule should be bypassed inspecting invoice lines count...

May a `sudo()` usage when retrieving invoiced lines solve this issue?

JoanSForgeFlow:
Hi @dalonsod,

Thank you for pointing out this issue. I was able to reproduce the error following your steps. I've applied your suggested sudo() solution for retrieving invoiced lines, and it seems to work well in resolving the issue.

dalonsod (approved):
:+1:

LoisRForgeFlow (approved):
Functional review LGTM

PR #2728, opened 2023-10-19
[ADD] new module sale_order_lot_selection_by_quant

HekkiMelody:
New module letting the user select a lot by choosing a specific quant through a widget.


francesco-ooops:
@grindtildeath slightly related to https://github.com/OCA/stock-logistics-workflow/pull/1391 , what do you think of this one? :)

geomer198 (approved):
LGTM

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@simahawk do you like this one? :)

francesco-ooops:
@pedrobaeza can this be merged?

sale_order_lot_selection_by_quant/readme/ROADMAP.rst
pedrobaeza, line 1: Don't add empty files.

pedrobaeza (commented):
Why do you need JS?

You may want to add some screenshots for the README

HekkiMelody:
> Why do you need JS?

My first attempted involved only adding a `quant_id` field and then changing the `lot_id` based on the selected quant. However, I think the user experience was too poor for it to be usable.

- The name of the quant was not good enough on its own. I would have had to override `name_get` to add some info such as warehouse.
- For the use case I'm covering, the user always wants to click "search more" to see info such as availability of the product in various warehouses
- Sales Order Lines are quite busy enough as it is, so I figured this solution would also have been cleaner. 
- With not much refactoring it's easily reusable or expandable into a more generic widget.

There were other problems too. Going the JS route was not my first solution either.

pedrobaeza:
OK, thanks. Merging based on reviews and comments:

/ocabot merge nobump

PR #2747, opened 2023-10-26
[FIX] sale_order_product_recommendation: freezegun hack

yajo:
As seen in https://github.com/spulec/freezegun/issues/485, when decorating a class with `@freezegun.freeze_time()`, class cleanups are not properly executed.

For Odoo, this means that the cursor is not properly closed. Thus, if you happen to run a test for another module after this one that needs a Postgres-level lock that conflicts with this module's, it will fail.

This change does the same, but without decorating the class. Then, the cleanups work as expected.

The error is reproducible with:

    odoo --stop-after-init -i sale_order_product_recommendation,sale_partner_selectable_option --test-enable --test-tags /sale_order_product_recommendation,/sale_partner_selectable_option

@moduon MT-1075

### DIFF
--- a/sale_order_product_recommendation/tests/test_recommendation.py
+++ b/sale_order_product_recommendation/tests/test_recommendation.py
@@ -2,15 +2,11 @@
 # Copyright 2020 Tecnativa - Pedro M. Baeza
 # Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from freezegun import freeze_time
-
 from odoo.exceptions import UserError
 
 from .test_recommendation_common import RecommendationCase
 
 
-@freeze_time("2021-10-02 15:30:00")
 class RecommendationCaseTests(RecommendationCase):
     def test_recommendations(self):
         """Recommendations are OK."""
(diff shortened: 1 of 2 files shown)


pedrobaeza:
But the issue is only with Python 3.11.

yajo:
No, check this: https://github.com/spulec/freezegun/issues/485#issuecomment-1542230502

And BTW you can actually deploy Odoo 15 with Python 3.11.

pedrobaeza (approved):
OK, it's a pity to have to do such ugly/weird things, but if there's no other solution...

yajo:
Yeah... 😕  This has been the result of several hours of debugging, and I know it might seem awkward at the beginning. But it's related to the way freezegun is implemented. At the end, the summary is that it's not safe to use the class decorator for time-freezing, and instead this is the way that gives zero problems.

Thanks for the review.

pedrobaeza:
I suppose it was hard to diagnose it.

I think the decorator is used in other places, so eventually you'll find the same problem. Not sure if this can be reflected in any document as reference, or maybe at Odoo level.

Shide (approved):
LGTM

sergio-teruel:
@yajo I'm sorry... but in all the v15 instances we have we do not experience this problem..

PR #2749, opened 2023-10-27
[DOC] sale_order_line_menu : Improve usage

TumbaoJu:


TumbaoJu:
@sbidoul : Here is the first functional contribution on Read Me.
We did something simple just to have an example.
@francesco-ooops @lfreeke For information

PR #2752, opened 2023-10-30
[FIX]sale_quotation_number: get_quotation_seq method

manuelregidor:
This fix improves abstraction by adding a new method in sale.order model to get the quotation sequence, so it can be inherited by other modules and change its default behaviour.


ValentinVinagre (approved):
👍🏻

HaraldPanten (approved):
LGTM

PR #2754, opened 2023-10-30
[14.0][IMP] update dotfiles

HekkiMelody:
Superseeds #2750 

cc @etobella


HekkiMelody:
Please do double check the changes to the templates where I removed `position="replace"` in favour of `t-if="False"` before merging.

pedrobaeza:
What you have to do is to put `priority="999"`, but don't remove the replace ones.

HekkiMelody:
Mmmh, ok.

HekkiMelody:
done

pedrobaeza (approved):

Issue #2597, opened 2023-07-07
sale_procurement_group_by_line: No "Purchases" button in SO from sale_purchase_stock module


TelmoSuarezConinpe:
module: sale_procurement_group_by_line
version: 15.0

**Steps to reproduce**
- Install sale_procurement_group_by_line and sale_purchase_stock
- Unarchive "Replenish on Order (MTO)" route 
![Screenshot 2023-07-06 at 12-54-42 Odoo - Replenish on Order (MTO)](https://github.com/OCA/sale-workflow/assets/86360008/1b9b9f46-b142-4442-b44a-f93761d8a8e7)
- Configure product with MTO and Buy.
![Screenshot 2023-07-07 at 11-54-56 Odoo - test MTO](https://github.com/OCA/sale-workflow/assets/86360008/01ebf0a5-6c92-469e-8324-edbfc2c31937)
- Confirm SO with MTO product
- Check if there is a new PO

**Current behavior**
In the SO there is no smart button with the related PO
![Screenshot 2023-07-07 at 11-37-34 Odoo - S00035](https://github.com/OCA/sale-workflow/assets/86360008/4fc3b956-0ac6-476f-bb02-171a2782d24a)

**Expected behavior**
Additional smart button on the SO with the related PO
![Screenshot 2023-07-07 at 11-52-57 Odoo - S00038](https://github.com/OCA/sale-workflow/assets/86360008/accb7525-445b-4ecf-a051-0f8c4616eed7)

support-skillreso:
Hello TelmoSuarezConinpe,

Did you find a workaround for this problem ? I have encountered it too and I can't find how to make this work.

Thanks.

support-skillreso:
I finally found a correction for this problem. For V14, you need to update the file sale_procurement_group_by_line/model/sale.py

First, you need to import odoo.api  :
`from odoo import fields, models, **api**`

Then you need to override SaleOrder class by adding this code before the SaleOrderLine class definition :
```
class SaleOrder(models.Model):
    _inherit = 'sale.order'

    @api.depends('order_line.procurement_group_id.stock_move_ids.created_purchase_line_id.order_id', 'order_line.procurement_group_id.stock_move_ids.move_orig_ids.purchase_line_id.order_id')
    def _compute_purchase_order_count(self):
        super(SaleOrder, self)._compute_purchase_order_count()

    def _get_purchase_orders(self):
[...]

PR #2613, opened 2023-07-20
[15.0][ADD] sale_purchase_procurement_group_by_line: new module

SilvioC2C:
Module ``sale_procurement_group_by_line`` moves the procurement group from the sale order to the sale order lines; but ``sale_purchase_stock`` module only reads sale orders' procurement groups when retrieving purchase orders linked to the sale orders.
This module solves this issue.

PR #2758, opened 2023-10-31
[FIX] sale_procurement_group_by_line: purchase order retrieval

support-skillreso:
Fix _get_purchase_orders to correctly retrieve purchase orders linked to a sale order via procurement group. Useful for products that use MTO route.
Fixes issue #2597

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -3,9 +3,18 @@
 # © 2016 Serpent Consulting Services Pvt. Ltd.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 from odoo.tools.float_utils import float_compare
 
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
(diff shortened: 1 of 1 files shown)


rousseldenis:
@support-skillreso This is not correct. You cannot add a code that depends on purchase without adding dependency in the module.

BUT you can't as we don't want that one.

Do a glue module instead.

support-skillreso:
Hello @rousseldenis,

Thanks for your feedback. I have prepared what I understand as a "glue module" between sale_procurement_group_by_line and sale_purchase_stock. I have linked a zip containing the code, but I'm not sure what to do with it, since I can't make a pull request on this repositery.
Also, I'm not sure about the name of the module : Is the concatenation of the two modules names right ? Should I use a prefix like I did ("sk_") ?

[sk_sale_procurement_group_by_line_sale_purchase_stock.zip](https://github.com/OCA/sale-workflow/files/13283193/sk_sale_procurement_group_by_line_sale_purchase_stock.zip)

SilvioC2C:
@support-skillreso the glue module was implemented in my PR some time ago: https://github.com/OCA/sale-workflow/pull/2613
Feel free to review it if you want to.

PR #2761, opened 2023-11-03
[14.0][MIG] sale_stock_secondary_unit: Migration to 14.0 

toita86:


francesco-ooops (approved):
Functional ok!

sale_stock_secondary_unit/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Sale Stock Secondary Unit",
     "summary": "Get product quantities in a secondary unit",
-    "version": "15.0.1.0.2",
+    "version": "14.0.1.0.2",
```
geomer198: I think module version must be `14.0.1.0.0`

geomer198 (commented):
Please check my comment.

sale_stock_secondary_unit/__manifest__.py
toita86: Thank you! Just fixed.

francesco-ooops:
@geomer198 good now?

GSLabIt (commented):
did u back port from v 15.0?

toita86:
> did u back port from v 15.0?

Yes

geomer198 (approved):
LGTM

GSLabIt:
@pedrobaeza an info on migration best pratics; it is ok to back port from 15.0 instead of migrating from 13.0?

pedrobaeza:
Well, if there's no feature difference, it doesn't matter too much. What I would advise is to migrate from the closest one in terms of datamodel / framework difference. In this case, it would be the same more or less, so backporting from 15 may be other option.

GSLabIt (approved):

francesco-ooops:
@pedrobaeza shall we proceed?

PR #2763, opened 2023-11-07
[16.0][FIX] sale_invoice_policy: re-structure module so that there is no need to change the original invoice_policy field

ghost:
FIX for: https://github.com/OCA/sale-workflow/issues/2725

Issue #2725, opened 2023-10-16
Broken runboat for 16.0


yajo:
Runboat is broken since fa0344a3ceae3d85a459e9b3a5e1ff223e3e891c:

![imagen](https://github.com/OCA/sale-workflow/assets/973709/b18fd702-705d-46d2-86c5-9bd955c0d353)

That happened after merging `sale_order_invoicing_finished_task`. It was detected in the revision, but maybe wrongly attributed to other unknown module: https://github.com/OCA/sale-workflow/pull/2513#issuecomment-1552245396

Please @Yadier-Tecnativa could you check what can be happening? You just have to see the logs of latest runboat builds. Extracting here the important part:

<details>

```
2023-09-26 14:36:28,978 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.modules.loading: Module sale_timesheet: loading demo 
2023-09-26 14:36:28,979 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.modules.loading: loading sale_timesheet/data/sale_service_demo.xml 
2023-09-26 14:36:29,151 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.tests: skip sending email in test mode 
2023-09-26 14:36:29,151 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.addons.mail.models.mail_mail: Mail with ID 47 and Message-Id '<789900971309150.1695738989.126296997070312-openerp-40-project.task@b2d5f6468-15ae-4b22-996f-67b8c7660b0e-initialize-746qd>' successfully sent 
2023-09-26 14:36:29,190 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.models.unlink: User #1 deleted mail.mail records with IDs: [47] 
2023-09-26 14:36:29,190 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.addons.mail.models.mail_mail: Sent batch 1 emails via mail server ID #False 
2023-09-26 14:36:29,225 52 INFO b2d5f6468-15ae-4b22-996f-67b8c7660b0e odoo.tests: skip sending email in test mode 
[...]

PR #2764, opened 2023-11-07
[16.0][IMP] sale_invoice_policy: Make it uninstallable

rousseldenis:


pedrobaeza:
Doing this, the module is not available in runboat.  It's preferable another solution that doesn't break the compatibility between modules. In fact, as is, `sale_invoice_policy` is incorrect and shouldn't be allowed in OCA, as it breaks some parts when installed.

Just now, we are just hiding the trash under the carpet.

rousseldenis:
> Doing this, the module is not available in runboat. It's preferable another solution that doesn't break the compatibility between modules. In fact, as is, `sale_invoice_policy` is incorrect and shouldn't be allowed in OCA, as it breaks some parts when installed.
> 
> Just now, we are just hiding the trash under the carpet.

Yes, I know but I want to let other people to run their PR before finding a good solution.

As the current implementation is depending on context.

pedrobaeza:
I think you should mark it as installable=False, and the CI is checked on the PR

rousseldenis:
> Just now, we are just hiding the trash under the carpet.

Yes, I'll do that instead and create an issue.

pedrobaeza:
The issue #2725 tackles it.

rousseldenis:
> The issue #2725 tackles it.

Yes, but with this I will solve it. Another should be recreated to reimplement the module.

rousseldenis:
@ljsalvatierra-factorlibre @yankinmax @Cedric-Pigeon sale_invoice_policy is breaking standard behavior. I will make it uninstallable here and will create an issue to reimplement it if necessary.

ljsalvatierra-factorlibre:
> @ljsalvatierra-factorlibre @yankinmax @Cedric-Pigeon sale_invoice_policy is breaking standard behavior. I will make it uninstallable here and will create an issue to reimplement it if necessary.

cc. @ACheung-FactorLibre

rousseldenis:
@Yadier-Tecnativa I've reverted your change during migration of sale_order_invoicing_finished_task module as test result depended on sale_invoice_policy side effect.

https://github.com/OCA/sale-workflow/commit/6e12d22507c5168299200e35000c8150cbb67d41#diff-6eb7f1f7a4095dcbbc28fdb5ecc49b6b69e38fdb15ab3a07d9499f5e68a712b6R163

pedrobaeza (approved):
Thanks for the investigation and changes:

/ocabot merge patch

legalsylvain:
Hi. @rousseldenis and @pedrobaeza. I'm working on the migration of the OCA modules we use in GRAP from 12.0 to 16.0 and this module is the last ones I have to migrate. is this module still usefull in V16, or can we consider it as "merged into sale" in Openupgrade apriori file.
kind regard.

pedrobaeza:
The module is still useful (although is debatable that is a good business flow to decide the invoicing policy per order instead of per product). The problem was the implementation, that conflicts and breaks any other module coexisting, as it changes an standard stored field by a non-stored one. It was disabled meanwhile.

The attempt of fixing this is in #2763

legalsylvain:
hi @pedrobaeza. Thanks a lot for your quick answer. I'll follow #2763.
have a nice day.

PR #2765, opened 2023-11-08
[15.0][IMP] product_supplierinfo_for_customer_sale: Change acls to noupdate

CarlosRoca13:
cc @Tecnativa TT45675


carlosdauden (approved):

sergio-teruel (approved):

carlosdauden:
It is ready to merge?

pedrobaeza:
I don't think this should be done this way. If you need other ACLs, just add them. They are additive, so why do you need to do it this way?

carlosdauden:
Everyone can use this option as they wish, but what is the problem with this particular commit?

pedrobaeza:
Introducing heterogeneity and complexity if you change such ACLs in the module development cycle, needing to add migration scripts, which will also break your customizations. That may seems very improbable, but that things have happened in the past, requiring a lot of debugging time.

pedrobaeza (changes requested):
See comments in the task

CarlosRoca13:
@pedrobaeza changes done

pedrobaeza (approved):
Thanks.

/ocabot merge minor

PR #2770, opened 2023-11-09
16.0 mig sale delivery state

sebastienbeau:
Following the discussion here : https://github.com/OCA/sale-workflow/pull/2361

As Odoo have introduced the state "delivery_status" in V16, we believed that the module was useless.
But as the native implementation is broken so we still need this module.
see comment here : https://github.com/OCA/sale-workflow/pull/2361#issuecomment-1801656951

In order to not add any extra new fields, the feature is split in 3 modules
- sale_delivery_state (this add the field but it's not visible in the view)
- sale_stock_delivery_state (glue module to overwrite the implementation of the computed field delivery_status)
- sale_delivery_state_view (glue module when sale_stock is not installed, it will show the field)

It's sad to have to do 3 modules for this, but we do not have choice...


sebastienbeau:
@rousseldenis @lmignon needed for having clean state on shopinvader

flotho:
hi @sebastienbeau , do you need a review here ?
Regards

sale_stock_delivery_state/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "license": "AGPL-3",
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "sale_stock",
+        "sale_delivery_state",
+    ],
+    "data": [],
```
yostashiro:
```suggestion
    "auto_install": True,
    "data": [],
```

yostashiro (commented):
How about integrating `sale_delivery_state_view` into `sale_delivery_state`, and make the added field invisible in `sale_stock_delivery_state`?

sebastienbeau:
@yostashiro seem to be  good idea, I going to change it today
@flotho a review will be welcome (maybe just after the change propose by @yostashiro)

sebastienbeau:
@flotho @yostashiro the PR is ready for the review. I will squash my commit before merging it

lmignon (approved):
Code review only

sebastienbeau:
@flotho @yostashiro ping, thanks in advance for your feedback

sebastienbeau:
@Kev-Roche can you review?

sale_delivery_state/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,20 @@
...
+is set, it will trigger the computation of delivery state.
+
+This module also works with delivery.carrier fees that are added as a
+sale order line. Thoses line are special as they will never be considered delivered.
+Delivery fees lines are ignored in the computation of the delivery state.
+
+
+This module do not add the "views" for the field "delivery_status".
+If you have the sale_stock module installed please installed "sale_stock_delivery_state".
+If not install the module "sale_delivery_state_view".
```
yostashiro:
```suggestion
When the 'sale_stock' module is installed, the glue module 'sale_stock_delivery_state' should also be installed; this module is designed to override the compute method of the delivery status field from 'sale_stock'.
```

yostashiro (commented):
@sebastienbeau Would you consider adding a migration script?

sebastienbeau:
@yostashiro I have added a post-init-hook. This should solve
- migration from 14 to 16 => the field is recomputed so we have the right value
- new installation on v16 => after the installation on an db that have sale_stock, the field is recomputed so all existing Sale Order will have the right value.

It's ok for you ?

yostashiro:
@sebastienbeau Thanks for the update.  I think `sale_delivery_state` still requires a migration script for the scenario of upgrading a database with this module but no `sale_stock`.  Simply moving `post_init_hook()` from `sale_stock_delivery_state` to `sale_delivery_state` might do the job, but I wasn't perfectly sure.

sebastienbeau:
@yostashiro When migrating a database to V16, as the field "delivery_status" have changed of name ("delivery_state" before),
Odoo will consider it as a new field and will compute it and fill the value on all existing sale order.
So everything should work without any extra logic.

yostashiro:
> @yostashiro When migrating a database to V16, as the field "delivery_status" have changed of name ("delivery_state" before), Odoo will consider it as a new field and will compute it and fill the value on all existing sale order. So everything should work without any extra logic.

@sebastienbeau Ah, I see. Missed that point. Thanks for the explanation!

yostashiro (approved):

sebastienbeau:
Thanks for the review let's merge it

/ocabot merge path

ValentinVinagre:
Hey @pedrobaeza @sebastienbeau @yostashiro @HaraldPanten @etobella 
Hello,
I thought that Oca was going to go down the path of using Odoo functionality. Will the native Odoo functionality be left aside in the end? According to the conversation with Pedrobaeza, it seemed that the way forward was going to be to use the native way of Odoo.

The discusion: https://github.com/OCA/purchase-workflow/pull/2081

pedrobaeza:
I prefer Odoo native feature, but there's a lot of people in OCA, and they can think different. At least a mention in the README should be done about the differences between this module and the standard. Or maybe they didn't know about the core feature?

HaraldPanten:
I think they knew about the native Odoo feature, but the decision was to keep the OCA one, because it's more complete than de native one.

PR #2772, opened 2023-11-14
[16.0][IMP] sale order product recommendation - add zero unit included + extended product recommendation domain

santostelmo:
This PR includes 2 improvements for the following use cases:
1. Currently product recommendation wizard shows all the products sold in the past according to some parameters. Recommendations with 0 units included will not be added to the sale order.  
But we have a use case where client wants to add **0 units included** lines in the sale order and then edit the quantities directly in the SO.  
We propose that as an option that can be enabled via settings (see demo):
force_add_0_recom

2. In addition to the default product recommendations, we want to give the option to include or excluded some products. (For example we could exclude any product of type service adding a extra domain `[("product_id.type", "!=", "service")]`
The extra domain can be added via settings:
image


cyrilmanuel (approved):

sale_order_product_recommendation/models/res_company.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    force_zero_units_included = fields.Boolean()
+    sale_line_recommendation_domain = fields.Char(default="[]")
```
leemannd, line 11: This will lead the user to set an invalid domain.
When creating/writing this value, I would test it with `is_leaf` https://github.com/odoo/odoo/blob/17.0/odoo/osv/expression.py#L712

leemannd (commented):
Base looks good. I think we should add a validation testing of the domain

leemannd (approved):

sale_order_product_recommendation/models/res_company.py
santostelmo: Seems a good idea!  💡 Do you know any example of code with that in OCA?

Ricardoalso:
```
from odoo.osv.expression import is_leaf

if is_leaf(sale_line_recommendation_domain)
```

santostelmo: Fortunately, the widget has built-in validation
image

Ricardoalso: Checked it with Telmo and the widget for domain editing already incorporates a friendly "invalid domain" message

Ricardoalso (approved):

sale_order_product_recommendation/models/res_company.py
leemannd: I didn't saw at first check that you where using the widget domain. Cool

cyrilmanuel:
Hello @yajo  @pedrobaeza could you take 5 min to review or merge this PR please  :))

sergio-teruel (approved):

PR #2773, opened 2023-11-14
[MIG] sale_pricelist_triple_discount

SirAionTech:
Migration from https://github.com/OCA/sale-workflow/tree/cc4d5d723ab0ba3006b76d311a3caba7a8cb2903/sale_pricelist_triple_discount (`8.0`).


dreispt:
I tested this, and I don't see the discount % being copied form the Priceline to the Sales Order Line.
Investigating further.

dreispt:
My bad - "Discount Policy" was not set properly. Looks good, thanks!

dreispt (approved):
Simple tests

are-agilebg (approved):

DavideSpagnuolo (changes requested):
There's a Use Case which cannot be managed if this module is installed:
1. Admin user defines a group that gives read-only permission to sale pricelists
2. Admin user add User A to this group
3. User A opens Sales->Products->Pricelists
4. User A selects a specific pricelist
5. Odoo opens the selected pricelist and shows settings and pricelist lines
6. Odoo doesn't allow the User A to operate any changes nut only see the settings
7. User A selects a specific pricelist line
8. Odoo shows settings of the selected pricelist line
9. Odoo doesn't allow the User A to operate any changes but only see the settings

Right now this Use Case at step 4, because Odoo reply with an error for missing permission.
This is caused by function: def _patch_triple_discount(self) in pricelist.py, which requires a write permission on field percent_price.

If possibile, permissions should be managed by external group not limited by the module, writing using sudo.

DavideSpagnuolo (approved):
Now it's possibile to complete the use case previously presented.
It's possibile to create a group that has only view rights on sale pricelists, without recieving errors

ioans73:
@rousseldenis @dreispt Could you merge, please?

PR #2774, opened 2023-11-14
[14.0][FIX] sale_force_invoiced: Proper tagging tests with post_install

mt-software-de:


mt-software-de:
@simahawk, @rousseldenis  could please check this one. 
I tried to tag the unittest properly, but the unittests for the first commit failed for the module sale_triple_discount. 
Normally this shouldn't have affected the test. 
I added the now a second commit where i comment out the tagged. 
As you can see the tests are still failing.

pedrobaeza:
What is the reason for this tagging? I don't think it's needed. You should use the same flow as Odoo and OCA does: only tests on installing, not later once installed.

mt-software-de:
> What is the reason for this tagging? I don't think it's needed. You should use the same flow as Odoo and OCA does: only tests on installing, not later once installed.

That is why i am adding it. It must be tagged as post_install because it needs a proper account set up. 
It is the same as here
https://github.com/OCA/brand/pull/128

When the account module is installed, the journal entries are created the post_init_hook https://github.com/odoo/odoo/blob/ad5a7263461e734e294685fb1406bec97311c4d6/addons/account/__manifest__.py#L95
https://github.com/odoo/odoo/blob/ad5a7263461e734e294685fb1406bec97311c4d6/addons/account/__init__.py#L71
And thats the reason why the test must be tagged as post_install otherwise it will fail with `odoo -i sale_force_invoice --test-enable`

It is working for the OCA test flow, because at first some base modules are installed like account, after that a second install is running for the modules of the repo.

pedrobaeza:
OK about the tags. If invoices are involved, we need to do it post-install for making sure a CoA is present. Please check CI though.

mt-software-de:
> OK about the tags. If invoices are involved, we need to do it post-install for making sure a CoA is present. Please check CI though.

Good. But do you know why the test currently failing? Pre-commit is clear but the test with Odoo/OCB are failing. With the last test commit, which doesn't make any sense to me

mt-software-de:
@pedrobaeza as you can see here https://github.com/OCA/sale-workflow/actions/runs/6946367207/job/18897683508?pr=2795 - https://github.com/OCA/sale-workflow/pull/2795
The test of sale_triple_discount failing also without my changes from this PR.
How we deal with that?

pedrobaeza:
Someone should fix that test, that can be a change upstream that provokes the problem. Ping the maintainers of the module.

mt-software-de:
> Someone should fix that test, that can be a change upstream that provokes the problem. Ping the maintainers of the module.

Is it possible to merge this PR in the meantime? Or do we have to wait for the fix?

pedrobaeza:
We should wait for the good of the state. If not, the red state will expand...

mt-software-de:
> We should wait for the good of the state. If not, the red state will expand...

There is already a PR for fixing https://github.com/OCA/account-invoicing/pull/1605.

StefanRijnhart (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2775, opened 2023-11-14
[14.0][ADD] product_supplierinfo_for_customer_sale_quick

toita86:
Auto-installing bridge module between `product_supplierinfo_for_customer_sale` and `sale_quick`.

The flow should be: the user adds a product from `sale_quick` > the fields "customer code" and "description" 
are populated as if the same product had been added through a standard sales order line using the features of `product_supplierinfo_for_customer_sale`.


product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    product_customer_name = fields.Char(
+        related="variant_customer_ids.product_name",
+        string="Product Customer Name",
+        readonly=True,
+    )
+
+    @api.model
+    def _search_customer_supplierinfo_data(self, operator, value):
```
HekkiMelody, line 22: **issue**: As far as I can tell, this function is not used anywhere

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,28 @@
...
+        related="variant_customer_ids.product_code",
+        string="Product Customer Code",
+        readonly=True,
+    )
+
+    product_customer_name = fields.Char(
+        related="variant_customer_ids.product_name",
+        string="Product Customer Name",
+        readonly=True,
+    )
```
HekkiMelody: **issue**: unfortunately these won't do, as the `related` field will always only take the first record it finds from `variant_customer_ids`, which may not be the one we want. 

In module product_supplierinfo_for_customer there's a function `_select_customerinfo()` https://github.com/OCA/product-attribute/blame/14.0/product_supplierinfo_for_customer/models/product_product.py#L122 that returns the correct `product.customerinfo` based on partner, quantity etc. (though that function could use a couple of fixes as well, but that's for another PR)

product_supplierinfo_for_customer_sale_quick/models/sale_order_line.py
```diff
@@ -0,0 +1,27 @@
...
+                    and supplierinfo.product_code
+                    and supplierinfo.product_name
+                ):
+                    line.name = (
+                        "["
+                        + str(supplierinfo.product_code)
+                        + "]"
+                        + " "
+                        + str(supplierinfo.product_name)
+                    )
```
HekkiMelody: **praise**: The `onchange` is an ok place for this, as `._complete_quick_line_vals()` calls every onchange linked to the fields. Good choice! 

**issue**: However, I don't think a change to `name` should trigger the onchange. I understand that just adding `product_id` to the `onchange` decorator doesn't trigger when you want it to with just `product_id`, but at worst maybe we could call the `onchange` manually. Besides, `product_supplierinfo_for_customer_sale` already implements this logic in `_compute_product_customer_code()`, so we shouldn't add the same logic again here.

HekkiMelody (changes requested):
Hello, and thank you for your first contribution to OCA!

Partial code review.

Also please don't commit empty files, and add tests!

product_supplierinfo_for_customer_sale_quick/views/product_view.xml
```diff
@@ -0,0 +1,31 @@
...
+        <field name="inherit_id" ref="sale_quick.product_tree_view4sale" />
+        <field name="arch" type="xml">
+            <xpath expr="//tree//field[@name='display_name']" position="after">
+                <field name="product_customer_code" optional="show" />
+            </xpath>
+
+            <xpath expr="//tree//field[@name='display_name']" position="after">
+                <field name="product_customer_name" optional="show" />
+            </xpath>
+
```
HekkiMelody: **suggestion**:

```suggestion
            <field name="display_name" position="after>
                <field name="product_customer_code" optional="show" />
                <field name="product_customer_name" optional="show" />
            </field>
```

I'd even drop the `optional` attribute here, but that's just preference on my part.

product_supplierinfo_for_customer_sale_quick/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+Auto-installing bridge module between `product_supplierinfo_for_customer_sale` and `sale_quick`.
+
+The flow should be: the user adds a product from `sale_quick` > the fields "customer code" and "description" 
+are populated as if the same product had been added through a standard sales order line using the features of `product_supplierinfo_for_customer_sale`.
```
HekkiMelody, line 4: **issue**: don't commit empty files (USAGE.rst)

(I can't comment on empty files)

product_supplierinfo_for_customer_sale_quick/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_context_add_products(self):
+        res = super()._get_context_add_products()
+        for record in self:
+            res.update({
+                "partner_id": record.partner_id.id,
+            })
```
HekkiMelody: **chore**: `_get_context_add_products()` is only called by `add_product()` which has an `ensure_one()` so you can be sure that `self` doesn't contain multiple records. Besides, if it ever did, this way you'd be overwriting `res` with only the partner of the last record in the recordset. Remove the `for` loop IMO.

product_supplierinfo_for_customer_sale_quick/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import product_product
+from . import sale_order
```
HekkiMelody: **todo**: Files should end with an empty line

product_supplierinfo_for_customer_sale_quick/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_product_customerinfo_sale_quick_manager,product.customerinfo.quick.sale_manager,product_supplierinfo_for_customer.model_product_customerinfo,sales_team.group_sale_manager,1,1,1,1
+access_product_customerinfo_sale_quick_user,product.customerinfo.sale.quick_user,product_supplierinfo_for_customer.model_product_customerinfo,sales_team.group_sale_salesman,1,0,0,0
```
HekkiMelody: **issue**: There's no need to add the same access rules already present in `product_supplierinfo_for_customer`.

product_supplierinfo_for_customer_sale_quick/views/product_view.xml
```diff
@@ -0,0 +1,31 @@
...
+    <record id="product_search_inherit" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_search_form_view" />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <field name="product_customer_name" />
+            </field>
+            <field name="name" position="after">
+                <field name="product_customer_code" />
+            </field>
```
HekkiMelody: **chore**: syntax can be simplified as above. See comment in model for other issues.

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.model
+    def _search_customer_supplierinfo_data(self, operator, value):
+        products = self.env["product.product"].search(
+            [
+                "|",
+                ("product_customer_code", operator, value),
+                ("product_customer_name", operator, value),
+            ]
+        )
```
HekkiMelody: **issue**: The two fields `product_customer_code` and `product_customer_name` are not stored (and they shouldn't be); that means they're calculated  in memory every time they need to be accessed.

Unfortunately, `search` defers the operation to the database, that cannot execute it because those fields do not exist as columns in the DB.

There's also a recursion error here because this method is called to perform the search. When the `search()` method is called here, Odoo encounters the same field again, and calls this method again since Odoo has been told it to use this method to perform a search on these fields.

My suggestion: 
- only associate this method with `product_customer_name`
- rewrite the search so it doesn't use either of these two fields. You may have to search directly on product.customerinfo` something like: 
```
self.env["product.customerinfo"].search([
            ('partner_id', '=', ...),
            "|",
                ('product_code', 'operator', ...),
                ('product_name', 'operator', ...),
        ])
```
- When adding the filter in xml, only add the filter for `product_customer_name`

Functional question: you add the filter to `product.product`, but these fields do not always have a value. They're only given a value if there's a `partner_id` in the context. What should happen if the user searches on these fields in a "normal" view, without `partner_id` in the context?

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,38 @@
...
+        return [("id", "in", products.ids)]
+
+    def _compute_product_customer_info(self):
+        for product in self:
+            partner_id = product.env.context.get("partner_id")
+            partner = self.env["res.partner"].browse(partner_id)
```
HekkiMelody: **todo**: These two lines could be moves outside of the loop, since they don't depend on the current `product` in the iteration

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,38 @@
...
+            supplier_info = product._select_customerinfo(partner=partner)
+            product.product_customer_code = supplier_info.product_code
+            product.product_customer_name = supplier_info.product_name
```
HekkiMelody: **todo**: We need to make sure this function always assigns a value. But if there's no `partner` we can avoid passing from `_select_customerinfo` and immediately assign `False` to both fields.

HekkiMelody (changes requested):
Good progress on this! A few issues and some code cleanup, but definitely better.

product_supplierinfo_for_customer_sale_quick/security/ir.model.access.csv
toita86: So, i should delete de file?

product_supplierinfo_for_customer_sale_quick/views/product_view.xml
toita86: I would like to keep the option just in case it may be usefull

product_supplierinfo_for_customer_sale_quick/security/ir.model.access.csv
HekkiMelody: If you don't need to add other rules, yes. You'll find a warning in the logs if there are modules without an access rule. (You'll also have issues when testing the module, because you won't be able to access those models)

product_supplierinfo_for_customer_sale_quick/models/product_product.py
toita86: I have implemented the changes. For the functional question, I think it should retrieve nothing, so i will have no output for that search. Alternatively, it might throw an error.

product_supplierinfo_for_customer_sale_quick/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product Supplierinfo for Customer Sale Quick",
+    "version": "14.0.0.0.0",
```
HekkiMelody: **issue**: Generally modules start at `14.0.1.0.0`. The first two (`14.0`) are the Odoo version, and then the last 3 numbers (`1.0.0`) are the major, minor and revision numbers for the module itself.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#L59-L77

product_supplierinfo_for_customer_sale_quick/models/product_product.py
HekkiMelody: I think these two new filters shouldn't go in the generic product search, since they're not going to be very useful.

`base_product_mass_addition` implements a separate `search` view https://github.com/OCA/product-attribute/blob/14.0/base_product_mass_addition/views/product_view.xml#L19

and `purchase_quick` correctly uses that `search` view https://github.com/OCA/purchase-workflow/blob/14.0/purchase_quick/models/purchase_order.py#L44-L46

The same does not happen with `sale_quick` but maybe it should! And then you'd only add your two new filters to `base_product_mass_addition.product_search_form_view`

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    @api.model
+    def _search_customer_supplierinfo_data(self, operator, value):
+        partner_id = self.env.context.get("partner_id")
+        partner = self.env["res.partner"].browse(partner_id)
+
+        products = self.env["product.customerinfo"].search(
+            [
+                ("name", "=", partner.name),
```
HekkiMelody: **issue**: it may be counterintuitive, but the field `name` on `product.customerinfo` is a many2one towards `res.partner`.

https://github.com/OCA/product-attribute/blob/14.0/product_supplierinfo_for_customer/models/product_customerinfo.py#L12

So this should be
```suggestion
                ("name", "=", partner_id),
```

product_supplierinfo_for_customer_sale_quick/models/product_product.py
```diff
@@ -0,0 +1,48 @@
...
+                "|",
+                ("product_name", operator, value),
+                ("product_code", operator, value),
+            ]
+        )
+        return [
+            ("id", "in", products.mapped("product_tmpl_id.product_variant_ids").ids)
```
HekkiMelody: **issue**: The search may return only specific variants, but this would make it so that all the variants of a specific `product.template` are returned. Is there a reason why this behaviour would be desirable?

HekkiMelody (commented):
Code review, generally looks good, and I'd say it's close to done. A few comments here and there.

product_supplierinfo_for_customer_sale_quick/models/product_product.py
toita86: While I was testing I noticed that all the variants need to be returned in the case that a product has an empty `Product Variant` field in the `sale` tab of a product. 
If I would use something like `return [("id", "in", products.mapped("product_id").ids)]` in this case I would have a empty research results.
I'm pretty sure I'm missing something or there is a better way implement this case.

HekkiMelody: This is ok now and can be resolved

product_supplierinfo_for_customer_sale_quick/views/product_view.xml
```diff
@@ -0,0 +1,29 @@
...
+            ref="base_product_mass_addition.product_search_form_view"
+        />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <field name="product_customer_name" />
+                <field name="product_customer_code" />
+            </field>
+        </field>
+    </record>
+</odoo>
```
HekkiMelody, line 29: Continuing from [here](https://github.com/OCA/sale-workflow/pull/2775#discussion_r1502983285) since Github won't let me reply directly for some reason

This is ok now, but I don't think you can see those filters, right?

I think in [sale_quick](https://github.com/OCA/sale-workflow/blob/14.0/sale_quick/models/sale_order.py#L42) you need to specify the use of this search view, just like it was done in [purchase_quick](https://github.com/OCA/purchase-workflow/blob/14.0/purchase_quick/models/purchase_order.py#L44-L46)

HekkiMelody (commented):
Just a minor change and then it's good to go IMO

HekkiMelody (approved):
Code review, LGTM

francesco-ooops (approved):
Functional ok!

@OCA/crm-sales-marketing-maintainers can this be merged? Thanks!

francesco-ooops:
@pedrobaeza can this be merged? thank you

PR #2338, opened 2023-01-13
[16.0][MIG][WIP] Migration sale_product_set_layout

angelmoya:

PR #2777, opened 2023-11-16
[16.0][MIG] sale_product_set_layout

CRogos:
supersede https://github.com/OCA/sale-workflow/pull/2338

@pedrobaeza : sale_product_set was splitted into product_set  and sale_product_set. Some of these extensions belong to product_set and other to sale_product_set now. I don't think it is a good idea to split this module and would rather add the functionality to product_set and sale_product_set. What do you think?
If you don't need need Section/Note than don't use it. If you agree, I will prepare updates for product_set and sale_product_set.


CRogos:
I've moved all changes into these two modules.
sale_product_set: https://github.com/OCA/sale-workflow/pull/2786
product_set: https://github.com/OCA/product-attribute/pull/1452

What is needed to do to add sale_product_set_layout into the upgrade script? In think this should be merged into sale_product_set.

PR #2605, opened 2023-07-13
[16.0][MIG] sale_order_price_recalculation

andrel-exo:

PR #2778, opened 2023-11-16
[16.0][MIG] sale_order_price_recalculation

CRogos:
supersede: #2605

Has anyone an idea what is wrong with the tests? Is this related to the hard disk swap outage?


CRogos:
Has anyone an idea why this test is failing? When I run it locally the tests passes and also the module is working fine as far as I can say.

MohamedOsman7 (approved):
code & functional changes LGTM

CRogos:
Is the recalculate price obsolete? When switching the price list, there is an update button: 
image

CRogos:
@arleite , @rousseldenis could you review this migration?

CRogos:
@arleite  can you also add yourself to the reviews?
image

andrel-exo (commented):
LGTM

CRogos:
@arleite. I should have been more precise. You have to select "approve", not "comment". 

Thank you.

andrel-exo (approved):

CRogos:
@pedrobaeza could you merge?

pedrobaeza:
This module is outdated after the integration of the feature in core, and if you want it for the description thing, it should be renamed IMO.

CRogos:
The update price button in the standard is not always visible and there is not possibility to reset the description. But I agree this feature mostly adds some UI improvements. 

The functionality is not changed, so I do not understand why we should change the name... but make a suggestion?

pedrobaeza:
I would maybe start from scratch a little module `sale_order_price_recalculation_visibility` with such function. And do you really use any time the reset description feature?

CRogos:
While refactoring this module I've exactly asked this question to my sales guys and they said that they are missing something like that. There themes to be a use case for that... but I agree that there could be a more beautiful solutions like these buttons.

pedrobaeza:
OK, you tell me then: should I press the button without thinking more?

CRogos:
Yes please... anyone who has this module installed would be able to migrate to v16 without any functionality change. And Odoo has improved but does not fulfil the whole functionality. 

I agree that the module is now very small and maybe we can combine it with other functions in the future.

PR #2779, opened 2023-11-17
[16.0][FIX] sale_order_general_discount_triple: remove attribute required f…

celm1990:
…or res.config.settings

- Field mark as required cause issues with other modules

PR #2780, opened 2023-11-17
[16.0][FIX] *.* remove requirement unnecesary

celm1990:
this dependence cause error into all CI

@pedrobaeza can you review this please, it was added on PR https://github.com/OCA/sale-workflow/pull/2382 but i not see is used and I get error on my PR https://github.com/OCA/sale-workflow/pull/2779

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,1 +1,0 @@
-odoo_test_helper


CRogos:
This is not necessary. I a think  the issues where related to the hardware fail. My PR tests are working again. https://github.com/OCA/sale-workflow/pull/2778

celm1990:
> This is not necessary. I a think the issues where related to the hardware fail. My PR tests are working again. #2778

Sorry for confusion, I see now working

PR #2783, opened 2023-11-17
[14.0] Fix pricelist_cache concurrent update issues

mmequignon:
see commit descriptions


pricelist_cache/models/product_pricelist_cache.py
```diff
@@ -193,8 +193,21 @@ def _update_pricelist_items_cache(self, pricelist_items):
...
-        pricelists = self.env["product.pricelist"].search([])
-        pricelist_ids = pricelists.ids
+        pricelist_model = self.env["product.pricelist"]
+        # Here, we split price computation in 2.
+        # Huge pricelists (root pricelists and factor pricelists) are computed
+        # on their own, in order to avoid long sql transactions.pricelist_model
+        # Smaller pricelists can be computed 3 by 3, since they are taking
+        # less time to process.
+        global_list_ids = pricelist_model._get_global_pricelist_ids()
+        # Belt and braces. Just to ensure higher level lists are executed first
```
simahawk, line 203: what grants that "higher level" is bound to the ID order?

pricelist_cache/tests/common.py
```diff
@@ -132,3 +133,27 @@ def setUpClass(cls):
         cls.set_currency()
         cls.setUpClassBaseCache()
         cls.partner = cls.env.ref("base.res_partner_12")
+
+    def _flush_cache(self):
+        self.cache_model.with_context(debug_test=True).flush_pricelist_cache()
```
simahawk: `debug_test` -> left over?

pricelist_cache/models/product_pricelist_cache.py
mmequignon: no, if the pricelist has not parent, or just adds N% on the parent list price, then the pricelist is global, meaning that we have to store all prices.

pricelist_cache/tests/common.py
mmequignon: yes, thanks

pricelist_cache/models/product_pricelist_cache.py
mmequignon: Sorry, I misunderstood your question.
Well, the biggest one should be the default one (id 1), and since each list depends on it, then it makes all pricelist caches unavailable if it isn't computed.
I want to be sure it's done first.
But there's no guarantee here.
I added a note in the ROADMAP, in order to add job dependencies.
If list 3 depends on 2 which depends on 1, I want job 3 to depend on job 2 etc…

simahawk: we should probably have an explicit flag :see_no_evil:

simahawk (approved):

sebalix (approved):
LG
Sidenote: better to avoid the use of personal pronouns in commit messages

mmequignon:
This pr conflicts with sale triple discount.
I'll have to have a look at this

PR #2785, opened 2023-11-18
[17.0][MIG]sale_order_type

guestisp:
First commit for migration to Odoo 17.0

Now the addon install properly on 17, but there are 2 things to address (i don't have enought Odoo knowledge to fix them on my own):

1. the incoterms were moved from `sale_stock` to accounting or sale_order, I think we should change the reference in the views, but i wasn't able to do so, so i've removed the `group` reference in the XML. I think we have to add them ,but how?

2. `states` are not supported anymore in 17, i've replaced them by keeping the model in `readonly = False` but we have to restore the value based on the state (Draft, Sale, Done, Cancel ,....)

if someone tell me how to fix these two things, i'll update the PR


CRogos:
Please have a look into the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

I use https://github.com/OCA/oca-port to partly automate the steps.

guestisp:
yes, i've used the migration guide but not al changes are listed there

PR #2788, opened 2023-11-19
[MIG] sale_order_type: Migration to 17.0

guestisp:
First commit for migration to Odoo 17.0

Now the addon install properly on 17, but there are 2 things to address (i don't have enought Odoo knowledge to fix them on my own):

the incoterms were moved from sale_stock to accounting or sale_order, I think we should change the reference in the views, but i wasn't able to do so, so i've removed the group reference in the XML. I think we have to add them ,but how?


celm1990:
duplicated of https://github.com/OCA/sale-workflow/pull/2836
Can join effort and keep a single PR?

pedrobaeza:
Superseded by #2836

PR #2790, opened 2023-11-21
[FIX] sale_packaging_default: reorder SO sub-form view too

yajo:
When creating SO from mobile view or from "Commercial Quotations" menu installed by `sale_order_product_picker`, the pop-up form view that appears when clicking on the line kanban kard didn't have the fields in the correct order.

When you install `sale_packaging_default`, you expect this order:
1. Packaging.
2. Packaging qty.
3. Product qty.

@moduon MT-3930


rafaelbn (approved):
Good point! 👍🏼

PR #2791, opened 2023-11-21
[15.0][IMP] sale_order_line_price_history: Manage price history from wizard

CarlosRoca13:
With this changes we avoid having 2 reserved columns on the tree order lines to make the same aproach.

The button is now just in one column and painted as red to show that the price can't be transfered to the line (This is happening when the line is still in new state).

cc @Tecnativa TT46085 

ping @pedrobaeza @chienandalu


pedrobaeza (approved):

chienandalu (commented):
:+1:

chienandalu (approved):

PR #2792, opened 2023-11-21
[FIX] sale_packaging_default: setting product_uom_qty to 0 empties packaging

yajo:
Before this patch, we had this problem:

1. Add new SO line.
2. Set a packaging and a packaging qty. The UoM qty is computed. Good.
3. Set UoM qty to zero. Packaging and packaging qty remain with previous value. Bad.

When user sets UoM to zero, Odoo should remove any packaging information.

Notice that, when adding a new line, that qty will be zero at the beginning, but we will want to still get the defaults. Thus, this change is only applied when the user *is* actually setting that value, and not on any other circumstance. I needed to override the `onchange()` method to propagate that information.

@moduon MT-3930

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,7 +1,8 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.17.2
+_commit: v1.19.2
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
+convert_readme_fragments_to_markdown: false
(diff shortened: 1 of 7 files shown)


Gelojr (approved):
LGTM https://www.loom.com/share/da7859d81be349bea4493e6a8cf06f86?sid=fa69fb2b-dfbd-451a-b1af-7c61d78a3087

Shide (approved):

PR #2793, opened 2023-11-21
[FIX] sale_order_product_picker: store packaging information

yajo:
Previous situation: when packagings are enabled in sale configuration, if a product had any packaging, the packaging and its qty wasn't saved when clicking on the "+1 unit" button.

This could lead to incongruent situations:
- A handpicked packaging and qty combination didn't match with the product uom qty.
- Possibly valid packaging and qty combinations were not computed automatically, as they were when filling from the normal sale order view.

@moduon MT-3930


CarlosRoca13 (approved):
Thanks :+1:

rafaelbn (approved):
👍🏼 Great! Thank you!

PR #2795, opened 2023-11-21
Test PR for checking sale_triple_discount tests

mt-software-de:
The test of sale_triple_discount running into an error within my PR https://github.com/OCA/sale-workflow/pull/2774.
Normally it should not have any affect.


francesco-ooops:
@mt-software-de check https://github.com/OCA/account-invoicing/pull/1605

PR #2796, opened 2023-11-21
[16.0][FIX] sale_order_product_picker: Changes not saved when line is already saved

### LINKED ISSUE #2794: [16.0] sale_order_product_picker: cannot use to edit qtys, after saving for the 1st time
## Module
sale_order_product_picker

## Describe the bug
After you use the product picker once and save the record, you cannot use it any more. What is worse, you don't get any errors or visual feedback. The changes are just ignored. See:

https://github.com/OCA/sale-workflow/assets/973709/6d2381d9-086b-4216-a5ad-c10cd783f660

## To Reproduce
**Affected versions**: 16.0

Steps to reproduce the behavior: See video above.

**Expected behavior**
Changes should get saved, just like when editing for the 1st time.

**Additional context**
New module from https://github.com/OCA/sale-workflow/pull/2762.

@moduon MT-3930
Ping @CarlosRoca13 @pedrobaeza

CarlosRoca13:
This PR solves the issue reported on https://github.com/OCA/sale-workflow/issues/2794

cc @Tecnativa 

ping @yajo @rafaelbn

### DIFF
--- a/sale_order_product_picker/static/src/kanban/kanban_record.esm.js
+++ b/sale_order_product_picker/static/src/kanban/kanban_record.esm.js
@@ -132,9 +132,9 @@ patch(KanbanRecord.prototype, "sale_order_product_picker.KanbanRecord", {
         const orderLines = x2mList.records.filter(
             (line) => line.data.product_id[0] === record.data.product_id[0]
         );
+        this.disableGlobalClick = true;
         if (!orderLines.length) {
             // Disable global click to avoid actions while record is added
-            this.disableGlobalClick = true;
             const lineRec = await x2mList.addNew({
                 position: "bottom",
                 context: ctx,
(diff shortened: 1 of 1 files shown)


yajo:
@rafaelbn @Gelojr please review. @moduon MT-3930

Gelojr (approved):
LGTM
https://www.loom.com/share/ff5a17456abd4e5ca6960bcd552eebbf?sid=38c9b40f-9567-487d-97bf-bd16b27a48aa

Shide (approved):
LGTM

yajo (changes requested):
The problem with this fix is that it turns the button slower.

For example, if you want to add 3 units and click the button 3 times quickly, chances are that only 1 or 2 of those clicks are actually handled.

I think it would make sense to throttle and defer the calls, and only when the user has stopped clicking for about 500ms, then upload the changes. Does that make sense?

Otherwise, at least it would help if the button appears as disabled while you can't use it. That visual feedback isn't the best, but at least gives a hint that you have to wait to click more.

yajo:
Hi @CarlosRoca13. Have you been able to review https://github.com/OCA/sale-workflow/pull/2796#pullrequestreview-1743843628? Any thoughts about it?

CarlosRoca13:
Yes it's true, that some clicks are lost, but right now I have no chance to make the change, sorry :pray: 

I will review it when I have a chance

rafaelbn:
@yajo could we merge this as is and later propose a new PR? O better to propose a new PR directly?

I would like to close this task 😄 

https://github.com/OCA/sale-workflow/pull/2796

@moduon MT-4423

yajo:
OK, let's merge. The new bug is better than the older one. I'll open a new issue about the new bug, so we can track it.

/ocabot merge patch

yajo:
New issue opened in #2842

PR #2804, opened 2023-11-24
[16.0][ADD] sale_triple_discount: Additional discounts missing in sal…

lethuthao2403:
ref: https://github.com/odoo/odoo/pull/145513

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -131,3 +131,108 @@ class SaleOrderLine(models.Model):
             discount=self._get_final_discount(),
             price_subtotal=self.price_subtotal,
         )
+
+    @api.depends("discount2", "discount3", "discounting_type")
+    def _compute_untaxed_amount_to_invoice(self):
+        to_recompute = self.filtered(lambda r: r.discount2 or r.discount3)
+        res = super(
+            SaleOrderLine, self - to_recompute
+        )._compute_untaxed_amount_to_invoice()
+        if to_recompute:
+            to_recompute._compute_untaxed_amount_to_invoice_process()
+        return res
+
+    def _compute_untaxed_amount_to_invoice_process(self):
+        """Total of remaining amount to invoice on the sale order line (taxes excl.) as
+            total_sol - amount already invoiced
+        where Total_sol depends on the invoice policy of the product.
+        Note: Draft invoice are ignored on purpose, the 'to invoice' amount should
+        come only from the SO lines.
+        """
+        # Code copied from Odoo sale.order.line._compute_untaxed_amount_to_invoice
+        #  in order to consider triple discount fields (cf Adjustment comments)
+        # FIXME: Module would need a refactor to use dedicated fields for triple discount
+        #  so that sale.order.line.discount field can be used to aggregate the triple discount
+        #  fields as suggested in https://github.com/odoo/odoo/pull/145513#pullrequestreview-1774865804  # noqa
+        for line in self:
+            amount_to_invoice = 0.0
+            if line.state in ["sale", "done"]:
+                # Note: do not use price_subtotal field as it returns zero when
+                # the ordered quantity is zero. It causes problem for expense
+                # line (e.i.: ordered qty = 0, deli qty = 4, price_unit = 20 ;
+                # subtotal is zero), but when you can invoice the line,
+                # you see an amount and not zero. Since we compute untaxed
+                # amount, we can use directly the price
+                # reduce (to include discount) without using `compute_all()` method on taxes.
+                price_subtotal = 0.0
+                uom_qty_to_consider = (
+                    line.qty_delivered
+                    if line.product_id.invoice_policy == "delivery"
+                    else line.product_uom_qty
+                )
+
+                # Adjustment: Get the discount from line._get_final_discount()
+                # instead of line.discount
+                price_reduce = line.price_unit * (
+                    1 - (line._get_final_discount() or 0.0) / 100.0
+                )
+                price_subtotal = price_reduce * uom_qty_to_consider
+                if len(line.tax_id.filtered(lambda tax: tax.price_include)) > 0:
+                    # As included taxes are not excluded from the computed subtotal,
+                    # `compute_all()` method has to be called to retrieve
+                    # the subtotal without them.
+                    # `price_reduce_taxexcl` cannot be used as it is computed
+                    # from `price_subtotal` field. (see upper Note)
+                    price_subtotal = line.tax_id.compute_all(
+                        price_reduce,
+                        currency=line.currency_id,
+                        quantity=uom_qty_to_consider,
+                        product=line.product_id,
+                        partner=line.order_id.partner_shipping_id,
+                    )["total_excluded"]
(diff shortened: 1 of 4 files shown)


ajaniszewska-dev:
@lethuthao2403  pre-commit to fix

lethuthao2403:
> @lethuthao2403 pre-commit to fix

@ajaniszewska-dev the PR has been updated! Thanks

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,95 @@ def _convert_to_tax_base_line_dict(self):
...
+        come only from the SO lines.
+        """
+        for line in self:
+            amount_to_invoice = 0.0
+            if line.state in ["sale", "done"]:
+                # Note: do not use price_subtotal field as it returns zero when the ordered quantity is
+                # zero. It causes problem for expense line (e.i.: ordered qty = 0, deli qty = 4,
+                # price_unit = 20 ; subtotal is zero), but when you can invoice the line, you see an
+                # amount and not zero. Since we compute untaxed amount, we can use directly the price
+                # reduce (to include discount) without using `compute_all()` method on taxes.
```
ajaniszewska-dev: @lethuthao2403  some of the lines are too long, so pre commit is still failing
```
sale_triple_discount/models/sale_order_line.py:155:104: B950 line too long (103 > 88 characters)
sale_triple_discount/models/sale_order_line.py:157:101: B950 line too long (100 > 88 characters)
sale_triple_discount/models/sale_order_line.py:158:102: B950 line too long (101 > 88 characters)
sale_triple_discount/models/sale_order_line.py:171:108: B950 line too long (107 > 88 characters)
sale_triple_discount/models/sale_order_line.py:173:124: B950 line too long (123 > 88 characters)
sale_triple_discount/models/sale_order_line.py:189:103: B950 line too long (102 > 88 characters)
```

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,95 @@ def _convert_to_tax_base_line_dict(self):
...
+                                    round=False,
+                                )
+                                * inv_line.quantity
+                            )
+
+                    amount_to_invoice = max(price_subtotal - amount, 0)
+                else:
+                    amount_to_invoice = price_subtotal - line.untaxed_amount_invoiced
+
+            line.untaxed_amount_to_invoice = amount_to_invoice
```
grindtildeath: Here you are copying a fairly big amount of code from the Odoo function `_compute_untaxed_amount_to_invoice`.
Can you please state this in a comment and use another comment to highlight what are you changing from this function?

grindtildeath (commented):
I guess the only part you had to change from the Odoo function is the `if` condition? In that case I'm wondering if we shouldn't submit a PR to Odoo to add a hook and mention that in a code comment.

sale_triple_discount/models/sale_order_line.py
lethuthao2403: hi @grindtildeath I want to calculate the 'price_reduce' based on three discount fields. However, the original function doesn't have any separate functions for calculating each discount individually.

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,21 @@ def _convert_to_tax_base_line_dict(self):
             discount=self._get_final_discount(),
             price_subtotal=self.price_subtotal,
         )
+
+    def _needs_remaining_calculation(self, inv_lines):
+        res = super()._needs_remaining_calculation(inv_lines)
```
nguyenminhchien: add 1 check:
```
if res:
    return res
```

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,21 @@ def _convert_to_tax_base_line_dict(self):
...
+        if self.discount2 or self.discount3:
+            res = any(
+                inv_lines.mapped(
+                    lambda inv_l: inv_l.discount != self.discount
```
nguyenminhchien: no need to check `inv_l.discount` because `super()` has done this

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,21 @@ def _convert_to_tax_base_line_dict(self):
...
+                    or inv_l.discount3 != self.discount3
+                )
+            )
+        return res
+
+    def _get_discount(self):
+        res = super()._get_discount()
+        if self.discount2 or self.discount3:
+            res = self._get_final_discount()
+        return res
```
nguyenminhchien:
```
    def _get_discount(self):
        if self.discount2 or self.discount3:
            return self._get_final_discount()
        return super()._get_discount()
```

nguyenminhchien (changes requested):

nguyenminhchien:
@grindtildeath branch updated, but the tests are failed at this moment, we need to wait for https://github.com/odoo/odoo/pull/145513 merged.

nguyenminhchien:
@grindtildeath 

[Feyensv](https://github.com/Feyensv) does not like the idea of the hook , i let you decide what is the next move.
image

grindtildeath:
@lethuthao2403 @nguyenminhchien Thanks for reporting this to Odoo.

I guess the best to do in a first time is to copy the whole function there is on Odoo - with code comments to highlight the change in the if condition - as it was in a previous version of this PR.

Then, I'd suggest to open an issue about the implementation of this particular module on this repository, because IMO the point of view of @Feyensv is legit. And having 3 discount fields, that would then be "consolidated" into the standard discount field could solve plenty of issues, and avoid costly overrides. But that's a major breaking change and I'm not sure we want to do it right now.

nguyenminhchien:
@grindtildeath @ajaniszewska-dev  it's updated. there are 2 changes which commented by **# Adjustment**

sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,103 @@ def _convert_to_tax_base_line_dict(self):
...
+
+    def _compute_untaxed_amount_to_invoice_process(self):
+        """Total of remaining amount to invoice on the sale order line (taxes excl.) as
+            total_sol - amount already invoiced
+        where Total_sol depends on the invoice policy of the product.
+        Note: Draft invoice are ignored on purpose, the 'to invoice' amount should
+        come only from the SO lines.
+        """
+        for line in self:
```
grindtildeath, line 157:
```suggestion
        # Code copied from Odoo sale.order.line._compute_untaxed_amount_to_invoice
        #  in order to consider triple discount fields (cf Adjustment comments)
        # FIXME: Module would need a refactor to use dedicated fields for triple discount
        #  so that sale.order.line.discount field can be used to aggregate the triple discount
        #  fields as suggested in https://github.com/odoo/odoo/pull/145513#pullrequestreview-1774865804  # noqa
        for line in self:
```

nguyenminhchien: updated.

grindtildeath (approved):

grindtildeath:
cf https://github.com/OCA/sale-workflow/pull/2885

PR #2821, opened 2023-11-30
[15.0][IMP] sale_resource_booking: add multiples attendees in booking

angelgarciadelachica:
With this FIX the adaptations of the IMP https://github.com/OCA/calendar/pull/110 are made.

On the other hand, we change the name of the type_id field of the transient model resource_booking_sale that causes an error when creating a quotation from a resource.booking.type with the module sale_order_type installed.

### DIFF
--- a/sale_resource_booking/models/sale_order_line.py
+++ b/sale_resource_booking/models/sale_order_line.py
@@ -74,7 +74,7 @@ class SaleOrderLine(models.Model):
             }
             rbc_rel = line.product_id.resource_booking_type_combination_rel_id
             context = {
-                "default_partner_id": line.order_id.partner_id.id,
+                "default_partner_ids": [(4, line.order_id.partner_id.id)],
                 "default_combination_auto_assign": not rbc_rel,
                 "default_combination_id": rbc_rel.combination_id.id,
             }
--- a/sale_resource_booking/tests/test_sale_resource_booking.py
+++ b/sale_resource_booking/tests/test_sale_resource_booking.py
@@ -61,11 +61,9 @@ class SaleResourceBookingsCase(TransactionCase):
         self.assertTrue(order.resource_booking_ids)
         self.assertTrue(self.rbt.booking_ids)
         self.assertEqual(self.rbt.booking_count, 2)
-        # Use wizard to quickly assign partners
-        wiz = self._run_action(action["actions"][0])
-        with Form(wiz) as wiz_f:
-            with wiz_f.resource_booking_ids.edit(1) as booking_f:
-                booking_f.partner_id = partner2
+        # Add new attendees
+        for booking in order.resource_booking_ids:
+            booking.partner_ids += partner2
         # Click on "Bookings" smart button
         action = order.action_open_resource_bookings()
         bookings = self._run_action(action)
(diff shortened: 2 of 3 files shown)


angelgarciadelachica:

sale_resource_booking/wizards/resource_booking_sale.py
```diff
@@ -9,7 +9,7 @@ class ResourceBookingSale(models.TransientModel):
     _name = "resource.booking.sale"
     _description = "Sale order generator for resource booking types"
 
-    type_id = fields.Many2one(
+    rb_type_id = fields.Many2one(
```
chienandalu: suggestion: a more expressive name:
```suggestion
    booking_type_id = fields.Many2one(
```

pedrobaeza: Anyway, this can be considered more a bug in `sale_order_type`, not cleaning the context (I assume with `default_type_id`). Changing one field name means that a migration script should be provided, so we should avoid it, or provide the script.

chienandalu: Well, this is a transient model, so maybe is not such a big deal

pedrobaeza: OK, being a transient, no migration script indeed, but anyway, the problem will arise with any transient using `type_id`, so I insist that should be fixed in `sale_order_type`.

ValentinVinagre: I find it totally necessary to change the name of the field of the sale_order_type module and the purchase_order_type. They are "common" names that can be stepped on as has happened with different modules.
But I would make the change for the v.17 migration. How do you see it?

angelgarciadelachica: In any case, I also think it is necessary to rename the field and avoid using common names. So I have changed the name as suggested by @chienandalu. 
What do you think?

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):
LGTM @chienandalu @pedrobaeza

pedrobaeza (changes requested):
I don't see the migration scripts, but I insist: this renaming is not needed at all. The fix should be done elsewhere. Anyway, please limit this PR to the multiple attendees thing, and we handle the other thing in another PR, stating the steps to reproduce the problem.

angelgarciadelachica:
> I don't see the migration scripts, but I insist: this renaming is not needed at all. The fix should be done elsewhere. Anyway, please limit this PR to the multiple attendees thing, and we handle the other thing in another PR, stating the steps to reproduce the problem.

IMO it is better not to separate it in two PRs, I think it is better to separate them in two commits, one for the IMP and another one for the FIX. 

On the other hand, in the commit with the FIX, instead of changing the name of the type_id field, I have removed it from the context when creating the sales order from the wizard. What do you think?

pedrobaeza:
No, it's not, as the fix for multiple attendees is going to be merged quick, but the other not, so you are not interested in delay the most urgent fix IMO. Anyway, the proper way to do it is to call `clear_context` (or similar) from tools, that remove all the `default_` whatever. And don't use `self._context`. That's an internal variable. Use `self.env.context`.

angelgarciadelachica:
@pedrobaeza

angelgarciadelachica:
Sorry @pedrobaeza , I didn't wanted to send an empty message. 

I have already made the changes, when you can, could you check it? 
THX!

pedrobaeza (approved):
Thanks.

/ocabot merge patch

PR #2825, opened 2023-12-01
[IMP] sale_order_product_picker: quick add 1 packaging

yajo:
When enabling packaging standard Odoo system, now the picker allows you to add +1 package.
imagen

For this to work, product packagings must be enabled in settings, and the product must have at least one packaging configured for sales.

Only the 1st one for sales will be used for the picker shortcut.

@moduon MT-4396

### DIFF
--- a/sale_order_product_picker/README.rst
+++ b/sale_order_product_picker/README.rst
@@ -59,6 +59,12 @@ Usage
    * On page Picker you can search for products.
    * You can:
       * Add a line by clicking on +1 button.
+      * Add a packaging by clicking on +1 button.
+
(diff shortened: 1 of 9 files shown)


yajo:
While reviewing functionally, please keep in mind that:
- if you click on +1 unit, it still is affected by https://github.com/OCA/sale-workflow/issues/2794
- if you click on +1 packaging, it still is affected by https://github.com/OCA/sale-workflow/pull/2796#pullrequestreview-1743843628

I pushed an update that hides the new button if the product has no packaging configured.

Shide (approved):
Code LGTM

PR #2826, opened 2023-12-01
[17.0][MIG] sale_automatic_workflow

sanderlienaerts:


pedrobaeza:
Please check https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#how-to

The "pre-commit auto-fixes" commit is missing.

sanderlienaerts:
The pre-commit auto fixes commit was merged with the migration into one commit, now restored in [1eb94b1](https://github.com/OCA/sale-workflow/pull/2826/commits/1eb94b1b951afcfd01a06598af085afd4c4b0002).

pedrobaeza:
> The pre-commit auto fixes commit was merged with the migration into one commit,

Thanks. Was there any reason for squashing them together? Maybe the documentation is not clear enough about it, and we should improve it.

/ocabot migration sale_automatic_workflow

sanderlienaerts:
> > The pre-commit auto fixes commit was merged with the migration into one commit,
> 
> Thanks. Was there any reason for squashing them together? Maybe the documentation is not clear enough about it, and we should improve it.
> 
The [documentation](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests) mentions PRs should have the least possible commits. I probably took that bit too literal by merging the commit for pre-commit because it was not really adding anything functional.

pedrobaeza:
Yeah, it's true that it seems not needed, but in this case it's made this way for easing the review of the manual changes done for making this to work on the new version. I always review that isolated commit when doing code review.

sanderlienaerts:
Thanks for clarifying 👍

sanderlienaerts:
Hi @pedrobaeza,

 Unsure how, but after reviewing this PR - somehow a binary file ([pandoc-3.1.9-x86_64-macOS.pkg](https://github.com/sanderlienaerts/sale-workflow/blob/17.0-mig-sale_automatic_workflow/pandoc-3.1.9-x86_64-macOS.pkg)) snuck into the auto fixes commit. I have force pushed the branch to remove the binary. No other changes were made.

pedrobaeza:
I think it's pre-commit not cleaning correctly resources.

smcqueen81 (approved):
Functional Ok

atchuthan (changes requested):
Except action_done in Sales order automatic workflow, other functions are working as expected.  Tested functionally :+1: 

https://github.com/sanderlienaerts/sale-workflow/blob/50fd7dfeb5db5db0c88bd4c3d57401f74d73b43d/sale_automatic_workflow/models/automatic_workflow_job.py#L139
> we could use action_lock() method as done state is removed in Odoo v17.

More info: https://github.com/odoo/odoo/commit/17bece3e797913bcba8dd7e07fc8541c0a45e3f7#diff-695321b9df7129604a59fbc9bff34f97ba90b3394aa1166ca24e36b27389ae82L962-R959

sale_automatic_workflow/data/automatic_workflow_data.xml
```diff
@@ -0,0 +1,112 @@
...
+            name="domain"
+        >[('state', 'in', ['draft', 'confirmed', 'assigned'])]</field>
+        <field name="user_id" ref="base.user_root" />
+    </record>
+    <record id="automatic_workflow_create_invoice_filter" model="ir.filters">
+        <field name="name">Automatic Workflow Create Invoice Filter</field>
+        <field name="model_id">sale.order</field>
+        <field
+            name="domain"
+        >[('state','in',['sale','done']),('invoice_status','=','to invoice')]</field>
```
celm1990: State done was removed, please update `domain`
```suggestion
        >[('state','in',['sale']),('invoice_status','=','to invoice')]</field>
```

celm1990 (changes requested):

sanderlienaerts:
Thanks for reviewing. The PR commit was changed accordingly.

celm1990 (approved):
LGTM

john-herholz-dt (approved):

john-herholz-dt:
@pedrobaeza is it ready to be merged? Can you have a look on it when you have time?
Thank you!

sale_automatic_workflow/data/automatic_workflow_data.xml
```diff
@@ -34,8 +33,8 @@
         >[('state', '=', 'draft'), ('posted_before', '=', False)]</field>
         <field name="user_id" ref="base.user_root" />
     </record>
-    <record id="automatic_workflow_sale_done_filter" model="ir.filters">
-        <field name="name">Automatic Workflow Sale Done Filter</field>
+    <record id="automatic_workflow_sale_lock_filter" model="ir.filters">
```
pedrobaeza: Please don't rename XML-ID records if there's no need. It's OK to rename the text for better clarity, but the technical XML-ID is OK (the value is still `done`), and you avoid potential compatibility/migration problems.

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -130,23 +130,22 @@ def _validate_pickings(self, picking_filter):
             with savepoint(self.env.cr):
                 self._do_validate_picking(picking, picking_filter)
 
-    def _do_sale_done(self, sale, domain_filter):
-        """Set a sales order to done, filter ensure no duplication"""
+    def _do_sale_lock(self, sale, domain_filter):
```
pedrobaeza: The same about renaming methods. It's not adding too much value in this case, as the goal is the same, and it has been the same name for several versions where done = lock.

sale_automatic_workflow/models/stock_picking.py
```diff
@@ -24,13 +24,13 @@ def validate_picking(self):
                 rounding = move.product_id.uom_id.rounding
                 if (
                     float_compare(
-                        move.quantity_done,
+                        move.quantity,
```
pedrobaeza, line 27: Why changing this? The field has been renamed?

sale_automatic_workflow/tests/test_multicompany.py
```diff
@@ -73,7 +72,9 @@ def setUpClass(cls):
         cls.env.user.company_ids |= cls.company_fr_daughter
 
         cls.env.user.company_id = cls.company_fr.id
-        coa.try_loading(company=cls.env.user.company_id)
+        cls.env["account.chart.template"].try_loading(
+            "generic_coa", cls.env.user.company_id
```
pedrobaeza: You mustn't depend on `l10n_generic_coa`. Put the tests as `post-install` for avoiding the problem.

sale_automatic_workflow/views/sale_workflow_process_view.xml
```diff
@@ -56,15 +56,17 @@
                             <div class="col-sm-8">
                                 <label
                                     for="order_filter_id"
-                                    attrs="{'required':[('validate_order','=',True)], 'invisible':[('validate_order','!=',True)]}"
+                                    required="validate_order == True"
+                                    invisible="validate_order != True"
```
pedrobaeza: Please avoid spaghetti code (the same for the rest):
```suggestion
                                    required="validate_order"
                                    invisible="not validate_order"
```

pedrobaeza (changes requested):

sale_automatic_workflow/models/stock_picking.py
john-herholz-dt: Yes, see https://www.youtube.com/watch?v=KNcK-TPTpRg&t=631s

pedrobaeza: In the slide, it seems `quantity_done` is now `picked`, not `quantity`, isn't it?

john-herholz-dt: reserved_quantity and quantity_done were merged together into 'quantity' and picked is a boolean (indicator for backorders as I understood it).

rousseldenis:
@sanderlienaerts Could you attend last comments ?

sale_automatic_workflow/tests/test_multicompany.py
sanderlienaerts: Hi @pedrobaeza, the `post_install` tags were already added. So not sure what the specific mentioned solution should have been.

Because I was unable to use `try_loading()` with country specific chart template codes in tests, I've managed to load chart templates by inheriting `setup_company()` from `AccountTestInvoicingCommon` instead. If you have any recommendations or if this is bad practice, please let me know.

sanderlienaerts:
> @sanderlienaerts Could you attend last comments ?

@rousseldenis the comments have been attended.

Thank you all for your critical reviews. The suggested changes have been applied and where applicable renaming fields, methods and XML-id's reverted.

sale_automatic_workflow/models/stock_picking.py
pedrobaeza: Then you should also check if `picked` is True.

sale_automatic_workflow/tests/test_multicompany.py
pedrobaeza: OK, I think it's enough to just do the tests in post-install and don't create a new company in the tests, but as you see it.

sale_automatic_workflow/models/stock_picking.py
rousseldenis: @sanderlienaerts Is this applicable ?

sanderlienaerts: hi @rousseldenis, the condition has been pushed to the latest commit

bizzappdev (approved):

bizzappdev:
IMO this is ready for the final review

celm1990:
@OCA/crm-sales-marketing-maintainers could somebody help me with review or merge, please

pedrobaeza:
As I see, my comments were attended. The only weird thing that I see now is that the migration commit contains the removal of some RST files that shouldn't be there and should be removed by the `pre-commit auto fixes` commit, and the same for some reformatting in XML files. @sanderlienaerts can you please fix these last things to proceed with the merge?

sonhd91 (approved):
I do believe some commits could be squashed together and the removing of some `.rst` file should be in `pre-commit` commit

sanderlienaerts:
All right, sorry this was on my to do list for a while. Not sure how commit https://github.com/OCA/sale-workflow/pull/2826/commits/f36d7c412c9b6004b5a4f35ec6e15aaeb063cf72 got in between the migration commits. Is it ok to stay this way?

sebalix:
I know this is a big change, but is it possible to port the refactoring done a long time ago in 13.0 to extract the dependency on `stock` from the base module?
https://github.com/OCA/sale-workflow/pull/1257

This work was not merged to keep existing installations stable, and it was better to wait a migration to add such changes. We could handle this on top of your PR, as soon as it is not merged (or we'll have to wait for 18.0 migration).

sale_automatic_workflow/static/description/index.html
```diff
@@ -0,0 +1,463 @@
+<?xml version="1.0" encoding="utf-8"?>
```
xaviedoanhduy, line 1: Can you remove this line?

xaviedoanhduy (approved):

trisdoan:
Hi, I opened https://github.com/OCA/sale-workflow/pull/3105 to follow-up on this PR:
- take occasion to squash commits regarding to fixup, pot update
- separate module stock from sale_automatic_workflow

sanderlienaerts:
> Hi, I opened #3105 to follow-up on this PR:
> 
> * take occasion to squash commits regarding to fixup, pot update
> * separate module stock from sale_automatic_workflow

Thanks, I will close this one!

PR #2827, opened 2023-12-01
[REF] sale_packaging_default: remove sale_default field

yajo:
It turns out that, since product packagings can be sorted, and can already be set as `sales=True`, we can safely consider that the 1st packaging with that field on is the default one for sales.

We loose the ability to set one product with sales packagings to not have a default one, but we win in the sense that we don't diverge from Odoo CE DB model.

That advantage means that the "packaging_default" family of modules are now just a plug-in that will emphasize the usage of packagings in the sale process, but other modules that use the same data model can reason similarly without depending directly on these ones.

@moduon MT-4396


Shide (approved):
Code LGTM

rafaelbn (approved):
OK, thank you! 👍🏼

PR #2828, opened 2023-12-03
[15.0][IMP] sale_stock_return_request: optional sales order filter / shortcut from sales order

dalonsod:
Fw-port of #2722 

Sometimes selecting some orders to be returned for a certain customer could be useful, specially when we want return a sales order with a lot of pickings. Adding a filter and/or having a shortcut from sale order it could be performed.


chienandalu (approved):
Thanks :)

pedrobaeza (approved):
/ocabot merge minor

PR #2829, opened 2023-12-05
[FIX] sale_order_line_date: format style of commitment_date

miguel-S73:
The `text-left` label is no longer used In Odoo 16.0. Instead, `text-start` is now utilized.
I have replaced this label and also added a name to the `th` and `td` elements to improve the code when inheriting.


hildickethan (approved):

flotho (approved):
easy code review,
thanks

PR #2832, opened 2023-12-05
[16.0][ADD] sale_loyalty_exclude

santostelmo:
This module allows you to exclude products from a loyalty programs. These products have no effect on the minimum amount required to apply for the loyalty program.


cyrilmanuel (approved):

sale_loyalty_exclude/models/sale_order.py
```diff
@@ -0,0 +1,16 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_no_effect_on_threshold_lines(self):
+        """
+        Returns the lines that have no effect on the minimum amount to reach
+        """
+        self.ensure_one()
+        lines = self.order_line.filtered(lambda line: line.product_id.loyalty_exclude)
+        return lines + super()._get_no_effect_on_threshold_lines()
```
Ricardoalso: Don't you want an `union` here instead of `+`?

Ricardoalso (approved):
Small comment, otherwise LG

sale_loyalty_exclude/models/sale_order.py
santostelmo: I think union and + are equivalent in the case because this is a recordset

Ricardoalso: image
Apparently not

Ricardoalso: image
image

santostelmo: Interesting with `+` you have a duplicated record. I thought that was not possible 🤔

santostelmo: Thank you @Ricardoalso in this case is better to use `|`

santostelmo: In standard 17.0 they are using `+` https://github.com/odoo/odoo/blob/11bd9e01b79ec94c6e8e637a30cd412f01f2d791/addons/sale_loyalty_delivery/models/sale_order.py#L25

santostelmo: Changed

PR #2833, opened 2023-12-05
[16.0] sale_product_packaging_container_deposit: fix qty delivered update

simahawk:
We should never write on related records in a computed field.
Now to detect the update of delivered qty we hook to the stock move update.

Depends on https://github.com/OCA/product-attribute/pull/1475

The same fix should be applied to the purchase module.

### DIFF
--- a/sale_product_packaging_container_deposit/models/stock_move.py
+++ b/sale_product_packaging_container_deposit/models/stock_move.py
@@ -1,0 +1,15 @@
+# Copyright 2023 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_done(self, cancel_backorder=False):
+        todo_moves = super()._action_done(cancel_backorder=cancel_backorder)
+        # ensure container deposit qty are updated on related SOs
+        if todo_moves:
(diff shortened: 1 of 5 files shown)


santostelmo (approved):

santostelmo:
I update https://github.com/OCA/purchase-workflow/pull/2034 with that fix

sale_product_packaging_container_deposit/models/stock_move.py
```diff
@@ -0,0 +1,15 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_done(self, cancel_backorder=False):
+        todo_moves = super()._action_done(cancel_backorder=cancel_backorder)
+        # ensure container deposit qty are updated on related SOs
+        if todo_moves:
+            todo_moves.sale_line_id.order_id.update_order_container_deposit_quantity()
+        return todo_moves
```
jbaudoux, line 15:
```suggestion
        done_moves = super()._action_done(cancel_backorder=cancel_backorder)
        # ensure container deposit qty are updated on related SOs
        if done_moves:
            done_moves.sale_line_id.order_id.update_order_container_deposit_quantity()
        return done_moves
```

jbaudoux (commented):
I find your variable name weird

jbaudoux (approved):

sale_product_packaging_container_deposit/models/stock_move.py
simahawk: yeah, I just respected odoo's name. That's what they return. I can change it tho.

PR #2836, opened 2023-12-06
[MIG] sale_order_type: Migration to 17.0

matiasperalta1:


nicolascol:
Functional LGTM

sale_order_type/models/account_move.py
```diff
@@ -12,7 +12,7 @@ class AccountMove(models.Model):
         string="Sale Type",
         compute="_compute_sale_type_id",
         store=True,
-        readonly=True,
+        readonly="state not in ['draft']",
```
celm1990: These attributes should now be in XML instead of .py files and must be evaluated as Booleans. Please refer to the logs of the `test` for more information.

sale_order_type/models/sale.py
```diff
@@ -16,7 +16,7 @@ class SaleOrder(models.Model):
         compute="_compute_sale_type_id",
         precompute=True,
         store=True,
-        readonly=False,
+        readonly="state in ['draft']",
```
celm1990: same here

celm1990 (changes requested):
The `CI` has failed, check test

lef-adhoc:
@celm1990 It's ready

celm1990 (approved):
LGTM

sale_order_type/models/sale.py
```diff
@@ -16,12 +16,6 @@ class SaleOrder(models.Model):
         compute="_compute_sale_type_id",
         precompute=True,
         store=True,
-        readonly=False,
```
pedrobaeza: Don't remove `readonly=False`. If you do this, you are saying to the ORM that the field is not computed writable.

sale_order_type/views/sale_order_type_view.xml
```diff
@@ -52,10 +52,7 @@
                                 name="pricelist_id"
                                 groups="product.group_product_pricelist"
                             />
-                            <field
-                                name="incoterm_id"
-                                groups="sale_stock.group_display_incoterm"
```
pedrobaeza: Why are you removing the group on this field?

pedrobaeza (changes requested):

sale_order_type/views/sale_order_type_view.xml
celm1990: This group Was removed, see https://github.com/odoo/odoo/commit/9ddc128d7bd3998fb3cf1f1c55d0d76fccc91196

sale_order_type/models/sale.py
lef-adhoc: I already put it back

pedrobaeza (approved):
/ocabot merge nobump

PR #2830, opened 2023-12-05
[16.0][ADD] partner_picking_note

EmilioPascual:
This module add two description fields in partner, one for internal use and other for customer use. 
When assign partner in sale order, the fields description are copied into the respective fields of the sales order.

https://www.loom.com/share/ffdec27ff32244cf902d26cacb26300a

@rafaelbn @Gelojr @Shide please review 😄

MT-4381 @moduon

PR #2841, opened 2023-12-11
[IMP] picking notes in partner

EmilioPascual:
Add two description fields in partner, one for internal use and other for customer use.
When assign partner in sale order, the fields description are copied into the respective fields of the sales order.

From pr https://github.com/OCA/sale-workflow/pull/2830. Better add behavior to this module and don't create module. 

https://www.loom.com/share/04281bb4c8c2431ca01dd08253ea55bf?sid=d3893e04-5a68-4c0e-bd67-b886ffadf295

@rafaelbn @Gelojr @Shide @yajo please review 😄

MT-4381 @moduon


sale_stock_picking_note/models/sale_stock.py
```diff
@@ -1,15 +1,28 @@
...
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    picking_note = fields.Html(string="Picking Internal Note")
-    picking_customer_note = fields.Text(string="Picking Customer Comments")
+    picking_note = fields.Html(
+        string="Picking Internal Note",
+        # compute="_compute_picking_note"
```
yajo:
```suggestion
```

sale_stock_picking_note/models/sale_stock.py
```diff
@@ -1,15 +1,28 @@
...
+    )
+    picking_customer_note = fields.Text(
+        string="Picking Customer Comments",
+        # compute="_compute_picking_note"
```
yajo:
```suggestion
```

sale_stock_picking_note/models/sale_stock.py
```diff
@@ -1,15 +1,28 @@
...
+    )
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id_picking_notes(self):
+        for order in self:
+            order.picking_note = order.partner_id.picking_note
+            order.picking_customer_note = order.partner_id.picking_customer_note
 
 
 class StockMove(models.Model):
```
yajo: **issue:** the `_get_new_picking_values()` method should fall back to partner notes if there's no sale order related to the picking.

sale_stock_picking_note/tests/test_sale_stock_picking_note.py
```diff
@@ -8,32 +8,78 @@ class TestSaleStockPickingNote(common.TransactionCase):
...
         )
-        cls.order = cls.env["sale.order"].create(
+        # cls.order = cls.env["sale.order"].create(
+        #     {
+        # "partner_id": cls.partner.id,
+        # "order_line": [
+        #     (0, 0, {"product_id": cls.product.id, "product_uom_qty": 1})
+        # ],
+        #     }
+        # )
```
yajo:
```suggestion
```

yajo (changes requested):
Code & functional review. Some comments below.

sale_stock_picking_note/models/stock_picking.py
```diff
@@ -1,10 +1,23 @@
...
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     customer_note = fields.Text(string="Customer Comments")
+
+    @api.onchange("partner_id", "picking_type_id")
+    def _onchange_partner_id_picking_notes(self):
+        for picking in self:
```
chienandalu, line 28: suggestion: you don't need to iterate over the recordset in an onchange

sale_stock_picking_note/models/stock_picking.py
```diff
@@ -1,10 +1,23 @@
...
+            if picking.picking_type_id.code != "incoming":
+                if self.group_id.sale_id:
+                    note = self.group_id.sale_id.picking_note
+                    customer_note = self.group_id.sale_id.picking_customer_note
+                else:
+                    note = picking.partner_id.picking_note
+                    customer_note = picking.partner_id.picking_customer_note
```
chienandalu: suggestion:
```suggestion
                self.note = self.sale_id.picking_note or self.partner_id.picking_note
                self.customer_note = self.sale_id.picking_customer_note or self.partner_id.picking_customer_note
```

question: should we keep manual modifications from being squashed by this onchange?

sale_stock_picking_note/views/res_partner_view.xml
```diff
@@ -0,0 +1,19 @@
...
+<!-- Copyright 2023 Moduon Team S.L.
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0) -->
+<odoo>
+    <record id="view_partner_form_inherit_partner_picking_note" model="ir.ui.view">
+        <field name="name">Partner Picking Note</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='sale']" position="inside">
+                <field name="picking_note" placeholder="Picking notes..." />
```
chienandalu: question: Maybe this should be under the stock user group? I mean, isn't it a problem that any user in the organization can add a default note to any contact?

sale_stock_picking_note/models/sale_stock.py
```diff
@@ -1,15 +1,26 @@
...
-    picking_note = fields.Html(string="Picking Internal Note")
-    picking_customer_note = fields.Text(string="Picking Customer Comments")
+    picking_note = fields.Html(
+        string="Picking Internal Note",
+    )
+    picking_customer_note = fields.Text(
+        string="Picking Customer Comments",
+    )
+
+    @api.onchange("partner_id")
```
chienandalu: suggestion: This will only work on UI. For a more resilient way use a storable not readonly compute.

sale_stock_picking_note/models/sale_stock.py
```diff
@@ -1,15 +1,26 @@
...
+    def _onchange_partner_id_picking_notes(self):
+        for order in self:
+            order.picking_note = order.partner_id.picking_note
```
chienandalu: question: what happens if there are user default values? Maybe we could avoid writing the notes if the user doesn't have any.

chienandalu (commented):
Thanks for your work @EmilioPascual 

Some comments:

sale_stock_picking_note/models/stock_picking.py
Shide: I think it's not necessary since the picking notes are related directly to the partner.
Maybe we could check if it's empty adding self.note and self.customer_note at the end of "or" chains.

```python
self.note = self.sale_id.picking_note or self.partner_id.picking_note or self.note
self.customer_note = self.sale_id.picking_customer_note or self.partner_id.picking_customer_note or self.customer_note
```

Shide: I think we should always propagate the data from the order and the partner. 
I don't see the Stock User adding notes on the stock picking directly by himself. Maybe a Salesman receiving a call and adding it to an order, but not for the Stock User.

```python
# I prefer a depends to allow batch update of the sales
@api.depends("sale_id.picking_note", "partner_id.picking_note", "sale_id.picking_customer_note", "partner_id.picking_customer_note")
```

sale_stock_picking_note/views/res_partner_view.xml
Shide: At least ensure Sale User group (don't remember if the tab has this group)

This will be propagated from Sales to Pickings in all cases.

Shide (changes requested):
Good job!

**suggestion:** the `@api.depends` will help in batch modifications, and precise propagation (only when field changes). I think @yajo suggest an onchange

**note:** check groups of partner fields to ensure Salesman+ can modify them.

sale_stock_picking_note/views/res_partner_view.xml
EmilioPascual: > question: Maybe this should be under the stock user group? I mean, isn't it a problem that any user in the organization can add a default note to any contact?


> At least ensure Sale User group (don't remember if the tab has this group)
> 
> This will be propagated from Sales to Pickings in all cases.

Added sales: own documents and stock user to fields

sale_stock_picking_note/models/stock_picking.py
EmilioPascual: Done both

sale_stock_picking_note/models/stock_picking.py
EmilioPascual: You are right

sale_stock_picking_note/models/sale_stock.py
EmilioPascual: Changed to compute

sale_stock_picking_note/models/sale_stock.py
EmilioPascual: I don't undestand you. User default values? Can you give me more context please?

sale_stock_picking_note/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+        string="Picking Customer Comments",
+        compute="_compute_picking_notes",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("partner_id.picking_note", "partner_id.picking_customer_note")
+    def _compute_picking_notes(self):
+        for order in self:
+            if order.state not in ("sale", "done", "cancel"):
```
yajo: Typical performance rule of thumb enhancement:

```suggestion
            if order.state not in {"sale", "done", "cancel"}:
```

sale_stock_picking_note/models/sale_stock.py
yajo: I think he means an `ir.default`. IMHO the partner defaults should override user defaults.

sale_stock_picking_note/models/sale_stock.py
yajo: I'm not sure I prefer this new approach based on computes.

Imagine this flow:

1. Wood Corner has a note: "never deliver on sundays". 
2. They buy a new chair. However, this one is urgent, and it can be delivered on sunday. In the sales order, the salesman writes: "this time, deliver on sunday".
3. Then he realizes that this happens often, so he decides to go to the partner and change the note to: "unless explicitly stated, don't deliver on sundays".

If we use an onchange, all goes as expected. If we use a compute, then all open orders and pickings will get that last note, and it will override the value that the salesman already changed by hand.

Thus, I believe an onchange is a better option, as we only want to get some values *by default*, but still leave room for human intervention later.

sale_stock_picking_note/models/stock_picking.py
```diff
@@ -1,10 +1,38 @@
...
+    )
+
+    @api.depends(
+        "sale_id.picking_note",
+        "partner_id.picking_note",
+        "sale_id.picking_customer_note",
+        "partner_id.picking_customer_note",
+        "picking_type_id",
+    )
+    def _compute_picking_notes(self):
```
yajo, line 27: For the same reasons as explained above in https://github.com/OCA/sale-workflow/pull/2841#discussion_r1423765223, I also think a compute is not the best solution here.

sale_stock_picking_note/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+        readonly=False,
+    )
+    picking_customer_note = fields.Text(
+        string="Picking Customer Comments",
+        compute="_compute_picking_notes",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("partner_id.picking_note", "partner_id.picking_customer_note")
```
chienandalu: suggestion: this isn't a very good idea as you'll squash a lot of custom notes without noticing. Just depend on `partner_id`.

yajo (changes requested):
I'm not sure this new approach is better. Reopening some threads for further analysis, please place your comments 😃

sale_stock_picking_note/models/stock_picking.py
```diff
@@ -1,10 +1,38 @@
...
+    )
+    note = fields.Html(
+        compute="_compute_picking_notes",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends(
+        "sale_id.picking_note",
+        "partner_id.picking_note",
```
chienandalu: suggestion: Same here. You don't want to squash the notes that are already set

sale_stock_picking_note/models/sale_stock.py
rafaelbn: @EmilioPascual imagine you don't have this improvement and you have a default vaule for sale order with this customer.

@chienandalu Default vaules has prioriti over the partner fields so I guess this will not impact to you, but please! tell us.

I have record a 3 minutes video! (check @EmilioPascual )

https://www.loom.com/share/4fa5b85fa6084a468fc65da91355f841?sid=ddccdcfa-54a1-4d84-94ee-215cb40859c6

sale_stock_picking_note/models/sale_stock.py
chienandalu: Sure, and that's why I commented not depending on the partner fields, so the compute behaves like an onchange.

In the other hand, I'm thinking in processes that create sale orders automatically (e.g.: purchase sale intercompany) that should seamlessly take those default notes. A compute is the way to go for this case, IMO

yajo: Ah OK, with that solution, then it makes sense. Thanks!

rafaelbn:
- [x] @chienandalu do you agree if we convert both fields to HTML (internal comment and customer comment)

sale_stock_picking_note/models/sale_stock.py
rafaelbn: > IMHO the partner defaults should override user defaults.

IMHO the oposite 😄, default in sales should overwrite partner notes and in fact actually functionality I think is RIGHT! 👍🏼 as expected. Take all care that ir.default will be never used with this improvement. Further more I said to Emilio to open a new module consciously of this problems impacting in a new one, we don't want or have time to discuss too much this trivial improvement 😄 https://github.com/OCA/sale-workflow/pull/2830 (that why it was other module)

chienandalu:
> * [ ]  @chienandalu do you agree if we convert both fields to HTML (internal comment and customer comment)

Not very friend of the idea, it might be a breaking change for somebody if they don't deal with that html properly. I can't devise the case, but I'm not sure that it won't be an issue either :sweat_smile:

chienandalu (approved):
Thanks :+1:

sale_stock_picking_note/models/sale_stock.py
yajo: > > IMHO the partner defaults should override user defaults.
> 
> IMHO the oposite 😄, default in sales should overwrite partner notes and in fact actually functionality I think is RIGHT! 👍🏼 as expected.

🤔 why?

A salesman can sell to 1000 customers, where 900 have no specific note, but 100 of them have. Doesn't it make more sense if those 100 customers get their specific notes, instead of getting the default one that the salesman just uses for everyone else?

EmilioPascual:
I think methods should be compute to work well in batch editing but not depend on notes fields. I have changed behavior, please review again. Thanks!

Shide (approved):
LGTM

chienandalu (approved):

yajo (approved):

rafaelbn:
@Gelojr detected a BUG, @EmilioPascual @chienandalu , notes should be from the delivery address and not from de customer. https://www.loom.com/share/d9af1396b9a8437283de2064fc9af858

Issue #2818, opened 2023-11-28
[IMP] [14-15-16] sale_delivery_state


ValentinVinagre:
Explanation of the limitation:

If in a sale order you have 2 lines:
* Consumable/storable type product
* Service type product that refers to transportation.

The "delivery_state" field will never indicate that it has been completely delivered because it takes into account the delivery of the service and will have to force that the entire sale has been delivered.

Solution:

Create a checkbox in sales configuration (by company), which indicates whether this field has to take into account service type products or not.
In the future it could be expanded with the sale_order_type module in case there are order types that do have to be taken into account.

What do you think about adding this new functionality? If you are convinced, we will carry out the PR with the extension.
@HaraldPanten @etobella @pedrobaeza

etobella:
I think you don't even need the flag, by default, the delivery shouldn't check services. Or did I understood something wrongly=

ValentinVinagre:
It makes sense in sales that create projects and invoices based on recorded hours (which count as delivered), so in those cases it should work.

HaraldPanten:
I think that delivery state should be focused in picking, not in hours (even if the system takes in consideration the delivered hours), but a checkbox will be an "intermediate solution" that will satisfy both points of view.

@yostashiro @sebastienbeau @lmignon @rousseldenis that might be interesting for you, as well.

yostashiro:
`sale_delivery_state` already skips order lines with `is_delivery` true, but I suppose the issue raised here is about service lines with `is_delivery` false (or when `delivery` module is not installed).

I tend to think that those service lines should also be accounted for, and users should manually update delivered quantities.  However, I think it's also good to have a boolean field in the order line to provide an option to be skipped in the delivery state computation. Adding a global configuration is good, and it'd be even better to be able to adjust skip/no-skip at the order line level.

ValentinVinagre:
> `sale_delivery_state` already skips order lines with `is_delivery` true, but I suppose the issue raised here is about service lines with `is_delivery` false (or when `delivery` module is not installed).
> 
> I tend to think that those service lines should also be accounted for, and users should manually update delivered quantities. However, I think it's also good to have a boolean field in the order line to provide an option to be skipped in the delivery state computation. Adding a global configuration is good, and it'd be even better to be able to adjust skip/no-skip at the order line level.

We could consider both cases, I see it well. Because it is well received, we will present a PR with the change in the next few days.

jbaudoux:
[...]

PR #2845, opened 2023-12-12
[IMP]sale_delivery_state: Skip sale order lines for delivery state

manuelregidor:
Based on https://github.com/OCA/sale-workflow/issues/2818


manuelregidor:
@HaraldPanten @ValentinVinagre @etobella @yostashiro Could you review, please? Thank you.

ValentinVinagre (approved):
LGTM 👍🏻

etobella (approved):

manuelregidor:
There is a failing test from sale_by_packaging but it doesn't seem to be related to this imp.

HaraldPanten:
@etobella We made a small change in order to solve the tests issue. Could you re-review, please? 🙏

sale_delivery_state/models/sale_order_line.py
```diff
@@ -0,0 +1,27 @@
...
+    skip_sale_delivery_state = fields.Boolean(
+        string="Skip Delivery State",
+        compute="_compute_skip_sale_delivery_state",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("company_id", "product_id")
+    def _compute_skip_sale_delivery_state(self):
+        for sel in self:
```
etobella: I do usually prefer to use for record in self: or for line in self:

etobella (approved):
Just a small nitpicking comment,

The main idea of the change is that it is more readable

sale_delivery_state/models/sale_order_line.py
manuelregidor: done

HaraldPanten (approved):

simahawk:
Reminder for the next time: when adding a field to a model that could be linked to thousands or millions of records, add a pre-mig script to create the field before upgrading the module to avoid issues w/ existing DBs :wink:

PR #2849, opened 2023-12-13
[FIX] 14.0 fix i18n fr

rdualsam:


rousseldenis:
@rdualsam Please use https://translation.odoo-community.org/ instead. Thanks

PR #2850, opened 2023-12-13
[15.0][FIX]*: Change type to detailed_type (correct field in products)

victoralmau:
Change `type` to `detailed_type` (correct field in products https://github.com/odoo/odoo/blob/15.0/addons/product/models/product_template.py#L50)

@Tecnativa

### DIFF
--- a/sale_invoice_policy/models/product_template.py
+++ b/sale_invoice_policy/models/product_template.py
@@ -31,19 +31,19 @@ class ProductTemplate(models.Model):
         for template in self.filtered("invoice_policy"):
             template.default_invoice_policy = template.invoice_policy
 
-    @api.depends("type", "default_invoice_policy")
+    @api.depends("detailed_type", "default_invoice_policy")
     @api.depends_context("invoice_policy")
     def _compute_invoice_policy(self):
         """
         Apply the invoice_policy given by context (if exist) otherwise use the
         default invoice policy given by the field with this same name.
-        If the product is type = 'service', we don't have to apply the invoice
+        If the product is detailed_type = 'service', we don't have to apply the invoice
         policy given by the context.
         :return:
         """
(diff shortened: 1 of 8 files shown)


sale_invoice_policy/models/product_template.py
```diff
@@ -31,19 +31,19 @@ def _inverse_invoice_policy(self):
...
         default invoice policy given by the field with this same name.
-        If the product is type = 'service', we don't have to apply the invoice
+        If the product is detailed_type = 'service', we don't have to apply the invoice
         policy given by the context.
         :return:
         """
         invoice_policy = self.env.context.get("invoice_policy")
         for tmpl in self:
-            if tmpl.type != "service" and invoice_policy:
+            if tmpl.detailed_type != "service" and invoice_policy:
```
chienandalu, line 46: I feel this isn't right. several `detailed_type` categories can be of `type` `service` (e.g.: `event`)

chienandalu:
What's the issue you're trying to solve?

victoralmau:
The initial reason for this change was to set the correct field (`detailed_type`) but odoo has a workaround to resolve this if the "_old_" field name (`type`) is defined.: https://github.com/odoo/odoo/blob/15.0/addons/product/models/product_template.py#L469

chienandalu:
But the two are correct. It only depends on what you need. The rationale for `detailed_type` is that there are serveral types of services with their own workflows. So:

- When you declare a product, the right thing will be to set its `detailed_type`: Example: https://github.com/odoo/odoo/blob/15.0/addons/event_booth_sale/data/product_data.xml#L11
- When you evaluate a product, is very possible that you need the `type` so you don't have to consider every possible detailed type. Example: https://github.com/OCA/OCB/blob/fa97bc043d8c292fa502580599761e39e1d4ebd5/addons/delivery/models/sale_order.py#L21

PR #2856, opened 2023-12-15
[ADD] sale_order_product_recommendation_elaboration

yajo:
New moduule to include elaborations when generating or accepting sale order product recommendations.

@moduon MT-4472

PR #2857, opened 2023-12-15
[IMP] sale_order_product_recommendation: favorites, categories

yajo:
New shiny feature: you can un/favorite products, and those will appear first in the list of recommendations.

imagen

New not-so-shiny feature: recommendations will be sortable by product category.

imagen

While testing the module in the field, it turned out that salespeople often use recommendations based on the product category. Other than that, the favorites option still helps.

@moduon MT-4472

### DIFF
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation.py
@@ -220,11 +220,17 @@ class SaleOrderRecommendation(models.TransientModel):
 class SaleOrderRecommendationLine(models.TransientModel):
     _name = "sale.order.recommendation.line"
     _description = "Recommended product for current sale order"
-    _order = "id"
+    _order = "product_priority desc, id"
 
     currency_id = fields.Many2one(related="product_id.currency_id")
     partner_id = fields.Many2one(related="wizard_id.order_id.partner_id")
     product_id = fields.Many2one("product.product", string="Product")
+    product_categ_id = fields.Many2one(
+        related="product_id.categ_id", readonly=True, store=True
+    )
+    product_priority = fields.Selection(
+        related="product_id.priority", store=True, readonly=False
+    )
     price_unit = fields.Monetary(compute="_compute_price_unit")
     pricelist_id = fields.Many2one(related="wizard_id.order_id.pricelist_id")
     times_delivered = fields.Integer(readonly=True)
--- a/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
+++ b/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml
@@ -25,9 +25,21 @@
                             colspan="2"
                             attrs="{'invisible': [('line_ids', '=', [])]}"
                         >
-                            <tree create="0" delete="0" editable="top">
+                            <tree
+                                create="0"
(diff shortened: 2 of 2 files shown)


sergio-teruel:
You are breaking the module behavior.

Add a setting option to manage this!!

sergio-teruel:
Now you always order by category first.... I know this problem... 
the categ orderd is incompatible with the more delivered order....

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -220,11 +220,20 @@ def action_accept(self):
 class SaleOrderRecommendationLine(models.TransientModel):
     _name = "sale.order.recommendation.line"
     _description = "Recommended product for current sale order"
-    _order = "id"
+    _order = (
+        "product_priority desc, product_categ_id, units_delivered desc, "
+        "times_delivered desc, product_id, id"
+    )
```
yajo: Before this patch, it was just ordered by id. That default order didn't seem very meaningful, that's why I changed it so freely, but I didn't mean to break the module.

I still think it makes sense to put first the favorites, isn't it?

So, the correct default order would be this?

```suggestion
    _order = "product_categ_id, units_delivered desc, times_delivered desc, id"
```

Or would it be just this other one?

```suggestion
    _order = "product_priority desc, id"
```

yajo: I decided to go with the 2nd option, which is more behavior-preserving.

That way, this improvement can be merged faster. Still, on desktop view, users can click on the category header to be able to sort by category.

We will see which other options we have for implementing that order by default later.

rafaelbn:
Hello @sergio-teruel ! Your comments have beed addressed 😄 Do you agree?

rafaelbn:
@yajo ,

> We will see which other options we have for implementing that order by default later.

Diferent companies have diferent default order criteria, so by parameter could be good options. The nice point would be order should be changed with buttons on demand by several options. The main ones are:

- delivered times
- category
- name

yajo:
@rafaelbn with this change, you can already do that in tablets and PC views. You only can't in mobile (because the mobile kanban view doesn't have an ordering option).

Shide (approved):
Code LGTM

I'm agree with this PR because of that comment:
> @rafaelbn with this change, you can already do that in tablets and PC views. You only can't in mobile (because the mobile kanban view doesn't have an ordering option).

rafaelbn:
Please @yajo could you improve this fields of categories?

imagen

yajo:
@rafaelbn If I do that, it'll affect the way categories are ordered. Are you sure you want that?

rafaelbn:
OK @yajo . Lets, by the moment this as is. I guess this column ins optional (property hide)

rafaelbn:
Don't merge this without @sergio-teruel approval please. This impact in actual implementations in production. Thank you.

yajo:
@fcvalgar was getting a bug fixed in https://github.com/OCA/sale-workflow/pull/2856 while doing functional tests, so I rebased.

fcvalgar (approved):
Functional review,

LGTM, thak you @yajo

sergio-teruel (approved):

PR #2858, opened 2023-12-15
[FIX] sale_order_product_recommendation_packaging_default: quick crea…

yajo:
…te packaging

Now it includes the default product in the context, so when you use quick creation actions, it will behave as expected.

@moduon MT-4472


yajo:
Fast-tracking, it's an obvious fix.

/ocabot merge minor

rafaelbn (approved):
👍🏼

PR #2859, opened 2023-12-15
[16.0][ADD] sale_loyalty_industry

santostelmo:
This module allow to restrict the industries in which the sale loyalty program applies.


sale_loyalty_industry/tests/test_program_rules.py
```diff
@@ -0,0 +1,93 @@
...
+        self.immediate_promotion_program.industry_ids = self.industry_A
+        self.steve.industry_id = self.industry_B
+        order = self._create_sale_order()
+        self._auto_rewards(order, self.immediate_promotion_program)
+        self.assertEqual(
+            len(order.order_line.ids),
+            2,
+            "The promo offer is not applied to industry_B.",
+        )
+        self.immediate_promotion_program.industry_ids = self.industry_B
```
simahawk, line 69: these are all different cases, right? If yes, rule of thumb: define 1 method for each of them unless it gets too complicated w/ the setup

simahawk (approved):
LG overall, minor remark for tests

sale_loyalty_industry/tests/test_program_rules.py
santostelmo: Done splitted unit test in several ones

santostelmo:
Moved that PR to sale-promotion: https://github.com/OCA/sale-promotion/pull/197

PR #2862, opened 2023-12-19
[16.0][IMP] sale_stock_picking_note: Notes from partner shipping

EmilioPascual:
When choosing delivery address in sale order, notes are taken from delivery address.
If the delivery address has no notes, they are taken from the customer.

https://www.loom.com/share/d1051894a29b42909ba075d35408964c

@yajo, @Gelojr, @rafaelbn, @Shide or @chienandalu please review.

MT-4381 @moduon

### DIFF
--- a/sale_stock_picking_note/models/sale_order.py
+++ b/sale_stock_picking_note/models/sale_order.py
@@ -22,9 +22,15 @@ class SaleOrder(models.Model):
         readonly=False,
     )
 
-    @api.depends("partner_id")
+    @api.depends("partner_id", "partner_shipping_id")
     def _compute_picking_notes(self):
         for order in self:
             if order.state not in {"sale", "done", "cancel"}:
-                order.picking_note = order.partner_id.picking_note
-                order.picking_customer_note = order.partner_id.picking_customer_note
+                order.picking_note = (
+                    order.partner_shipping_id.picking_note
+                    or order.partner_id.picking_note
+                )
+                order.picking_customer_note = (
+                    order.partner_shipping_id.picking_customer_note
+                    or order.partner_id.picking_customer_note
+                )


carlosdauden (approved):

fcvalgar (approved):
Functional review,

LGTM, thank you @EmilioPascual

chienandalu (approved):

sale_stock_picking_note/models/sale_order.py
```diff
@@ -22,9 +22,15 @@ class SaleOrder(models.Model):
         readonly=False,
     )
 
-    @api.depends("partner_id")
+    @api.depends("partner_shipping_id")
```
yajo: According to the code, these should be the dependencies, right?

```suggestion
    @api.depends("partner_id", "partner_shipping_id")
```

yajo (changes requested):

Gelojr (approved):
Thanks @EmilioPascual 
https://www.loom.com/share/dee104c8e4714fc6b02fd0d052a07569?sid=ddf1f705-d6de-402a-bb49-5777bbca31a6
LGTM

rafaelbn (approved):
👍🏼

sale_stock_picking_note/models/sale_order.py
EmilioPascual: I think it's not neccesary, if you change partner, partner_shipping also changes by this compute https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order.py#L345

pedrobaeza: You shouldn't rely on "external logic" for your depends. If the defining field is `partner_shipping_id`, then put this field as its depends. If tomorrow Odoo (or your customizations) decides that changing partner doesn't change the delivery address, your code will continue working without problems.

EmilioPascual: Ok, I changed it

yajo: Apart from that, if you define dependencies appropriately, you indirectly instruct Odoo the order in which it should process the computes involved in a change.

yajo (approved):
/ocabot merge patch

EmilioPascual:
Tests don't pass, fix https://github.com/OCA/sale-workflow/pull/2868

EmilioPascual:
fix https://github.com/OCA/sale-workflow/pull/2868 merged.
@yajo can you try to merge again, please?

PR #2863, opened 2023-12-19
[15.0][FIX] sale_payment_sheet: Don't set name on stamement lines

pedrobaeza:
Since v14, statement lines are extensions of journal entries, so its fields are the move fields. Setting `name` field of the line means setting the journal entry number, which is not desired and provokes incorrect numbering scheme.

@Tecnativa TT46659

### DIFF
--- a/sale_payment_sheet/models/sale_payment_sheet.py
+++ b/sale_payment_sheet/models/sale_payment_sheet.py
@@ -153,7 +153,6 @@ class SalePaymentSheet(models.Model):
 
                 if key not in vals_dic:
                     vals_dic[key] = {
-                        "name": line.name,
                         "date": line.date,
                         "amount": amount_line,
                         "partner_id": line.partner_id.id,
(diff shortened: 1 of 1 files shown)


sale_payment_sheet/models/sale_payment_sheet.py
```diff
@@ -164,7 +163,6 @@ def button_confirm_sheet(self):
                     }
                 else:
                     vals_dic[key]["amount"] += amount_line
-                    vals_dic[key]["name"] += " {}".format(line.invoice_id.name)
```
sergio-teruel, line 167: Better use vals_dic[key]["name"] += f" {line.invoice_id.name}" ??

sergio-teruel (changes requested):

sale_payment_sheet/models/sale_payment_sheet.py
pedrobaeza: No, `name` mustn't be used under any circumstance, as it's the journal entry number. You have already put this text on other places where they are shown in the statement.

sergio-teruel (approved):

victoralmau (approved):

PR #2865, opened 2023-12-19
[16.0][FIX] sale_payment_sheet: Don't set name on stamement lines

pedrobaeza:
Forward-port of #2863

Since v14, statement lines are extensions of journal entries, so its fields are the move fields. Setting `name` field of the line means setting the journal entry number, which is not desired and provokes incorrect numbering scheme.

@Tecnativa TT46659


pedrobaeza:
Let's fix the problem the same as in previous version:

/ocabot merge patch

PR #2869, opened 2023-12-22
[FIX] sale_order_product_recommendation_elaboration: don't assume the current product is the same than the one in the elaboration

yajo:
Correcting some mistakes due to a misconception on my side, I designed the module and tests assuming the product linked in the elaboration should be the one we're currently selling.

Instead, as the README on `sale_elaboration` says, the elaboration product is an additional product to be added to the SO when delivered, producing a price increase.

@moduon MT-4472


EmilioPascual (approved):

fcvalgar (approved):
Functional review. 


LGTM, thank you @yajo

rafaelbn (approved):
/ocabot merge patch

PR #2873, opened 2023-12-28
[14.0][FIX] name position correction in views

CristianoMafraJunior:
Change in two view files that had the wrong position

### DIFF
--- a/sale_order_revision/view/sale_order.xml
+++ b/sale_order_revision/view/sale_order.xml
@@ -16,7 +16,7 @@
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="priority">15</field>
         <field name="arch" type="xml">
-            <xpath expr="//header" postion='inside'>
+            <xpath expr="//header" position='inside'>
                 <button
                     name="create_revision"
                     states="cancel,sent"
--- a/pricelist_cache/views/res_partner.xml
+++ b/pricelist_cache/views/res_partner.xml
@@ -11,7 +11,7 @@
         <xpath expr="//group[@name='sale']" position="inside">
             <field name="pricelist_cache_product_filter_id" />
         </xpath>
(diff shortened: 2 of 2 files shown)


antoniospneto (approved):

sale_order_revision/view/sale_order.xml
```diff
@@ -16,7 +16,7 @@
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="priority">15</field>
         <field name="arch" type="xml">
-            <xpath expr="//header" postion='inside'>
+            <xpath expr="//header" position='inside'>
```
rousseldenis, line 19: @CristianoMafraJunior I don't see it on runboat:

image

CristianoMafraJunior: primeiraimagem
segundaimagem

This button is associated with an action called create_revision and must be available when the sales order status is 'cancel' or 'sent'.

antoniospneto:
@rousseldenis @CristianoMafraJunior 

It seems that the module is conflicting with another one in the repository, as when only the `sale_order_revision `is installed, the behavior is as expected. I was able to test this using runboat with a clean database:

(https://github.com/OCA/sale-workflow/assets/634278/183fb6c7-2d61-4ac2-833d-5dddba9914cd)
image

sale_order_revision/view/sale_order.xml
CristianoMafraJunior: https://github.com/OCA/sale-workflow/assets/142639425/56c0c526-d6ed-43ac-adf8-920471677150

As demonstrated here in the video, when you click on the convert to order button, the “New Revision of Quotation” button will appear and when the email is sent, the “New Revision of Quotation” button will also appear and this behavior is necessary because of the module = "sale_isolated_quotation", it is this module that adds that "convert to order" button

PR #2879, opened 2024-01-05
[16.0][IMP] sale_stock_picking_note: Not duplicated notes

EmilioPascual:
Duplicate sales or pickings shouldn't duplicated notes

https://www.loom.com/share/2c1063cf16fe41a98be56a7d63474d11

@Gelojr, @yajo please, @chienandalu please review

@moduon MT-4381

### DIFF
--- a/sale_stock_picking_note/models/sale_order.py
+++ b/sale_stock_picking_note/models/sale_order.py
@@ -14,12 +14,14 @@ class SaleOrder(models.Model):
         compute="_compute_picking_notes",
         store=True,
         readonly=False,
+        copy=False,
     )
     picking_customer_note = fields.Text(
         string="Picking Customer Comments",
         compute="_compute_picking_notes",
         store=True,
         readonly=False,
+        copy=False,
     )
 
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

victoralmau (approved):

carlosdauden (changes requested):
This change is somewhat controversial because depending on the flow of each company and the type of note, in some instances it may go well and in others badly.
Therefore, I propose two options:
 - Close this PR and move this changes to the custom code of the instance that needs it. 
 - Add a configurable option at company or system parameter level to determine the behaviour.

EmilioPascual:
> This change is somewhat controversial because depending on the flow of each company and the type of note, in some instances it may go well and in others badly. Therefore, I propose two options:
> 
> * Close this PR and move this changes to the custom code of the instance that needs it.
> * Add a configurable option at company or system parameter level to determine the behaviour.

The team has been thinking about this and we think you are right. The normal odoo flow when duplicating a sale order is duplicate these types of fields.
Thank you @carlosdauden.
I close PR.

PR #2881, opened 2024-01-06
[16.0][IMP] sale_tier_validation: simplify tests

bosd:
Simplify the tests, IMO no need to run the tests twice.

### DIFF
--- a/sale_tier_validation/tests/test_tier_validation.py
+++ b/sale_tier_validation/tests/test_tier_validation.py
@@ -4,7 +4,7 @@ from odoo.exceptions import ValidationError
 from odoo.tests import common, tagged
 
 
-@tagged("post_install")
+@tagged("post_install", "-at_install")
 class TestSaleTierValidation(common.TransactionCase):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
There should be a reason for executing it post install, so instead of removing the tag, just remove the at_install tag.

bosd:
what is the reason for this module to run test at post_install?

I went trough all the tier_validation modules across the repos. It looked random, some where at install other at post install.

Odoo defaults to at_install.
In that case no tag is needed. 


As of this module. In the migration to v16 the `-at_install` got removed.
W/o any comment about reason..
Making the tests run twice

Dranyel-Bosd (approved):
LGTM :+1:

pedrobaeza:
Then please re-add `-at_install` for keeping the conditions in previous version. A possible reason is if there's any invoice creation involved, that the only way to assure that the chart template is loaded is to do it post-load.

rousseldenis (approved):
Makes sense

PR #2883, opened 2024-01-06
[FIX][16.0] sale_quotation_number: deprecated call to Savepointcase

bosd:


pedrobaeza (approved):
Fast tracking as obvious

PR #2885, opened 2024-01-08
[16.0][REF] sale_triple_discount: Consolidate discount in std field

grindtildeath:
Until now, the triple discount feature did use the standard Odoo field as the first discount field, adding only extra fields for the second and the third discount. This implied we had to override any function using discount to consider the other discounts properly.

By adding an extra field discount1 to store the first discount, it allows to redefine the standard discount field to a computed field that will consolidate the triple discount from the other fields, and avoid the need to redefine anything relying on the discount field as it will already consider the triple discounts.

Relates to:
 - [ ] https://github.com/OCA/account-invoicing/pull/1638

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -15,12 +15,21 @@ class SaleOrderLine(models.Model):
 
     discount = fields.Float(
         string="Total discount",
+        store=True,
+        compute="_compute_discount_consolidated",
+        compute_sudo=True,
+        precompute=True,
         readonly=True,
     )
     discount1 = fields.Float(
         string="Disc. 1 (%)",
         digits="Discount",
+        store=True,
         default=0.0,
+        compute="_compute_discount",
+        compute_sudo=True,
+        precompute=True,
+        readonly=False,
     )
     discount2 = fields.Float(
         string="Disc. 2 (%)",
@@ -69,24 +78,54 @@ class SaleOrderLine(models.Model):
         final_discount = 1
         for discount in discounts:
             final_discount *= discount
-        return 100 - final_discount * 100
+        result = 100 - final_discount * 100
+        dp = self.env.ref("product.decimal_discount").precision_get("Discount")
+        return round(result, dp)
 
     @api.model
     def _discount_fields(self):
         return ["discount1", "discount2", "discount3"]
 
-    # pylint: disable=W8110
-    @api.depends("discount1", "discount2", "discount3", "discounting_type")
+    # Copy of Odoo function to change field being assigned from discount to discount1
+    @api.depends("product_id", "product_uom", "product_uom_qty")
     def _compute_discount(self):
-        super()._compute_discount()
         for line in self:
-            # Reset discount if not done in super
+            if not line.product_id or line.display_type:
+                line.discount1 = 0.0
+
             if not (
                 line.order_id.pricelist_id
                 and line.order_id.pricelist_id.discount_policy == "without_discount"
             ):
-                line.discount = 0.0
-            line.discount += line._get_final_discount()
+                continue
+
+            line.discount1 = 0.0
+
+            if not line.pricelist_item_id:
+                # No pricelist rule was found for the product
+                # therefore, the pricelist didn't apply any discount/change
+                # to the existing sales price.
+                continue
+
+            line.discount1 = line._calc_discount_from_pricelist()
+
+    def _calc_discount_from_pricelist(self):
+        self.ensure_one()
+        self = self.with_company(self.company_id)
+        pricelist_price = self._get_pricelist_price()
+        base_price = self._get_pricelist_price_before_discount()
+
+        if base_price != 0:  # Avoid division by zero
+            discount = (base_price - pricelist_price) / base_price * 100
+            if (discount > 0 and base_price > 0) or (discount < 0 and base_price < 0):
+                # only show negative discounts if price is negative
+                # otherwise it's a surcharge which shouldn't be shown to the customer
+                return discount
+
+    @api.depends("discount1", "discount2", "discount3", "discounting_type")
+    def _compute_discount_consolidated(self):
+        for line in self:
+            line.discount = line._get_final_discount()
 
     _sql_constraints = [
         (
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -151,6 +151,7 @@ class SaleOrderLine(models.Model):
         more discount fields to the invoice lines
         """
         res = super()._prepare_invoice_line(**kwargs)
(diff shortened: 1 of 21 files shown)


grindtildeath:
ping @nicomacr @jjscarafia @tafaRU @chienandalu @SirAionTech @jbaudoux @leemannd @lethuthao2403 @nguyenminhchien

chienandalu (approved):
I like the approach :+1: (we don't maintain this module currently though)

francesco-ooops:
@grindtildeath any plan to bring this to completion?

grindtildeath:
@francesco-ooops I didn't have any time during the last few weeks and I'm going on holidays the next 2 ones, so I won't be able to work on it before some time.

IMO the issue to get this merged is recomputation of old moves in periods closed by fiscal year lock date as discussed in https://github.com/OCA/account-invoicing/pull/1638

sale_triple_discount/models/sale_order_line.py
```diff
@@ -11,6 +11,11 @@
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    discount1 = fields.Float(
+        string="Disc. 1 (%)",
+        digits="Discount",
+        default=0.0,
+    )
```
ajaniszewska-dev: @grindtildeath  any reason to not add this field  in the view: https://github.com/OCA/sale-workflow/blob/16.0/sale_triple_discount/views/sale_order_report.xml#L7 ?

jjscarafia:
Hi @grindtildeath 
We like the approach. Actually we've a similar legacy approach that we're using [here](https://github.com/ingadhoc/sale/tree/16.0/sale_three_discounts)

We've created this module before this OCA one and never migrate to use the OCA one due to lack of time but also because somewhow we like more our approach (same as you share).

Last thing, I believe this kind of changes are not good to be done on an already working version since long time. Maybe better for 17 or 18?

grindtildeath:
@jjscarafia Thanks for sharing your work.

It's not easy finding a good solution while the `discount` field is already computed by Odoo to store a discount coming from the pricelist.
I found a compromise here in the last version, would love to have feedback from those who are used to this module.

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo16-addon-account-invoice-triple-discount @ git+https://github.com/OCA/account-invoicing@refs/pull/1638/head#subdirectory=setup/account_invoice_triple_discount
```
rousseldenis:
```suggestion
odoo-addon-account-invoice-triple-discount @ git+https://github.com/OCA/account-invoicing@refs/pull/1638/head#subdirectory=setup/account_invoice_triple_discount
```
@grindtildeath

ajaniszewska-dev (approved):

grindtildeath:
@legalsylvain rebased

legalsylvain:
- Could you take a look on red CI ?
- you have to remove changes in test-requirement
- I think cherry pick https://github.com/OCA/sale-workflow/pull/3598 could help.

grindtildeath:
@legalsylvain I'm sorry but I don't have any time left to spend on that :disappointed: 

I still did rebase this PR on 16.0 and dropped changes in test requirements/oca dependencies. However I tried cherry-picking your commits and there are conflicts on `.github/workflows/test.yml`

Maybe you can fix the conflict and open a PR targeting this branch? I'll merge it as soon as possible.

Otherwise feel free to open a new PR and take over this work if it's easier for you.

legalsylvain:
Hi @grindtildeath. Thanks for your time.

> I still did rebase this PR on 16.0 and dropped changes in test requirements/oca dependencies. However I tried cherry-picking your commits and there are conflicts on .github/workflows/test.yml

No worry with test.yml, it is an autogenerated file. you can recreate it with ``copier copy gh:oca/oca-addons-repo-template . --trust`` when there is conflict.

> Maybe you can fix the conflict and open a PR targeting this branch? I'll merge it as soon as possible.

[Done here.](https://github.com/grindtildeath/sale-workflow/pull/12)

> Otherwise feel free to open a new PR and take over this work if it's easier for you.

That's not possible. I don't use sale_triple_discount. i just want to unblock many PRs that are stucked, since account_triple_discount refactor has been merged and not this one. 
In my PR, I fixed an error sale_order_general_discount. But I faced other errors with sale_pricelist_triple_discount for exemple. I think this current PR has to be finished.
Could you take a look ? 
Thanks !

victoralmau:
Can we unlock this? Another possibility will be to exclude these modules from the CI because they are locked to different PRs.

rousseldenis:
@grindtildeath @legalsylvain Due to : https://github.com/OCA/account-invoicing/pull/1840

All repos depending on them are not running tests. We should land on this.

rousseldenis:
@grindtildeath My question: shouldn't it be possible to make both modules compatible after refactor?

victoralmau:
This is still pending, is there any proposal to correct tests?

rousseldenis:
@legalsylvain In fact, this is quite bad that https://github.com/OCA/account-invoicing/pull/1840 has been merged and not this one...

I think I will try to make current sale_triple_discount to work with a freezed version of account_triple_discount (before refactor).

legalsylvain:
> @legalsylvain In fact, this is quite bad that https://github.com/OCA/account-invoicing/pull/1840 has been merged and not this one...

Indeed ! It's a pity. Thanks for your quick fix.

sale_triple_discount/models/sale_order_line.py
```diff
@@ -60,19 +75,61 @@ def _multiplicative_discount(self):
...
-    def _compute_amount(self):
-        with self._aggregated_discount() as lines:
-            res = super(SaleOrderLine, lines)._compute_amount()
-        return res
+    # Copy of Odoo function to change field being assigned from discount to discount1
+    @api.depends("product_id", "product_uom", "product_uom_qty")
+    def _compute_discount(self):
+        for line in self:
+            if not line.product_id or line.display_type:
+                line.discount1 = 0.0
```
hildickethan, line 91: @grindtildeath does this not break every `_compute_discount` that inherits the base Odoo one? they either won't run, will error or won't write onto `discount` since it isn't its compute anymore
any reason for this approach here and not on the invoice version?
regardless I will continue looking into bringing this up to date as I know you don't have time

hildickethan:
I made a new PR https://github.com/OCA/sale-workflow/pull/3907 based on this one to take another look at this refactoring, there are some different changes compared to this one but the general idea is the same
The use cases and module combinations I checked all seemed to work as expected, but I would appreciate it if some people involved with this PR could check it out and see if we can close this issue and be able to properly migrate higher versions

rousseldenis:
@grindtildeath

grindtildeath:
Closed in favour of https://github.com/OCA/sale-workflow/pull/3907

PR #2886, opened 2024-01-09
[17.0][MIG] sale_product_set: Migration to 17.0

NICO-SOLUTIONS:
standard migration to 17.0


NICO-SOLUTIONS:
Depends on:

- [x] https://github.com/OCA/product-attribute/pull/1554

MohamedOsman7:
Code LGTM. The failed check needs to be resolved in order to properly test the functionality using the runboat.

NICO-SOLUTIONS:
> Code LGTM. The failed check needs to be resolved in order to properly test the functionality using the runboat.

The test fails, because product_set is not merged yet and test-requirements is using this PR for it. I rebased it for rebuilding runboat. Feel free to check it

MohamedOsman7 (approved):
code & functional review LGTM

CRogos (changes requested):
Could you also cleanup the commit history?

And I think I've found a bug.
When you switch between product sets, The product list below is not updating.
image

sale_product_set/wizard/product_set_add.xml
```diff
@@ -31,7 +31,7 @@
                             <field
                                 name="product_id"
                                 options="{'no_open': True, 'no_create': True}"
-                                attrs="{'required': [('display_type', '=', False)]}"
+                                required="display_type == 'False'"
```
CRogos: Shouldn't this be?
`required="not display_type"`

CRogos (changes requested):
The above mentioned behavior is already in Odoo 16.

sale_product_set/wizard/product_set_add.xml
NICO-SOLUTIONS: yep...

NICO-SOLUTIONS:
> The above mentioned behavior is already in Odoo 16.

you mean that the lines are not updating is still present in V16? i´ll have a look later on. Here it does now

CRogos (approved):
LGTM code + tested.
(I did the Odoo 16 test with our DB, and at least we have this issue. But maybe we are not on the latest version.)

Just a question... shouldn't we move this button?
image
Not necessary as part of this migration but in general I think it would be much more natural there?

NICO-SOLUTIONS:
Yeah I like the idea. But I guess there will be be returning questions why this disappeared 😉
But you're right. It would fit better

CRogos:
> Yeah I like the idea. But I guess there will be be returning questions why this disappeared 😉 But you're right. It would fit better

maybe, but when changing a major version there are also a lot other changes and the user is used to it. 😉

NICO-SOLUTIONS:
True 😉
I'll attack this soon.

pedrobaeza (approved):
You can't put the button as another option of adding lines in the one2many, as the way this is built doesn't allow to do that I'm afraid.

/ocabot merge nobump

PR #2899, opened 2024-01-15
[16.0][ADD]sale_quotation_capable_vendor_rfq: add module

DemchukM:
This module allows you to create RFQs directly from sales quotation. Requests for quotations will be created for all suppliers who are capable of supplying products of the sales quotation.


sale_quotation_capable_vendor_rfq/models/purchase_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    sale_order_id = fields.Many2one(
+        "sale.order",
+        string="Sale Order",
```
dmitriypaulov: Unnecessary string parameter in field definition

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+        return vendors_dict.keys()
+
+    def _create_purchase_orders(self):
+        """
+        Create purchase orders from this sale order
+        """
+        self.ensure_one()
+        vendors_dict = self._get_vendor_dict()
+        for vendor, value in vendors_dict.items():
+            vals_purchase = {
```
dmitriypaulov: In order to make this method easier to override later, I suggest you to move values preparation into a separate method
```python
purchase = self.env["purchase.order"].create(
    self._prepare_purchase_order_vals(vendor, lines)
)
```

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+                "partner_id": vendor.id,
+                "origin": self.name,
+                "sale_order_id": self.id,
+                "currency_id": self.currency_id.id,
+                "company_id": self.company_id.id,
+            }
+            purchase = self.env["purchase.order"].create(vals_purchase)
+            for line in value:
+                vals_purchase_line = {
```
dmitriypaulov: Same as previous comment here

dmitriypaulov (changes requested):

sale_quotation_capable_vendor_rfq/models/product_product.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # get the lowest price from each vendor
+        min_prices = {}
+        for info in supplierinfo:
+            if info.partner_id not in min_prices:
+                min_prices[info.partner_id] = info
+            else:
+                if info.price < min_prices[info.partner_id].price:
+                    min_prices[info.partner_id] = info
+        return min_prices.values()
```
ivs-cetmix: I think it will return a list not a recordset

sale_quotation_capable_vendor_rfq/models/purchase_order.py
ivs-cetmix: string="Quotation"

sale_quotation_capable_vendor_rfq/models/purchase_order.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2024 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    sale_order_id = fields.Many2one(
```
ivs-cetmix: Better use something like `sale_quotation_id`
Because `sale_order_id` is too common name which might be occasionally reused by some other module.

sale_quotation_capable_vendor_rfq/models/purchase_order.py
```diff
@@ -0,0 +1,14 @@
...
+        "sale.order",
+        string="Sale Order",
+        help="Sale order that created this purchase order",
```
ivs-cetmix: Quotation this purchase order was created from

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    purchase_order_ids = fields.One2many(
```
ivs-cetmix: We are actually creating RFQs not purchase orders. Also here we can run into the same issue when some other module will reimplement the same fields. Because its name is to generic.
So `rfq_ids` would be better.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+        "purchase.order",
+        "sale_order_id",
+        string="Purchase Orders",
```
ivs-cetmix: RFQs

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+        help="Purchase orders created from this sale order",
+    )
+    purchase_count = fields.Integer(
```
ivs-cetmix: `rfq_count`

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,118 @@
...
+        string="Purchase Orders",
```
ivs-cetmix: RFQs

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+            purchase = self.env["purchase.order"].create(
+                self._prepare_purchase_order_vals(vendor)
+            )
+            for line in value:
+                self.env["purchase.order.line"].create(
+                    self._prepare_purchase_order_line_vals(purchase, line)
+                )
+
+    def _prepare_purchase_order_vals(self, vendor):
+        return {
```
ivs-cetmix, line 86: Docstring is missing

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+            "partner_id": vendor.id,
+            "origin": self.name,
+            "sale_order_id": self.id,
+            "currency_id": self.currency_id.id,
+            "company_id": self.company_id.id,
+        }
+
+    def _prepare_purchase_order_line_vals(self, purchase, line):
+        return {
```
ivs-cetmix, line 103: Docstring is missing

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+            "product_uom": line["line"].product_uom.id,
+            "price_unit": self._convert_to_currency(
+                line["supplierinfo"].price,
+                line["supplierinfo"].currency_id,
+                purchase.currency_id,
+            ),
+            "date_planned": fields.Datetime.now(),
+        }
+
+    def _get_vendor_dict(self, lead_time=0):
```
ivs-cetmix: Docstring is missing

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+                    "line": line,
+                }
+                if supplierinfo.partner_id not in vendors_dict:
+                    vendors_dict[supplierinfo.partner_id] = [vals]
+                else:
+                    vendors_dict[supplierinfo.partner_id].append(vals)
+        return vendors_dict
+
+    def _convert_to_currency(self, price, currency_id, currency_to_id):
+        """
```
ivs-cetmix, line 141: Docsting is incomplete

ivs-cetmix (changes requested):
This is just some code review. Functional is to be done yet

angelinaanaki:
@DemchukM 

Functional review: 

The RFQ is generated for all vendors of the product template, regardless of the selected variant in the Sales Quotation.

Example: [Sale Quotation](http://oca-sale-workflow-16-0-pr2899-8f89053a85a9.runboat.odoo-community.org/web#id=31&cids=1&menu_id=349&action=573&model=sale.order&view_type=form) and [Product Template](http://oca-sale-workflow-16-0-pr2899-8f89053a85a9.runboat.odoo-community.org/web#id=9&cids=1&menu_id=349&action=585&model=product.template&view_type=form)

angelinaanaki (approved):
Functional review: LGTM

sale_quotation_capable_vendor_rfq/demo/vendors.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo noupdate="0">
+    <record model="res.partner" id="partner_1">
```
ivs-cetmix: Please use readable ids. Because it's much easies to read the code later
`id="partner_doge"`
Same for all other records

ivs-cetmix (changes requested):

ivs-cetmix (approved):
Functional LGTM

rousseldenis:
@DemchukM Still draft ?

sale_quotation_capable_vendor_rfq/models/purchase_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    sale_quotation_id = fields.Many2one(
+        "sale.order",
+        string="Quotation",
+        help="Quotation this purchase order was created from",
+    )
```
yelizariev, line 14: Maybe use name `sale_order_id` similar to the field in [sale_purchase](
https://github.com/odoo/odoo/blob/a21cbb54121d97d7b10a48beb5c1fad4d79bfe01/addons/sale_purchase/models/purchase_order.py#L70-L74) module.

`sale_quotation_id` is attached to the workflow, which might be changed later.

sale_quotation_capable_vendor_rfq/models/product_product.py
```diff
@@ -0,0 +1,38 @@
...
+        for info in supplierinfo:
+            if info.partner_id not in min_prices:
+                min_prices[info.partner_id] = info
+            else:
+                if info.price < min_prices[info.partner_id].price:
+                    min_prices[info.partner_id] = info
+        supplierinfo_ids = [info.id for info in min_prices.values()]
+        if not supplierinfo_ids:
+            return []
+        return self.env["product.supplierinfo"].browse(supplierinfo_ids)
```
yelizariev: This should work

```suggestion
        return self.env["product.supplierinfo"].concat(*min_prices.values())
```

It also fixes the issue that result type must be the same even on empty record set

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,175 @@
...
+
+        :param lead_time: Integer of days to match
+
+        :return: dict
+        """
+        vendors_dict = {}
+        for line in self.order_line:
+            supplierinfo_ids = line.product_id._get_matching_vendor_pricelists(
+                line.product_uom_qty, lead_time
+            )
```
yelizariev: `supplierinfo_ids` is  not a good name in this context. The variable names with  `_ids` ending is usually for list of id, not for a record set.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,175 @@
...
+            for supplierinfo in supplierinfo_ids:
+                vals = {
+                    "supplierinfo": supplierinfo,
+                    "line": line,
+                }
+                if supplierinfo.partner_id not in vendors_dict:
+                    vendors_dict[supplierinfo.partner_id] = [vals]
+                else:
+                    vendors_dict[supplierinfo.partner_id].append(vals)
```
yelizariev: You can use `defaultdict`.

`supplierinfo = defaultdict(list)`

```suggestion
                vendors_dict[supplierinfo.partner_id].append(vals)               
```

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,175 @@
...
+        :param currency_id: res.currency
+        :param currency_to_id: res.currency
+
+        :return: float
+        """
+        if currency_id != currency_to_id:
+            price = currency_id._convert(
+                price, currency_to_id, self.env.company, fields.Date.today()
+            )
+        return price
```
yelizariev, line 154: I don't think this method is really needed. First, it's not related the sale order model. Second, you can do the same by calling currency's [method](https://github.com/odoo/odoo/blob/a21cbb54121d97d7b10a48beb5c1fad4d79bfe01/odoo/addons/base/models/res_currency.py#L278
) directly:

```py
line["supplierinfo"].currency_id._convert(
    line["supplierinfo"].price,
    rfq.currency_id,
    self.env.company,
    fields.Date.today()
)
```

It still makes sense to make a wrapper for the method, but since it's used in a single place only, the new method is not needed

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+    def _get_alternative_group_id(self):
+        """
+        Get alternative group id
+
+        :return: purchase.order.group
+        """
+        self.ensure_one()
+        return self.env["purchase.order.group"].create({}).id
```
yelizariev: Method names are confusing.

Do we really need a new method to create empty record?

Also, why word "alternative" is used in this context?

sale_quotation_capable_vendor_rfq/models/purchase_order.py
DemchukM: At first it was like that, but later they asked to change the name to the one it is now, here is the link to the comment
https://github.com/OCA/sale-workflow/pull/2899#discussion_r1453852181

ivs-cetmix:
> @DemchukM Still draft ?

Hi @rousseldenis we have fixed all the issued we managed to find so now this PR is ready for review.

yelizariev (approved):
I've checked updates related to my previous review. It looks good.

sale_quotation_capable_vendor_rfq/models/product_product.py
```diff
@@ -0,0 +1,35 @@
...
+
+        :param quantity: Quantity to match
+        :param output_time: Integer of days to match
+        :return: Recordset of product.supplierinfo
+        """
+        self.ensure_one()
+
+        # get all vendors that match the criteria
+        supplierinfo = self.seller_ids.filtered(
+            lambda s: (not output_time or s.delay >= output_time)
```
ivs-cetmix: Let's assume `output_time == 10`. This means that we need to find all vendors that can deliver faster than 10 days.
Shouldn't it be `s.delay <= output_time`?

sale_quotation_capable_vendor_rfq/models/product_product.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (C) 2024 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_matching_vendor_pricelists(self, quantity, output_time=0):
```
ivs-cetmix: `output_time` is a confusing name for this parameter. Should be something like `days_max`.
Let's also add a lower bottom and check it too. So there will be two of them:
`days_min`: minimum delivery lead time, days
`days_max`: maximum delivery lead time, days

ivs-cetmix (changes requested):

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+        Returns a dict of res.partner records that match the criteria
+
+        :param lead_time: Integer of days to match
+
+        :return: dict
+        """
+        vendors_dict = defaultdict(list)
+        for line in self.order_line:
+            supplierinfos = line.product_id._get_matching_vendor_pricelists(
+                line.product_uom_qty, lead_time
```
ivs-cetmix: You are providing `days_min` here. However we need to select all vendors that can deliver **same or faster** number of days. Which means that we need to provide `days_max` instead.
However we should add both `days_min` and `days_max` to the parent function. So need to provide both of them here too.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Create purchase orders from this sale order
+        """
+        self.ensure_one()
+        self._create_rfq()
+        if self.rfq_count:
+            return self.action_view_rfq()
+        raise UserError(_("No RFQs created"))
+
+    def _get_capable_vendors(self, lead_time=0):
```
ivs-cetmix: Let's add `days_min` and `days_max` here too to be able to filter vendors based on lead time.

ivs-cetmix (changes requested):

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+        self._create_rfq()
+        if self.rfq_count:
+            return self.action_view_rfq()
+        raise UserError(_("No RFQs created"))
+
+    def _get_capable_vendors(self, day_min=0, day_max=0):
+        """
+        Returns a recordset of res.partner records that match the criteria
+
+        param lead_time: Integer of days to match
```
angelinaanaki: Fix docstring.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+                fields.Date.today(),
+            ),
+            "date_planned": fields.Datetime.now(),
+        }
+
+    def _get_vendor_dict(self, day_min=0, day_max=0):
+        """
+        Returns a dict of res.partner records that match the criteria
+
+        :param lead_time: Integer of days to match
```
angelinaanaki: Also here.

angelinaanaki (changes requested):

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Create purchase orders from this sale order
+        """
+        self.ensure_one()
+        self._create_rfq()
+        if self.rfq_count:
+            return self.action_view_rfq()
+        raise UserError(_("No RFQs created"))
+
+    def _get_capable_vendors(self, day_min=0, day_max=0):
```
angelinaanaki:
```suggestion
    def _get_capable_vendors(self, days_min=0, days_max=0):
```

Please ensure the names match the previous changes.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,165 @@
...
+            "price_unit": line["supplierinfo"].currency_id._convert(
+                line["supplierinfo"].price,
+                rfq.currency_id,
+                self.env.company,
+                fields.Date.today(),
+            ),
+            "date_planned": fields.Datetime.now(),
+        }
+
+    def _get_vendor_dict(self, day_min=0, day_max=0):
```
angelinaanaki:
```suggestion
    def _get_vendor_dict(self, days_min=0, days_max=0):
```
Also here.

angelinaanaki (changes requested):

sale_quotation_capable_vendor_rfq/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+    "version": "16.0.1.0.0",
+    "category": "Sales",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Cetmix, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "sale_management",
+        "purchase",
```
GabbasovDinar: you can remove, because purchase_requisition depends on purchase module

sale_quotation_capable_vendor_rfq/models/product_product.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        self.ensure_one()
+
+        # get all vendors that match the criteria
+        supplierinfo = self.seller_ids.filtered(
+            lambda s: (not days_min or s.delay >= days_min)
+            and (not days_max or s.delay <= days_max)
+            and s.min_qty <= quantity
+            and (s.product_id == self or not s.product_id)
+        )
```
GabbasovDinar, line 27: maybe will faster:

```python
        domain = [
            ('product_tmpl_id', '=', self.product_tmpl_id.id),
            ('min_qty', '<=', quantity),
        ]
        if days_min:
            domain.append(('delay', '>=', days_min))
        if days_max:
            domain.append(('delay', '<=', days_max))
        
        supplierinfo = self.seller_ids.filtered_domain(domain)

        min_prices = {}
        for info in supplierinfo:
            if info.partner_id not in min_prices or info.price < min_prices[info.partner_id].price:
                min_prices[info.partner_id] = info
        
        return self.env['product.supplierinfo'].concat(*min_prices.values())
```

GabbasovDinar (changes requested):

sale_quotation_capable_vendor_rfq/models/product_product.py
angelinaanaki: Using the domain `('product_tmpl_id', '=', self.product_tmpl_id.id)` retrieves all supplier information associated with the product template, regardless of the specific product variant. This can lead to issues if different variants have distinct supplier information. Please correct me if I'm wrong.

GabbasovDinar (approved):

dmitriypaulov (approved):
LGTM

ivs-cetmix (approved):
Functional LGTM

angelinaanaki (approved):

youring:
Nice feature addition! Thanks.
Minor typo: readme/CONTEX > CONTEXT

BTW, would you also consider adding permission control before the first merge?
It seems now only Administration/Settings group user can use the Create RFQs button.
I tested with demo user:
`You are not allowed to access 'Action Window' (ir.actions.act_window) records.
This operation is allowed for the following groups:
	- Administration/Settings
Contact your administrator to request access if necessary.`

Additional enhancement suggestion:

- Check if RFQs already created/exist upon click the Create RFQ button
- Select preferred supplier only before create RFQs, i.e. a capable vendor list pop up window upon click the Create RFQ button.

ivs-cetmix:
> Nice feature addition! Thanks. Minor typo: readme/CONTEX > CONTEXT
> 
> BTW, would you also consider adding permission control before the first merge? It seems now only Administration/Settings group user can use the Create RFQs button. I tested with demo user: `You are not allowed to access 'Action Window' (ir.actions.act_window) records. This operation is allowed for the following groups: - Administration/Settings Contact your administrator to request access if necessary.`
> 
> Additional enhancement suggestion:
> 
> * Check if RFQs already created/exist upon click the Create RFQ button
> * Select preferred supplier only before create RFQs, i.e. a capable vendor list pop up window upon click the Create RFQ button.

 @youring thank you for the review and proposed features! Will add them on the list 👍

hoangtrann:
Nice pull request, this is what we stumble upon as well working with a client who source products for their customer with dropshipping.

I have a couple of comments as follows:

1/ Create RFQ should only be visible in Quotation (`draft`), Quotation Sent (`sent`), and maybe Sales Order (`sale`) with limited accessibility.
2/ Currently, a user can hit `Create RFQ` multiple times, and nothing prevents them not to. E.g user creates a quote, hits `Create RFQ`, goes back to the quote, and hits `Create RFQ` again --> there will be new RFQs created.
3/The Icon being used for RFQ might be replaced with another one to avoid confusion with Delivery.
4/ If Call for Tender is what you are looking for, it can be confusing when comes to the scenario that you would like to purchase a product from this vendor and another product from another vendor. RFQs for these 2 vendors can have the same set of products for instance. Confirming one RFQ and Odoo will prompt for canceling or keeping the other, which I find confusing.
5/ RFQ count should only count anything that isn't canceled I think.
6/ Cancelling a quotation having RFQs confirmed should be prevented and request the user to cancel the Purchase Orders first. Or we should have a wizard prompts for action, or a warning.
7/ I haven't checked dropshipping, but selected vendor/products should be tied to the Quotation so that later confirming sales order will create dropshipping via the pre-created Purchase Orders.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        for order in self:
+            order.rfq_count = len(order.rfq_ids)
+
+    def action_view_rfq(self):
+        """
+        View the purchase orders created from this sale order
+        """
+        self.ensure_one()
+        action = self.env.ref("purchase.purchase_rfq").sudo().read()[0]
```
hoangtrann, line 37:
```suggestion
        action = self.env["ir.actions.actions"]._for_xml_id("purchase.purchase_rfq")
```
This seems to be more align with new Odoo code-base these days

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def action_create_rfq(self):
+        """
+        Create purchase orders from this sale order
+        """
+        self.ensure_one()
+        self._create_rfq()
+        if self.rfq_count:
+            return self.action_view_rfq()
+        raise UserError(_("No RFQs created"))
```
hoangtrann, line 49: I think we should use `Warning` here instead of `UserError`, it isn't user's fault to not knowing if the product doesn't have matching supplierinfo

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+        Create purchase orders from this sale order
+        """
+        self.ensure_one()
+        vendors_dict = self._get_vendor_dict()
+        for vendor, value in vendors_dict.items():
+            purchase = self.env["purchase.order"].create(self._prepare_rfq_vals(vendor))
+            for line in value:
+                self.env["purchase.order.line"].create(
+                    self._prepare_rfq_line_vals(purchase, line)
+                )
```
hoangtrann, line 75:
```suggestion
        vendors_dict = self._get_vendor_dict()
        vals_list = []
        for vendor, value in vendors_dict.items():
            vals = self._prepare_rfq_vals(vendor)
            vals["line_ids"] = [
                Command.create(self._prepare_rfq_line_vals(line))
                for line in value
            ]
            vals_list.append(vals)
        self.env["purchase.order"].create(vals_list)
```

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+            "name": line["line"].name,
+            "product_qty": line["line"].product_uom_qty,
+            "product_uom": line["line"].product_uom.id,
+            "price_unit": line["supplierinfo"].currency_id._convert(
+                line["supplierinfo"].price,
+                rfq.currency_id,
+                self.env.company,
+                fields.Date.today(),
+            ),
+            "date_planned": fields.Datetime.now(),
```
hoangtrann, line 115:
```suggestion
            "product_id": line["line"].product_id.id,
            "name": line["line"].name,
            "product_qty": line["line"].product_uom_qty,
            "product_uom": line["line"].product_uom.id,
            "price_unit": line["supplierinfo"].currency_id._convert(
                line["supplierinfo"].price,
                line["supplierinfo"].partner_id.property_purchase_currency_id or self.currency_id,
                self.env.company,
                fields.Date.today(),
            ),
            "date_planned": fields.Datetime.now(),
```

hoangtrann: Should also add `self.ensure_one()`

sale_quotation_capable_vendor_rfq/models/sale_order.py
hoangtrann: Should also add `self.ensure_one()`

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+            order.rfq_count = len(order.rfq_ids)
+
+    def action_view_rfq(self):
+        """
+        View the purchase orders created from this sale order
+        """
+        self.ensure_one()
+        action = self.env.ref("purchase.purchase_rfq").sudo().read()[0]
+        action["domain"] = [("id", "in", self.rfq_ids.ids)]
+        return action
```
hoangtrann, line 39: I would suggest returning a form view if count = 1, else return list view.

sale_quotation_capable_vendor_rfq/models/sale_order.py
```diff
@@ -0,0 +1,167 @@
...
+
+        :param vendor: res.partner
+
+        :return: dict
+        """
+        return {
+            "partner_id": vendor.id,
+            "origin": self.name,
+            "sale_quotation_id": self.id,
+            "currency_id": self.currency_id.id,
```
hoangtrann, line 90:
```suggestion
            "currency_id": vendor.property_purchase_currency_id.id or self.currency_id.id,
```

sale_quotation_capable_vendor_rfq/models/sale_order.py
hoangtrann: I think we can remove the `rfq` param for this method and combine into a single `create` call

sale_quotation_capable_vendor_rfq/models/sale_order.py
hoangtrann: Should combine `create` into a single call if possible

PR #2900, opened 2024-01-16
[14.0][IMP] pricelist_cache: add some indexes

TDu:


rousseldenis:
@TDu So many ? Could you point out the searches that uses those fields ?

TDu:
> @TDu So many ? Could you point out the searches that uses those fields ?

I can not because I did not analyze the queries on the related production system.
I will get back with some of answer.
But I understand it is a lot of indexes and that could be counter productive.

sebalix:
The searches are done by Odoo methods that `pricelist_cache` module triggers. Queries were spotted after some months with `pg_stat_statements`. Trying to find the search:
- `company_id`: done automatically by the ORM I guess (should be in Odoo core?)
- `base_pricelist_id`: in `unlink`: https://github.com/odoo/odoo/blob/14.0/addons/product/models/product_pricelist.py#L352-L353 (should be in Odoo core?)
- `product_id`/`product_tmpl_id`: when deleting a product, the `ondelete="cascade"` needs it (in Odoo core?)
- `applied_on`, `min_quantity`, `categ_id`: triggered by the default order https://github.com/odoo/odoo/blob/14.0/addons/product/models/product_pricelist.py#L370
- `date_start`/`date_end`: triggered by the SQL query in `_compute_price_rule_get_items` https://github.com/odoo/odoo/blob/14.0/addons/product/models/product_pricelist.py#L93

These indexes should be included in Odoo core, but not sure Odoo will merge this? And as `pricelist_cache` is using a lot the pricelists mechanism, these indexes are very welcome for this module.

rousseldenis (approved):
That makes sense in this context

mmequignon (approved):

sebalix:
@mmequignon as this requires an update of the module to be effective (to create index), `minor` would have been better here, not a big deal

PR #2907, opened 2024-01-18
[FIX] sale_order_line_field_from_product_attribute: run on create as well instead of only on write + support Booleans

dreispt:


sale_order_line_field_from_product_attribute/models/sale_order_line.py
```diff
@@ -26,5 +25,15 @@ def write(self, vals):
                     )
                     pav_vals[field_name] = value
                 if pav_vals:
-                    super().write(pav_vals)
+                    sol.write(pav_vals)
```
rousseldenis: @dreispt So, you call write() in another write() ???

rousseldenis (changes requested):
IMHO, this change is not wanted

sale_order_line_field_from_product_attribute/models/sale_order_line.py
dreispt: @rousseldenis Indeed, it was working but it isn't proper code, so introduced a per record super().

dreispt:
Added a commit to properly support falsy values on Boolean fields.

PR #2910, opened 2024-01-19
[FIX] sale_order_product_recommendation: don't destroy choices when criteria changes

yajo:
Before this patch:
1. Open the recommendations wizard.
2. Generate recommendations.
3. Use those recommendation lines to add some products to the SO.
4. Change some filter parameter.
5. 💣 All changes you made are lost.

Now, to change filter criteria, you have to click on a button that will warn you about changes being lost. Then, you can choose to not do that, and instead save before opening a new recommendations wizard.

@moduon MT-4472


rafaelbn (approved):
OK, thank you! 👍🏼

PR #2913, opened 2024-01-22
[IMP] sale_order_product_recommendation: choose order on mobile

yajo:
Before, recommendations on mobile screens couldn't be sorted because they are using the kanban widget, which doesn't allow it.

Recommendation lines order is particularly important to improve salesperson efficiency, so a special feature is added here so that it can be done while using the module in mobile.

The previous default behavior is respected.

@moduon MT-4472

### DIFF
--- a/sale_order_product_recommendation/README.rst
+++ b/sale_order_product_recommendation/README.rst
@@ -88,6 +88,8 @@ To use this module, you need to:
 #. Create a new sale order.
 #. Assign its customer.
 #. Press *Recommended Products* button.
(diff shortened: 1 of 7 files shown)


rafaelbn (approved):
Tested in runbot 👍🏼 

Now you can order but category and by product name 😄 

🚀

Shide (approved):
Code LGTM

Gelojr (approved):
LGTM Thanks @yajo

EmilioPascual (approved):

PR #2918, opened 2024-01-23
[16.0][IMP]sale_exception: set context param for unlocking an order

dalonsofl:
Hi!

I have found the following problem.

Suppose there is an order that initially it does not have any exception or the exception is desactived and it is activated when the order has a 'done' state. If you try to unlock it, you will receive the warning message and you will not be able to unlock it.

### To Repoduce

1. Create an order

image

2. Activate the sale exception rule

image

3. Try to unlock the order

image

(images from runboat OCA)

With this change, it will be possible to unlock the order. It only adds the param "check_exception" to the context in order to avoid the warning message [here](https://github.com/OCA/server-tools/blob/d799fb940d0c52d0f4b7db30c1576a53cbeb150b/base_exception/models/base_exception.py#L100)

### DIFF
--- a/sale_exception/tests/test_sale_exception.py
+++ b/sale_exception/tests/test_sale_exception.py
@@ -204,3 +204,21 @@ class TestSaleException(TransactionCase):
         so_except_confirm.action_confirm()
         self.assertFalse(sale_order.ignore_exception)
         self.assertTrue(sale_order.state == "draft")
+
+    def test_exception_no_validation_for_unlock(self):
+        partner = self.env.ref("base.res_partner_1")
+        partner.zip = False
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        self.assertEqual(sale_order.state, "sale")
+        exception = self.env.ref("sale_exception.excep_no_zip")
+        exception.active = True
+        sale_order.action_done()
+        self.assertEqual(sale_order.state, "done")
+        # Just reverting the state to 'sale' would trigger an error.
+        with self.assertRaises(ValidationError):
+            sale_order.write({"state": "sale"})
+        sale_order.action_unlock()
+        self.assertEqual(sale_order.state, "sale")
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -65,3 +65,8 @@ class SaleOrder(models.Model):
     @api.model
     def _get_popup_action(self):
         return self.env.ref("sale_exception.action_sale_exception_confirm")
+
+    def action_unlock(self):
+        return super(
+            SaleOrder, self.with_context(check_exception=False)
+        ).action_unlock()


kikopeiro (approved):
LGTM :+1:

sale_exception/tests/test_sale_exception.py
```diff
@@ -204,3 +204,17 @@ def test_exception_no_free(self):
...
+        exception = self.env.ref("sale_exception.excep_no_stock")
+        exception.active = True
+        partner = self.env.ref("base.res_partner_1")
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        sale_order.action_done()
+        self.assertEqual(sale_order.state, "done")
+        sale_order.action_done()
```
AungKoKoLin1997: Why twice `action_done`()?

dalonsofl: Absolutely right! I just changed it, thanks!

AungKoKoLin1997 (approved):
LGTM (Code Review)

sale_exception/tests/test_sale_exception.py
```diff
@@ -204,3 +204,16 @@ def test_exception_no_free(self):
...
+        exception.active = True
+        partner = self.env.ref("base.res_partner_1")
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        sale_order.action_done()
+        self.assertEqual(sale_order.state, "done")
+        sale_order.action_unlock()
+        self.assertEqual(sale_order.state, "sale")
```
yostashiro: This test scenario doesn't look valid.  We may first process the SO up to done, then activate an exception rule that should raise an exception without the `check_exception=False` context.

yostashiro (changes requested):
Thanks for the fix!

sale_exception/tests/test_sale_exception.py
dalonsofl: Hi!

Thanks for your suggestion! I have change the scenario as you proposed.

Firstly, the SO is created and confirmed, after that, the exception rule is activated (It is supposed to be raised because the partner does not have any zip code). Secondly, the SO is locked and unlock, without any exception.

Regards

sale_exception/tests/test_sale_exception.py
```diff
@@ -204,3 +204,18 @@ def test_exception_no_free(self):
         so_except_confirm.action_confirm()
         self.assertFalse(sale_order.ignore_exception)
         self.assertTrue(sale_order.state == "draft")
+
+    def test_exception_no_stock_warning(self):
```
yostashiro:
```suggestion
    def test_exception_no_validation_for_unlock(self):
```
or whatever that describes the test scenario better.

sale_exception/tests/test_sale_exception.py
```diff
@@ -204,3 +204,18 @@ def test_exception_no_free(self):
...
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        self.assertEqual(sale_order.state, "sale")
+        exception = self.env.ref("sale_exception.excep_no_zip")
+        exception.active = True
+        sale_order.action_done()
+        self.assertEqual(sale_order.state, "done")
+        sale_order.action_unlock()
```
yostashiro, line 223: To make the context clearer.

```suggestion
        # Just reverting the state to 'sale' would trigger an error.
        with self.assertRaises(ValidationError):
            sale_order.write({"state": "sale"})
        # action_unlock(), which also writes the state to 'sale', should be free from validation.
        sale_order.action_unlock()
```

sale_exception/tests/test_sale_exception.py
dalonsofl: Yes! This name makes sense for this situation

yostashiro (approved):
Thanks for the update!

aliciagaarzo (approved):
LGTM!

sale_exception/tests/test_sale_exception.py
rousseldenis: @dalonsofl Could you comment the function as suggested?

Maybe:

# Avoid triggering sale exceptions when trying to unlock sale order

rousseldenis: @dalonsofl

dalonsofl: Hi @rousseldenis ! I have already added the context, thank you!

rousseldenis: Oops didn't see. It was above

PR #2346, opened 2023-01-26
[14.0][IMP] sale_order_revision: 'New Revision' quick button on confirmed Sales Orders

dreispt:

PR #2920, opened 2024-01-23
[14.0][FIX] sale_order_revision: wrong revision list if user has a fav filter

dreispt:
If the user created a default afvorite filter on the Sales Order list,
it will override the default filter set by the Revisons button.

So, the Sale Order menu Action should not be reused for this.


dreispt:
<s>NOTE: please prefer reviewing and merging #2930, that includes this, and already has the merge conflicts with  #2346 resolved.</s>

EDIT: please ignore

dreispt:
Ping!

StefanRijnhart:
Mmm, I feel that the implications of this are larger than just this particular case. Should this be a guideline somewhere? I would also appreciate a docstring to ensure this survives the next migration.

PR #2924, opened 2024-01-24
[14.0] [ADD] product_supplierinfo_for_customer_sale_blanket_order

renda-dev:
New module to apply supplier info for customers on blanket orders

Similar to [product_supplierinfo_for_customer_sale](https://github.com/OCA/sale-workflow/tree/14.0/product_supplierinfo_for_customer_sale)


product_supplierinfo_for_customer_sale_blanket_order/models/sale_blanket_order_line.py
```diff
@@ -0,0 +1,62 @@
...
+        if hasattr(super(), "onchange_product"):
+            super().onchange_product()
+        for line in self.filtered(
+            lambda sbol: (
+                sbol.product_id.product_tmpl_id.customer_ids
+                and sbol.order_id.pricelist_id.item_ids
+            )
+        ):
+            product = line.product_id
+            items = self.env["product.pricelist.item"].search(
```
rousseldenis, line 38: Hich, I'm not sure about performances of this. Couldn't you do the search outside of the loop?

Moreover, adding tests should be great

renda-dev: Thanks for the review!

I totally agree with you, but I'm not quite sure if that could be possible since the whole search is based on the iterating line.

Please, let me know if you have any other idea on how this could be optimized.

francesco-ooops (approved):
Functional ok!

francesco-ooops (approved):
Functional ok!

product_supplierinfo_for_customer_sale_blanket_order/models/sale_blanket_order_line.py
HekkiMelody: Only thing that comes to mind for improving performance is to not use `filtered` but to use `if (condition): continue` inside the loop, so you don't loop twice, but that's really a minimal gain, you're not going to have very many lines.

But also, in this case, in an onchange you're guaranteed to only have one record, so you don't even have to loop in the first place.

Aside from that, all other similar modules are implemented this way...

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@rousseldenis good to go? :)

francesco-ooops:
@rousseldenis gentle reminder :)

francesco-ooops:
@pedrobaeza can we merge based on existing reviews? thanks

PR #2927, opened 2024-01-26
[14.0][IMP]sale_blanket_order: refactor module making it similar to sale_order

PicchiSeba:
This refactor aims to make `sale.blanket.order` even more similar to the normal `sale.order`.

This includes changing the state from "expired" to "cancel" as it is called in `sale.order`s


dreispt (changes requested):
This breaks the module's API, at least the change to states.
It will can cause existing customizations on top of it to break.
These API breaking changes should happen in the latest version only.

PicchiSeba:
> This breaks the module's API, at least the change to states. It will can cause existing customizations on top of it to break. These API breaking changes should happen in the latest version only.

I completely forgot about that, thanks for pointing it out.

I just added the migration script, but I'm unsure about the module version.

@dreispt

sale_blanket_order/models/blanket_orders.py
```diff
@@ -264,9 +264,21 @@ def onchange_partner_id(self):
...
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if self.env.company.enable_numbered_bo and vals.get(
+                "name", _("Draft")
+            ) == _("Draft"):
+                vals["name"] = self.env["ir.sequence"].next_by_code(
+                    "sale.blanket.order"
+                ) or _("Draft")
```
renda-dev: **suggestion:** I'm not quite sure if "Draft" will need to be translated three times, but just to be safe, place `_("Draft")` inside of a variable and use that instead

sale_blanket_order/migrations/14.0.1.2.2/pre-migrate.py
```diff
@@ -0,0 +1,19 @@
...
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("Updating states for sale_blanket_order " + str(version))
+
+    cr.execute(
+        """
+        UPDATE sale_blanket_order sbo
+        SET state = 'cancel'
+        WHERE sbo.state = 'expired'
+        """
+    )
```
renda-dev: **nitpick(non-blocking):** I personally think that `openupgrade.logged_query()` fits better here :)

renda-dev (changes requested):
Just a couple of really small changes and it should be good to go

sale_blanket_order/models/blanket_orders.py
PicchiSeba: Nice catch!

renda-dev (approved):
Code Review, LGTM

francesco-ooops (approved):
Functional ok!

@dreispt good for you too?

sale_blanket_order/migrations/14.0.1.2.2/pre-migrate.py
```diff
@@ -16,4 +16,3 @@ def migrate(env, version):
         WHERE sbo.state = 'expired'
         """,
     )
-    env.cr.commit()
```
HekkiMelody: **issue**: This is a breaking change, it definitely deserves a major version --> `14.0.2.0.0`. Also, when you add a migration script, you need to also change the version in the `__manifest__.py` manually to make sure they are the same. (maybe you did, github isn't showing me all the changes in this PR for some reason)

sale_blanket_order/views/sale_blanket_order_views.xml
```diff
@@ -42,6 +42,11 @@
                         string="Cancel"
                         class="btn-secondary"
                         type="object"
+                        attrs="{'invisible': [
+                            '|',
+                            ('state', 'not in', ['draft', 'open']),
+                            ('id', '=', False)
```
HekkiMelody, line 47: **chore(non-blocking)**: isn't the second condition `('id', '=', False)` redundant? When a new SBO is created, it's in `draft` state. I think this condition could be simplified by only leaving the first tuple.

HekkiMelody (changes requested):
Partial code review. I don't think I can see all the changes I'm meant to see in this PR, from the github interface

sale_blanket_order/views/sale_blanket_order_views.xml
PicchiSeba: The point was to hide the Cancel button when creating a new record, because it doesn't exist yet so we cannot cancel it

HekkiMelody: For sure. But a new record is also in 'draft', so the button should already be hidden by the first condition, no?

Anyway, this is a non-blocking comment, so feel free to ignore it.

HekkiMelody:
Some changes, such as the change to the name of the SBO state, have definitely been lost.

sale_blanket_order/views/sale_blanket_order_views.xml
PicchiSeba: It was done this way to mimic as close as possible the `sale` module.

See how it was done in the form view for `sale.order`

https://github.com/odoo/odoo/blob/c77881763c11871dc061c47eefcca707b24e8a4e/addons/sale/views/sale_views.xml#L290

PicchiSeba:
> Some changes, such as the change to the name of the SBO state, have definitely been lost.

@aleuffre 

I recovered the missing pieces

sale_blanket_order/models/res_company.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    enable_numbered_bo = fields.Boolean("Enable numbered BO", default=False)
```
HekkiMelody: **issue**: the name of this setting leads me to believe that without it, BOs wouldn't be numbered. Instead, this setting only changes WHEN they are numbered. A clearer name and string would help avoid possible confusion IMO. `blanket_order_seq_number_from_draft`? You can probably pick something better

(Also, when you look at res_company there are settings from all over the place, so I'd avoid abbreviations that could be unclear when taken out of context, such as `bo`)

sale_blanket_order/models/blanket_orders.py
```diff
@@ -264,9 +264,24 @@ def onchange_partner_id(self):
...
+        for vals in vals_list:
+            if (
+                self.env.company.enable_numbered_bo
+                and vals.get("name", translated_draft) == translated_draft
+            ):
+                vals["name"] = (
+                    self.env["ir.sequence"].next_by_code("sale.blanket.order")
+                    or translated_draft
+                )
+        result = super().create(vals_list)
```
HekkiMelody: **issue**: `action_confirm` wasn't changed.

The behaviour right now, with the setting activated, is that:
- I create a BO --> It takes number BO0001
- I confirm it --> It takes number BO0002

sale_blanket_order/views/sale_blanket_order_views.xml
```diff
@@ -31,17 +31,22 @@
...
-                        states="expired,done"
+                        states="cancel,done"
                         string="To Draft"
                         class="btn-secondary"
                         type="object"
                     />
                     <button
                         name="action_cancel"
-                        states="open"
+                        states="draft,open"
```
HekkiMelody: **Chore**: Since you added the `attrs` below, you can remove `states`, which is now unused and overwritten by `attrs`.

HekkiMelody (changes requested):
Code review

sale_blanket_order/models/blanket_orders.py
```diff
@@ -298,11 +313,15 @@ def set_to_draft(self):
...
         self._validate()
         for order in self:
-            sequence_obj = self.env["ir.sequence"]
-            if order.company_id:
-                sequence_obj = sequence_obj.with_company(order.company_id.id)
-            name = sequence_obj.next_by_code("sale.blanket.order")
-            order.write({"confirmed": True, "name": name})
+            update_data = {"confirmed": True}
+            # if the order already had a sequence number we don't need to set it again
+            if not self.env.company.blanket_order_seq_number_from_draft:
```
HekkiMelody: **issue**: I believe the company here should not be `self.env.company` but `order.company_id` (and maybe `self.env.company` as a fallback, in case `order.company_id` is empty.

An alternative approach (that also sidesteps the `company` problem) could be to not check the setting at all, but instead replicate the check in the create: simply, if the field `name` is still empty/or it still says "Draft", give it a number. We for sure want it to have a number, we only need to make sure we're not overwriting an existing number.

sale_blanket_order/models/blanket_orders.py
```diff
@@ -264,9 +264,24 @@ def onchange_partner_id(self):
             values["team_id"] = self.partner_id.team_id.id
         self.update(values)
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        translated_draft = _("Draft")
+        for vals in vals_list:
+            if (
+                self.env.company.blanket_order_seq_number_from_draft
```
HekkiMelody: **issue**: this `self.env.company` should be `vals["company_id"]` (possibly with `self.env.company` as a fallback)

My active company could be company A, but I could still be creating an order for company B that has a different setting. The BO needs to follow the setting specified in `company_id` regardless of what `self.env.company` is.

HekkiMelody (changes requested):

sale_blanket_order/models/blanket_orders.py
```diff
@@ -264,9 +264,25 @@ def onchange_partner_id(self):
             values["team_id"] = self.partner_id.team_id.id
         self.update(values)
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        translated_draft = _("Draft")
+        for vals in vals_list:
+            company = vals.get("company_id") or self.env.company
```
HekkiMelody: **issue**: Have you tried this? I do not believe that this will work, since `vals.get("company_id")` returns an integer, whereas `self.env.company` is a record. The following should work, but please do check it.

```suggestion
            company_id = vals.get("company_id")
            company = self.env["res.company"].browse(company_id) if company_id else self.env.company
```

HekkiMelody (changes requested):

sale_blanket_order/models/blanket_orders.py
PicchiSeba: I haven't tried via the interface but I run the tests and they succeeded.

But yeah, looking again, done this way it shouldn't work

HekkiMelody: Confirmed on runboat, on creating a BO.

image

HekkiMelody (approved):
Code review, LGTM

sale_blanket_order/models/blanket_orders.py
```diff
@@ -98,7 +98,7 @@ def _compute_amount_all(self):
             ("draft", "Draft"),
             ("open", "Open"),
             ("done", "Done"),
-            ("expired", "Expired"),
```
rousseldenis, line 101: I understand the idea behind this change BUT:

I think, from a functional point of view, a blanket order has an expiration date because of its nature (of blanket order) and should have that 'expired' state. Because of that, blanket order should differ from sale order.

For the remaining changes, I'm ok

PicchiSeba: I updated the module with both states. If you see anything else that smells please let me know

renda-dev (approved):
Code Review, LGTM

francesco-ooops (approved):
Functional ok!

@dreispt good for you?

sale_blanket_order/views/sale_blanket_order_views.xml
```diff
@@ -117,7 +121,7 @@
                         <page string="Order Lines" name="order_lines">
                             <field
                                 name="line_ids"
-                                attrs="{'readonly': [('state', 'in', ('open','expired'))]}"
+                                attrs="{'readonly': [('state', 'in', ('open','cancel'))]}"
```
HekkiMelody: **issue**: if you leave both states ('cancel' and 'expired') then this `attrs` and other attributes and functions in the module need to be updated accordingly.

HekkiMelody (changes requested):
Existing code is no longer correct with the state change partially reverted

sale_blanket_order/models/blanket_orders.py
HekkiMelody: > I understand the idea behind this change BUT:
> 
> I think, from a functional point of view, a blanket order has an expiration date because of its nature (of blanket order) and should have that 'expired' state. Because of that, blanket order should differ from sale order.
> 
> For the remaining changes, I'm ok

The main idea driving this change is to make the module compatible with `base_revision`, which expects a state `cancel`. This is why the first draft of this PR changed `expire` to `cancel`.

What do you think about having the state be called `cancel` in the backend for compatibility, and "Canceled/Expired" for the User, in order to cover both use cases?

sale_blanket_order/wizard/create_sale_orders.py
```diff
@@ -20,9 +20,9 @@ def _default_order(self):
         blanket_order = self.env["sale.blanket.order"].search(
             [("id", "=", self.env.context["active_id"])], limit=1
         )
-        if blanket_order.state == "expired":
+        if blanket_order.state == "cancel":
```
HekkiMelody: **issue**: 

```suggestion
        if blanket_order.state in ("expired", "cancel"):
```

sale_blanket_order/wizard/create_sale_orders.py
```diff
@@ -20,9 +20,9 @@ def _default_order(self):
...
             raise UserError(
-                _("You can't create a sale order from " "an expired blanket order!")
+                _("You can't create a sale order from " "a cancelled blanket order!")
```
HekkiMelody:
```suggestion
                _("You can't create a sale order from a cancelled or expired blanket order!")
```

HekkiMelody (changes requested):

HekkiMelody (approved):
Code review, LGTM.

Structurally, I'm not sure about keeping both "cancel" and "expired" states, but :shrug:

francesco-ooops (approved):
Functional ok!

Now "cancelled" is an independent state, much more similar to the standard implementation in SO

PR #2930, opened 2024-01-27
[14.0][IMP] sale_order_revision: carry Invoice links to new revision Sales Orders

dreispt:
- [IMP] sale_order_revision: 'New Revision' quick button on confirmed Sales Orders
- [IMP] sale_order_revision: carry Invoice links to new revision Sales Orders

Includes:
- [ ] #2346  
- [ ] #2920


IvanHRN (approved):

sale_order_revision/models/sale_order.py
```diff
@@ -34,11 +34,37 @@ def _prepare_revision_data(self, new_revision):
...
+            new_line.invoice_lines = old_line.invoice_lines
+        return action
+
+    def action_cancel_create_revision(self):
+        for sale in self:
+            sale.action_cancel()
+            action = sale.create_revision()
+        if len(self) == 1:
+            return action
+        return {}
```
victoralmau, line 70:
```suggestion
    def action_cancel_create_revision(self):
        self.ensure_one()
        sale.action_cancel()
        return sale.create_revision()
```

If you can only call from the form view for a record, I think it could be simplified like this.

victoralmau (changes requested):

pedrobaeza (changes requested):
Revisions shouldn't be done in confirmed sales orders, as that impacts in a lot of workflows. If done, it should be in a separate module, but not forced to be done in the existing one.

dreispt:
@pedrobaeza I understand.
I will make the button work for Quotation state only, and move the confirmed SO support to a separate module.

I wish I didn't have to do revisions os confirmed SOs, but Portuguese Tax Authority requires SOs to be locked once sent to customers. I'm trying to make abiding to this less painful for PT companies.

pedrobaeza:
> I wish I didn't have to do revisions os confirmed SOs, but Portuguese Tax Authority requires SOs to be locked once sent to customers. I'm trying to make abiding to this less painful for PT companies.

That sound a bit weird, Anyway, sent to customers don't mean confirmed. You have the `sent` state in the sales order, so I think it's better to forbid the changes in that state that changing all of this, don't you think?

dreispt:
> That sound a bit weird, Anyway, sent to customers don't mean confirmed. You have the `sent` state in the sales order, so I think it's better to forbid the changes in that state that changing all of this, don't you think?

The forbidding changes is not the issue, that's not hard to enforce.
The issue is when, after sent and confirmed, we need to make changes to the order, such as correcting ordered quantities, prices, or even applicable taxes. Then we need to do that in a copy of the original order.

pedrobaeza:
OK, but the flow in that case is to cancel the sales order, and then issue a new revision. What you can't do is to try to link things that are generated when you confirm the sales order to the new revision. That's something out of scope of any module IMO.

dreispt:
Additional work was on on this feature, implementing it in a separate module.
This PR is now superseded by https://github.com/OCA/sale-workflow/pull/3275

PR #2931, opened 2024-01-27
[ADD] sale_order_revision_stock: carry Delivery links to new revision Sales Orders

dreispt:
Depends (includes):

- [ ] #2930


rousseldenis:
@dreispt My question is : Is it normal to do revisions on sale orders that have already stock moves (and so, has already been confirmed) ?

dreispt:
@rousseldenis Good question.
The use case is caused by Portuguese Tax Authorities, that require Quotations or Sale orders sent to customers to be inalterable, signed with a cryptographic hash, and be electronically reported to them.
So if I send ou a Sales order to a customer, and later I need to make changes to it , such adding a line, or correcting something, I have to do it in a copy of the original order.

Today you need to create a copy of the SO, with all the disadvantages of doing that :-(
This is an attempt for a better solution.

dreispt:
PS: I could move this to l10n-portugal in case it is a bit absurd anywhere else...

rousseldenis:
@dreispt I would say this should be configurable (configuration parameter on company)

dreispt:
Superseded by https://github.com/OCA/sale-workflow/pull/3275

PR #2934, opened 2024-01-30
[14.0][FIX]product_supplierinfo_for_customer_sale: set explicit quantity=None

PicchiSeba:
explicitly assign `quantity=None`

requires https://github.com/OCA/product-attribute/pull/1506

### DIFF
--- a/product_supplierinfo_for_customer_sale/models/sale_order_line.py
+++ b/product_supplierinfo_for_customer_sale/models/sale_order_line.py
@@ -19,7 +19,8 @@ class SaleOrderLine(models.Model):
         for line in self:
             if line.product_id:
                 supplierinfo = line.product_id._select_customerinfo(
-                    partner=line.order_partner_id
+                    partner=line.order_partner_id,
+                    quantity=None,
                 )
                 code = supplierinfo.product_code
             else:
@@ -56,7 +57,7 @@ class SaleOrderLine(models.Model):
             )
             if items:
                 supplierinfo = line.product_id._select_customerinfo(
-                    partner=line.order_partner_id
+                    partner=line.order_partner_id, quantity=None
                 )
                 if supplierinfo and supplierinfo.min_qty:
                     line.product_uom_qty = supplierinfo.min_qty


product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
```diff
@@ -79,7 +79,10 @@ def _create_pricelist_item(self, name, pricelist, product):
 
     def test_product_supplierinfo_for_customer_sale(self):
         so = self.env["sale.order"].create(
-            {"partner_id": self.customer.id, "pricelist_id": self.pricelist.id}
+            {
+                "partner_id": self.customer.id,
+                "pricelist_id": self.pricelist.id,
+            }
```
HekkiMelody: **chore**: Please remove this diff

product_supplierinfo_for_customer_sale/models/sale_order_line.py
```diff
@@ -52,11 +53,11 @@ def product_id_change(self):
                     "&",
                     ("product_id", "=", product.id),
                     ("applied_on", "=", "0_product_variant"),
-                ]
+                ],
```
HekkiMelody: **chore**: please remove this diff

HekkiMelody (approved):
Code review ok

minor code cleanup.

francesco-ooops:
@pedrobaeza good for you? :)

pedrobaeza:
I think this should depend also on `product_uom_qty` and pass such quantity to the method.

HekkiMelody:
> I think this should depend also on `product_uom_qty` and pass such quantity to the method.

@pedrobaeza We tried this, it was also our first line of thinking, but the module (and the tests) explicitly assume that you should be able to get a customer code even if you don't respect the minimum quantity. 

For example, in the tests it creates supplierinfo with minimum quantity set to 15, and then expects to get the customer code even if the quantity is set to 1.

https://github.com/OCA/sale-workflow/blob/14.0/product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py#L47-L57

I'm unsure if that's how it's supposed to work, so for now we restored the original functionality (I haven't had time to look deeply into how it works for the original supplierinfo in base Odoo). If you think we should change it we can adapt the tests.

pedrobaeza:
OK, let's continue for now as it is:

/ocabot merge patch

PR #2938, opened 2024-01-31
[17.0][MIG] sale_start_end_dates

alexis-via:
Migration from v16 to v17. Code cleanup in tests: stop using demo data, use setupClass(cls)


sale_start_end_dates/tests/test_sale_start_end_dates.py
```diff
@@ -8,40 +8,47 @@
...
 class TestSaleStartEndDates(TransactionCase):
-    def setUp(self):
-        super().setUp()
-        self.partner = self.env.ref("base.res_partner_3")
-        self.product_id = self.env.ref(
-            "account_invoice_start_end_dates.product_insurance_contract_demo"
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis, line 14: @alexis-via FYI, best is to use BaseCommon class instead : https://github.com/odoo/odoo/blob/c11e5fea5dd5a357c17b12c0c5f383909fa44bb8/odoo/addons/base/tests/common.py#L21

rousseldenis (approved):
Code review

sale_start_end_dates/tests/test_sale_start_end_dates.py
alexis-via: I don't think so. BaseCommon adds stuff for the tests of the base module. Nobody uses this class for tests except the product module. I didn't find any OCA module using it... do you know one ?

StefanRijnhart (approved):

PR #2939, opened 2024-02-01
[16.0][FIX] sale_packaging_default: Keep packaging order in SO lines

p-tombez:
When multiple packaging are set on a product, during the SO creation, the default package is sometime not the first one in the list :

1. Have a product with 2 packagings
2. Drag and drop the first packaging to be the second to change its sequence
3. Create a SO and add a line with the product

Before : package A is selected instead of package B
After : package B is selected

### DIFF
--- a/sale_packaging_default/models/sale_order_line.py
+++ b/sale_packaging_default/models/sale_order_line.py
@@ -27,7 +27,7 @@ class SaleOrderLine(models.Model):
                 line.product_packaging_id = (
                     line.product_id.packaging_ids.filtered_domain(
                         [("sales", "=", True)]
-                    )[:1]
+                    ).sorted("sequence")[:1]
                 )
         result = super()._compute_product_packaging_id()
         # If there's no way to package the desired qty, remove the packaging.
--- a/sale_packaging_default/tests/test_sale_packaging_default.py
+++ b/sale_packaging_default/tests/test_sale_packaging_default.py
@@ -75,6 +75,20 @@ class SalePackagingDefaultCase(ProductCommon):
             self.assertEqual(line_f.product_packaging_qty, 0)
             self.assertEqual(line_f.product_uom_qty, 0)
 
+    def test_default_packaging_order_on_sale_order(self):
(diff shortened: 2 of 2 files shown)


santostelmo (approved):

sale_packaging_default/models/sale_order_line.py
```diff
@@ -27,7 +27,7 @@ def _compute_product_packaging_id(self):
                 line.product_packaging_id = (
                     line.product_id.packaging_ids.filtered_domain(
                         [("sales", "=", True)]
-                    )[:1]
+                    ).sorted("sequence")[:1]
```
yajo, line 30: You know what's weird? The test is correctly written, and the bug is supposed to not exist, but if I remove this change, the test still goes green.

That's what I'd expect because [the default order](https://github.com/odoo/odoo/blob/4adda36260f3ba0fdf75688b419593006b80adb7/addons/product/models/product_packaging.py#L15) is supposed to do automatically what you expect here.

Do you have any other addons installed that alter that default order, or something similar?

yajo (commented):
In general I agree that, if this was happening, it's a bug. However, I can't reproduce. See below please.

Thanks for your contribution!

p-tombez:
@yajo I thought I tested the "issue" with a bare Odoo, but I found out later that, indeed, there is another module changing the default order : https://github.com/OCA/product-attribute/blob/16.0/product_packaging_level/models/product_packaging.py#L11
I'm currently not sure how to handle my use cases with these 2 modules installed together, therefore this PR may be irrelevant

yajo:
If a user drags the packagings and sorts them in a certain order, why does that module ignore that order? Conceptually speaking it seems like a bug in `product_packaging_level`. If it even makes sense somehow, then that module could alter the views where it needs that model to be sorted differently, instead of changing the default order for the model.

IMHO if that's the module that breaks the behavior, it's probably a better fit for a fix.

Let me close this PR for now while you search for a solution there. If something blocks you from progressing in that direction, then just reopen and we'll see what we can do here. 😉 

Thanks!

PR #2941, opened 2024-02-04
[FIX] sale_order_price_recalculation: Remove button using action_update_prices

ajaniszewska-dev:
Action not implemented in the code

### DIFF
--- a/sale_order_price_recalculation/views/sale_order_view.xml
+++ b/sale_order_price_recalculation/views/sale_order_view.xml
@@ -7,14 +7,6 @@
         <field name="arch" type="xml">
             <field name="state" position="before">
                 <button
-                    name="action_update_prices"
-                    class="ml-2 btn btn-default"
-                    string="Recalculate prices"
-                    Help="Apply the current pricelist"
-                    type="object"
-                    attrs="{'invisible':[('state','not in', ['draft', 'sent'])]}"
-                />
-                <button
                     name="action_update_names"
                     class="ml-2 btn btn-default"
                     string="Reset descriptions"


sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -6,14 +6,6 @@
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
             <field name="state" position="before">
-                <button
-                    name="action_update_prices"
```
ajaniszewska-dev, line 10: afaik this is already implemented in sale in v16, not sure why it was kept

rousseldenis:
@CRogos @arleite What do you think of this ?

rousseldenis:
What's the difference with :

https://github.com/odoo/odoo/blob/a21cbb54121d97d7b10a48beb5c1fad4d79bfe01/addons/sale/views/sale_order_views.xml#L270C43-L270C63

@ajaniszewska-dev Following this conversation, this module is still needed : https://github.com/OCA/sale-workflow/pull/2778#issuecomment-1856531619

rousseldenis:
Indeed, the function call is the same as in core.

The difference I see is the visibility condition

CRogos:
The button is visible always, while the odoo 16 refresh button only shows when the pricelist has changed.
If we remove the button, we should change the visibility of the native refresh button either to be always visible, or if at least one SO-line price value is not according to the pricelist.

rousseldenis:
> The button is visible always, while the odoo 16 refresh button only shows when the pricelist has changed. If we remove the button, we should change the visibility of the native refresh button either to be always visible, or if at least one SO-line price value is not according to the pricelist.

@CRogos What I see here is the fact original button can be visible at the same time as this one. I would say this should take into account the inverse of 'show_update_pricelist' field too.

PR #2943, opened 2024-02-06
[15.0][FIX] sale_invoice_plan: add permission menu invoice plan

Saran440:
This PR fixed user is not sale can visible menu sale because `sale invoice plan` don't have permission.
They shouldn't see this menu, if not sale.

Selection_002


rousseldenis (approved):
Code review. Seems fair

kittiu (approved):

PR #2946, opened 2024-02-07
[FIX] sale_product_multi_add: fix discount

majouda:
Previously, when the 'Discount Policy' on the Price list was configured to 'Show public price & discount to customer,' the inclusion of multiple products in the sales order line did not take the price list into account. Now, this functionality is compatible with the specified price list configuration.

### DIFF
--- a/sale_product_multi_add/wizards/sale_import_products.py
+++ b/sale_product_multi_add/wizards/sale_import_products.py
@@ -46,7 +46,10 @@ class SaleImportProducts(models.TransientModel):
             }
         )
         line_values = sale_line._convert_to_write(sale_line._cache)
-        line_values = sol_obj.play_onchanges(line_values, [])
+        line_values = sol_obj.play_onchanges(
+            line_values,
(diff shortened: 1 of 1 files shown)


sale_product_multi_add/wizards/sale_import_products.py
```diff
@@ -46,7 +46,10 @@ def _get_line_values(self, sale, item):
             }
         )
         line_values = sale_line._convert_to_write(sale_line._cache)
-        line_values = sol_obj.play_onchanges(line_values, [])
+        line_values = sol_obj.play_onchanges(
+            line_values,
+            ["product_id", "price_unit", "product_uom", "product_uom_qty", "tax_id"],
```
rousseldenis, line 51: @majouda Thanks for this.

In fact I think the good syntax is to use `line_values.keys()` on the second argument of play_onchanges() to be compatible with side modules.

Could you maybe add a test case for the one you describe ?

Thanks

rousseldenis:
Closing this as very old.

PR #2947, opened 2024-02-07
[14.0][FIX] sale_mrp_bom: accept BoMs without variant

HekkiMelody:
This commit fixes a bug where the module would not accept a BoM on a Sale Order Line if the product in question had multiple variants and the BoM did not specify a variant.

Now, regardless of whether a product has multiple variants or not,  if the variant is specified on the BoM, it must match the product on the Sale Order Line (like before), but if it's not specified, matching on the product template is enough.

### DIFF
--- a/sale_mrp_bom/models/sale_order_line.py
+++ b/sale_mrp_bom/models/sale_order_line.py
@@ -19,16 +19,18 @@ class SaleOrderLine(models.Model):
     @api.constrains("bom_id", "product_id")
     def _check_match_product_variant_ids(self):
         for line in self:
-            if line.bom_id:
-                bom_product_tmpl = line.bom_id.product_tmpl_id
-                bom_product = bom_product_tmpl.product_variant_ids
-            else:
-                bom_product_tmpl, bom_product = None, None
-            line_product = line.product_id
-            if not bom_product or line_product == bom_product:
+            if not line.bom_id:
+                continue
+            bom_product = line.bom_id.product_id
+            bom_product_tmpl = line.bom_id.product_tmpl_id
+
+            if bom_product and bom_product == line.product_id:
+                continue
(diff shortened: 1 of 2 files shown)


francesco-ooops (approved):
Functional ok!

sale_mrp_bom/models/sale_order_line.py
```diff
@@ -19,16 +19,20 @@ class SaleOrderLine(models.Model):
...
-                bom_product_tmpl, bom_product = None, None
-            line_product = line.product_id
-            if not bom_product or line_product == bom_product:
+            if not line.bom_id:
                 continue
+            bom_product = line.bom_id.product_id
+            bom_product_tmpl = line.bom_id.product_tmpl_id
+
+            if bom_product:
+                if bom_product == line.product_id:
```
aayartsev: i think this expression will be better: `bom_product and bom_product == line.product_id:`

sale_mrp_bom/models/sale_order_line.py
```diff
@@ -19,16 +19,20 @@ class SaleOrderLine(models.Model):
...
+                    continue
+            else:
```
aayartsev: i think you can delete this `else`

aayartsev (changes requested):
@aleuffre Check please my suggestions.

HekkiMelody:
thank you for your review @aayartsev , I simplified the code according to your suggestions.

aayartsev (approved):
Looks good.)

francesco-ooops:
@pedrobaeza is this good to go? Thanks!

geomer198 (approved):
LGTM!

francesco-ooops:
@pedrobaeza can you relaunch merge command?

PR #2951, opened 2024-02-13
[14.0][ADD] sale_discount_display_amount: add discount_subtotal

matthieusaison:
Add fields discount_subtotal  (discount without tax)

### DIFF
--- a/sale_discount_display_amount/migrations/14.0.2.2.0/post-compute-subtotal-columns.py
+++ b/sale_discount_display_amount/migrations/14.0.2.2.0/post-compute-subtotal-columns.py
@@ -1,0 +1,21 @@
+import logging
+
+from odoo import SUPERUSER_ID
+from odoo.api import Environment
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Compute discount columns")
+    env = Environment(cr, SUPERUSER_ID, {})
+
+    query = """
+    select distinct order_id from sale_order_line where discount > 0.0;
+    """
+
+    cr.execute(query)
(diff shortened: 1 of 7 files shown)


sergiocorato:
Thanks for the PR @matthieusaison 

Could you please add the new field in `hooks.py` [here](https://github.com/OCA/sale-workflow/blob/39045c8e3d017753f54812d37b879e8dafe64794/sale_discount_display_amount/hooks.py#L14) and [here](https://github.com/OCA/sale-workflow/blob/39045c8e3d017753f54812d37b879e8dafe64794/sale_discount_display_amount/hooks.py#L17)? So at the new installation this will be pre-computed as the others.

If you can bump the module version and add the migration script with the same version too, for the new field pre-computation in existing installation.

BTW, I made a PR: https://github.com/akretion/sale-workflow/pull/31

sebastienbeau:
@sergiocorato thanks for your contribution !

luisDIXMIT (approved):
LGTM!

sale_discount_display_amount/migrations/14.0.2.2.0/post-compute-subtotal-columns.py
```diff
@@ -0,0 +1,21 @@
...
+
+    query = """
+    select distinct order_id from sale_order_line where discount > 0.0;
+    """
+
+    cr.execute(query)
+    order_ids = cr.fetchall()
+
+    orders = env["sale.order"].search([("id", "in", order_ids)])
+    orders.mapped("order_line")._update_discount_display_fields()
```
rousseldenis, line 21: IMHO, this should be done through SQL as it is much more efficient than python.

PR #2954, opened 2024-02-15
[IMP]sale_quoatation_number: New sequence after change in state

manuelregidor:
In the current version of the module, the change in the sequence, from the quotation sequence to the confirmed order sequence, is performed before the actual change of the state in the sale order.

With this improvement, the change in the sequence is dones after the change of the state in the sale order. The aim is to avoid a write action (the one to change the sequence and other attributes related to it) if the change of the state fails or cannot be completed for any reason.

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -48,16 +48,19 @@ class SaleOrder(models.Model):
         self.ensure_one()
         return self.env["ir.sequence"].next_by_code("sale.order")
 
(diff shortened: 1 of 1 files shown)


manuelregidor:
@HaraldPanten @ValentinoDefelice

manuelregidor:

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):
LGTM

PR #2955, opened 2024-02-15
[FIX] sale_order_product_recommendation: make UoM work

yajo:
Before this patch, UoM wasn't being properly inherited into generated lines when those lines didn't exist previously in the SO:

image

Also, changing it manually didn't work as expected in that same case.

This is because the UoM was related to the SOL UoM. But if there was no SOL, it did nothing.

Now it's properly computed. UoM comes from the product if there's no SOL or it can't provide it. The user can change it to other UoM of the same category. The change is propagated when applying the wizard, both when creating lines and updating them. Tests added.

@moduon MT-4472

### DIFF
--- a/sale_order_product_recommendation/README.rst
+++ b/sale_order_product_recommendation/README.rst
@@ -128,6 +128,12 @@ Contributors
 
   * Telmo Santos
 
+* `Moduon <https://www.moduon.team>`_:
+
+  * Rafael Blasco
+  * Gelo Joga
+  * Jairo Llopis
+
 Maintainers
 ~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


fcvalgar (approved):
Functional review,

LGTM thank you @yajo 

image

image

Shide (approved):
Code and Functional review

EmilioPascual (approved):

Gelojr (approved):
LGTM Thanks @yajo https://www.loom.com/share/d92457038ada460cb3b7623e2cef8916?sid=df8340c6-601a-4019-be9f-0213a6a7722b

rafaelbn (approved):
/ocabot merge patch

PR #2959, opened 2024-02-19
[16.0] sale_packaging_default: make packaging lookup reusable

simahawk:
Add _get_default_packaging. Allows replicating the same logic when needed.


yajo (approved):
Thanks!

/ocabot merge minor

PR #2961, opened 2024-02-19
[17.0] [MIG] sale_force_invoiced: migration to 17.0

siemenv:


sale_force_invoiced/model/sale_order.py
```diff
@@ -21,6 +20,6 @@ class SaleOrder(models.Model):
     def _compute_invoice_status(self):
         res = super()._compute_invoice_status()
         self.filtered(
-            lambda so: so.force_invoiced and so.state in ("sale", "done")
+            lambda so: so.force_invoiced and (so.state == "sale" or so.locked)
```
StefanRijnhart: In principle, only orders in `sale` status can be locked. I think you can simplify the module by ignoring the `locked` field everywhere and just go by state `sale`. Note that orders in any other state will not be marked as invoiceable by Odoo: https://github.com/odoo/odoo/blob/3f6deca/addons/sale/models/sale_order.py#L530-L531 so even if orders were to be locked in any other status, this override might be considered meaningless for such orders. Or am i missing anything?

rousseldenis:
@JordiBForgeFlow @MiquelRForgeFlow @StefanRijnhart @siemenv 

I would say this module is quite dangerous and something has to be put in README. If a sale order is put as force invoiced and a line is added that generates a delivery order, the sale order won't never been invoiced for that outgoing product.

StefanRijnhart:
Maybe a roadmap item can be added for a feature to ensure that only locked orders can be forced invoiced?

@siemenv thanks for the update. Due to squashing, your commit message now has two lines:

```
 [MIG] sale_force_invoiced: ignore locked state

[MIG] sale_force_invoiced: Migration to 17.0
```

Only the first line is visible in the commit history here on Github. It must say `[MIG] sale_force_invoiced: Migration to 17.0`.

Can you fix the commit message?

StefanRijnhart (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #2962, opened 2024-02-19
[17.0] [MIG] sale_order_line_date: migration to 17.0

siemenv:


sale_order_line_date/models/sale_order_line.py
```diff
@@ -0,0 +1,43 @@
...
+            )
+        return vals
+
+    def write(self, vals):
+        res = super().write(vals)
+        moves_to_upd = set()
+        if "commitment_date" in vals:
+            for move in self.move_ids:
+                if move.state not in ["cancel", "done"]:
+                    moves_to_upd.add(move.id)
```
celm1990: `write` work over multi records, please avoid singleton error if write is done with multi records
```suggestion
            for line in self:
              for move in line.move_ids:
                  if move.state not in ["cancel", "done"]:
                      moves_to_upd.add(move.id)
```

celm1990 (changes requested):

sale_order_line_date/models/sale_order_line.py
yannoliv: Both self and move_ids are recordsets, so I think this code is viable as self.move_ids will just get all move_ids of all records in self. @celm1990

bizzappdev:
Thank you for the contribution and fix.
Functional test okay.
Code LGTM.
Only the commit message should follow the OCA standard 
$ git commit -m "[MIG] $module: Migration to 17.0"
Please make the necessary changes.

salvorapi:
LGTM

celm1990 (approved):
LGTM

sale_order_line_date/models/sale_order_line.py
BialArno: Was wondering, why iterate when you can map/filter instead?

```python
        if 'commitment_date' in values and (moves_to_update := self.mapped('move_ids').filtered(lambda move: move.state not in ['cancel', 'done'])):
            moves_to_update.date_deadline = values.get('commitment_date')
```

rousseldenis: @BialArno In fact, mapped.filtered do the same as the current code. I was using Odoo provided methods before but now I'd prefer this writing as clearer.

rousseldenis (approved):
Code review

rousseldenis:
@celm1990 @yannoliv Do you want to update your review ?

bizzappdev (approved):
LGTM Code review and functional testing both

sale_order_line_date/models/sale_order_line.py
BialArno: @rousseldenis well yes, they do the same, more or less it's pick of personal preferrence, but then we should consider implementing framework methods over sticking to python basics or go plain python and at that point, in my opinion, in Odoo these methods weren't added for nothing :) Ofc competition of speed might be useful to check if for loop or mapped/filtered does same task faster and/or more resource efficient.

celm1990 (approved):
LGTM

DorianMAG (approved):
the code looks good
Can't test in runboat to make a functional review

salvorapi (approved):

Deriman-Alonso (approved):
LGTM 👍, code review, Thanks!

traitrantrobz (approved):
The test file and functional is LGTM 👍
Please rebase the commit.
Ảnh chụp Màn hình 2024-07-22 lúc 11 40 18

HaraldPanten:
@siemenv Could you "clean" the commit history a little bit?

THX

PieterPaulussen (approved):

BT-etejeda (approved):
Code and functional review!
@siemenv could you please squash the administrative commits?
If you don't have time to do it, let us now, please. Thanks.

StefanRijnhart:
I do apologize but with 9 approvals I think we can live with a handful of redundant administrative commits.

/ocabot merge nobump

PR #2969, opened 2024-02-22
[16.0][IMP] sale_stock_picking_blocking: allow lock state on sale order

trisdoan:
### Steps to reproduce:

1. Activate "Lock Confirmed Sales"
2. Create a new sale order, add a product and configure a Delivery Block Reason
3. Save the sale order / confirm, you get this error message : `You cannot block a sale order with "auto_done_setting" active`.

### This change

- Remove `_check_not_auto_done` to allow feature sale order lock.

### DIFF
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -2,8 +2,7 @@
 #   (http://www.eficent.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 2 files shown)


trisdoan:
Hi @ , I created this PR because I don't understand why this check was added and it prevents the module from working in a legitimate use case, do you remember why this module was considered as incompatible with the Sale Lock feature option

trisdoan:
Hi @MiquelRForgeFlow  , I created this PR because I don't understand why this check was added and it prevents the module from working in a legitimate use case

Do you remember why this module was considered as incompatible with the Sale Lock feature option ?

AaronHForgeFlow:
I think the problem was the SO was automatically locked when confirmed, so the delivery was blocked forever. Don't remember the exact version

trisdoan:
Hi @AaronHForgeFlow, thanks for your reply. However, I believe there might be some confusion or a workaround that I've missed.

In my understanding, there is a button available to unlock the sale order, which subsequently releases the delivery. Could you please provide more context or clarify the scenario where this button doesn't resolve the blocking issue?

trisdoan:
Hi @AaronHForgeFlow, the `_check_not_auto_done` should be removed in odoo 16, IMO

User can unlock the SO and `release delivery block` easily

AaronHForgeFlow:
@trisdoan  It is a bit strange the user has to unlock the delivery block in order to create the stock transfer, but ok.

yvaucher (approved):

AaronHForgeFlow (approved):

M-tripleorange:
Hi, I am trying to use this module in Odoo V17 and i get the same issue. How can I fix this?

AaronHForgeFlow:
@M-tripleorange FW port available here: https://github.com/OCA/sale-workflow/pull/3437 please review.

PR #2970, opened 2024-02-22
[16.0][IMP] sale_auto_remove_zero_quantity_lines: exclude lines that don't have a product_id

sonhd91:
In the action that removes zero qty lines, exclude lines that don't have a product set on them like notes and sections


santostelmo (approved):

rousseldenis (approved):
Code review. Of course!

Issue #2813, opened 2023-11-27
[V16.0] sale_discount_display_amount: Wrong discount total calculation on discount included amounts


PieterPaulussen:
Source: https://github.com/OCA/sale-workflow/blob/50f5707635f93a8b5868e20bbee1726c4063bb9c/sale_discount_display_amount/models/sale_order_line.py#L40C46-L40C60

The implementation of this seems off. Are discounts not always calculated on the amount excluding VAT?

Atteqany:
You are right.  wrong calculation.
Odoo makes the % discount before adding the TAX, but this module calculate the %discount on the amount which includes the TAX

For example: Sale order for a product with a price 100 and 10% discount, Odoo wrote the subtotal 90 
So the value of the 10% discount is only 10 because all this are before applying the TAX
but this module made the 10% discount value is 11.50, which means it calculated wrong.
Now, the discount amount which the module display is more than the actual which Odoo counted.

PR #2972, opened 2024-02-22
[17.0][MIG] sale discount display amount: Migration to 17.0

FSchiepers:
This version adds the possibility to choose whether the discount is calculated based on the non-taxed subtotal or the taxed total, see: https://github.com/OCA/sale-workflow/issues/2813


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

Check also pre-commit stuff.

PR #2974, opened 2024-02-22
[FIX] sale_order_general_discount: fix compute_discount

jdoutreloux:
Set general discount as line discount only if there's a general discount.
No need to redefine discount fields, it's already a computed field.

### DIFF
--- a/sale_order_general_discount/models/sale_order_line.py
+++ b/sale_order_general_discount/models/sale_order_line.py
@@ -1,16 +1,10 @@
 # Copyright 2018 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import api, fields, models
+from odoo import api, models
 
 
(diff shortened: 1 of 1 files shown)


sale_order_general_discount/models/sale_order_line.py
```diff
@@ -26,8 +20,8 @@ def create(self, vals_list):
 
     @api.depends("order_id", "order_id.general_discount")
     def _compute_discount(self):
-        if hasattr(super(), "_compute_discount"):
-            super()._compute_discount()
+        res = super()._compute_discount()
         for line in self:
-            line.discount = line.order_id.general_discount
-        return
+            if line.order_id.general_discount:
```
rousseldenis: @jdoutreloux So, if a discount of 0 is set on order level, lines should not be affected ?

rousseldenis (approved):

ThomasBinsfeld (approved):

PR #2979, opened 2024-02-26
[15.0][FIX] sale_planner_calendar: Avoid infinite loop recursion when recurrence end is forever

CarlosRoca13:
cc @Tecnativa TT48016


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2981, opened 2024-02-26
[FIX] sale_advance_payment field amount_residual

sistecem:


sistecem:
There is a small error in the current version,  when canceling a paid Invoice , It still sums up in the invoice_paid_amount , resulting in wron amount_residual in sale order

rousseldenis (changes requested):
@sistecem Thanks for this.

Could you change your commit message to make module name appear ? ([FIX] sale_advance_payment ...).

Moreover, adding a test should be great in order to avoid regressions

sistecem:
I am going to do the Test in another PR

PR #2983, opened 2024-02-27
[15.0][FIX] sale_custom_rounding: Do not call super with empty recordset

manuelregidor:
Before this fix, super method could be called with an empty recordset, which can cause errors depending on the order that methods execute.

After this fix, super methods are called only when some of the records meet certain criteria to avoid calling methods with empty recordsets.

It also fixes the imports and names of tests.

### DIFF
--- a/sale_custom_rounding/models/sale_order.py
+++ b/sale_custom_rounding/models/sale_order.py
@@ -34,21 +34,31 @@ class SaleOrder(models.Model):
 
     @api.depends("tax_calculation_rounding_method")
     def _compute_tax_totals_json(self):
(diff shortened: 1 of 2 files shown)


ValentinVinagre (changes requested):
add tests please

manuelregidor:
@ValentinVinagre It's a very technical feature and I don't think that adding additional tests would make any difference. If you have any idea, please, let me know and I will add them.

ValentinVinagre (approved):

HaraldPanten (approved):

PR #2988, opened 2024-03-02
[16.0][MIG] sale_invoice_plan, standard migration to 16.0

kittiu:
Standard migration

PR #2989, opened 2024-03-03
[16.0][ADD] sale_invoice_plan_batch

kittiu:
Dependencies:
- sale_invoice_plan https://github.com/OCA/sale-workflow/pull/2988

This module add new way to create invoice from sales invoice plan in batch
rather than having to open each sales order to create invoice by plan.

A new `Sales Invoice Plan` Batch document allow user to filter all matched installments
with many criteria, i.e., plan date from, plan date to, customers, sales orders.

Then, all installment can be created as invoices in one button click.


kittiu:
Replaced with #2990

PR #2995, opened 2024-03-06
[16.0][FIX] product_supplierinfo_for_customer_sale: handle no supplierinfo

alexey-pelykh:


alexey-pelykh:
@AaronHForgeFlow @simahawk related to https://github.com/OCA/product-attribute/commit/03d244c9e172d2fff93c3c5796c7cb3083c29f34

AaronHForgeFlow (approved):
LGTM!

simahawk:
Wouldn't be better to restore the empty recordset as you mentioned on the prev comment?

alexey-pelykh:
@simahawk imho we should do both so that empty string is explicitly used for empty record set.

simahawk:
Just do `supplierinfo.product_code or ""` if you care about the type although is not so relevant because Odoo sets empty fields as False anyway.

alexey-pelykh:
Reasonable, then let's merge the other PR and close this one

AaronHForgeFlow (changes requested):
as per comments

PR #2996, opened 2024-03-06
[17.0][MIG] sale_restricted_qty

alexey-pelykh:

PR #2997, opened 2024-03-07
[17.0][IMP] sale_restricted_qty: UX

alexey-pelykh:
Depends on #2996


alexey-pelykh:
@rousseldenis @rafaelbn prior to diving too deep, there are two ways I can tackle this:

A. Migrate to v17 and then refractor the hell out if this module.
B. Refactor the hell out of it in v14 and then migrate.

Either way, I want to get some feedback on the UX refactor itself first :)

rafaelbn:
Hello @alexey-pelykh ,

I've never used this module 😢 

Maybe the opinion of original author from Akretion are more valuable

@Amourad-ehm @victoralmau @sbiosca-s73 @SirAionTech @rvalyi 

Best regards! 😄 ❤️

bosd:
Just unlocked :unlock: a feature:
Have a SO which does not adhere to a multiple of 50 (e.g. 101 pce)
Go your product. Set the multiple to 50 and enforce restrict multiple.

No you will get an error message, unable to save the product.

alexey-pelykh:
Yep, by design - you can have either a warning or an error

bosd:
https://github.com/user-attachments/assets/8f23e0eb-7ca0-4591-a71c-ad231b7219ea
> Yep, by design - you can have either a warning or an error

aah. so this is intended behaviour?
Or someting you want to change?

My use case, client has a messy system. (rescue project)
Want to enforce a multiple, while there already are SO's which do not have a proper multiple.

alexey-pelykh:
@bost you have Restrict selected as "yes" - this means it won't allow to save if validation fails, regardless if there was mistake before or no.

BostX:
No, I don't. But @bosd has 😜

alexey-pelykh:
Ahah, sorry, was typing from the phone :)

bosd:
Hmm.. Looks like I;m still not getting this ux. Or I'm confused.
Can you please help me.
So, with this module, it is not possible to enforce a multiple quantity if you have historical orders?

What is the use case for checking the historical so's, when making changes to the product form?

Setting a restrict to no:
image
still allows me to set this on a SO.
image
But I've filled in a multiple. What is the use of setting a multiple, without using it?
The earlier version of this module mentioned in the description end of product life or selling old stock. But then, one could just easily remove the value from this field.

alexey-pelykh:
> So, with this module, it is not possible to enforce a multiple quantity if you have historical orders?

It is possible, but you won't be able to save-after-edit those historical orders that violate the rule. That's the "enforcement" part. Historical orders mean "no change"

alexey-pelykh:
Restrict = No means that you'll get only the yellow warning highlight of the bad line

bosd:
> > So, with this module, it is not possible to enforce a multiple quantity if you have historical orders?
> 
> It is possible, but you won't be able to save-after-edit those historical orders that violate the rule. That's the "enforcement" part. Historical orders mean "no change"

aah. are we talking about the same item? Having an historical order, which violate the rule now block saving of the product template. The error message shown is not very clear about that. And as end user it is quite confusing.
I would vote, to always be able to edit save your product. regardless what happened in the past.

A strong use case of this module, would be as described, MOQ went wrong in the past. They install, this module to prevent it from happening in the future.

> Restrict = No means that you'll get only the yellow warning highlight of the bad line

Aaha, clear... That would be a valuable addition to the docs..

alexey-pelykh:
Rebased on latest 17.0 (the dependency #2996 is now merged), undrafted, and ready for review.

This PR improves the UX of `sale_restricted_qty`:
- Refactors the mixin for cleaner inheritance (category → template → variant)
- Adds per-entity checkbox overrides instead of force flags
- Shows inline qty constraint indicators on the sale order form
- Splits the monolithic test into per-model test files

@bosd — you reviewed the migration (#2996), would you mind taking a look at this UX follow-up?
@rousseldenis — would appreciate your review when you have a moment.

Thanks!

sale_restricted_qty/models/product_restricted_qty_mixin.py
```diff
@@ -4,120 +4,549 @@
 
 from odoo import api, fields, models
 
+RESTRICTION_ENABLED = "1"
+RESTRICTION_DISABLED = "0"
+RESTRICTION_SELECTION = [
+    (RESTRICTION_ENABLED, "Yes"),
+    (RESTRICTION_DISABLED, "No"),
+]
```
marielejeune, line 12: Would be better to have real chars as selection values, it eases the code comprehension:

I would suggest something like `RESTRICTION_ENABLED = "enabled"` or `RESTRICTION_ENABLED = "restricted"` and `RESTRICTION_DISABLED = "disabled"` or `RESTRICTION_DISABLED = "unrestricted"`

marielejeune (changes requested):

sale_restricted_qty/models/product_restricted_qty_mixin.py
rousseldenis: @alexey-pelykh @bosd Could we finish this before going to 18.0 ?

rousseldenis: @alexey-pelykh

rousseldenis:

PR #3001, opened 2024-03-08
[16.0][ADD] sale_spread_cost_revenue

kittiu:
This module works exactly the same as account_spread_cost_revenue.
Except that it allow create spread revenue from sales order line
and if the spread is created from sales order, it will be linked
automatically with the invoice created by this sales order.


rousseldenis:
@kittiu What's the status of this?

kittiu:
The code is good. But I only use it on one of my project. Not sure others will find it useful.

PR #3004, opened 2024-03-11
[FIX] sale_order_product_recommendation: product category field string

yajo:
Product category was labeled as "Complete name" due to a typo in field definition, which didn't make any sense.

@moduon MT-5404


yajo:
Obvious fast-tracking candidate:

/ocabot merge patch

PR #3006, opened 2024-03-11
[17.0][MIG] sale_shipping_info_helper

AndreuOForgeFlow:
Standard migration

@ForgeFlow


sale_shipping_info_helper/models/sale_order.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    item_amount_total = fields.Float(
+        compute="_compute_shipping", digits="Account", store=True
+    )
+    item_amount_untaxed = fields.Float(
+        compute="_compute_shipping", digits="Account", store=True
+    )
+    item_amount_tax = fields.Float(
+        compute="_compute_shipping", digits="Account", store=True
+    )
```
GuillemCForgeFlow: As a general improvement to the module, perhaps it could be great to add some help descriptions to the fields to better understand which is the purpose for each one. What do you think? 😄

GuillemCForgeFlow (commented):
The rest looks fine 👍🏽

GuillemCForgeFlow (changes requested):
Perfect, thank you for the descriptions! I see that the version bump got done in the previous commit: https://github.com/OCA/sale-workflow/pull/3006/commits/76ad6655a72d6c78b7905afa620a763a01d336c5#diff-d0de6d5d2f3650f12b7d8c0526b108685a25da2d8d0c8913174cc7dd0c4e0ecaR8. It should be done in the migration one!

sale_shipping_info_helper/models/sale_order.py
rousseldenis: +1

sale_shipping_info_helper/models/sale_order.py
```diff
@@ -0,0 +1,68 @@
...
+        store=True,
+        help="Tax portion of the item amount",
+    )
+
+    @api.depends("amount_total", "amount_untaxed")
+    def _compute_shipping(self):
+        for record in self:
+            shipping_amount_untaxed = shipping_amount_total = shipping_amount_tax = 0
+            for line in record.order_line:
+                if not line.is_delivery:
```
rousseldenis: I think the best is to use '_is_delivery()' function instead.

rousseldenis (commented):
Minor change

GuillemCForgeFlow (approved):
LGTM! 👍🏽

rousseldenis (approved):
Code review

PR #3009, opened 2024-03-12
[16.0][IMP] sale_product_set: Move the base logic of the transient model to product_set

pilarvargas-tecnativa:
Pending from:

- [x] https://github.com/OCA/product-attribute/pull/1548

The reason to move this logic is that there are other modules that extend
product_set as for example stock_product_set but at the same time make use
of the transient model to define a wizard. For this reason it is better to
have the logic available in product_set module and avoid duplicating code or
inheriting from sale_product_set and what this implies in their respective
dependencies with the only need to extend the transient model.
In addition, the transient model is renamed to make it clearer to identify
that it is this type of model.

cc @Tecnativa TT48100

@pedrobaeza @victoralmau please review

### DIFF
--- a/sale_product_set/README.rst
+++ b/sale_product_set/README.rst
@@ -91,6 +91,9 @@ Contributors
 * Adria Gil Sorribes <adria.gil@forgeflow.com>
 * Phuc (Tran Thanh) <phuc@trobz.com>
 * Manuel Regidor <manuel.regidor@sygel.es>
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Pilar Vargas
 
 Other credits
 ~~~~~~~~~~~~~
--- a/sale_product_set/__manifest__.py
+++ b/sale_product_set/__manifest__.py
@@ -13,7 +13,7 @@
         "security/ir.model.access.csv",
         "views/product_set.xml",
         "views/product_set_line.xml",
-        "wizard/product_set_add.xml",
+        "wizard/sale_product_set_wizard_view.xml",
         "views/sale_order.xml",
     ],
     "demo": ["demo/product_set_line.xml"],
--- a/sale_product_set/readme/CONTRIBUTORS.rst
+++ b/sale_product_set/readme/CONTRIBUTORS.rst
@@ -6,3 +6,6 @@
 * Adria Gil Sorribes <adria.gil@forgeflow.com>
 * Phuc (Tran Thanh) <phuc@trobz.com>
 * Manuel Regidor <manuel.regidor@sygel.es>
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Pilar Vargas
--- a/sale_product_set/security/ir.model.access.csv
+++ b/sale_product_set/security/ir.model.access.csv
@@ -1,4 +1,4 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 sale_product_set_manager,Read-Write-Create access on product.set to Sale Manager,product_set.model_product_set,sales_team.group_sale_manager,1,1,1,1
 sale_product_set_line_manager,Read-Write-Create access on product.set.line to Sale Manager,product_set.model_product_set_line,sales_team.group_sale_manager,1,1,1,1
-product_set_add_wizard_user,Read-Write-Create-Delete access on product.set.add to users,model_product_set_add,base.group_user,1,1,1,1
+sale_product_set.access_sale_product_set_wizard,access_sale_product_set_wizard,sale_product_set.model_sale_product_set_wizard,base.group_user,1,1,1,1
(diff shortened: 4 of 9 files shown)


sale_product_set/wizard/product_set_add.py
```diff
@@ -6,8 +6,8 @@
 
 
 class ProductSetAdd(models.TransientModel):
```
petrus-v:
```suggestion
class SaleProductSetWizard(models.TransientModel):
```

sale_product_set/wizard/product_set_add.py
```diff
@@ -6,8 +6,8 @@
```
petrus-v, line 6: Could you rename this file `sale_product_set_wizard.py` ?

pilarvargas-tecnativa: changes done, thanks!

pedrobaeza (approved):
/ocabot merge major

lmignon:
@pedrobaeza  How is it possible to merge this breaking change into a stable version!!!!!!! Only one approve and it's merged. This must be reverted.

pedrobaeza:
This is a forward-port, and it's not breaking change. What is broken in your side?

lmignon:
the model `product.set.add` has been removed. Therefore it's a breaking change.

lmignon:
@pedrobaeza Where is the original PR since you say it's a foward port.

pedrobaeza:
Sorry, not forward-port, but chained PR: https://github.com/OCA/product-attribute/pull/1548. Without this merge, half of the suite won't work.

That model is a transient one, and you are mentioning a change (the same as renaming a method or similar, that may impact somehow), not something broken. If there's any OCA module affected, please tell us and we change it, as we have made with all the known affected ones like this PR.

lmignon:
I fixed the impacted addons into shopinvader and into our customer project. It is not courteous to merge this kind of change without ensuring a backward compatibility on a stable branch. In addition, the OCA merge policy has not been respected. Only 1 reviewer, less than 2 days after the PR creation and no review by initial contributors ... It looks more like a forceful passage to avoid any comment.

simahawk:
I would add that there's no good reason to remove the model completely, you could just deprecate it.
You could add a new model extending the new one and make the old one log deprecation for a smooth transition.
Please, keep this approach in mind next time :pray:

pedrobaeza:
The original PR has been fully supported by 4 reviewers: https://github.com/OCA/product-attribute/pull/1548, including one of your company, Laurent, so this one must be merged for completing the chain and not having incompatible modules inside OCA.

You have done similar things in the past and we required to do changes in modules outside OCA without us complaining. It's just how the evolution inside the same version works.

And this is not a regular model, just a transient. There's no way to keep backwards compatibility in this. And if so, please propose how, because I don't know how to do it. If you keep the old model, then what? No error on update, but no visibility of it when clicking the button? And the old model was shared between all the modules in all the areas, doing dirty things like putting a field like not required. Clearly a bad initial design, but it became from a restricted scope that was later expanded. And now the base is OK. And it must be done here in v16, where the `product_set` was split and where we, as contributors, are investing time/money in improving the landscape.

We are one of the companies that assure the smoothest transitions, including migration scripts across versions or inside them, and care about all the affected modules inside OCA, so please don't accuse us for something that others make worst than us. And specially you, @simahawk, can't speak a word, where we have suffered from your continuous changes in OCA/edi, provoking crashes and problems even inside OCA modules.

simahawk:
> The original PR has been fully supported by 4 reviewers: https://github.com/OCA/product-attribute/pull/1548, including one of your company, Laurent, so this one must be merged for completing the chain and not having incompatible modules inside OCA.

That's true.

> And this is not a regular model, just a transient. There's no way to keep backwards compatibility in this. 

This is not true and you know it better than me. The fact that is a transient model does not mean is not important.
And you know that such changes are normally introduced in migrations.
Now, if you tell me any motivation I buy it but not this one, sorry.

> And specially you, @simahawk, can't speak a word, where we have suffered from your continuous changes in OCA/edi, provoking crashes and problems even inside OCA modules.

1st of all: You are taking this too personal.

2nd: not fair at all because before introducing changes I always ask @etobella if there's anything I can do to not break existing implementations in ES localization.  Here's one of the examples https://github.com/OCA/edi/pull/797#pullrequestreview-1585641057

If it happened, it was definitely because of an error and not for a careless approach. AND you know very well that if you ask me for help for something that I break (or even someone else breaks) I'm always available for help and I always say "sorry" first w/o pointing the finger at others.

We are all humans and we fail. We all broke something at some point.
So.... my friend, just cool down and accept the critics. No bad feelings.

With love,
S

pedrobaeza:
Simone, as said, there's no way to avoid it except letting things as they were, which is not an option for decoupling correctly the wizard. It's not a big deal, detected on CI part, so let's continue and accept this as part of the game.

And the example you are putting in the link happened after 2 changes that broke l10n-spain and then Enric told you about that. The first breakage on OCA/edi impacted a customer of mine not sending the Facturae files until it was detected, as any CI flows weren't aware. I was the same annoyed, but accepted it having yours and @etobella's support. I offered the same to Laurent: if something is broken, tell me and I'll fix it.

PR #3011, opened 2024-03-13
[16.0][IMP] sale_product_packaging_container_deposit: Avoid forbidden modifying locked SO when updating order container deposit quantity

QuocDuong1306:
Depend on:
- https://github.com/OCA/product-attribute/pull/1545

Rationale:
- `product_packaging_container_deposit` needs to update SOL
- But when a Sale Order is locked, some fields are protected (hence the error: "It is forbidden to modify the following fields in a locked order: Quantity")
- Given that this is a modification intended by the system and not by the user, we had a context key + an override to allow it.

### DIFF
--- a/sale_product_packaging_container_deposit/models/sale_order_line.py
+++ b/sale_product_packaging_container_deposit/models/sale_order_line.py
@@ -13,3 +13,9 @@ class SaleOrderLine(models.Model):
 
     def _get_product_qty_delivered_received_field(self):
         return "qty_delivered"
+
+    def _get_protected_fields(self):
+        protected_fields = super()._get_protected_fields()
+        if self.env.context.get("update_order_container_deposit_quantity", False):
+            protected_fields.remove("product_uom_qty")
+        return protected_fields
--- a/sale_product_packaging_container_deposit/tests/test_sale_product_packaging_container_deposit.py
+++ b/sale_product_packaging_container_deposit/tests/test_sale_product_packaging_container_deposit.py
@@ -1,5 +1,6 @@
 # Copyright 2023 Camptocamp (<https://www.camptocamp.com>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.exceptions import UserError
 from odoo.tests import Form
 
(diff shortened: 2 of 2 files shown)


sale_product_packaging_container_deposit/models/sale_order_line.py
```diff
@@ -13,3 +13,9 @@ def _get_product_qty_field(self):
 
     def _get_product_qty_delivered_received_field(self):
         return "qty_delivered"
+
+    def _get_protected_fields(self):
+        protected_fields = super()._get_protected_fields()
+        if self.env.context.get("update_order_container_deposit_quantity", False):
+            protected_fields.remove("product_uom_qty")
```
santostelmo, line 20: Could you please add a unit test showing that we are able to deliver products with container deposit when the SO is locked ?

sale_product_packaging_container_deposit/models/sale_order_line.py
```diff
@@ -13,3 +13,9 @@ def _get_product_qty_field(self):
 
     def _get_product_qty_delivered_received_field(self):
         return "qty_delivered"
+
+    def _get_protected_fields(self):
+        protected_fields = super()._get_protected_fields()
+        if self.env.context.get("update_order_container_deposit_quantity", False):
```
simahawk, line 19: pls see my comment on https://github.com/OCA/product-attribute/pull/1545

QuocDuong1306:
Could you have a look, i updated

simahawk (approved):

simahawk:
dependency merged

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo-addon-product_packaging_container_deposit @ git+https://github.com/OCA/product-attribute.git@refs/pull/1545/head#subdirectory=setup/product_packaging_container_deposit
```
santostelmo: Dependency was merged

QuocDuong1306:
Thank @simahawk , @santostelmo , I updated

cyrilmanuel (approved):

PR #3017, opened 2024-03-15
[15.0][IMP] sale_planner_calendar: Do not create attendees is no mail option selected

sergio-teruel:
cc @Tecnativa 


carlosdauden (approved):
Code LGTM

pedrobaeza:
If the salesman is not added as attendee, its calendar shows the event?

sergio-teruel:
Yes... We do not fill the attendees (calendar.attendee) but the event's attendants (partner_ids) is filled, so the event is shown on calendar view

pedrobaeza (approved):
/ocabot merge minor

PR #3019, opened 2024-03-15
[REF] sale_order_type: compute of pricelist and payment_terms

matiasperalta1:
The error is when the payment term is set in the partner as well as pricelist.
The case of replication by creating a quotation and introducing a partner with the payment term and pricelist set in its configuration. Then we manually change the payment term and pricelist to another. And finally we change the type of sale. Here we observe that the aforementioned fields are recomputed to those configured in the partner and this is an error. With this PR the error no longer persists since all cases are considered.

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -96,7 +96,6 @@ class SaleOrder(models.Model):
                 order.picking_policy = order_type.picking_policy
         return res
 
-    @api.depends("type_id")
     def _compute_payment_term_id(self):
         res = super()._compute_payment_term_id()
         for order in self.filtered("type_id"):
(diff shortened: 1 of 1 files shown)


rousseldenis:
Same comment as https://github.com/OCA/sale-workflow/pull/3046#issuecomment-2056508201

rousseldenis:
@matiasperalta1

alexey-pelykh (changes requested):
Thanks for working on this @matiasperalta1. The problem you describe (manual changes to payment term / pricelist getting overwritten when changing the sale type) is real and worth fixing, but I think the approach here introduces regressions.

Two main concerns:

1. **Removing `@api.depends("type_id")` from the compute methods breaks server-side propagation.** The compute body still contains the `type_id` logic, but without the decorator the compute will no longer re-trigger when `type_id` is changed programmatically (XML-RPC, server actions, automated workflows, tests). Only the UI `onchange` path would work. This is a functional regression for any non-UI caller.

2. **`@api.onchange` should be avoided in Odoo 16+.** The stored computed writable field pattern (`compute=..., store=True, readonly=False`) is the standard approach in 16.0 and is what this module already uses. Introducing onchanges goes against that pattern and was specifically flagged by @rousseldenis on the sibling PR #3046 — "We should avoid onchanges."

The underlying issue is about compute dependency ordering: when `type_id` changes, the base compute (which reads from `partner_id`) runs and overwrites the type-based value. The fix should stay within the compute pattern — for example, by adjusting the conditional logic so that if `type_id` has a value set, it takes priority over the partner default, without removing the dependency.

A test case covering this scenario would also be needed, as previously requested.

rousseldenis:

PR #3020, opened 2024-03-15
[14.0][IMP] sale_delivery_state: improve xpath in view inheritance

alexis-via:
The current xpath used in the inherit of the form view of sale.order could match a tree view embedded in the form view of sale.order.line. This commit fixes this.


matthieusaison (approved):

rvalyi (approved):

alexis-via:
No need to forward-port, because the field skip_sale_delivery_state is not present in v16/v17, so the xpath is not present in v16/v17.

PR #3022, opened 2024-03-16
[16.0][MIG] sale_order_secondary_unit

alexey-pelykh:


.copier-answers.yml
```diff
@@ -17,6 +17,7 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - sale_packaging_default,sale_order_product_recommendation,sale_order_product_recommendation_packaging_default,sale_order_product_recommendation_elaboration
+- sale_order_qty_change_no_recompute
```
alexey-pelykh: @pedrobaeza @victoralmau was there a specific reason why this module was not marked as rebel?

alexey-pelykh: https://github.com/OCA/sale-workflow/pull/3023

sale_order_secondary_unit/views/sale_order_views.xml
```diff
@@ -12,20 +12,25 @@
...
-                    optional="show"
-                />
-                <field
-                    name="secondary_uom_id"
-                    class="oe_inline"
-                    domain="[('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
-                    attrs="{'readonly': [('state', 'in', ('done', 'cancel'))]}"
-                    optional="show"
-                />
-                <field name="secondary_uom_unit_price" readonly="1" optional="show" />
```
alexey-pelykh: @sergio-teruel Why the field was deleted from the view yet kept in the model?

alexey-pelykh:
@pedrobaeza see, all tests are good :)

ivs-cetmix (approved):
Functional LGTM

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -11,52 +12,44 @@ class SaleOrderLine(models.Model):
         "uom_field": "product_uom",
     }
 
-    secondary_uom_unit_price = fields.Float(
+    secondary_uom_price_unit = fields.Float(
```
pedrobaeza, line 15: Please add a migration script renaming this field, or let it as it was.

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -11,52 +12,44 @@ class SaleOrderLine(models.Model):
...
-    @api.depends("secondary_uom_qty", "secondary_uom_id", "product_uom_qty")
+    @api.depends("secondary_uom_qty", "secondary_uom_id")
     def _compute_product_uom_qty(self):
+        res = super()._compute_product_uom_qty()
         self._compute_helper_target_field_qty()
+        return res
 
     @api.onchange("product_uom")
-    def onchange_product_uom_for_secondary(self):
+    def _onchange_product_uom(self):
```
pedrobaeza, line 30: If everything is now computed writable fields, you should convert this as well. If not, you don't know which one is going to act first (the compute or the onchange).

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -11,52 +12,44 @@ class SaleOrderLine(models.Model):
...
-            if line_uom_qty == 1.0:
-                self.secondary_uom_qty = 1.0
-                self.onchange_product_uom_for_secondary()
-            else:
-                self.product_uom_qty = line_uom_qty
-        return res
-
-    @api.depends("secondary_uom_qty", "product_uom_qty", "price_unit")
-    def _compute_secondary_uom_unit_price(self):
+    def _onchange_product_id(self):
```
pedrobaeza, line 34: The same about the onchange.

pedrobaeza (dismissed):

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: What's the point if it was computed and not stored? And now is computed and stored?

pedrobaeza: Was it not stored? And why is not stored now? I don't think the store attribute should change then.

alexey-pelykh: It was not stored, it is stored now. Truth be told - the field is not used anywhere since removal of that one from the UI - we've talked about that not that long ago.

pedrobaeza: Then don't do it stored if it's even not used in UI.

alexey-pelykh: Now that I know that that removal was not rationalized enough, I'm thinking if I should revert that as well

pedrobaeza: OK, I didn't know exactly the history, but I think this is not needed at all.

rousseldenis:
@alexey-pelykh Do you plan to attend comments ?

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: https://github.com/OCA/product-attribute/blob/16.0/product_secondary_unit/models/product_secondary_unit_mixin.py#L23-L25

I'm not sure that's an option - write does not replace onchange in any way

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: This one is needed to ensure smooth UX

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: Well, I guess since it's not in the UI since earlier version, I'll just drop the field entirely

alexey-pelykh: Or expose it in the UI..

alexey-pelykh: Yeah, I'll expose it in the UI. I don't see why migration would be needed, if that's a computed-stored field. Also, it's stored and precomputed alike price_unit

alexey-pelykh:
@rousseldenis yep, yet there's some common ground to be found

sale_order_secondary_unit/models/sale_order.py
rousseldenis: @alexey-pelykh What @pedrobaeza said is that in v16 (and even before), onchanges have been moved to computed stored fields functions. These methods trigger onchange behaviors as well (if depends is well defined of course).

sale_order_secondary_unit/models/sale_order.py
rousseldenis: Could you show the UX problem you have with compute method ?

alexey-pelykh: To ensure that we're keeping the quantity in primary UoM unchanged if there's a secondary UoM on a new product and recompute the quantity in secondary UoM

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: I'm not sure we're on the same page. It's literally implemented as product_secondary_unit in v16 asks for - and it works as a result.

As a side note - if it's computed, so what if there's an action to be performed on onchange?

rousseldenis: The current implementation of product_secondary_unit is not correct. As the code in the compute method is duplicated in onchange...

I think the migration should have merged both in a compute method.

alexey-pelykh: Well, yet with current merged implementation of product_secondary_unit I don't see other way for this to work. Yet as far as I can see, you've approved that in https://github.com/OCA/product-attribute/pull/1325 ? I take it your position has changed since then?

rousseldenis: We cannot see all the changes effects at each time :sweat_smile: 

Nevertheless, my question is still open. Why is the computed method does not work ?

alexey-pelykh: I don't have answer to your question, neither I don't think it's relevant - it works now and it didn't work in another migration. In the end my only interest is having a properly working module, rather then discovering why when following directions provided by product_secondary_unit module it actually works 🤷‍♂️

pilarvargas-tecnativa:
Please include https://github.com/OCA/sale-workflow/pull/3199

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: @rousseldenis do you have any does that answer your question?

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: @rousseldenis do you have any does that answer your question?

sale_order_secondary_unit/models/sale_order.py
alexey-pelykh: > If everything is now computed writable fields, you should convert this as well. If not, you don't know which one is going to act first (the compute or the onchange).

@pedrobaeza I'm not sure that's the case. The onchange and compute are not replacement and are called at different time. Compute is called on save and onchange can be called on an unsaved record with NewId - at least to my maybe outdated knowledge.

alexey-pelykh:
@pilarvargas-tecnativa thanks, included

sale_order_secondary_unit/models/sale_order.py
pedrobaeza: That's not really true. The computation of the computed fields is delayed until they are needed, before saving or not, being finally computed on write if they haven't been yet computed (but only if they are not stored). @api.onchange methods are only executed when doing things via UI, and they are executed immediately, possible being less performant in general. That's why is convenient to move everything to computed writable when possible.

alexey-pelykh: > That's why is convenient to move everything to computed writable when possible.

Which is exactly the opposite of what we want here - the conversion between primary and secondary UoM must happen on UI changes, not on save.

pedrobaeza: That's not what happen. If you have any field in screen that is recomputed in the computed writable methods, it's executed immediately. Please don't walk in circles and convert it to computed writable. Then, you tell me the problems you find.

alexey-pelykh: I don't walk in circles. There's a working solution implemented according to the https://github.com/OCA/product-attribute/blob/16.0/product_secondary_unit/models/product_secondary_unit_mixin.py#L9-L28 and at the moment you're requesting me to explore an approach that goes against those guidelines. Once those guidelines are changed I'd be happy to factor those changes in. Until that time I see no reason allocating time towards that, sorry.

pedrobaeza: OK

odooNextev:
@pedrobaeza Are there any updates? Could you merge it, or does it need some changes?
Thanks

pedrobaeza:
I have dismissed my review, as @sergio-teruel should give his final blessing.

sergio-teruel:
@alexey-pelykh Can you add d3d859ed27339c5560a2653aecf43e84d429ae71?

I will Review after done.. thanks

santostelmo:
@alexey-pelykh do you have a moment to cherry pick https://github.com/OCA/sale-workflow/commit/d3d859ed27339c5560a2653aecf43e84d429ae71 mentioned by @sergio-teruel

alexey-pelykh:
@santostelmo @sergio-teruel done

macagua:
@pedrobaeza @rousseldenis @sergio-teruel can you check this pull request? and if all is ok merge it, please, I need this improvement

sale_order_secondary_unit/views/sale_order_views.xml
sergio-teruel: I think is not needed add to view but if someone added the field I do not remove from model data

sergio-teruel:
ping @pilarvargas-tecnativa

PR #3023, opened 2024-03-16
[16.0][IMP] sale_order_qty_change_no_recompute: mark as rebel

alexey-pelykh:


.copier-answers.yml
```diff
@@ -1,6 +1,6 @@
 # Do NOT update manually; changes here will be overwritten by Copier
 _commit: v1.20
-_src_path: gh:oca/oca-addons-repo-template
+_src_path: git+https://github.com/oca/oca-addons-repo-template
```
pedrobaeza: Don't change this.

alexey-pelykh: Not my personal change, `pipx run copier copy git+https://github.com/oca/oca-addons-repo-template . --trust` did that

alexey-pelykh: I mean, I'll revert yet what did I do wrong?

pedrobaeza: You have to run it with `pipx run copier copy gh:oca/oca-addons-repo-template . --trust`

alexey-pelykh: Done

pedrobaeza (commented):
If there's no other option... Anyway, such incompatibility will arise if you install both modules (`sale_product_secondary_unit` and this one). In fact, we have installations with both installed. Which is the incompatibility?

alexey-pelykh:
I'm uncertain if this one will actually be needed as I'm still connecting the dots with `sale_product_secondary_unit` :)

Yet even w/o that - such unconditional alteration of default behavior in theory is a rebel one. Yet I'm not sure if that will be _required_

pedrobaeza:
I think it's better to keep them together while no unreconciliable problem is found.

alexey-pelykh:
Actually, now I do have at least some insight - `sale_product_secondary_unit` tests as they are now can't be true unless the side effects of `sale_order_qty_change_no_recompute` are there, namely `_compute_price_unit` being triggered, resetting the value.

pedrobaeza:
Then a redesign of the test?

alexey-pelykh:
In this very specific case I concur as the test case is wrong. However, it was _allowed_ to be wrong specifically by the unconditional behavior change. Have you considered also remaking the no_compute module by overriding `_compute_price_unit` somehow?

pedrobaeza:
I remember trying several techniques and that one was the only acceptable, but I don't remember details.

alexey-pelykh:
👍 So, that said - I still think module should be a rebel one, for reasons that if affects other modules unconditionally and is a behavioral breaking change. I wish there was a way to test other modules with and without this module automagically - to ensure they are compatible both ways - that'd be ideal

pedrobaeza:
Yes, but not being able, if we don't find any cause for not isolating it, although risky, we should keep them together for finding incompatibilities to arise.

alexey-pelykh:
So about the cause:

Here where `sale_order_qty_change_no_recompute` fails: https://github.com/OCA/sale-workflow/actions/runs/8308848147/job/22739527563?pr=3022#step:8:381

It appears that:

1. `_compute_product_uom_qty` gets called
2. `_compute_helper_target_field_qty` sets `product_uom_qty` to 1.0 as `default_qty_field_value`

Relevant logs:
```
2024-03-17 06:46:38,118 520 INFO odoo odoo.addons.sale_order_qty_change_no_recompute.tests.test_sale_order_qty_change: Starting TestSaleOrderQtyChange.test_sale_line_misc ... 
2024-03-17 06:46:38,142 520 WARNING odoo odoo.addons.sale_order_secondary_unit.models.sale_order: SaleOrderLine._compute_product_uom_qty (1): product_uom_qty = 2.0; self._origin['product_uom_qty'] = 0.0 secondary_uom_qty = 0.0 
2024-03-17 06:46:38,145 520 WARNING odoo odoo.addons.sale_order_secondary_unit.models.sale_order: SaleOrderLine._compute_product_uom_qty (2): product_uom_qty = 2.0; self._origin['product_uom_qty'] = 0.0 secondary_uom_qty = 0.0 
2024-03-17 06:46:38,145 520 WARNING odoo odoo.addons.sale_order_secondary_unit.models.sale_order: SaleOrderLine._compute_product_uom_qty (3): product_uom_qty = 1.0; self._origin['product_uom_qty'] = 0.0 secondary_uom_qty = 0.0 
2024-03-17 06:46:38,146 520 WARNING odoo odoo.addons.sale_order_secondary_unit.models.sale_order: SaleOrderLine._onchange_product_id (1): product_uom_qty = 1.0 
2024-03-17 06:46:38,164 520 INFO odoo odoo.addons.sale_order_qty_change_no_recompute.tests.test_sale_order_qty_change: ====================================================================== 
2024-03-17 06:46:38,164 520 ERROR odoo odoo.addons.sale_order_qty_change_no_recompute.tests.test_sale_order_qty_change: FAIL: TestSaleOrderQtyChange.test_sale_line_misc
  File "/__w/sale-workflow/sale-workflow/sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py", line 37, in test_sale_line_misc
    self.assertEqual(self.line_form.price_subtotal, 60)
AssertionError: 30.0 != 60
```

It happens because when `product_id` is assigned, we eventually get here I suppose:

```
@api.depends("secondary_uom_qty", "secondary_uom_id")
    def _compute_product_uom_qty(self):
        res = super()._compute_product_uom_qty()
        self._compute_helper_target_field_qty()
        return res
```

And relevant piece of code:
```
    @api.depends("secondary_uom_qty", "secondary_uom_id")
    def _compute_product_uom_qty(self):
        logger.warning(f"SaleOrderLine._compute_product_uom_qty (1): product_uom_qty = {self.product_uom_qty}; self._origin['product_uom_qty'] = {self._origin['product_uom_qty']} secondary_uom_qty = {self.secondary_uom_qty}")
        res = super()._compute_product_uom_qty()
        logger.warning(f"SaleOrderLine._compute_product_uom_qty (2): product_uom_qty = {self.product_uom_qty}; self._origin['product_uom_qty'] = {self._origin['product_uom_qty']} secondary_uom_qty = {self.secondary_uom_qty}")
        self._compute_helper_target_field_qty()
        logger.warning(f"SaleOrderLine._compute_product_uom_qty (3): product_uom_qty = {self.product_uom_qty}; self._origin['product_uom_qty'] = {self._origin['product_uom_qty']} secondary_uom_qty = {self.secondary_uom_qty}")
        return res
```

So it appears that this code in `product_secondary_unit`:

https://github.com/OCA/product-attribute/blob/0bc9700847546d52af97ba964f79f9805090fbe6/product_secondary_unit/models/product_secondary_unit_mixin.py#L100-L105

should've used the "current computed value", or I'm missing something:

```
rec[rec._secondary_unit_fields["qty_field"]] = (
    rec._origin[rec._secondary_unit_fields["qty_field"]]
    or rec[rec._secondary_unit_fields["qty_field"]]
    or default_qty_field_value
)
```

cc @sergio-teruel

legalsylvain:
Hi. I just discovered the module ``sale_order_qty_change_no_recompute`` which completely fills a need I've had for a long time.
However, I need to be able to enable/disable recompute on demand. I hope to be able to make a PR in the next few days.

I think that if my PR works and is accepted, there will be no need to mark the module as rebel.

PR #3028, opened 2024-03-20
[16.0][IMP] sale_order_product_recommendation: improve performance of propagating recommendations to sale order

trisdoan:
- As part of [improvements on 12.0](https://github.com/OCA/sale-workflow/pull/1100), `odoo.tests.Form()` was introduced to deal with onchange issues
- But meanwhile, in 16.0, `sale_*` modules were heavily refactored, mostly to get rid of onchanges in favor of computed fields:
  - see [here](https://github.com/odoo/odoo/commit/74e13888ee15c445e066b8098045d66581922566)
  - and [here](https://github.com/odoo/odoo/commit/8c7bd64e3e77a69a12053da4ae56b7d7bb458177)
- Given that `odoo.tests.Form()` is having a big impact on performances, and given the changes above, we consider it's ok now to get rid of it for 16.0+
- product_id is already indexed since v16: https://github.com/odoo/odoo/commit/318888fd835233dc17bb0d9f200544266a345ed4

- In [test_computes](https://github.com/OCA/sale-workflow/blob/cbccbb3da53c4be927d08ac198fcd9d56cfc8ae4/sale_order_product_recommendation_packaging_default/tests/test_recommendation.py#L89) and [test_preexisting_product](https://github.com/OCA/sale-workflow/blob/cbccbb3da53c4be927d08ac198fcd9d56cfc8ae4/sale_order_product_recommendation_packaging_default/tests/test_recommendation.py#L157), I forced product_packaging_qty=9 to avoid change package from Dozen to Box.


trisdoan:
DRAFT: working on tests

trisdoan:
Hi @pedrobaeza, please help to review my changes 🙏

sale_order_product_recommendation/models/sale_order_line.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
```
pedrobaeza: Please don't change the name of the file. It's OK to host in the "parent" model adaptations of the children one, so that if in the future, something at `sale.order` needs to be added, you don't have to rename the file again.

sale_order_product_recommendation/models/sale_order_line.py
```diff
@@ -0,0 +1,18 @@
...
+
+    # Add database index to improve performance of _recommendable_sale_order_lines_domain
+    qty_delivered = fields.Float(
+        string="Delivery Quantity",
+        compute="_compute_qty_delivered",
+        digits="Product Unit of Measure",
+        store=True,
+        readonly=False,
+        copy=False,
+        index="btree_not_null",
```
pedrobaeza: Indexes on floats are not interesting AFAIK.

sale_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -6,11 +6,13 @@
 from freezegun import freeze_time
 
 from odoo.exceptions import UserError
+from odoo.tests import tagged
 from odoo.tests.common import Form
 
 from .test_recommendation_common import RecommendationCase
 
 
+@tagged("post_install", "-at_install")
```
pedrobaeza: Why do you need this? It's better to not do it if not needed.

sale_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -154,6 +156,7 @@ def test_change_uom(self):
                 }
             ],
         )
+
```
pedrobaeza: Don't introduce empty lines inside a method.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -220,21 +218,16 @@ def action_accept(self):
...
                 continue
             if wiz_line.sale_line_id:
-                index = self.order_id.order_line.ids.index(wiz_line.sale_line_id.id)
                 if wiz_line.units_included or force_zero_units_included:
-                    with order_form.order_line.edit(index) as line_form:
-                        wiz_line._prepare_update_so_line(line_form)
+                    wiz_line.sale_line_id.update(wiz_line._prepare_update_so_line())
                 else:
-                    to_remove.append(index)
+                    wiz_line.sale_line_id.unlink()
```
pedrobaeza: Do this on batch as it was.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -385,3 +362,22 @@ def _get_unit_price_from_pricelist(self):
             product_currency=self.currency_id,
         )
         return price_unit
+
+    def _prepare_update_so_line(self):
```
pedrobaeza: Why changing the order of these methods increasing the diff or make difficult to review them?

sale_order_product_recommendation_elaboration/wizards/sale_order_recommendation.py
```diff
@@ -38,21 +38,26 @@ class SaleOrderRecommendationLine(models.TransientModel):
     _name = "sale.order.recommendation.line"
     _inherit = ["sale.order.recommendation.line", "product.elaboration.mixin"]
 
-    def _prepare_elaborations_line_form(self, line_form):
+    def _prepare_elaborations_line_form(self, vals):
```
pedrobaeza: Rename method accordingly:

```suggestion
    def _prepare_elaboration_line_vals(self, vals):
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: And maybe the names of the methods should be updated accordingly the changes:

```suggestion
    def _prepare_update_so_line_vals(self):
```

sale_order_product_recommendation_elaboration/wizards/sale_order_recommendation.py
```diff
@@ -38,21 +38,26 @@ class SaleOrderRecommendationLine(models.TransientModel):
...
         """Transfer elaborations to a sale order line form."""
-        line_form.elaboration_ids.clear()
-        for elaboration in self.elaboration_ids:
-            line_form.elaboration_ids.add(elaboration)
-        line_form.elaboration_note = self.elaboration_note
+        if self.sale_line_id:
+            self.sale_line_id.update({"elaboration_ids": [(fields.Command.clear())]})
```
pedrobaeza: Why are you changing here directly the sale line?

pedrobaeza (changes requested):

sale_order_product_recommendation_elaboration/wizards/sale_order_recommendation.py
trisdoan: Hi @pedrobaeza, do you think it's better to remove it ?
As I understood, elaboration_ids will be updated anyway at the end.

trisdoan:
DRAFT: working on pedro's requests

trisdoan:
Hi @pedrobaeza, I updated as you requested.

Please help to review it again 🙏

sale_order_product_recommendation/models/sale_order.py
```diff
@@ -1,11 +0,0 @@
...
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
-from odoo import fields, models
-
-
-class SaleOrderLine(models.Model):
-    _inherit = "sale.order.line"
-
-    # Add database index to get last sale price query performance
-    product_id = fields.Many2one(index=True)
```
pedrobaeza, line 11: Is this index not needed anymore?

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -220,21 +219,20 @@ def action_accept(self):
...
+                    )
                 else:
-                    to_remove.append(index)
+                    to_remove |= wiz_line.sale_line_id
             else:
                 sequence += 1
-                with order_form.order_line.new() as line_form:
-                    wiz_line._prepare_new_so_line(line_form, sequence)
-        # Remove at the end and in reverse order for not having problems
-        to_remove.reverse()
```
pedrobaeza, line 234: You are removing this part that is important.

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -311,21 +309,24 @@ def _compute_price_unit(self):
...
                 line.price_unit = line._get_last_sale_price_product()
 
-    def _prepare_update_so_line(self, line_form):
-        """So we can extend SO update"""
+    def _prepare_update_so_line_vals(self):
+        vals = {"product_uom_qty": self.units_included}
         if not self.product_uom_readonly:
-            line_form.product_uom = self.sale_uom_id
-        line_form.product_uom_qty = self.units_included
+            vals.update({"product_uom": self.sale_uom_id.id})
```
pedrobaeza: Cleaner syntax when it's only one element:

```suggestion
            vals["product_uom"] = self.sale_uom_id.id
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -311,21 +309,24 @@ def _compute_price_unit(self):
...
+        vals = {
+            "product_id": self.product_id.id,
+            "sequence": sequence,
+            "product_uom_qty": self.units_included,
+            "order_id": self.wizard_id.order_id.id,
+        }
         if not self.product_uom_readonly:
-            line_form.product_uom = self.sale_uom_id
-        line_form.product_uom_qty = self.units_included
+            vals.update({"product_uom": self.sale_uom_id.id})
```
pedrobaeza:
```suggestion
            vals["product_uom"] = self.sale_uom_id.id
```

sale_order_product_recommendation/wizards/sale_order_recommendation.py
```diff
@@ -311,21 +309,24 @@ def _compute_price_unit(self):
...
         if self.wizard_id.sale_recommendation_price_origin == "last_sale_price":
-            line_form.price_unit = self.price_unit
+            vals.update({"price_unit": self.price_unit})
```
pedrobaeza:
```suggestion
            vals["price_unit"] = self.price_unit
```

sale_order_product_recommendation_elaboration/wizards/sale_order_recommendation.py
```diff
@@ -38,21 +38,24 @@ class SaleOrderRecommendationLine(models.TransientModel):
     _name = "sale.order.recommendation.line"
     _inherit = ["sale.order.recommendation.line", "product.elaboration.mixin"]
 
-    def _prepare_elaborations_line_form(self, line_form):
+    def _prepare_elaborations_line_vals(self, vals):
```
pedrobaeza: I think these extra prepare methods are not needed, as you can do (and override the same) in the main method.

pedrobaeza (commented):
Thanks for the changes. Another round of suggestions and some important remarks, and for me will be good to go.

sale_order_product_recommendation/models/sale_order.py
trisdoan: Hello @pedrobaeza, thank you for you in-depth review

Indeed, product_id is already indexed since v16: https://github.com/odoo/odoo/commit/318888fd835233dc17bb0d9f200544266a345ed4

sale_order_product_recommendation/tests/test_recommendation.py
trisdoan: Hi @pedrobaeza,

The reason I used that decorator was to avoid **NotNullViolation**.
Traceback:
psycopg2.errors.NotNullViolation: null value in column "warehouse_id" of relation "sale_order" violates not-null constraint

pedrobaeza: But that's something on your end when you repeat the tests once the module is installed. Please remove it.

trisdoan:
Hello @pedrobaeza, I updated as you suggested

sale_order_product_recommendation/wizards/sale_order_recommendation.py
pedrobaeza: The reverse is still important, and don't know why are you doing `to_remove[::-1]`

trisdoan:
Hello @pedrobaeza, I updated the 'reverse' part

pedrobaeza (approved):

cyrilmanuel (approved):
Not an expert but LGTM

PR #3029, opened 2024-03-20
[FIX] sale_elaboration: propagate note-only lines to stock moves

yajo:
Before this patch, if a sale order line had no elaborations selected, but had elaboration notes, those notes didn't land in the created `stock.move` records when confirming the SO.

@moduon MT-5511


Shide (changes requested):
Thanks! Great job!

Sorry, I've wrong clicked request changes

fcvalgar (approved):
Functional review,

LGTM thank you @yajo

Shide (approved):

rafaelbn (approved):
👍🏼

PR #3036, opened 2024-03-21
[14.0][ADD] sale_automatic_workflow_stock_location_route

nayatec:
This module allows to automate the setting of a specific route on sale order lines.


rousseldenis:
@nayatec Thanks for this. In order to attract people to review this, try to make your PR's tittle meaningful, thanks.

Moreover, your commit message should look like `[14.0][ADD] sale_automatic_workflow_stock_location_route`

nayatec:
Thanks for the advice @rousseldenis !

hparfr (approved):

sale_automatic_workflow_stock_location_route/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+    def automatic_set_route_on_sol(self):
+        for sale in self:
+            route_id = sale.workflow_process_id.sale_line_route_id
+            if route_id:
+                sale.order_line.filtered(
+                    lambda sol: sol.qty_delivered_method == "stock_move"
+                ).route_id = route_id
+
+    def _action_confirm(self):
+        self.automatic_set_route_on_sol()
```
florian-dacosta, line 26: I think it is arguable to force it at confirmation, potentially erasing a route set manually.
At least it can bring confusion (as the user has no idea the route he could set manually won't be actually used).

An other way would be to set the route during the onchange (or make it a computed editable field) depending on workflow_process_id).

I guess it may be wanted to really erase the route on confirm, but it won't be ok for all use cases.

sale_automatic_workflow_stock_location_route/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This allow to automatically set a specific route on sale lines with product delivered by stock move, when the sale is confirmed.
```
florian-dacosta: I would rephrase, like on sale lines with physical (or any better way) product. I think "delivered by stock move" is not very clear for a none odoo developer person. I may be wrong.

sale_automatic_workflow_stock_location_route/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
...
+Just set the desired route on the sale automatic workflow.
```
florian-dacosta: Maybe add a small comment to indicates the configured route will erase any other route set manually?

florian-dacosta (approved):
Code review OK

Functionally I am not sure it is the best like explained but it can correspond to some use case I guess.
I wonder if something like this could make more sense configured on a `sale.channel` for instance.

But you could use sale_automatic_workflow and not sale_channel so why not.

rousseldenis:
@nayatec

nayatec:
FYI @florian-dacosta i took your remarks into account  by :
- adding a route policy to add a bit of flexibility
- review the onchange to actually do what will happen at confirmation time
- rewrite the readme for clarity

@rousseldenis i don't intend to further dev. that module in that version of Odoo. If i did not do any mistake that is to be corrected it is ready for merging imho.

rousseldenis:
@nayatec You didn't changed commit messages. They should respect formatting rules as already said here above. This will be closed as quite old.

rousseldenis:
@nayatec These are commit messages :sweat_smile: 

image

nayatec:
@rousseldenis i'm somehow relunctant ro rewrite the history when something is already publicly pushed, but i understand that it is ok. So i just did.

rousseldenis:
> @rousseldenis i'm somehow relunctant ro rewrite the history when something is already publicly pushed, but i understand that it is ok. So i just did.

@nayatec It is not pushed yet as in your branch and not on OCA one.

Thanks for the first commit but for updates you should also make the module name appear like `[IMP] sale_automatic_workflow_stock_location_route: (...)`

Many thanks

rousseldenis:
@nayatec Could you rebase ?

sale_automatic_workflow_stock_location_route/models/sale_order.py
rousseldenis: @nayatec Is this use case covered ? Manual route will be erased. Is this wanted ?

nayatec: @rousseldenis no, this use case is not covered, and yes manual routes will be erased, as it was initially wanted for the use case we had. That said i fully agree with Florian and it would certainly be a nice improvement to cover this, maybe with a kind of settable policy on the workflow like "always override" or "do not override if already set".
As i said, this module is no longer used in the project i developed it for, so i have no budget for it anymore, but if anybody find it useful and want to improve it i would be really happy !

rousseldenis: Ok, but I think if one configure the workflow with that feature, they are conscious of that route change.

We can go forward

rousseldenis (approved):

rousseldenis:
@sbidoul Pypi failed

sbidoul:
@rousseldenis please retry the merge

PR #3037, opened 2024-03-22
[16.0][IMP] sale_order_product_picker: Support use of discounts

### LINKED ISSUE #2842: sale_order_product_picker: quick-clicking on +1 button fails
For example, if you want to add 3 units and click the button 3 times quickly, chances are that only 1 or 2 of those clicks are actually handled.

I think it would make sense to throttle and defer the calls, and only when the user has stopped clicking for about 500ms, then upload the changes. Does that make sense?

Otherwise, at least it would help if the button appears as disabled while you can't use it. That visual feedback isn't the best, but at least gives a hint that you have to wait to click more.

_Originally posted by @yajo in https://github.com/OCA/sale-workflow/pull/2796#pullrequestreview-1743843628_

CarlosRoca13:
cc @Tecnativa TT48323

### DIFF
--- a/sale_order_product_picker/models/sale_order_picker.py
+++ b/sale_order_product_picker/models/sale_order_picker.py
@@ -1,9 +1,10 @@
 # Copyright 2023 Tecnativa - Sergio Teruel
 # Copyright 2023 Tecnativa - Carlos Dauden
 # License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from datetime import datetime, time
 
 from odoo import api, fields, models
-from odoo.tools import float_compare
+from odoo.tools import float_compare, ormcache
 
 
 class SaleOrderPicker(models.Model):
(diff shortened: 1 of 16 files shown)


sale_order_product_picker/models/sale_order_picker.py
```diff
@@ -100,21 +287,32 @@ def _get_last_sale_price_product(self, use_delivery_address=False):
...
             .with_context(prefetch_fields=False)
         )
-        return so_line.price_unit or 0.0
+        return so_line.price_unit, so_line.discount, so_line.price_reduce
 
     def add_to_cart(self):
         self.ensure_one()
         so_line = self.order_id.order_line.new({"product_id": self.product_id.id})
-        so_line.product_id_change()
+        # so_line._product_id_change()
```
sergio-teruel: Is ok this commented line??

sergio-teruel (changes requested):

sale_order_product_picker/models/sale_order_picker.py
CarlosRoca13: The method doesn't exist now, removed line

sale_order_product_picker/static/src/x2many/x2many_field.esm.js
```diff
@@ -45,16 +72,21 @@ patch(X2ManyField.prototype, "sale_order_product_picker.X2ManyField", {
...
+                        line_price_reduce: price_discount[1],
                     });
                 }
                 return res;
             };
-            const newUpdateRecord = (record) => {
+            const newUpdateRecord = async (record) => {
                 // Function used when updating a record.
-                const res = updateRecord(record);
+                lineList = this.props.record.model.root.data.order_line;
```
chienandalu, line 85: You aren't declaring it... doesn't this fail?

sale_order_product_picker/static/src/x2many/x2many_field.esm.js
```diff
@@ -67,28 +99,41 @@ patch(X2ManyField.prototype, "sale_order_product_picker.X2ManyField", {
...
+                        line_price_reduce: price_discount[1],
                     });
                 }
                 return res;
             };
-            const newRemoveRecord = (record) => {
+            const newRemoveRecord = async (record) => {
                 // Function used when removing a record.
-                const res = removeRecord(record);
+                lineList = this.props.record.model.root.data.order_line;
```
chienandalu, line 115: Same here..

CarlosRoca13:
That var is defined in line 29. I added comments explaining why need to redefine the list.

CarlosRoca13:
Added changes to make the button +1 faster. With these changes, what we do is to store the changes to be made while pressing the +1 button. Each time the button is pressed, the timer that will execute the writing of the lines is reset. 

This relieves the server of unnecessary calls, since everything will be written at the same time. And the user will not have the feeling of slowness when writing on 3G networks.

Here is a sample:
Example picker16 0

As you can see in the example, the time to write the pending lines is 1 sec, but it can be increased using the system parameter `sale_order_product_picker.delay` and the number of seconds to wait.

CarlosRoca13:
@yajo I believe this should resolve this issue :smile: https://github.com/OCA/sale-workflow/issues/2842

sergio-teruel:
@CarlosRoca13 Can you put this PR as draft until september?

CarlosRoca13:
This PR is ready for review

yajo (approved):
Thanks!

sergio-teruel (approved):

PR #3038, opened 2024-03-25
[14.0][IMP] sale_order_invoice_amount: make sure to respect currencies

DavidJForgeFlow:
Using this module in a system with multiple currencies I have found some issues with this ones.

When computing the amount from invoice to the sale order it will firstly convert correctly the currency.

Also when the sale order is marked as fully invoiced, the amount invoiced is automatically settled as 0.

Lastly, now the uninvoiced amount is not always total_amount - invoiced_amount. Now it checks the qty invoiced to compute the uninvoiced amount.

@ForgeFlow

### DIFF
--- a/sale_order_invoice_amount/models/sale_order.py
+++ b/sale_order_invoice_amount/models/sale_order.py
@@ -29,12 +29,27 @@ class SaleOrder(models.Model):
     )
     def _compute_invoice_amount(self):
         for rec in self:
-            if rec.state != "cancel" and rec.invoice_ids:
+            if (
+                rec.state != "cancel"
+                and rec.invoice_ids
+                and rec.invoice_status != "invoiced"
+            ):
                 rec.invoiced_amount = 0.0
                 for invoice in rec.invoice_ids:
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
Tested, LGTM

LoisRForgeFlow:
It will need forward port once merged.

rousseldenis:
@DavidJForgeFlow I don't see a test case with different currencies. Maybe this should be mandatory to avoid regressions ?

LoisRForgeFlow:
> @DavidJForgeFlow I don't see a test case with different currencies. Maybe this should be mandatory to avoid regressions ?

@DavidJForgeFlow I think this is a good remark, could you attend it when you have time?

LoisRForgeFlow:
@rousseldenis tests have been added

rousseldenis (approved):

LoisRForgeFlow:
@DavidJForgeFlow This one is pending to fwported, isn't it?

PR #3041, opened 2024-03-25
[FIX] sale_advance_payment field amount_residual

sistecem:
There is an error in the current version, when canceling a paid Invoice , It still sums up in the invoice_paid_amount , resulting in wrong amount_residual in sale order


rousseldenis (approved):
Code review.

sale_advance_payment/models/sale.py
```diff
@@ -86,7 +86,8 @@ def _compute_advance_payment(self):
             for inv in order.invoice_ids.filtered(
                 lambda x: x.move_type == "out_invoice"
             ):
-                invoice_paid_amount += inv.amount_total - inv.amount_residual
+                if inv.state == "posted":
```
AungKoKoLin1997, line 89: How about we just add this condition in `filtered` instead of using both `filtered` and `if`?

rousseldenis (approved):
Code review

rousseldenis:
@sistecem Could you squash your commits ?

sale_advance_payment/models/sale.py
rousseldenis: @sistecem Minor enhancement

rousseldenis:
@sistecem

alexey-pelykh (commented):
Thanks for the fix @sistecem, the root cause analysis is correct -- a cancelled invoice zeroes out `amount_residual` while keeping `amount_total`, so `amount_total - amount_residual` wrongly reports the full invoice amount as "paid". Filtering on `state == "posted"` is the right approach.

Two items still open from earlier feedback:

1. As both @AungKoKoLin1997 and @rousseldenis suggested, the state check would read better folded into the existing `filtered()` lambda rather than a separate `if` inside the loop:

```python
for inv in order.invoice_ids.filtered(
    lambda x: x.move_type == "out_invoice" and x.state == "posted"
):
    invoice_paid_amount += inv.amount_total - inv.amount_residual
```

This keeps the filtering logic in one place and avoids the extra indentation level.

2. The branch still has merge commits from syncing with upstream -- please squash as @rousseldenis requested.

The test coverage for the cancel scenario looks good.

rousseldenis:

PR #3043, opened 2024-03-26
 [MIG] sale_order_line_date: Migration to 17.0

yannoliv:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

Check also the pre-commit error.

rousseldenis:
@yannoliv Commit history is still wrong

rousseldenis:
Moreover, there is already another PR for this. Could you close this and review #2962 instead ?

PR #3044, opened 2024-03-26
[ADD] sale_stock_partner_warehouse

santostelmo:
**Description**
Allow to choose by default a warehouse on SO based on a Partner parameter.
If the warehouse parameter is completed, then this value would be the default one on the sale.
If empty, then the warehouse on sale would be the standard default warehouse.


sale_stock_partner_warehouse/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "data": [
+        "views/partner_view.xml",
+    ],
+    "depends": [
+        "sale_stock",
+        "delivery",
```
simahawk: why `delivery`?
If it's only for the view, you can simply use `<group name="sale" position="inside">`.

sale_stock_partner_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    warehouse_id = fields.Many2one(
+        compute="_compute_warehouse_id", store=True, precompute=True
```
simahawk, line 11: why `precompute`?

simahawk (changes requested):

sale_stock_partner_warehouse/models/sale_order.py
santostelmo: To compute the value before the record creation

sale_stock_partner_warehouse/__manifest__.py
santostelmo: I removed dependency

luisDIXMIT (approved):
LGTM!!

sale_stock_partner_warehouse/models/res_partner.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    sale_warehouse_id = fields.Many2one("stock.warehouse", string="Sale Warehouse")
```
rousseldenis, line 10: IMHO, this should be company dependent.

sale_stock_partner_warehouse/models/sale_order.py
rousseldenis: @santostelmo As warehouse_id is defined in sale_stock, you don't need to repeat the properties. Just override _compute_warehouse_id()

sale_stock_partner_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+    @api.depends("user_id", "company_id", "partner_id")
```
rousseldenis: You don't need to repeat user_id and company_id. Odoo will append partner_id to depends list

sale_stock_partner_warehouse/models/res_partner.py
santostelmo: I agree. The company of the partner, if defined must be consistent with the one set in the warehouse

Jaimermaccione (approved):
Functional review on runboat! LGTM!

rousseldenis:
@simahawk Could you update your review ?

sale_stock_partner_warehouse/models/res_partner.py
santostelmo: Actually after thinking about that I think we don't want to have  `sale_warehouse_id` dependent of the company. We want to be able to select any warehouse from active companies.
With that option a Sale can be created in a company A and Delivered from the Warehouse of the company B

sale_stock_partner_warehouse/models/sale_order.py
santostelmo: Thank you I updated that

rousseldenis:
@simahawk

sale_stock_partner_warehouse/models/sale_order.py
simahawk: @santostelmo I know, I mean... why do you need that? I guess it's because the wh is supposed to be already there on create.
If there's a motivation, it's worth to add a comment there.

simahawk (approved):
Minor remark, LG otherwise

HaraldPanten (approved):

PR #1436, opened 2021-01-29
14.0 pricelist tax

bguillot:
This modules alllows you to set a pricelist in tax included.

It will automatically replace the tax excluded tax by the included tax on the sale order line.

It is usefull if you want to set a fixed price tax included (for instance : 9.99€) on the pricelist for B2C and an untaxed fixed price on the pricelist of B2B

PR #3045, opened 2024-03-26
[FIX] sale_product_rating_verified: Don't set computed method to required

rousseldenis:
As the False value is set for all records at the beginning of the compute method, don't set it to required as previous created records before this module install will avoid to set the NOT NULL constraint at module install

This is due to project module install before this one in this PR: https://github.com/OCA/sale-workflow/pull/1436


rousseldenis:
/ocabot merge patch

As this is obvious and preventing merging elsewhere

PR #3046, opened 2024-03-26
[REF] sale_order_type: compute of pricelist and payment_terms

matiasperalta1:
The error is when the payment term is set in the partner as well as pricelist.
The case of replication by creating a quotation and introducing a partner with the payment term and pricelist set in its configuration. Then we manually change the payment term and pricelist to another. And finally we change the type of sale. Here we observe that the aforementioned fields are recomputed to those configured in the partner and this is an error. With this PR the error no longer persists since all cases are considered.

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -91,7 +91,6 @@ class SaleOrder(models.Model):
                 order.picking_policy = order_type.picking_policy
         return res
 
-    @api.depends("type_id")
     def _compute_payment_term_id(self):
         res = super()._compute_payment_term_id()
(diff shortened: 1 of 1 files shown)


rousseldenis:
@matiasperalta1 I'm not sure about the solution. We should avoid onchanges.

A test case should be welcomed.

rousseldenis:
@matiasperalta1

matiasperalta1:
@rousseldenis 
Hi! sorry for the delay in the response. I analyzed another type of solution for the case but for the moment I understand that it is the best. I know that you should avoid using onchange but in some cases I can't think of how we could replace it.

PR #3047, opened 2024-03-27
[14.0][IMP] sale_product_multi_add: pricelist compatibility

FrancescoBallerini:
Before commit:

Adding sale order lines from `sale.import.products` assign `list_price`
as so line `price_unit`

After commit:

Adding sale order lines from `sale.import.products` will fetch the price
accordingly to the selected pricelist. Pricelist price will be assigned
as so line `price_unit` if a pricelist item matches so line `product_id`.
If no pricelist items matches so line `product_id`, the `list_price` will
be assigned as sale order line `price_unit`


AungKoKoLin1997 (approved):
Code Review: LGTM

rousseldenis (approved):
LGTM

PR #3049, opened 2024-03-28
[17.0][MIG] sale_order_line_menu: Migration to 17.0

MohamedOsman7:


sale_order_line_menu/models/sale_order_line.py
```diff
@@ -0,0 +1,20 @@
...
+        readonly=True,
+        store=True,
+        index=True,
+    )
+    account_analytic_id = fields.Many2one(
+        related="order_id.analytic_account_id",
+        readonly=True,
+        store=True,
+        index=True,
+    )
```
CRogos: I think we should fix the typo and rename field the same as in sale.order:
 account_analytic_id -> analytic_account_id

of cause this needs also to be updated in the views.

CRogos (changes requested):

celm1990:
Please split commits according guideslines
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#technical-method-to-migrate-a-module-from-160-to-170-branch
image

rafaelbn:
/ocabot migration sale_order_line_menu

@moduon @EmilioPascual

CRogos:
@celm1990 I think you are looking for these two commits.

I agree that the naming is not 100%, this is because we are using [oca-port ](https://github.com/OCA/oca-port) which generates the first commit automatically and the commit message is still from migration 14->15.

image

CRogos (approved):

EmilioPascual (approved):

angelmoya (approved):

CRogos:
@EmilioPascual could you retry with "nobump" at the end?

rafaelbn (approved):
/ocabot merge nobump

PR #3051, opened 2024-04-02
Add chinese translation

ryanhex53:


rousseldenis:
@ryanhex53 Thanks for this. Could you please use this instead ? https://odoo-community.org/resources/translate

Thanks

PR #3055, opened 2024-04-04
[15.0][IMP] sale_automatic_workflow: Confirm pickings with delivery_package_number module installed

manuelregidor:
When the module delivery_package_number is installed, the cron that confirm pickings do not work when the following conditions in the file stock_picking.py in delivery_package_number are met, as it tries to open the number of packages selection wizard:

`
if (res and test_condition and isinstance(res, bool) and any(picking.ask_number_of_packages for picking in self) and not self.env.context.get("bypass_set_number_of_packages"))
`
Including delivery_package_number as a dependency of sale_automatic_workflow would not make much sense, but considering the "bypass_set_number_of_packages" context parameter seems logic as both modules can be used together.

I-6003

### DIFF
--- a/sale_automatic_workflow/models/stock_picking.py
+++ b/sale_automatic_workflow/models/stock_picking.py
@@ -32,5 +32,10 @@ class StockPicking(models.Model):
                 ):
                     for move_line in move.move_line_ids:
                         move_line.qty_done = move_line.product_uom_qty
-            picking.with_context(skip_immediate=True, skip_sms=True).button_validate()
+            picking.with_context(
+                skip_immediate=True,
+                skip_sms=True,
+                # Necessary in case module delivery_package_number is in use
+                bypass_set_number_of_packages=True,
+            ).button_validate()
         return True


manuelregidor:
@luis-ron @ValentinVinagre

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@HaraldPanten

HaraldPanten (approved):

HaraldPanten:
@pedrobaeza @sergio-teruel @rousseldenis WDYT?

sergio-teruel:
Is there no other way to proceed?
The automatic workflow has a context to skip in the other modules??

On the other hand, what about other wizards like as expired lots, backorders etc.. Should any context clues be taken into account here?

I do not like the automatic picking validation because the picking can required manual actions... But if there is no other solution..

sergio-teruel:
Mybe call _action_done method instead button_validate??

sale_automatic_workflow/models/stock_picking.py
```diff
@@ -32,5 +32,10 @@ def validate_picking(self):
                 ):
                     for move_line in move.move_line_ids:
                         move_line.qty_done = move_line.product_uom_qty
-            picking.with_context(skip_immediate=True, skip_sms=True).button_validate()
+            picking.with_context(
+                skip_immediate=True,
+                skip_sms=True,
+                # Necessary in case module delivery_package_number is in use
+                bypass_set_number_of_packages=True,
```
rousseldenis, line 39: Nope. This is not correct. You should find a way to do a glue module instead.

With your way, this context could grow indefinitely with keys this module direct dependencies are not aware of.

HaraldPanten:
THX guys 👍

manuelregidor:
We'll make the changes in our repository.

PR #3061, opened 2024-04-09
[FIX] sale_triple_discount: apply rounding 

ajaniszewska-dev:
sometimes in case of multiplicative discount, subtotal on the line is not same as untaxed amount. Seems due to rounding as so far result returned by method could be something like: `284.64151515`

before:
Screenshot from 2024-04-09 13-11-50

after:
image

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -58,7 +58,9 @@ class SaleOrderLine(models.Model):
         final_discount = 1
         for discount in discounts:
             final_discount *= discount
-        return 100 - final_discount * 100
+        result = 100 - final_discount * 100
+        dp = self.env.ref("product.decimal_discount").precision_get("Discount")
+        return round(result, dp)
 
     def _discount_fields(self):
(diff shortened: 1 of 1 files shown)


JuMiSanAr (approved):

sale_triple_discount/models/sale_order_line.py
```diff
@@ -58,7 +58,8 @@ def _multiplicative_discount(self):
         final_discount = 1
         for discount in discounts:
             final_discount *= discount
-        return 100 - final_discount * 100
+        result = 100 - final_discount * 100
+        return round(result, 3)
```
rousseldenis: @ajaniszewska-dev Why 3 (and not another number ...)?

rousseldenis: IMHO, you should (at least) use the configuration one:

image

ajaniszewska-dev: @rousseldenis thank you for remark, corrected code.

rousseldenis (approved):

PR #3062, opened 2024-04-09
[17.0][MIG] sale cancel reason

maciej-wichowski:


sale_cancel_reason/wizard/cancel_reason_view.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="inherit_id" ref="sale.sale_order_cancel_view_form" />
+            <field name="arch" type="xml">
+                <xpath expr="//footer" position="before">
+                    <br />
+                    <p>
+                        Choose the reason for the cancellation of the
+                        sale order.
+                    </p>
+                    <group>
+                        <field name="reason_id" widget="selection" />
```
celm1990, line 15: Please remove the selection widget. With many records, filtering becomes complicated, and not all records are displayed.
```suggestion
                        <field name="reason_id" options="{'no_create': True, 'no_open': True}" />
```

celm1990 (changes requested):
Could you add a `Reason` field to this wizard?
image
image

maciej-wichowski:
Hello @celm1990 , thank you for your review! I added possibility to choose cancel reason on mass quotation cancel wizard.
Regarding removing selection widget: I don't think we should make such improvements during module migration. I suggest to create an Issue in this repo to properly discuss your suggestion with the maintainers.

sale_cancel_reason/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Sale Cancel Reason",
-    "version": "16.0.1.0.1",
+    "version": "17.0.1.0.1",
```
rousseldenis: Usually we reset to `17.0.1.0.0`

rousseldenis (approved):
Minor remark

sale_cancel_reason/wizard/mass_cancel_orders_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="mass_cancel_orders_view_form" model="ir.ui.view">
+        <field name="name">sale.mass.cancel.orders.form</field>
+        <field name="model">sale.mass.cancel.orders</field>
+        <field name="inherit_id" ref="sale.mass_cancel_orders_view_form" />
+        <field name="arch" type="xml">
+            <footer position="before">
+                <group>
+                    <field name="reason_id" />
```
celm1990: image

```suggestion
                    <field name="reason_id" options="{'no_create': True, 'no_open': True}"/>
```

celm1990 (changes requested):
For consistency, both wizard fields must be the same (either many2one with widget=selection or not)

sale_cancel_reason/wizard/mass_cancel_orders_view.xml
maciej-wichowski: Forgot about it, thanks.

celm1990 (approved):
I think these two commits can be squashed into a single commit
image

maciej-wichowski:
@rousseldenis , since it's approved, could you merge?

rousseldenis:
> I think these two commits can be squashed into a single commit image

@celm1990 FYI, if the commit is reflecting an improvement or is not related to the preceding one (migration), it's common to let them in place. As maybe, it can be easily forward/backported.

PR #3064, opened 2024-04-09
16.0 mig sale stock secondary unit

jcadhoc:


rousseldenis:
@jcadhoc Thanks for this.

Could you check pre-commit ?

Moreover, if there are pending PR's dependencies, please add pointers in this PR description in order to help reviewers.

You can also add temporary references here in order to run tests and runboat that allows to review functionally too. https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

Thanks

PR #3068, opened 2024-04-11
[IMP] sale_elaboration: display elaboration notes in picking operations report

yajo:
~If the option to print elaboration data on delivery slips is enabled, they also will appear now in the picking operations report.~

~The XML ID of that security group is kept for historical reasons, but it is re-labeled now.~

@moduon MT-5508

### DIFF
--- a/sale_elaboration/__manifest__.py
+++ b/sale_elaboration/__manifest__.py
@@ -24,6 +24,7 @@
         "views/stock_move_views.xml",
         "views/stock_picking_views.xml",
         "reports/report_deliveryslip.xml",
+        "reports/report_picking_operations.xml",
     ],
     "pre_init_hook": "pre_init_hook",
     "maintainers": ["rafaelbn", "yajo"],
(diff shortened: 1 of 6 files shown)


yajo:
This is how it looks: 
image

Gelojr (changes requested):
Hi @yajo ,
Can you check this?
https://www.loom.com/share/6d265d7760b944b38938671588ae594e?sid=063f5f85-ce73-468d-b829-54cb6541512e
Thanks

yajo:
Did you make sure to enable this option?
image

Shide (approved):
Functional and code review

image

Shide:
@Gelojr you need permissions to show it or activate on settings for every user

rafaelbn:
@yajo @Shide this option should be independent for delivery slips and picking operations

imagen

In our case:

- Always should be printed in the picking operations
- Optional (never) in delivery slip

But module should be generic so 4 combinations must be supported.

OK? Thanks

yajo:
> this option should be independent for delivery slips and picking operations

Fixed, please review again.

rafaelbn (approved):
/ocabot merge minor

PR #3081, opened 2024-04-17
[16.0][IMP] sale_order_carrier_auto_assign: assign carrier on create

jbaudoux:
Allow to assign carrier on create

Support multi-company

I'll drop this feature from https://github.com/OCA/delivery-carrier/tree/16.0/delivery_auto_refresh so that auto_refresh on refresh the delivery line but don't assign carrier.

image

cc @sebalix @santostelmo


jbaudoux:
I'll move those improvements to v17.0. For 16.0, I'll keep the new methods in https://github.com/OCA/delivery-carrier/pull/799

PR #2601, opened 2023-07-12
[15.0][MIG] pricelist_cache

juanpgarza:
Standard migration

PR #3086, opened 2024-04-19
[15.0][MIG] pricelist_cache

valentincastravete:
Standard migration


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0.

valentincastravete:
Hi @pedrobaeza 
Thanks for the explanation, but I only wanted to try to fix the failed tests from it including all the commits on the 14.0 branch, as there is another active PR for the MIG of pricelist_cache #2601.
But I will close this PR and try the tests on a local env instead.

PR #3088, opened 2024-04-19
[16.0][FIX] sale_order_carrier_auto_assign: fix all service

jbaudoux:
Do not add carrier when SO contains only service lines

Also increased test coverage

Removed Alpha status

@sebalix @simahawk @santostelmo


santostelmo (approved):

sale_order_carrier_auto_assign/__manifest__.py
```diff
@@ -4,10 +4,10 @@
     "name": "Sale Order Carrier Auto Assign",
     "summary": "Auto assign delivery carrier on sale order confirmation",
     "version": "16.0.1.1.0",
-    "development_status": "Alpha",
```
sebalix, line 7: If it's not `Alpha`, is it `Beta`/`Mature`/`Production/Stable`?
I think we should not remove it, I do not know the default value...?

jbaudoux: Default is Beta

jbaudoux:
@sebalix Can we merge this?

jbaudoux:
ping @sebalix

rousseldenis (approved):

PR #3092, opened 2024-04-23
[14.0][IMP] product_supplierinfo_for_customer_sale: menu for customer info

toita86:


francesco-ooops (approved):
Functional ok!

product_supplierinfo_for_customer_sale/views/sale_view.xml
```diff
@@ -22,4 +22,79 @@
...
+        <field
+            name="inherit_id"
+            ref="product_supplierinfo_for_customer.product_customerinfo_form_view"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="before">
+                <field name="product_tmpl_id" required="1" />
+            </xpath>
+        </field>
+    </record>
```
HekkiMelody: **issue**: both of these add the field `product_tmpl_id` to every `customerinfo` view, but we don't want these fields, e.g., in the tree view embedded in the product

image


(also, it'd make sense visually for `product_tmpl_id` to be next to `product_id`.)

A solution could be to mark these as `primary=True` with a priority of `99`. That way, these views won't be the default views. Then in the `act_window` below, specify that you want to use these views in particular.

HekkiMelody (changes requested):
Code and functional review: Mostly looks good, except one point.

product_supplierinfo_for_customer_sale/views/sale_view.xml
```diff
@@ -22,4 +22,84 @@
...
```
HekkiMelody: **issue**: I think this view deserves the same treatment as the tree view. Don't forget to add it to the `action` otherwise it won't be used!

HekkiMelody (approved):
Code review, LGTM

francesco-ooops (approved):
As per USAGE: 

In Sales > Product > Product Customer Info, user can access a list view with all supplierinfos set in products, as well as create new ones.

@pedrobaeza good IMP?

francesco-ooops:
@rousseldenis or @simahawk anyone available for merge? Thanks!

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
```diff
@@ -0,0 +1,9 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "product.customerinfo"
+
+    @api.onchange("product_tmpl_id")
+    def onchange_product_tmpl_id(self):
```
pedrobaeza: This shouldn't be done this way, but with a computed writable.

product_supplierinfo_for_customer_sale/views/sale_view.xml
```diff
@@ -22,4 +22,89 @@
...
+
+    <record id="action_product_customerinfo" model="ir.actions.act_window">
+        <field name="name">Products Customer info</field>
+        <field name="res_model">product.customerinfo</field>
+        <field name="view_mode">tree,form</field>
+        <field
+            name="view_ids"
+            eval="[
+            (5,0,0),
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('product_supplierinfo_for_customer_sale.product_customerinfo_tree_view')}),
```
pedrobaeza, line 97: Why do you need to indicate explicitly the views?

toita86: I created an independent view starting from the inherited one, and without explicitly indicating that new one, it wouldn't work. The new view it's used to display the product_template_id field, which is not normally shown.

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
toita86: Hi @pedrobaeza, It serves to enhance the user experience, clearing the field and requiring them to select a new variant.

product_supplierinfo_for_customer_sale/views/sale_view.xml
toita86: @pedrobaeza let me know if there's anything I'm missing

francesco-ooops:
@pedrobaeza good to go?

pedrobaeza:
@francesco-ooops I'm not comfortable lately reviewing and merging your PRs since you don't forward-port these changes to upper versions.

francesco-ooops:
We will gladly forward port any applicable PR, excluding: 

- Modules that have not been ported
- Modules that have been heavily refactored in later versions, whose changes have not been backported

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
pedrobaeza: But as said, switch to computed writable as the way to implement it.

pedrobaeza:
OK, I hope that promise is committed in the future. I continue helping you.

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
```diff
@@ -0,0 +1,14 @@
...
+    _inherit = "product.customerinfo"
+
+    product_id = fields.Many2one(
+        "product.product", compute="_compute_product_id", readonly=False
+    )
+
+    @api.depends("product_tmpl_id")
+    def _compute_product_id(self):
+        for record in self:
+            record.product_id = False
```
pedrobaeza: No need of loop.

```suggestion
        self.product_id = False
```

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "product.customerinfo"
+
+    product_id = fields.Many2one(
+        "product.product", compute="_compute_product_id", readonly=False
+    )
```
pedrobaeza:
```suggestion
    # convert it to computed writable
    product_id = fields.Many2one(
        compute="_compute_product_id", readonly=False, store=True
    )
```

although this can be put in the main module for avoiding incompatibilities.

product_supplierinfo_for_customer_sale/models/product_customerinfo.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
```
pedrobaeza:
```suggestion
class ProductCustomerinfo(models.Model):
```

product_supplierinfo_for_customer_sale/views/sale_view.xml
pedrobaeza: OK. You can do it directly using `view_id` field for simplicity.

toita86: In this case, I believe that I need to use `view_ids` in order to specify both `tree` and `form` views. With `view_id` i can only specify one view.

pedrobaeza (approved):
/ocabot merge minor

pedrobaeza:
To be fw-ported to upper versions.

PR #3095, opened 2024-04-24
[17.0][MIG] sale_order_qty_change_no_recompute

CRogos:
not including: https://github.com/OCA/sale-workflow/pull/3023


MohamedOsman7 (approved):
Functional & code review LGTM

Anxo82 (approved):
LGTM :+1:

omalbastin:
https://github.com/OCA/sale-workflow/blob/71e12fd11ec4c54df0b8afabb5801f3edcbb2269/sale_order_qty_change_no_recompute/monkeypatching.py#L20

What if we need to apply similar functionality in purchase and account move? Do we have to monkeypatch again and again? just curious

pedrobaeza:
Yes, the call to the original method will assure that everything is executed and put in place. At the end, this acts like a regular method override. Do you have any better solution?

omalbastin:
@pedrobaeza Nope.
I thought  the monkeypack will apply on latest one only. That why I asked for clarification. In similar cases what  I done was rewriting the method again. Didnt thought of moneypatch after odoo removed auditlog module from core.  :)

pedrobaeza:
We save the previous reference and call that one, so they are not overwritten. Have you check it with an example?

Arozeo:
Hi everyone,

Sorry to bother but I cannot seem to get this module working in v17, or I am missing something

I tried with a Quote and with a confirmed Sale Order, doesn't work
Also saving between setting the custom sale price and changing the Qty, doesn't work too

Each time the price get cleared with the base one set on the product template

Thanks

CRogos:
@Arozeo could you solve your issue?

I have tried to retest this issue, but even without installing this module, the unit price is not reset.
I can reproduce the behavior in the "Optional Products" that the price is reset when changing the quantity, but this module does not help on Optional products. 

Maybe this module is obsolete due to this odoo change?
https://github.com/odoo/odoo/commit/fbf36f02e2b4dd3ec351294fd835cc03d57cd4f7

Arozeo:
Hi @CRogos 

We just upgraded from v14 to v17 last Thursday and for now I'm a bit overwhelmed by change management and post-upgrade tasks 😅

It's on my to-do list though, I will let you know in Septembre when I will have the time to try it again on a staging

Thanks
Eliott

Issue #3093, opened 2024-04-23
sale_force_invoiced: set amount_to_invoice to zero


alessandro-fiorino:
The sale_force_invoiced module just updates the invoice_status to invoiced.
It should also set the amount_to_invoice to zero on the lines to be more consistent, without that for example the sale_order_amount_to_invoice module shows non-zero amount to invoice values for "fully invoiced" orders, which can confuse the user.

legalsylvain:
Makes sense, could you make a PR ?

alessandro-fiorino:
Just did it!

alessandro-fiorino:
@legalsylvain would you review it ?

alessandro-fiorino:
The code has been merged. Thanks

PR #3101, opened 2024-04-25
[16.0][IMP] sale_force_invoiced: update untaxed_amount_to_invoice

alessandro-fiorino:
Recompute untaxed_amount_to_invoice on the sale order lines when force_invoiced is changed.
Issue https://github.com/OCA/sale-workflow/issues/3093

### DIFF
--- a/sale_force_invoiced/models/sale_order.py
+++ b/sale_force_invoiced/models/sale_order.py
@@ -24,3 +24,14 @@ class SaleOrder(models.Model):
             lambda so: so.force_invoiced and so.state in ("sale", "done")
         ).update({"invoice_status": "invoiced"})
         return res
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.depends("order_id.force_invoiced")
+    def _compute_untaxed_amount_to_invoice(self):
+        force_invoiced = self.filtered(lambda x: x.order_id.force_invoiced)
+        force_invoiced.update({"untaxed_amount_to_invoice": 0.0})
+        not_forced = self - force_invoiced
+        return super(SaleOrderLine, not_forced)._compute_untaxed_amount_to_invoice()


sale_force_invoiced/models/sale_order.py
```diff
@@ -24,3 +24,14 @@ def _compute_invoice_status(self):
             lambda so: so.force_invoiced and so.state in ("sale", "done")
         ).update({"invoice_status": "invoiced"})
         return res
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.depends("order_id.force_invoiced")
```
legalsylvain, line 32: Original dependencies are : 

``    @api.depends('state', 'price_reduce', 'product_id', 'untaxed_amount_invoiced', 'qty_delivered', 'product_uom_qty')
``

I guess you should write : 
``    @api.depends('state', 'price_reduce', 'product_id', 'untaxed_amount_invoiced', 'qty_delivered', 'product_uom_qty', 'order_id.force_invoiced')
``
Don't you think ?

sale_force_invoiced/models/sale_order.py
```diff
@@ -24,3 +24,14 @@ def _compute_invoice_status(self):
             lambda so: so.force_invoiced and so.state in ("sale", "done")
         ).update({"invoice_status": "invoiced"})
         return res
+
+
+class SaleOrderLine(models.Model):
```
legalsylvain, line 29: could you move this code into models/sale_order_line.py dedicated file ?

legalsylvain (commented):
some remarks inline. 
Otherwise, LGTM.

However, I don't use this module. I prefer original authors confirm this change is correct.

CC : @JordiBForgeFlow

sale_force_invoiced/models/sale_order.py
pedrobaeza: No, they are accumulative. You don't need to rewrite all of them.

legalsylvain: oh ! thanks, I didn't know !

sale_force_invoiced/models/sale_order.py
alessandro-fiorino: Split in dedicated file done

pedrobaeza: I prefer to not split it, as this a "child" model, and it's better to have everything together.

pedrobaeza (approved):
Thanks!

flotho (approved):
code review 
LGTM
thanks

PR #3102, opened 2024-04-25
[16.0][FIX] sale_invoice_policy: compute qty to invoice

ioans73:
I have cherry-picked the commit of PR [#2763](https://github.com/OCA/sale-workflow/pull/2763), to try to fix the module

Services should be computed based on the policy established in the product, right? Because it makes no sense to do it based on the quantity delivered, if that is the policy established in the order

@ACheung-FactorLibre @pedrobaeza @rousseldenis


pedrobaeza:
Yes, the override should be in the quantity to invoice. I wrote it correctly at first instance in the other PR, but then put received quantity in the second comment (now also fixed). Please check pre-commit.

ioans73:
@pedrobaeza 🍏

sale_invoice_policy/models/product_template.py
```diff
@@ -1,54 +1,18 @@
...
-        for tmpl in self:
-            if tmpl.type != "service" and invoice_policy:
-                tmpl.invoice_policy = invoice_policy
-            else:
-                tmpl.invoice_policy = tmpl.default_invoice_policy
-
-    @api.model
-    def _search_invoice_policy(self, operator, value):
-        return [("default_invoice_policy", operator, value)]
+    invoice_policy = fields.Selection(default=_default_invoice_policy)
```
pedrobaeza: If preserved, for being clearer it should be:

```suggestion
    # existing field only adding the default
    invoice_policy = fields.Selection(default=_default_invoice_policy)
```

sale_invoice_policy/models/res_config_settings.py
```diff
@@ -7,11 +7,6 @@
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
-    sale_default_invoice_policy = fields.Selection(
```
pedrobaeza, line 10: Why removing this?

sale_invoice_policy/models/product_template.py
```diff
@@ -1,54 +1,18 @@
...
-        string="Default Invoicing Policy",
-        help="Ordered Quantity: Invoice based on the quantity the customer "
-        "ordered.\n"
-        "Delivered Quantity: Invoiced based on the quantity the vendor "
-        "delivered (time or deliveries).",
-        default=lambda self: self.env["ir.default"].get(
-            "res.config.settings", "default_invoice_policy"
-        ),
-    )
+    def _default_invoice_policy(self):
```
pedrobaeza: Is this really needed? Isn't this already done by default?

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -5,64 +5,25 @@
...
-                    # assign the value in the modified context to self
-                    line_by_id[line.id].invoice_status = line.invoice_status
-        # Not to break function if (it could not happen) some records
-        # were not in so_lines
-        super(SaleOrderLine, self - done_lines)._compute_invoice_status()
+    def _compute_qty_to_invoice(self):
+        other_lines = self.env["sale.order.line"]
+        for line in self:
+            if line.product_id.type == "service" or not line.order_id.invoice_policy:
+                other_lines |= line
```
pedrobaeza: Instead of going one by one and adding it to the recordset, which means to recreate it each time, do `filtered` for getting the result in one shot.

sale_invoice_policy/models/product_template.py
ioans73: By default it's a computed field. I think we can leave this (by configuration you can choose the default value)

ioans73: Sorry, we don't need it because we already have this by default https://github.com/odoo/odoo/blob/16.0/addons/sale/models/res_config_settings.py#L11

sale_invoice_policy/models/res_config_settings.py
ioans73: `default_get` uses this field https://github.com/odoo/odoo/blob/16.0/addons/sale/models/res_config_settings.py#L11

sale_invoice_policy/models/sale_order_line.py
ioans73: done

pedrobaeza (approved):

fuentes73 (approved):

ioans73:
@pedrobaeza Could it be merged?

ioans73:
> @pedrobaeza Could it be merged?

Don't do it yet, because I want to check if I need to modify anything in this field https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order_line.py#L255

ioans73:
Update: compute untaxed amount to invoice

pedrobaeza:
Please check CI

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -1,68 +1,99 @@
...
-                super(SaleOrderLine, so_lines)._compute_qty_to_invoice()
-                for line in so_lines:
-                    # due to the change of context in compute methods,
-                    # assign the value in the modified context to self
-                    line_by_id[line.id].qty_to_invoice = line.qty_to_invoice
-        # Not to break function if (it could not happen) some records
-        # were not in so_lines
-        super(SaleOrderLine, self - done_lines)._compute_qty_to_invoice()
+        other_lines = self.filtered(
+            lambda l: l.product_id.type == "service" or not l.order_id.invoice_policy
```
ghost: Hi, 

I think you need to also check if `l.order_id.invoice_policy_required` is set on `False`, because if it is, you don't need to modify the line's `qty_to_invoice`, since it is not required.

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -1,68 +1,99 @@
...
-            so_lines = (
-                self.with_context(invoice_policy=invoice_policy)
-                .filtered(lambda x, p=invoice_policy: x.order_id.invoice_policy == p)
-                .with_prefetch()
+    def _compute_untaxed_amount_to_invoice(self):
+        other_lines = self.filtered(
+            lambda line: line.product_id.type == "service"
+            or not line.order_id.invoice_policy
+            or line.order_id.invoice_policy == line.product_id.invoice_policy
+            or line.state not in ["sale", "done"]
```
ghost, line 46: Same here, I think you need to also check if `l.order_id.invoice_policy_required` is set on `False`, because if it is, you don't need to modify the line, since it is not required.

ghost (changes requested):
Hello @ioans73,

Please make sure you are only modifying the lines whose order has `invoice_policy_required` set to True. By doing this, all tests will pass without any problem.

Thank you

sale_invoice_policy/models/sale_order_line.py
ioans73: done

sale_invoice_policy/models/sale_order_line.py
ioans73: done

ghost:
Hi @ioans73,

Please do a rebase so that tests are executed. 

Thank you

ioans73:
@ACheung-FactorLibre I have done it before, but I don't know why the actions are not being executed

pedrobaeza:
https://github.com/OCA/oca-addons-repo-template/issues/250

ioans73:
> Hi @ioans73,
> 
> Please do a rebase so that tests are executed.
> 
> Thank you

The tests are now ok

sale_invoice_policy/tests/test_sale_invoice_policy.py
```diff
@@ -36,6 +39,9 @@ def test_sale_order_invoice_order(self):
                 "invoice_policy": "order",
             }
         )
+        so._compute_invoice_policy_required()
```
pedrobaeza: Why do you need to manually call the compute method? It shouldn't be needed.

ioans73: I'm not really sure, because there is code that comes from the old PR, but I can review it

ioans73: Removed

pedrobaeza (approved):
/ocabot merge major

jguenat:
> Hello @ioans73,
> 
> Please make sure you are only modifying the lines whose order has `invoice_policy_required` set to True. By doing this, all tests will pass without any problem.
> 
> Thank you

Hello @ACheung-FactorLibre 

By doing this qty_to_invoice won't be recomputed for any sale order line if the config parameter sale_invoice_policy_required is set to false.

Meaning this modules does nothing unless you set sale_invoice_policy_required to True in the settings.

It was not the case in previous versions.

rousseldenis:
> > Hello @ioans73,
> > Please make sure you are only modifying the lines whose order has `invoice_policy_required` set to True. By doing this, all tests will pass without any problem.
> > Thank you
> 
> Hello @ACheung-FactorLibre
> 
> By doing this qty_to_invoice won't be recomputed for any sale order line if the config parameter sale_invoice_policy_required is set to false.
> 
> Meaning this modules does nothing unless you set sale_invoice_policy_required to True in the settings.
> 
> It was not the case in previous versions.

@ioans73 @jguenat In fact, this is not the expected behavior.

The initial one was :

- If no invoice policy on SO, we should use the product invoice policy
- The required parameter should avoid this (but the parameter should stay)
- If the invoice policy is set on SO level, we should use it

So, the required parameter should not avoid using the SO invoice policy.


I will make a PR to improve it.

rousseldenis:
> > > Hello @ioans73,
> > > Please make sure you are only modifying the lines whose order has `invoice_policy_required` set to True. By doing this, all tests will pass without any problem.
> > > Thank you
> > 
> > 
> > Hello @ACheung-FactorLibre
> > By doing this qty_to_invoice won't be recomputed for any sale order line if the config parameter sale_invoice_policy_required is set to false.
> > Meaning this modules does nothing unless you set sale_invoice_policy_required to True in the settings.
> > It was not the case in previous versions.
> 
> @ioans73 @jguenat In fact, this is not the expected behavior.
> 
> The initial one was :
> 
> * If no invoice policy on SO, we should use the product invoice policy
> * The required parameter should avoid this (but the parameter should stay)
> * If the invoice policy is set on SO level, we should use it
> 
> So, the required parameter should not avoid using the SO invoice policy.
> 
> I will make a PR to improve it.

https://github.com/OCA/sale-workflow/pull/3376

PR #3105, opened 2024-04-29
[17.0][MIG] sale_automatic_workflow: Migration to 17.0

trisdoan:
### This change
- This PR separates the module stock from sale_automatic_workflow


sebalix:
@trisdoan thank you. Can you rename your PR `[17.0][MIG] sale_automatic_workflow` and add the stock separation comment in the PR body instead?

sale_automatic_workflow_stock/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+
+class SaleOrder(models.Model):
+    """Extend to add stock related workflow."""
+
+    _inherit = "sale.order"
+
+    # pylint: disable=W8110
+    @api.onchange("workflow_process_id")
+    def _onchange_workflow_process_id(self):
+        """Override to add stock related workflow."""
```
sebalix: Better to use a simple comment to not replace the docstring of the original method
```
# Override to add stock related workflow
```

sale_automatic_workflow_stock/tests/common.py
```diff
@@ -0,0 +1,38 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.sale_automatic_workflow.tests.common import TestAutomaticWorkflowMixin
+
+
+class TestAutomaticWorkflowStockMixin(TestAutomaticWorkflowMixin):
+    """Extend to add stock related workflow."""
+
+    def create_sale_order(self, workflow, override=None, product_type="product"):
+        """Override to create stock operations for each product."""
```
sebalix: Same, use comment instead of docstring

sale_automatic_workflow_stock/tests/common.py
```diff
@@ -0,0 +1,38 @@
...
+                        "product_id": line.product_id.id,
+                        "location_id": self.env.ref("stock.stock_location_stock").id,
+                        "inventory_quantity": line.product_uom_qty,
+                    }
+                )
+                inventory._apply_inventory()
+        return order
+
+    def create_full_automatic(self, override=None):
+        """Override to include default stock related values."""
```
sebalix: Here too

trisdoan:
Hello @sebalix, I updated as you suggest :pray:

sebalix (approved):

sale_automatic_workflow/models/sale_order.py
```diff
@@ -44,8 +55,6 @@ def _onchange_workflow_process_id(self):
         if not self.workflow_process_id:
```
rousseldenis: @trisdoan @sebalix Maybe it's time to change the field to computed/stored/readonly=False one to remove onchanges.

UPDATE: In fact, you should override the _compute_team_id() method and to add 'workflow_process_id' in depends.

trisdoan: Hi @rousseldenis @sebalix , changes done. Please help to review :pray:

trisdoan:
DRAFT: working on failed tests

sale_automatic_workflow/models/sale_order.py
trisdoan: Hello @rousseldenis @sebalix, I updated as you suggest

I added this [commit](https://github.com/OCA/sale-workflow/pull/3105/commits/028edd74b2aafc33b7a8c6f9133242fa5c79ed30).
I adapt field `picking_policy` in sale.order to writable computed field. Unfortunately, there is a module: `sale_order_type`, which does the same thing but depends on field 'type_id' instead.
-> So one solution is to mark sale_order_type as rebel module. What do you guys think?

sebalix: Not sure it's a good idea :thinking: 
What you could do is to extract the `@api.depends` values in a method and make use of `hasattr` like it's done in the compute method:
```python
# in sale_order_type
def _depends_picking_policy(self):
    depends = []
    if hasattr(super(), "_depends_picking_policy"):
        depends = super()._depends_picking_policy()
    depends.append("type_id")
    return depends

@api.depends(lambda self: self._depends_picking_policy())
def _compute_picking_policy():
    [...]
```
```python
# in sale_automatic_workflow_stock
def _depends_picking_policy(self):
    depends = []
    if hasattr(super(), "_depends_picking_policy"):
        depends = super()._depends_picking_policy()
    depends.append("workflow_process_id")
    return depends

@api.depends(lambda self: self._depends_picking_policy())
def _compute_picking_policy():
    [...]
```
This way we are dynamically compatible with both modules installed (and probably still with an undefined behavior, but that's better than something else I guess).

sebalix: @trisdoan any news about this?

trisdoan: commit is coming today. Will ping you when it's done

trisdoan: Hello @sebalix, updated as you suggest + tests are green :pray:

rousseldenis (approved):
Code review

sale_order_type/models/sale.py
```diff
@@ -80,7 +80,14 @@ def _compute_warehouse_id(self):
                 order.warehouse_id = order_type.warehouse_id
         return res
 
-    @api.depends("type_id")
+    def _depends_picking_policy(self):
```
sebalix, line 83: Can you bump the version number of this module?  As a patch version. So we can merge the whole PR with `nobump` option.

sebalix (approved):
LG

sale_order_type/models/sale.py
trisdoan: Sure, will do

sebalix: Thanks!

sebalix (approved):

jaenbox (approved):

PR #3106, opened 2024-04-29
[17.0][MIG] sale_product_multi_add

AnaisLForgeFlow:
Standard Migration

@ForgeFlow


sale_product_multi_add/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+{
+    "name": "Sale Product Multi Add",
+    "summary": """
+        Sale Product Multi Add """,
+    "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "category": "Sale Management",
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["sale", "sale_management"],
```
MiquelRForgeFlow:
```suggestion
    "depends": ["sale_management"],
```
but I would go for not adding sale_management and left it as it was (because this module doesn't inherit/use anything from sale_management):
```suggestion
    "depends": ["sale"],
```

sale_product_multi_add/wizards/sale_import_products_view.xml
```diff
@@ -0,0 +1,54 @@
...
+                <field name="items">
+                    <tree create="0" editable="bottom">
+                        <field name="product_id" readonly="1" />
+                        <field name="quantity" />
+                    </tree>
+                </field>
+                <footer>
+                <button
+                        name="select_products"
+                        string="Select"
```
GuillemCForgeFlow:
```suggestion
                        string="Confirm"
```
I believe it is more user-friendly to see Confirm instead of select, as in the end, once you are on that view, you are confirming to add the Sales Order Lines. WDYT?

GuillemCForgeFlow (commented):
for the rest, functional and code review LGTM 👍🏿

sale_product_multi_add/wizards/sale_import_products_view.xml
AnaisLForgeFlow: Yes, I agree. Thank you for the suggestion!

MiquelRForgeFlow (approved):

GuillemCForgeFlow (approved):

sale_product_multi_add/wizards/sale_import_products_view.xml
```diff
@@ -33,7 +33,7 @@
                 <footer>
                 <button
                         name="select_products"
-                        string="Select"
+                        string="Confirm"
```
rousseldenis, line 36: Why changing this ? Maybe a text help could be valuable in this wizard form. Something like 'Please select the products you want to add to the Sale Order, then click on 'Select' ? @jbaudoux

GuillemCForgeFlow: > Why changing this ? Maybe a text help could be valuable in this wizard form. Something like 'Please select the products you want to add to the Sale Order, then click on 'Select' ? @jbaudoux

I suggested to change to text from 'Select' to 'Confirm' as IMHO it makes more sense to be like so. As we are in the last step of dynamically adding various Sales Order Lines, we are not selecting anything, we are just confirming to add one/multiple order lines.

Then, a help text with the text you are suggesting would make sense on "wizard num 1", but not on the second one where you are just assigning the quantities for each product, and then confirming to add the products to the Sale Order. The text could be useful on "wizard num 1", nevertheless, I believe it is quite intuitive already that you need to add some products, but we can discuss on that of course.

What do you think?

antoniocanovas (approved):
Functional review OK !!

pedroguirao (approved):
Functional OK!

sale_product_multi_add/wizards/sale_import_products_view.xml
jbaudoux: I'm fine with this change.
It's better to put changes in a separate commit.

jbaudoux (approved):

FrancoMaxime (approved):

PR #3108, opened 2024-04-29
[17.0][ADD] sale_product_approval

patrickrwilson:
This PR adds the sale_product_approval module to v17 - Note that this module was introduced in earlier versions but PR's were not merged.


sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+from odoo.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    exceptions_sale_approval_confirm = fields.Boolean(
+        compute="_compute_exceptions", string="Exception", default=False
+    )
+    override_exception = fields.Boolean(default=False)
```
rousseldenis: IMHO, this name is a little bit too broad. Should reflect the 'sale product approval' flow to avoid names collisions.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+
+from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    exceptions_sale_approval_confirm = fields.Boolean(
+        compute="_compute_exceptions", string="Exception", default=False
```
rousseldenis:
```suggestion
        compute="_compute_exceptions_sale_approval_confirm", string="Exception", default=False
```

sale_product_approval/models/sale_order.py
rousseldenis: No need of `default=False`as default

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+        for order in self:
+            note = self._render_product_state_excep(order, product_id)
+            order.activity_schedule(
+                "mail.mail_activity_data_warning",
+                date.today(),
+                note=note,
+                user_id=order.user_id.id or SUPERUSER_ID,
+            )
+
+    def _render_product_state_excep(self, order, product_id):
```
rousseldenis: Prefer explicit naming than abbreviations

sale_product_approval/models/sale_order_line.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    approved_sale_confirm = fields.Boolean(
```
rousseldenis, line 10: Is this really needed as stored value ?

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+                not line.approved_sale_confirm
+                for line in so.order_line.filtered(lambda x: not x.display_type)
+            )
+
+    def _log_exception_activity_sale(self, product_id):
+        for order in self:
+            note = self._render_product_state_excep(order, product_id)
+            order.activity_schedule(
+                "mail.mail_activity_data_warning",
+                date.today(),
```
rousseldenis: Is this timezone safe ?

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+        copy=False,
+        readonly=True,
+    )
+    candidate_sale = fields.Boolean(
+        string="Candidate to add to Quote",
+    )
+    candidate_sale_confirm = fields.Boolean(
+        string="Candidate to be Sold",
+    )
+    can_edit_candidate = fields.Boolean(compute="_compute_can_edit_candidate")
```
rousseldenis, line 32: This name should maybe reflect 'product approval' flow.

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _compute_can_edit_candidate(self):
+        for product in self:
+            product.can_edit_candidate = self.env.user.has_group(
```
rousseldenis: You can do without loop (as record value depends on self value):

```suggestion
            self.update({"can_edit_candidate": self.env.user.(...)})
```

sale_product_approval/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+                "sale_product_approval.group_product_administrator"
+            )
+
+    @api.onchange("candidate_sale_confirm")
+    def _onchange_candidate_sale_confirm(self):
```
rousseldenis: Change the field `candidate_sale` as compute/readonly=False/stored field to avoid onchanges.

sale_product_approval/tests/test_product_approval.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.exceptions import UserError
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderLineDates(TransactionCase):
+    def setUp(self):
```
rousseldenis: Please use classmethod instead

sale_product_approval/models/sale_order_line.py
rousseldenis: Ok, see. For the search.

rousseldenis (changes requested):

sale_product_approval/models/product_product.py
patrickrwilson: We went over the naming of these fields internally quite a bit and ended up settling on 'candidate' because the approval part is in the states and having the 'approved to be' on the products confused the users because they struggled to make the connection between the setting on the product depends on the state. So we settled on 'Candidate' to say basically its configured to be 'Sold', 'Purchased', etc as long as the state it's in allows it.

patrickrwilson:
@rousseldenis I think all your comments have been addressed and ready for a re-review, thanks for your effort.

sale_product_approval/models/sale_order.py
```diff
@@ -0,0 +1,59 @@
...
+        values = {"sale_order_ref": order, "product_ref": product_id}
+        return self.env["ir.ui.view"]._render_template(
+            template=self.env.ref("sale_product_approval.exception_on_product").id,
+            values=values,
+        )
+
+    def action_confirm(self):
+        res = super().action_confirm()
+        for sale in self:
+            if (
```
rousseldenis, line 49: @patrickrwilson You should add (either a global either a company) a configuration parameter that enable the restriction.

This is a good practice in order to not break standard behaviors.

See tests results.

rousseldenis (changes requested):

sale_product_approval/models/sale_order.py
rousseldenis: @patrickrwilson

rousseldenis:

User seems not active anymore

PR #3113, opened 2024-05-02
[15.0][IMP] sale_credit_point-update call wizard from list view

santostelmo:
Allow to call wizard from list view to update partner credit in batch,


rousseldenis (approved):
Functional review

FrancoMaxime (approved):
LGTM

PR #3118, opened 2024-05-06
[13.0][FIX] sale_order_line_packaging_qty

acsonefho:
In some case, the `product_packaging` is empty so the `product_uom_id` too and the rounding is not possible.
This fix force a default value (`2`) for the `rounding`.

### DIFF
--- a/sale_order_line_packaging_qty/models/sale_order_line.py
+++ b/sale_order_line_packaging_qty/models/sale_order_line.py
@@ -35,16 +35,19 @@ class SaleOrderLine(models.Model):
                 )
             else:
                 product_qty = sol.product_uom_qty
+            uom = (
+                sol.product_packaging.product_uom_id
+                or sol.product_uom
+                or sol.product_id.uom_id
+            )
+            rounding = uom.rounding or 2
             qty_mod = float_round(
-                product_qty % sol.product_packaging.qty,
-                precision_rounding=sol.product_packaging.product_uom_id.rounding,
+                product_qty % sol.product_packaging.qty, precision_rounding=rounding,
             )
             # After the rounding, the value could be equals to sol.product_packaging.qty
             # So just re-apply the '%'
             qty_mod = qty_mod % sol.product_packaging.qty
-            if not float_is_zero(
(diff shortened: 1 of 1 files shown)


AnizR (approved):
Code LGTM, Thanks!

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -35,16 +35,14 @@ def _compute_product_packaging_qty(self):
                 )
             else:
                 product_qty = sol.product_uom_qty
+            rounding = sol.product_packaging.product_uom_id.rounding or 2
```
rousseldenis: @acsonefho Why not using uom from the line instead and why 2?

https://github.com/odoo/odoo/blob/13.0/addons/sale/models/sale.py#L1284

```suggestion
            rounding = sol.product_packaging.product_uom_id.rounding or sol.product_uom or sol.product_id.uom_id
```

sale_order_line_packaging_qty/models/sale_order_line.py
```diff
@@ -35,16 +35,14 @@ def _compute_product_packaging_qty(self):
...
-                precision_rounding=sol.product_packaging.product_uom_id.rounding,
+                product_qty % sol.product_packaging.qty, precision_rounding=rounding,
             )
             # After the rounding, the value could be equals to sol.product_packaging.qty
             # So just re-apply the '%'
             qty_mod = qty_mod % sol.product_packaging.qty
-            if not float_is_zero(
-                qty_mod, precision_digits=sol.product_packaging.product_uom_id.rounding,
-            ):
+            if not float_is_zero(qty_mod, precision_digits=rounding):
```
rousseldenis:
```suggestion
            if not float_is_zero(qty_mod, precision_rounding=rounding):
```

rousseldenis (changes requested):
At least, change the float_is_zero() argument

FrancoMaxime (approved):
LGTM: code review

rousseldenis:
@acsonefho Could you please update your commit message first line to add some words to summarize ? Thanks

rousseldenis (approved):

PR #3127, opened 2024-05-08
[17.0][MIG] partner_contact_sale_info_propagation

Deriman-Alonso:
Standart migration.


partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -0,0 +1,86 @@
...
+        for vals in vals_list:
+            if "parent_id" in vals:
+                if "user_id" not in vals:
+                    vals.update(user_id=self.browse(vals["parent_id"]).user_id.id)
+                if "team_id" not in vals:
+                    vals.update(team_id=self.browse(vals["parent_id"]).team_id.id)
+        return super().create(vals_list)
+
+    @api.onchange("parent_id")
+    def onchange_parent_id(self):
```
rousseldenis: IMHO, this should be changed to use the compute method instead.

See: https://github.com/odoo/odoo/blob/17.0/odoo/addons/base/models/res_partner.py#L370

Deriman-Alonso: You're right, its changed now, please if possible check the changes.
Thanks for the review.

Deriman-Alonso: Yeah you're right, ill work on it.

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -25,13 +25,15 @@ def write(self, vals):
...
+                    lambda r, record=record: not r.user_id
+                    or r.user_id == record.user_id
                 )
                 if childs:
                     childs.write({"user_id": vals["user_id"]})
             if "team_id" in vals:
                 childs = record.mapped("child_ids").filtered(
-                    lambda r: not r.team_id or r.team_id == record.team_id
+                    lambda r, record=record: not r.team_id
+                    or r.team_id == record.team_id
```
FrancoMaxime: Hi @Deriman-Alonso ,
 Thanks a lot for your pull request.


Why are we keeping record.mapped("child_ids") and not using directly record.child_ids since record contains only one element. 
If I read the documentation correctly ([Doc mapped](https://www.odoo.com/documentation/17.0/developer/reference/backend/orm.html?highlight=mapped#map)) doing record.mapped(child_ids) is equivalent to record.child_ids.

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -50,17 +52,17 @@ def create(self, vals_list):
         return super().create(vals_list)
 
     @api.onchange("parent_id")
-    def onchange_parent_id(self):
+    def _compute_user_id(self):
```
FrancoMaxime: Since this [commit](https://github.com/odoo/odoo/commit/f0b7600314fee4e99de7ff48c91132f2754a9a40) user_id is a computed field, so its need an @api.depends and not an @api.onchange

FrancoMaxime (changes requested):
Hello, thanks a lot for your pull request.

There is one question and one request change. 

Great work.

partner_contact_sale_info_propagation/models/res_partner.py
Deriman-Alonso: Hi, absolutely, its changed now. Nice point and Ty for review.

partner_contact_sale_info_propagation/models/res_partner.py
Deriman-Alonso: Yeah mb there, should be good now, Ty for review 👍.

sanderlienaerts (approved):
Thanks for contributing. The migration works functionally and the code changes are fine. 

There are some commits that need to be squashed. You can follow the instructions in: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests to squash these commits.

partner_contact_sale_info_propagation/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Partner contact sale info propagation",
+    "summary": "Propagate Salesperson and Sales Channel " "from Company to Contacts",
```
FrancoMaxime, line 5:
```suggestion
    "summary": "Propagate Salesperson and Sales Channel from Company to Contacts",
```

FrancoMaxime (commented):
LGTM: minor comment.

juancarlosonate-tecnativa:
Can you squash de commits? @Deriman-Alonso  https://github.com/OCA/sale-workflow/pull/3127#pullrequestreview-2217800518

juancarlosonate-tecnativa:
Definitely, the commit history is completely incorrect, the user is also inactive, i will proceed to suppress this pr

PR #3132, opened 2024-05-10
[FIX] sale_stock_product_recommendation: wizard compatibility with sale_order_product_recommendation_packaging_default

yajo:
Without this patch, the non-deterministic order of view replacing sometimes made the wrapping of the `units_included` field done by this module wrap other fields that it shouldn't. Those fields lost their labels and the form became awkward:
image

To test this works, both `sale_stock_product_recommendation` and `sale_order_product_recommendation_packaging_default` must be installed simultaneously in the DB, and the form should appear correctly:
image

@moduon MT-6109


fcvalgar (approved):
LGTM, thank you @yajo 

image

rafaelbn (approved):
OK, great!

PR #3134, opened 2024-05-13
[15.0][IMP] product_supplierinfo_for_customer_sale: Add compatibility with Form() to change line name properly

victoralmau:
Add compatibility with Form() to change line name properly

@Tecnativa TT49073

### DIFF
--- a/product_supplierinfo_for_customer_sale/models/sale_order_line.py
+++ b/product_supplierinfo_for_customer_sale/models/sale_order_line.py
@@ -1,6 +1,8 @@
 # Copyright 2013-2017 Agile Business Group sagl
 #     (<http://www.agilebg.com>)
 # Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2024 Tecnativa - Víctor Martínez
+# Copyright 2024 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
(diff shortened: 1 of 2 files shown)


product_supplierinfo_for_customer_sale/models/sale_order_line.py
```diff
@@ -9,28 +10,40 @@
...
-                supplierinfo = line.product_id._select_customerinfo(
-                    partner=line.order_partner_id
-                )
-                code = supplierinfo.product_code
-            else:
-                code = ""
-            line.product_customer_code = code
-
-    @api.onchange("product_id")
-    def product_id_change(self):
```
pedrobaeza, line 29: Why is this not working?

victoralmau: This onchange is not a problem, but if we have already set the `customerinfo_id` field I think it is better to make the onchange with that field to avoid redoing the search.

pedrobaeza: The approach is not valid, as you are replacing any possible text, not exactly the reference we are looking for. Please revert to the previous condition.

victoralmau: Changed to another approach (and explained why)

pedrobaeza (commented):
Why do we need to search for pricelist items? I don't get that need. About your new approach, it's doing again the customerinfo selection, and also overriding any other custom behavior, so it doesn't feel OK. Why can't this be done at once using the same customerinfo and not overriding the text?

victoralmau:
Sorry, can you indicate exactly what change you want to do to solve the problem? The previous approach was incorrect and so is this one. The current implementation of all the code may look strange, although all the logic is in `product_supplierinfo_for_customer`.

pedrobaeza (approved):

pedrobaeza:
@victoralmau changed approach to be friendlier with other extensions.

PR #3136, opened 2024-05-14
[FIX] sale_elaboration: use parent.state in domain

yajo:
Without this patch, and with some unknown but feasible combination of sale-workflow modules, these steps were producing an exception:

1. Switch to mobile view.
2. Edit one sale order line.
3. Change packaging.
4. Save line (but not parent record).
5. Repeat steps 2-4.

The error was:

    modifier "readonly": Unknown field state in domain

I'm not sure why just `state` wasn't working reliably, but I can tell Odoo upstream uses `parent.state` since https://github.com/odoo/odoo/commit/a7883df9eac4bd93417846a44f185e6c2bc36eca and this fixes the issue.

@moduon MT-5351

### DIFF
--- a/sale_elaboration/views/sale_order_views.xml
+++ b/sale_elaboration/views/sale_order_views.xml
@@ -14,12 +14,12 @@
                     <field
                         name="elaboration_ids"
                         widget="many2many_tags"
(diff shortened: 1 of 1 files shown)


fcvalgar (approved):

yajo:
Using maintainer superpowers

/ocabot merge patch

PR #3139, opened 2024-05-16
[15.0][FIX] sale_planner_calendar: Do not take into account the user time zone when the planner events are created

sergio-teruel:
cc @Tecnativa TT49169


CarlosRoca13 (approved):
LGTM :+1:

pedrobaeza (approved):
/ocabot merge patch

PR #3150, opened 2024-05-25
[17.0][FIX] sale_fixed_discount: SOL discount propagation to invoice line

astirpe:
**Context:**
when creating or editing a sale order line by using the UI, the `_onchange_discount_fixed()` ensures that the `discount` field is computed based on the value of `discount_fixed` and `price_unit` fields. Keeping the `discount` updated is important because this value will be propagated to the invoice line at the moment the invoice is generated.

**Description of the issue:**
The onchange method is not triggered in case we create or modify the sale order line by using an API (and so without using the UI) or when we update `discount_fixed` and `price_unit` field in custom code.
This means that the wrong `discount` value (that stays `0.0`) will be propagated to the invoice line at the moment the invoice is generated.

**Proposed fix:**
with this PR the onchange method is replaced by extending the `_compute_discount()` method, ensuring that the `discount` field is consistent with `discount_fixed`

### DIFF
--- a/sale_fixed_discount/models/sale_order_line.py
+++ b/sale_fixed_discount/models/sale_order_line.py
@@ -78,9 +78,12 @@ class SaleOrderLine(models.Model):
 
         return super()._convert_to_tax_base_line_dict()
 
-    @api.onchange("discount_fixed", "price_unit")
-    def _onchange_discount_fixed(self):
-        self.discount = self._get_discount_from_fixed_discount()
+    @api.depends("discount_fixed", "price_unit")
+    def _compute_discount(self):
+        lines_with_discount_fixed = self.filtered(lambda sol: sol.discount_fixed)
+        for line in lines_with_discount_fixed:
+            line.discount = line._get_discount_from_fixed_discount()
+        return super(SaleOrderLine, self - lines_with_discount_fixed)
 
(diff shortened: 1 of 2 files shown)


TimGlimmerveen-360ERP (approved):

CasVissers-360ERP (approved):
Functional review

astirpe:
Can this one be merged?

sale_fixed_discount/models/sale_order_line.py
```diff
@@ -78,9 +78,14 @@ def _convert_to_tax_base_line_dict(self, **kwargs):
 
         return super()._convert_to_tax_base_line_dict()
 
-    @api.onchange("discount_fixed", "price_unit")
-    def _onchange_discount_fixed(self):
-        self.discount = self._get_discount_from_fixed_discount()
+    @api.depends(
+        "product_id", "product_uom", "product_uom_qty", "discount_fixed", "price_unit"
```
rousseldenis:
```suggestion
       "discount_fixed", "price_unit"
```

ORM will append new fields to existing depends.

astirpe: Right, thanks!

rousseldenis (approved):

PR #3153, opened 2024-05-27
[IMP]Improved code to avoid error when there is no company or product 

jupical-oca:
Improved code.

### DIFF
--- a/sale_product_set/models/product_set_line.py
+++ b/sale_product_set/models/product_set_line.py
@@ -5,6 +5,7 @@ from odoo import fields, models
 
 
 class ProductSetLine(models.Model):
+
     _inherit = "product.set.line"
 
     discount = fields.Float(string="Discount (%)", digits="Discount", default=0.0)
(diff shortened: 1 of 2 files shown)


sale_product_set/models/product_set_line.py
```diff
@@ -15,8 +16,8 @@ def prepare_sale_order_line_values(self, order, quantity, max_sequence=0):
             "order_id": order.id,
             "product_id": self.product_id.id,
             "product_uom_qty": self.quantity * quantity,
-            "product_uom": self.product_id.uom_id.id,
+            "product_uom": self.product_id and self.product_id and self.product_id.uom_id and self.product_id.uom_id.id or False,
```
rousseldenis, line 19: You can simplify by:

```suggestion
            "product_uom": self.product_id.uom_id and self.product_id.uom_id.id or False,
```

rousseldenis: Even if self.product_id.uom_id.id will be False if any relation is not defined. So, I'm wondering in which case you'll have an error

rousseldenis:
@jupical-oca

rousseldenis:
Closing this as no answer and user is no more active on github.

PR #3155, opened 2024-05-27
[16.0] [FIX] sale_order_line_date - bug if commitment_date is reset to False

BenjaHe:
Bug if a user remove the commitment date (from the sale order line) after the order confirmation (and a stock.picking is generated)


StefanRijnhart:
Can you reproduce the issue with a unit test?

PR #3156, opened 2024-05-27
[17.0][MIG] sale_advance_payment: Migration to 17.0

Abranes:
Migration to 17.0


Abranes:
I've reviewed the tests and it seems the error is because the currency conversion is not applied correctly. If anyone can provide some insight, otherwise I'll review it again when I get a chance.

vvrossem:
Please Squash administrative commits with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

sale_advance_payment/models/payment.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    sale_id = fields.Many2one(
+        "sale.order", "Sale", readonly=True, states={"draft": [("readonly", False)]}
+    )
```
vvrossem, line 12: Since Odoo 17, property `states` is no longer supported.
You should define the `readonly` attribute on the related views instead.
something of the like of `<field name="sale_id" readonly="state != 'draft'">`

sale_advance_payment/wizard/sale_advance_payment_wzd.py
```diff
@@ -99,24 +103,26 @@ def default_get(self, fields_list):
                 {
                     "order_id": sale.id,
                     "amount_total": sale.amount_residual,
-                    "currency_id": sale.pricelist_id.currency_id.id,
+                    "currency_id": sale.pricelist_id.currency_id.id
+                    or sale.currency_id.id,
```
vvrossem, line 107: https://github.com/odoo/odoo/blob/17.0/addons/sale/models/sale_order.py#L392-L401
The pricelist is not set anymore on draft SO. 
The Travis issue might come from this. 
I've ran the tests on the module in its version 16.0, and the currency_id doesn't match with the one of 17.0

vvrossem (changes requested):

vvrossem:
Have a look at [[17.0][MIG] purchase advance payment](https://github.com/OCA/purchase-workflow/pull/2234), it might be useful on what changes to do on the unit tests

Abranes:
Thank you for the comments @vvrossem. I'm a bit busy right now, but I will review everything and make the necessary changes by the end of the week.

yankinmax (approved):

rousseldenis:
@Abranes

pandoc-3.2-1-amd64.deb
Camille0907, line 1: This file should be removed, cf. https://github.com/OCA/oca-addons-repo-template/issues/262

sale_advance_payment/views/sale_view.xml
```diff
@@ -11,7 +11,7 @@
                     string="Pay sale advanced"
                     type="action"
                     groups="account.group_account_invoice"
-                    attrs="{'invisible': [('state', 'in', ['done','cancel'])]}"
+                    invisible="state == 'cancel'"
```
Camille0907, line 14: We don't want it anymore to be invisible when the state is done ? It does not look like a migration action

Camille0907 (changes requested):
Please remove the `.deb` file. Except that, migration looks technically good to me

gurneyalex (changes requested):
Don't merge the PR while the debian file is in there.

yankinmax:
This PR will be superseded by:
- [ ] https://github.com/OCA/sale-workflow/pull/3244

gurneyalex:
superseeded by https://github.com/OCA/sale-workflow/pull/3244

PR #3159, opened 2024-05-28
[FIX] sale_order_type: Fix when changing the currency in a customer payment

rov-adhoc:
When you change the currency of a customer payment, the currency of the payment gets changed to the company's set currency. 

For example:
When you are changing from ARS to EUR and the company's currency is set to USD, the payment currency changes to USD.

I attach a video to represent the error.
https://drive.google.com/file/d/1xsZmIiFM5-bOAV8oFZw_vieeUTe0p7oS/view

### DIFF
--- a/sale_order_type/models/__init__.py
+++ b/sale_order_type/models/__init__.py
@@ -5,3 +5,4 @@ from . import sale
 from . import res_partner
 from . import account_move
 from . import res_currency
+from . import account_payment
--- a/sale_order_type/models/account_payment.py
+++ b/sale_order_type/models/account_payment.py
@@ -1,0 +1,14 @@
+from odoo import api, models
+
+
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    @api.depends("journal_id")
+    def _compute_currency_id(self):
+        for rec in self:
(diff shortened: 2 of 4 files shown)


rousseldenis:
@rov-adhoc I don't catch why it should be integrated into sale_order_type module...

rov-adhoc:
@rousseldenis because is this module who is generating the error in the customer payments.

I attach a video to represent the error.
https://drive.google.com/file/d/1xsZmIiFM5-bOAV8oFZw_vieeUTe0p7oS/view

Regards

rousseldenis:
Yes I understand, but the code you introduce does not depends at all on sale_order_type one...

You should at least show the problem in a test case.

rov-adhoc:
@rousseldenis  I just wanted to let you know that I already added the test case you mentioned in your previous comment.
Thanks in advance!
Regards

nicolascol (approved):
Functional review OK

cav-adhoc (approved):
Technical review OK

rov-adhoc:
@rousseldenis Hi! I Would like to ask you if you can help me with the merge of this PR.
Thanks in advance

PR #3160, opened 2024-05-29
[17.0] [MIG] Sale discount display amount

FSchiepers:
This version adds the possibility to choose whether the discount is calculated based on the non-taxed subtotal or the taxed total, see: https://github.com/OCA/sale-workflow/issues/2813

Related to https://github.com/OCA/sale-workflow/issues/2766


DorianMAG (approved):
Tested in ruboat.

LGTM thx

rousseldenis (approved):

PR #3163, opened 2024-05-31
[16.0][IMP] product_supplierinfo_for_customer_sale: Add compatibility with Form() to change line name properly

victoralmau:
FWP from 15.0: https://github.com/OCA/sale-workflow/pull/3134 + https://github.com/OCA/sale-workflow/pull/3158

Add compatibility with Form() to change line name properly

@Tecnativa TT49073

### DIFF
--- a/product_supplierinfo_for_customer_sale/models/sale_order_line.py
+++ b/product_supplierinfo_for_customer_sale/models/sale_order_line.py
@@ -33,12 +33,11 @@ class SaleOrderLine(models.Model):
         This also takes from context the possible customerinfo already searched in
         product_id_change for avoiding duplicated searches.
         """
-        if "customerinfo" in self.env.context:
-            customerinfo = self.env.context["customerinfo"]
-        else:
-            customerinfo = self.product_id._select_customerinfo(
-                partner=self.order_partner_id
-            )
+        # We need to repeat the search here, as passing the value by context or any
+        # other trick makes the ORM to do ugly things in "onchange" mode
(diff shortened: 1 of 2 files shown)


product_supplierinfo_for_customer_sale/models/sale_order_line.py
```diff
@@ -25,38 +27,30 @@ def _compute_product_customer_code(self):
                 code = ""
             line.product_customer_code = code
 
+    def _compute_name(self):
+        """Sobrescribimos el compute para re-definir el name."""
```
pedrobaeza: This comment should be in English and be more descriptive. What you are stating here is the what, which is clear - a method call `_compute_name` does that, to compute the name -, not the why.

pedrobaeza (changes requested):
Please put 2 commits, one with the changes as they were in v1, and another with the v16 adaptations.

pedrobaeza:
The thumb rule for squashing in the same commit or not is if the changes needed in the new version are relevant enough to be kept separated. In this case, that the methods have totally changed in this new version, they are.

product_supplierinfo_for_customer_sale/models/sale_order_line.py
victoralmau: Sorry, the comment has been changed.

victoralmau:
Changes done (v15 commits + v16 commit)

pedrobaeza (approved):
Merging as done in previous version:

/ocabot merge patch

PR #3168, opened 2024-06-03
[16.0][IMP-FIX] sale_payment_sheet: Archive all reconciliation models to avoid them interfering with the tests

pilarvargas-tecnativa:
When in an instance this module coincides with the bank reconciliation module account_reconcile_oca it causes the tests to fail as the statement lines are automatically reconciled. For this reason it is necessary to archive all records of the reconciliation models.

cc @tecnativa TT44331

@carolinafernandez-tecnativa @victoralmau please review

ping @pedrobaeza

### DIFF
--- a/sale_payment_sheet/tests/test_sale_payment_sheet.py
+++ b/sale_payment_sheet/tests/test_sale_payment_sheet.py
@@ -16,6 +16,8 @@ class TestSaleInvoicePayment(TransactionCase):
         super().setUpClass()
         # Remove time zone from user to avoid to time local representation
         cls.env.user.partner_id.tz = False
+        # Archive all reconciliation models to avoid them interfering with the tests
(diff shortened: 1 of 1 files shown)


carolinafernandez-tecnativa (approved):

pedrobaeza:
According to your explanation, then the problem is on the other module (`account_reconcile_oca`), not here. Anyway, one thing you can do that is cleaner than this is to archive all the reconciliation models on `setUpClass`.

pilarvargas-tecnativa:
> According to your explanation, then the problem is on the other module (`account_reconcile_oca`), not here. Anyway, one thing you can do that is cleaner than this is to archive all the reconciliation models on `setUpClass`.

I have archived the records in the test, thanks.

pedrobaeza (approved):
/ocabot merge patch

PR #3169, opened 2024-06-04
[FIX] product_supplierinfo_for_customer_sale: fix tests

xmglord:
In #3163 the test where changed and try to set a field restricted to a group that the user used in the test don't have.


xmglord:
@pedrobaeza please review.

cc: @desdelinux @luisg123v

pedrobaeza (approved):
OK, we can fast-track this:

/ocabot merge patch

PR #3175, opened 2024-06-05
[16.0][ADD] sale_order_recurrence

quentinDupont:
This module provides a wizard that facilitates the duplication of sales with a certain recurrence.

User defines a begin date and recurrence (week, monthly) and the wizard creates draft quotations with the dates selected in the wizard

It's useful for users who regularly sell the same products to the same customer, such as bakers who receive the same orders from shops every week.

It's a custom module that we migrate from 12.0 to 16.0. We use it since several years in production

sale_order_duplication_wizard_form

quotations_created_with_wizard


legalsylvain (approved):
LGTM. 
Thanks !

duplication_sale_order/README.rst
```diff
@@ -0,0 +1,99 @@
...
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module provide a wizard to duplicate a given Sale order.
+
+User define:
+
+* a begin date
+* a frequency (week, monthly)
```
rousseldenis: @quentinDupont @legalsylvain IMHO, you should explain a little bit more this one. As I understand (after having read the code), the wizard will create several quotations based on dates.

quentinDupont: thanks for your review ! what do you think about the changes on this last commit ?

rousseldenis:
@quentinDupont Great! My last question is from a functional point of view. Could you explain briefly in which context this is necessary and in which manner this differs from contract flow (that could generates sales or invoices) ? Thanks

quentinDupont:
> @quentinDupont Great! My last question is from a functional point of view. Could you explain briefly in which context this is necessary and in which manner this differs from contract flow (that could generates sales or invoices) ? Thanks

Hello and thanks for your time !
I just tested sale_contract_generation which is very great !

However, I think this module fits better "my" users need.
They need to be able to anticipate their sales for the month in the software, and to see them in concrete terms.
The fact that 'contract' works via cron is very 'clean' but can also be disconcerting for the user.
Having said that, surely we could make "contract" allow you to create all the sales at once? 

But I think we're still looking for a simple way of creating sales in advance, and that the solution I'm proposing in this "PR" is more simple and effective than adding a new contract concept for the user.

To answer your question about the process more directly. A baker knows that he's going to sell bread to a shop once a week during the month: he wants to anticipate his administrative work by creating the 4 sales of the month so that he can easily generate the stock pickings each week.
The following month, he doesn't sell to the shop for a couple of weeks for one reason or another, so there are no sales. But he starts again the following week for 4 weeks and so on.

quentinDupont:
ping @rousseldenis :-)

rousseldenis:
> ping @rousseldenis :-)

Yes. Busy on other things.

FYI, there is also another way to do those flows:

- https://github.com/OCA/sale-workflow/pull/3436

Could you check also ?


I want to avoid too much modules with same functionalities.

IMHO, `duplication` is quite confusing.

Maybe a better name could be great like `sale_order_recurrence` or ...

quentinDupont:
> > ping @rousseldenis :-)
> 
> Yes. Busy on other things.
> 
> FYI, there is also another way to do those flows:
> 
>     * [[16.0][Add] sale_order_blanket_order #3436](https://github.com/OCA/sale-workflow/pull/3436)
> 
> 
> Could you check also ?
> 
> I want to avoid too much modules with same functionalities.
> 
> IMHO, `duplication` is quite confusing.
> 
> Maybe a better name could be great like `sale_order_recurrence` or ...

Hey thanks for sharing this module that i didn't know !
After checking, reading and testing, it seems that blanket_order doesn't fit my module's needs : easily duplicate sale_order with recurrence

I think the name `sale_order_recurrence` is a great idea :+1:

quentinDupont:
Not related issue in runboat  ?

`odoo.exceptions.UserError: Modules "Account Fixed Discount" and "Account Invoice Triple Discount" are incompatible.`

rousseldenis:
> Not related issue in runboat ?
> 
> `odoo.exceptions.UserError: Modules "Account Fixed Discount" and "Account Invoice Triple Discount" are incompatible.`

@quentinDupont Those have been fixed. Could you rebase ?

Jaimermaccione (approved):
LGTM! Functional review on runboat.

quentinDupont:
Hello @OCA/crm-sales-marketing-maintainers, this PR has bee approved and ready to merge :)

rousseldenis:
@quentinDupont Is it normal there are two modules here ?

quentinDupont:
> @quentinDupont Is it normal there are two modules here ?


Shit no :laughing:

quentinDupont:
Duplicate module removed and commits squashed @rousseldenis thanks !

sale_order_recurrence/models/sale_order_recurrence_wizard.py
```diff
@@ -0,0 +1,126 @@
...
+        inverse_name="wizard_id",
+        string="New Dates",
+    )
+
+    @api.model
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        order_id = self.env.context.get("active_id", False)
+        if not order_id:
+            res.update({"order_id": False})
```
rousseldenis, line 60: FYI, this is more efficient:

```suggestion
            res.update({
                "order_id": False,
                "partner_id": False,
                "begin_date": False,
            })
```

rousseldenis (approved):

PR #3177, opened 2024-06-11
[16.0][FIX] sale_order_qty_change_no_recompute: Avoid error in tests if pricelist_id field is invisible

victoralmau:
Avoid error in tests if pricelist_id field is invisible

`AssertionError: can't write on invisible field pricelist_id`

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #3063, opened 2024-04-09
[FIX] sale_order_lot_selection: remove deprecated method for domain

TheMule71:
`_onchange_product_id_set_lot_domain()` causes a warning as 'domain' is deprecated as a return value, and also, doesn't work as intented, the domain is specified in the view. In short you can't customize it by overriding this method anyway, as its output is ignored.

So we take the other approach, as describe here:
https://github.com/odoo/odoo/pull/41918#issuecomment-824946980

Tests have been modified accordingly.

PR #3179, opened 2024-06-12
[14.0] sale_order_lot_selection: fix perf issue

sebastienbeau:
Following PR : https://github.com/OCA/sale-workflow/pull/3063 have replaced the onchange by a compute method
The issue is that the implementation return in the variable "allowed_lot_ids" all the lot matching for the product.
When having huge database it generate performance issue (you can have million of lot).

As the domain was already defined in the view, the onchange was just useless (it was dead code from previous migration and refactor).

So I think the best is to just use a domain in the view and nothing more.

This is a breaking change (as it change the API) so before merging it, it need approve of reviewer of the PR that have introduced the change.

@TheMule71 @rousseldenis @bodedra  @sergiocorato  @simahawk 

Thanks for your feedback.

Note: In version 16 the implementation just use a domain and this is working perfectly.

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -17,7 +17,3 @@ class SaleOrderLine(models.Model):
         res = super().product_id_change()
         self.lot_id = False
         return res
-
-    @api.onchange("product_id")
-    def _onchange_product_id_set_lot_domain(self):
-        return {"domain": {"lot_id": [("product_id", "=", self.product_id.id)]}}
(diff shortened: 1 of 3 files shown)


bodedra (approved):
LGTM :+1: 

Thanks @sebastienbeau for take care of it. Appreciate your time!

sale_order_lot_selection module is a stable and many people uses into their Production database. This module should avoid any new field introduction and restriction.

Kev-Roche (approved):
LGTM

rousseldenis (approved):
LGTM

PR #3180, opened 2024-06-12
[14.0] Add module sale_order_split_strategy

grindtildeath:
This module provides the possibility to define split strategies to be 
applied on sales order for splitting.


sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,109 @@
...
+        return {"order_line": False}
+
+    def _prepare_line_move_defaults(self):
+        """Hook to customize values used on new sale order line from split"""
+        self.ensure_one()
+        return {"order_id": self.id}
+
+    def _select_lines_to_split(self):
+        self.ensure_one()
+        line_filter = self.split_strategy_id.line_filter_id
```
mmequignon: shouldn't we append domain with section and note rules depending on `split strategy` configuration ?
The search would return everything relevant, and you wouldn't have to filter in the huge method above.
I would create a `split.strategy._eval_domain()` method to use here.

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,109 @@
...
+from odoo.osv.expression import AND
+from odoo.tools.safe_eval import safe_eval
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    split_strategy_id = fields.Many2one("sale.order.split.strategy")
+
+    def action_split(self):
```
mmequignon: With the right domain coming from the strategy, i'd say we should easily retrieve lines to split.
```python3
lines_to_split = self.order_line.search(self.strategy.eval_domain + ("order_id", "=", self.id)
# return if no line to split
lines_to_keep = self.order_line - lines_to_split
new_order = self.copy(order_line=lines_to_split)
self.order_line = lines_to_keep
```
Of course I probably missed things, but I believe this could be simplified.
This method is very long.

sale_order_split_strategy/models/sale_order.py
grindtildeath: I'm not sure how you could define if a note or a section is to be split except according to their sequences?

sale_order_split_strategy/models/sale_order.py
grindtildeath: The issue is handling the sections and notes. 

A line is part of a section if a section is defined _before_ said line, and to define if the section must be kept, be copied, or moved entirely, it depends on the lines defined _after_ the section... Therefore I don't see any other option than building the OrderedDict based on the selected lines for split.

As for notes, it's not easy to define what they relates to (if it relates to another line, a section, or the sale order). That's why I left a FIXME at the moment.

EDIT: BTW I agree the method is very long and still needs to be split in parts we can override if we want to implement more concepts on the strategy.

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+        orders_without_split = self.filtered(lambda o: not o.split_strategy_id)
+        if orders_without_split:
+            raise UserError(
+                _("Cannot split orders %s without any split strategy defined")
```
rousseldenis:
```suggestion
                _("Cannot split orders %(order_names)s without any split strategy defined", order_names=", ".join(orders_without_split.mapped("name") )
```

is better

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+                % ", ".join(orders_without_split.mapped("name"))
+            )
+        new_order_ids = []
+        for order in self:
+            copy_sections = order.split_strategy_id.copy_sections
+            copy_notes = order.split_strategy_id.copy_notes
+            lines_to_split = order.split_strategy_id._select_lines_to_split(order)
+            if not lines_to_split:
+                raise UserError(
+                    _(
```
rousseldenis: Same as above

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+                        "Cannot split order %s according to its strategy because there are no matching lines"
+                    )
+                    % order.name
+                )
+            new_order = order.copy(self._prepare_order_copy_defaults())
+            # Prepare an OrderedDict from sale order lines and their sections
+            sections_dict = OrderedDict()
+            section_line_ids = None
+            for line in order.order_line.sorted():
```
rousseldenis:
```suggestion
            for line in order.order_line:
```
As no key has been added, so default sort order

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+                raise UserError(
+                    _(
+                        "Cannot split order %s according to its strategy because there are no matching lines"
+                    )
+                    % order.name
+                )
+            new_order = order.copy(self._prepare_order_copy_defaults())
+            # Prepare an OrderedDict from sale order lines and their sections
+            sections_dict = OrderedDict()
+            section_line_ids = None
```
rousseldenis:
```suggestion
            section_line_ids = defaultdict(lambda: self.env["sale.order.line].browse)
```

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+            for line in order.order_line.sorted():
+                if line.display_type == "line_section":
+                    section_line_ids = sections_dict.setdefault(line.id, [])
+                else:
+                    if section_line_ids is None:
+                        section_line_ids = sections_dict.setdefault(False, [])
```
rousseldenis: Using defaultdict will avoid this

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def _prepare_message_split_from(self, order):
+        self.ensure_one()
+        template = self.env["ir.ui.view"].search(
+            [("key", "=", "sale_order_split_strategy.split_from")], limit=1
+        )
+        if not template:
+            return _(
+                "This sale order was created after splitting lines from %s using strategy %s"
+            ) % (order.name, order.split_strategy_id.name)
```
rousseldenis: Same as above

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def _prepare_message_split_to(self, order):
+        self.ensure_one()
+        template = self.env["ir.ui.view"].search(
+            [("key", "=", "sale_order_split_strategy.split_to")], limit=1
+        )
+        if not template:
+            return _(
+                "This sale order had some of its lines split to %s using strategy %s"
+            ) % (order.name, self.split_strategy_id.name)
```
rousseldenis: Same as above

rousseldenis (changes requested):
Code review

sale_order_split_strategy/models/sale_order.py
grindtildeath: We still need to make sure order lines are sorted using the default order, what is not necessarily the case with a recordset.

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,147 @@
...
+            )
+        else:
+            return template._render(values={"from_order": self, "to_order": order})
+
+    def _prepare_order_copy_defaults(self):
+        """Hook to customize values used on new sale order from split"""
+        self.ensure_one()
+        return {"order_line": False}
+
+    def _prepare_line_move_defaults(self):
```
sebalix: No risk to introduce conflicts with method names, depending on the set of modules installed? I would have named it `_prepare_line_split_defaults` perhaps

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,147 @@
...
+            return _(
+                "This sale order had some of its lines split to %(order_name)s"
+                " using strategy %(strategy)s",
+                order_name=order.name,
+                strategy=self.split_strategy_id.name,
+            )
+        else:
+            return template._render(values={"from_order": self, "to_order": order})
+
+    def _prepare_order_copy_defaults(self):
```
sebalix:
```suggestion
    def _prepare_order_split_copy_defaults(self):
```

sale_order_split_strategy/models/sale_order.py
rousseldenis: What do you mean by default order ? A recordset is respecting _order property by default.

grindtildeath: Not necessarily :wink: 
```
>>> order = env["sale.order"].create({"partner_id": env["res.partner"].search([], limit=1).id})
>>> order
sale.order(27,)
>>> order.order_line
sale.order.line()
>>> product = env["product.product"].search([("sale_ok", "=", True)], limit=1)
>>> order.order_line.create({"order_id": order.id, "product_id": product.id, "product_uom_qty": 1})
sale.order.line(65,)
>>> order.order_line
sale.order.line(65,)
>>> order.order_line.sequence
10
>>> order.order_line.create({"order_id": order.id, "product_id": product.id, "product_uom_qty": 1, "sequence": 20})
sale.order.line(66,)
>>> order.order_line
sale.order.line(65, 66)
>>> order.order_line.create({"order_id": order.id, "product_id": product.id, "product_uom_qty": 1, "sequence": 5})
sale.order.line(67,)
>>> order.order_line
sale.order.line(65, 66, 67)
>>> order.order_line.sorted()
sale.order.line(67, 65, 66)
```

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,151 @@
...
+        for line in self.order_line.sorted():
+            if line.display_type == "line_section":
+                section_id = line.id
+            sections_dict[section_id].append(line.id)
+        return sections_dict
+
+    def _prepare_message_split_from(self, order):
+        self.ensure_one()
+        template = self.env["ir.ui.view"].search(
+            [("key", "=", "sale_order_split_strategy.split_from")], limit=1
```
simahawk: why not `self.env.ref(..., raise_if_not_found=False)`?

sale_order_split_strategy/views/mail_message_template.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="split_from">
+        <t t-name="sale_order_split_strategy.split_from">
```
simahawk: there's no need to define `t-name`

sale_order_split_strategy_delivery/models/sale_order.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _has_only_lines_to_split(self, lines_to_split):
+        return self.order_line.filtered(lambda l: not l.is_delivery) == lines_to_split
```
simahawk: IMO these lines can be excluded beforehand w/o this module since the method `_is_delivery` is available in `sale` module.

sale_order_split_strategy_delivery/models/sale_order_split_strategy.py
```diff
@@ -0,0 +1,12 @@
...
+from odoo import models
+from odoo.osv.expression import AND
+
+
+class SaleOrderSplitStrategy(models.Model):
+    _inherit = "sale.order.split.strategy"
+
+    def _get_lines_to_split_domain(self, orders):
+        domain = super()._get_lines_to_split_domain(orders)
+        return AND([domain, [("is_delivery", "=", False)]])
```
simahawk: these lines should be excluded a priori, no need for this module IMO

sale_order_split_strategy/models/sale_order_split_strategy.py
```diff
@@ -0,0 +1,30 @@
...
+        domain="[('model_id', '=', 'sale.order.line')]",
+        required=True,
+    )
+    copy_sections = fields.Boolean()
+    copy_notes = fields.Boolean()
+
+    def _select_lines_to_split(self, orders):
+        self.ensure_one()
+        domain = self._get_lines_to_split_domain(orders)
+        return self.env["sale.order.line"].search(domain)
```
simahawk: just filter here delivery lines w/ `not line._is_delivery()`

sale_order_split_strategy/tests/test_split_strategy.py
```diff
@@ -0,0 +1,124 @@
...
+            }
+        )
+        order_lines = order.order_line.sorted()
+        self.assertEqual(order_lines[0], first_section)
+        self.assertEqual(order_lines[2], middle_section)
+        self.assertEqual(order_lines[5], last_section)
+        self.product_type_not_service_strategy.copy_sections = True
+        order.split_strategy_id = self.product_type_not_service_strategy
+        new_order = order.action_split()
+        self.assertEqual(len(new_order.order_line), 4)
```
simahawk, line 118: not sure to understand what are you testing here... I guess this part has to be completed.

sale_order_split_strategy/models/sale_order.py
grindtildeath: IMO it's safer to get the record this way :shrug:

simahawk: why?

setup/sale_order_split_strategy_delivery/odoo/addons/sale_order_split_strategy_delivery
```diff
@@ -0,0 +1 @@
+../../../../sale_order_split_strategy_delivery
```
simahawk: left over :wink:

sale_order_split_strategy/models/sale_order_split_strategy.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+from odoo.osv.expression import AND
+from odoo.tools.safe_eval import safe_eval
+
+
+class SaleOrderSplitStrategy(models.Model):
+    _name = "sale.order.split.strategy"
+    _description = "Order split strategy"
```
simahawk, line 10:
```suggestion
    _description = "Sale order split strategy"
```

sale_order_split_strategy/models/sale_order.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            return template._render(values={"from_order": order})
+
+    def _prepare_message_split_to(self, order):
+        self.ensure_one()
+        template = self.env["ir.ui.view"].search(
+            [("key", "=", "sale_order_split_strategy.split_to")], limit=1
+        )
+        if not template:
+            return _(
```
simahawk:
```suggestion
        if template:
            return template._render(values={"from_order": self, "to_order": order})
        return _(...)
```

simahawk (approved):
Minor remarks, LG otherwise. Pre-approving :)

sale_order_split_strategy/models/sale_order.py
simahawk: As it seems linked to caching I'd rather invalidate the cache of `order_line` instead.

sale_order_split_strategy/models/sale_order.py
grindtildeath: Actually we need to use `_render_template` because standard user don't have access to `ir.ui.view` to search

mmequignon (approved):
LGTM, can you please rebase?

PR #3192, opened 2024-06-14
IMP sale_recalculation: allow to recompute several orders

bealdav:
It  adds an action to update prices on several sale orders.

@pedrobaeza @cubells @CRogos

### DIFF
--- a/sale_order_price_recalculation/action/action.xml
+++ b/sale_order_price_recalculation/action/action.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="recalculate_prices_act_serv" model="ir.actions.server">
+        <field name="name">Recalculate Prices</field>
+        <field name="model_id" ref="model_sale_order" />
+        <field name="binding_model_id" ref="model_sale_order" />
+        <field name="state">code</field>
+        <!-- We already have a button, then no need to also add action on form view -->
+        <field name="binding_view_types">list</field>
+        <field name="code">records._recalculate_prices()</field>
+    </record>
+
+</odoo>
--- a/sale_order_price_recalculation/models/sale_order.py
+++ b/sale_order_price_recalculation/models/sale_order.py
@@ -5,6 +5,7 @@
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # Copyright 2023 glueckkanja AG - CRogos
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
 from odoo import models
 
 
@@ -15,3 +16,7 @@ class SaleOrder(models.Model):
         lines = self.mapped("order_line")
         lines._compute_name()
         return True
+
+    def _recalculate_prices(self):
+        for rec in self:
+            rec.action_update_prices()
--- a/sale_order_price_recalculation/README.rst
+++ b/sale_order_price_recalculation/README.rst
@@ -37,6 +37,8 @@ It is launched manually as a button to get the user decide if he/she wants to
 recalculate prices when pricelist is changed or after duplicating a sale order
 to update or not sales information.
 
+It also adds an action to update prices on several sale orders.
+
 **Table of contents**
 
 .. contents::
@@ -58,6 +60,8 @@ custom descriptions.
 
 .. image:: https://raw.githubusercontent.com/sale_order_price_recalculation/static/description/sale_order_price_recalculation.png
     :alt: Sale order price recalculation
+
+You may also select several orders to recompute prices at once.
 
 Known issues / Roadmap
 ======================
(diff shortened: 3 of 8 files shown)


sale_order_price_recalculation/action/action.xml
```diff
@@ -0,0 +1,14 @@
...
+<odoo>
+
+    <record id="recalculate_prices_act_serv" model="ir.actions.server">
+        <field name="name">Recalculate Prices</field>
+        <field name="model_id" ref="model_sale_order" />
+        <field name="binding_model_id" ref="model_sale_order" />
+        <field name="state">code</field>
+        <!-- We already have a button, then no need to also add action on form view -->
+        <field name="binding_view_types">list</field>
+        <field name="code">env["sale.order"]._recalculate_prices()</field>
```
rousseldenis:
```suggestion
        <field name="code">records._recalculate_prices()</field>
```

is more correct.

@bealdav

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -5,8 +5,12 @@
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # Copyright 2023 glueckkanja AG - CRogos
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+import logging
+
 from odoo import models
 
+logger = logging.getLogger(__name__)
```
CRogos: _logger instead of logger is a lot more common.

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -15,3 +19,11 @@ def action_update_names(self):
         lines = self.mapped("order_line")
         lines._compute_name()
         return True
+
+    def _recalculate_prices(self):
+        self = self.browse(self._context.get("active_ids"))
```
CRogos: I am not familiar with this line, but I think the last change "records._recalculate_prices()" makes this obsolete?

CRogos: The name action_update_prices is already used, but would the method name action_recompute_prices better?

rousseldenis: @bealdav _context is deprecated since long time ago : https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L5160

bealdav: Yes deprecated since 10.0, 11.0 ? but yet here, arf !
Yes I can remove

bealdav: about name, the module is named recalculation, but I can change

bealdav: in this case it's intentional to prefix method name by `_` to not allow to call via xml-rpc. Are you ok ?
Then `action_` convention  is for method which are accessible by xml-rpc isn't it ?

bealdav: Then `_recalculate_prices` and `_recompute_prices` are candidates but the last one could be collide other methods name in other modules ? re-compute is a common word used, isn't it ?

 `_recalculate_prices` seems better

CRogos (approved):

sale_order_price_recalculation/models/sale_order.py
CRogos: You are right, that action_ can be called by the API, but I do not see any problem to call this method because you could also call action_update_prices() in a loop from the API.

And **action_update_prices** calls **_recompute_prices** and I thought having **_recalculate_prices** and **_recompute_prices**  is also confising. 

When **action_recompute_prices** calls **_recompute_prices** via **action_update_prices** it's a little more consistant.

I think the best name would be **action_update_prices** but this is already used. Maybe **action_update_prices_list**

image

But I agree there is a lot personal taste.

sale_order_price_recalculation/models/sale_order.py
```diff
@@ -15,3 +19,10 @@ def action_update_names(self):
         lines = self.mapped("order_line")
         lines._compute_name()
         return True
+
+    def _recalculate_prices(self):
+        _logger.info(
```
rousseldenis: Do we really want to maybe have a big stack in normal logs ? Maybe debug should be sufficent no ?

bealdav: I think it'll not a big stack because it's only one log triggered by server action.
On my side I never use debug log level because it's too much noise. If you think this log is too noisy, it might  better to remove it completely.
The use case of recalculation, is an exceptional operation, don't you think ?

Here is my opinion, but I'm Ok to apply your final choice cc @CRogos

CRogos: I am fine with removing it. If someone is interested in tracking the changes caused by this action, it could install tracking_manager and track the changes on the fields.

bealdav: Thanks for your quick reply, fixed

rousseldenis (approved):
Code review

bealdav:
@rousseldenis could you merge please.

Thanks

PR #3194, opened 2024-06-18
[15.0][IMP] sale_order_revision: add filters for archived/active SO

ThiagoMForgeFlow:
@ForgeFlow

### DIFF
--- a/sale_order_revision/view/sale_order.xml
+++ b/sale_order_revision/view/sale_order.xml
@@ -8,6 +8,18 @@
             <field name="partner_id" position="after">
                 <field name="current_revision_id" invisible="1" />
             </field>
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
:+1:

DavidJForgeFlow (approved):

rousseldenis:
Mlmmmh, not sure this should be here

rousseldenis:
Why not using/enhancing this instead ? https://github.com/OCA/sale-workflow/blob/15.0/sale_order_archive

As if both module are installed,  view will be incoherent.

LoisRForgeFlow:
Hi @rousseldenis 

Thanks for your suggestion, we'll use that module instead.

The PR can be closed.

PR #3201, opened 2024-06-19
[17.0][IMP] sale_fixed_discount: prevent using replace in xml view extension

CRogos:
My initial issue was that another addon adds an td node in the front, so that td[3] wasn't hitting the correct node anymore:
//section[@id='details']//tbody[hasclass('sale_tbody')]//t[@t-foreach='lines_to_report']/tr//t[@t-if='not line.display_type']//td[3]/div

Therefore I've removed this xpath and replaced it with setting attributes, which is to prefer anyway.


rousseldenis (approved):
Great !

MohamedOsman7 (approved):
LGTM

CRogos:
@pedrobaeza  could you merge?

PR #3205, opened 2024-06-25
[16.0][IMP] sale_pricelist_from_commitment_date: enable feature on pricelists

sebalix:
Enable it by default (retro-compatibility).

### DIFF
--- a/sale_pricelist_from_commitment_date/__manifest__.py
+++ b/sale_pricelist_from_commitment_date/__manifest__.py
@@ -9,5 +9,8 @@
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "depends": ["sale"],
+    "data": [
+        "views/product_pricelist.xml",
+    ],
     "installable": True,
 }
--- a/sale_pricelist_from_commitment_date/__manifest__.py
+++ b/sale_pricelist_from_commitment_date/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Sale Pricelist From Commitment Date",
     "summary": "Use sale order commitment date to compute line price from pricelist",
-    "version": "16.0.1.0.1",
+    "version": "16.0.1.1.0",
     "category": "Sale",
     "website": "https://github.com/OCA/sale-workflow",
     "author": "Camptocamp, Odoo Community Association (OCA)",
--- a/sale_pricelist_from_commitment_date/migrations/16.0.1.1.0/post-migrate.py
+++ b/sale_pricelist_from_commitment_date/migrations/16.0.1.1.0/post-migrate.py
@@ -1,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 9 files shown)


rousseldenis:
@sebalix Could you check tests?

sebalix:
We have some tests related to `sale_global_discount` failing, like:
```
TestSaleGlobalDiscount.test_02_global_sale_discounts_from_partner
  File "/__w/sale-workflow/sale-workflow/sale_global_discount/tests/test_sale_global_discount.py", line 161, in test_02_global_sale_discounts_from_partner
    self.assertAlmostEqual(self.sale.amount_global_discount, 162.49)
AssertionError: 3.25 != 162.49 within 7 places (159.24 difference)
```
I cannot reproduce the issue locally with `sale_global_discount` installed alongside `sale_pricelist_from_commitment_date` :thinking: I still cannot confirm its related.

EDIT: it's definitely linked, setting the module as `installable: False` make the CI :green_circle:

cyrilmanuel (approved):
not an expert but LGTM

santostelmo (approved):

cyrilmanuel (approved):

alexey-pelykh (commented):
Thanks for this improvement, @sebalix. Making the commitment date pricing opt-in per pricelist is a solid design choice -- it gives users proper control rather than silently overriding the standard date logic for all pricelists.

The implementation looks good overall. The refactoring of `_compute_price_unit` and `_compute_pricelist_item_id` to use `groupby` and batch lines per order is a welcome improvement over the previous per-line loop, and extracting `_get_pricelist_date()` on `sale.order` provides a clean extension point. The migration correctly enables the flag on all existing pricelists to preserve current behavior on upgrade.

One thing that still seems unresolved though: @sebalix confirmed back in July 2024 that this PR causes `sale_global_discount` tests to fail (the `test_02_global_sale_discounts_from_partner` assertion). Setting this module as non-installable makes CI green again, so the interaction is confirmed. Has there been any progress on diagnosing or fixing that? The PR won't be mergeable until cross-module compatibility is sorted out.

Minor observation on the view: the `price_based_on_delivery_date` field could benefit from a `string` and `help` attribute to make it clear to end users what this toggle does (e.g. "Use the commitment/expected delivery date instead of the order date for pricelist rule matching").

cyrilmanuel:
for order, lines in groupby(self, key=lambda line: line.order_id):

odoo.tools.groupby requires the recordset to be sorted by the same key. If not sorted, grouping becomes inconsistent and some lines are skipped or mis-grouped. That silently corrupts the price computation. In standard 

perhaps this came from that point. to be check if someone have another idea. i've try to change the group by and sort but nothing change.

sale_pricelist_from_commitment_date/__manifest__.py
```diff
@@ -3,11 +3,14 @@
 {
     "name": "Sale Pricelist From Commitment Date",
     "summary": "Use sale order commitment date to compute line price from pricelist",
-    "version": "16.0.1.0.1",
+    "version": "16.0.1.1.0",
```
rousseldenis, line 6: @sebalix Don't change it. ocabot will do it

rousseldenis: migration script ok

rousseldenis:
@sebalix Could you rebase ?

rousseldenis:

PR #3206, opened 2024-06-25
[15.0] [IMP] sale_order_product_assortment: Use original product_id field keeping domain. Use cache to performance

toita86:
FWP of https://github.com/OCA/sale-workflow/pull/3135


carlosdauden:
This code was merged, the fact, the original code:
https://github.com/OCA/sale-workflow/pull/2538
is the base backport in:
https://github.com/OCA/sale-workflow/pull/3135

pedrobaeza:
Then this can be closed, right?

PR #3208, opened 2024-06-26
[16.0][ADD] sale_stock_order_line_quantity_readonly

trisdoan:


rousseldenis:
@trisdoan Can you explain in which case the 'Locked' state is not enough to do that ?

ivantodorovich (commented):
Technically LG

Though I think the readme needs to better explain the use case

santostelmo:
Implementation LG

Improvement on the readme is welcome explaining that the module is restricting quantities update on sale state and why locked is not enough.

cyrilmanuel (approved):

trisdoan:
Hello @rousseldenis, @santostelmo , @ivantodorovich, thank you all for the in-depth review.

I enhanced the readme :pray:

sale_stock_order_line_quantity_readonly/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module extends the functionality of the Sales module to restrict users from updating the quantity of products in sale order lines:
+
+- When the sale order is locked/confirmed, users are still able to adjust the quantities.
```
rousseldenis, line 3: This is not the standard. You should not write this or mention the module that change that behavior.

sale_stock_order_line_quantity_readonly/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
...
+- This becomes a problem if the related pickings are already (or partially) released, as it generates a return operation.
```
rousseldenis, line 4: Are you sure ? I've had the same problem but after install custom modules. That depends on the stock move merge process that uses some fields like 'date_deadline'

sale_stock_order_line_quantity_readonly/readme/DESCRIPTION.rst
simahawk: And you can test this on Runbot ;)

sale_stock_order_line_quantity_readonly/readme/DESCRIPTION.rst
simahawk: @trisdoan he's right. You should mention that this is only a way to have more fine grained control and that it offers an hook to customize the behavior for custom modules.

trisdoan: Hello @rousseldenis @simahawk, you're right. After checking with BA, Locked state is enough for them.

I will close this PR

PR #3211, opened 2024-06-27
[16.0] sale_triple_discount: public invalidate_cache is deprecated, changed to private _invalidate_cache

IJOL:


sale_triple_discount/models/sale_order_line.py
```diff
@@ -102,7 +102,7 @@ def triple_discount_preprocess(self):
         this method is called multiple times.
         Updating the cache provides consistency through re-computations."""
         prev_values = dict()
-        self.invalidate_cache(fnames=self._discount_fields(), ids=self.ids)
+        self._invalidate_cache(fnames=self._discount_fields(), ids=self.ids)
```
rousseldenis: This is not correct: https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L6004

You should change to `self.invalidate_recordset(fnames=self._discount_fields())`

rousseldenis (changes requested):
@IJOL Thanks for this.

Code review

sale_triple_discount/models/sale_order_line.py
IJOL: No so sure, I did it firstly the way you ask for, but after taking a look at how it's implemented in core, just changed it to save some cycles, and because doing so the modification is minimal, in the other place i_c is used... but i'l do shortly..

PR #3212, opened 2024-06-27
[17.0][ADD] sale_order_catalog_previously_sold

angelmoya:
If you use new feature catalog on sale order is very usefull filter products by products that has ben sold previously to de current customer.

This new feature could be extended a lot... maybe we need new repository product-catalog.


Deriman-Alonso (approved):
LGTM 👍.

rousseldenis:
@angelmoya There are already a set of modules doing the same : https://github.com/OCA/sale-workflow/tree/17.0/sale_order_product_recommendation

Could you check ?

alexey-pelykh (changes requested):
Thanks for the contribution, @angelmoya.

I agree with @rousseldenis's question above -- `sale_order_product_recommendation` already covers "recommend products based on sales history to current customer" with a mature, well-maintained wizard. Could you clarify how this module differs from or complements that one? If the value-add is specifically the catalog search filter (vs. wizard), that might be worth calling out, but as it stands the overlap is significant and would need justification before adding a new module for it.

Beyond the scope question, a few technical issues I noticed:

1. **Search method ignores `operator`/`value`**: `_search_product_catalog_product_previously_sold` validates that operator is `=`/`!=` and value is `bool`, but then always returns `[("id", "in", product_ids)]` regardless of the actual operator/value combination. Searching for `("product_catalog_product_previously_sold", "=", False)` or `("...", "!=", True)` would incorrectly return the products that *were* sold. The logic needs to invert the domain when the effective condition is negation.

2. **`order_id` default as empty string**: In the search method, `self.env.context.get("order_id", "")` passes an empty string to `browse()` which will raise. Should default to `False` and guard accordingly (like the compute method does).

3. **Wrong copyright in tests**: The test file header says `Copyright 2020 Camptocamp SA` -- looks copied from another module. Should be the actual author.

4. **Test method naming**: `test_sale_order_carrier_auto_assign` does not describe what it tests. Appears to be leftover from a copy-paste.

5. **`development_status` should not be `Production/Stable`** for a brand-new module that has never been released. `Beta` would be more appropriate.

6. **Typo in DESCRIPTION.md**: "catalago" should be "catalog".

Please address the overlap question from @rousseldenis first -- if there is a clear use case that `sale_order_product_recommendation` doesn't cover, the code issues above would need fixing before merge.

PR #3213, opened 2024-06-28
[16.0][FIX] sale_quotation_number: Fix hardcoded quotation sequence p…

aaladro-kernet:
…refix on sale order confirm

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -4,11 +4,13 @@
 # © 2020 Manuel Regidor  <manuel.regidor@sygel.es>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
+
+    confirmed_before = fields.Boolean(copy=False)
 
     @api.model_create_multi
     def create(self, vals_list):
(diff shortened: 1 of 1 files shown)


sale_quotation_number/models/sale_order.py
```diff
@@ -4,12 +4,14 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    confirmed_before = fields.Boolean(copy=False)
```
rousseldenis, line 13: @aaladro-kernet Thanks for this. I would say this is a little bit too generic. Maybe `quotation_number_set` or something like that.

rousseldenis (changes requested):

sale_quotation_number/models/sale_order.py
rousseldenis: Moreover, a little test and a migration script (that set that field for 'SQ' quotations) are welcome

aaladro-kernet: This field is simply to control the order, once confirmed, go back to draft and reconfirm. Do not assign a new sequence number. Are we agreed?

This allows us to modify the prefix of the quotation sequence as we wish.

MohamedOsman7:
@rousseldenis @aaladro-kernet i opened another PRs for this issue in all 3 versions.
(17: #3310 , 16: #3309 , 15: #3311)

PR #3215, opened 2024-06-28
[FIX] [16.0] sale_order_line_effective_date: Add commitment date

Shide:
Add real commitment date to the lines.

Is just a relation with the real commitment date.

Checked with [sale_order_line_date](https://github.com/OCA/sale-workflow/tree/16.0/sale_order_line_date) module and doesn't do the same. This only adds the real commitment date set on the order. The other one performs another operations.

MT-6075 @moduon @rafaelbn @Gelojr @fcvalgar please review if you want :)

### DIFF
--- a/sale_order_line_effective_date/README.rst
+++ b/sale_order_line_effective_date/README.rst
@@ -29,7 +29,8 @@ Sale Order Line Effective Dates
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module extends the functionality of sale stock and adds "Effective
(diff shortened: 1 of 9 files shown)


Gelojr (approved):
Nice job @Shide LGTM
https://www.loom.com/share/fc5798d248eb42ffa8a6d4a22b86f920?sid=482221f9-c18e-412c-b89e-fa9a4bf20504

yajo (approved):

yajo (approved):
/ocabot merge minor

PR #3216, opened 2024-06-29
[16.0][IMP] sale_global_discount: refactor _compute_amounts

trisdoan:
### Context

In simple case (tax is not included in price), flow of the module as follow:
   - Let tax is computed as usual, as odoo native supports
   - Afterwards, checking if global discount is applied on Sale order:
      1. If not, only add amt_untaxed_before_global_discount and amt_total_before_global_discount,
      2. If yes, precompute tax

**A example of Sale order is below:**
- Tax is not included in price
- With 20% global discount

<div style="display: flex;">

<div style="flex: 1;">

| **Product Name** | **Quantity** | **Unit Price** | **SubTotal** |
|------------------|--------------|----------------|-----------------|
| Example Product  | 2            | $75        | $150        |

</div>

<div style="flex: 0; margin-left: 20px;">

| **Untaxed Amt** |
|------------------|
| $120      |

| **Tax** |
|------------------|
| $18.0      |

| **Total** |
|------------------|
| $138.0      |

**=> So tax is recalculated based on Final price (after global discount)**

- Before this change, if enabling included-in-price tax, it breaks the intended flow as it triggers the part 'handling price include' again with the new price

### This change
- pass param: `handle_price_include=False` because it was handle in super(), by default `handle_price_include=True` . Now, as we recalculate tax to apply global discount, there is no need to handle price included or not.

- With above example, now it is:
   **- but this time, Tax is included in price**

<div style="display: flex;">

<div style="flex: 1;">

| **Product Name** | **Quantity** | **Unit Price** | **SubTotal** |
|------------------|--------------|----------------|-----------------|
| Example Product  | 2            | $75        | $130.43       |

</div>

<div style="flex: 0; margin-left: 20px;">

| **Untaxed Amt** |
|------------------|
| $104.34      |

| **Tax** |
|------------------|
| $15.65      |

| **Total** |
|------------------|
| $119.9900      |

- Add a check to only recompute tax when global discount is applied on the sale order. Before this change, it results in rounding issue. It happens when products are configured as tax-included in price. For example:

sample_sale_global_discount

### DIFF
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -112,9 +112,13 @@ class SaleOrder(models.Model):
                     product=line.product_id,
                     partner=line.order_id.partner_shipping_id,
                 )
-                amount_discounted_tax += sum(
-                    t.get("amount", 0.0) for t in discounted_tax.get("taxes", [])
-                )
+                for tax in discounted_tax.get("taxes", []):
+                    if line.tax_id.browse(tax["id"]).amount_type == "fixed":
+                        amount_discounted_tax += (
+                            tax.get("amount", 0.0) * line.product_uom_qty
+                        )
+                        continue
+                    amount_discounted_tax += tax.get("amount", 0.0)
             order.update(
                 {
                     "amount_untaxed_before_global_discounts": (
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -94,48 +94,51 @@ class SaleOrder(models.Model):
         res = super()._compute_amounts()
         for order in self:
             order._check_global_discounts_sanity()
-            amount_untaxed_before_global_discounts = order.amount_untaxed
-            amount_total_before_global_discounts = order.amount_total
+            vals = {
+                "amount_untaxed_before_global_discounts": order.amount_untaxed,
+                "amount_total_before_global_discounts": order.amount_total,
+            }
             discounts = order.global_discount_ids.mapped("discount")
-            amount_discounted_untaxed = amount_discounted_tax = 0
-            for line in order.order_line:
-                discounted_subtotal = line.price_subtotal
-                if not line.product_id.bypass_global_discount:
-                    discounted_subtotal = self.get_discounted_global(
-                        line.price_subtotal, discounts.copy()
-                    )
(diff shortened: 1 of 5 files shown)


sale_global_discount/models/sale_order.py
```diff
@@ -94,44 +94,52 @@ def _compute_amounts(self):
...
+                    discounted_subtotal = line.price_subtotal
+                    if not line.product_id.bypass_global_discount:
+                        discounted_subtotal = self.get_discounted_global(
+                            line.price_subtotal, discounts.copy()
+                        )
+                    amount_discounted_untaxed += discounted_subtotal
+                    discounted_tax = line.tax_id.compute_all(
+                        discounted_subtotal,
+                        line.order_id.currency_id,
+                        1.0,
```
rousseldenis: Why using 1.0 and not mimic Odoo method ? https://github.com/odoo/odoo/blob/45154cfdb044aea85ad72bb36348af7b77afd844/addons/sale/models/sale_order_line.py#L879

cyrilmanuel (approved):

sale_global_discount/models/sale_order.py
```diff
@@ -94,44 +94,52 @@ def _compute_amounts(self):
...
-                discounted_tax = line.tax_id.compute_all(
-                    discounted_subtotal,
-                    line.order_id.currency_id,
-                    1.0,
-                    product=line.product_id,
-                    partner=line.order_id.partner_shipping_id,
+                    for tax in discounted_tax.get("taxes", []):
+                        if line.tax_id.browse(tax["id"]).amount_type == "fixed":
+                            amount_discounted_tax += (
+                                tax.get("amount", 0.0) * line.product_uom_qty
```
santostelmo: why do you need to multiply `discounted_tax` by `product_uom_qty`?
`compute_all` method with `product_uom_qty` passed as arguments does not give the same result ?

trisdoan:
Hello @rousseldenis, @santostelmo, thank you for your in-depth review.

I just simplified the tax calculation. Briefly, discounted price_unit is passed as base to calculate tax for each line (similarly how untaxed amount is calculated)

santostelmo (approved):

rousseldenis (approved):
Code review

PR #3219, opened 2024-07-02
[16.0][sale_company_currency] Added amount_untaxed_curr and amount_tax

renatonlima:
Currently the sale_company_currency module only has the total field converted into the company currency, this PR implements the total tax-free and tax total fields converted into the company currency.

image

Fields are hidden in the tree view

image


alexey-pelykh (changes requested):
Thanks for adding the untaxed and tax company-currency fields -- useful addition.

A few things need attention before this can be merged:

**1. Dependency escalation from `sale` to `sale_margin`**

The original module only depends on `sale`. This PR changes it to `sale_margin` (which pulls in `sale_management`) solely to add the `margin_curr` field. That's a significant dependency escalation for users who want company-currency amounts but don't use margins.

I'd suggest splitting this: keep `amount_untaxed_curr` and `amount_tax_curr` in this module (with `"depends": ["sale"]`), and add `margin_curr` in a separate module (e.g. `sale_company_currency_margin`) or in a glue module that depends on both `sale_company_currency` and `sale_margin`.

**2. Conversion direction changed from `*` to `/` -- undocumented bug fix**

The existing code computes `amount_total * currency_rate`. Since `currency_rate` on `sale.order` is the rate *from company currency to order currency* (i.e. `_get_conversion_rate(from_currency=company, to_currency=order)`), dividing is indeed the correct way to convert back. So the PR silently fixes a bug in the existing `amount_total_curr` field.

This should be called out explicitly -- it's a behavioral change for existing users and their stored values will need recomputation. Please mention this in the PR description and commit message.

**3. Division by zero**

With the switch to division, `currency_rate` being 0 will raise a `ZeroDivisionError`. Please add a guard, e.g.:

```python
rate = order.currency_rate or 1.0
```

**4. No tests**

The module has no test coverage. Adding tests for the computed fields -- especially the currency conversion logic and the same-currency short-circuit -- would be expected for an OCA contribution.

sale_company_currency/models/sale_order.py
```diff
@@ -14,6 +14,22 @@ class SaleOrder(models.Model):
         readonly=True,
         store=True,
     )
+    amount_untaxed_curr = fields.Monetary(
```
rousseldenis, line 17: @renatonlima Explicit names is better than implicit (amount_untaxed_currency`)

sale_company_currency/models/sale_order.py
```diff
@@ -23,11 +39,31 @@ class SaleOrder(models.Model):
...
     def _compute_amount_company(self):
         for order in self:
             if order.currency_id.id == order.company_id.currency_id.id:
+                to_untaxed = order.amount_untaxed
+                to_tax = order.amount_tax
                 to_amount = order.amount_total
+                to_margin = order.margin
             else:
-                to_amount = order.amount_total * order.currency_rate
+                to_untaxed = order.amount_untaxed / order.currency_rate
```
rousseldenis, line 62: Could you check division by zero is protected ?

rousseldenis:

PR #3221, opened 2024-07-03
Added setting on partner

drzaphod85:
Added the ability to set a standard invoice policy for a partner so that it will always be sett on the sales order. It will select the default setting from Sales->Settings->Invoice->Invoice Policy as the default setting when a new partner is created. Also added an extra menu item in the Action menu that can update all partners with the default setting if no setting is set from the beginning.


rousseldenis:
@drzaphod85 Could you check this ? https://github.com/OCA/sale-workflow/pull/3376

rousseldenis:

PR #3225, opened 2024-07-05
[15.0][FIX] sale_planner_calendar: 'calendar.event' object has no attribute 'start_datetime'

carolinafernandez-tecnativa:
- With the router we can establish a recurring calendar event of type "Delivery" that means that when a sales order is confirmed that generates a delivery, it establishes the delivery date on the order on the next day that has been defined in the event. calendar.
- Before this fix, when trying to confirm a sale order got this error message:

image

@Tecnativa
TT49977
@pedrobaeza @sergio-teruel

### DIFF
--- a/sale_planner_calendar/models/sale_order.py
+++ b/sale_planner_calendar/models/sale_order.py
@@ -28,7 +28,7 @@ class SaleOrder(models.Model):
                 limit=1,
             )
             if delivery_event:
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Clear patch, so fast-tracking it:

/ocabot merge patch

PR #3228, opened 2024-07-11
[14.0] [FIX] sale_product_set_packaging_qty: Fix quantity update on product.set.line

ajaniszewska-dev:
Changing qty on packaging in product form is not reflected on product.set.line record having this package set. To avoid this situation, we should recompute packaging_qty and update values on the product.set.line whenever packaging_ids changes on product.

### DIFF
--- a/sale_product_set_packaging_qty/models/product_set.py
+++ b/sale_product_set_packaging_qty/models/product_set.py
@@ -17,7 +17,12 @@ class ProductSetLine(models.Model):
         digits="Product Unit of Measure",
     )
 
-    @api.depends("quantity", "product_packaging_id", "product_packaging_id.qty")
+    @api.depends(
+        "quantity",
+        "product_packaging_id",
+        "product_packaging_id.qty",
+        "product_id.packaging_ids",
+    )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please follow proper commit message format: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

ajaniszewska-dev:
@pedrobaeza thank you for reminder, should be better now

sale_product_set_packaging_qty/models/product_set.py
```diff
@@ -31,6 +36,7 @@ def _compute_product_packaging_qty(self):
                 line.product_packaging_qty = 0
                 continue
             line.product_packaging_qty = line.quantity / line.product_packaging_id.qty
+            line.write(line._prepare_product_packaging_qty_values())
```
rousseldenis: You should not write in a compute

```suggestion
            line.update(line._prepare_product_packaging_qty_values())
```

rousseldenis (changes requested):

simahawk (approved):

PR #3030, opened 2024-03-21
[17.0][MIG] sale_order_general_discount: 17.0

miguel-S73:
Standard migration from 16.0 to 17.0 and I have added  `cherry-pick` https://github.com/OCA/sale-workflow/pull/2524 and this suggestion https://github.com/OCA/server-backend/pull/221#pullrequestreview-1664274601 too.

PR #3157, opened 2024-05-28
[17.0][MIG] sale_triple_discount: Migration to 17.0

Boxnia:
Standard migration to 17.0
Depends on:

[ [17.0][MIG] account_invoice_triple_discount: Migration to 17.0  account-invoicing#1626](https://github.com/OCA/account-invoicing/pull/1626)

PR #3235, opened 2024-07-15
[17.0][MIG] sale_order_general_discount_triple: Migration to 17.0

arantxa-s73:
Depends on:

- [ sale_order_general_discount ] https://github.com/OCA/sale-workflow/pull/3030
- [ sale_triple_discount ] https://github.com/OCA/sale-workflow/pull/3929


dalonsod:
I've superseded `sale_triple_discount` PR migration with #3406 , you can use it instead of #3157

alexey-pelykh (commented):
Thanks for the migration work here.

I went through the changes comparing with the 16.0 source. The core adaptation from `discount` to `discount1` is handled correctly, and the pre-migration script to update existing config parameter values is a nice touch.

A few things I noticed:

**POT/translation inconsistency** -- The `.pot` file still has `Project-Id-Version: Odoo Server 16.0` and contains stale msgids (`Disc. 2 (%)`, `Disc. 3 (%)`) from the 16.0 field labels. The `it.po` has the same stale entries with `Project-Id-Version: Odoo Server 14.0`. Meanwhile `es.po` was apparently regenerated and has the correct 17.0 msgids (`Discount 1 (%)`, `Discount 2 (%)`, etc.). The `.pot` and `it.po` should be regenerated to match the actual 17.0 field definitions.

**`_recompute_prices` and `_create_delivery_line`** -- The `_recompute_prices` override is a good addition for 17.0. One thing: in `_create_delivery_line`, calling `_compute_discount1/2/3` directly on every order line (not just the new delivery line) looks like it could reset user-edited values on existing lines. The 16.0 version had the same pattern with `_compute_discount`, so this may be intentional, but worth double-checking whether the scope should be narrower (e.g. only the newly created delivery line).

**`bypass_general_discount` guard** -- Good alignment with what `sale_order_general_discount` does in 17.0. However, this check is only applied in the `general_discount` branch of each compute method. In the zero-reset branch (when the discount field is not configured for either pricelist or general), the loop still runs through all lines including reward lines. The `_check_is_reward_line` guard is applied there, which is fine, but just noting the asymmetry.

**Dependency PR #3929** -- The PR body references `sale_triple_discount` PR #3929 which is still open. The `test-requirements.txt` correctly pulls it from the PR head for CI. Just flagging that this PR is blocked until #3929 merges.

Overall the migration logic looks sound -- the discount→discount1 rename is correctly propagated throughout, and the patterns align with how the dependency modules work in 17.0.

Reyes4711-S73 (approved):
LGTM

RabbitJon-S73 (approved):

rousseldenis:
@arantxa-s73 Could you rebase ?

rousseldenis:
@sbiosca-s73

rousseldenis:

PR #3240, opened 2024-07-19
[FIX] sale_company_currency: Changes on currency rate application

dalonsofl:
### Module

sale_company_currency

### Describe the bug

The calculation of Total (Company Currency) field is not correct, it is diving currency rate instead of multiplying

### To Reproduce

1. Create a sale order
2. Use a pricealist with the same currency as the company and check the field
3. Change the pricelist for one with different currency of the company

(image from runboat OCA)

Settings

Captura desde 2024-07-19 09-15-38

USD

image

image

EUR

Captura desde 2024-07-19 09-16-58

Captura desde 2024-07-19 09-17-54

The 36.37 is not correct because it is dividing the total in the currency of the order (in this case €) by the rate of the company's currency (which is USD) when it should be multiplying it to get the real value of the company's currency (55.60€ * 1.528900 = 85.00$) which is the value it gives at the beginning without currency changes. This applies for all the **Show Currency Rate** types in the configuration.

**Affected versions:**

16.0

### DIFF
--- a/sale_company_currency/models/sale_order.py
+++ b/sale_company_currency/models/sale_order.py
@@ -29,5 +29,5 @@ class SaleOrder(models.Model):
             if order.currency_id.id == order.company_id.currency_id.id:
                 to_amount = order.amount_total
             else:
-                to_amount = order.amount_total / order.currency_rate
+                to_amount = order.amount_total * order.currency_rate
             order.amount_total_curr = to_amount


flalexg (approved):
Functional review! LGTM!

renatonlima (approved):

PR #3244, opened 2024-07-22
[17.0][MIG] sale_advance_payment: Migration to 17.0

yankinmax:
This PR is opened to fast track PR rewiews suggestions asked here:
 - [x] https://github.com/OCA/sale-workflow/pull/3156


yankinmax:
@Camille0907 @vvrossem @gurneyalex 
can you pls take a look?

sale_advance_payment/models/payment.py
```diff
@@ -7,6 +7,4 @@
 class AccountPayment(models.Model):
     _inherit = "account.payment"
 
-    sale_id = fields.Many2one(
-        "sale.order", "Sale", readonly=True, states={"draft": [("readonly", False)]}
-    )
+    sale_id = fields.Many2one(comodel_name="sale.order", string="Sale", readonly=True)
```
vvrossem, line 10: Depending if `sale_id` could be imported:
> If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.

cf. [Migration to 17.0](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0)

sale_advance_payment/models/sale.py
```diff
@@ -9,16 +9,17 @@ class SaleOrder(models.Model):
...
     account_payment_ids = fields.One2many(
-        "account.payment", "sale_id", string="Pay sale advanced", readonly=True
+        comodel_name="account.payment",
+        inverse_name="sale_id",
+        string="Pay sale advanced",
     )
     amount_residual = fields.Float(
-        "Residual amount",
-        readonly=True,
+        comodel_name="Residual amount",
```
vvrossem: `string="Residual amount",`

vvrossem (approved):
pre-approving, thank you :pray:

sale_advance_payment/models/payment.py
yankinmax: `sale_id` in this case won't be imported and is used as a inverse field for one2many relation
It's not shown on any form and it doesn't make sense to show it, so we can simply leave `readonly=True` here.
same as here:
- [ ] https://github.com/OCA/purchase-workflow/blob/43ee24d83333ed82492abfa37d34fb6d8c7a2041/purchase_advance_payment/models/payment.py#L13

gurneyalex (approved):

gurneyalex (approved):

PR #3246, opened 2024-07-23
[FIX] sale_order_line_date: add new condition

adrip-s73:
Check if "commitment_date" exists in vals to aviod the error bellow when we change line quantity.

image


ferran-S73 (approved):

StefanRijnhart:
https://github.com/OCA/sale-workflow/pull/3155/files seems to be fixing the same issue by using the order's expected date as a fallback. What do you think of that?

StefanRijnhart:
Closing in favor of https://github.com/OCA/sale-workflow/pull/3364 which includes tests.

PR #3249, opened 2024-07-24
[16.0][IMP] sale_packaging_default: Add context to support bypass setting default packaging

QuocDuong1306:
In some cases, we don't want to get the default for packaging. This PR will provide a context to use for those cases.

### DIFF
--- a/sale_packaging_default/models/sale_order_line.py
+++ b/sale_packaging_default/models/sale_order_line.py
@@ -22,6 +22,9 @@ class SaleOrderLine(models.Model):
     @api.depends("product_id", "product_uom_qty", "product_uom")
     def _compute_product_packaging_id(self):
         """Set a default packaging for sales if possible."""
+        if not self.env.context.get("apply_default_packaging", True):
+            return super()._compute_product_packaging_id()
+
         for line in self:
             if line.product_id != line.product_packaging_id.product_id:
                 line.product_packaging_id = line._get_default_packaging(line.product_id)


yajo:
In which cases?

sale_packaging_default/models/sale_order_line.py
```diff
@@ -22,6 +22,9 @@ def onchange(self, values, field_name, field_onchange):
     @api.depends("product_id", "product_uom_qty", "product_uom")
     def _compute_product_packaging_id(self):
         """Set a default packaging for sales if possible."""
+        if self.env.context.get("bypass_default_packaging", False):
```
yajo: issue: please [avoid negative booleans](https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/)

```suggestion
        if not self.env.context.get("apply_default_packaging", True):
```

yajo (changes requested):

sale_packaging_default/models/sale_order_line.py
QuocDuong1306: Thank @yajo  for your suggestion

QuocDuong1306:
>In which cases?

Hi @yajo , this context will be helpful for those cases:
- With specific products, users don't want to get the default for product packaging but keep using the default for others.
- Or using `sale_packaging_default` and `sell_only_by_packaging` together in 1 DB, using this context in the glue module will help for the force quantities feature

yajo:
> * With specific products, users don't want to get the default for product packaging but keep using the default for others.

If this is your use case, have you considered simply prepending a packaging of 1 to the list of packagings? It will be used by default in that product and give you the default Odoo behavior without any extra code.

Otherwise, the fix is incomplete, as this won't allow a product to do that.

> * Or using `sale_packaging_default` and `sell_only_by_packaging` together in 1 DB, using this context in the glue module will help for the force quantities feature

If this is your case, I think it is more appropriate to add this is in the PR where you add the glue module. Or at least, to keep it as draft until you open such PR. And please add some tests.

sale_packaging_default/models/sale_order_line.py
```diff
@@ -22,6 +22,9 @@ def onchange(self, values, field_name, field_onchange):
     @api.depends("product_id", "product_uom_qty", "product_uom")
     def _compute_product_packaging_id(self):
         """Set a default packaging for sales if possible."""
+        if not self.env.context.get("apply_default_packaging", True):
```
rousseldenis, line 25: Depending on context for stored fields is usually bad. Maybe you can explain a little bit your use case ?

rousseldenis: But I see @yajo has already asked.

QuocDuong1306:
>Or using sale_packaging_default and sell_only_by_packaging together in 1 DB, using this context in the glue module will help for the force quantities feature

Hello @yajo @rousseldenis , actually, this is my case

>If this is your case, I think it is more appropriate to add this is in the PR where you add the glue module. Or at least, to keep it as draft until you open such PR. And please add some tests.

I'll double-check for this one. Temporarily, marking this PR as `Draft` is a good idea. Thanks

cyrilmanuel:
@QuocDuong1306 any new on that PR ? can we go on with that ?

rousseldenis:
Closing this as very old. Feel free to reopen it if needed. User seems inactive

PR #3252, opened 2024-07-26
[fix] sale_by_packaging: update can_be_sold on change

henrybackman:
When changing the packaging type, the `can_be_sold` checkbox did not update on user form to match the selected type

Added onchange function to resolve this.

### DIFF
--- a/sale_by_packaging/models/product_packaging.py
+++ b/sale_by_packaging/models/product_packaging.py
@@ -26,6 +26,14 @@ class ProductPackaging(models.Model):
         for record in self:
             record.can_be_sold = record.packaging_type_id.can_be_sold
 
+    # This shouldn't be required due to the compute method.
+    # However, we experienced some cases where nested form inside the product
+    # form fails to update the flag without this onchange.
+    @api.onchange("packaging_type_id")
+    def _onchange_packaging_type(self):
+        for record in self:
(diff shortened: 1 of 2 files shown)


sale_by_packaging/models/product_packaging.py
```diff
@@ -26,6 +26,11 @@ def _compute_can_be_sold(self):
         for record in self:
             record.can_be_sold = record.packaging_type_id.can_be_sold
 
+    @api.onchange("packaging_type_id")
+    def _onchange_packaging_type(self):
```
simahawk, line 33: strange... this should be covered by the compute method above :thinking:

simahawk (approved):

ivantodorovich (approved):
weird, indeed..

sale_by_packaging/models/product_packaging.py
sebalix: Why not using a simple related readonly=False + store=True here?
Can't say if it'll solve the issue but I do not understand the current implementation of `_compute_can_be_sold` (unless it is here to be overridden?)

henrybackman:
After further investigation, this change is not anymore needed. There was a missing `@api.depends()` on a compute function override that was causing the values not to update on the form. Closing this PR.

PR #3258, opened 2024-07-29
[FIX] sale_fixed_discount: fix fixed discount on API-created order lines

sefirosweb:
This fix when you want to create order via API and you want use discount_fixed instead "discount" rate

Without that the sales_order_line is created with discount_fixed  but discount is 0

image


sale_fixed_discount/models/sale_order_line.py
```diff
@@ -15,6 +15,13 @@ class SaleOrderLine(models.Model):
         help="Fixed amount discount.",
     )
 
+    @api.model
```
rousseldenis:
```suggestion
    @api.model_create_multi
```

And change the function accordingly

rousseldenis (changes requested):
@sefirosweb Could you fix this?

Moreover, adding a test case for that flow should be great.

sale_fixed_discount/models/sale_order_line.py
sefirosweb: You are right, done

rousseldenis: @sefirosweb Of course, the implementation of the function should be changed accordingly :sweat_smile:

sefirosweb:
@rousseldenis I saw a bit bug when is creating the records i fixed them

sale_fixed_discount/models/sale_order_line.py
```diff
@@ -15,6 +15,14 @@ class SaleOrderLine(models.Model):
         help="Fixed amount discount.",
     )
 
+    @api.model_create_multi
```
rousseldenis: @sefirosweb 

You don't need to override create():

```suggestion
    @api.depends("discount_fixed", "price_unit")
    def _compute_discount(self):
        super()._compute_discount()
        for line in self:
            if line.discount_fixed:
                line.discount = line._get_discount_from_fixed_discount()
```

And then, remove onchange.

sefirosweb: Yes you are right, I test it and works for me

rousseldenis:
@sefirosweb Could you improve your commits messages like `[IMP] sale_fixed_discount: (...)` and fix pre-commit ?

rousseldenis:
@sefirosweb Do you plan to finish this ?

sefirosweb:
OK!

alexey-pelykh (changes requested):
Thanks for the fix @sefirosweb, the core idea of overriding `_compute_discount` is the right approach here.

A few things still need attention before this can be merged:

1. **Remove `_onchange_discount_fixed`** -- as @rousseldenis mentioned in the earlier inline comment, once `_compute_discount` is in place the onchange is redundant. The compute method fires in both UI and API contexts, so the onchange should be dropped entirely.

2. **Test indentation** -- the body of `test_05_discount_fixed_when_created_by_api` is indented with 12 spaces instead of the expected 8 (4 for class + 4 for method). Please align with the rest of the test file.

3. **Missing newline at end of file** -- the test file is missing a trailing newline, which will also trip pre-commit.

4. **Reuse existing fixtures** -- the new test creates its own `res.partner` and `product.product` when `self.partner` and `self.product` (or `self.product2`) from `setUpClass` could be reused. Keeps tests consistent and leaner.

5. **Commit messages** -- @rousseldenis already flagged this: commits should follow OCA format, e.g. `[FIX] sale_fixed_discount: ...`. The intermediate commits should be squashed into a single clean commit.

Points 1-4 are likely what's causing the CI failures. Should be straightforward to fix up.

rousseldenis:
> Thanks for the fix @sefirosweb, the core idea of overriding `_compute_discount` is the right approach here.
> 
> A few things still need attention before this can be merged:
> 
> 1. **Remove `_onchange_discount_fixed`** -- as @rousseldenis mentioned in the earlier inline comment, once `_compute_discount` is in place the onchange is redundant. The compute method fires in both UI and API contexts, so the onchange should be dropped entirely.
> 2. **Test indentation** -- the body of `test_05_discount_fixed_when_created_by_api` is indented with 12 spaces instead of the expected 8 (4 for class + 4 for method). Please align with the rest of the test file.
> 3. **Missing newline at end of file** -- the test file is missing a trailing newline, which will also trip pre-commit.
> 4. **Reuse existing fixtures** -- the new test creates its own `res.partner` and `product.product` when `self.partner` and `self.product` (or `self.product2`) from `setUpClass` could be reused. Keeps tests consistent and leaner.
> 5. **Commit messages** -- @rousseldenis already flagged this: commits should follow OCA format, e.g. `[FIX] sale_fixed_discount: ...`. The intermediate commits should be squashed into a single clean commit.
> 
> Points 1-4 are likely what's causing the CI failures. Should be straightforward to fix up.

@alexey-pelykh Comments for specific changes should be done in PR's code line level. Please avoid such reviews. Thanks

rousseldenis:

PR #3260, opened 2024-08-02
[FIX] sale_order_currency_rate : Change curency_rate calculation

dalonsofl:
### Module

sale_order_currency_rate

### Describe the bug

The currency rate applied is not correct and the value in report is incorrect too

### To Reproduce

1. Create a sale order and confirm
2. Check the value in sale report

(image from runboat OCA)

image

Captura desde 2024-08-02 16-01-00

Instead of divining by the unit per USD is using the USD per unit and the value is incorect. Because this reason, the best option is using the base odoo calculation of currency_rate.

image

**Affected versions:**

16.0

### DIFF
--- a/sale_order_currency_rate/models/sale_order.py
+++ b/sale_order_currency_rate/models/sale_order.py
@@ -7,13 +7,8 @@ from odoo import api, fields, models
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    currency_rate = fields.Float(
-        compute="_compute_currency_rate",
-        digits="Currency Rate Precision",
-    )
-
     inverse_currency_rate = fields.Float(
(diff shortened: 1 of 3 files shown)


aliciagaarzo (approved):
LGTM!

rousseldenis:
@dalonsofl Could you add a test case that highlight the bug and avoid regression ?

dalonsofl:
Hi @rousseldenis !

I have added a simple test recently (I hope it will be enough), if anything else is needed, text me again without problem.

Regards

amarcosg (approved):
LGTM!

rousseldenis (approved):
LGTM. Of course using core computation is the best

PR #3008, opened 2024-03-12
[MIG] sale_cancel_reason and sale_mrp_bom

dhp-denero:
Hello Community,

Please do review my work, I have migrated sale_cancel_reason and sale_mrp_bom modules to version 17.0

Thanks

Dhaval Patel

PR #3262, opened 2024-08-06
[17.0][MIG] sale_mrp_bom

antonioburic:
superseeds https://github.com/OCA/sale-workflow/pull/3008

includes this fix from v14 too https://github.com/OCA/sale-workflow/pull/2947

_edit: tests fail seemingly for an unrelated reason_


StefanRijnhart:
/ocabot migration sale_mrp_bom

@antonioburic sale_mrp_bom looking good with all the tests running and no warnings from this module, but there are linting changes to the RST files of some other modules. Maybe it's best to remove them from this PR, and if you feel like going all the way, propose them separately for bonus points. What do you think?

antonioburic:
> /ocabot migration sale_mrp_bom
> 
> @antonioburic sale_mrp_bom looking good with all the tests running and no warnings from this module, but there are linting changes to the RST files of some other modules. Maybe it's best to remove them from this PR, and if you feel like going all the way, propose them separately for bonus points. What do you think?

@StefanRijnhart nice catch, oca-port automation and I missed to notice :) 
corrected, will create a separate PR for those changes

StefanRijnhart (approved):
Thanks!

gbrito (approved):

sale_mrp_bom/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+                bom_product_tmpl = line.bom_id.product_tmpl_id
+                bom_product = bom_product_tmpl.product_variant_ids
+            else:
+                bom_product_tmpl, bom_product = None, None
+            line_product = line.product_id
+            if not bom_product or line_product == bom_product:
+                continue
+            raise ValidationError(
+                _(
+                    "Please select BoM that has matched product with the line `{}`"
```
rousseldenis: It's better to use translate function args:

```suggestion
                    "Please select BoM that has matched product with the line `%(product_name)s`", product_name=line_product.name)
```

sale_mrp_bom/tests/test_sale_mrp_bom.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2020 Akretion Renato Lima <renato.lima@akretion.com.br>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleMrpLink(TransactionCase):
+    def setUp(self):
```
rousseldenis: Could you switch to class method instead ?

sale_mrp_bom/tests/test_sale_mrp_bom.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+    def test_define_bom_in_sale_line(self):
+        """Check manufactured order is created with BOM definied in Sale."""
+        so, bom_a, bom_b = self._prepare_so()
+
+        # Check manufacture order
+        mos = self.env["mrp.production"].search([("origin", "=", so.name)])
+        for mo in mos:
+            self.assertEqual(mo.bom_id, self.boms.get(mo.product_id.id))
```
rousseldenis, line 94: It seems this line is not reached by tests.

Could you check why ?

rousseldenis (changes requested):
Code review

sale_mrp_bom/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+    @api.constrains("bom_id", "product_id")
+    def _check_match_product_variant_ids(self):
+        for line in self:
+            if line.bom_id:
+                bom_product_tmpl = line.bom_id.product_tmpl_id
+                bom_product = bom_product_tmpl.product_variant_ids
+            else:
+                bom_product_tmpl, bom_product = None, None
```
CasVissers-360ERP: Constraint is incorrect, can be a lot more simple:
```

    @api.constrains("bom_id", "product_id")
    def _check_match_product_id(self):
        for line in self:
            if line.product_id != line.bom_id.product_id:
                raise ValidationError(
                    _(
                        "Please select BoM that has matched product with the line `{}`"
                    ).format(line.product_id.name)
                )

```

StefanRijnhart: Agree that the code looks all wrong, but don't we want to allow BoMs that only specify a product template and no specific variant? It was recently fixed in 14.0 like that: https://github.com/OCA/sale-workflow/pull/2947/files#diff-c6c892ccbd2c51f5c8a648ee366fd8c32f9ecd9ebce7c649b8c48f1f29aad900

CasVissers-360ERP: https://github.com/OCA/sale-workflow/pull/2947/files we will forward port this

CasVissers-360ERP (approved):
Functional review (after constraint is fixed)

StefanRijnhart:
@rousseldenis Is it alright to merge this one?

alexey-pelykh (approved):
Thanks for the migration @antonioburic, and for incorporating all the previous feedback.

I went through the current state of the code and it looks good to me:

- Constraint properly handles both variant-specific BoMs and template-only BoMs (the v14 fix from #2947 is included)
- `_prepare_procurement_values` correctly injects `bom_id` into the procurement values, which the mrp `_get_matching_bom` picks up
- Tests cover the main flow (BoM propagation to MO), procurement values check, mismatch validation, and the no-variant case
- `setUpClass` pattern, named translation args -- all addressed from earlier review rounds
- Manifest, views, security group all look correct for 17.0

This has been sitting for a while and the previous concerns have been resolved. Ready to merge from my side.

rousseldenis (approved):

rousseldenis:
@antonioburic Could you check tests ?

PR #2738, opened 2023-10-23
[FIX] sale_advance_payment: residual amount

Drsmartinkus:
Resolve the problem in the sale order residual amount display when a credit note is issued.

Calculation before fix example:

`invoice_paid_amount += inv.amount_total - inv.amount_residual`

As there are 2 invoices (one out invoice and once refund invoice both paid fully)

```
invoice_paid_amount += inv.amount_total - inv.amount_residual
invoice_paid_amount = (320.0 - 0) + (320 - 0)
```

Sale order residual amount is going to be calculated like:

```
amount_residual = order.amount_total - advance_amount - invoice_paid_amount
amount_residual = 320.0 - 0 - 640  
amount_residual = -320 
```

Sale order is going to display residual amount as -320.

Invoices-Odoo-single-reversal
S00035-Odoo-single-case

PR #3266, opened 2024-08-08
[FIX] sale_advance_payment: Residual Amount - Port PR#2738 to v16.0

ellbristow:
See PR https://github.com/OCA/sale-workflow/pull/2738

This is a port of that fix for 16.0


ellbristow:
@OCA/core-maintainers

FYI the failure on this one appears to be a false negative. It's asking for tests to cover the new test

theangryangel (approved):

rousseldenis:
@ellbristow Could you check latest version of the module as I see some changes have been made. Thanks

ellbristow:
@rousseldenis I've tweaked my suggestions, but that left the tests in conflict. I've modified the original test to what I believe should be the correct result, but this poses a question of "which approach is right". For the clients we have, including the credit notes in the amount_residual calculation leads to absurd results, hence the need to filter them out.

alexey-pelykh (approved):
Thanks for the port and for resolving the rebase conflicts, @ellbristow.

The fix is correct -- filtering to `move_type == "out_invoice"` when computing `invoice_paid_amount` avoids double-counting credit note payments, which was producing negative/inflated residual amounts. The test assertion change to `0` is the expected result: the order is fully invoiced and paid, so the residual should be zero regardless of the credit note.

CI is all green. LGTM.

rousseldenis:
@ellbristow Could you rebase to solve conflicts and squash your commits ? Thanks

rousseldenis:

PR #3275, opened 2024-08-19
[14.0][ADD] sale_order_revision_replace[_stock]

dreispt:
Allow revisions of confirmed Sale Order, replacing the previous revision.
Adds a "New Revision" button on confirmed SOs, and the new revision will keep the Delivered and Invoiced information.

For extra UX see also https://github.com/OCA/server-ux/pull/936


rousseldenis:
@dreispt I'm wondering why you needs this as the base module does allow revisions on canceled sale orders only ?

PR #3255, opened 2024-07-29
[17.0][MIG] sale_blanket_order

chrisb-c01:
Based on / supersedes https://github.com/OCA/sale-workflow/pull/3014

PR #3277, opened 2024-08-20
[17.0] [ADD] sale_blanket_order_tier_validation

Nikul-OSI:
Depends on 

- [x] sale_blanket_order https://github.com/OCA/sale-workflow/pull/3255

- [X] base_tier_validation https://github.com/OCA/server-ux/tree/17.0/base_tier_validation


sale_blanket_order_tier_validation/models/sale_blanket_order.py
```diff
@@ -0,0 +1,56 @@
...
+            ):
+                order.state = "done"
+            # Removing else statement from original method
+            # else:
+            #     order.state = "open"
+
+    def action_confirm(self):
+        """Set the state to 'Open' when clicking confirmed instead of during compute."""
+        for rec in self:
+            rec.write({"state": "open"})
```
StefanRijnhart, line 55: FYI I did https://github.com/OCA/server-ux/pull/939 so that tier validations can be applied to computed state fields as in this model.

rousseldenis:
@Nikul-OSI Could you rebase ?

Nikul-OSI:
> Sponsor

Done

Nikul-OSI:
> @Nikul-OSI Could you rebase ?

Done

sale_blanket_order_tier_validation/models/sale_blanket_order.py
StefanRijnhart: Would you consider reviewing that change and refactoring this module to work with it? You wouldn't have to jump through hoops like you do in the code above.

StefanRijnhart: I did the refactoring in https://github.com/ursais/sale-workflow/pull/14. What do you think?

alexey-pelykh (changes requested):
Thanks for working on this, tier validation for blanket orders is a useful addition.

However, since @StefanRijnhart's [base_tier_validation improvement](https://github.com/OCA/server-ux/pull/939) (support for computed state fields) was merged in October 2024, the full `_compute_state` override in this module should no longer be necessary. I'd strongly recommend refactoring to leverage that, as Stefan already suggested — and even prototyped in https://github.com/ursais/sale-workflow/pull/14. The current approach copies the entire upstream compute method, which is fragile and will silently break whenever `sale_blanket_order` updates its `_compute_state`.

Beyond that, there are a few concrete issues:

**Bug in `_compute_state`**: When `confirmed=True`, `validity_date > today`, and remaining qty is non-zero, no state gets assigned at all (the `else: order.state = "open"` was removed but nothing replaces it for that case). This leaves the state stale/undefined for active confirmed orders.

**`action_confirm` write ordering**: `rec.write({"state": "open"})` runs before `super().action_confirm()`, but since `state` is a computed field that depends on `confirmed`, the `super()` call setting `confirmed = True` will trigger recomputation, which — given the bug above — won't set state to `"open"` either.

**Docs reference wrong model**: `readme/CONFIGURE.md` says "Create as many tiers as you want for Purchase Order model" — should say Sale Blanket Order.

**`needs_review` filter domain**: References states `cancel` and `to approve` which don't exist on `sale.blanket.order` (valid states: draft, open, done, expired). This filter would match records it shouldn't.

**Missing copyright header** in `tests/__init__.py`.

The main ask: please rebase, adopt the computed state support from `base_tier_validation`, and drop the `_compute_state` override entirely. That should also resolve the state assignment bug and the `action_confirm` ordering issue naturally.

rousseldenis:

PR #3284, opened 2024-08-27
[17.0][MIG] sale_order_currency_rate: Migration to 17.0

SodexisTeam:


alexey-pelykh (changes requested):
Thanks for working on the migration. I found a few issues that need to be addressed before this can be merged.

**1. `currency_rate` field redefines and breaks the core Odoo 17 field**

In Odoo 17 (same as 16), `sale.order` already has a `currency_rate` field defined in core (`store=True, precompute=True`, depends on `currency_id`, `date_order`, `company_id`, uses `_get_conversion_rate`).

In the 16.0 version of this module, this was handled correctly -- only `inverse_currency_rate` was added, and the core `currency_rate` was reused as-is.

However, this PR redefines `currency_rate` as a non-stored computed field with a different computation method (`_convert` instead of `_get_conversion_rate`) and incomplete depends (missing `date_order`). This will:
- Override the stored core field with a non-stored one, breaking any code relying on it being stored
- Use a different computation direction/method than what core expects
- Miss recomputation triggers when `date_order` changes

The fix should be to drop the `currency_rate` field definition entirely (as in 16.0) and only keep `inverse_currency_rate`. The `_compute_currency_rate` method should be renamed to something like `_compute_inverse_currency_rate` and only compute `inverse_currency_rate`.

**2. `attrs` to inline `invisible` migration looks correct**

The view correctly migrates from `attrs="{'invisible': [...]}"` (16.0 syntax) to `invisible="..."` (17.0 syntax). This is good.

**3. Settings view XPath needs updating**

The 16.0 version uses `//div[@name='quotation_order_setting_container']` while the 17.0 version uses `//block[@name='quotation_order_setting_container']`. In Odoo 17 settings views, `div` was indeed changed to `block`, so this migration is correct. However, the settings view structure changed from `o_setting_box` divs to the `<setting>` tag in 17.0 -- this also looks properly adapted.

**4. Test file has incorrect license headers**

The test files (`tests/__init__.py` and `tests/test_sale_order_compute_currency_rate.py`) use `OPL-1` license which contradicts the module's `LGPL-3` license declared in the manifest. These should use the same LGPL-3 header as the rest of the module.

**5. Test uses `setUp` instead of `setUpClass`**

The 16.0 version correctly used `@classmethod` + `setUpClass`. The 17.0 migration changed this to `setUp` (instance method) which runs before every test method -- less efficient and not the OCA convention. Also the parameter is `cls` but it's actually `self` now which is misleading. Should be reverted to `setUpClass`/`cls`.

**6. Test file was renamed but the old test was more complete**

The 16.0 test (`test_sale_order_currency_rate.py`) tested the config setting integration and pricelist-driven currency changes. The new test (`test_sale_order_compute_currency_rate.py`) only tests the compute method directly and lost the configuration and pricelist integration coverage. Also, the 16.0 `tests/__init__.py` imported `test_sale_order_currency_rate` but the new one imports `test_sale_order_compute_currency_rate`, so the old test name is completely gone. Consider preserving the more comprehensive 16.0 test and adapting it to 17.0.

**7. `.pot` and `.po` files still reference `Odoo Server 16.0`**

The `Project-Id-Version` in `.pot` and `.po` files still says `Odoo Server 16.0`. These should be regenerated for 17.0 (typically handled by `oca-gen-addon-readme` / the OCA bot after merge, but worth noting).

sale_order_currency_rate/tests/test_sale_order_compute_currency_rate.py
```diff
@@ -0,0 +1,62 @@
+# copyright 2024 Sodexis
+# license OPL-1 (see license file for full copyright and licensing details).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderCurrencyRate(TransactionCase):
+    def setUp(cls):
+        super().setUp()
```
rousseldenis, line 9: Thanks for the tests ! Could you switch this to `classmethod` instead ?

rousseldenis (approved):
Ok, let's move forward

rousseldenis:

PR #3287, opened 2024-08-28
[FIX] sale_advance_payment: Residual Amount - Port PR#2738 to v17.0

ellbristow:
See PR https://github.com/OCA/sale-workflow/pull/2738

This is a port of that fix for 17.0

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -80,7 +80,10 @@ class SaleOrder(models.Model):
                     advance_amount += line_amount
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
-            for inv in order.invoice_ids:
+            # Filter out credit notes (returns)
+            for inv in order.invoice_ids.filtered(
+                lambda x: x.move_type == "out_invoice"
+            ):
                 invoice_paid_amount += inv.amount_total - inv.amount_residual
             amount_residual = order.amount_total - advance_amount - invoice_paid_amount
             payment_state = "not_paid"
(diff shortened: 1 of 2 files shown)


ellbristow:
@rousseldenis My previous PR got "corrupted" due to force pushes, so I closed it and created a new one. This one uses a cherry-pick.

However it's failing tests for something that wasn't in the commit

rousseldenis (approved):

rousseldenis:
@ellbristow Could you rebase?

alexey-pelykh (commented):
Thanks for working on this, @ellbristow.

Looking at the current state of the 17.0 branch, this fix has already been addressed through a different (and more robust) approach. Specifically:

- `900b1c5b` ("[FIX] sale_advance_payment: Use signed amount when computing residuals", Jan 2025) switched to `amount_total_in_currency_signed` which naturally handles credit notes correctly since the signed amount is negative for refunds.
- `60aeec64` ("[FIX] sale_advance_payment: fix invoice paid amount", Sep 2025) further refined the filter to exclude cancelled invoices and use invoice currency amounts.

The current 17.0 code already filters with `lambda x: x.state != "cancel"` and uses the signed amount, which is a better solution than filtering by `move_type == "out_invoice"` -- the signed amount approach correctly accounts for partial credit notes and mixed scenarios. There's also already a `test_05_residual_amount_credit_note` covering this case, with the correct assertion that after a full refund cycle the residual returns to the order total (3600), not 0.

This also explains why the ocabot rebase failed with conflicts in both `sale.py` and `test_sale_advance_payment.py` -- the underlying code has diverged significantly.

I believe this PR can be closed as the issue is resolved on 17.0.

rousseldenis:

PR #3297, opened 2024-09-02
[15.0][FIX] sale_elaboration, product_supplierinfo_for_customer_elaboration: Set elaboration note with a onchange instead of compute method

sergio-teruel:
cc @Tecnativa  TT50584

Use onchange method instead of compute because if any other data of any line force a recompute of all lines and the elaboration custom notes are reset.

ping  @carlosdauden @pilarvargas-tecnativa

### DIFF
--- a/product_supplierinfo_for_customer_elaboration/models/sale_order_line.py
+++ b/product_supplierinfo_for_customer_elaboration/models/sale_order_line.py
@@ -23,8 +23,20 @@ class SaleOrderLine(models.Model):
             if customer_info:
                 line.elaboration_ids = customer_info.elaboration_ids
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why?

carlosdauden (approved):

sergio-teruel:
If you read the chages the reason is explained

pedrobaeza:
Yeah, but you should provide the explanation in the commit message/main comment of the PR, apart from the docstring of a file.

pilarvargas-tecnativa (approved):
:+1:

sergio-teruel:
Ready

PR #3305, opened 2024-09-06
[17.0][FIX] sale_quotation_number: fix numbering to be compatible with sale_order_revision

CRogos:
This module is incompatible with sale_order_revision. New revision got always the next number instead of -02 suppliment.

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -13,20 +13,21 @@ class SaleOrder(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
-            if self.is_using_quotation_number(vals):
-                company_id = vals.get("company_id", self.env.company.id)
-                sequence = (
-                    self.with_company(company_id)
-                    .env["ir.sequence"]
-                    .next_by_code("sale.quotation")
-                )
-                vals["name"] = sequence or "/"
+            if not vals.get("name"):
+                if self.is_using_quotation_number(vals):
+                    company_id = vals.get("company_id", self.env.company.id)
+                    sequence = (
+                        self.with_company(company_id)
+                        .env["ir.sequence"]
+                        .next_by_code("sale.quotation")
+                    )
+                    vals["name"] = sequence or "/"
         return super().create(vals_list)
 
     @api.model
     def is_using_quotation_number(self, vals):
         company = False
-        if "company_id" in vals:
+        if vals.get("company_id"):
             company = self.env["res.company"].browse(vals.get("company_id"))
         else:
             company = self.env.company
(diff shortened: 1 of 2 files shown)


MohamedOsman7 (approved):
Functional & code review LGTM

CRogos:
@diggy128  could you also have a look on this change?

CRogos:
@RogerSans would you be so nice and add another review? You were involved in the migration to 17.

MohamedOsman7:
@MiquelRForgeFlow  Could you please make a review?

RogerSans (approved):

RogerSans:
Review LGTM

rousseldenis:
@CRogos Maybe adding a test case will help avoiding regressions ?

rousseldenis:
Oops, wrong button

sale_quotation_number/models/sale_order.py
```diff
@@ -13,14 +13,16 @@ class SaleOrder(models.Model):
...
-            if self.is_using_quotation_number(vals):
-                company_id = vals.get("company_id", self.env.company.id)
-                sequence = (
-                    self.with_company(company_id)
-                    .env["ir.sequence"]
-                    .next_by_code("sale.quotation")
-                )
-                vals["name"] = sequence or "/"
+            # Check if it's a revision copy and maintain the original name
+            if "name" not in vals:
```
zamberjo: How about checking that the value of name is not False or None?

```suggestion
            if not vals.get("name"):
```

rousseldenis (approved):

zamberjo (approved):

sale_quotation_number/models/sale_order.py
```diff
@@ -13,20 +13,22 @@ class SaleOrder(models.Model):
...
         for vals in vals_list:
-            if self.is_using_quotation_number(vals):
-                company_id = vals.get("company_id", self.env.company.id)
-                sequence = (
-                    self.with_company(company_id)
-                    .env["ir.sequence"]
-                    .next_by_code("sale.quotation")
-                )
-                vals["name"] = sequence or "/"
+            # Check if it's a revision copy and maintain the original name
```
rousseldenis: @CRogos  Don't put such comment as this module is not aware of sale_order_revision one.

rousseldenis: Something more like `if the sale order name is not provided, ...`

CRogos: I thought as explanation why this is important to keep. But I agree that the unit test is the better choice.

CRogos:
@rousseldenis are you able to merge?

rousseldenis (approved):

PR #3306, opened 2024-09-09
[IMP] sale_loyalty_exclude: Avoid to  select excluded products in rul…

miguelcb2003:
Avoid to  select excluded products in rewards

The excluded products can be selected at loyalty rewards. This is pointless and only makes confussion because the rewards would be ignored.

Besides, the "loyalty_exclude" check will no longer appear at product variant views.

[Grabación de pantalla desde 09-09-24 09:18:52.webm](https://github.com/user-attachments/assets/744bd60e-3316-4f2c-b146-9dc5fc6d3644)

FL-556-3740

### DIFF
--- a/sale_loyalty_exclude/models/__init__.py
+++ b/sale_loyalty_exclude/models/__init__.py
@@ -1,2 +1,3 @@
+from . import loyalty_reward
 from . import product_template
 from . import sale_order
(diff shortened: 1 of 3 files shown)


Pablocce (approved):
LGTM!

rousseldenis (approved):
LGTM

PR #3309, opened 2024-09-12
[16.0][PORT] sale_quotation_number: allow customizable quotation sequence

MohamedOsman7:
Same Changes: #3310

PR #3310, opened 2024-09-12
[17.0][IMP] sale_quotation_number: allow customizable quotation sequence

MohamedOsman7:
Enable users to customize the sale.quotation sequence prefix, instead of it being restricted to the default 'SQ'.

Solution from version 14: https://github.com/OCA/sale-workflow/pull/2126

(PR 16: #3309 )
(PR 15: #3311 )

(Follow-up PR: https://github.com/OCA/sale-workflow/pull/3213)

### DIFF
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -43,8 +43,11 @@ class SaleOrder(models.Model):
         return super().copy(default)
 
     def action_confirm(self):
+        sequence = self.env["ir.sequence"].search(
+            [("code", "=", "sale.quotation")], limit=1
+        )
         for order in self:
-            if self.name[:2] != "SQ":
+            if sequence and self.name[: len(sequence.prefix)] != sequence.prefix:
                 continue
             if order.state not in ("draft", "sent") or order.company_id.keep_name_so:
                 continue


CRogos (approved):
LGTM code + tested

rousseldenis:
@MohamedOsman7 Ok, next time, the better is to cheery-pick the original fix and then do the modification (the search before the loop) to keep original author. Thanks

CRogos:
@rousseldenis could you add a review and merge?

CRogos:
@pedrobaeza could you also merge this?

v17 of: https://github.com/OCA/sale-workflow/pull/3309

manuelregidor:
@MohamedOsman7 @CRogos @pedrobaeza @rousseldenis @ValentinVinagre @HaraldPanten 

This last improvement applied to versions from 14 up to 17 is not working properly.

First of all, there was an improvement already merged in v15 that solved the issues pointed out here. The PR was https://github.com/OCA/sale-workflow/pull/2719/files and there was a fix aftewards: https://github.com/OCA/sale-workflow/pull/2752/files. Is there any reason why these commits were not cherry-picked and another approach was developed instead?

Second, we think the approach previously available in v15 is more convenient, for two reasons basically:

1. The approach in this PR does not work if the quotation's sequence uses a formula in its prefix. For example, if the sequences's prefix is "Q/%(year)s/" the following line will not work as expected:
`if sequence and self.name[: len(sequence.prefix)] != sequence.prefix:`
As the quotation number will contain the year itself (and not the string "%(year)"), the sequence will be not changed to the sale order sequence when the quotation gets confirmed.

2. The approach in this PR only considers the sequences with codes "sale.quotation" and "sale.order", whereas the approach previsouly available in v15 uses methods get_quotation_seq and get_sale_order_seq to get the sequence to be applied. Those methods can easily be inherited so other sequences can be used depending on different requirements (for example, in module sale_order_type_quotation_numer https://github.com/OCA/sale-workflow/tree/15.0/sale_order_type_quotation_number).

Some of our clients currently use this module in v15 and are having issues as they use a formula in the quotation sequence prefix. We should find a way to proceed to get this sorted, as the old v15 of the module with an improvement, which worked fine, has stopped working and the new version of the module in the rest of versions will cause problems.

Thank you.

HaraldPanten:
IMO:

- I think we should revert this last commit (At least in V15). It has been done in a stable version and it's not needed there.
- Then we should discuss about the best approach for V16 and V17.

WDYT?

HaraldPanten:
Solving the issue in V15 before discussing the best solution.

https://github.com/OCA/sale-workflow/pull/3416

PR #3314, opened 2024-09-16
[14.0][FIX] sale_order_line_input: access to action views

toita86:
Previously of this fix when a user with access to all the SO tries to access one of them through the "Open Order" widget there will produce an error:
image

### DIFF
--- a/sale_order_line_input/models/sale_order.py
+++ b/sale_order_line_input/models/sale_order.py
@@ -54,9 +54,8 @@ class SaleOrderLine(models.Model):
 
     def action_sale_order_form(self):
         self.ensure_one()
-        action = self.env.ref("sale.action_orders")
+        action = self.env["ir.actions.act_window"]._for_xml_id("sale.action_orders")
         form = self.env.ref("sale.view_order_form")
-        action = action.read()[0]
         action["views"] = [(form.id, "form")]
         action["res_id"] = self.order_id.id
         return action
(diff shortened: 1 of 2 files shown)


sale_order_line_input/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_window_action_form_user,access_window_action_form_user,base.model_ir_actions_act_window,sale_order_line_input.group_sale_order_access,1,0,0,0
```
rousseldenis: @toita86 You cannot do that as it will give you access to the whole model.

The problem is here: https://github.com/OCA/sale-workflow/blob/14.0/sale_order_line_input/models/sale_order.py#L57

You should change the code to retrieve the action through `for_xml_id`function. Look into core code.

rousseldenis (changes requested):

sale_order_line_input/security/ir.model.access.csv
toita86: Thank you for the explanation!
I have changed my approach based on your suggestion, now should be correct.

GSLabIt (approved):
code

francesco-ooops (approved):
Functional ok!

@rousseldenis what do you say?

francesco-ooops:
@OCA/crm-sales-marketing-maintainers can this be merged? Thanks!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers ping :)

simahawk (approved):

francesco-ooops:
@toita86 I think this was fixed in v15 and later here: https://github.com/OCA/sale-workflow/pull/2503

so no need for fw-porting

PR #3315, opened 2024-09-17
[15.0][FIX] sale_fully_invoiced: tests not being called

AaronHForgeFlow:
I wonder why this is not giving any errors

@ForgeFlow

BTW: the tests still make sense, the invoice status does not depend in the delivery status, indeed, this module does not depend on sale_stock

PR #3316, opened 2024-09-17
[16.0][MIG] sale_fully_invoiced: Migration to v16

AaronHForgeFlow:
No changes needed, just fix in tests, same as in 15.0:  https://github.com/OCA/sale-workflow/pull/3315

Also added myself as maintainer

@ForgeFlow


sale_fully_invoiced/models/sale.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    @api.depends("invoice_status", "invoice_ids.state")
+    def _compute_invoice_status_validated(self):
+        # Same as invoice status, execpt it is not invoiced unless there is no invoice
+        # in draft
+        for order in self:
+            if order.invoice_status in ("no", "to invoice"):
+                order.invoice_status_validated = order.invoice_status
+            elif order.invoice_status in ("no", "upselling"):
```
rousseldenis: I'm quite bugging on this line...

If the order status is "no" on the first `if`, then, the `invoice_status_validated`== `order.invoice_status`== `no`

So, the `elif`should be `elif order.invoice_status == "upselling"` 

No ?

HaraldPanten:
You need to change the PR name and the migration commit message: full --> fully 😝

AaronHForgeFlow:
Thank you for the review. I need to improve the tests for the upselling case. Also fix the module in v15: https://github.com/OCA/sale-workflow/blob/15.0/sale_fully_invoiced/models/sale.py#L29

rousseldenis (approved):

BhaveshHeliconia (approved):
Functional review LGTM!

PR #3317, opened 2024-09-19
[17.0][ADD] sale_stock_dropshipping_note: new module

vvrossem:
This glue module between `stock_dropshipping` and `sale_stock_picking_note` transfers the external (customer) and internal picking note from the sales order to the dropship associated.


vvrossem:
closing as no added value. It's enough to install `stock_dropshipping` to have `sale_stock_picking_note` working as expected with Dropship Orders

PR #3319, opened 2024-09-20
[FIX] pricelist_cache: item expiration fix

henrybackman:
The pricelist cache was not taking into account the `date_end` and `date_start` of items when calculating the prices. This caused outdated prices to be shown in the pricelist cache.

Fixed by modifying the SQL query that finds the parent pricelists to check the item expiration.

### DIFF
--- a/pricelist_cache/models/product_pricelist.py
+++ b/pricelist_cache/models/product_pricelist.py
@@ -141,15 +141,20 @@ class Pricelist(models.Model):
         globally, and based on another pricelist
         """
         self.ensure_one()
+        today = str(date.today())
+        self.flush()
         query = """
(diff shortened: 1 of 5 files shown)


simahawk:
> The pricelist cache was not taking into account the `date_end` and `date_start` of items when calculating the prices. This caused outdated prices to be shown in the pricelist cache.
> 
> Fixed by modifying the SQL query that finds the parent pricelists to check the item expiration.

This should be in the commit description (which next time will give you an automatic description of the PR) :wink:

simahawk (approved):
LG, minor remark

mmequignon (approved):

PR #3321, opened 2024-09-24
[16.0][FIX] sale_order_type: Avoid unwanted triggering of compute functions

grindtildeath:
Having account.move._compute_journal_id overriden with an added dependency to sale_type_id is messing with computation triggers, because compute functions depending on journal_id will get recomputed, potentially overwriting existing value, as is the case with bank account on account.payment.

Use instead the hook to search for default journal, in order to return the journal defined on sale order type to its compute function.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -48,6 +48,11 @@ class AccountMove(models.Model):
                 )
                 if sale_type:
                     record.sale_type_id = sale_type
+            # Add manually field to compute, as using api.depends on its compute function would
+            #  trigger unwanted recomputation of partner_bank_id on vendor bills
+            self.env.add_to_compute(
+                self.env["account.move"]._fields["journal_id"], record
+            )
 
     @api.depends("sale_type_id")
     def _compute_invoice_payment_term_id(self):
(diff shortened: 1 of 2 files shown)


sale_order_type/models/account_move.py
```diff
@@ -48,6 +48,11 @@ def _compute_sale_type_id(self):
                 )
                 if sale_type:
                     record.sale_type_id = sale_type
+            # Add manually field to compute, as using api.depends on its compute function would
+            #  trigger unwanted recomputation of partner_bank_id on vendor bills
+            self.env.add_to_compute(
+                self.env["account.move"]._fields["journal_id"], record
+            )
```
ivantodorovich, line 55: Careful, AFAIK, when using `add_to_compute`, the field dependencies aren't recomputed unless you explicitly call `modified`, too.

henrybackman (approved):

grindtildeath:
Since Odoo accepted my fix and fast tracked it, I'm not sure if this change is still needed :question:

sale_order_type/models/account_move.py
rousseldenis: IMHO, this is more a bad core architecture as `partner_bank_id` field is computed/stored and not its dependency `bank_partner_id`.

rousseldenis: Have you tried to change the depends of `partner_bank_id`field using `commercial_partner_id`instead ?

https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L1315

grindtildeath:
IMO we can close since https://github.com/odoo/odoo/pull/181278 is merged

PR #3325, opened 2024-09-25
[16.0][FIX] sale_triple_discount: Remove call to deprecated

lmignon:
Method 'invalidate_cache' is deprecated. To invalidate specific field's value into the cache for a recordset we must use the 'invalidate_recordset' method on the recordset itself.


rousseldenis (approved):
Trivial

sbejaoui (approved):

glitchov (approved):
ok

PR #3327, opened 2024-09-25
[14.0][IMP] sale_exception: sale price unit warning

DiegoParadeda:


mileo (changes requested):
image

The price compared may not be correct due to taxes.

Issue #3323, opened 2024-09-25
[16.0] - sale_triple_discount - updating discounts not working after sale.order.line creation


IJOL:
## Module
sale_triple_discount

## Describe the bug
After line creation updating discounts is not working the previous value is reapplied 

1. Create a valid sale.order.line applying some discounts 
2. Try to update any of them
3. try to save the updates, line will recover previous unmodified state

**Expected behavior**
Been able to update discounts after line creation

IJOL:
https://github.com/OCA/sale-workflow/pull/3324 seems to solve it, but it maybe not totally ok

IJOL:
a better solution in https://github.com/OCA/sale-workflow/pull/3329

PR #3324, opened 2024-09-25
[16.0] - sale_triple_discount  - [BUG] preprocess cache invalidation seems to make updates impossible

IJOL:
I know is really naive but until now it solves the issue AFAIK

PR #3329, opened 2024-09-26
[16.0][IMP] sale_triple_discount: Improves perfs and consistency

lmignon:
When computing the amount total on a sale.order.line, we now avoid to play with the cache since this change introduced in https://github.com/OCA/sale-workflow/commit/e3e59ba15ebcf1f91124eda60a694c0a76fabb69 introduces a decline in performance. When the cache in invalidated, fields are flushed to the database and depending computed fields are recomputed..... This ultimately leads to temporary inconsistencies breaking others addons relaying on this one. Moreover, the method used in this original commit 'self.invalidate_cache(fnames=self._discount_fields(), ids=self.ids)' is deprecated and the use of the recommended new one lead to others side effects. The original approach is replaced by a contextual method modifying only the discount field when we need it's needed to don't break code relaying on the fact that the discount field is the total discount to apply to a line. This contextual method ensures that the original value is properly restored on exit and that it will not trigger additional recompute due to the temporary change of the discount value

### DIFF
--- a/sale_triple_discount/models/sale_order_line.py
+++ b/sale_triple_discount/models/sale_order_line.py
@@ -4,6 +4,8 @@
 # Copyright 2018 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from contextlib import contextmanager
+
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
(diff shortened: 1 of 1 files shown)


lmignon:
ping @renda-dev @toita86 @technativa

renda-dev (approved):
This solution is really clean and I love it.

Code review, LGTM!

toita86 (approved):
Code review, LGTM.

IJOL:
Just tested this solution and works nicely and is much more clean and less magic, i'm all about less roundtrips and no cache messing, although my latest iteration in https://github.com/OCA/sale-workflow/pull/3324 worked nicely too, I much favor this one over ours.

And closes  https://github.com/OCA/sale-workflow/issues/3323

francesco-ooops:
@IJOL approve please :)

IJOL (approved):

IJOL (approved):
Ok

mariancuadranetkia:
Tested and it recovers funcionality
@pedrobaeza  how do you see this fix?

mariancuadranetkia (approved):

pedrobaeza:
I'm not using the module, but with current review status + coming from a trusted contributor, let's merge it:

/ocabot merge patch

PR #3340, opened 2024-10-01
[14.0][FIX] sale_triple_discount: Fixed tax amount on report

toita86:


francesco-ooops (approved):
Functional ok!

This is solving an issue (seems exclusive to v14) where "taxes" amount on SO report was not taking into account discounts 2 and 3

Taxes on SO form are correct, issue is only on report

francesco-ooops:
@pedrobaeza could you take a look? thanks!

pedrobaeza:
Hi from Odoo XP. The changes seems so broad for me not knowing the module. Other qualified reviewer should check.

francesco-ooops:
@lmignon @grindtildeath could I ask you to take a look? thanks!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers ping :)

StefanRijnhart (commented):
I see that the method has indeed been rewritten and renamed in 14: https://github.com/odoo/odoo/commit/149eb38b07cace0ec2118d214c56995402008efc

The only change is the injection of the discounted price, and it is marked as such in the method, so agreed in principle with the change.

Small remark: the call to super is void as all the values are overwritten. I would prefer to see it replaced by a fat warning saying this is an overwrite of the method in sale in the method's docstring.

toita86:
@StefanRijnhart Thank you for the review. I have removed the super() call and added a docstring. Please let me know if everything looks good now.

PicchiSeba (approved):
Code Review: OK

StefanRijnhart:
Thanks for the update! Forgot to ask, but would it be possible to add a small test to check the result of the method for an order with triple discounts? It can probably be a single line change in `test_03_sale_order_complex_triple_discount`.

toita86:
@StefanRijnhart Thanks for the suggestion! I've added a test to verify the result of the `_amount_by_group` method after each discount is added.

Let me know if any further adjustments are needed!

StefanRijnhart (approved):
Excellent, thanks for the added test!

For the record, I see that the overridden method is no longer present in Odoo 15 so no forward porting seems necessary.

PR #3347, opened 2024-10-09
[FIX] sale_sourced_by_line: fix conflict with precompute

anikeenko-viktor:
Built-in filed is marked as precompute=True.
When field is marked as precomputed and readonly, Odoo will remove it's value from create values. (See `odoo.models.BaseModel._prepare_create_values`) Therefore, it was not possible to set `warehouse_id` in create as it would reset to default computed warehouse.

Now, it's fixed and readonly enforced on view level.

### DIFF
--- a/sale_sourced_by_line/__manifest__.py
+++ b/sale_sourced_by_line/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Sale Sourced by Line",
     "summary": "Multiple warehouse source locations for Sale order",
-    "version": "17.0.1.0.0",
+    "version": "17.0.1.0.1",
     "author": "Camptocamp,"
     "Eficent,"
     "SerpentCS,"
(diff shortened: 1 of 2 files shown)


alexey-pelykh (approved):
Thanks for the fix.

The root cause is clear: upstream `warehouse_id` on `sale.order` is `precompute=True, store=True, readonly=False`, but this module was overriding it with `readonly=True`. That combination causes `_prepare_create_values` to strip the value from create vals, making it impossible to set a per-line warehouse at creation time.

Removing `readonly=True` from the Python field definitions is the right approach -- the view XML already enforces `readonly="state not in ['draft', 'sent']"` for the UI, and the existing tests (which create SO lines with explicit `warehouse_id`) effectively serve as regression coverage for this fix.

LGTM.

rousseldenis (approved):
Seems legit

PR #3351, opened 2024-10-10
[16.0][IMP] sale_loyalty_exclude

Pablocce:
Description: 
The functionality provided by the module allows to set which products are excluded from promotions in a specific field. Currently, if there is a promotion for a product and it has the "Loyalty_exclusion" field as true, the claim reward button is displayed on the e-commerce site, but when clicked, it shows an error preventing the promotion from being added.

With this PR, the goal is to avoid displaying this button altogether in such cases.

Solution: 
To resolve this problem, we have made the following modifications:

- Inherited method that returns the available promotions to filter them based on the products in the cart.
- Inherited method that applies discount codes to control their application based on the products in the cart.

This fix will prevent contradictory information from being displayed on the e-commerce site.

### DIFF
--- a/sale_loyalty_exclude/i18n/es.po
+++ b/sale_loyalty_exclude/i18n/es.po
@@ -19,7 +19,7 @@ msgstr ""
 #. module: sale_loyalty_exclude
 #: model:ir.model.fields,field_description:sale_loyalty_exclude.field_loyalty_reward__discount_product_ids
 msgid "Discounted Products"
-msgstr ""
+msgstr "Productos con descuento"
 
 #. module: sale_loyalty_exclude
 #: model_terms:ir.ui.view,arch_db:sale_loyalty_exclude.product_template_form_view
@@ -35,7 +35,7 @@ msgstr "Excluir producto en programas de descuento"
 #. module: sale_loyalty_exclude
 #: model:ir.model,name:sale_loyalty_exclude.model_loyalty_reward
 msgid "Loyalty Reward"
-msgstr ""
+msgstr "Recompensa de fidelidad"
 
 #. module: sale_loyalty_exclude
 #: model:ir.model,name:sale_loyalty_exclude.model_product_template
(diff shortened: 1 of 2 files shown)


danielduqma (approved):

alejandro-vallejoFL (approved):

rousseldenis:
@Pablocce Could you rebase ?

sale_loyalty_exclude/i18n/es.po
```diff
@@ -19,7 +19,7 @@ msgstr ""
 #. module: sale_loyalty_exclude
 #: model:ir.model.fields,field_description:sale_loyalty_exclude.field_loyalty_reward__discount_product_ids
 msgid "Discounted Products"
-msgstr ""
+msgstr "Productos con descuento"
```
rousseldenis, line 22: Please don't include translations here. Do it through weblate

rousseldenis (commented):
Could you maybe add tests to avoid regression ? Thanks

rousseldenis:
@Pablocce @danielduqma Do you plan to attend comments ?

alexey-pelykh (commented):
Thanks for working on this -- hiding the reward button when it can't actually apply is a real UX improvement.

A few points to consider:

**1. Outstanding reviewer feedback not addressed**

@rousseldenis already requested a rebase, removal of the `es.po` translation changes (those should go through Weblate), and test coverage. These still seem pending. I'd suggest addressing those first since the PR appears stalled on them.

**2. Filtering logic in `_has_all_products_loyalty_exclude`**

The method iterates over `self.order_line` which includes all line types (reward lines, sections, notes, delivery, etc.). A note or section line with no product would have `product_id.loyalty_exclude` evaluate to `False`, so the method would return `False` even if every real product is excluded. Consider filtering to only product lines, e.g.:

```python
def _has_all_products_loyalty_exclude(self):
    product_lines = self.order_line.filtered(lambda l: l.product_id and not l.is_reward_line)
    return product_lines and all(l.product_id.loyalty_exclude for l in product_lines)
```

This also handles the empty-order edge case (no product lines should not return `True`).

**3. Error message in `_try_apply_program`**

`"This code is invalid."` is misleading -- the code itself may be perfectly valid, but the products in the cart are excluded from promotions. Something like `"No applicable products in the order for this program."` would be more accurate and less confusing for the end user.

**4. Minor style**

The `else` after a `return` is unnecessary in both `_get_claimable_rewards` and `_try_apply_program`:

```python
def _get_claimable_rewards(self, forced_coupons=None):
    if self._has_all_products_loyalty_exclude():
        return {}
    return super()._get_claimable_rewards(forced_coupons=forced_coupons)
```

**5. Tests**

+1 to @rousseldenis's request -- tests covering both the "all excluded" and "mixed" scenarios would be valuable here to prevent regressions.

rousseldenis:

PR #3353, opened 2024-10-11
[16.0] [MIG] sale_stock_delivery_address

SimaElisabeta:


rrebollo (commented):
Hi @SimaElisabeta thanks for contributing. Apart from missing tests I think is a great job but ... I would love a deeper level of details in README. It's not a matter of taste look: while I was reviewing this, I was also reviewing #3347. The procurement feature looks at some point similar to me. U can check the code. Then I wonder if there is common ground. Can these two addons work together? Are they exclusive? I know is only a migration but I think a bit of documentation by both sides would be an added value in both cases as a first step. Then it should be easier to define whether there is place to further improvements.

rrebollo:
> Hi @SimaElisabeta thanks for contributing. Apart from missing tests I think is a great job but ... I would love a deeper level of details in README. It's not a matter of taste look: while I was reviewing this, I was also reviewing #3347. The procurement feature looks at some point similar to me. U can check the code. Then I wonder if there is common ground. Can these two addons work together? Are they exclusive? I know is only a migration but I think a bit of documentation by both sides would be an added value in both cases as a first step. Then it should be easier to define whether there is place to further improvements.

@anikeenko-viktor :point_up:

alexey-pelykh (changes requested):
Thanks for the migration, @SimaElisabeta. The test coverage looks solid and the 16.0-specific model/field renames (`stock.route`, `location_dest_id`) are handled well. I did find a couple of issues that need attention though:

**`sale_order.py` -- `_prepare_procurement_group_by_line` no longer exists**

The method `_prepare_procurement_group_by_line` was present on `SaleOrder` in older versions of `sale_procurement_group_by_line` (v12/v13), but it was removed and refactored into `_prepare_procurement_group_vals()` on `SaleOrderLine` by the time v15 came around. In the 16.0 branch of `sale_procurement_group_by_line`, this method does not exist at all -- the `super()` call in `sale_order.py` will raise an `AttributeError` at runtime.

Additionally, even if the method did exist, the condition `line._get_procurement_group_key()[0] == 16` would never be `True` since `sale_order_line.py` sets `priority = 15`. So there's a priority mismatch on top of the missing method.

This file needs to be reworked. The procurement group name customization should likely move to an override of `_prepare_procurement_group_vals()` on `SaleOrderLine`, or be handled via a different hook appropriate for the 16.0 API.

**`sale_order_line.py` -- behavioral change in `_get_procurement_group_key`**

The 15.0 version returned the original parent key when `dest_address_id` was not set:
```python
if key[0] < priority:
    if self.dest_address_id:
        return (priority, self.dest_address_id)
return key
```

The 16.0 version now wraps the key:
```python
if self.dest_address_id:
    return (priority, self.dest_address_id)
return (priority, key)
```

That final `return (priority, key)` changes the grouping behavior for lines _without_ a destination address -- they now get priority 15 instead of whatever the parent returned. Was this intentional? If not, the last line should just be `return key` to preserve the 15.0 behavior.

**Minor: test `setUp` vs `setUpClass`**

OCA 16.0 convention prefers `setUpClass` with `@classmethod` for test data setup (more efficient since the data is created once per class rather than per test method). Worth updating while you're at it.

rousseldenis:

PR #3354, opened 2024-10-11
[16.0][FIX] sale_order_type: assign base company in non-multicompany environments

iTecan:
We have modified the following company field in data/default_type.xml from:

        <field name="company_id" eval="False" />
To:

        <field name="company_id" ref="base.main_company" />

This change is due to the fact that in an environment without the multi-company parameter, the warehouse field cannot be filled because it is not assigned to a company.

### DIFF
--- a/sale_order_type/data/default_type.xml
+++ b/sale_order_type/data/default_type.xml
@@ -1,7 +1,7 @@
 <odoo noupdate="1">
     <record id="normal_sale_type" model="sale.order.type">
         <field name="name">Normal Order</field>
-        <field name="company_id" eval="False" />
+        <field name="company_id" ref="base.main_company" />
         <field name="sequence_id" ref="sale.seq_sale_order" />
     </record>
 </odoo>


sale_order_type/data/default_type.xml
```diff
@@ -1,7 +1,7 @@
 <odoo noupdate="1">
     <record id="normal_sale_type" model="sale.order.type">
         <field name="name">Normal Order</field>
-        <field name="company_id" eval="False" />
+        <field name="company_id" ref="base.main_company" />
```
rousseldenis, line 4: @iTecan This is not correct.

As there is this record rule : https://github.com/OCA/sale-workflow/blob/16.0/sale_order_type/security/security.xml#L7

So, if no company is defined, everybody can see it and it will be available of course in non-multicompany environments

StefanRijnhart: To be fair, so many fields provided by the sale order type (journal/warehouse/payment term/pricelist) do require a company that order types without company are probably not used in practice.

alexey-pelykh (commented):
Thanks for the contribution. I see the point about `check_company` constraints making a company-less type impractical — @StefanRijnhart's observation is spot on there.

However, hardcoding `ref="base.main_company"` introduces a different problem: on multi-company databases, the default "Normal Order" type would be pinned to company 1 at install time. Users in other companies would not get this default type (the `_default_type_id` search filters by `self.env.company`), so they'd need to manually create their own, which is a worse out-of-the-box experience than today.

The current `eval="False"` + model default of `self.env.company` already handles the common case: when users create types via the UI, they get the correct company assigned automatically. The XML record is `noupdate="1"`, so it only matters at install. If the goal is to ensure the initial record also gets a company, a `post_init_hook` that sets `company_id = self.env.company` would respect the installing user's company context rather than always picking `base.main_company`.

That said, the warehouse linkage issue is real — it would be worth checking whether the root cause is better addressed by ensuring the `onchange`/compute logic gracefully handles `company_id=False` on the type, rather than hardcoding a company on the demo data.

rousseldenis:

PR #3357, opened 2024-10-15
[16.0][ADD]sale_order_line_multi_warehouse: Split sale order in multiple pickings

manuelregidor:
Split sale order lines into multiple sub-lines assigned to different warehouses. When a sale order is confirmed, multiple pickings can be created depending on the warehouses selected.

@ValentinVinagre @HaraldPanten @luis-ron 

T-6563


pedrobaeza:
Isn't this the same as https://github.com/OCA/sale-workflow/tree/16.0/sale_sourced_by_line ?

manuelregidor:
@pedrobaeza Not exactly. sale_sourced_by_line lets you select a warehouse in a sale order line. Our module lets you split a single sale order line into multiple sublines, each one related to a different warehouse.

pedrobaeza:
I think there's a shared code on how to do such split when confirming the sales order, so it should be at least an expansion module. cc @rousseldenis

HaraldPanten:
> I think there's a shared code on how to do such split when confirming the sales order, so it should be at least an expansion module. cc @rousseldenis

Hi Pedro, thanks for your comments.

Could you give us some examples of the shared code that you mention? The modules have different purposes and we don't want to create a "too complex" module, but we are open to improve our work. You know...

pedrobaeza:
I was just talking from the "feature" perspective, as both should split the pickings (at procurement level). I haven't dug deeper, as I'm not using any of them. If you think there's no overlap, OK then.

HaraldPanten:
> I was just talking from the "feature" perspective, as both should split the pickings (at procurement level). I haven't dug deeper, as I'm not using any of them. If you think there's no overlap, OK then.

@manuelregidor @ValentinVinagre  Could you deeper analise Pedro's comments and verify that we're not overlaping anything? As we're not using sale_sourced_by_line maybe there's something that we are not taking in consideration.

Denis's comments would be appreciated, as well 😉

THX!

manuelregidor:
@pedrobaeza It is kind of similar from a concept perspective. However, in this proposal not only a sale order line can be assigned to a specific warehouse, but they can also be split into separate warehouses. Because of that, it has been necessary to create a new model (sale.order.line.warehouse). Its class instances are related to a sale order line, and a sale order line can have multiple instances of sale.order.line.warehouse, so the logic to separate the sale order into multiple warehouses follows a different and much more complex path, as it shifts from sale.order.line model to sale.order.line.warehouse model.

Basically, in module sale_sourced_by_line the way of groupping lines in multiple procurement groups is done in the sale order line level, so using the methods to apply additional rules (a warehouse assignment in this case) is enough to group the sale order lines into different pickings. In our proposal, we have to achieve the separation from another level.

That is why an expansion module would be possible, yet the methods in sale_sourced_by_line module would be strongly modified as they should be adapted to the sale.order.line.warehouse level. Another issue is that module sale_sourced_by_line depends on sale_procurement_group_by_line, which contains methods also used in this module; again, those inheritances would be pretty much ignored for the same reason.

pedrobaeza:
OK, thanks for the extra explanations.

etobella (approved):
Functionaly tested and seems to work properly.

However, I would prefer to see an specific widget on the warehouse selection :wink:

sale_order_line_multi_warehouse/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 Manuel Regidor <manuel.regidor@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale Order Line Multi Warehouse",
+    "summary": "Sale Order Line Multi Warehouse",
+    "version": "16.0.1.0.0",
+    "category": "Uncategorized",
```
ValentinVinagre: "category": "Sales", please change

sale_order_line_multi_warehouse/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "website": "https://github.com/OCA/sale-workflow",
```
ValentinVinagre, line 10: add 'development_status'

ValentinVinagre (commented):
some changes

luis-ron (approved):
Functional review: LGTM 👍🏻

sale_order_line_multi_warehouse/README.rst
```diff
@@ -0,0 +1,115 @@
...
+
+To activate the sale orders multi-warehouse options you need to:
+
+#. Go to Inventory > Configuration > Settings.
+#. Activate the "Multi Warehouse in Sale Orders" option, under the "Warehouse" section. This field is only visible for users in group "Inventory / Administrator". When a sale order is created, multi warehouse options in sale order lines will be enabled/disabled depending on the value in this field on creation.
+
+It is necessary to set alternative warehouses to a warehouse so they can be selected in sale order lines when the warehouse is set in the "Warehouse" field in a sale order. To set alternative warehouses you need to:
+
+#. Go to Inventory > Configuration > Warehouses and see a warehouse's form view.
+#. Select the warehouse's alternative warehouses in the "Altenative Warehouses" field. This field is only visible for users in group "Inventory / Administrator".
```
Tisho99, line 49: Group stock.group_stock_multi_warehouses is also needed

sale_order_line_multi_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,69 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    allow_sale_multi_warehouse = fields.Boolean(
+        compute="_compute_allow_sale_multi_warehouse", store=True
```
Tisho99: Why not related?

sale_order_line_multi_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,69 @@
...
+            order.suitable_warehouse_ids = (
+                order.warehouse_id + order.warehouse_id.alternative_warehouse_ids
+            )
+
+    def get_incompatible_multi_warehouse_lines(self, warehouse):
+        self.ensure_one()
+        compatible_warehouse_ids = warehouse + warehouse.alternative_warehouse_ids
+        return self.order_line.mapped("sale_order_line_warehouse_ids").filtered(
+            lambda a: a.warehouse_id.id not in compatible_warehouse_ids.ids
+        )
```
Tisho99, line 39: I think you can use:

`return self.order_line.sale_order_line_warehouse_ids - self.suitable_warehouse_ids`

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    sale_order_line_warehouse_ids = fields.One2many(
+        string="SO Line Warehouses",
+        comodel_name="sale.order.line.warehouse",
+        inverse_name="order_line_id",
+    )
+    qty_assigned_to_warehouse = fields.Float(
```
Tisho99, line 16: Did you thought making the product_uom_qty computed instead of creating this field, and using the compute + compute_reverse function instead of overriding create and write functions?

sale_order_line_multi_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,67 @@
...
+        compatible_warehouse_ids = warehouse + warehouse.alternative_warehouse_ids
+        return self - compatible_warehouse_ids
+
+    def _action_confirm(self):
+        for order in self:
+            if order.get_incompatible_multi_warehouse_lines(self.warehouse_id):
+                raise ValidationError(
+                    _(
+                        "Some sale order line assignment lines are not compatible with "
+                        "the selected warehouse. Quotation {} could not be validated.".format(
```
rousseldenis: This is not correct. Please use translation function arguments instead

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+    suitable_warehouse_ids = fields.Many2many(
+        strting="Suitable Warehouses",
+        comodel_name="stock.warehouse",
+        related="order_id.suitable_warehouse_ids",
+    )
+
+    @api.constrains("sale_order_line_warehouse_ids")
+    def _check_warehouses(self):
+        for line in self:
+            warehouse_lines_group = self.env["sale.order.line.warehouse"].read_group(
```
rousseldenis: Can't you do this outside the loop ?

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        # Automatically create a warehouse distributions line when the
+        # sale order allows multi warehouse.
+        for line in lines.filtered("allow_sale_multi_warehouse"):
+            if line.move_ids:
+                # This part is executed when a sale order line is created by adding
+                # a stock move in a picking related to the sale order.
+                self.env["sale.order.line.warehouse"].create(
```
rousseldenis: I would have build a list of arguments to create records in batch after the loop

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+        return self.sale_order_line_warehouse_ids.filtered(
+            lambda a: a.warehouse_id == self.order_id.warehouse_id
+        )
+
+    def adjust_qty_assigned_to_warehouse_decrease(self, qty):
+        self.ensure_one()
+        reduce_line = (
+            self._get_adjustment_default_warehouse_line()
+            or self.sale_order_line_warehouse_ids.filtered(
+                lambda a: a.product_uom_qty > 0.0
```
rousseldenis: Using `float_compare()` is more correct.

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+            )
+        )
+        pending_qty = qty
+        while pending_qty > 0.0:
```
rousseldenis: Same as above.

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,266 @@
...
+        return ret_vals
+
+    def action_show_warehouse_lines(self):
+        self.ensure_one()
+        view = self.env.ref(
+            "sale_order_line_multi_warehouse.sale_order_line_warehouse_view"
+        )
+
+        if self.product_id.tracking == "serial" and self.state == "assigned":
+            self.next_serial = self.env["stock.lot"]._get_next_serial(
```
rousseldenis: Mmmmh, I'm not sure this should be in this module.

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_line_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo import fields, models
+
+
+class SOMultiWarehouseChangeLineWizard(models.TransientModel):
+    _name = "so.multi.warehouse.change.line.wizard"
+    _description = "Sale Order Multi Warehouse Change Line Wizard"
+
+    so_multi_warehouse_change_wizard_id = fields.Many2one(
+        string="SO Multi Warehouse Change Wizard",
+        comodel_name="so.multi.warehouse.change.wizard",
```
rousseldenis, line 13:
```suggestion
        comodel_name="so.multi.warehouse.change.wizard",
        required=True,
        ondelete="cascade"
```

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_line_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    sale_order_line_warehouse_id = fields.Many2one(
+        string="Current Warehouse",
```
rousseldenis, line 18: Shouldn't it be required too ?

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
```diff
@@ -0,0 +1,78 @@
...
+                    sel.sale_order_id.get_incompatible_multi_warehouse_lines(
+                        sel.new_warehouse_id
+                    )
+                )
+                if incompatible_lines:
+                    for incompatible_line in incompatible_lines:
+                        vals.append(
+                            {"sale_order_line_warehouse_id": incompatible_line.id}
+                        )
+                    lines = lines.create(vals)
```
rousseldenis: It's not correct to call a `create()` in a `compute()`

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
```diff
@@ -0,0 +1,78 @@
...
+            for line in self.so_multi_warehouse_change_line_ids:
+                so_line_warehouse = line.sale_order_line_warehouse_id
+                related_lines = (
+                    so_line_warehouse.order_line_id.sale_order_line_warehouse_ids
+                )
+                same_warehouse_line = related_lines.filtered(
+                    lambda a: a.warehouse_id == self.new_warehouse_id
+                )
+                if same_warehouse_line:
+                    same_warehouse_line[0].write(
```
rousseldenis: And if several same lines ?

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
```diff
@@ -0,0 +1,78 @@
...
+                    line.sale_order_line_warehouse_id.unlink()
+                else:
+                    line.sale_order_line_warehouse_id.write(
+                        {"warehouse_id": self.new_warehouse_id.id}
+                    )
+            self.sale_order_id.write({"warehouse_id": self.new_warehouse_id.id})
+        else:
+            raise ValidationError(
+                _(
+                    "Warehouse cannot be changed as sale order is in {} state.".format(
```
rousseldenis: Please use translate function arguments instead.

rousseldenis (changes requested):
I would say this module can be great. i would not say this is much simpler than `sale_sourced_by_line`...

I haven't tested it functionally yet.

But some refinements in code should be done.

rousseldenis:
Using float_compare everywhere a field with decimal precision is used should be done.

manuelregidor:
@rousseldenis Thanks for your feedback. I'll have a look as soon as I can.

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
manuelregidor: There should not be more than 1 as the _check_warehouses constraint in sale.order.line avoids that.

Tisho99 (approved):
LGTM

manuelregidor:
@rousseldenis @pedrobaeza This module is not compatible with sale_procurement_group_by_line. In the future, we could develope a module to make them compatible. However, we'd like to know what's the best way to proceed in these situations, as we cannot include sale_procurement_group_by_line in the "excludes" list in manifest. Thank you.

pedrobaeza:
Well, I know by experience that such future may not come. On other modules, I did some checks against `config["test_enable"]` and detect if there's an injected context to only activate the behavior in other cases, but I know others like Roussel doesn't like this. What I totally dislike is to have more splits in the CI (we already have 2 splits) to test them apart.

Daniel-Melian (commented):
Functionally reviewed LGTM.

rousseldenis:
> but I know others like Roussel doesn't like this. What I totally dislike is to have more splits in the CI (we already have 2 splits) to test them apart.

And the solution that introduce a configuration parameter to activate the feature avoid both :-)

manuelregidor:
> > but I know others like Roussel doesn't like this. What I totally dislike is to have more splits in the CI (we already have 2 splits) to test them apart.
> 
> And the solution that introduce a configuration parameter to activate the feature avoid both :-)

I don't know exactly what you mean by "introduce a configuration parameter". There is already a configuration field in res.company to activate the features introduced by this module. What kind of additional parameter do we need here? Thank you.

HaraldPanten:
> > > but I know others like Roussel doesn't like this. What I totally dislike is to have more splits in the CI (we already have 2 splits) to test them apart.
> > 
> > 
> > And the solution that introduce a configuration parameter to activate the feature avoid both :-)
> 
> I don't know exactly what you mean by "introduce a configuration parameter". There is already a configuration field in res.company to activate the features introduced by this module. What kind of additional parameter do we need here? Thank you.

@rousseldenis Could you give us more details, please?

HaraldPanten:
@rousseldenis  Could you give some feedback to us? We're not sure about how to continue with that.

THX!

ValentinVinagre:
@rousseldenis Could you give us some advice? We are waiting to finalize this PR based on your comments.
Thanks 😄

manuelregidor:
@rousseldenis I've removed the incompatibility with sale_procurement_group_by_line from the manifest and explained in roadmap why both modules should be used together. I don't think there's need to create a system parameter to activate the features provided by this module as there is already a field in company to activate them.

sale_order_line_multi_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,69 @@
...
+        related="company_id.allow_sale_multi_warehouse", store=True
+    )
+    suitable_warehouse_ids = fields.Many2many(
+        string="Suitable Warehouses",
+        comodel_name="stock.warehouse",
+        compute="_compute_suitable_warehouse_ids",
+    )
+
+    @api.depends("company_id")
+    def _compute_allow_sale_multi_warehouse(self):
```
rousseldenis: You can remove this as field is related

sale_order_line_multi_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,69 @@
...
+            lambda a: a.warehouse_id.id not in compatible_warehouse_ids.ids
+        )
+
+    def _action_confirm(self):
+        for order in self:
+            if order.get_incompatible_multi_warehouse_lines(self.warehouse_id):
+                raise ValidationError(
+                    _(
+                        "Some sale order line assignment lines are not compatible with "
+                        "the selected warehouse. Quotation {order} could not be validated."
```
rousseldenis: You should use translation arguments instead `format()`

sale_order_line_multi_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,321 @@
...
+                reduce_line = self.sale_order_line_warehouse_ids.filtered(
+                    lambda a: float_compare(
+                        a.product_uom_qty,
+                        0.0,
+                        precision_rounding=a.product_uom_id.rounding,
+                    )
+                    > 0
+                )
+            else:
+                raise ValidationError(_("Amount cannot be reduced."))
```
rousseldenis: IMHO, this message could be enhanced.

rousseldenis, line 162: IMHO, this message could be enhanced.

rousseldenis (commented):
Some comments.

Testing the wizard could be great

manuelregidor:
@rousseldenis Changes applied. New test has been added too.

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def change_warehouse(self):
+        self.ensure_one()
+        if self.sale_order_id.state not in ["sale", "done"] and self.new_warehouse_id:
+            for line in self.so_multi_warehouse_change_line_ids:
+                so_line_warehouse = line.sale_order_line_warehouse_id
+                related_lines = (
+                    so_line_warehouse.order_line_id.sale_order_line_warehouse_ids
+                )
+                same_warehouse_line = related_lines.filtered(
```
rousseldenis: Adding this will allow to avoid `same_warehouse_line[0].write(` but simply `same_warehouse_line.write(`

```suggestion
                same_warehouse_line = fields.first(related_lines.filtered(
```

manuelregidor: done

sale_order_line_multi_warehouse/wizard/so_multi_warehouse_change_wizard.py
```diff
@@ -0,0 +1,103 @@
...
+                    line.sale_order_line_warehouse_id.unlink()
+                else:
+                    line.sale_order_line_warehouse_id.write(
+                        {"warehouse_id": self.new_warehouse_id.id}
+                    )
+            self.sale_order_id.write({"warehouse_id": self.new_warehouse_id.id})
+        else:
+            raise ValidationError(
+                _(
+                    "Warehouse cannot be changed as sale order is in {state} state."
```
rousseldenis: Oops, forgot this one

manuelregidor: Is it necessary to change it method name_get in sale_order_line_warehouse.py as well? Thank you.

rousseldenis: No, it's only needed when there are translations.

manuelregidor: Changed

manuelregidor:
@rousseldenis I applied the changes you suggested and added a new test. Could have a look, please? Thank you.

ValentinVinagre:
@rousseldenis There is currently a very similar case to this module in https://github.com/OCA/sale-workflow/pull/3598/files. The solution has been to change the OCA CI in the repository for the incompatible modules. Don't you think that would be the solution to apply? I mean that it avoids creating parameters in odoo, etc. only by not adapting the CI.
If you agree, we will apply it.

manuelregidor:
@rousseldenis @Daniel-Melian Tests passed. Could you have a look? Thank you

HaraldPanten:
@rousseldenis you asked for changes some weeks ago. Could you have a look?

THX!

rousseldenis (approved):

PR #3362, opened 2024-10-15
[16.0] [FIX] fix invoiced_amount when invoice currency == sale curren…

BenjaHe:
We have a bug when:
invoice.currency_id == rec.currency_id  and invoice.currency_id != invoice.company_currency_id
-> This may append in an inter-company scénario

In that case the invoiced amount in the sale order is not correct:
invoice
Saleorder

### DIFF
--- a/sale_order_invoice_amount/models/sale_order.py
+++ b/sale_order_invoice_amount/models/sale_order.py
@@ -42,6 +42,13 @@ class SaleOrder(models.Model):
                                 invoice.company_id,
                                 invoice.invoice_date or fields.Date.today(),
                             )
+                        elif (
+                            invoice.currency_id == rec.currency_id
+                            and invoice.currency_id != invoice.company_currency_id
+                        ):
+                            rec.invoiced_amount += (
+                                invoice.amount_total_in_currency_signed
+                            )
                         else:
                             rec.invoiced_amount += invoice.amount_total_signed
                 # Uninvoiced amount could not be equal to total - invoiced amount.


BenjaHe:
@luc-demeyer

sale_order_invoice_amount/models/sale_order.py
```diff
@@ -42,6 +42,13 @@ def _compute_invoice_amount(self):
                                 invoice.company_id,
                                 invoice.invoice_date or fields.Date.today(),
                             )
+                        elif (
```
flotho, line 45: Hy,

Thanks for the proposal.
I totally agree on the case. 
Yet I think it should be great to simplify. Actually this line L37 is the same as your addition.
So in the end we have 2 main cases and not 3 : 
* One case when `rec.currency_id == invoice.company_currency_id` fair case in L53
* the other one when `rec.currency_id != invoice.company_currency_id` andIMHO, we should always use the convert method to prevent the enumeration of cases and improve readability.

Of course this is my humble opinion

flotho (changes requested):

rousseldenis:
@BenjaHe Thanks for this. A great thing should be tests addition for that use case. Thanks

alexey-pelykh (commented):
Thanks for addressing this, the inter-company scenario where `invoice.currency_id == rec.currency_id != invoice.company_currency_id` is a real gap — using `amount_total_signed` there gives the company-currency amount instead of the document-currency amount, so the fix is correct in choosing `amount_total_in_currency_signed`.

That said, I agree with @flotho's suggestion. The three-branch structure can be collapsed to two cases:

- `rec.currency_id == invoice.company_currency_id` → use `amount_total_signed` (already in the right currency)
- otherwise → convert from `invoice.currency_id` to `rec.currency_id` via `_convert`

When invoice and SO currencies happen to match, `_convert` with rate 1.0 is a no-op, so the conversion path covers both the "all different" and "invoice == SO but != company" cases without a dedicated branch. This avoids enumerating currency combinations and is easier to reason about going forward.

Also, as @rousseldenis mentioned, a test covering this specific scenario (SO and invoice in the same non-company currency) would be needed to prevent regressions.

rousseldenis:

PR #3364, opened 2024-10-16
[16.0][FIX] sale_order_line_date: fix setting empty commitment date on stock moves

### LINKED ISSUE #3363: [16.0] sale_order_line_date : removing the date after sale has been confirmed raises TypeError
## Module
sale_order_line_date

## Describe the bug
When updating the Delivery date (commitment_date field) on a sale order line, this module updates the stock moves related to the line.
During the computation of new dates for stock moves, the delta with previous date is used in module stock to avoid recomputation of all delays.
Now, if we just remove the commitment date, the delta is trying to compare previous date with nothing... which leads to :
```
TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType'
```

## To Reproduce
**Affected versions**: 16.0

1. Open a brand new runboat for 16.0
2. Create a sale order with some stockable product (let's say FURN_6666)
3. Set the delivery date on the line
4. Confirm the sale order
5. Remove the delivery date
6. Save the order

**Expected behavior**
This shouldn't raise a TypeError.
All dates should be updated in stock moves and ideally be the same as when the order is first confirmed with no delivery date set.

**Additional context**
By default, when no delivery date is set, procurements are created with date_deadline :
https://github.com/odoo/odoo/blob/c991ad65821e2022868b9bb8b1e899795ceba0a7/addons/sale_stock/models/sale_order_line.py#L259

metaminux:
…sale order line

Closes #3363

### DIFF
--- a/sale_order_line_date/models/sale_order_line.py
+++ b/sale_order_line_date/models/sale_order_line.py
@@ -31,13 +31,17 @@ class SaleOrderLine(models.Model):
 
     def write(self, vals):
         res = super().write(vals)
-        moves_to_upd = set()
         if "commitment_date" in vals:
-            for move in self.move_ids:
-                if move.state not in ["cancel", "done"]:
-                    moves_to_upd.add(move.id)
-        if moves_to_upd:
-            self.env["stock.move"].browse(moves_to_upd).write(
-                {"date_deadline": vals.get("commitment_date")}
-            )
+            if vals.get("commitment_date"):
+                self.move_ids.filtered(
+                    lambda sm: sm.state not in ["cancel", "done"]
+                ).write({"date_deadline": vals.get("commitment_date")})
+            else:
+                for line in self:
+                    date_deadline = (
+                        line.order_id.commitment_date or line._expected_date()
+                    )
+                    line.move_ids.filtered(
+                        lambda sm: sm.state not in ["cancel", "done"]
+                    ).write({"date_deadline": date_deadline})
         return res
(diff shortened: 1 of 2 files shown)


metaminux:
First commit is a test confirming issue #3363 
Second one is my attempt to fix it.

@ferran-S73 @rousseldenis @Kev-Roche Ready for your reviews...

Regards,

sale_order_line_date/models/sale_order_line.py
```diff
@@ -31,13 +31,17 @@ def _prepare_procurement_values(self, group_id=False):
...
+                    lambda sm: sm.state not in ["cancel", "done"]
+                ).write({"date_deadline": vals.get("commitment_date")})
+            else:
+                for line in self:
+                    date_deadline = (
+                        line.order_id.commitment_date or line._expected_date()
+                    )
+                    line.move_ids.filtered(
+                        lambda sm: sm.state not in ["cancel", "done"]
+                    ).write({"date_deadline": date_deadline})
```
rousseldenis, line 46: @metaminux Here, try to build ids set as before to write just once

metaminux: I think we need to loop on lines because _expected_date calls ensure_one at first and we could call the write method on lines that don't belong to the same order, so date_deadline wont be the same for each line...

StefanRijnhart (approved):
Thanks! I just closed duplicate https://github.com/OCA/sale-workflow/pull/3540 of this one.

sale_order_line_date/models/sale_order_line.py
rousseldenis: Ok, nitpicking even if a group by date write() could be great

rousseldenis (approved):

StefanRijnhart:
Porting to 18.0 in https://github.com/OCA/sale-workflow/pull/3565

PR #3367, opened 2024-10-21
[17.0][MIG] sale_stock_picking_validation_blocking: Migrate to 17.0

kevinkhao:


rousseldenis:
@kevinkhao Could you check tests ?

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -31,11 +31,9 @@ def setUp(self):
     def test_sale_stock_picking_validation_blocked(self):
         self.sale_order.action_confirm()
         picking = self.sale_order.picking_ids
-        picking.move_lines.write({"quantity_done": 1})
+        picking.move_ids.write({"quantity": 1})
         self.sale_order.action_block_picking_validation()
-        self.assertFalse(picking.show_validate)
```
StefanRijnhart: Do you want to check the value of `hide_button_picking_validation_blocked` instead, here and below?

sale_stock_picking_validation_blocking/views/sale_order.xml
```diff
@@ -2,32 +2,35 @@
...
+                <field name="hide_button_picking_validation_blocked" invisible="True" />
+            </field>
             <button name="action_draft" position="after">
                 <button
                     name="action_block_picking_validation"
                     type="object"
                     string="Block Delivery Validation"
                     groups="sales_team.group_sale_manager"
-                    attrs="{'invisible': ['|', '|', ('state', 'not in', ('sale', 'done')), ('delivery_count', '=', 0), ('picking_validation_blocked', '=', True)]}"
+                    invisible="picking_validation_blocked or hide_button_picking_validation_blocked"
```
StefanRijnhart, line 23: For better readability here, maybe use `picking_validation_blocked` in the compute method of `hide_button_picking_validation_blocked`?

sale_stock_picking_validation_blocking/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "summary": """
         This module adds the opportunity to prevent
         the validation of delivery order from the SO.""",
-    "version": "13.0.1.0.1",
+    "version": "17.0.1.0.1",
```
StefanRijnhart: 17.0.1.0.0

sale_stock_picking_validation_blocking/views/sale_order.xml
kevinkhao: It's what I went for at the start, but there is also:
- the other button that does the opposite, and
- conditions where you want to hide both block and unblock 
- you have to reason with invisible=... instead of display=...
in that case it gets confusing since both are using the same field and you can't have a simple logic to cover different sets of conditions. I think this approach is just simple to reason with. Does this seem reasonable to you ?

StefanRijnhart: Yes, alright!

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleStock(TransactionCase):
+    def setUp(self):
```
StefanRijnhart: Oh, please refactor this using setUpclass

sale_stock_picking_validation_blocking/views/stock_picking.xml
```diff
@@ -0,0 +1,28 @@
...
+        >stock.picking.form (in sale_stock_picking_validation_blocking)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//sheet" position="before">
+                <field name="validation_blocked_by_so" invisible="1" />
+                <div
+                    class="alert alert-danger text-center o_form_header"
+                    role="alert"
+                    invisible="validation_blocked_by_so == False"
```
StefanRijnhart:
```suggestion
                    invisible="not validation_blocked_by_so"
```

kevinkhao:
@StefanRijnhart done thanks for feedback

sale_stock_picking_validation_blocking/tests/test_sale_stock.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleStock(TransactionCase):
+    def setUpClass(cls):
```
StefanRijnhart, line 10: `TypeError: TestSaleStock.setUpClass() missing 1 required positional argument: 'cls'`
setUpClass requires `@classmethod` decorator.

kevinkhao: my bad, corrected

StefanRijnhart (approved):
Thanks for the updates!

astirpe (approved):

StefanRijnhart:
@kevinkhao Can you rebase to resolve the conflict due to the automated linting of unrelated module sale_order_general_discount?

alexey-pelykh (commented):
Thanks for the migration, looks good overall. Just one minor thing I noticed:

In `stock_picking.py`, the translation string uses %-formatting inside `_()`:

```python
_("Validation is blocked by SO for picking %s" % picking.name)
```

The recommended pattern since Odoo 15+ is to pass the arguments to `_()` directly so the string stays lazy-translatable:

```python
_("Validation is blocked by SO for picking %s", picking.name)
```

Also the `.pot` file still references `Project-Id-Version: Odoo Server 13.0` -- might be worth regenerating it after rebase.

Aside from that, the `attrs` to inline `invisible` conversion and the `move_lines`/`quantity_done` renames are done correctly. The computed `hide_button_picking_validation_blocked` field is a clean approach for the button visibility logic.

Waiting on the rebase Stefan requested to resolve the `sale_order_general_discount` conflict before this can go in.

rousseldenis (approved):

BhaveshHeliconia (approved):
LGTM!

BhaveshHeliconia:
@kevinkhao Could you please rebase the PR?

PR #3371, opened 2024-10-22
[17.0][ADD]: add new module add_sale_order_qty_and_uom_change_no_reco…

victorvermot:
This new module is based on sale_order_qty_change_no_recompute. For some people, we also want the unit of measure to not recompute the price unit

# Before Merge
Changing the unit of measure and qty on sale order lines will reset the price unit to its base value

# Ater Merge
Changing the unit of measure and qty on sale order lines will not reset the price unit to its base value


alexey-pelykh (changes requested):
Thanks for the contribution, but I have significant concerns about this PR's approach.

**This should be a patch to the existing module, not a new module.**

Comparing `monkeypatching.py` here with the existing `sale_order_qty_change_no_recompute`, the only meaningful difference is two added lines:

```python
if "parent_line_id" in depends:
    depends.remove("parent_line_id")
```

The existing module already removes `product_uom` from the depends, so the UOM part of the claim is already covered there. This new module is essentially a full copy of the existing one with a two-line addition. Shipping both modules would also be problematic -- they both monkeypatch `Field.get_depends` at module-import time, so installing both would cause the second import to silently overwrite the first, leading to unpredictable behavior depending on module load order.

The right approach would be to open a PR against `sale_order_qty_change_no_recompute` adding the `parent_line_id` removal.

Additional issues if this were to go forward as a separate module:

1. The `__manifest__.py` name and summary are identical to the base module -- they should differentiate if this is meant to be a standalone alternative.
2. The `.pot` filename is `sale_order_qty_change_no_recompute.pot` (base module name) instead of `sale_order_qty_and_uom_change_no_recompute.pot`.
3. The `i18n/fr.po` is an empty file.
4. The tests are copy-pasted from the base module with no additional coverage for the UOM change behavior or the `parent_line_id` change that is supposedly the point of this module.
5. No `depends` on the base module, which confirms these two modules would conflict if co-installed.

rousseldenis:

PR #3373, opened 2024-10-23
[17.0][ADD] sale_partner_utm_source

natuan9:
This module provides a way to classify partners using Odoo’s native utm.source model. Each partner can be assigned a Source, which is added just below the partner's name on the form view. When creating a Quotation, the Source field of the partner is automatically copied to the Quotation. Users can manually adjust the Source in the Quotation if needed.

**Depend on:** 
- [[17.0][ADD] partner_utm_source](https://github.com/OCA/partner-contact/pull/1886)


rousseldenis:
@natuan9 Thanks for this. Could you maybe change the module name to `sale_partner_utm_source` in order to reflect that this is not for 'any`other source. Thanks

rousseldenis:
I advise too to install `pre-commit` tool locally before commiting.

sale_partner_source/models/res_partner.py
```diff
@@ -0,0 +1,9 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    source_id = fields.Many2one('utm.source', string='Source')
```
rousseldenis: Why not putting this outside of a 'sale` module to keep the utm source of the partner without having to install sale module ?

natuan9: Thanks for your review.
This field is used for creating quotations. When a quotation is created, it automatically copies the `source_id` from partner if it's empty

rousseldenis: Yes, I understood that part. What I want to say is a module that stores the utm source on partner level could be great without the sale feature

natuan9: Oh, I understand your point now. I'll make adjustments

rousseldenis (commented):
@natuan9 Maybe a little test could be great

rousseldenis:
@natuan9 https://github.com/OCA/partner-contact/pull/1886 has been merged

hailangvn:
The new PR https://github.com/OCA/sale-workflow/pull/3732 takes over this one. It includes small test per requested.

PR #3375, opened 2024-10-23
[17.0][MIG] sale_automatic_workflow_periodicity

henrybackman:
Migration from 14.0 to 17.0


sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -176,9 +176,12 @@ def _register_payment_invoice(self, invoice):
     def _handle_pickings(self, sale_workflow):
         pass
 
+    def _sale_workflow_domain(self, workflow):
```
rousseldenis, line 179: @henrybackman Please put these changes in `sale_automatic__workflow`in another PR (moreover those changes should not be in same commit as migration of the other module)

henrybackman: Thank you! I've now separated the commits

alexey-pelykh (changes requested):
Thanks for picking up this migration.

A few things to address:

**View XML -- duplicate `name` attribute**

The inherited view inserts a `<div class="container" name="sale_order_confirm">` before the existing `<div ... name="sale_order_confirm">`. Having two elements with the same `name` in the form will break xpath resolution for any other module targeting `//div[@name='sale_order_confirm']`. The original 14.0 version used `<group name="workflow_periodicity_exec">` with its own unique name -- please use a distinct name here as well (e.g. `name="workflow_periodicity_exec"`).

Also, the 14.0 view used `<group>` elements which provide automatic label alignment. The 17.0 version switches to manual `<div class="row">` / `<label>` / `<span>` / `<field>` layout to match the parent form style, which is fine, but the `next_execution` field lost its `readonly` attribute. Since this field is computed by the cron and shouldn't be edited manually, consider adding `readonly="1"` on the field in the view (or on the model definition as in 14.0).

**Translation files reference Odoo 14.0**

Both `i18n/sale_automatic_workflow_periodicity.pot` and `i18n/es.po` have `Project-Id-Version: Odoo Server 14.0`. These should be regenerated for 17.0 (the .pot at minimum -- `oca-gen-addon-readme` / the export tool will take care of it).

**`sale_automatic_workflow` changes scope**

As @rousseldenis already noted, the hook-point additions to `sale_automatic_workflow` (the new `_sale_workflow_domain` and `_workflow_process_to_run_domain` methods) should be submitted as a separate PR against `sale_automatic_workflow`. This PR should only contain the new module. That prerequisite PR needs to land first, and this migration PR should be rebased on top of it.

rousseldenis:

PR #3376, opened 2024-10-24
[IMP] sale_invoice_policy: Several improvements

rousseldenis:
- [x] Don't use the required parameter to compute the quantity to invoice
- [x] Simplify `depends()`
- [x] Don't use the required parameter to compute the untaxed amount to invoice
- [x] Don't rewrite the Odoo core compute method
- [x] Improve README

Refactor:

I think the default policy that was applied on sale orders from the default product configuration (defined in sale module) was the wrong approach as it is not contextually correct (products are not sale orders).

That said, I introduced an invoice policy on sale order level more explicitly than before (void) which is 'Products Invoicing Policy'. That let the user free to choose between standard behavior or sale order level configuration.

The question is:

- As the required attribute was used before means that the sale order invoicing policy should not use the product configurations. Should we reintroduce that behavior (to constrain sale orders to NOT use product configuration at all ?) ? @ioans73 @fuentes73 @sergio-teruel @ACheung-FactorLibre @Cedric-Pigeon @acsonefho

### DIFF
--- a/sale_invoice_policy/__manifest__.py
+++ b/sale_invoice_policy/__manifest__.py
@@ -10,9 +10,11 @@
     "category": "Sales Management",
     "version": "16.0.2.0.0",
     "license": "AGPL-3",
-    "depends": ["sale_stock"],
+    "depends": ["sale_stock", "base_partition"],
+    "external_dependencies": {"python": ["openupgradelib"]},
     "data": [
         "views/res_config_settings_view.xml",
         "views/sale_view.xml",
     ],
+    "pre_init_hook": "pre_init_hook",
 }
--- a/sale_invoice_policy/models/sale_order.py
+++ b/sale_invoice_policy/models/sale_order.py
@@ -9,41 +9,27 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     invoice_policy = fields.Selection(
-        [("order", "Ordered quantities"), ("delivery", "Delivered quantities")],
-        readonly=True,
+        [
+            ("product", "Products Invoice Policy"),
+            ("order", "Ordered quantities"),
+            ("delivery", "Delivered quantities"),
+        ],
+        compute="_compute_invoice_policy",
+        store=True,
+        readonly=False,
+        required=True,
         states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
+        precompute=True,
         help="Ordered Quantity: Invoice based on the quantity the customer "
         "ordered.\n"
         "Delivered Quantity: Invoiced based on the quantity the vendor "
-        "delivered (time or deliveries).",
-    )
-    invoice_policy_required = fields.Boolean(
-        compute="_compute_invoice_policy_required",
-        default=lambda self: self.env["ir.default"].get(
-            "res.config.settings", "sale_invoice_policy_required"
-        ),
+        "delivered (time or deliveries). This applies for storable products only.",
     )
 
-    @api.model
-    def default_get(self, fields_list):
-        res = super().default_get(fields_list)
-        default_invoice_policy = (
-            self.env["res.config.settings"]
-            .sudo()
-            .default_get(["default_invoice_policy"])
-            .get("default_invoice_policy", False)
-        )
-        if "invoice_policy" not in res:
-            res.update({"invoice_policy": default_invoice_policy})
-        return res
-
-    @api.depends("partner_id")
-    def _compute_invoice_policy_required(self):
-        invoice_policy_required = (
-            self.env["res.config.settings"]
-            .sudo()
-            .default_get(["sale_invoice_policy_required"])
-            .get("sale_invoice_policy_required", False)
-        )
-        for sale in self:
-            sale.invoice_policy_required = invoice_policy_required
+    @api.depends("company_id")
+    def _compute_invoice_policy(self) -> None:
+        """
+        Get default sale order invoice policy
+        """
+        for company, sale_orders in self.partition("company_id").items():
+            sale_orders.invoice_policy = company.sale_default_invoice_policy
--- a/sale_invoice_policy/models/sale_order_line.py
+++ b/sale_invoice_policy/models/sale_order_line.py
@@ -15,10 +15,11 @@ class SaleOrderLine(models.Model):
         "order_id.invoice_policy",
     )
     def _compute_qty_to_invoice(self):
+        """
+        Exclude lines that have their order invoice policy filled in
+        """
         other_lines = self.filtered(
-            lambda l: l.product_id.type == "service"
-            or not l.order_id.invoice_policy
-            or not l.order_id.invoice_policy_required
+            lambda l: l.product_id.type == "service" or not l.order_id.invoice_policy
         )
         super(SaleOrderLine, other_lines)._compute_qty_to_invoice()
         for line in self - other_lines:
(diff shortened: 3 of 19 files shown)


rousseldenis:
Don't hesitate to test it on runboat

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -29,74 +25,27 @@ def _compute_qty_to_invoice(self):
...
-        return True
+        saved_product_values = dict()
+        lines = self - other_lines
+        for line in lines:
+            # Save product invoice policy, change it to sale order one, add context manager
+            # to not trigger other computes. Then, restablish former one.
+            with self.env.norecompute():
+                saved_product_values[line.product_id] = line.product_id.invoice_policy
+                line.product_id.invoice_policy = line.order_id.invoice_policy
+            # We should call the super() per line as each sale order can have different policy
```
lmignon: To avoid 'recomputes' and marking the field as modified (i.e. writing it to DB), you need to change the approach and use the ‘protecting’ method on the env. You could also improve readability and the batch processing by grouping your lines by invoice policy and using a context manager to apply the invoice policy to the products when computing the untaxed amount to invoice.

```python

@contextmanager
def _sale_invoice_policy(self, lines):
    """ Apply the sale invoice policy to the products
   
    This method must be called with lines sharing the same invoice policy
    """
    invoice_policy = lines.mapped("order_id.invoice_policy")
    if len(invoice_policy) > 1:
        raise Exception("The mehtod _sale_invoice_policy must be called with lines sharing the same invoice policy")
    invoice_policy = invoice_policy[0]
    invoice_policy_field = self.env["product.product"]._fields["invoice_policy"]
    products = lines.product_id
    with self.env.protecting([invoice_policy_field], products):
        old_values = {}
        for product in products:
            old_values[product] = product.invoice_policy
            product_id.invoice_policy = invoice_policy
        yield
        for product, invoice_policy in old_values.items():
            product.invoice_policy = invoice_policy

@api.depends("order_id.invoice_policy")
def _compute_untaxed_amount_to_invoice(self) -> None:
    other_lines = self.filtered(
        lambda line: line.product_id.type == "service"
        or line.order_id.invoice_policy == "product"
        or line.order_id.invoice_policy == line.product_id.invoice_policy
        or line.state not in ["sale", "done"]
    )
    super(SaleOrderLine, other_lines)._compute_untaxed_amount_to_invoice()
    for lines in (self - other_lines).partition("invoice_policy").values():
        with self._sale_invoice_policy(lines):
            super(SaleOrderLine, lines)._compute_untaxed_amount_to_invoice()
```

lmignon (changes requested):

sale_invoice_policy/models/sale_order_line.py
rousseldenis: Done

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -29,74 +50,18 @@ def _compute_qty_to_invoice(self):
...
-                            )
-                            * inv_line.quantity
-                        )
-                amount_to_invoice = max(price_subtotal - amount, 0)
-            else:
-                amount_to_invoice = price_subtotal - line.untaxed_amount_invoiced
-            line.untaxed_amount_to_invoice = amount_to_invoice
-        return True
+        for lines in (
+            (self - other_lines).partition("product_id.invoice_policy").values()
```
lmignon:
```suggestion
            (self - other_lines).partition("order_id.invoice_policy").values()
```

lmignon (changes requested):

ferran-S73 (approved):
Code review OK. As per your question I see no problem in allowing product configuration for sale orders' invoice policy. It adds more flexibility to the module

phschmidt (approved):
Functionary review

marielejeune (approved):
LGTM (code + functional review)
Thanks for the work

anajuaristi:
> Don't hesitate to test it on runboat

Hello @rousseldenis 
I was trying to test in runboat but I can´t find or see the build for this PR. Could you help me with that please? 

Thank you!!!

sale_invoice_policy/models/sale_order.py
```diff
@@ -9,41 +9,27 @@ class SaleOrder(models.Model):
...
+        compute="_compute_invoice_policy",
+        store=True,
+        readonly=False,
+        required=True,
         states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
+        precompute=True,
         help="Ordered Quantity: Invoice based on the quantity the customer "
         "ordered.\n"
         "Delivered Quantity: Invoiced based on the quantity the vendor "
         "delivered (time or deliveries).",
```
jbaudoux: Add that is applies only for stock products

jbaudoux (commented):
Can you add in the readme that the SO invoice policy affects only stock products. Service products continue to be invoiced based on the product invoice policy.

sale_invoice_policy/__manifest__.py
```diff
@@ -10,9 +10,10 @@
     "category": "Sales Management",
     "version": "16.0.2.0.0",
     "license": "AGPL-3",
-    "depends": ["sale_stock"],
+    "depends": ["sale_stock", "base_partition"],
     "data": [
         "views/res_config_settings_view.xml",
         "views/sale_view.xml",
     ],
+    "pre_init_hook": "pre_init_hook",
```
jbaudoux, line 19:
```suggestion
    "pre_init_hook": "pre_init_hook",
    "external_dependencies": {"python": ["openupgradelib"]},
```

jbaudoux (commented):
And generate `requirement.txt` as it is not part of it in this repo

```
2025-01-25 16:00:20,521 417 INFO odoo odoo.modules.loading: Loading module sale_invoice_policy (66/67) 
2025-01-25 16:00:20,522 417 CRITICAL odoo odoo.modules.module: Couldn't load module sale_invoice_policy 
2025-01-25 16:00:20,522 417 CRITICAL odoo odoo.modules.module: No module named 'openupgradelib' 
```

marielejeune:
@rousseldenis Could you have a look at the last reviews please?

rousseldenis:
@jbaudoux Could you update your review ?

hildickethan (approved):
Looks good!

jbaudoux (approved):
Can you drop auto-generated files from the commits?

alexey-pelykh (commented):
Thanks for the refactor Denis, the move to a proper three-valued selection with `"product"` as a default is a much cleaner design than the old `invoice_policy_required` boolean approach. The delegation to Odoo core in `_compute_untaxed_amount_to_invoice` instead of reimplementing the whole calculation is a welcome simplification too.

A few things I noticed:

There is a copy-paste typo in `readme/CONTEXT.rst` (and consequently in the generated README.rst and index.html) -- the two bullet points both say "Invoicing on ordered quantities". The second one should be "Invoicing on delivered quantities".

In `_sale_invoice_policy`, a bare `Exception` is raised when called with mixed policies. Using `ValueError` (or `UserError` if the intent is to surface this to the UI) would be more conventional.

As @jbaudoux mentioned in his last review, the auto-generated files (`README.rst`, `static/description/index.html`) should be dropped from the commits. OCA CI regenerates them.

One note on the context manager pattern: the `env.protecting` + temporary product field patching approach is clever, but it does mean that if an exception occurs between setting the new values and the `yield`, the old values would not be restored. In practice this seems unlikely to cause issues since any such exception would typically roll back the transaction, but worth being aware of.

Otherwise the implementation looks solid -- `pre_init_hook` with `openupgradelib` for the migration, the company-level default, tests covering both policy paths and the context manager guard. Nice work.

PR #3381, opened 2024-10-29
[16.0][FIX] sale_tier_validation: Basic compatibility with sale_loyalty

pedrobaeza:
If you install `sale_loyalty`, this piece of code:

https://github.com/odoo/odoo/blob/2a7876538e9ea630563e39ee8402b27147d1e428/addons/sale_loyalty/models/sale_order.py#L740

is always executed when confirming a sales order, so it blocks it, no matter if no coupons are being handled.

This little code avoids this blocking without requiring a glue module, although further work would be needed if some coupons are applied.

@Tecnativa TT51484

### DIFF
--- a/sale_tier_validation/models/sale_order.py
+++ b/sale_tier_validation/models/sale_order.py
@@ -20,3 +20,18 @@ class SaleOrder(models.Model):
 
     def _get_rejected_notification_subtype(self):
         return "sale_tier_validation.sale_order_tier_validation_rejected"
+
+    def _get_fields_to_write_validation(self, vals, records_exception_function):
+        # Don't block order validation when sale_loyalty is installed and no coupon
+        # is being handled, due to this code assigning always an empty value:
+        # https://github.com/odoo/odoo/blob/2a7876538e9ea630563e39ee8402b27147d1e428/
+        # addons/sale_loyalty/models/sale_order.py#L740
+        # Done here for not creating a glue module as we can detect the situation. The
+        # only drawback is that this doesn't have any regression test
+        (
+            allowed_field_names,
+            not_allowed_field_names,
+        ) = super()._get_fields_to_write_validation(vals, records_exception_function)
+        if "applied_coupon_ids" in vals and not vals.get("applied_coupon_ids"):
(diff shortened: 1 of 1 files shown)


SirAionTech (changes requested):
> No, you didn't. You pressed on "Comment", not on "Request changes". You did code suggestions, but that's something different.
> 
> imagen

Sorry to pollute your PR, with this fake review, but I have to try this.
This is what I'm doing
image

SirAionTech:
And the result is the same as it was in https://github.com/OCA/purchase-workflow/pull/2216, what am I missing?

pedrobaeza:
That's very weird... Look how it looks my "Request changes":

https://github.com/OCA/helpdesk/pull/644#pullrequestreview-2460636774

imagen

Maybe it's because you don't have enough permissions. The key is that yours is very easy to be missed. In fact, in this PR is the same like you didn't oppose to it.

SirAionTech:
> That's very weird... Look how it looks my "Request changes":
> 
> [OCA/helpdesk#644 (review)](https://github.com/OCA/helpdesk/pull/644#pullrequestreview-2460636774)
> 
> imagen
> 
> Maybe it's because you don't have enough permissions. The key is that yours is very easy to be missed. In fact, in this PR is the same like you didn't oppose to it.

Yes, I think only reviews of maintainers are red and flashy and also appear in the summary along with CI checks.
When simple contributors do the same thing, it shows like mine in showing.

P.S.: For completeness, this is how my review is showing 
image

pedrobaeza:
Yeah, I see, then that's something I should pay attention from now on.

carlos-lopez-tecnativa (approved):
LGTM

alexey-pelykh (approved):
Thanks Pedro, nice catch. The approach is solid -- detecting the empty `applied_coupon_ids` write from `sale_loyalty` and allowing it through the validation gate is the right trade-off vs. introducing a glue module for this.

LGTM.

pedrobaeza:
Merging, as the requested changes was a test from @SirAionTech 

/ocabot merge patch

PR #3382, opened 2024-10-30
18.0 mig sale tier validation

valentinagiuliani-rapsodoo:
Needed also modulo base_tier_validation https://github.com/OCA/server-ux/pull/964


rousseldenis:
Already migrated

PR #3387, opened 2024-11-04
[15.0][ADD] sale_pricelist_global_rule: new module to apply pricelist based on cumulative quantities across sales order lines

carlos-lopez-tecnativa:
- by product template
- by product category

TT51247

@Tecnativa @pedrobaeza @victoralmau @chienandalu could you please review this.


sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,84 @@
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    need_recompute_pricelist_global = fields.Boolean()
```
pedrobaeza, line 8:
```suggestion
    need_recompute_pricelist_global = fields.Boolean(store=False, default=True)
```

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,84 @@
...
+    has_pricelist_global = fields.Boolean(compute="_compute_has_pricelist_global")
+
+    @api.depends("pricelist_id")
+    def _compute_has_pricelist_global(self):
+        for sale in self:
+            sale.has_pricelist_global = any(
```
pedrobaeza: This is very limited, as the rule may not be applicable to the current lines.

sale_pricelist_global_rule/models/product_pricelist.py
```diff
@@ -0,0 +1,320 @@
...
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    def _compute_price_rule_get_items(
+        self, products_qty_partner, date, uom_id, prod_tmpl_ids, prod_ids, categ_ids
+    ):
+        items = super()._compute_price_rule_get_items(
+            products_qty_partner, date, uom_id, prod_tmpl_ids, prod_ids, categ_ids
+        )
+        # ignore new global rules on Odoo standard
```
pedrobaeza, line 14: Why do you need this if you are cleaning the values on create/write?

sale_pricelist_global_rule/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_
+
+  * Carlos López
```
pedrobaeza, line 4: You should include me as the designer (and now reviewer).

sale_pricelist_global_rule/models/sale_order.py
carlos-lopez-tecnativa: This doesn't work with `store=False`. Odoo handles it similarly in the sale order when the pricelist is changed, by using a storable field. Please see https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/sale/models/sale_order.py#L288

sale_pricelist_global_rule/models/sale_order.py
carlos-lopez-tecnativa: Done

sale_pricelist_global_rule/models/product_pricelist.py
carlos-lopez-tecnativa: This is needed because Odoo does not evaluate the `applied_on` field; it only checks if `product_tmpl_id` or `categ_id` is NULL. In this case, both conditions are True, so the rules are considered. To discard these lines, this step is necessary.
https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/product/models/product_pricelist.py#L106-L108

sale_pricelist_global_rule/readme/CONTRIBUTORS.rst
carlos-lopez-tecnativa: done

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,94 @@
...
+            categ_ids = {}
+            prod_tmpl_ids = {}
+            for line in sale.order_line:
+                prod_tmpl_ids[line.product_id.product_tmpl_id.id] = True
+                categ = line.product_id.categ_id
+                while categ:
+                    categ_ids[categ.id] = True
+                    categ = categ.parent_id
+            categ_ids = list(categ_ids)
+            prod_tmpl_ids = list(prod_tmpl_ids)
```
pedrobaeza: This code is repeated with pricelist one. Extract it to a method.

carlos-lopez-tecnativa: done

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

victoralmau (changes requested):
Functional review OK.

PR #3392, opened 2024-11-07
[15.0][ADD] account_invoice_sale_attachment

sbejaoui:
This module allows attaching a document to the sale order and automatically including it when an invoice is sent by email.


samirGuesmi (approved):
LGTM

qgroulard (approved):
LGTM (code review)

account_invoice_sale_attachment/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    sale_document_attachment = fields.Binary(copy=False, attachment=True)
+    sale_document_filename = fields.Char(copy=False)
+
+    def _get_sale_document_attachments(self):
+        for rec in self:
+            attachment = self.env["ir.attachment"].search(
```
ivs-cetmix, line 17: You should fetch all attachments using a single query to improve the performance. Something like this:

```python

attachments = self.env["ir.attachment"].search(                [
                    ("res_id", "in", self.ids),
                    ("res_model", "=", self._name),
                    ("res_field", "=", "sale_document_attachment"),
                ]
            )
for attachment in attachments:
    order = self.browse(attachment["res_id"])
    if order and order.sale_document_filename:
        attachment.name = order.sale_document_filename
```

ivs-cetmix (changes requested):
Hi @sbejaoui, thank you for your contribution!

Please check the comments. Also please add a CONFIGURE.rst file with a single line "No configuration is required" to avoid possible misunderstanding.

I would also suggest the following improvements:

- Change the attachment field to m2m to allow to attach multiple files.
- Add screenshots to the documentation to improve the UX.
- Add the CONTEXT.rst file which would explain the idea behind this module. Eg 

 "Sometime you need to attach specific Sales Order documents to all customer invoices. For example payment instructions or a user manual."

rousseldenis:
> Please check the comments. Also please add a CONFIGURE.rst file with a single line "No configuration is required" to avoid possible misunderstanding.

@ivs-cetmix CONFIGURE.rst and CONTEXT.rst are optional, see: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

They are not required to be merged. Please update your review

ivs-cetmix:
> > Please check the comments. Also please add a CONFIGURE.rst file with a single line "No configuration is required" to avoid possible misunderstanding.
> 
> @ivs-cetmix CONFIGURE.rst and CONTEXT.rst are optional, see: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
> 
> They are not required to be merged. Please update your review

@rousseldenis  Yes, I agree. However let's add at least the CONTEXT because it improves the search experience and helps other people to locate the module.

@sbejaoui I have another idea after I checked the module once again that might improve the UX and also simplify the code as well. Please review and let me know what you think about it:

- User wants to ensure that some attachments are added automatically from the Sales Order to all invoices that are created and sent from this SO.
- User opens the Sales order and adds those files as attachments directly in chatter.
- User opens the "Other Info" tab of the sales order and selects attachments that he wants to be automatically added in the Invoice Attachments field. Similar to how it's done in the [mail_attach_exiasting_attachment](https://github.com/OCA/social/tree/16.0/mail_attach_existing_attachment) module.
image
- bonus: we might also consider integration with this [module](https://github.com/OCA/social/tree/16.0/mail_attach_existing_attachment_account). 


Pros:  
- just one field instead of two
- can attach multiple attachments
- uses the standard attachment mechanism

rousseldenis:
> @rousseldenis Yes, I agree. However let's add at least the CONTEXT because it improves the search experience and helps other people to locate the module

Just I wanted you to pay attention to blocking reviews on such things that can refrain contributions. Prefer normal comment.

That said, I agree 100% that the more content we put in modules, the best it is to help people finding stuffs.

ivs-cetmix:
> > @rousseldenis Yes, I agree. However let's add at least the CONTEXT because it improves the search experience and helps other people to locate the module
> 
> Just I wanted you to pay attention to blocking reviews on such things that can refrain contributions. Prefer normal comment.
> 
> That said, I agree 100% that the more content we put in modules, the best it is to help people finding stuffs.

Noted, thank you! 😉

PR #3393, opened 2024-11-08
[15.0][IMP] sale_elaboration: Based report view elaborations in stock moves instead of sale.order.lines

sergio-teruel:
cc @Tecnativa TT51619


sale_elaboration/models/sale_order.py
```diff
@@ -53,11 +52,6 @@ class SaleOrderLine(models.Model):
     elaboration_price_unit = fields.Float(
         "Elab. Price", compute="_compute_elaboration_price_unit", store=True
     )
-    is_prepared = fields.Boolean(
-        compute=lambda self: None,
-        search="_search_is_prepared",
-        help=("Dummy field to be able to find prepared lines"),
-    )
```
CarlosRoca13, line 60: Why is this beeing removed?

sergio-teruel:
@CarlosRoca13 With this new approach the records of elaboration wiew are stock moves so is not needed have a fiel in sale order line to get is a sol is prepared or no

CarlosRoca13:
Ok! Thanks for the clarification :smile:

carlosdauden (approved):

CarlosRoca13 (approved):

PR #3394, opened 2024-11-08
[18.0][MIG] sale_order_carrier_auto_assign: Migration to 18.0

lef-adhoc:


matiasperalta1 (approved):
LGTM

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+    def action_confirm(self):
+        for rec in self:
+            if not rec.company_id.carrier_auto_assign:
+                continue
+            rec._add_delivery_carrier_on_confirmation()
+        return super().action_confirm()
+
+    def _add_delivery_carrier_on_confirmation(self):
+        """Automatically add delivery.carrier on sale order confirmation"""
+        for order in self:
```
xaviedoanhduy, line 18: I see `action_confirm` has iterated over the records in sale.order and called `_add_delivery_carrier_on_confirmation` function for each individual record. So why do we need to iterate over it again, you can consider it with function that directly handles that record and call `ensure_one` function will make the code clearer

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+            delivery_wiz_action = order.action_open_delivery_wizard()
+            delivery_wiz_context = delivery_wiz_action.get("context", {})
+            if not delivery_wiz_context.get("default_carrier_id"):
+                continue
+            delivery_wiz = (
+                self.env[delivery_wiz_action.get("res_model")]
+                .with_context(**delivery_wiz_context)
+                .create({})
+            )
+            delivery_wiz._get_delivery_rate()
```
xaviedoanhduy, line 29: do we really need to call this function, I checked and found that it is mainly called in `onchange` functions and did not see it appear in any test in odoo codebase. I tried removing it and found that the module still works as expected and I guess that just calling `button_confirm` is enough, since in `button_confirm` there is a call to `set_delivery_line` function. I'm not sure if it's completely correct because I might miss something, do you think so?

sale_order_carrier_auto_assign/models/sale_order.py
xaviedoanhduy, line 17: I see `action_confirm` has iterated over the records in `sale.order` and called `_add_delivery_carrier_on_confirmation` function for each single record. So why do we need to iterate over it again, you can consider it with function that directly handles that record and call `ensure_one` function will make the code clearer

xaviedoanhduy (commented):
hi @lef-adhoc, thanks for migration this module.
I tested the functional and it works fine.
btw, could you squash administrative commits.

sale_order_carrier_auto_assign/models/sale_order.py
lef-adhoc: I deleted the line and I get an error in the tests and if I don't delete it I don't get an error. Can you check?

sale_order_carrier_auto_assign/models/sale_order.py
lef-adhoc: I did it another way, what do you think?

sale_order_carrier_auto_assign/models/sale_order.py
xaviedoanhduy: thanks for taking the time, maybe it was worth appearing on that line.

sale_order_carrier_auto_assign/models/sale_order.py
xaviedoanhduy: it looks fine to me. thanks

xaviedoanhduy (approved):
LGTM

jbaudoux (changes requested):
You are missing this commit https://github.com/OCA/sale-workflow/commit/bfa00bb62fcdee6e8252983bb2396fb08fe4dbbb

and this change https://github.com/OCA/delivery-carrier/commit/bdff1b38c7664f96f79c9ed8fe71ad4c4fbe0ff4

and integrate those commits that were postponed for next version https://github.com/OCA/sale-workflow/pull/3081/commits

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,10 +7,9 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     def action_confirm(self):
-        for rec in self:
-            if not rec.company_id.carrier_auto_assign:
-                continue
-            rec._add_delivery_carrier_on_confirmation()
+        self.filtered(
+            lambda so: so.company_id.carrier_auto_assign
+        )._add_delivery_carrier_on_confirmation()
```
jbaudoux: This change should not be part of the migration commit

lef-adhoc:
@jbaudoux This is the migration to 18 of the module that is already in 17 and has the same commits that are in 17 https://github.com/OCA/sale-workflow/commits/17.0/sale_order_carrier_auto_assign shouldn't you do the PRs first to 17 before including them in 18?

sale_order_carrier_auto_assign/models/sale_order.py
lef-adhoc: I have already changed it to another commit

sebalix: IMO it doesn't optimize previous code: now we are doing two loops, the `filtered` + the one in `_add_delivery_carrier_on_confirmation`.
With previous code we were looping only once on `self` (the loop in `_add_delivery_carrier_on_confirmation` can be ignored, it doesn't hurt with current recordset being of size 1).

So this commit can be removed.

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/3635 to commit history before migration commit.

simahawk:
@lef-adhoc are you going to finish this one?

jbaudoux:
Superseded by https://github.com/OCA/sale-workflow/pull/3709

PR #3349, opened 2024-10-09
[17.0][FW] [IMP]sale_delivery_state: Skip sale order lines for delivery state

henrybackman:
Port of #2845 from 14.0 to 17.0.

PR #2864, opened 2023-12-19
[IMP]sale_delivery_state: Skip sale order lines for delivery state

manuelregidor:
Based on https://github.com/OCA/sale-workflow/issues/2818

PR #3395, opened 2024-11-09
[18.0][MIG] sale_delivery_state

Virendrasinh1909:


dreispt (approved):

dreispt:
As a note,https://github.com/OCA/sale-workflow/pull/3349 will need forward porting to 18.0.

sale_delivery_state/views/sale_order_views.xml
```diff
@@ -0,0 +1,46 @@
...
+                <button
+                    name="action_unforce_delivery_state"
+                    type="object"
+                    string="Unforce delivery done"
+                    invisible="not force_delivery_state or state != 'sale' or not locked"
+                />
+            </field>
+            <group name="sale_shipping" position="inside">
+                <field name="delivery_status" invisible="state != 'sale'" />
+                <field name="force_delivery_state" invisible="1" />
```
thienvh332, line 24: FYI, Since `https://github.com/odoo/odoo/pull/137031` the invisible fields are add automatically (invisible and readonly) if they are used by python expressions in views (invisible, required, readonly, context, domain...).
```suggestion
```

sale_delivery_state/views/sale_order_views.xml
```diff
@@ -0,0 +1,46 @@
...
+    <record id="view_order_tree_inherit_delivery_status" model="ir.ui.view">
+        <field name="name">sale.order.tree</field>
+        <field name="inherit_id" ref="sale.view_order_tree" />
+        <field name="model">sale.order</field>
+        <field name="arch" type="xml">
+            <field name="invoice_status" position="before">
+                <field
+                    name="delivery_status"
+                    widget="badge"
+                    optional="hide"
```
thienvh332, line 38:
```suggestion

```

thienvh332 (approved):
LG overall,
In addition to `PR 3349` that the previous person suggested, you also need to pick  #3020.

Additionally, I think we should have a migration for the **`sale_stock_delivery_state`** module. To avoid problems when sale_stock is installed.

sale_delivery_state/views/sale_order_views.xml
sebalix: Why hiding? I thought it was a global recommendation in OCA to hide new fields in tree/list views to not clutter them. By default the std field is hidden too in `sale_stock`.

sebalix (approved):

sebalix:
Thank you!

/ocabot merge nobump

sebalix:
If `sale_stock` (adding too `delivery_status` field) is installed alongside `sale_delivery_state`, this generates these warning logs:
```
2025-03-05 14:37:15,717 275 WARNING odoo odoo.fields: sale.order.delivery_status: selection=[('pending', 'Not Delivered'), ('partial', 'Partially Delivered'), ('full', 'Fully Delivered')] overrides existing selection; use selection_add instead 
```
Not sure how to tackle this, maybe by setting the `selection` dynamically in `_register_hook` or `post_load`?

simahawk:
Notes from our discussion: the validation checks only if the options of the selection are not the same. We should then include the `started` status (which says that at least a picking, maybe an internal one, has been done).
We don't compute the started one here but we add a hook to be overridden in the stock glue module.

simahawk (changes requested):
To be fixed to address `started` state.

ValentinVinagre:
please add the following commits:
https://github.com/OCA/sale-workflow/pull/2864/commits/ad4c714617735d15b9c5391681c2c82f32fa3d4f
https://github.com/OCA/sale-workflow/pull/3697/commits/cf761f5b06f25ad502a2faf02ec0e05cd5b3431d

simahawk:
> please add the following commits: [ad4c714](https://github.com/OCA/sale-workflow/commit/ad4c714617735d15b9c5391681c2c82f32fa3d4f) [cf761f5](https://github.com/OCA/sale-workflow/commit/cf761f5b06f25ad502a2faf02ec0e05cd5b3431d)

@ValentinVinagre thanks for the pointer. I'll finish the base migration on #3712 and I'll open a new one to fwd port those commits.

PR #3396, opened 2024-11-11
[17.0][FIX] sale_order_line_sequence: fix SingletonError in rare cases

StefanRijnhart:


CasVissers-360ERP (approved):
Functional review.

sale_order_line_sequence/models/account_move_line.py
```diff
@@ -14,11 +14,20 @@ class AccountMoveLine(models.Model):
 
     @api.depends("move_id.invoice_line_ids")
     def _compute_related_so_sequence(self):
+        """Reflect the sale line sequence(s) on the invoice line"""
         for rec in self:
-            if len(rec.move_id.mapped("line_ids.sale_line_ids.order_id")) > 1:
-                rec.related_so_sequence = "{}/{}".format(
-                    rec.sale_line_ids.order_id.name,
-                    rec.sale_line_ids.visible_sequence,
+            if (
```
rousseldenis, line 19: @StefanRijnhart Thanks, I would say I won't do specific code for len(sale lines) > 1 and remove the condition and else part to use only your new code. What do you think?

StefanRijnhart: Note that the formatting is more elaborate in the complex case, whereas in the simple case it is simply the numeric sequence of the sale line. Isn't it neater to be able to use the bare sale line sequence if possible (this should be the common case)?

rousseldenis (approved):

PR #3405, opened 2024-11-11
[16.0] [IMP] sale_mrp_bom: ensure bom_id can only be selected if right routing (manufacturing) is settle

### LINKED ISSUE #3403: [16.0] sale_mrp_bom: currently U can choose a bom in sale order line for non properly configured as manufactured products which is misleading
I suggest enhancing the bom_id field by extending the domain to allow BOM selection only when the product is configured with the right routes. Specifically, the product should have either:

- Manufacture and MTO routes enabled, or
- Manufacture with both MTO + MTS route (if applicable).

This improvement would help ensure that defining a BOM on the Sales Order leads to a Manufacturing Order upon confirmation aligning with expected behavior. I noticed this feature isn’t present in other version either, so this adjustment could done across versions.

What do you think?

_Originally posted by @arielbarreiros96 in https://github.com/OCA/sale-workflow/issues/3380#issuecomment-2446793455_

rrebollo:
Fixes #3403.


rrebollo:
@arielbarreiros96 dear friend could you test this approach? I'm doing the bom_id field readonly if the routing is not properly settle.
I think right now I have covered the "Manufacture + MTO" case, "Manufacture + MTO + MTS" is not so clear to me. Could U show me some concrete example to reproduce?

rrebollo:
As a matter of fact, I believe this proposal aligns with the add-on's original intent, as stated in the usage section of the README: 'if the routing is manufacturing then...'

arielbarreiros96 (commented):
From a functional POV, for me, it seems quite better, now it doesn't allow you to select BoM on the sale.order.line unless the product has the manufacture and MTO routes. I can't recall a case in which I needed to set the BoM without using MTO.
I wasn't able to test the behavior when stock_mts_mto_rule is installed, which is slightly different from pure MTO

rrebollo:
thank U @arielbarreiros96 , do you mean [this one](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_mts_mto_rule)?

PR #3407, opened 2024-11-13
[17.0][FIX] sales_team_security: error uninstalling - use deprecated `api.Environment.manage`

dalonsod:
Starting from v15, `api.Environment.manage()` is deprecated and its usage should be simply be removed.

At least for v17 using it raises an error (for v15 a warning was raised instead), so removing it is actually needed.

Error raised uninstalling `sales_team_security`:

```
  File "/home/.../sale-workflow/sales_team_security/hooks.py", line 25, in uninstall_hook
    with api.Environment.manage():
AttributeError: type object 'Environment' has no attribute 'manage'

```

cc @pedrobaeza @ivantodorovich

### DIFF
--- a/sales_team_security/hooks.py
+++ b/sales_team_security/hooks.py
@@ -2,8 +2,6 @@
 # Copyright 2020 - Iván Todorovich
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import api
-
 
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #3412, opened 2024-11-15
[17.0][FIX] sale_blanket_order: sale.blanket.order note field should be Html

bobslee:
The `sale.blanket.order` field (type) should be Html, as the default (copied) `res.company` `invoice_terms` field is also Html.

### DIFF
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -90,7 +90,7 @@ class BlanketOrder(models.Model):
         string="Customer Reference",
         copy=False,
     )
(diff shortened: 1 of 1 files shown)


alexey-pelykh (approved):
Thanks for this fix. The `invoice_terms` field on `res.company` is indeed `Html` in Odoo 17, so using `fields.Text` here loses the formatting when copying the default. The field type change is safe since both `Text` and `Html` map to the same PostgreSQL `text` column type, so no migration script is needed -- existing plain-text data will render fine.

LGTM.

rousseldenis (approved):

BhaveshHeliconia (approved):

PR #3422, opened 2024-11-25
[15.0][IMP] sale_planner_calendar: Backport improvements from 16.0

carlosdauden:
@Tecnativa TT51739

cc @sergio-teruel @pilarvargas-tecnativa


pilarvargas-tecnativa:
There is an error in the recurrence test. With the previous routing event model, when events were reassigned, a new recurrence was created for the modified events. In v16 I remember I had to modify the wizard logic in the migration because of the changes in the new version of Odoo. It is possible that with the reactorisation of the model change the bug in this version will come to light.

pilarvargas-tecnativa (approved):

carlosdauden:
@sergio-teruel can you review?

sergio-teruel (approved):

PR #2461, opened 2023-03-31
[16.0][MIG] sales_team_security_crm

SoporteOperu:

PR #3429, opened 2024-11-25
[17.0][MIG] sales_team_security_crm: Migration to 17.0

lef-adhoc:
I migrated the module through this PR #2461


cvinh (approved):
LGTM

sales_team_security_crm/security/security.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="crm_lead_team_rule" model="ir.rule">
+        <field name="name">Team Leads/Opportunities</field>
+        <field name="model_id" ref="crm.model_crm_lead" />
+        <field
+            name="domain_force"
+        >['|', '|', ('user_id', '=', user.id), ('user_id', '=', False), '|', ('team_id', '=', user.sale_team_id.id),
```
carlos-lopez-tecnativa: I think this can be simplified to: What do you think?

```suggestion
        >['|', ('user_id', 'in', [user.id, False]), ('team_id', 'in', [user.sale_team_id.id, False]),
```

sales_team_security_crm/security/security.xml
```diff
@@ -0,0 +1,39 @@
...
+            name="groups"
+            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
+        />
+    </record>
+    <record id="crm_activity_report_team" model="ir.rule">
+        <field name="name">Team Activities Analysis</field>
+        <field name="model_id" ref="crm.model_crm_activity_report" />
+        <field
+            name="domain_force"
+        >['|', '|', ('user_id', '=', user.id), ('user_id', '=', False), '|', ('team_id', '=', user.sale_team_id.id),
```
carlos-lopez-tecnativa: the same here

carlos-lopez-tecnativa (changes requested):
LGTM just a minor change

TT52441

@Tecnativa

sales_team_security_crm/security/security.xml
cvinh: Moreover, you must take into account the multi team that has been introduced in odoo v15. In that case, user.sale_team_id can be false if a user is in several teams (ie https://github.com/odoo/odoo/blob/b7abee1f64d6812d66c5ac61d1d8f841310c81d7/addons/sales_team/models/res_users.py#L32)
IMHO the correct rule should be  
['|', '|', ('user_id', 'in', [user.id, False]), ('team_id', '=', user.sale_team_id.id), ('team_id', 'in', user.crm_team_ids.ids)]

sales_team_security_crm/security/security.xml
cvinh: same here
['|', '|', ('user_id', 'in', [user.id, False]), ('team_id', '=', user.sale_team_id.id), ('team_id', 'in', user.crm_team_ids.ids)]

sales_team_security_crm/security/security.xml
```diff
@@ -0,0 +1,39 @@
...
+        </field>
+        <field
+            name="groups"
+            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
+        />
+    </record>
+    <record id="sales_team_team_rule" model="ir.rule">
+        <field name="name">Own Sale Teams</field>
+        <field name="model_id" ref="sales_team.model_crm_team" />
+        <field name="domain_force">[('id', '=', user.sale_team_id.id)]</field>
```
cvinh: and here
['|', ('id', '=', user.sale_team_id.id), ('id', 'in', user.crm_team_ids.ids)]

cvinh (changes requested):

sales_team_security_crm/security/security.xml
cvinh: in our case, we did not want the users in Team Only group to access to leads without a user_id so we finally did
['|', '|', ('user_id', '=', user.id), ('team_id', '=', user.sale_team_id.id), ('team_id', 'in', user.crm_team_ids.ids)]

pedrobaeza: The original module should include leads without user.

cvinh: > The original module should include leads without user.

You are right considering that odoo standard has the same policy with crm ('my document only' group includes lead without user)
Anyway, the rule can be customized after on.
But natively, please include crm_team_ids because if not, the rule does not work out of the box in multi-team configuration

sales_team_security_crm/security/security.xml
lef-adhoc: Wouldn't [('id', 'in', user.crm_team_ids.ids)] be enough?

lef-adhoc:
@pedrobaeza @cvinh 
How do you see it?
Do you know why the tests failed?

lef-adhoc:
@cvinh the tests fail with the domains you gave me, could you check it?

cvinh:
> @cvinh the tests fail with the domains you gave me, could you check it?

Humm there should be something to do in sales_team_security

sales_team_security_crm/tests/test_sales_team_security_crm.py
```diff
@@ -0,0 +1,21 @@
...
+        super().setUpClass()
+        cls.record = cls.env["crm.lead"].create(
+            {
+                "name": "Test lead",
+                "partner_id": cls.partner.id,
+                "user_id": cls.user.id,
+                "team_id": cls.team.id,
+            }
+        )
+
```
cvinh, line 24: Tests with the case multi-team should be added and see why this one fails

cvinh:
> @cvinh the tests fail with the domains you gave me, could you check it?

Can you commit again to see where it fails ?

lef-adhoc:
@cvinh The failure occurs in the following line https://github.com/adhoc-dev/sale-workflow/blob/4cd5e8211dceceece3b54121c9b0e6c7c7420e63/sales_team_security/tests/common.py#L88

When searching with the self.user in the following line it does not find anything https://github.com/adhoc-dev/sale-workflow/blob/4cd5e8211dceceece3b54121c9b0e6c7c7420e63/sales_team_security/tests/common.py#L69

lef-adhoc:
@cvinh I think that's it, I also added a multi-team test, what do you think?

sales_team_security_crm/security/security.xml
```diff
@@ -0,0 +1,37 @@
...
+        </field>
+        <field
+            name="groups"
+            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
+        />
+    </record>
+    <record id="sales_team_team_rule" model="ir.rule">
+        <field name="name">Own Sale Teams</field>
+        <field name="model_id" ref="sales_team.model_crm_team" />
+        <field name="domain_force">[('id', 'in', user.crm_team_ids.ids)]</field>
```
cvinh, line 30:
```suggestion
        <field name="domain_force">['|', ('id', '=', 'user.sale_team_id.id'), ('id', 'in', user.crm_team_ids.ids)]</field>
```

cvinh (commented):
I think if a user is in one team only, the crm_team_ids are empty
If I'm wrong then it's ok as you did

sales_team_security_crm/security/security.xml
lef-adhoc: When there is only one team, crm_team_ids is not empty; it contains that single team

lef-adhoc:
@cvinh 
In this pr #3430 I have changed the domain to adapt it to multi team and added tests, what do you think ?

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

PR #3433, opened 2024-11-26
[MIG] sale_order_line_sequence: Migration to 18.0

BT-mchechlacz:


BT-nfrancisco (approved):

sale_order_line_sequence/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_sale_order_line_sequence
```
BT-crodriguez, line 1: **[change]** Missing license header

`# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).`

sale_order_line_sequence/views/account_move_view.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
BT-crodriguez, line 1: **[change]** Missing license header

```xml
<?xml version="1.0" encoding="utf-8" ?>
<!--
  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-->
```

sale_order_line_sequence/views/report_invoice.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
BT-crodriguez, line 1: **[change]** Missing license header

```xml
<?xml version="1.0" encoding="utf-8" ?>
<!--
  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-->
```

sale_order_line_sequence/views/report_saleorder.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
BT-crodriguez, line 1: **[change]** Missing license header

sale_order_line_sequence/views/sale_view.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
BT-crodriguez, line 1: **[change]** Missing license header

BT-crodriguez (changes requested):
LGTM just some licenses missing

sale_order_line_sequence/tests/__init__.py
rousseldenis: @BT-crodriguez License headers are not needed in `__init__.py` files.

sale_order_line_sequence/views/account_move_view.xml
StefanRijnhart: Are licenses needed in each file? https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst only mentions the manifest.

sale_order_line_sequence/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+
+    related_so_sequence = fields.Char(
+        string="SO Line Number",
+        compute="_compute_related_so_sequence",
+    )
+
+    @api.depends("move_id.invoice_line_ids")
+    def _compute_related_so_sequence(self):
+        for rec in self:
+            if len(rec.mapped("move_id.line_ids.sale_line_ids.order_id")) > 1:
```
StefanRijnhart, line 18: Given that you are touching this line in the migration commit, better go all the way with 
```suggestion
            if len(rec.move_id.line_ids.sale_line_ids.order_id) > 1:
```

sale_order_line_sequence/views/report_saleorder.xml
```diff
@@ -0,0 +1,21 @@
...
+<odoo>
+    <template
+        id="report_saleorder_document_sequence"
+        inherit_id="sale.report_saleorder_document"
+    >
+        <xpath expr="//table[hasclass('table')]/thead/tr/th[1]" position="before">
```
StefanRijnhart, line 7: Can you make the table selection more specific? I see multiple tables with this class in the document. I think each does have a unique class that you can add here.

sale_order_line_sequence/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2017 ForgeFlow S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale Order Line Sequence",
+    "summary": "Propagates SO line sequence to invoices and stock picking.",
+    "version": "18.0.1.1.0",
```
StefanRijnhart, line 8:
```suggestion
    "version": "18.0.1.0.0",
```

sale_order_line_sequence/views/sale_view.xml
```diff
@@ -0,0 +1,24 @@
...
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <field name="order_line" position="before">
+                <field name="max_line_sequence" invisible="1" />
```
StefanRijnhart, line 9: Fields like this, that will never be visible and are only used in attributes of other fields are now added automatically in Odoo 18. I think we can consider removing them from our views for a slightly cleaner layout. What do you think?

I only added this to the migration guidelines yesterday, so you will not have seen it before.
> Missing fields used in domains and other attributes are added to the view as invisible automatically. If you encounter fields in views with invisible="True", they can probably be removed. Reference: https://github.com/odoo/odoo/pull/137031

StefanRijnhart:
Superseded by https://github.com/OCA/sale-workflow/pull/3562

PR #3423, opened 2024-11-25
[16.0][ADD] sale_stock_prebook

lmignon:
Port of `sale_stock_prebook` addon from #2624 to 16.0

PR #3436, opened 2024-11-27
[16.0][Add] sale_order_blanket_order

lmignon:
requires:

* [ ] #3423 
* [ ] #3457 

# Context

When a company sells the same products to the same customers on a regular basis, it's a common business practice to create a sale framework that defines the terms and conditions of the sales.

If you need a way to define:
* the terms and conditions of the sales,
* the payment terms,
* the delivery terms,

and also secure the quantities of the products to be delivered, the sale framework module is the right choice.

This module introduces 2 new kinds of sale orders:

1. Sale Blanket Order: This is a sale order that defines the terms and conditions of the sales, the payment terms, the delivery terms, and secures the quantities of the products to be delivered. It is used to create sale orders that will be delivered in the future.   

2. Call of order: This is a sale order that is created to consume the quantities of the products secured in the sale blanket order.

Others modules can be used to provide the same kind of features. For example, the module (sale_blanket_order)[https://pypi.org/project/odoo-addon-sale-blanket-order] also defines the concept of sale blanket order. The main difference between the two modules is that the sale framework module extends the sale order model to add the sale blanket order and the call of order. This allows to keep the benefits of all the extensions made on the sale order model by other modules without having to adapt them to the sale blanket order model (discount, invoicing; inventory process, ...).

This module extends the functionality of Sale Order to support Blanket Order and Call of Order.

# Blanket Order

A Blanket Order is a standard sales order with the following specific features:

* Type: Classified as "Blanket Order".
* Defined Duration: Includes a validity period (end date).
* Payment Terms: Allows selection of preferred terms (e.g., 90 days end of month, upon delivery, etc.).
* Invoicing Policy: Can be based on product settings or the order itself.
* Stock Reservation: Allows advance reservation of sold quantities.
* Handling Unfulfilled Quantities: Provides options for dealing with undelivered quantities upon order expiration.
* Prices are calculated based on existing rules since it is a standard sales order type.

The blanket order serves as the central element triggering stock management and invoicing mechanisms.

## Stock Management
Delivered quantities are tracked on the sales order lines as with regular sales orders.
The goods are not delivered upon confirmation of the blanket order, but stock can be reserved for the customer using OCA modules:

- **sale_manual_delivery:** Marks sales orders as "manual delivery," preventing stock reservation or delivery preparation.
- **sale_stock_prebook:** Marks sales orders as "to reserve," triggering stock reservation but blocking delivery preparation for such cases.

## Invoicing

Standard invoicing policies apply (e.g., invoice on order or on delivery). Payment terms are configurable per order. Prepayment can be enforced by configuring the invoicing policy at the order level using the OCA module (sale_invoice_policy)[https://pypi.org/project/odoo-addon-sale-invoice-policy/].

## Consumption Management

A wizard will be available on the blanket order to initiate a delivery. It allows users to select products and quantities for delivery. This action creates a Call-off Order linked to the blanket order.

# Call-off Order

A Call-off Order is a standard sales order with these specific characteristics:

* Type: Classified as "Call-off Order".
* Linked to Blanket Order: Only includes products from the blanket order.
* Delivery Release: Enables the release of reserved stock for delivery.
* No Invoicing or Stock Management: These are handled via the linked blanket order.

## Stock Management

No delivery is generated directly from the call-off order.

It triggers:
* Release of the reserved quantity in the blanket order.
* Adjustment of stock reservations for the remaining quantities.

# Standard Sales Orders

To support existing workflows (e.g., e-commerce), call-off orders can be generated transparently from standard sales orders based on product and availability:

Entire orders may be converted into call-off orders if all products are linked to a blanket order.
Mixed orders split call-off items into a new call-off order, with both confirmed within the available quantities of the blanket order.


sale_framework/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Sale Framework",
+    "summary": """Manage blanket order and call of ordr""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": [
+        "sale_manual_delivery",
+        "sale_stock_prebook",
```
pedrobaeza: These dependencies seem too much for a "framework".

pedrobaeza (commented):
There's also `sale_blanket_order` in OCA.

lmignon:
> There's also `sale_blanket_order` in OCA.

see https://github.com/OCA/sale-workflow/pull/3436/files#diff-69478c720e8bc6a98546ecc5284bf2f95193e55918008899fe0160fc5e10972cR150

rvalyi:
> > There's also sale_blanket_order in OCA.

..and even a "vault"  module in server-auth I dare to say (désolé ; french only).

---->[]

pedrobaeza:
In my experience in the past, extending a base model to behave different, requires more code to contain the undesired side effects than to add such features. Anyway, as said, having such heavy dependencies for the base module is too much IMO.

sale_framework/__manifest__.py
lmignon: It's probably not the best name. I have no other idea as the name “sale_blanket_order” is already in use.

pedrobaeza: Yeah, but it's not only a question of the name, but to require such dependencies, that I think they should be extension modules: `sale_framework_manual_delivery` and `sale_framework_stock_prebook`. It also highlights that even using `sale.order` as base, it requires you some glue code for making it work. I would include also `order` on the name to be more specific. It's true that is not easy to name this. What about `sale_order_multi_purpose`?

lmignon:
> In my experience in the past, extending a base model to behave different, requires more code to contain the undesired side effects than to add such features. Anyway, as said, having such heavy dependencies for the base module is too much IMO.

We analyzed the existing code and the Odoo code in depth with @jbaudoux  and given the need to manage invoicing, stock reservation, etc. at the blanket order level, it seemed simpler to us to define new SO types.  Otherwise, all the specific flows for price management, invoicing, etc. would have had to be duplicated for new models. I've tried to clearly present the constraints and objectives in the 'readme' files. 

The module is not a base nor a simple module and may be split into several modules if possible. Work in progress

pedrobaeza:
> ..and even a "vault" module in server-auth I dare to say (désolé ; french only).

I don't get the joke I'm afraid.

-------

I don't insist more on the architecture, but a module with a generic name should be generic, not so specific of your project, including what `prebook` may be. And this repo is also very crowded to have even more deformations on standard objects. I wonder if this would be something to be on a vertical like `wms`.

sale_framework/__manifest__.py
jbaudoux: rename this module to `sale_order_blanket_order` as it allows to define the sale order type of an order as a blanket order.

jbaudoux: also remove the dependency to stock from this module and make another one for the glue with stock

sale_framework/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale Framework",
+    "summary": """Manage blanket order and call of ordr""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
jbaudoux: :O
```suggestion
    "author": "ACSONE SA/NV,BCIM,Odoo Community Association (OCA)",
```

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,479 @@
...
+from odoo.addons.sale.models.sale_order import (
+    LOCKED_FIELD_STATES,
+    READONLY_FIELD_STATES,
+)
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    order_type = fields.Selection(
```
youring, line 20: More specific naming suggestion: order_blanket_type or order_frame_type

youring (changes requested):
Hi @lmignon thanks for the huge work. Just a minor suggestion for naming.
IMHO, order_type is too common to be used here to just distinguish whether a sales order is a normal, blanket or call-off, and note that there's already a module `sale_order_type` which specializes in managing various order types. 
From a functional user perspective, I think a more specific field name here would give the user a better idea of what it represents, thus less headaches.

lmignon:
> Hi @lmignon thanks for the huge work. Just a minor suggestion for naming. IMHO, order_type is too common to be used here to just distinguish whether a sales order is a normal, blanket or call-off, and note that there's already a module `sale_order_type` which specializes in managing various order types. From a functional user perspective, I think a more specific field name here would give the user a better idea of what it represents, thus less headaches.

Thank you the review. @youring  I'll think about your comment about the naming even if I don't have a better idea at the moment. Once our functional tests are completed, I'll mark this PR as ready for review.

sale_order_blanket_order/models/sale_order.py
jbaudoux: `order_type` is referring to the ISO naming here https://docs.peppol.eu/poacc/upgrade-3/codelist/UNCL1001_T01/

if it's not conflicting with another module, I find it a proper name.

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,495 @@
...
+        [
+            ("normal", "Normal Sale Order"),
+            ("blanket", "Blanket Order"),
+            ("call_off", "Call-off Order"),
```
jbaudoux:
```suggestion
            ("order", "Order"),
            ("blanket", "Blanket Order"),
            ("call_off", "Call off Order"),
```

sale_order_blanket_order/models/sale_order_line.py
```diff
@@ -0,0 +1,554 @@
...
+                        product=line.product_id.display_name,
+                    )
+                )
+
+    @api.constrains(
+        "blanket_validity_start_date",
+        "blanket_validity_end_date",
+        "product_id",
+    )
+    def _check_blanket_product_not_overlapping(self):
```
jbaudoux: Why do you want to prevent this?

sale_order_blanket_order/models/sale_order_line.py
```diff
@@ -0,0 +1,554 @@
...
+
+    @api.constrains(
+        "state",
+        "blanket_order_id",
+        "order_type",
+        "product_id",
+        "product_uom_qty",
+        "display_type",
+    )
+    def _check_call_off_order_line(self):
```
jbaudoux: I would call this check on confirm and not as a constrain. If you receive a wrong order, it's better to create it and then fix the issues than preventing to save it.

sale_order_blanket_order/views/sale_order.xml
```diff
@@ -0,0 +1,343 @@
...
+                        name="call_off_order_count"
+                        widget="statinfo"
+                        string="Call-offs"
+                    />
+                </button>
+
+            </xpath>
+            <field name="partner_id" position="before">
+                <field name="order_type" />
+            </field>
```
jbaudoux: I would place it before order name (sequence) like odoo is doing for invoice_type

sale_order_blanket_order/models/sale_order_line.py
lmignon: Otherwise it becomes difficult to know which blanket line to choose in case of automatic creation of call off order at normal order confirmation. (Reading your comment and the list of declared tiggers , I see a potential bug since all the fields used to detect 'double' are not into the list of triggers )

lmignon: done

sale_order_blanket_order/i18n/fr.po
```diff
@@ -0,0 +1,530 @@
...
+
+#. module: sale_order_blanket_order
+#: model:ir.model.fields,field_description:sale_order_blanket_order.field_sale_order__call_off_order_count
+msgid "Call-off Order Count"
+msgstr "Nbr de commandes d'appel"
+
+#. module: sale_order_blanket_order
+#: model:ir.model.fields,field_description:sale_order_blanket_order.field_sale_order_line__call_off_line_ids
+msgid "Call-off Order Lines"
+msgstr "Linges de commande d'appel"
```
jbaudoux:
```suggestion
msgstr "Lignes de commande d'appel"
```

sale_order_blanket_order/i18n/fr.po
```diff
@@ -0,0 +1,530 @@
...
+"système libérera la réservation. Si la stratégie est 'Livrer les quantités "
+"restantes', le système créera automatiquement un bon de livraison pour les "
+"quantités restantes."
+
+#. module: sale_order_blanket_order
+#: model:ir.model.fields,help:sale_order_blanket_order.field_sale_order__order_type
+#: model:ir.model.fields,help:sale_order_blanket_order.field_sale_order_line__order_type
+msgid "Specifies the type of sale order: Order, Blanket, or Call-off."
+msgstr ""
+"Spécifie le type de commande client : Normal, Commande-cadre ou Commande "
```
jbaudoux:
```suggestion
"Spécifie le type de commande client : Commande, Commande-cadre ou Commande "
```

sale_order_blanket_order/models/res_company.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    create_call_off_from_so_if_possible = fields.Boolean(
+        string="Create Call-Off from SO if possible",
+        help="If checked, when a sale order is confirmed and some lines refer to a "
```
jbaudoux:
```suggestion
        help="If checked, when a sales order is confirmed and some lines refer to a "
```

sale_order_blanket_order/models/res_config_settings.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    create_call_off_from_so_if_possible = fields.Boolean(
+        related="company_id.create_call_off_from_so_if_possible",
+        readonly=False,
+        string="Create Call-Off from SO if possible",
+        help="If checked, when a sale order is confirmed and some lines refer to a "
```
jbaudoux:
```suggestion
        help="If checked, when a sales order is confirmed and some lines refer to a "
```

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+                and order.blanket_order_id.order_type != "blanket"
+            ):
+                raise ValidationError(_("A call-off order must have a blanket order."))
+            if order.order_type == "order" and order.blanket_order_id:
+                raise ValidationError(_("An order cannot have a blanket order."))
+
+    @api.constrains(
+        "order_type", "blanket_validity_start_date", "blanket_validity_end_date"
+    )
+    def _check_validity_dates(self):
```
jbaudoux: Should happen on confirm and not as a constrain

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+    @api.constrains("order_type", "blanket_order_id")
+    def _check_order_type(self):
+        for order in self:
+            if order.order_type == "blanket" and order.blanket_order_id:
+                raise ValidationError(_("A blanket order cannot have a blanket order."))
+            if (
+                order.order_type == "call_off"
+                and order.blanket_order_id.order_type != "blanket"
+            ):
+                raise ValidationError(_("A call-off order must have a blanket order."))
```
jbaudoux, line 142: This one should be on confirm

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+                if order.blanket_validity_end_date < order.blanket_validity_start_date:
+                    raise ValidationError(
+                        _(
+                            "The validity end date must be greater than the "
+                            "validity start date."
+                        )
+                    )
+
+    @api.constrains("order_type", "blanket_order_id", "date_order", "commitment_date")
+    def _check_call_of_link_to_valid_blanket(self):
```
jbaudoux: on confirm

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+            ):
+                raise ValidationError(
+                    _(
+                        "The call-off order must be within the validity period of "
+                        "the blanket order."
+                    )
+                )
+
+    @api.constrains("order_type", "blanket_order_id")
+    def _check_blanket_order_state(self):
```
jbaudoux: on confirm

jbaudoux (commented):
Minor remarks.
You should not prevent importing & saving.
Most constrains should be evaluated on confirm.

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+
+        This method is called for orders. If some order lines are related
+        to a blanket order, it will create a call-off order for each of them and
+        remove them from the original order.
+
+        The method returns the call-off orders that have been created or an empty
+        recordset if no call-off orders have been created.
+        """
+        if any(self.filtered(lambda order: order.order_type != "order")):
+            raise ValidationError(_("Only orders can be split."))
```
jbaudoux: Can you raise such outside a constrains ?

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+                    0.0,
+                    precision_rounding=blanket_lines[0].product_uom.rounding,
+                )
+                <= 0
+            ):
+                continue
+            for line in lines:
+                for blanket_line in blanket_lines:
+                    blanket_order = blanket_line.order_id
+                    call_off = new_call_off_by_order[line.order_id]
```
jbaudoux, line 483: I would have put an ensure_one, I don't see what you gain from processing lines from different orders except risking to mix up stuffs

Nevertheless, I would make a `new_call_off_by_blanket` as matching several blanket orders will be soon a requested use case

lmignon: > Nevertheless, I would make a `new_call_off_by_blanket` as matching several blanket orders will be soon a requested use case

While it's not a requested use case, I'll not add more complexity here.

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+                            qty_deliverable,
+                            line.product_uom_qty,
+                            precision_rounding=rounding,
+                        )
+                        >= 0
+                    ):
+                        # The blanket line has enough remaining quantity for the line
+                        # We move the line to the call-off order
+                        line.price_unit = 0
+                        line.order_id = call_off
```
jbaudoux, line 500: what you do with the original order if you moved all the lines to a call-off ? Shouldn't you cancel it ?

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,497 @@
...
+        self.ensure_one()
+        return self.copy(self._get_default_call_off_order_values(blanket_order_id))
+
+    def _get_single_blanket_order_candidates_domain(self):
+        """Get the domain to search for a blanket order candidates."""
+        self.ensure_one()
+        validity_date = self.commitment_date or self.date_order or fields.Date.today()
+        validity_date = fields.Date.to_string(validity_date)
+        return [
+            ("partner_id", "=", self.partner_id.id),
```
jbaudoux, line 596: I would also add the `partner_shipping_id`

sale_order_blanket_order/models/sale_order.py
lmignon: The order is done and empty. You'll see into the shatter that all the lines are moved to a calloff

sale_order_blanket_order/models/sale_order_line.py
```diff
@@ -0,0 +1,590 @@
...
+            )
+        return res
+
+    def _link_to_blanket_order_line(self):
+        """Link the call-off order lines to the corresponding blanket order lines.
+
+        This method is called at the confirmation time of call-off orders. It will
+        link each call-off line to the corresponding blanket line. If the quantity on
+        the call-off line is greater than the quantity on the blanket line, the
+        call-off line will be split to ensure taht the quantity on the call-off line
```
jbaudoux:
```suggestion
        call-off line will be split to ensure that the quantity on the call-off line
```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
+When a company sells the same products to the same customers on a regular basis, it's a common business practice to create a sale framework that defines the terms and conditions of the sales.
+
+If you need a way to define:
+* the terms and conditions of the sales,
+* the payment terms,
+* the delivery terms,
+
+and also secure the quantities of the products to be delivered, the sale framework module is the right choice.
+
+This module introduces 2 new kinkds of sale orders:
```
jbaudoux:
```suggestion
This module introduces 2 new kinds of sales orders:
```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
+When a company sells the same products to the same customers on a regular basis, it's a common business practice to create a sale framework that defines the terms and conditions of the sales.
+
```
jbaudoux:
```suggestion
When a company sells the same products to the same customers on a regular basis, it's a common business practice to create a blanket order that defines the terms and conditions of the sales.

```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
...
+If you need a way to define:
+* the terms and conditions of the sales,
+* the payment terms,
+* the delivery terms,
+
+and also secure the quantities of the products to be delivered, the sale framework module is the right choice.
```
jbaudoux:
```suggestion
and also secure the quantities of the products to be delivered, the sale order blanket order module is the right choice.
```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
...
+
+This module introduces 2 new kinkds of sale orders:
+
+1. Sale Blanket Order: This is a sale order that defines the terms and conditions of the sales, the payment terms, the delivery terms, and secures the quantities of the products to be delivered. It is used to create sale orders that will be delivered in the future.
```
jbaudoux:
```suggestion
1. Blanket Order: This is a sales order that defines the terms and conditions of the sales, the price, the payment terms, the delivery terms, and secures the quantities of the products to be delivered.
```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
...
+* the payment terms,
+* the delivery terms,
+
+and also secure the quantities of the products to be delivered, the sale framework module is the right choice.
+
+This module introduces 2 new kinkds of sale orders:
+
+1. Sale Blanket Order: This is a sale order that defines the terms and conditions of the sales, the payment terms, the delivery terms, and secures the quantities of the products to be delivered. It is used to create sale orders that will be delivered in the future.   
+
+2. Call of order: This is a sale order that is created to consume the quantities of the products secured in the sale blanket order.
```
jbaudoux:
```suggestion
2. Call of order: This is a sales order linked to a blanket order that is created to trigger the delivery of quantities of the products secured in the blanket order.
```

sale_order_blanket_order/readme/CONTEXT.md
```diff
@@ -0,0 +1,16 @@
...
+
+Others modules can be used to provide the same kind of features. For example, the module (sale_blanket_order)[https://pypi.org/project/odoo-addon-sale-blanket-order] also defines the concept of sale blanket order. The main difference between the two modules is that the sale framework module extends the sale order model to add the sale blanket order and the call of order. This allows to keep the benefits of all the extensions made on the sale order model by other modules without having to adapt them to the sale blanket order model (discount, invoicing; inventory process, ...).
```
jbaudoux:
```suggestion
Others modules can be used to provide the same kind of features. For example, the module (sale_blanket_order)[https://pypi.org/project/odoo-addon-sale-blanket-order] also defines the concept of sale blanket order. The main difference between the two modules is that the sale order blanket order module extends the sale order model to add the sale blanket order and the call off order. This allows to keep the benefits of all the extensions made on the sale order model by other modules without having to adapt them to the sale blanket order model (discount, invoicing; inventory process, ...).
```

sale_order_blanket_order/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,56 @@
+This module extends the functionality of Sale Order to support Blanket Order and Call of Order.
```
jbaudoux:
```suggestion
This module extends the functionality of Sale Order to support Blanket Order and Call-off Order.
```

sale_order_blanket_order/readme/USAGE.md
```diff
@@ -0,0 +1,3 @@
+By default, the automatic creation of call-off orders from normal sale orders containing products
+part of a blanket order is disabled. To enable this feature, you need to go into the sales settings
+and enable the option "CCreate Call-Off from SO if possible".
```
jbaudoux:
```suggestion
and enable the option "Create Call-Off from SO if possible".
```

sale_order_blanket_order/models/sale_order.py
```diff
@@ -0,0 +1,580 @@
...
+        """
+        if any(self.filtered(lambda order: order.order_type != "order")):
+            raise ValidationError(_("Only orders can be split."))
+
+        splitable_orders = self.filtered(
+            lambda order: order.company_id.sudo().create_call_off_from_so_if_possible
+        )
+        if not splitable_orders:
+            return self.browse()
+        blanket_order_candidates = splitable_orders._get_blanket_order_candidates()
```
jbaudoux, line 450: my main remark is to loop on order here

sale_order_blanket_order/models/sale_order_line.py
```diff
@@ -0,0 +1,604 @@
...
+    def _get_call_off_line_to_blanked_line_matching_fields(self):
+        """Get the fields used to match call-off order lines to blanket order lines.
+
+        Be careful to override this method if you want to add new fields to the matching
+        key. You must most probably extend the list of fields triggering the constraint
+        `_check_blanket_product_not_overlapping` since these fields are used to check
+        that a product is not part of multiple blanket orders with overlapping validity
+        periods.
+        """
+        return ["product_id", "product_packaging_id", "order_partner_id"]
```
jbaudoux, line 297: and keep here only line related matching key (so no check again on partner)

jbaudoux (commented):
my main remark concerns this `_split_for_blanket_order` where I would lop on orders after the search. You can still keep one search and reapply the domain on the search result if you have multiple orders in self. This should simplify a bit the code and reduce the risk of mixing stuffs. It will also reduce the set of lines when doing the lines matching as you have multiple for loops in cascade.

nilshamerlinck:
you probably noticed it already as I see you isolated them, but fyi tests seem to get stuck there:

image

and because builds were triggered for each of the last commits, it created a bit of traffic jam in the ci queue for some moment :sweat_smile: (until Stéphane killed pending jobs)

sale_order_blanket_order/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+
+{
+    "name": "Sale Order Blanket Order",
+    "summary": """Manage blanket order and call of ordr""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,BCIM,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": [
+        "sale_manual_delivery",
```
rousseldenis, line 12: @lmignon The install of this addon does not work on large database.

See: https://github.com/OCA/sale-workflow/issues/3476

lmignon: @rousseldenis 
It worked on my very large database.... (even if it was not fast and should be improved)

lmignon:
Modules moved to new repository https://github.com/OCA/sale-blanket/pull/2

PR #3439, opened 2024-11-28
[14.0][ADD] new module sale_order_qty_change_no_zero

clementmbr:
Prevent unit price to be zero when sale order line quantity is changed

Variant/alternative of [sale_order_qty_change_no_recompute](https://github.com/OCA/sale-workflow/tree/14.0/sale_order_qty_change_no_recompute)

cc @Kev-Roche @sebastienbeau


clementmbr:
This is the sale version of https://github.com/OCA/purchase-workflow/pull/2474

PR #3442, opened 2024-11-29
[FIX] [16.0] sale_order_product_recommendation_packaging_default: Do not update packaging if quantities has not changed

Shide:
Do not update packaging quantities if quantities has not changed.

This [line](https://github.com/OCA/sale-workflow/blob/16.0/sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py#L84) triggers recomputation and clear `discount`, so avoid calling unncessary function.

The `if` when creating is used for maintain consistency in case we modify something in the future.

MT-7993 @moduon @rafaelbn @Gelojr @yajo @fcvalgar please review if you want :)

### DIFF
--- a/sale_order_product_recommendation_packaging_default/tests/test_recommendation.py
+++ b/sale_order_product_recommendation_packaging_default/tests/test_recommendation.py
@@ -188,6 +188,35 @@ class PackagingRecommendationCase(test_recommendation_common.RecommendationCase)
             ],
         )
 
+    def test_preexisting_product_without_modification(self):
+        self.new_so.order_line = [
+            Command.create(
+                {
+                    "product_id": self.prod_1.id,
+                    "product_uom_qty": 12,
+                    "qty_delivered_method": "manual",
+                    "discount": 12.0,
+                }
+            )
+        ]
+        # Ensure pricelist wants to recompute discounts
+        self.new_so.sudo().pricelist_id.discount_policy = "without_discount"
+        # Do not modify anything
+        wiz = self.wizard()
+        wiz.action_accept()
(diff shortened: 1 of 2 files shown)


Gelojr (approved):
LGTM Nice job @Shide 
https://www.loom.com/share/1657a0284dfb481a98b7735894a18d6d?sid=fc96614b-c532-4f01-9bae-15cdb6b4880d

sale_order_product_recommendation_packaging_default/tests/test_recommendation.py
```diff
@@ -188,6 +188,33 @@ def test_preexisting_product(self):
...
+        wiz.action_accept()
+        self.assertRecordValues(
+            self.new_so.order_line,
+            [
+                {
+                    "product_id": self.prod_1.id,
+                    "product_packaging_id": self.prod_1_dozen.id,
+                    "product_packaging_qty": 1,
+                    "product_uom_qty": 12,
+                    "discount": 12.0,  # Discount is preserved
```
yajo, line 215: Are you sure this test is actually working? If you don't set a pricelist, I think it would be preserved anyway.

Shide: Default pricelist is always applied. 

Every time a pricelist with Discount Policy "Show discount and price" recomputes something, discounts are reset.
If the Discount Policy on the pricelist is "Discount included in price", preserves used discounts.

I'll set the discount policy on the pricelist just in case.

yajo (approved):
/ocabot merge patch

PR #3441, opened 2024-11-29
[18.0][MIG] sale_automatic_workflow: Migration to 18.0

chaule97:
- I format-patch from https://github.com/OCA/sale-workflow/pull/3375

PR #3447, opened 2024-12-03
[18.0][MIG] sale_automatic_workflow_stock: Migration to 18.0

chaule97:
-  Depend on: [sale_automatic_workflow: Migration to 18.0 ](https://github.com/OCA/sale-workflow/pull/3441)
- This migration includes the integration of the `sale_automatic_workflow_delivery_state` module. The `all_qty_delivered` field from the `sale_automatic_workflow` module will now rely on the `delivery_status` field in the `sale_stock` module for its calculations


sale_automatic_workflow_stock/tests/common.py
```diff
@@ -8,14 +8,22 @@
 class TestAutomaticWorkflowStockMixin(TestAutomaticWorkflowMixin):
     """Extend to add stock related workflow."""
 
-    def create_sale_order(self, workflow, override=None, product_type="product"):
+    def create_sale_order(
+        self, workflow, override=None, product_type="consu", extra_product_values=None
+    ):
+        extra_product_values = extra_product_values or {}
+        extra_product_values["is_storable"] = True
```
sebalix: I would add a condition on this:
```python
if product_type == "consu":
    extra_product_values["is_storable"] = True
```

sebalix (approved):
LG, minor comment

ivantodorovich (approved):

sale_automatic_workflow_stock/readme/CREDITS.md
```diff
@@ -0,0 +1 @@
+The migration of this module from 17.0 to 18.0 was financially supported by Camptocamp.
```
rvalyi, line 1: This is only my opinion, but I think this is a pretty trivial module migration and if everytime someone migrates a modules he starts adding a section like that telling company X put the money for it, then the OCA appstore will quickly end up like the click bait blinking gifs of Odoo appstore. This is a bit concerning if Camptocamp starts showing some bad example here... I mean come on guys this doesn't scale if every little trivial migration start doing that... we need some self moderation and common sense... Eventually you guys can try to push so that the OCA better emphatize the modules authors, contributors and maintainers and this is a place where CampToCamp already shines. But starting to add such sections for trivial migrations is just commercial noise that doesn't scale IMHO.

sebalix: I disagree, this `CREDITS` fragment is here for that reason. And IMO even a migration that doesn't introduce big changes to work on a new version, requires in fact some burden: checking every pieces of code extending the Odoo std if they are still OK, testing and making sure it works as expected.

And IMO we should get only one migration text regarding this for a given version, for Odoo 19.0 the "17.0 to 18.0" sentence should be removed.

mmequignon: This discussion belongs on the contributors mailing list, I guess.

simahawk: It's also clear this is not a "pretty trivial migration".

PR #3449, opened 2024-12-03
[18.0][MIG] sale_force_whole_invoiceability: Migration to 18.0

Naraka:
Migration to 18.0


pedrobaeza (approved):
/ocabot migration sale_force_whole_invoiceability

BhaveshHeliconia (approved):
Functional review LGTM!

alexey-pelykh (changes requested):
Thanks for the migration, the 18.0 API changes look good (`self.env._()`, `self.env.user.has_group()`).

There's a logic bug in the view though — the `invisible` domain on the button was changed from `or` to `and`:

```xml
invisible="state not in ['sale'] and invoice_count > 0"
```

In 17.0 this was `or`, which is correct: hide the button when the state is not `sale` **or** when there are already invoices. With `and`, the button becomes visible in draft (since `invoice_count > 0` is false) and also stays visible on confirmed orders that already have invoices (since `state not in ['sale']` is false). Should be reverted to `or`:

```xml
invisible="state not in ['sale'] or invoice_count > 0"
```

pedrobaeza:
@Naraka please include #3621 before migration commit.

flotho (approved):
hi,
thanks,
easy code review
LGTM,
Thanks

rousseldenis:
This has to be fixed but @Naraka is not active anymore.

PR #3400, opened 2024-11-11
[14.0] [MIG] sale_order_product_availability_inline

Tiago370:
---

PR #3451, opened 2024-12-03
[14.0][ADD] sale_order_line_display_stock_per_warehouse

Kev-Roche:
module to show available stock per warehouse on sale order lines


sale_order_line_display_stock_per_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,41 @@
...
+                        .read_group(
+                            domain=[
+                                ("product_id", "=", line.product_id.id),
+                                ("location_id", "child_of", warehouse.lot_stock_id.id),
+                            ],
+                            fields=["quantity:sum"],
+                            groupby=[],
+                        )[0]["quantity"]
+                    )
+                    qty_available = qty_data if qty_data else 0
```
sebastienbeau: Make it simplier and correct. 
```suggestion
                    qty_available = line.product_id.with_context(warehouse=warehouse.id).qty_available
```

With the current implementation you have the quantity that you have physically in stock. What we need is the quantity that you can sell. The field qty_available depends on the context passed in the warehouse so it' simplier to reuse this field.

Kev-Roche: thank you for your review, I fixed it.

rousseldenis:
@Kev-Roche @sebastienbeau Have you explored the possibility to extend this? https://github.com/OCA/sale-workflow/pull/3400

rousseldenis:
@Kev-Roche @sebastienbeau

rousseldenis:
@Kev-Roche Could we move forward on this or we close ?

Kev-Roche:
Hi @rousseldenis, I don't think this module has the same purpose as this one: https://github.com/OCA/sale-workflow/pull/3400. I'd propose keeping them both, if that's OK.

FYI @metaminux, related to your proposal: I added the choice between qty_available, free_qty and virtual_available. Depending on the use case, one of these quantity fields will be more relevant.

metaminux (approved):

rousseldenis (approved):

FrancoMaxime (approved):
LGTM: code review

Some strange indentation in the xml files.

rousseldenis:
Name reserved.

/ocabot merge nobump

PR #3457, opened 2024-12-04
[16.0][FIX] sale_order_product_picker: Fix addon installation when this addon is installed

lmignon:
It's a bad idea to use the registry at initialization phase. When you access the registry some info are set into the cache. For exemple the field's ids list of a given model. That means that new field added to the model later into the installation process will not be returned by a new call to the _get_ids method of the ir.model.fields model. At the end, the installation of new addon adding, for exemple, a selection field to the sale.order model is no more possible.

This fix solve https://github.com/OCA/sale-workflow/actions/runs/12144814546/job/33865050704?pr=3436#step:8:2078

### DIFF
--- a/sale_order_product_picker/models/sale_order.py
+++ b/sale_order_product_picker/models/sale_order.py
@@ -60,12 +60,23 @@ class SaleOrder(models.Model):
     def _get_partner_picker_field(self):
         # HACK: To avoid installation error when get default value to be used in a
         # depends of a computed field
-        if "sale_order_product_picker" in self.env.registry._init_modules and self.env[
-            "ir.default"
-        ].get("sale.order", "use_delivery_address"):
-            return "partner_shipping_id"
-        else:
-            return "partner_id"
+        if "sale_order_product_picker" in self.env.registry._init_modules:
+            use_delivery_address_defined = self.env["ir.default"].get(
+                "sale.order", "use_delivery_address"
+            )
+            # accessing the registry into the initialization phase is a really bad
+            # idea and could lead to a lot of issues. For example, it will load the
+            # cache of the field's id of the model. That means that starting from this
+            # new field added to the model will not be returned by the method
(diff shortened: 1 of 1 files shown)


lmignon:
ping @CarlosRoca13

CarlosRoca13:
Thanks, my colleague @carlosdauden is going to review the PR as he is the author of this code.

sale_order_product_picker/models/sale_order.py
```diff
@@ -60,12 +60,23 @@ def _list_product_picker_filters(self):
     def _get_partner_picker_field(self):
         # HACK: To avoid installation error when get default value to be used in a
         # depends of a computed field
-        if "sale_order_product_picker" in self.env.registry._init_modules and self.env[
-            "ir.default"
-        ].get("sale.order", "use_delivery_address"):
-            return "partner_shipping_id"
-        else:
-            return "partner_id"
+        if "sale_order_product_picker" in self.env.registry._init_modules:
```
rousseldenis, line 63: My question here:

The use of ir.default is questionning for me as it will not indicate which field is used through all sale orders... as this is a default for new ones.

So, we can have a mix through all sale orders.

IMHO, the implementation should be changed.

lmignon: @rousseldenis I'm not the author of the original code and I don't use this module. I only need to fix an error that prevent the installation of others addons once this one is installed... Plz don't block this.

rousseldenis: I'm not blocking this as this solves installs. I'm questioning about the implementation which seems wrong to me.

rousseldenis (approved):

rousseldenis:
@carlosdauden

carlosdauden (approved):

dreispt (approved):

PR #3458, opened 2024-12-04
[18.0][MIG] sale_order_line_input: Migration to 18.0

Naraka:
Migration to 18.0


sale_order_line_input/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+    def _onchange_order_partner_id(self):
+        """Create order to correct compute of taxes"""
+        if not self.order_partner_id or self.order_id:
+            return
+        SaleOrder = self.env["sale.order"]
+        new_so = SaleOrder.new({"partner_id": self.order_partner_id})
+        for onchange_method in new_so._onchange_methods["partner_id"]:
+            onchange_method(new_so)
+        order_vals = new_so._convert_to_write(new_so._cache)
+        self.order_id = SaleOrder.create(order_vals)
```
rousseldenis, line 53: I'm still reluctant to creates in onchanges...

Naraka: If I understand correctly, you mean that `create` shouldn't be used in `onchange`. Perhaps using `@api.depends` could solve this problem?

```python
    order_id = fields.Many2one("sale.order", string="Sale Order", compute="_compute_order_id", store=True)

    @api.depends("order_partner_id")
    def _compute_order_id(self):
        """Create order automatically when order_partner_id changes"""
        for line in self:
            if line.order_partner_id and not line.order_id:
                SaleOrder = self.env["sale.order"]
                new_so = SaleOrder.new({"partner_id": line.order_partner_id.id})
                # Apply the onchange methods (e.g., computing taxes)
                for onchange_method in new_so._onchange_methods["partner_id"]:
                    onchange_method(new_so)
                order_vals = new_so._convert_to_write(new_so._cache)
                line.order_id = SaleOrder.create(order_vals)
```
I also have a question that is not clear to me regarding the process,  shouldn't these types of changes be handled in a section other than migration? Thanks!

sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -0,0 +1,264 @@
...
+                    force_save="1"
+                    options='{"no_open": True, "no_create": True}'
+                />
+                <button
+                    title="Open order"
+                    name="action_sale_order_form"
+                    type="object"
+                    icon="fa-external-link"
+                    class="oe_stat_button"
+                />
```
carlosdauden, line 45:
```suggestion
                    options='{"no_create": True}'
                />
```

sale_order_line_input/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def action_sale_order_form(self):
+        self.ensure_one()
+        return self.order_id.get_formview_action()
```
carlosdauden, line 57:
```suggestion
```

carlosdauden (approved):
Maybe we can remove the button as clicking on the order_id gives the same result.

victoralmau:
Superseded by https://github.com/OCA/sale-workflow/pull/3795

PR #3460, opened 2024-12-05
[18.0][MIG] sale_automatic_workflow_job: Migration to 18.0 

chaule97:
-  Depend on:
    - https://github.com/OCA/sale-workflow/pull/3441
- Because [this commit](https://github.com/OCA/sale-workflow/pull/3105/commits/1d0e79e5accf69e71eef92284726b0e10af5e518) separated the module `stock` from `sale_automatic_workflow`.  I removed everything related to `stock`.
- because `_do_validate_sale_order` will be run in job, I add `_do_send_order_confirmation_mail` in job and it will be run after 2 minutes
- because queue_job affects test case in `sale_automatic_workflow`, I have removed tagged in `TestAutomaticWorkflow`


henrybackman (approved):

sale_automatic_workflow_job/models/automatic_workflow_job.py
```diff
@@ -9,20 +11,30 @@ class AutomaticWorkflowJob(models.Model):
...
+        description = self.env._("Validate sales order %s", sale.display_name)
         return {
             "description": description,
             "identity_key": identity_exact,
         }
 
+    def _do_send_order_confirmation_mail_job_options(self, sale):
+        description = self.env._("Send order %s confirmation mail", sale.display_name)
+        eta = datetime.now() + timedelta(minutes=2)
+        return {"description": description, "identity_key": identity_exact, "eta": eta}
```
sebalix: I won't do that, we could imagine big SO that takes more than 2min to be validated.
Instead, we should spawn a new job chained to the previous one (dependencies), so this one will start once the first is Done.

chaule97: I have updated it

yankinmax (approved):
I don't get finally, why don't you replace existing string formatting with f-string. Instead you use old style.

ivantodorovich (approved):

sale_automatic_workflow_job/models/automatic_workflow_job.py
```diff
@@ -9,20 +10,44 @@ class AutomaticWorkflowJob(models.Model):
     _inherit = "automatic.workflow.job"
 
     def _do_validate_sale_order_job_options(self, sale, domain_filter):
-        description = _("Validate sales order {}").format(sale.display_name)
+        description = self.env._("Validate sales order %s", sale.display_name)
```
vvrossem, line 13:
```suggestion
        description = self.env._(f"Validate sales order {sale.display_name}")

File: sale_automatic_workflow_job/models/automatic_workflow_job.py [hunk 3]
```diff
@@ -9,20 +10,44 @@ class AutomaticWorkflowJob(models.Model):
...
         return {
             "description": description,
             "identity_key": identity_exact,
         }
 
+    def _do_send_order_confirmation_mail_job_options(self, sale):
+        description = self.env._("Send order %s confirmation mail", sale.display_name)
```
vvrossem, line 20:
```suggestion
        description = self.env._(f"Send order {sale.display_name} confirmation mail")

File: sale_automatic_workflow_job/models/automatic_workflow_job.py [hunk 4]
```diff
@@ -9,20 +10,44 @@ class AutomaticWorkflowJob(models.Model):
...
+        )
         return super(AutomaticWorkflowJob, with_context)._validate_sale_orders(
             domain_filter
         )
 
     def _do_create_invoice_job_options(self, sale, domain_filter):
-        description = _("Create invoices for sales order {}").format(sale.display_name)
+        description = self.env._(
+            "Create invoices for sales order %s", sale.display_name
+        )
```
vvrossem, line 50:
```suggestion
        description = self.env._(
            f"Create invoices for sales order {sale.display_name}"
        )
```

sale_automatic_workflow_job/models/automatic_workflow_job.py
```diff
@@ -33,7 +58,7 @@ def _create_invoices(self, domain_filter):
         return super(AutomaticWorkflowJob, with_context)._create_invoices(domain_filter)
 
     def _do_validate_invoice_job_options(self, invoice, domain_filter):
-        description = _("Validate invoice {}").format(invoice.display_name)
+        description = self.env._("Validate invoice %s", invoice.display_name)
```
vvrossem, line 61:
```suggestion
        description = self.env._(f"Validate invoice {invoice.display_name}")
```

sale_automatic_workflow_job/models/automatic_workflow_job.py
```diff
@@ -45,21 +70,8 @@ def _validate_invoices(self, domain_filter):
...
-
-    def _validate_pickings(self, domain_filter):
-        with_context = self.with_context(auto_delay_do_validation=True)
-        return super(AutomaticWorkflowJob, with_context)._validate_pickings(
-            domain_filter
-        )
-
     def _do_sale_done_job_options(self, sale, domain_filter):
-        description = _("Mark sales order {} as done").format(sale.display_name)
+        description = self.env._("Mark sales order %s as done", sale.display_name)
```
vvrossem, line 74:
```suggestion
        description = self.env._(f"Mark sales order {sale.display_name} as done")
```

sale_automatic_workflow/tests/test_automatic_workflow.py
```diff
@@ -7,12 +7,10 @@
 from freezegun import freeze_time
 
 from odoo import fields
-from odoo.tests import tagged
 
 from .common import TestAutomaticWorkflowMixin, TestCommon
```
vvrossem: I would adapt [`TestCommon`](https://github.com/OCA/sale-workflow/blob/18.0/sale_automatic_workflow/tests/common.py#L11-L15) so it uses [`BaseCommon`](https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/tests/common.py#L20) instead of `TransactionCase`

> Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

sale_automatic_workflow/tests/test_automatic_workflow.py
```diff
@@ -7,12 +7,10 @@
...
 
 
-@tagged("post_install", "-at_install")
```
vvrossem: why removing?

vvrossem (approved):

sale_automatic_workflow_job/models/automatic_workflow_job.py
StefanRijnhart: No, in the case of the translation wrapper we want to keep the substitution values out of scope. It's fine to use the built-in substitution mechanism which uses the percent marker.

sale_automatic_workflow/tests/test_automatic_workflow.py
StefanRijnhart: The PR description explains that this is due to interference of queue_job (and presumably this module), but @chaule97 isn't that what https://github.com/OCA/sale-workflow/blob/18.0/sale_automatic_workflow/tests/test_automatic_workflow.py#L23-L26 is for? Currently, the tests are depending on a preinstalled chart of accounts whereas they should be inheriting from AccountTestInvoicingCommon which requires to be run post-install. This can be considered technical debt, and we should not be moving further away from its implementation.

gurneyalex:
I'm wondering why we are keeping the cron jobs in this module? 

This seems so inefficient, when we could have the queue jobs created on creation of the various records (sale, picking, invoice) based on the configured automatic workflow. 

Maybe I'm missing a use case... 


[EDIT] use case I'm missing: external event changing something on the record which makes it match the configured domain to trigger the workflow. But this could be handled by an automated action.

sale_automatic_workflow_job/models/automatic_workflow_job.py
simahawk: @vvrossem FTR using f strings will break translations because the string will be evaluated at run time making the string always different. Ideally we should have named placeholders like `_("Foo %(baz)s", baz=self.baz)`, but is not a blocker.

simahawk:
> I'm wondering why we are keeping the cron jobs in this module?
> 
> This seems so inefficient, when we could have the queue jobs created on creation of the various records (sale, picking, invoice) based on the configured automatic workflow.
> 
> Maybe I'm missing a use case...
> 
> [EDIT] use case I'm missing: external event changing something on the record which makes it match the configured domain to trigger the workflow. But this could be handled by an automated action.

I can give you another use case: having a delay between the creation of the SO and the automatic approval (real customer use case).

IMO it's better to keep the cron because it also helps not to spawn useless jobs and to not add too much overhead on save or confirm or whatever action you want to hook to.

StefanRijnhart:
Agreed. With the cron queuing the jobs, you can plan the load on your system.

sale_automatic_workflow_job/models/automatic_workflow_job.py
StefanRijnhart: Other reviewers suggested to use a positional argument if there is only one (as is the case here), and switch to named arguments in case of multiple. I like that one myself, but maybe that's all that it is, a preference?

chaule97:
Ciao @simahawk, can you review and merge again? I have found that module `queue_job` changed parameter `_job_force_sync` to `queue_job__no_delay`

PR #3465, opened 2024-12-06
[17.0][IMP] sale_discount_display_amount: improve layout sale discount total

xaviedoanhduy:
There are some interface changes on the Sales Order form view that cause the total summary to be a bit broken.

before:
image

after:
image

### DIFF
--- a/sale_discount_display_amount/views/sale_view.xml
+++ b/sale_discount_display_amount/views/sale_view.xml
@@ -19,37 +19,35 @@
         <field name="priority">99</field>
         <field name="arch" type="xml">
             <data>
-                <xpath expr="//field[@name='tax_totals']" position="after">
-                    <field
-                        name="display_discount_with_tax"
-                        invisible="1"
-                        position="attributes"
+                <xpath expr="//group[@name='sale_total']" position="after">
+                    <group
+                        class="oe_subtotal_footer"
+                        colspan="2"
+                        name="sale_discount_total"
                     >
-                    </field>
-                    <field
-                        name="price_subtotal_no_discount"
-                        invisible="display_discount_with_tax == True"
-                        position="attributes"
-                    >
-                    </field>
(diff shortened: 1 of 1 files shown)


rousseldenis:
@xaviedoanhduy Thanks for this. Could you check as there are some glitches on runboat:

image

xaviedoanhduy:
hi @rousseldenis, `sale_order_general_discount` has class `oe_subtotal_footer_separator` will create a horizontal bar above the discount value. (see: https://github.com/OCA/sale-workflow/blob/a0a79cbbad3545d290212444b0383e37400e5255/sale_order_general_discount/views/sale_order_view.xml#L11)
image

LucasTran380381:
> @xaviedoanhduy Thanks for this. Could you check as there are some glitches on runboat:
> 
> image

image

that show in my local machine. no problem at all

xaviedoanhduy:
about this, `base_tier_validation` will make them `readonly=False`, that makes the currency unit space far enough apart that we can edit those 2 values, do you think I should take this opportunity to set 2 `readonly=True` values ​​on the view form for those 2 fields (see: https://github.com/OCA/server-ux/blob/1be5f5af82105d0ba25bbe494ca26ccd313a8f2c/base_tier_validation/models/tier_validation.py#L811)
image

result: when remove `oe_subtotal_footer` class on `general_discount` field and add `readonly=True` on them.
image

rousseldenis:
@xaviedoanhduy There is the tax totals component. Why not extending it to comply with core features?

xaviedoanhduy:
> @xaviedoanhduy There is the tax totals component. Why not extending it to comply with core features?

tax totals component (from account module) is used in account, purchase and sales. i think there should be 1 account related module to do that, and for now to simplify i just improve a little on view form of sale order. but you are right extending that component will make the layout better.

PR #3466, opened 2024-12-06
  [18.0][MIG] sales_team_security_sale: Migration to 18.0

lav-adhoc:


bo-eska:
Hi, when user's team members are salesperson, sales orders are visible but linked invoices do not open with just these `ir.rules`. This rules solves the problem; 
```
<record id="account_move_team_rule" model="ir.rule">
        <field name="name">Team Invoices</field>
        <field
            name="domain_force"
        >['|', ('invoice_user_id', 'in', [user.id, False]), ('team_id', 'in', user.crm_team_ids.ids + [False])]
        </field>
        <field
            name="groups"
            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
        />
    </record>
    <record id="account_move_line_team_rule" model="ir.rule">
        <field name="name">Team Invoices Lines</field>
        <field name="model_id" ref="account.model_account_move_line" />
        <field
            name="domain_force"
        >['|', ('move_id.invoice_user_id', 'in', [user.id, False]), ('move_id.team_id', 'in', user.crm_team_ids.ids + [False])]</field>
        <field
            name="groups"
            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
        />
    </record>
```

alexey-pelykh (commented):
Thanks for the migration.

The security rules and manifest look fine for 18.0 -- `salesman_id`, `user_id`, and `team_id` are all still present on their respective models.

One thing I noticed: the 17.0 version had a `test_sale_order_permissions_multi_team` test that verified permissions when a user belongs to multiple teams. That test (along with the `team3` setup and the `ir.config_parameter` for `sales_team.membership_multi`) was removed in this PR. Dropping the config parameter call makes sense since multi-team membership is default in 18.0, but the test itself was still exercising a valid scenario (user in team1 sees an order assigned to team3). Could you bring back a version of that test without the config parameter? It would help maintain the coverage we had before.

Also a minor nit: the `.pot` file still references `Odoo Server 15.0` in the `Project-Id-Version` header. Not blocking, but worth regenerating.

rousseldenis (approved):

rousseldenis:

PR #3467, opened 2024-12-06
[14.0][IMP] sale_order_product_assortment: performance in so

HekkiMelody:
This commit improves performance when computing the products allowed in a Sales Order, especially when creating an order and the partner is empty

### DIFF
--- a/sale_order_product_assortment/models/sale_order.py
+++ b/sale_order_product_assortment/models/sale_order.py
@@ -19,61 +19,72 @@ class SaleOrder(models.Model):
 
     @api.depends("partner_id", "partner_shipping_id", "partner_invoice_id")
     def _compute_product_assortment_ids(self):
-        self.allowed_product_ids = False
-        self.has_allowed_products = False
         partner_field = (
             self.env["ir.config_parameter"]
             .sudo()
             .get_param("sale_order_product_assortment.partner_field", "partner_id")
         )
-        product_domain = []
-        if self[partner_field]:
-            filters_partner_domain = self.env["ir.filters"].search(
-                [("is_assortment", "=", True)]
-            )
-            filters_partner_domain = filters_partner_domain.filtered(
-                lambda x: x.all_partner_ids & self[partner_field]
-            )
-            if filters_partner_domain:
-                # use OR to combine allowed products
-                # and negate the blacklisted ones with AND
-                eval_product_domain = expression.OR(
-                    safe_eval(
-                        domain,
-                        {"datetime": datetime, "context_today": datetime.datetime.now},
+        filters_partner_domain = None
+        for so in self:
+            products = False
+            partner = so[partner_field]
+            if partner:
+                if filters_partner_domain is None:
+                    # search for filters only once
+                    filters_partner_domain = self.env["ir.filters"].search(
+                        [("is_assortment", "=", True)]
                     )
-                    for domain in filters_partner_domain.mapped("domain")
+                filters_partner_domain = filters_partner_domain.filtered(
+                    lambda x, partner=partner: partner in x.partner_ids
+                    # handle empty domain [] as False
+                    or partner.filtered_domain(
+                        x._get_eval_partner_domain() or expression.FALSE_DOMAIN
+                    )
                 )
-                whitelist_product_domain = expression.OR(
-                    [
-                        eval_product_domain,
+                if filters_partner_domain:
+                    # use OR to combine allowed products
+                    # and negate the blacklisted ones with AND
+                    eval_product_domain = expression.OR(
+                        safe_eval(
+                            domain,
+                            {
+                                "datetime": datetime,
+                                "context_today": datetime.datetime.now,
+                            },
+                        )
+                        for domain in filters_partner_domain.mapped("domain")
+                    )
+                    whitelist_product_domain = expression.OR(
                         [
-                            (
-                                "id",
-                                "in",
-                                filters_partner_domain.mapped(
-                                    "whitelist_product_ids.id"
-                                ),
-                            )
-                        ],
-                    ]
-                )
-                blacklist_product_domain = [
-                    (
-                        "id",
-                        "not in",
-                        filters_partner_domain.mapped("blacklist_product_ids.id"),
+                            eval_product_domain,
+                            [
+                                (
+                                    "id",
+                                    "in",
+                                    filters_partner_domain.mapped(
+                                        "whitelist_product_ids.id"
+                                    ),
+                                )
+                            ],
+                        ]
                     )
-                ]
-                product_domain = expression.AND(
-                    [whitelist_product_domain, blacklist_product_domain]
-                )
-            if product_domain:
-                self.allowed_product_ids = self.env["product.product"].search(
-                    product_domain
-                )
-            if self.allowed_product_ids:
-                self.has_allowed_products = True
+                    blacklist_product_domain = [
+                        (
+                            "id",
+                            "not in",
+                            filters_partner_domain.mapped("blacklist_product_ids.id"),
+                        )
+                    ]
+                    product_domain = expression.AND(
+                        [whitelist_product_domain, blacklist_product_domain]
+                    )
+                    products = self.env["product.product"].search(product_domain)
+            so.update(
+                {
+                    "allowed_product_ids": products,
+                    "has_allowed_products": bool(products),
+                }
+            )
 
     @api.onchange("partner_id", "partner_shipping_id", "partner_invoice_id")
     def _onchange_partner_id(self):


CarlosRoca13:
Hi @aleuffre, first of all thanks for the changes :smile: can you show a comparison of the times between the current version and after adding these changes?

ping @carlosdauden

HekkiMelody:
Thank you for checking the PR @CarlosRoca13 !

I tried using `timeit` but it wouldn't behave for some reason, so, apologies for the "homemade" solution in Odoo shell.

This is tested on a production database that is having some performance issues, that's why we're doing a performance pass on a handful of modules

**Before:**

```
>>> from datetime import datetime as dt
>>> so = env["sale.order"].search([], limit=1)
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
3.842279 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.563739 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.670987 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.658426 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.532004 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.629248 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.555761 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.684823 seconds
```

**After**:

```
>>> from datetime import datetime as dt
>>> so = env["sale.order"].search([], limit=1)
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
3.460797 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.454272 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.428768 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.399137 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.413592 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.444747 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.380238 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.397538 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.412193 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.373276 seconds
>>> start = dt.now(); so._compute_product_assortment_ids(); print((dt.now() - start).total_seconds(), "seconds")
0.430472 seconds
```

To add to this, when pressing "create" from the interface, and the partner is still empty, the new function checks that the partner is not set and is not executed, whereas the old function still performs the full search.

CarlosRoca13 (approved):
It seems to be ok :smile: 

can you review too @carlosdauden @pedrobaeza

francesco-ooops (approved):
LGTM

HekkiMelody:
Sorry, I made a new change, specifically

```py
                filters_partner_domain = filters_partner_domain.filtered(
                    lambda x, partner=partner: partner in x.partner_ids
                    or partner.filtered_domain(x._get_eval_partner_domain())
                )
```

This lets us skip computing `all_partner_ids` and the first invocation of the method goes down from 3.8 seconds to 1.8 seconds on my DB.
Also did some cleanup on the onchange below, though the impact in terms of performance is almost negligible.

HekkiMelody:
I need to fix the tests.

carlosdauden (changes requested):
This version does not have the changes applied properly, this backport was made, but it was not quite right:
https://github.com/OCA/sale-workflow/pull/3135
as in addition to these changes there were others that did not carry over.
Just take a look at the file in 15.0:
https://github.com/OCA/sale-workflow/blob/ad9d3f630683f6ba32ec29270b484f9c1255b2ad/sale_order_product_assortment/models/sale_order.py#L18-L38
Test the performance with these changes by simply copying and pasting.

carlosdauden:
You need changes in product_assortment which I see are not at 14.0 either:
https://github.com/OCA/product-attribute/blob/15.0/product_assortment/models/ir_filters.py#L62-L71

HekkiMelody:
> This version does not have the changes applied properly, this backport was made, but it was not quite right: #3135 as in addition to these changes there were others that did not carry over. Just take a look at the file in 15.0:
> 
> https://github.com/OCA/sale-workflow/blob/ad9d3f630683f6ba32ec29270b484f9c1255b2ad/sale_order_product_assortment/models/sale_order.py#L18-L38
> 
> Test the performance with these changes by simply copying and pasting.

I have vague memories of this PR. I think the problem, as you pointed out, was that there were changes not ported in another module in another repo, and due to the fact it can take a while to get PRs merged and we were in a hurry, it was decided to adapt the changes to the module as is.

Having to compute `all_partner_ids` just to see if a single partner is in a recordset of hundreds, if not thousands, every time, is really heavy, as you can imagine.

I'll check the PRs you mentioned.

Thank you for your time!

carlosdauden (approved):
Ok.
I'm going to approve the changes anyway in case you want to merge this, as I understand it works better than what is currently there and I don't want to block in this version either.

HekkiMelody:
I still need to fix the tests, but thank you

francesco-ooops:
@CarlosRoca13 tests fixed, can this be merged? thanks!

HekkiMelody:
@carlosdauden Hello, sorry to bother you, but since you worked on the performance of this module, I figured it wouldn't hurt to ask for a suggestion.

Me and @francesco-ooops were running some more performance tests and while this PR helps a lot when the method in question is called (for example, when the SO is opened for the first time, or when the partner is changed), a lot of other operations (such as adding a SOL) are still very slow. And notably, no code in this module should really affect them. Removing this module makes SOs usable, whereas with this module (even with this PR) SOs are painfully slow, we're talking 10+ seconds to add a SOL and choose a product.

After a few tests, commenting parts of the view etc, I think I found the culprit. The field `allow_product_ids` in our case can contain over 35k records. When calling the onchange on the sale order line, the client sends the list of 35k ids to the server (that's ~~250kb of data!), and I assume the server reads it somehow because the onchange itself is painfully slow (over 5 seconds) even if the field `allow_product_ids` is not really used.

Commenting this part of the view

```xml
                <attribute
                    name="domain"
                    operation="domain_add"
                    condition="parent.has_allowed_products"
                >[('id', 'in', parent.allowed_product_ids)]</attribute>
```
didn't help at all. Only when I also commented

```xml
                <field name="allowed_product_ids" invisible="1" />
```

I saw a noticeable performance gain.

Do you have any suggestions on how to proceed? Ideally I think I'd like to add the domain that is now used for the `.search()` in the compute directly to the view...

carlosdauden:
Are you tested this in v15.0?

HekkiMelody:
> Are you tested this in v15.0?

Not yet, but I will

sale_order_product_assortment/models/sale_order.py
```diff
@@ -19,61 +19,68 @@ class SaleOrder(models.Model):
...
-                    [whitelist_product_domain, blacklist_product_domain]
-                )
-            if product_domain:
-                self.allowed_product_ids = self.env["product.product"].search(
-                    product_domain
-                )
-            if self.allowed_product_ids:
-                self.has_allowed_products = True
+                    products = self.env["product.product"].search(product_domain)
+            so.allowed_product_ids = products
```
rousseldenis:
```suggestion
            so..update({
                "allowed_product_ids": products,
                "has_allowed_products": bool(products)
            })
```

sale_order_product_assortment/models/sale_order.py
```diff
@@ -19,61 +19,68 @@ class SaleOrder(models.Model):
...
-                        domain,
-                        {"datetime": datetime, "context_today": datetime.datetime.now},
+        filters_partner_domain = None
+        for so in self:
+            products = False
+            partner = so[partner_field]
+            if partner:
+                if filters_partner_domain is None:
+                    # search for filters only once
+                    filters_partner_domain = self.env["ir.filters"].search(
```
rousseldenis, line 34: Is the case where partner field is void common? Either, put the search once before the loop.

sale_order_product_assortment/models/sale_order.py
```diff
@@ -19,61 +19,68 @@ class SaleOrder(models.Model):
...
-                product_domain = expression.AND(
-                    [whitelist_product_domain, blacklist_product_domain]
-                )
-            if product_domain:
-                self.allowed_product_ids = self.env["product.product"].search(
-                    product_domain
-                )
-            if self.allowed_product_ids:
-                self.has_allowed_products = True
+                    products = self.env["product.product"].search(product_domain)
```
rousseldenis, line 81: I would say, doing a search once before the loop should be the aim to get to.

IMHO, it is possible to do it, then filtering the products

HekkiMelody:
> Are you tested this in v15.0?

I have tested it. From a clean DB with `sale_order_product_assortment`
1. Add 35k products (`for i in range(35000): env["product.product"].create({"name": i})`)
2. Create an assortment that includes all products, all partners (I used the domain `("id", ">", 0)`)
3. Create a SO

Odoo is noticeably slow when selecting the partner, and when adding each Sales order line. Of course it's not quite as slow as a full production database with several other modules installed, but you can see the problem appearing there already.

sale_order_product_assortment/models/sale_order.py
HekkiMelody: There's at least one execution of this method (when the SO is just created from the interface, and all fields are empty) where the partner is empty

sale_order_product_assortment/models/sale_order.py
HekkiMelody: The vast, vast majority of the times (every time?) this method is executed with only one SO in `self`.

Either way, how would you suggest to go about it? We have over 50k products, and filtering them in python afterwards seems like it would slow things down significantly.

sale_order_product_assortment/models/sale_order.py
HekkiMelody: Changed as suggested, thanks

HekkiMelody:
> > Are you tested this in v15.0?
> 
> I have tested it. From a clean DB with `sale_order_product_assortment`
> 
>     1. Add 35k products (`for i in range(35000): env["product.product"].create({"name": i})`)
> 
>     2. Create an assortment that includes all products, all partners (I used the domain `("id", ">", 0)`)
> 
>     3. Create a SO
> 
> 
> Odoo is noticeably slow when selecting the partner, and when adding each Sales order line. Of course it's not quite as slow as a full production database with several other modules installed, but you can see the problem appearing there already.

I think the best solution would be to extend base_view_inheritance_extension in order to add an operation similar to `domain_add` but which would allow me to point at a binary field holding the domain I want to add. I don't know how feasible it would be.

That way, instead of adding `[('id', 'in', [list of ids])]` I could compute and return the domain computed from the assortments, without running the `search()` directly in the backend, and then combine that domain with the existing domain on `sale.order.line.product_id`

Does that make sense?

carlosdauden:
Yes, it makes sense

PR #3469, opened 2024-12-06
[18.0][MIG] sales_team_security_crm: Migration to 18.0

lav-adhoc:


alexey-pelykh (commented):
Thanks for the migration.

The security rules and manifest look good -- `crm.activity.report` and the record rule domains are still valid on 18.0.

A couple of things I noticed:

**`.pot` file -- stale version reference**

The `Project-Id-Version` in `i18n/sales_team_security_crm.pot` says `Odoo Server 15.0`. The 17.0 branch had this correctly set to `17.0`. Should be `Odoo Server 18.0` here (or it will be regenerated -- just flagging it).

**Dropped multi-team test**

The 17.0 version had `test_crm_lead_permissions_multi_team` which tested that a user belonging to multiple teams (via `crm.team.member`) could access leads from any of those teams. That test and the associated `team3` setup were removed in this migration. Since `sales_team.membership_multi` still exists in 18.0, this seems like unintentional coverage loss rather than a deliberate simplification. Could you restore it?

PR #3473, opened 2024-12-09
[17.0][MIG] sale_invoice_no_mail

alexeirivera87:


pandoc-3.6-1-amd64.deb
pedrobaeza, line 1: This file shouldn't be here

alexeirivera87: Hi @pedrobaeza 

Copied, I removed it. I think it was autogenerated during pre-commit.

regards

pedrobaeza:
Thanks. Now you need to check the CI.

alexeirivera87:
Hi @pedrobaeza 

It's done. Thanks

pedrobaeza:
OK, now reviewers can check.

/ocabot migration sale_invoice_no_mail

sale_invoice_no_mail/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "sale.advance.payment.inv"
+
+    def create_invoices(self):
+        skip = self._skip_invoice_mail()
+        wiz = self.with_context(mail_auto_subscribe_no_notify=skip)
+        return super(SaleAdvancePaymentInv, wiz).create_invoices()
+
+    def _skip_invoice_mail(self):
+        """Hook method, can be overridden. Returns True by default"""
+        return True
```
rousseldenis, line 17: IMHO this should come from a configuration. The default standard behavior is changed with this

alexey-pelykh (commented):
Thanks for the migration @alexeirivera87.

I looked through the diff and compared it against the 15.0 source. The wizard and test code are identical to 15.0, which is understandable given the module's simplicity. A few observations:

1. **`.pot` file still references Odoo 15.0** — `"Project-Id-Version: Odoo Server 15.0\n"` in `i18n/sale_invoice_no_mail.pot` should be regenerated for 17.0.

2. **Wizard flow in 17.0** — In Odoo 17, the `sale.advance.payment.inv` wizard gained its own `_create_invoices()` method that wraps the call to `sale_orders._create_invoices()`. The current approach of setting `mail_auto_subscribe_no_notify` on the wizard context before calling `super().create_invoices()` still works because the context propagates through the environment, but it is worth noting that this context key only suppresses the auto-subscribe notification — it does not suppress other mail mechanisms (e.g. `message_post` calls). For the "delivered" invoice path this is fine since `account.move.create()` is where the auto-subscribe triggers, and the context reaches it through the ORM chain.

3. I agree with @rousseldenis's point above — installing this module unconditionally changes the standard behavior for all users. A boolean on the company or in settings (with `True` as default to preserve current module behavior) would make this more flexible and easier to adopt. That said, this was already the design in 15.0, so this is a pre-existing consideration rather than a migration blocker.

The code itself is clean, the test covers the happy path, and the manifest/readme are properly updated for 17.0. The `.pot` update is the only concrete fix needed.

rousseldenis:
> Thanks for the migration @alexeirivera87.
> 
> I looked through the diff and compared it against the 15.0 source. The wizard and test code are identical to 15.0, which is understandable given the module's simplicity. A few observations:
> 
> 1. **`.pot` file still references Odoo 15.0** — `"Project-Id-Version: Odoo Server 15.0\n"` in `i18n/sale_invoice_no_mail.pot` should be regenerated for 17.0.
> 2. **Wizard flow in 17.0** — In Odoo 17, the `sale.advance.payment.inv` wizard gained its own `_create_invoices()` method that wraps the call to `sale_orders._create_invoices()`. The current approach of setting `mail_auto_subscribe_no_notify` on the wizard context before calling `super().create_invoices()` still works because the context propagates through the environment, but it is worth noting that this context key only suppresses the auto-subscribe notification — it does not suppress other mail mechanisms (e.g. `message_post` calls). For the "delivered" invoice path this is fine since `account.move.create()` is where the auto-subscribe triggers, and the context reaches it through the ORM chain.
> 3. I agree with @rousseldenis's point above — installing this module unconditionally changes the standard behavior for all users. A boolean on the company or in settings (with `True` as default to preserve current module behavior) would make this more flexible and easier to adopt. That said, this was already the design in 15.0, so this is a pre-existing consideration rather than a migration blocker.
> 
> The code itself is clean, the test covers the happy path, and the manifest/readme are properly updated for 17.0. The `.pot` update is the only concrete fix needed.

@alexey-pelykh  Do you still do this with AI ? @alexeirivera87 seems no more active on github... Closing this as very old.

PR #3477, opened 2024-12-10
[IMP] sale_manual_delivery: Improve init script performances

### LINKED ISSUE #3476: [16.0] sale_manual_delivery: Install on large database is too long
The pre init script is too long.

IMHO:

- The creation of `qty_procured`field should be initialized with `0` default value there:

https://github.com/OCA/sale-workflow/blob/2a2908b25b20495b28112c29bef34d8820fe4760/sale_manual_delivery/hook.py#L4

- The filling in for that value should be done for lines that have their `qty_delivered_method` to `stock_move`only. See compute method:

https://github.com/OCA/sale-workflow/blob/2a2908b25b20495b28112c29bef34d8820fe4760/sale_manual_delivery/models/sale_order_line.py#L41

@ChrisOForgeFlow @nguyenhk @lmignon

rousseldenis:


rousseldenis:
@lmignon

lmignon (approved):
LGTM (Code review only)

leemannd (approved):

bosd:
Merge? :eyes: :pray:

alexey-pelykh (approved):
Thanks @rousseldenis, nice optimization.

Splitting the init into a fast path for non-stock_move lines and narrowing the heavy join to only `qty_delivered_method = 'stock_move' AND state = 'sale'` should make a real difference on databases with large sale_order_line tables.

Column type switch to `double precision` is correct too -- these are Odoo Float fields, `numeric` was unnecessary overhead.

LGTM

WesleyOliveira98 (approved):
LGTM

PR #3480, opened 2024-12-11
[15.0][FIX] sale_order_type: sale order warehouse changed when the user of a sale order is changed

sergio-teruel:
[15.0][FIX] sale_order_type: sale order warehouse changed when the user of a sale order is changed
when a default warehouse is defined in order type

cc @Tecnativa TT52158

### DIFF
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -71,6 +71,14 @@ class SaleOrder(models.Model):
             vals = {}
             if order_type.warehouse_id:
                 vals.update({"warehouse_id": order_type.warehouse_id})
+            else:
+                vals.update(
+                    {
+                        "warehouse_id": self.user_id.with_company(self.company_id.id)
+                        ._get_default_warehouse_id()
+                        .id
+                    }
+                )
             if order_type.picking_policy:
                 vals.update({"picking_policy": order_type.picking_policy})
             if order_type.payment_term_id:
@@ -98,6 +106,17 @@ class SaleOrder(models.Model):
                 line_vals.update({"route_id": order_type.route_id.id})
             if line_vals:
                 order.order_line.update(line_vals)
+
+    @api.onchange("user_id")
+    def onchange_user_id(self):
+        res = super().onchange_user_id()
+        if (
+            self.state in ["draft", "sent"]
+            and not self.user_id.property_warehouse_id
+            and self.type_id.warehouse_id
+        ):
+            self.warehouse_id = self.type_id.warehouse_id
+        return res
 
     @api.model
     def create(self, vals):


sale_order_type/models/sale.py
```diff
@@ -71,6 +71,14 @@ def onchange_type_id(self):
             vals = {}
             if order_type.warehouse_id:
                 vals.update({"warehouse_id": order_type.warehouse_id})
+            else:
+                vals.update(
```
pedrobaeza, line 75: I don't get why don't you let the previous one here.

carlosdauden (approved):

sale_order_type/models/sale.py
```diff
@@ -99,6 +107,17 @@ def onchange_type_id(self):
             if line_vals:
                 order.order_line.update(line_vals)
 
+    @api.onchange("user_id")
+    def onchange_user_id(self):
+        res = super().onchange_user_id()
+        if (
+            self.state in ["draft", "sent"]
+            and not self.user_id.property_warehouse_id
+            and self.type_id.warehouse_id
```
chienandalu, line 116: Not very sure about this. Should user preference have precedence over sale type options?

carlosdauden: Note that user_id is equivalent to salesman.

The option for a salesman to have a warehouse assigned to him is very very specific.
Normally salesmen don't have any assigned to them, but if they do this option should override the warehouse assigned to the order type.

Maybe you can find a case where you want a different behaviour, but at the moment the problem reported to us is this.

Anyway these are still default values that can be changed by the user.

pedrobaeza: But this is not doing that, but overwriting whatever is already assigned, and put the picking type warehouse if not warehouse defined in the salesman (which will be mostly always). Moreover, this is an onchange, and not a computed method, so I think there's a mix of both around.

chienandalu: I can think in both scenarios :sweat_smile: or even in mixed ones :smile: And it seems that the logic is inverted from the one defined avobe where the sale type warehouse has the precedence...

carlosdauden: Well, the problem is the above, can you make a better code suggestion to solve it?

chienandalu: For me the sane option would be that the sale type warehouse had the precedence as it happens when the order is created, otherwise the user's default.

carlosdauden: I've tried the sane option, but it doesn't solve the problem... :wink: 
Any other suggestions?

I repeat the objective:
The warehouse of the salesperson (if established) is much more specific than the order type warehouse.

sale_order_type/models/sale.py
rousseldenis: @sergio-teruel I think this comment is still applicable. Why isn't this sufficent: https://github.com/odoo/odoo/blob/15.0/addons/sale_stock/models/sale_order.py#L39

rousseldenis:

PR #3481, opened 2024-12-11
[16.0][IMP] sale_stock_prebook: Specific picking type for booking

lmignon:
* [ ] depends on #3423


lmignon:
moved to https://github.com/OCA/sale-prebook/pull/3

PR #3072, opened 2024-04-12
[17.0][MIG] sale_substate

Wodran14:
Depends on:

- [https://github.com/OCA/server-ux/pull/855]

PR #3483, opened 2024-12-12
[17.0][FIX] sale_substate: fix inheritance order to restore tracking

len-foss:
With the current order, the super to _track_template from the mixin was never called. Which was a good thing since it was also buggy;
however the override in this model was removed on the assumption that it was.

Depends on https://github.com/OCA/server-ux/pull/982 since the super function is actually borked.

(it's therefore expected tests crash before the dependencies are updated)

Also, the problem is present in other substate modules like `purchase_substate` but their migration is in limbo so it's all good.

### DIFF
--- a/sale_substate/models/sale_order.py
+++ b/sale_substate/models/sale_order.py
@@ -13,5 +13,5 @@ class BaseSubstateType(models.Model):
 
 
 class SaleOrder(models.Model):
-    _inherit = ["sale.order", "base.substate.mixin"]
+    _inherit = ["base.substate.mixin", "sale.order"]  # order matters
     _name = "sale.order"
--- a/sale_substate/tests/test_sale_substate.py
+++ b/sale_substate/tests/test_sale_substate.py
@@ -61,6 +61,15 @@ class TestBaseSubstate(TransactionCase):
         self.assertTrue(so_test1.state == "sale")
         self.assertTrue(so_test1.substate_id == self.substate_valid_docs)
 
+        # manually change the substate; this one has a mail template
+        self.env.cr.flush()  # this is to post the messages (it's in after commit hook)
+        messages = so_test1.message_ids
+        so_test1.write({"substate_id": self.substate_in_delivery.id})
+        self.env.cr.flush()  # this is to post the messages (again)
+        new_messages = so_test1.message_ids - messages
+        expected_msg = "You order is being prepared for delivery."
+        self.assertTrue(expected_msg in "".join(new_messages.mapped("body")))
+
         # Test that substate_id is set to false if
         # there is not substate corresponding to state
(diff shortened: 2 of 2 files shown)


len-foss:
Hello @rousseldenis 
This fixes comment https://github.com/OCA/sale-workflow/pull/3072#discussion_r1565386377

StefanRijnhart (commented):
You can fix the tests by adding your version of base_substate in test-requirements.txt

Something like
```
git+https://github.com/lambdao-dev/server-ux@17.0-fxsbsttmx-len#subdirectory=base_substate
```
Would it be possible to add a test to validate that tracking has taken place, so that it is signalled when this goes wrong again after some future Odoo internal refactoring or so.

len-foss:
Working test added, it should be the same for other repair_substate or purchase_substate modules.

sale_substate/models/sale_order.py
```diff
@@ -13,5 +13,5 @@ class BaseSubstateType(models.Model):
 
 
 class SaleOrder(models.Model):
-    _inherit = ["sale.order", "base.substate.mixin"]
+    _inherit = ["base.substate.mixin", "sale.order"]  # order matters!!!
```
StefanRijnhart: I understand that finding out about something as subtle as this deserves the appropriate amount of exclamation marks in the heat of the moment, but for future generations of developers who will maintain this code the message will still be clear without them :smile: 

```suggestion
    _inherit = ["base.substate.mixin", "sale.order"]  # order matters
```

StefanRijnhart (commented):
Great work on the test! One more suggestion.

sale_substate/models/sale_order.py
len-foss: I think that in previous python versions, you would get the function non-deterministically, unless there are other conditions that dictate which function is loaded in which order. But it's more that it's not the first time and people seem quite unaware of the problem (if you want to do worse, do that with product.product and product.template!)

I'll do an absorb rebase when removing the temporary dependency, change is in a fixup commit.

StefanRijnhart (approved):

astirpe (approved):

StefanRijnhart:
Please drop the test requirement.

alexey-pelykh (commented):
Thanks for tracking this down, the inheritance order issue is a subtle one.

The fix and the test both look good to me. Now that OCA/server-ux#982 is merged, the temporary `test-requirements.txt` entry for the fork should be dropped (as Stefan mentioned).

len-foss:
@alexey-pelykh done, all green :+1:

StefanRijnhart:
@len-foss This shouldn't have needed the bot spam to get unstuck 😄 

/ocabot merge patch

PR #3484, opened 2024-12-12
[14.0][IMP]product_supplierinfo_for_customer_sale: Search SO by customerinfo

toita86:
Add a new search field that retrieves all the SOs that contains the product with the given `product_name` or `product_code`.

### DIFF
--- a/product_supplierinfo_for_customer_sale/models/sale_order_line.py
+++ b/product_supplierinfo_for_customer_sale/models/sale_order_line.py
@@ -12,20 +12,53 @@ class SaleOrderLine(models.Model):
     product_customer_code = fields.Char(
         compute="_compute_product_customer_code",
         string="Product Customer Code",
+        search="_search_product_customer_code",
+    )
+    product_customer_name = fields.Char(
+        compute="_compute_product_customer_code",
+        string="Product Customer Name",
+        search="_search_product_customer_name",
     )
 
-    @api.depends("product_id")
+    @api.depends("order_id.partner_id", "product_id", "company_id")
     def _compute_product_customer_code(self):
         for line in self:
-            if line.product_id:
-                supplierinfo = line.product_id._select_customerinfo(
+            product_customer_name = product_customer_code = False
+
+            if line.product_id and line.order_id and line.order_id.partner_id:
+                customerinfo = line.product_id._select_customerinfo(
                     partner=line.order_partner_id,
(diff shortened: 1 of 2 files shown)


francesco-ooops (approved):
LGTM

Use case: if product "Acoustic bloc screens" is sold to customer with customer name "Megablock", customer can ask you to retrieve an order containing the product "Megablock" => user needs to be able to perform search using customer name and/or customer code

With this PR we add a search field with the same properties as "Product" field in SO list view, searching for customer name and code

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@pedrobaeza good to go? :)

product_supplierinfo_for_customer_sale/models/sale_order_line.py
```diff
@@ -11,21 +11,30 @@ class SaleOrderLine(models.Model):
 
     product_customer_code = fields.Char(
         compute="_compute_product_customer_code",
+        store=True,
```
pedrobaeza: This is a no-go. You can't convert a field to stored on a table that can be populated with a lot of records.

pedrobaeza (changes requested):

product_supplierinfo_for_customer_sale/models/sale_order_line.py
toita86: Thank you for the review, while I get your point I will like to ask how do you would you implement such a feature.

My approach derives by a similar implementation found here https://github.com/OCA/stock-logistics-workflow/blob/14.0/product_supplierinfo_for_customer_picking/models/stock_move.py .
The situation is very similar so I thought that was a possible solution. 
For backward compatibility I can also make a migration script for the `store`  attribute of the field.

But I will like to have you opinion first.

HekkiMelody: One possibility is to not store the field, and implement the `search` attribute

toita86: @aleuffre Thank you for your suggestion. I have implemented the changes based on your feedback.

francesco-ooops (approved):
Functional still ok!

@pedrobaeza code is good now?

pedrobaeza (approved):
/ocabot merge minor

PR #3486, opened 2024-12-13
[18.0][MIG] product_form_sale_link: Migration to 18.0

BhaveshHeliconia:


JessBrandl (approved):
LGTM, tested locally and worked as intended

product_form_sale_link/models/product_template.py
```diff
@@ -11,7 +11,7 @@
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
-    sale_lines_count = fields.Float(compute="_compute_sale_lines_count", string="Sales")
```
CRogos: Is there a reason why you removed the string attribute and added it to the xml?

CRogos (commented):
Squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

image

BhaveshHeliconia:
@CRogos, Please Review.Thanks for suggestions!

BhaveshHeliconia:
@CRogos, The administrative commits have been squashed. Test cases in the pipeline are failing due to the **sales_line_count** field. This issue needs to be addressed by applying the fix in the XML configuration.
Screenshot from 2025-03-19 12-11-33

CRogos (approved):

CRogos:
@pedrobaeza could you merge?

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,172 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        # Create test users with different access rights
+        sale_manager_group = cls.env.ref("sales_team.group_sale_manager")
+        sale_user_group = cls.env.ref("sales_team.group_sale_salesman")
+        product_manager_group = cls.env.ref("product.group_product_manager")
+
+        cls.user_salesman = cls.env["res.users"].create(
```
pedrobaeza: Use `new_test_user` and `@users` from odoo.tests.common instead.

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,172 @@
...
+class TestProductSalesCount(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        # Create test users with different access rights
+        sale_manager_group = cls.env.ref("sales_team.group_sale_manager")
+        sale_user_group = cls.env.ref("sales_team.group_sale_salesman")
+        product_manager_group = cls.env.ref("product.group_product_manager")
+
```
pedrobaeza: Don't leave empty lines inside the method. They are not adding value.

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,172 @@
...
+                        [
+                            cls.env.ref("base.group_user").id,
+                            product_manager_group.id,
+                        ],
+                    )
+                ],
+            }
+        )
+
+        cls.currency = cls.env["res.currency"].create(
```
pedrobaeza: Don't create all this useless stuff. You don't care about the sales orders amounts. You only care about the number.

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,172 @@
...
+        cls.product_template = cls.env["product.template"].create(
+            {
+                "name": "Test Product Template",
+                "list_price": 100.0,
+                "taxes_id": False,
+                "currency_id": cls.currency.id,
+            }
+        )
+
+        cls.size_attribute = cls.env["product.attribute"].create(
```
pedrobaeza: Don't overcomplicate the test case. You don't need attributes for this case.

pedrobaeza (changes requested):
It seems the IA is generating the tests cases, bloating them a lot, so please don't use it.

BhaveshHeliconia:
@pedrobaeza, Thank you for the help! The issue is resolved now. I'll take care of it from here on out.Please Review.

product_form_sale_link/tests/test_product.py
pedrobaeza: I insist this is not needed.

pedrobaeza: I mean the attribute/values creation. You can do the test case without creating a muti-variant product.

CRogos: @BhaveshHeliconia could you cleanup the test as from pedro suggested?

BhaveshHeliconia: @CRogos, @pedrobaeza, I updated test cases Please Review.

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,75 @@
...
+            {
+                "name": "Test Product 2",
+            }
+        )
+        cls.product_template = cls.product1.product_tmpl_id
+        cls.partner = cls.env["res.partner"].create(
+            {
+                "name": "Test Partner",
+            }
+        )
```
pedrobaeza: Remove the trailing comma for not having so many useless lines (the same for all).

```suggestion
        cls.partner = cls.env["res.partner"].create(
            {"name": "Test Partner"}
        )
```

product_form_sale_link/tests/test_product.py
```diff
@@ -0,0 +1,75 @@
...
+            {
+                "partner_id": cls.partner.id,
+            }
+        )
+
+    @users("test_salesman")
+    def test_product_variant_sales_count_with_salesman(self):
+        self.sale_order.action_confirm()
+        product = self.product1
+        product._compute_sale_lines_count()
```
pedrobaeza: If you have to call directly this method, something it's incorrect.

BhaveshHeliconia:
@pedrobaeza, I updated test cases

pedrobaeza:
When I make a comment, my intention is to not repeat the same thing across whole code, but you have to check everywhere for the same occurrence. That's the case with the removal of `_compute_sale_lines_count` explicit call.

BhaveshHeliconia:
@pedrobaeza, Thanks for the note! I’ve cleaned up all such occurrences across the code.this is now resolved.

product_form_sale_link/models/product_template.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
```
pedrobaeza, line 4: Remove logging

pedrobaeza (commented):
I'm not very convinced of the efficiency of the tests, but let's not block more.

/ocabot merge nobump

product_form_sale_link/models/product_template.py
BhaveshHeliconia: @pedrobaeza, https://github.com/OCA/sale-workflow/pull/3681 Please Review.

PR #3487, opened 2024-12-13
[18.0][MIG] sale_order_archive: Migration to 18.0

BhaveshHeliconia:


sale_order_archive/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    active = fields.Boolean(default=True)
```
StefanRijnhart:
```suggestion
    active = fields.Boolean(copy=False, default=True)
```

StefanRijnhart (changes requested):
Tested this module locally and while not a migration issue, there is a slightly dangerous flaw in this module: only done or cancelled orders can be archived, but once a cancelled order is archived, it can be reset to state draft. The order is still cancelled at that point. Can you make it so that cancelled orders cannot be reset to draft (hide button)?

sale_order_archive/models/sale_order.py
BhaveshHeliconia: @StefanRijnhart , Thnaks for the suggestions! it's fixed can you review it?

StefanRijnhart (approved):
Thanks for the improvements!

sale_order_archive/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    active = fields.Boolean(copy=False, default=True)
+
+    def toggle_active(self):
+        if self.filtered(
+            lambda so: not so.locked and so.state != "cancel" and so.active
+        ):
+            raise UserError(_("Only 'Locked' or 'Canceled' orders can be archived"))
```
JessBrandl: Potential performance improvement possible here according to the documentation: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

> You can replace the call to translate a string ( _ ) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

JessBrandl: Minor spelling mistake Canceled -> Cancelled

sale_order_archive/tests/test_sale_order_archive.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_sale_order_archive_toggle_active_locked(self):
+        self.assertTrue(self.sale_order.active)
+        self.sale_order.locked = True
+        self.sale_order.toggle_active()
+        self.assertFalse(self.sale_order.active)
+
+    def test_sale_order_archive_donot_toggle_active(self):
+        self.assertTrue(self.sale_order.active)
+        regex = r"Only 'Locked' or 'Canceled' orders can be archived"
```
JessBrandl, line 36: Same as above: Minor spelling mistake Canceled -> Cancelled

JessBrandl (commented):
Tested in Runboat and worked as intended, commented some minor issues

sale_order_archive/models/sale_order.py
BhaveshHeliconia: @JessBrandl, Thanks for the suggestions!Please review.

sale_order_archive/tests/test_sale_order_archive.py
StefanRijnhart: You need to pick a fight with @jbaudoux cause he's been asking to replace Cancelled by Canceled. FYI it's both correct, it's just that Canceled is more common in American English, and Cancelled is more common in British English.

jbaudoux: Indeed, we must use en_US terms. So it is `Canceled`

sale_order_archive/models/sale_order.py
jbaudoux: `Canceled` was right

sale_order_archive/tests/test_sale_order_archive.py
JessBrandl: I see, I cross-compared it to the status seen on the quotations overview page, there it is written as "Cancelled". But as far as I can tell it is not consistent within Odoo overall. 
Cancelled

jbaudoux (changes requested):
Please do not do non migration changes in the migration commit.

Please revert to `Canceled`

sale_order_archive/tests/test_sale_order_archive.py
StefanRijnhart: Yep, you'd expect Odoo to have en_US as its basis but actually it is leaning towards Cancelled.
```
~/odoo/18.0/odoo$ grep -i cancelled --exclude-dir=i18n * -r 2>/dev/null|wc -l 
761
~/odoo/18.0/odoo$ grep -i canceled --exclude-dir=i18n * -r 2>/dev/null|wc -l 
343
```
Maybe the point is that both options are not wrong, and we should not be asking for these trivial corrections during code review?

StefanRijnhart:
@jbaudoux matter of opinion. It seems fine to me to include trivial changes like spelling changes in the migration commit.

sale_order_archive/models/sale_order.py
CRogos: I agree with @jbaudoux . The odoo default should be en_US, and the GB should be in the en_GB.po translation file. (even though it is currently not always clean/consistent.)

StefanRijnhart: Judging from the selection values on a database with all community modules installed, Odoo might actually have tried to standardize on `Cancelled`. Again, I would prefer if we just left this to the whim of each developer, but If we're going to be consistent across OCA modules, there is something to say to be in line with Odoo as well.
image

CRogos: I think cancel is really an edge case and I agree that there is no clear standard in Odoo when searching the code.

image
image

Neverthless when you install a database, the selected language is also clear en-US:
image

Therefore best practice should be using en-US and not en-GB in my opinion.

StefanRijnhart: There is a difference in code occurrences and UI occurrences, which is why I think selection value labels give a better indication than code search.

CRogos:
@HeliconiaSolutions here are also some commits (I've marked all) which should be squashed: 
image

oca-ci, oca-git-bot, weblate, ...

sale_order_archive/models/sale_order.py
StefanRijnhart: If we really want to make this productive, maybe you could try to get someone from Odoo to make a statement on the matter. Until then, we could refrain from asking contributors to make the change in OCA modules.

CRogos: I'll open a ticket at Odoo, but I am not very optimistic.

CRogos: @pedrobaeza how should we handle this in OCA? Is there a (strict) best practice regarding en_US/en_GB?

StefanRijnhart: @CRogos To make it more concrete, you could try to quickly make a PR with all occurrences of Cancelled replaced by Canceled and see what the response is.

pedrobaeza: There's no strict guideline of this in OCA, as this is an edge case that didn't happen before. My take on this is indeed to go to "English (US)" form, as the header language variant is that one. We already have it in "Color", "Behavior", etc instead of "Colour" / "Behaviour". I think this specific problem is because Odoo developers are not so aware on this regional differences, and they put that in a similar way as a typo, and the corresponding dictionaries in their IDE don't warn them.

jbaudoux (approved):

CRogos (approved):

PR #3420, opened 2024-11-22
[18.0][MIG] sale_procurement_group_by_line: Migration to 18.0.

MeritxellAForgeFlow:
Migration to 18.0

@ForgeFlow

PR #3496, opened 2024-12-17
[18.0][MIG] sale_sourced_by_line: Migration to 18.0

ajay-odoogap:
**Note**

- depends on
- #3420


MiquelRForgeFlow:
Please rebase

sale_sourced_by_line/model/sale.py
```diff
@@ -0,0 +1,68 @@
...
+    _inherit = "sale.order.line"
+
+    @api.depends(
+        "route_id", "order_id.warehouse_id", "product_packaging_id", "product_id"
+    )
+    def _compute_warehouse_id(self):
+        """compute the warehouse for the lines only
+        if it has not already been set."""
+        lines = self.filtered(lambda rec: not rec.warehouse_id)
+        return super(SaleOrderLine, lines)._compute_warehouse_id()
```
antonioburic, line 32: why do we need to do this exactly, given that it's already computed in the core? 
the depends also do not change ^^

antonioburic (commented):
Technical review only.  
Thanks for the port! It would also be nice to squash some administrative commits where possible.

sale_sourced_by_line/model/sale.py
ajay-odoogap: @antonioburic to ensure that the warehouse manually set on the sale order line is not recomputed when the warehouse is changed at the sale order level

antonioburic: ah indeed, thx!

antonioburic (approved):

Ricardo-MC (approved):
looks good to me, 
thanks for the work

celm1990 (approved):
LGTM
@pedrobaeza Could you please review this? Thanks in advance.

PR #3498, opened 2024-12-17
[FIX] [16.0] sale_elaboration: Apply groups to views correctly

Shide:
Groups are not being applied correctly because the template of elaboration has a check for the group.

Now you can select Groups independently and shows elaborations correctly depending on what group you choose.

MT-8419 @moduon @rafaelbn @yajo @fcvalgar please reivew if you want 😄

### DIFF
--- a/sale_elaboration/README.rst
+++ b/sale_elaboration/README.rst
@@ -50,6 +50,10 @@ To configure this module you need to:
 #. Create a new record.
 #. Set a product linked to the elaboration.
 #. Also you can select a route to procure this elaboration.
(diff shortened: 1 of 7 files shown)


rafaelbn (approved):
Tested functionally in runbot 👍🏼 

http://oca-sale-workflow-16-0-pr3498-c65c7ac4f0c2.runboat.odoo-community.org/web#id=29&cids=1&menu_id=219&action=326&active_id=25&model=stock.picking&view_type=form

fcvalgar (approved):
Functional review,

LGTM thank you @Shide

PR #3500, opened 2024-12-17
[ADD] sale_order_type: change of analytical account by project

matiasperalta1:


nicolascol (approved):
LGTM

nicolascol:
Hi @rvalyi !
Can you merge this PR?
We added the "project_id" because of the new analytic approach in v18.
[Odoo PR with new approach](https://github.com/odoo/odoo/pull/169868/)

rousseldenis:
@matiasperalta1 Could you rename your commit and this description as this is more correct: `[IMP] sale_order_type: Use project instead of analytic account on types`.

But, with this change, existing installs will loose data. It should contain migration scripts if possible.

rousseldenis (changes requested):
But, IMHO, this is not correct as analytic account still exists.

This is a change of behavior as an analytic account is not a project. That's why you've added a new dependency from sale_project.

If you want to assign a project to the type, you should do a new module that allow that.

nicolascol:
Hi @rousseldenis 
The "analytic_account_id" field doesn't exist in sale.order v18, Odoo replaced it with "project_id". [More info](https://github.com/odoo/odoo/pull/169868/)
This PR was made to adapt sale_order_type to v18. In sale_order_type v17 an "analytic_account_id" could be configured. [In the v18 migration PR](https://github.com/OCA/sale-workflow/pull/3352) we were evaluating this change to adapt it but that PR was merged and that is why we are now making this new PR.

sale_order_type/views/sale_order_type_view.xml
```diff
@@ -39,8 +39,8 @@
                             />
                             <field name="sequence_id" />
                             <field
-                                name="analytic_account_id"
-                                groups="analytic.group_analytic_accounting"
+                                name="project_id"
+                                groups="project.group_project_user,analytic.group_analytic_accounting"
```
StefanRijnhart, line 39: Analogous to https://github.com/odoo/odoo/blob/18.0/addons/sale_project/views/sale_order_views.xml#L33, `project.group_project_user` seems enough here. Can you also take `context="{'default_allow_billable': True}"` here as well?

StefanRijnhart (commented):
Can you add a migration script to assign the project with the `account_id` matching the original `analytic_account_id` values on the order types?

sale_order_type/models/sale_order_type.py
```diff
@@ -52,10 +52,11 @@ class SaleOrderTypology(models.Model):
...
     )
-    analytic_account_id = fields.Many2one(
-        comodel_name="account.analytic.account",
-        string="Analytic account",
-        check_company=True,
+    project_id = fields.Many2one(
+        comodel_name="project.project",
+        domain=[("allow_billable", "=", True)],
+        string="Project",
+        help="Select to define the analytics account (new approach in v18)",
```
StefanRijnhart: We don't need to add a changelog here.
```suggestion
        help="Select to define the analytics account",
```

But maybe consider to add `check_company=True` here.

sale_order_type/views/sale_order_type_view.xml
StefanRijnhart: @matiasperalta1 You resolved this comment (I prefer you leave it to the reviewer to resolve), but `groups` is still inconsistent with the project_id field on the sale order view.

sale_order_type/models/sale_order_type.py
StefanRijnhart: Any type of company check between order type and project (like `check_company` on the old analytic account field)  still seems mising.

rousseldenis:

jjscarafia:
@rousseldenis @StefanRijnhart @matiasperalta1 
As I understand:
1. the analytic_account_id is not used anymore, could be removed

2. if we want to add some new functionality related to "analytics" maybe we should add :
2a. add an "analytic_distribution" field (I believe this would be cleaner)
2b. or extends analytic distribution models to be able to map with types 

3. If there is need to set "project_id" by a sale order type (not linked to analytics need), I agree that a new module would be needed

What do you think of going  with "1" and "2a"?

PR #3503, opened 2024-12-18
[FIX] sale_exception: Don't order by main exception

### LINKED ISSUE #3495: sale_exception: sale.order _order property
In https://github.com/OCA/sale-workflow/commit/8e7b446da609600f9cd06a63464868d23fff6804#diff-dc48d49fa585bc657a22c490084f3d40dedddc12b8d973a22fcafe1471f73eb1R69

The main exception has been put in `_order` attribute.

The thing is changing default model `_order` can lead to misbehaviors for module that are not aware of it.

IMHO, this should be avoided.

@sebastienbeau What do you think ?

rousseldenis:
As this is breaking the standard behavior, don't order on model side.

@sebastienbeau

### DIFF
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -9,7 +9,6 @@ from odoo import api, models
 class SaleOrder(models.Model):
     _inherit = ["sale.order", "base.exception"]
     _name = "sale.order"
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

lmignon (approved):

rousseldenis:
@hparfr

PR #3415, opened 2024-11-20
[18.0][MIG] sale_commercial_partner: Migration to 18.0

BertVGroenendael:

PR #3504, opened 2024-12-18
[MIG] sale_commercial_partner: Migration to 18.0 

LucasTran380381:
base on: https://github.com/OCA/sale-workflow/pull/3415


StefanRijnhart (approved):
Thanks!

alexis-via (changes requested):
Please don't remove the inherit of sale.view_order_form that adds the field commercial_partner_id as invisible in the view. It's very convenient when you want to add (in another module) a field that has a domain that uses commercial_partner_id.

StefanRijnhart:
@alexis-via That's what I thought, but then @LucasTran380381 [pointed out](https://github.com/OCA/sale-workflow/pull/3415#discussion_r1867411779) correctly that Odoo 18 will automatically add fields that are only used in domains and other attributes (as per https://github.com/odoo/odoo/pull/137031)

alexis-via (approved):

PR #3505, opened 2024-12-18
[18.0][MIG] sale_start_end_dates

alexis-via:
Migrate from v17 to v18
Add fields in sale.report (improvement made in a separate commit)


rvalyi:
@alexis-via I take the opportunity to remind this demand from @sbidoul to move the account_invoice_start_end_dates dependency to LGPL https://github.com/OCA/account-closing/pull/300 which IMHO would be a great move for such a small utility module. 

My personal opinion on this is:

- the AGPL OCA ecosystem keeps growing massively, it's not a big deal if we relicense a few trivial modules to LGPL from time to time, in fact I would say it's natural that very simple modules become a permissive commodity. In the meantime our AGPL ecosystem keeps making a pressure for an open source ecosystem for more added value modules.
- trivial modules like account_invoice_start_end_dates are simply copied in non open source modules whenever people need it. Nobody will discover it, the OCA or the authors won't get any license infrigement compensation. But if such module had a permissive license instead, the adoption would be larger, that would help the maintenance and grow the open source ecosystem overall.

sale_start_end_dates/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Sale Start End Dates",
-    "version": "17.0.1.0.0",
+    "version": "18.0.1.0.0",
```
rousseldenis, line 8: @alexis-via Could you put this change in migration commit ? Thanks

fabian-digital (approved):

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+    @api.constrains("default_start_date", "default_end_date")
+    def _check_default_start_end_dates(self):
+        for order in self:
+            if (
+                order.default_start_date
+                and order.default_end_date
+                and order.default_start_date > order.default_end_date
+            ):
+                raise ValidationError(
+                    _(
```
chaule97, line 32:
```suggestion
                    self.env._(
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+            if line.start_date and line.end_date:
+                days = (line.end_date - line.start_date).days + 1
+            line.number_of_days = days
+
+    @api.onchange("number_of_days")
+    def _inverse_number_of_days(self):
+        res = {"warning": {}}
+        for line in self:
+            if line.number_of_days < 0:
+                res["warning"]["title"] = _("Wrong number of days")
```
chaule97, line 98:
```suggestion
                res["warning"]["title"] = self.env._("Wrong number of days")
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+                res["warning"]["message"] = _(
```
chaule97, line 99:
```suggestion
                res["warning"]["message"] = self.env._(
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+        return res
+
+    @api.constrains("product_id", "start_date", "end_date")
+    def _check_start_end_dates(self):
+        for line in self:
+            if line.product_id.must_have_dates:
+                if not line.end_date:
+                    raise ValidationError(
+                        _("Missing End Date for sale order line with Product '%s'.")
+                        % (line.product_id.display_name)
```
chaule97, line 124:
```suggestion
                        self.env._("Missing End Date for sale order line with Product '%s'.", line.product_id.display_name)
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+                    )
+                if not line.start_date:
+                    raise ValidationError(
+                        _(
+                            "Missing Start Date for sale order line with "
+                            "Product '%s'."
+                        )
+                        % (line.product_id.display_name)
```
chaule97, line 132:
```suggestion
                        self.env._(
                            "Missing Start Date for sale order line with "
                            "Product '%s'.",
                            line.product_id.display_name
                        )
```

sale_start_end_dates/models/sale_order.py
```diff
@@ -0,0 +1,181 @@
...
+                    )
+                if line.start_date > line.end_date:
+                    raise ValidationError(
+                        _(
```
chaule97, line 136:
```suggestion
                        self.env._(
```

rousseldenis:
@alexis-via

chaule97:
Salut @rousseldenis , do you want me to follow up it?

alexey-pelykh (approved):
Thanks @alexis-via for the migration, looks good.

The model code is unchanged from 17.0 which is fine -- the ORM API used here (`@api.constrains`, `@api.depends`, computed fields with `store=True, readonly=False`) is stable across versions. The `_prepare_invoice_line` override correctly uses `super()` with `**optional_values`.

View changes are correct for 18.0: the xpath in the form correctly targets `product_uom_readonly` (instead of `product_id` in 17.0), and the list views use `list` instead of `tree`. The new `sale_order_line.xml` views extending `view_order_line_tree` and `sale_order_line_view_form_readonly` are a nice addition over 17.0.

The `sale.report` extension is properly implemented using `_select_additional_fields` and `_group_by_sale` which are the correct hooks in Odoo 18.

Regarding the `self.env._()` suggestions in the other review thread -- those should be disregarded. Odoo 18 core itself still uses the global `from odoo import _` pattern, so the current code is correct.

Manifest and pyproject.toml look fine.

sale_start_end_dates/__manifest__.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

sale_start_end_dates/models/sale_order.py
florian-dacosta: Done

florian-dacosta (approved):
@rousseldenis Comments attended, can we merge this please ?

rousseldenis (approved):

PR #3507, opened 2024-12-19
[16.0][ADD] sale_stock_prebook_cancel_line: Glue addon

lmignon:
This addon is a technical module that ensures that the computation of the cancelled quantity on a sale order line is correct even when a prebooking is involved. Cancelled prebooked quantities must not be taken into account in the computation of the cancelled quantity on the sale order line.

depends on

* [ ] #3481 
* [ ] #2357


lmignon:
moved to https://github.com/OCA/sale-prebook/pull/5

PR #3512, opened 2024-12-23
[IMP] sale_advance_payment: Use method lines

etobella:
We are not using the default behavior of accounting.

### DIFF
--- a/sale_advance_payment/wizard/sale_advance_payment_wzd_view.xml
+++ b/sale_advance_payment/wizard/sale_advance_payment_wzd_view.xml
@@ -6,6 +6,7 @@
         <field name="type">form</field>
         <field name="arch" type="xml">
             <form string="Advance Payment">
+                <field name="available_payment_method_line_ids" invisible="1" />
                 <group>
                     <field name="order_id" invisible="1" />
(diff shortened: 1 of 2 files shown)


luisDIXMIT (approved):
Tested on runboat, LGTM!

HaraldPanten:
@etobella could you rebase, please?

sale_advance_payment/wizard/sale_advance_payment_wzd_view.xml
```diff
@@ -14,7 +15,12 @@
                             domain="[('type','in',['bank', 'cash'])]"
                             widget="selection"
                             select="1"
+                        />
+                        <field
+                            name="payment_method_line_id"
+                            options="{'no_create': True, 'no_open': True}"
```
ValentinVinagre, line 21: why the field is editable by model but not in the view ?

etobella: It is editable, but We don't allow creation or opening the item.

ValentinVinagre (approved):
LGTM 👍🏻

PR #3513, opened 2024-12-23
[17.0][FIX] sale_fixed_discount: Fixed discount consideration for por…

fd-oerp:
…tal orders

Issue:
When a pricelist is applied to a specific product with a discount, the discount is visible on the portal product page. However, during checkout, the discount is not reflected in the cart and is not considered in the total.
Attached is a screenshot showing how the discount was not reflected in the portal cart before this fix: 
image

Fix:
The issue occurred because the module was calling the base method _compute_discount but did not call the super method, which resulted in this behavior. This has now been addressed with the fix in this commit.

Expected Behavior:
Products with discounts applied via pricelists will now correctly show the discount in the cart during checkout.
Attached is a screenshot showing how the discount is now correctly reflected in the cart after applying this fix : 
image

### DIFF
--- a/sale_fixed_discount/models/sale_order_line.py
+++ b/sale_fixed_discount/models/sale_order_line.py
@@ -83,7 +83,9 @@ class SaleOrderLine(models.Model):
         lines_with_discount_fixed = self.filtered(lambda sol: sol.discount_fixed)
         for line in lines_with_discount_fixed:
             line.discount = line._get_discount_from_fixed_discount()
-        return super(SaleOrderLine, self - lines_with_discount_fixed)
+        return super(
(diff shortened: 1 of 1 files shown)


fd-oerp:
Hi @rousseldenis 

Could you please review this fix/PR and let me know if anything needs to be updated from my end?

Thank you

fd-oerp:
Hi @rousseldenis 

Could you please review this PR and let me know if anything needs to be updated from my end?

Thank you

rousseldenis (approved):
This is trivial.

@fd-oerp Sorry to get back here late. Could you add a test for that to avoid regression ?

rousseldenis:
This has been fixed

PR #3515, opened 2024-12-26
[IMP] sale_advance_payment: Allow edition of the related payments

etobella:
Actually, you can press buttons on the related payment that would make it editable, but you cannot edit it.

### DIFF
--- a/sale_advance_payment/views/account_payment.xml
+++ b/sale_advance_payment/views/account_payment.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <record id="view_account_payment_tree" model="ir.ui.view">
+        <field name="name">account.payment.tree</field>
+        <field name="model">account.payment</field>
+        <field name="mode">primary</field>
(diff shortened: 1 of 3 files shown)


luisDIXMIT (approved):
Tested on runboat, LGTM!

HaraldPanten:
@etobella could you rebase?

sale_advance_payment/views/account_payment.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <record id="view_account_payment_tree" model="ir.ui.view">
+        <field name="name">account.payment.tree</field>
+        <field name="model">account.payment</field>
+        <field name="mode">primary</field>
```
ValentinVinagre, line 7: Should priority="999999" be specified to apply all other inheritances first?

etobella: There is no need.

ValentinVinagre (approved):
LGTM 👍🏻

PR #3519, opened 2024-12-31
[18.0][MIG] sale_order_line_product_attribute_values: Migration to 18.0

BhaveshHeliconia:


rousseldenis (approved):

u0f (changes requested):
Thanks for this migration, @BhaveshHeliconia. Could you please [reduce the commit history](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests)?

Also, please rebase the PR, as Runboat is not accessible to check the migration. Thanks!

BhaveshHeliconia:
@u0f : There are only 5 commits in this PR, so nothing is reducible at this point.
I’ve completed the rebase, and Runboat is available now.

Thanks!

u0f (approved):
LGTM. Code and functional review 

https://www.loom.com/share/45bbf8cb932d4355b441381e38368d16 @moduon

Gelojr (commented):
During functional testing of the module sale_order_line_product_attribute_values on Odoo 18, the following results were obtained:

- Test 1 – Computation of all attributes on a sale order line (OK)
- Test 2 – Display of extra price in attribute list (OK)
- Test 3 – Search by Attribute Values in sale order lines (NOT OK)
Reproduction steps:
Create an attribute “Trademark” with values “Adidas” and “Nike” (create_variant = “never”).
Sell a product using these attribute values in several sale orders.
Open the sale order lines view and use the search field Attribute Values (all_product_attribute_value_ids).
Search for the value “Adidas”.
Current behavior: No matching lines are returned, as if the attribute values were not present.

Image

Image

Expected behavior: All sale order lines containing the attribute value “Adidas” in all_product_attribute_value_ids should appear in the results.

- Test 4 – Group by Attribute Values in sale order lines (NOT OK)
Reproduction steps:
Using the same dataset as Test 3, open the sale order lines view.
Use the “Attribute Values” group-by option.
Current behavior: Lines are not grouped by their attribute values. Groups appear empty or inconsistent.

Image

Expected behavior: The system should group sale order lines based on the values in all_product_attribute_value_ids (e.g., a group for “Adidas”, another for “Nike”, etc.).

- Test 5 – Inclusion of variant-creating attributes (OK, with clarification)
Note: Variant-creating attributes (such as “Outer Material: Corduroy”) correctly appear in the tags and in the attribute table of the sale order line form. However, these values do not appear in the textual description under the product name in the sale order line, because Odoo only shows non-variant and custom attributes in that area. This is standard Odoo behavior and not an issue with the module, but it is important to acknowledge.

Gelojr (commented):
Review requested.

## Module configuration (change in the wizard default behavior)

**Changes introduced by the PR (default wizard configuration):**
For the following models, the listed fields are included in the wizard but unchecked by default:

* `account.tax.group`: `sequence`
* `account.tax`: `sequence`, `children_tax_ids`
* `account.account`: `root_id`
* `account.group`: `parent_id`, `code_prefix_end`
* `account.fiscal.position`: `sequence`

## Tests performed and results

* TEST1 — OK
* TEST2 — OK
* TEST3 — NOT OK (Taxes: Sequence)

  * Starting point Image
  * During the test, the **Sequence** field was explicitly selected in the wizard for `account.tax`. Image
* However, after running the update, **the tax sequence order was not updated**, so the expected change was not applied. Image

## Reason for requesting a review

The test performed on Taxes shows that **selecting the Sequence field does not result in an actual sequence update**. This behavior does not match the expected outcome.

PR #3522, opened 2025-01-03
[FIX] sale_advance_payment: Use signed amount when computing residuals

etobella:
Otherwise, we might get incorrect residuals when we get refunds


luisDIXMIT (approved):
Tested on runboat, LGTM!

rvalyi (approved):

PR #3070, opened 2024-04-11
[16.0][MIG] sale_order_line_description

MarioLM-23:
Last PR of migration module in open stage: [#2450](https://github.com/OCA/sale-workflow/pull/2450)

PR #3523, opened 2025-01-07
[16.0][MIG] sale_order_line_description

tafaRU:
Superseeds https://github.com/OCA/sale-workflow/pull/3070


sale_order_line_description/models/sale_order_line.py
```diff
@@ -7,9 +7,9 @@
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    @api.onchange("product_id")
-    def product_id_change(self):
-        res = super(SaleOrderLine, self).product_id_change()
+    @api.depends("product_id")
```
rousseldenis: You can remove the depends as already defined in sale module.

sale_order_line_description/models/sale_order_line.py
```diff
@@ -7,9 +7,9 @@
...
+    def _compute_name(self):
+        res = super()._compute_name()
         if not self.product_id:  # pragma: no cover
```
rousseldenis: You should change the function implementation to support multi recordset.

sale_order_line_description/models/sale_order_line.py
tafaRU: done, thanks!

sale_order_line_description/models/sale_order_line.py
tafaRU: done, thanks!

Jaimermaccione (commented):
Initially, everything seems to be working correctly. However, as mentioned above, if you add the product using the 'Add Multiple Products' function, it does not work as expected.

tafaRU:
@Jaimermaccione thanks a lot for your review!

Concerning the following:
> if you add the product using the 'Add Multiple Products' function, it does not work as expected.
it's due to in runboat in also installed https://github.com/OCA/sale-workflow/tree/16.0/sale_product_multi_add.

I suppose you have the the same behavior in previous versions (i.e `15.0`).

I prefer not to add `sale_product_multi_add` as depends. The ideal solution to get two modules compatible should be to a create a module bridge. I'd now just add a ROADMAP section where to highlight the behavior.

tafaRU:
@rousseldenis @Jaimermaccione could you please update your review?

rousseldenis (approved):

DorianMAG (approved):
Functional test
LGTM
Thx for this work

Jaimermaccione (approved):
LGTM! Functional review on runboat.

DorianMAG:
ping @flotho,  @gaelTorrecillas, @JulienMartinez

gaelTorrecillas (approved):

Jaimermaccione:
@Tisho99 I have seen that now it is displayed correctly in the tree view, but it can no longer be consulted in the pivot view. Is that correct? I understand that this is due to your previous conversation with @rousseldenis , but does this mean that in the pivot view, we would just have to deduce the calculation manually? I'm asking without knowing the technical implications in depth.

Thank you!

dreispt (approved):

flotho:
pop @DorianMAG , could you have a look to this one ?

PR #3524, opened 2025-01-08
[16.0][FIX] sale_quotation_number: fix numbering to be compatible with sale_order_revision

Guidomadormo:
I implement the changes made in this PR: https://github.com/OCA/sale-workflow/pull/3305, as the same problem occurs in version 16.0.

### DIFF
--- a/sale_quotation_number/__manifest__.py
+++ b/sale_quotation_number/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Sale Quotation Numeration",
     "summary": "Different sequence for sale quotations",
-    "version": "16.0.1.1.0",
+    "version": "16.0.1.1.1",
     "category": "Sales Management",
     "website": "https://github.com/OCA/sale-workflow",
     "author": "Elico Corp, "
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -13,20 +13,21 @@ class SaleOrder(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
-            if self.is_using_quotation_number(vals):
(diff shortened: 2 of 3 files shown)


alexey-pelykh (changes requested):
Thanks for the backport effort -- the underlying fix from #3305 is definitely needed on 16.0 as well.

However, the PR currently has merge conflicts and the diff appears to have been developed against a codebase that doesn't match the current 16.0 branch. A few specific issues:

1. **Missing `quotation_seq_used` handling**: On 16.0, the `create` method sets `vals["quotation_seq_used"] = True` alongside the name. This PR drops that entirely, which would break `action_confirm` -- it relies on `quotation_seq_used` to decide whether to reassign a sale order sequence on confirmation.

2. **Removed `get_quotation_seq()` call**: The 16.0 module uses `self.with_company(company_id).get_quotation_seq()` rather than the inline `self.env["ir.sequence"].next_by_code(...)` seen here. The PR replaces the helper method call with an inline sequence lookup, which is an unrelated change and breaks the extension point that `get_quotation_seq` provides.

3. **`is_using_quotation_number` check ordering**: On the merged 17.0 fix, the `not vals.get("name")` guard is placed *inside* the `is_using_quotation_number` check, so `quotation_seq_used` is still set to `True` even when a name is already provided. Your version wraps both checks together, which means `quotation_seq_used` would never be set.

I'd suggest rebasing on the current 16.0 branch and adapting the fix to match the actual 16.0 code structure. Looking at the 17.0 result after #3305 was merged should give a good reference -- the key changes needed for 16.0 are:
- Add `if not vals.get("name"):` guard around the name assignment (but keep `quotation_seq_used = True` outside it)
- Change `if "company_id" in vals:` to `if vals.get("company_id"):`

PR #3528, opened 2025-01-09
[16.0] [IMP] Make packaging required on orders for configured products.

edlopen:
This PR allows a configured product not to be allowed to be sold without packaging.

This video explains the improvement
https://www.loom.com/share/b4b6a6fc9cdb406086115195a1eeb1c4

@moduon: MT-8489

### DIFF
--- a/sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
+++ b/sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
@@ -2,7 +2,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
 
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class SaleOrderRecommendationLine(models.TransientModel):
@@ -26,6 +27,7 @@ class SaleOrderRecommendationLine(models.TransientModel):
         digits="Product Unit of Measure",
         help="Quantity of packagings to sell.",
     )
+    require_packaging = fields.Boolean(compute="_compute_require_packaging")
 
     @api.depends("product_id", "units_included", "sale_uom_id")
     def _compute_product_packaging(self):
@@ -109,3 +111,22 @@ class SaleOrderRecommendationLine(models.TransientModel):
             # Nothing to create
             return result
         return self._prepare_packaging_line_vals(result)
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
+
+    @api.depends("product_id", "product_uom_qty", "product_uom")
+    def _warning_product_packaging_required(self):
+        """Notify the user when a packaging is required."""
+        result = {}
+        for line in self:
+            if line.require_packaging and not line.product_packaging_id:
+                raise UserError(
+                    _(
+                        "The product %s requires a packaging to be sold"
+                        % line.product_id.name
+                    )
+                )
+        return result
--- a/sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation_view.xml
+++ b/sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation_view.xml
@@ -10,6 +10,14 @@
             ref="sale_order_product_recommendation.sale_order_recommendation_view_form"
         />
         <field name="arch" type="xml">
+
+            <xpath
+                expr="//field[@name='line_ids']/form//field[@name='units_included']"
+                position="before"
+            >
+                <field name="require_packaging" invisible="1" />
+            </xpath>
+
             <xpath
                 expr="//field[@name='line_ids']/tree/field[@name='units_included']"
                 position="before"
@@ -36,12 +44,13 @@
                     name="product_packaging_id"
                     groups="product.group_stock_packaging"
                     context="{'default_product_id': product_id}"
+                    attrs="{'required': [('require_packaging', '=', True)]}"
                 />
                 <field
                     name="product_packaging_qty"
                     groups="product.group_stock_packaging"
                     widget="numeric_step"
-                    attrs="{'readonly': [('product_packaging_id', '=', False)]}"
+                    attrs="{'readonly': ['|', ('product_packaging_id', '=', False), ('require_packaging', '=', True)]}"
                 />
             </xpath>
         </field>
--- a/sale_packaging_default/models/product_template.py
+++ b/sale_packaging_default/models/product_template.py
@@ -1,0 +1,9 @@
+# Copyright 2024 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    require_packaging = fields.Boolean(default=False)
--- a/sale_packaging_default/models/sale_order_line.py
+++ b/sale_packaging_default/models/sale_order_line.py
@@ -2,11 +2,14 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 from contextlib import suppress
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
+
+    require_packaging = fields.Boolean(compute="_compute_require_packaging")
 
     def onchange(self, values, field_name, field_onchange):
         """Record which field was being changed."""
@@ -43,6 +46,23 @@ class SaleOrderLine(models.Model):
                     and line.product_uom_qty % line.product_packaging_id.qty
                 ):
                     line.product_packaging_id = False
+                    res = line._warning_product_packaging_required()
+                    if res:
+                        return res
+        return result
+
+    @api.depends("product_id", "product_uom_qty", "product_uom")
+    def _warning_product_packaging_required(self):
+        """Notify the user when a packaging is required."""
+        result = {}
+        for line in self:
+            if line.require_packaging and not line.product_packaging_id:
+                raise UserError(
+                    _(
+                        "The product %s requires a packaging to be sold"
+                        % line.product_id.name
+                    )
+                )
         return result
 
     @api.model
@@ -84,3 +104,8 @@ class SaleOrderLine(models.Model):
         _self = self.with_context(keep_product_packaging=True)
         result = super(SaleOrderLine, _self)._compute_product_uom_qty()
         return result
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
--- a/sale_packaging_default/views/product_template_view.xml
+++ b/sale_packaging_default/views/product_template_view.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2024 Moduon Team S.L. <info@moduon.team>
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<data>
+
+    <record id="product_template_form_view_inherit" model="ir.ui.view">
+        <field name="name">Simplify packaging fields</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']" position="inside">
+                <group string="Packaging Division">
+                    <field
+                        name="require_packaging"
+                        placeholder="Indicates if the product can be sold individually or if it requires keeping the original packaging."
+                    />
+                </group>
+            </xpath>
+        </field>
+    </record>
+
+</data>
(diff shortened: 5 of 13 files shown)


sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
...
             return result
         return self._prepare_packaging_line_vals(result)
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
+
+    @api.depends("product_id", "product_uom_qty", "product_uom")
+    def _warning_product_packaging_required(self):
```
yajo, line 121: This is a constraint instead:

```suggestion
    @api.constrains("product_id", "product_packaging_id")
    def _check_product_packaging_required(self):
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
...
+        """Notify the user when a packaging is required."""
+        result = {}
+        for line in self:
+            if line.require_packaging and not line.product_packaging_id:
+                raise UserError(
```
yajo, line 126: issue: [within constrains, you should raise `ValidationError`s](https://www.odoo.com/documentation/17.0/developer/reference/backend/orm.html?highlight=constrains#odoo.api.constrains):

```suggestion
                raise ValidationError(
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
...
+                    _(
+                        "The product %s requires a packaging to be sold"
+                        % line.product_id.name
```
yajo, line 129: issue: you must render the message before composing it for translations to work fine:

```suggestion
                        "The product %s requires a packaging to be sold",
                        line.product_id.name
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
...
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
+
+    @api.depends("product_id", "product_uom_qty", "product_uom")
+    def _warning_product_packaging_required(self):
+        """Notify the user when a packaging is required."""
+        result = {}
```
yajo, line 123: issue: no need for a result while constraining.

```suggestion
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
...
+        for line in self:
+            if line.require_packaging and not line.product_packaging_id:
+                raise UserError(
+                    _(
+                        "The product %s requires a packaging to be sold"
+                        % line.product_id.name
+                    )
+                )
+        return result
```
yajo, line 132: continuing from above...

```suggestion
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -26,6 +27,7 @@ class SaleOrderRecommendationLine(models.TransientModel):
         digits="Product Unit of Measure",
         help="Quantity of packagings to sell.",
     )
+    require_packaging = fields.Boolean(compute="_compute_require_packaging")
```
yajo, line 30: issue: you can use a readonly related field instead, which is essentially the same:

```suggestion
    require_packaging = fields.Boolean(related="product_id.require_packaging", readonly=True)
```

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation.py
```diff
@@ -109,3 +111,22 @@ def _prepare_new_so_line_vals(self, sequence):
             # Nothing to create
             return result
         return self._prepare_packaging_line_vals(result)
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
```
yajo, line 118: issue: if using a related field, you don't really need to add a compute method.

```suggestion
```

sale_packaging_default/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    require_packaging = fields.Boolean(default=False)
```
yajo, line 9: issue: add some help and make sure you cannot require packaging without packaging:

```suggestion
    require_packaging = fields.Boolean(help="Activate it to require that the product is always sold in a complete packaging.")

    @api.constrains("require_packaging", "packaging_ids")
    def _check_packaging_required(self):
        wrong = (self - self.filtered("packaging_ids")).filtered("require_packaging")
        if wrong:
            raise ValidationError(_("Products %s require packaging but have none", ", ".join(wrong.mapped("display_name"))))
```

sale_packaging_default/models/sale_order_line.py
```diff
@@ -84,3 +104,8 @@ def _compute_product_uom_qty(self):
         _self = self.with_context(keep_product_packaging=True)
         result = super(SaleOrderLine, _self)._compute_product_uom_qty()
         return result
+
+    @api.depends("product_id")
+    def _compute_require_packaging(self):
+        for line in self:
+            line.require_packaging = line.product_id.require_packaging
```
yajo, line 111: use related instead

```suggestion
```

sale_packaging_default/models/sale_order_line.py
```diff
@@ -2,12 +2,15 @@
...
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    require_packaging = fields.Boolean(compute="_compute_require_packaging")
```
yajo, line 12: use related instead
```suggestion
    require_packaging = fields.Boolean(related="product_id.require_packaging", readonly=True)
```

sale_packaging_default/models/sale_order_line.py
```diff
@@ -43,6 +46,23 @@ def _compute_product_packaging_id(self):
...
+        result = {}
+        for line in self:
+            if line.require_packaging and not line.product_packaging_id:
+                raise UserError(
+                    _(
+                        "The product %s requires a packaging to be sold"
+                        % line.product_id.name
+                    )
+                )
         return result
```
yajo, line 66: issue: use a constrain instead, and I summarize all the comments from above here too:

```suggestion

    @api.constrains("product_id", "product_packaging_id")
    def _check_product_packaging_required(self):
        """Notify the user when a packaging is required."""
        for line in self:
            if line.require_packaging and not line.product_packaging_id:
                raise ValidationError(
                    _(
                        "The product %s requires a packaging to be sold",
                        line.product_id.name
                    )
                )
```

sale_packaging_default/views/product_template_view.xml
```diff
@@ -0,0 +1,22 @@
...
+    <record id="product_template_form_view_inherit" model="ir.ui.view">
+        <field name="name">Simplify packaging fields</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']" position="inside">
+                <group string="Packaging Division">
+                    <field
+                        name="require_packaging"
+                        placeholder="Indicates if the product can be sold individually or if it requires keeping the original packaging."
```
yajo, line 15: issue: a placeholder doesn't make sense if the input is a checkbox. I already commented above about adding help to the field, so this is unnecessary.

```suggestion
```

sale_packaging_default/views/product_template_view.xml
```diff
@@ -0,0 +1,22 @@
...
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<data>
+
+    <record id="product_template_form_view_inherit" model="ir.ui.view">
+        <field name="name">Simplify packaging fields</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']" position="inside">
+                <group string="Packaging Division">
```
yajo, line 12: issue: since you're creating a new group, it can land far away from the packaging configuration. Better, put the new checkbox within the same visual group as the rest of packaging configuration:

image

sale_packaging_default/views/sale_order_view.xml
```diff
@@ -36,6 +36,33 @@
...
+
+            <!-- Make packaging qty readonly when the product is not allow to divide the packaging -->
+            <xpath
+                expr="//field[@name='order_line']/tree/field[@name='product_packaging_qty']"
+                position="attributes"
+            >
+                <attribute
+                    name="attrs"
+                >{'readonly': [('require_packaging', '=', True)]}</attribute>
+            </xpath>
```
yajo, line 65: issue: so, we require packaging, but we don't allow to set how many? It doesn't make sense:
image

Instead, please do the opposite thing. The field that you should make readonly is `product_uom_qty`. That way, the user is not allowed to break the packaging. They're only allowed to increase or decrease whole packaging sets.

sale_order_product_recommendation_packaging_default/wizards/sale_order_recommendation_view.xml
```diff
@@ -36,12 +44,13 @@
...
                     context="{'default_product_id': product_id}"
+                    attrs="{'required': [('require_packaging', '=', True)]}"
                 />
                 <field
                     name="product_packaging_qty"
                     groups="product.group_stock_packaging"
                     widget="numeric_step"
-                    attrs="{'readonly': [('product_packaging_id', '=', False)]}"
+                    attrs="{'readonly': ['|', ('product_packaging_id', '=', False), ('require_packaging', '=', True)]}"
                 />
```
yajo, line 54: issue: just like below, this field should be writable always when there's a packaging. But instead, do the opposite and make the `product_uom_qty` field readonly when `require_packaging == True`.

yajo (changes requested):
It is close, but I added a lot of implementation details to help for making it more maintainable and usable.

Thanks!

Shide:
I think the best approach will be using the same workflow as Odoo does:

### On product.category
Add a selection field as `packaging_reserve_method` called `packaging_sell_method` with 2 options:
- **full**: Sell only full packagings
- **partial**: Sell partial packagings

The workflow will be the same as the `packaging_reserve_method` field, but checking packagings when selling products.

yajo:
Since those fields exist already, would it make more sense to reuse them,
instead of creating new ones?

pedrobaeza:
FYI, this change in 19.0 will totally impact you: https://github.com/odoo/odoo/pull/184131

Shide:
> Since those fields exist already, would it make more sense to reuse them, instead of creating new ones?
> […](#)
> Message ID: ***@***.***>

Could be interesting to force commercial users to sell entire packages, but maybe in certain cases allow to send partial packages...

mmequignon (changes requested):
We already have something similar in v14.0 https://github.com/OCA/sale-workflow/tree/14.0/sale_by_packaging
I would migrate it to 16.0 instead.

rafaelbn:
Hello @mmequignon ,

Thank you so much for your feedback and for pointing us to the `sale_by_packaging` module. We’re already familiar with it, but we believe it addresses a different use case. 😄  

The main difference lies in flexibility. Your module revolves around **strictly selling by packaging**, while our use case is much more **dynamic and flexible**. 🚀  

Here’s what we mean:  

- Sometimes we sell by packaging, and sometimes we don’t.  
- In some scenarios, packaging is used optionally, allowing for different combinations.  

For our clients, this flexibility is critical, and unfortunately, the `sale_by_packaging` module doesn’t meet their needs.  

Additionally, we’ve noticed that your module introduces extra complexity and dependencies that are unnecessary for our use case, such as:  

- `sale_order_line_packaging_qty`  
- `product_packaging_type`  

If we’re mistaken here, we’d love to see more clarity! 🙏 When migrating the module, it would be super helpful if you could improve the README and even record a short video 🖥️🎥 explaining how to handle these use cases:  

1. **Products without packaging**: Ensure packaging is not required.  
2. **No forced packaging type**: Allow products with packaging but without enforcing a type (not needed for us).  
3. **Optional packaging**: Products with packaging (no type) that can be sold either with or without packaging.  
4. **Conditional enforcement**: Products with packaging (no type) where selling by packaging is only enforced in specific cases.  

If your module already covers all these scenarios, we’d be delighted to reconsider it! 😊💡  

Thanks again for your input, and let us know your thoughts!  

Best regards,  
Rafael ❤️

rafaelbn:
Closing as final solutions not found

PR #3424, opened 2024-11-25
[16.0][ADD] sale_stock_prebook_stock_available_to_promise_release

lmignon:
Port of `sale_stock_prebook_stock_available_to_promise_release` addon from https://github.com/OCA/sale-workflow/pull/2624 to 16.0

* [ ] depends on #3423

PR #3532, opened 2025-01-10
[16.0] [ADD] Sale Order Blanket Order Stock Prebook Release

lmignon:
When a call-off order is confirmed for a product, the priority date used to compute the quantity available at promised date must be the start date of the blanket order.

This is required to ensure that the principle of first-come, first-served is respected. 
In the case of a product part of a blanket order the date to consider must be the start date of the blanket order and not the date of the call-off order.

depends on

* [ ] #3481
* [ ] #3424
* [ ] #3436
* [ ] #3457


lmignon:
addon moved to https://github.com/OCA/sale-blanket/pull/4

PR #3526, opened 2025-01-08
[17.0][MIG] sale_pricelist_global_rule: Migration to version 17.0

carlos-lopez-tecnativa:
- Reused Odoo's base code for computing pricelists, adapted it to new functions, and simplified the logic.
- Renamed selection options for the `applied_on` field: `4_global_product_template` to `3_1_global_product_template` and `5_global_product_category` to `3_2_global_product_category` to ensure the correct order in pricelist items.
- Fixed an assertion in the tests. Previously, changing the UoM did not update the price, but this behavior was introduced in PR https://github.com/OCA/sale-workflow/pull/2473.
- Added a new assertion for the `pricelist_item_id` field.

TT52357
@Tecnativa @pedrobaeza @victoralmau could you please review this.

PR #3533, opened 2025-01-10
[16.0][MIG] sale_pricelist_global_rule

lmchlfl:
Migration of the sale_pricelist_global_rule module from version 15.0 to 16.0

FL-556-4652


sale_pricelist_global_rule/models/product_pricelist.py
```diff
@@ -2,16 +2,50 @@
...
+            ORDER BY
+                item.applied_on, item.min_quantity desc, categ.complete_name desc, item.id desc
+            """,
+            (prod_tmpl_ids, prod_ids, categ_ids, self.id, date, date),
+        )
+        # NOTE: if you change `order by` on that query, make sure it matches
+        # _order from model to avoid inconstencies and undeterministic issues.
+
+        item_ids = [x[0] for x in self.env.cr.fetchall()]
+        items = self.env["product.pricelist.item"].browse(item_ids)
```
mmequignon: Are you re implementing the whole thing, and ignoring the recursion, or am I missing something?

mmequignon (changes requested):

carlos-lopez-tecnativa:
Hi, I started the migration to 17 and did a significant refactor to reuse Odoo functions. These functions have been available since V16. I think it’s a good idea to take into account the changes from V17, as they may also be applicable here. Please check this PR: https://github.com/OCA/sale-workflow/pull/3526/

sale_pricelist_global_rule/models/product_pricelist.py
lmchlfl: We have reimplemented the functionality by doing a backport from v17 and there should no longer be any problems

lmchlfl:
Hi @carlos-lopez-tecnativa, we have reimplemented the functionality by doing a backport from v17.

sale_pricelist_global_rule/views/sale_order_views.xml
```diff
@@ -0,0 +1,29 @@
...
+        <field name="name">view.sale.order.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='action_cancel']" position="before">
+                <button
+                    string="Recompute pricelist global"
+                    name="button_compute_pricelist_global_rule"
+                    type="object"
+                    attrs="{'invisible': [('state', 'not in', ['draft', 'sent']), ('has_pricelist_global', '=', False), ('need_recompute_pricelist_global', '=', False)]}"
```
yostashiro:
```suggestion
                    attrs="{'invisible': ['|', '|', ('state', 'not in', ['draft', 'sent']), ('has_pricelist_global', '=', False), ('need_recompute_pricelist_global', '=', False)]}"
```

sale_pricelist_global_rule/views/sale_order_views.xml
```diff
@@ -0,0 +1,29 @@
...
+                    class="oe_highlight"
+                />
+                <button
+                    string="Recompute pricelist global"
+                    name="button_compute_pricelist_global_rule"
+                    type="object"
+                    attrs="{'invisible': [('state', 'not in', ['draft', 'sent']), ('has_pricelist_global', '=', False), ('need_recompute_pricelist_global', '=', True)]}"
```
yostashiro:
```suggestion
                    attrs="{'invisible': ['|', '|', ('state', 'not in', ['draft', 'sent']), ('has_pricelist_global', '=', False), ('need_recompute_pricelist_global', '=', True)]}"
```

yostashiro (changes requested):
Thanks for the PR! Just some adjustments to the buttons. Otherwise, it seems to work fine.

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+            for line in sale.order_line:
+                suitable_rule = pricelist._get_product_rule(
+                    line.product_id,
+                    quantity=line.product_uom_qty or 1.0,
+                    uom=line.product_uom,
+                    date=line.order_id.date_order,
+                )
+                if suitable_rule:
+                    break
+            sale.has_pricelist_global = bool(suitable_rule)
```
yostashiro, line 58: I think the logic should be as follows, if I understand the intention correctly. Otherwise, `has_pricelist_global` becomes true for practically all cases.
```suggestion
            global_rule = self.env["product.pricelist.item"]
            for line in sale.order_line:
                suitable_rule = pricelist._get_product_rule(
                    line.product_id,
                    quantity=line.product_uom_qty or 1.0,
                    uom=line.product_uom,
                    date=line.order_id.date_order,
                )
                if not suitable_rule:
                    continue
                rule = global_rule.browse(suitable_rule)
                if rule.applied_on in ["3_1_global_product_template", "3_2_global_product_category"]:
                    global_rule = rule
                    break
            sale.has_pricelist_global = bool(global_rule)
```

diggy128:
Hi @LuciaMarchalFactorLibre 
Any chance to finish the backport for this?

lmchlfl:
Hi @diggy128,

I made the backport respecting what was in v17, if another person wants to take it up again to add any modifications, they are welcome.

sale_pricelist_global_rule/migrations/17.0.1.0.0/pre-migration.py
carlos-lopez-tecnativa, line 1: This folder must be renamed to `16.0.1.0.0`; the migration script is needed and must not be deleted.

sale_pricelist_global_rule/views/sale_order_views.xml
carlos-lopez-tecnativa: @LuciaMarchalFactorLibre please fix the `attrs`; this must be evaluated as an `OR (|),` as suggested by @yostashiro.
This is the equivalent of the code in V17.

sale_pricelist_global_rule/views/sale_order_views.xml
carlos-lopez-tecnativa: the same here

carlos-lopez-tecnativa (changes requested):
@LuciaMarchalFactorLibre could you please apply these changes so we can proceed with the merge?

sale_pricelist_global_rule/models/sale_order.py
carlos-lopez-tecnativa: I don't understand your change, the current logic looks fine to me.
The intention is to activate the field when any sale line has a product with a global pricelist that can be applied.
Note that the `_is_applicable_for` function handles this logic:
 https://github.com/factorlibre/sale-workflow/blob/5d1e29774a3986653a456c72e2c69581ec6246be/sale_pricelist_global_rule/models/product_pricelist.py#L158C42-L163

lmchlfl:
@carlos-lopez-tecnativa I have already applied all the changes

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+            by_template: {product.template: qty},
+            by_categ: {product.category: qty}}
+        }
+        """
+        self.ensure_one()
+        qty_data = {
+            "by_template": {},
+            "by_categ": {},
+        }
+        for line in self.order_line.filtered("product_id"):
```
carlos-lopez-tecnativa, line 22: This is not working for me. I tested it in V17 and it works fine, but in V16 it doesn't work.
**Configuration in V16:**

- Set a minimum quantity = 12
- Set a product template with variants (like Customizable Desk, which has a sale price of 750)

image

I created a new sale order and selected the product, set quantity = 6, and the pricelist is applied incorrectly.
I tested with quantity = 5 and the pricelist is not applied.
I suspect that when Odoo iterates over the lines, there's a problem with onchange or NewId causing the order lines not to accumulate the quantities correctly.
Please review.
image


**V17 (same configuration):**
image

The sale order computes the price correctly with 6 units (750).
image


When I add 12 or more units in a single line, the pricelist is applied.
image

Or, if I add several lines with the same product template, the accumulated quantity applies the pricelist when I click the respective button.
image

carlos-lopez-tecnativa (changes requested):

DucTruongKomit (approved):
Everything seems ok

rousseldenis:
@LuciaMarchalFactorLibre @DucTruongKomit It seems this still behaves incorrectly.

Tested on runboat. If you lower a quantity after having a price discount, the original price is not set

DucTruongKomit:
> boat. If you lower a quantity after having a price discount, the original price is not set

Hello @rousseldenis ! Sorry for reply so late!
Did you click Recompute Pricelist Global button after changing the quantity?

sale_pricelist_global_rule/models/sale_order.py
DucTruongKomit: I tested on v16 on your case, it works for me even do not need to click recompute button. Could you check again please?
image

image

http://oca-sale-workflow-16-0-pr3829-159090c77cc8.runboat.odoo-community.org/web#cids=1&menu_id=178&action=297&model=sale.order&view_type=form&id=22

DucTruongKomit:
@LuciaMarchalFactorLibre Please help to solve review.

> > boat. If you lower a quantity after having a price discount, the original price is not set
> 
> Hello @rousseldenis ! Sorry for reply so late! Did you click Recompute Pricelist Global button after changing the quantity?

Is everything okay?

> > boat. If you lower a quantity after having a price discount, the original price is not set
> 
> Hello @rousseldenis ! Sorry for reply so late! Did you click Recompute Pricelist Global button after changing the quantity?



@rousseldenis Could you share your test case?

Here is mine:

Category:

Cat A
has parent category 

Product

Prod A | Cat A
Prod B | Cat A
Prod C | Cat A
Prod D | Cat A
Prod E  | Cat A


image

image

sale_pricelist_global_rule/models/sale_order.py
carlos-lopez-tecnativa: Please rebase the branch to include the latest commits so that Runboat can be recreated and tested again in this PR. This URL is related to another PR, and I don’t know if it includes these changes.

DucTruongKomit: @LuciaMarchalFactorLibre Could you help to squash into 1 commit only, please? Then I can help to test again.

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+        for sale in self:
+            if not sale.pricelist_id:
+                sale.has_pricelist_global = False
+                continue
+            qty_data = self._get_cummulative_quantity()
+            pricelist = sale.pricelist_id.with_context(
+                pricelist_global_cummulative_quantity=qty_data
+            )
+            suitable_rule = self.env["product.pricelist.item"]
+            for line in sale.order_line:
```
nobuQuartile, line 49: Thank you for your contribution.

```suggestion
            for line in sale.order_line.filtered("product_id"):
```

If a sales order line without a product_id (section, memo, etc) is passed to _get_product_rule, it causes _compute_price_rule to return {}, which then triggers an error indicating that the product.id key is missing.



_get_product_rule in odoo/addons
https://github.com/odoo/odoo/blob/c27d978ade9bcbea056933d8fb8b5a924e983bde/addons/product/models/product_pricelist.py#L110-L111

alexey-pelykh (changes requested):
Thanks for the migration work and the thorough test suite.

I went through the current state of the diff and there are a few things that still need attention beyond what previous reviewers already flagged:

**Bug in `_compute_has_pricelist_global`** (`sale_order.py`): Inside the `for sale in self` loop, line 38 calls `self._get_cummulative_quantity()` instead of `sale._get_cummulative_quantity()`. Since `_get_cummulative_quantity` calls `ensure_one()`, this will crash on multi-record recordsets and even on single records it computes against the wrong reference when called from a batch.

**Section/note lines crash** (`sale_order.py`): As @nobuQuartile already pointed out, `_compute_has_pricelist_global` iterates `sale.order_line` without filtering out lines that have no `product_id` (sections, notes). Passing an empty product to `_get_product_rule` will blow up. The same `.filtered("product_id")` guard used in `_get_cummulative_quantity` should be applied here too.

**Blanket discount reset** (`sale_order.py`): `button_compute_pricelist_global_rule` does `self.order_line.write({"discount": 0.0})` before recomputing. This nukes any discount that came from other sources (manual adjustments, standard pricelist rules, other modules). It should probably only reset discount on lines that were previously affected by a global rule, or better yet, let `_compute_discount` handle the reset naturally.

**POT file needs regeneration**: The `.pot` still says `Project-Id-Version: Odoo Server 15.0` and the XML IDs for selection values reference the old v15 keys (`__4_global_product_template`, `__5_global_product_category`) rather than the current `3_1_` / `3_2_` prefixed values. Running `oca-gen-addon-readme` and exporting translations fresh on 16.0 should fix both.

The reported functional issue from @rousseldenis (prices not reverting when quantity drops below threshold) is likely a consequence of the blanket discount reset combined with the `self` vs `sale` bug above -- worth verifying after fixing those.

PR #3534, opened 2025-01-13
[17.0][IMP]sale_quotation_number: Boolean to determine whether quotation sequence has been used

Tisho99:
FW Port of #2719

The current fix doesn't work because it stops working if you use formulas in the prefixes of the succesors.

T-7243


Tisho99:
FW port not neccesary because it has already been fixed.

https://github.com/OCA/sale-workflow/blob/b001b66630989ab81b298b43e492245c164becb8/sale_quotation_number/models/sale_order.py#L51

Tisho99:
@ValentinVinagre @HaraldPanten @manuelregidor

ValentinVinagre (approved):
LGTM 👍🏻

manuelregidor (approved):
Technical review LGTM

HaraldPanten (approved):
LGTM

PR #3545, opened 2025-01-26
[17.0][IMP] sale_force_invoiced: recompute invoice amounts when changing force boolean

kevinkhao:


sale_force_invoiced/tests/test_sale_force_invoiced.py
```diff
@@ -80,8 +89,17 @@ def test_sales_order(self):
...
+        )
         so.force_invoiced = False
         self.assertEqual(
             so.invoice_status, "to invoice", "The invoice status should be To Invoice"
         )
+        self.assertEqual(
+            so.amount_to_invoice,
+            5.0,
+            "Amount to invoice when invoice is not forced should be original amount",
+        )
```
StefanRijnhart, line 107: This test fails.
```
2025-01-26T13:25:41.3342860Z 2025-01-26 13:25:41,333 298 [1;31m[1;49mERROR[0m odoo odoo.addons.sale_force_invoiced.tests.test_sale_force_invoiced: FAIL: TestSaleForceInvoiced.test_sales_order
2025-01-26T13:25:41.3343574Z Traceback (most recent call last):
2025-01-26T13:25:41.3344183Z   File "/__w/sale-workflow/sale-workflow/sale_force_invoiced/tests/test_sale_force_invoiced.py", line 101, in test_sales_order
2025-01-26T13:25:41.3345029Z     self.assertEqual(
2025-01-26T13:25:41.3345466Z AssertionError: 3.45 != 5.0 : Amount to invoice when invoice is not forced should be original amount
2025-01-26T13:25:41.3345930Z  
```

StefanRijnhart (commented):
Makes sense. Is there a reason not to add `force_invoiced` as an @api.depends though? Something like

```
@api.depends("force_invoiced")
def _compute_amount_to_invoice(self):
    return super()._compute_amount_to_invoice()
```

-- edit -- Mmm, given that `_compute_amount_to_invoice` actually uses `invoice_status` but does not depend on it may be meaningful, so may be the override above should just set the amount to zero explicitely for sale orders that are force-invoiced.

StefanRijnhart:
Implemented my own suggestion in the port to 18: https://github.com/OCA/sale-workflow/pull/3593

rousseldenis:
@kevinkhao

StefanRijnhart (approved):
Thanks for the update!

alexey-pelykh (commented):
## Review: sale_force_invoiced - recompute invoice amounts

Nice improvement. The approach is correct: extending `_compute_amount_to_invoice` with `@api.depends("force_invoiced")` ensures the stored computed field is properly recomputed when the force boolean toggles. Calling `super()` first and then overriding for force-invoiced orders is the right pattern, since the base Odoo method's depends does not include `force_invoiced` and compute ordering between `_compute_invoice_status` and `_compute_amount_to_invoice` is not guaranteed.

The test improvements (explicit `price_unit` and `tax_id: False`) make the test deterministic regardless of default taxes configured in the test database -- good practice.

### Observations

**Manifest version not bumped** -- The current version is `17.0.1.1.1`. Per OCA versioning conventions, an `[IMP]` change should bump the version to `17.0.1.2.0`. This is a minor point since the OCA bot typically handles the version bump at merge time, but worth noting.

**No `state == "sale"` filter in `_compute_amount_to_invoice`** -- The existing `_compute_invoice_status` override filters with `so.force_invoiced and so.state == "sale"`, but the new `_compute_amount_to_invoice` only filters on `"force_invoiced"`. This is functionally fine because: (a) `force_invoiced` is typically only set after `action_lock()` which requires sale state, and (b) setting `amount_to_invoice = 0` on a non-sale force-invoiced order is harmless. But for consistency with the existing pattern, you might consider adding the same state check. This is purely stylistic -- not blocking.

Already approved by @StefanRijnhart. CI is green. Looks good to merge.

rousseldenis:
> ## Review: sale_force_invoiced - recompute invoice amounts
> Nice improvement. The approach is correct: extending `_compute_amount_to_invoice` with `@api.depends("force_invoiced")` ensures the stored computed field is properly recomputed when the force boolean toggles. Calling `super()` first and then overriding for force-invoiced orders is the right pattern, since the base Odoo method's depends does not include `force_invoiced` and compute ordering between `_compute_invoice_status` and `_compute_amount_to_invoice` is not guaranteed.
> 
> The test improvements (explicit `price_unit` and `tax_id: False`) make the test deterministic regardless of default taxes configured in the test database -- good practice.
> 
> ### Observations
> **Manifest version not bumped** -- The current version is `17.0.1.1.1`. Per OCA versioning conventions, an `[IMP]` change should bump the version to `17.0.1.2.0`. This is a minor point since the OCA bot typically handles the version bump at merge time, but worth noting.
> 
> **No `state == "sale"` filter in `_compute_amount_to_invoice`** -- The existing `_compute_invoice_status` override filters with `so.force_invoiced and so.state == "sale"`, but the new `_compute_amount_to_invoice` only filters on `"force_invoiced"`. This is functionally fine because: (a) `force_invoiced` is typically only set after `action_lock()` which requires sale state, and (b) setting `amount_to_invoice = 0` on a non-sale force-invoiced order is harmless. But for consistency with the existing pattern, you might consider adding the same state check. This is purely stylistic -- not blocking.
> 
> Already approved by @StefanRijnhart. CI is green. Looks good to merge.
> 
> _Review posted via [CorporateHub](https://corporatehub.eu) OCA review campaign_

@alexey-pelykh Could you refrain doing such reviews ?

With this, you don't use github review mechanism as each remark should be linked to a particular line of the PR. This is time consuming for other reviewers and for the author of the PR.

Many thanks

PR #3547, opened 2025-01-27
[FIX] [16.0] sale_order_line_price_history: Add button to kanban view

Shide:
Add button also to kanban view (same as tree + class) of sale order lines.

The red squares are from other modules.
image

MT-8825 @moduon @rafaelbn @yajo @ernestotejeda @CarlosRoca13 please review if you want! 😄

PR #3549, opened 2025-01-28
[18.0][MIG] sale_order_line_price_history: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT52961

ping @carlos-lopez-tecnativa @sergio-teruel 

Added too the change set on https://github.com/OCA/sale-workflow/pull/3547


sale_order_line_price_history/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+        "wizards/sale_order_line_price_history.xml",
+        "views/sale_views.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "sale_order_line_price_history/static/src/js/*.js",
+            "sale_order_line_price_history/static/src/xml/*.xml",
+        ],
+    },
+    "maintainers": ["ernestotejeda", "CarlosRoca13", "Shide"],
```
pedrobaeza: Finally remove Ernesto from here.

CarlosRoca13: Done!

sale_order_line_price_history/wizards/sale_order_line_price_history.py
```diff
@@ -0,0 +1,156 @@
...
+        This method is invoke by action_set_price method in this model.
+        """
+        self.ensure_one()
+        return {"price_unit": self.price_unit, "discount": self.discount}
+
+    def action_set_price(self):
+        if self.history_sale_order_line_id and self.sale_order_line_id:
+            self.history_sale_order_line_id.write(
+                self._prepare_set_price_history_vals()
+            )
```
carlos-lopez-tecnativa, line 152: From my point of view, changing the price should be forbidden if the sales order is `invoiced`. Odoo does not allow price changes in this case, but this wizard does. What do you think?

carlos-lopez-tecnativa (commented):
Tested functionally and working. I left a comment that is not related to the migration but rather to functionality.

sale_order_line_price_history/wizards/sale_order_line_price_history.py
CarlosRoca13: I have set a condition to hide the button for opening the wizard in case the invoiced amount is greater than 0, just as Odoo does.

CarlosRoca13: I have set a condition to hide the button for opening the wizard in case the invoiced amount is greater than 0, just as Odoo does.

sale_order_line_price_history/wizards/sale_order_line_price_history.py
```diff
@@ -113,7 +113,6 @@ class SaleOrderLinePriceHistoryline(models.TransientModel):
     )
     sale_order_line_id = fields.Many2one(
         comodel_name="sale.order.line",
-        string="Sale order line",
```
rousseldenis: Why removing this as this will generate 'Sale Order Line Id' as base string for translations?

pedrobaeza: In fact it will just generate `'Sale Order Line'` (with upper), and I think it's because a pylint linter bug that didn't take into account upper and lower, saying that "it's redundant". It's fixed, but maybe the version with the fix is not yet here.

CarlosRoca13: I remove it because I had problems with `target_sale_order_line_id` label

sergio-teruel (changes requested):
Hi @CarlosRoca13 Can we recover the space of the column??
image

CarlosRoca13:
> Hi @CarlosRoca13 Can we recover the space of the column?? image

I think it has to do with how Odoo distributes the column sizes. In this example, the gap doesn't look as exaggerated:
image

Anyway, I will see if it's possible to limit the size in some way.

CarlosRoca13:
@sergio-teruel done :smile:

sergio-teruel (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #3552, opened 2025-01-29
[16.0] [FIX] sale_order_line_price_history: Add discount field only if discount group is enabled

Shide:
If you try to add historic prices with discounts group deactivated, an error raises because discount field is not present on the view (kanban and tree)
This adds a check to see if discount is present on the record fields and add the key-value of discount if exists.

Added discount with the group to the kanban view (when defining fields)

Also added a known issue:
If you add the historic price from the kanban view, the sale order save button doesn't work. You need to go into the line and click save.

MT-8826 @moduon @CarlosRoca13 @yajo please review if you want :)

### DIFF
--- a/sale_order_line_price_history/README.rst
+++ b/sale_order_line_price_history/README.rst
@@ -59,6 +59,9 @@ Known issues / Roadmap
 ======================
 
 * A backend tour would be nice to have.
+* When you are in Kanban View, you can set an historic price from the wizard,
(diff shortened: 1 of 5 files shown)


yajo (approved):

Shide:
@CarlosRoca13 please check this PR! 😄

rousseldenis (approved):

rafaelbn (approved):
/ocabot merge patch

PR #3437, opened 2024-11-27
[17.0][FWD][IMP] sale_stock_picking_blocking: allow lock state on sale order

AaronHForgeFlow:
FW port of https://github.com/OCA/sale-workflow/pull/2969/

PR #3553, opened 2025-01-29
[MIG] sale_stock_picking_blocking: Migration to 18.0

absal-smile:


sale_stock_picking_blocking/views/sale_stock_picking_blocking_reason_view.xml
```diff
@@ -4,14 +4,14 @@
 <!--License
 AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).-->
 <odoo>
-    <record id="sale_stock_picking_blocking_reason_tree_view" model="ir.ui.view">
-        <field name="name">sale.delivery.block.reason.tree</field>
+    <record id="sale_stock_picking_blocking_reason_list_view" model="ir.ui.view">
```
sebalix: Updating the XML-ID is not a good idea IMO, this will force a lot of XML-IDs mapping during migration if we do that in all modules.

Is there any guidance / best practice regarding this? cc @OCA/crm-sales-marketing-maintainers

absal-smile:
@sebalix  how can I run the checks?

StefanRijnhart:
@absal-smile Checks approved.

sale_stock_picking_blocking/views/sale_stock_picking_blocking_reason_view.xml
StefanRijnhart: ~~Views can usually be renamed without a problem. They don't need to be mapped because they are not noupdate and they are not going to be referred to in other noupdate or custom data. Upon renaming, they are simply replaced by a new record. It's just that any inheriting views will have to adapt, but the impact of that for views from OCA modules is limited. I'm sure that at some point, this `tree` terminology will seem really obsolete and it's going to be a guideline to weed it out, so may as well do it now.~~

sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
```diff
@@ -2,10 +2,10 @@
 #   (http://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 from odoo.exceptions import ValidationError
-from odoo.tests import Form, common
+from odoo.tests import Form, TransactionCase
```
StefanRijnhart, line 5: While you are at it, inherit from `odoo.addons.base.tests.common.BaseCommon` to circumvent tracking for better performance.

pedrobaeza:
@StefanRijnhart please don't advise to change XML-IDs of the views. It's not needed, and it can cause problems if manual inherited views have been created. The old view won't be removed on the update process, having conflicts.

In fact, I added a clarification on the migration wiki about this specific thing the other day.

pedrobaeza:
In fact, in OpenUpgrade, when we detect a clear view XML-ID renaming, we perform the renaming on the scripts for maximum compatibility. It's true that it's an edge case, but let's try to minimize migration troubles.

StefanRijnhart:
@pedrobaeza Sure, whatever.

absal-smile:
@StefanRijnhart so what i can do??

StefanRijnhart:
@absal-smile Please keep the same view names as before.

leemannd (approved):
Migration looks good.

Please check already added comments

sebalix (commented):
Hello @absal-smile, you created two migration commits, can you squash them into one please?

JordiMForgeFlow (approved):
code + functional

sale_stock_picking_blocking/tests/test_sale_stock_picking_blocking.py
```diff
@@ -28,15 +28,17 @@ def setUpClass(cls):
...
+            "standard_price": 60.0,
+            "uom_id": cls.env.ref("uom.product_uom_unit").id,
+            "uom_po_id": cls.env.ref("uom.product_uom_unit").id,
         }
-        product = cls.env["product.product"].with_user(cls.user_test).create(prod_dict)
+        cls.product = (
+            cls.env["product.product"].with_user(cls.user_test).create(prod_dict)
+        )
         # Create Sale order:
-        # TODO/TMP:
```
StefanRijnhart: You're removing the TODO, but the sale order is still created with a predefined name rather than it being assigned by a sequence. If we now know why this is necessary, can you document that here?

AaronHForgeFlow:
@absal-smile  Could you please cherry-pick the commit in this PR? https://github.com/OCA/sale-workflow/pull/3437/

bizzappdev (approved):
Code + Functional LGTM

sale_stock_picking_blocking/views/sale_stock_picking_blocking_reason_view.xml
```diff
@@ -5,13 +5,13 @@
 AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).-->
 <odoo>
     <record id="sale_stock_picking_blocking_reason_tree_view" model="ir.ui.view">
-        <field name="name">sale.delivery.block.reason.tree</field>
+        <field name="name">sale.delivery.block.reason.view_list</field>
```
StefanRijnhart: Maybe there was some confusion due to the discussion about guidelines, but please keep the original view xml id.

JordiMForgeFlow:
@absal-smile could you attend @StefanRijnhart comment?

StefanRijnhart (approved):
/ocabot merge nobump

PR #3555, opened 2025-01-30
[16.0][FIX]sale_delivery_state: add state to silence warning

PicchiSeba:
Both `sale_delivery_state` and `sale_stock` add a `delivery_status` field. The two modules add the same selection options except for one, missing in `sale_delivery_state`. This is the "started" option.

With both modules installed we are witnessing a lot of warnings like the following

```
sale.order.delivery_status: selection=[('pending', 'Not Delivered'), ('started', 'Started'), ('partial', 'Partially Delivered'), ('full', 'Fully Delivered')] overrides existing selection; use selection_add instead 
```

With this PR we add the same option as `sale_stock` but we ignore it during the computation.

### DIFF
--- a/sale_delivery_state/models/sale_order.py
+++ b/sale_delivery_state/models/sale_order.py
@@ -13,6 +13,7 @@ class SaleOrder(models.Model):
     delivery_status = fields.Selection(
         [
             ("pending", "Not Delivered"),
+            ("started", "Started"),  # Odoo delivery_status compat, unused
             ("partial", "Partially Delivered"),
             ("full", "Fully Delivered"),
         ],
--- a/sale_delivery_state/models/sale_order.py
+++ b/sale_delivery_state/models/sale_order.py
@@ -14,6 +14,7 @@ class SaleOrder(models.Model):
     delivery_status = fields.Selection(
         [
             ("pending", "Not Delivered"),
+            ("started", "Started"),  # Odoo delivery_status compat, unused
             ("partial", "Partially Delivered"),
             ("full", "Fully Delivered"),
         ],


PicchiSeba:
**question**: given that both modules add the same field how do we ensure the execution of the right compute function when both of them are installed?

alexey-pelykh (commented):
## Review: sale_delivery_state selection compatibility fix

**Approach**: Adding the `started` selection option to align with Odoo core's `sale_stock` field definition is a pragmatic way to silence the override warning. The inline comment documenting that the value is unused by this module's compute is appreciated.

### Observations

**1. Manifest version not bumped**

The `__manifest__.py` still shows `"version": "16.0.2.0.1"`. Per OCA guidelines, any change -- even a fix -- should bump the version (at minimum the last segment, e.g., `16.0.2.0.2`).

**2. Tree view decoration gap**

The tree view (`view_order_tree_inherit_delivery_status`) defines `decoration-info`, `decoration-warning`, and `decoration-success` for `pending`, `partial`, and `full` respectively, but has no decoration for `started`. If `sale_stock`'s compute ever sets this value on a record, the badge would render without styling. Consider adding a `decoration-info` or `decoration-warning` for `started` to be safe.

**3. Unanswered question about compute method collision**

The PR author raised a very valid question in the comments: when both `sale_delivery_state` and `sale_stock` are installed, both define a compute method for `delivery_status` (`_compute_oca_delivery_status` vs `_compute_delivery_status`). Odoo's last-loaded-wins behavior for compute methods means the effective compute depends on module load order, which is non-deterministic unless explicitly controlled via `depends`. This PR does not address that concern, and the behavior when both modules coexist deserves clarification before merging.

**4. Alternative approach: `selection_add`**

Rather than redefining the full selection list (which is what triggers the warning), the more idiomatic Odoo pattern would be to use `selection_add` on the field definition when extending it. However, since `sale_delivery_state` defines the field independently (not inheriting from `sale_stock`), this approach of aligning the selection values is a reasonable workaround.

### Verdict

The one-line change itself is correct and harmless, but the broader question about compute method collision (point 3) should be addressed -- at minimum documented -- before this is merged. The version bump (point 1) is also needed per OCA conventions.

---

sale_delivery_state/models/sale_order.py
```diff
@@ -13,6 +13,7 @@ class SaleOrder(models.Model):
     delivery_status = fields.Selection(
         [
             ("pending", "Not Delivered"),
+            ("started", "Started"),  # Odoo delivery_status compat, unused
```
alexey-pelykh, line 16: The selection value alignment is correct and the comment is helpful.

Two things to consider:

1. **Version bump needed**: `__manifest__.py` version should be incremented (e.g., `16.0.2.0.2`) per OCA conventions.

2. **Compute method coexistence**: When both `sale_delivery_state` and `sale_stock` are installed, which `compute` method wins for this field? Odoo's MRO means the last class in the inheritance chain takes precedence. Since `_compute_oca_delivery_status` never sets `started`, and `_compute_delivery_status` (from `sale_stock`) never sets the OCA-specific logic, the effective behavior depends on load order. This might deserve a note in the module's README or a test covering the dual-install scenario.

rousseldenis:
@PierrickBrun @sebastienbeau Do you have an advice here ?

HekkiMelody:
The same fix has been applied in v18 here: https://github.com/OCA/sale-workflow/pull/3713

rousseldenis (approved):

rousseldenis:
@HekkiMelody Could you approve ?

HekkiMelody (approved):

PR #3557, opened 2025-01-31
[FIX] sale_order_type: Incompatibility with l10n_latam_check module

maq-adhoc:
This PR corrects an error that occurs in the core's l10n_latam_check module when creating third-party checks, when the sale_order_type module is installed.
 The error comes up when adding the sale_type_id dependency to the _compute_journal_id method of the account.move model.

While adding the dependency is technically correct, it creates a problem due to the relationship between the account.move and account.payment models, both during the creation of payments from the backend and during import.

The modification proposed in this PR solves the problem in the following ways:

The sale_type_id dependency in _compute_journal_id is removed and replaced by the calculation in _search_default_journal. This avoids the conflict in the relationship between account.move and account.payment.

An onchange is added for the order_type_id field in the form. This ensures that the journal is recalculated correctly when the order type is modified.

It is validated that the journal is present in suitable_journal_ids. This prevents the user from selecting a journal to which they do not have access.

Behavior before the PR

- A third-party check payment is created.
 - The l10n_latam_check_issuer_vat field is modified.
- The form is saved.
 - The l10n_latam_check_issuer_vat field is recomputed to its default value, losing the modified value.

Behavior after the PR

- A third-party check payment is created.
- The l10n_latam_check_issuer_vat field is modified.
- The form is saved.
- The l10n_latam_check_issuer_vat field retains the loaded value

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -48,9 +48,15 @@ class AccountMove(models.Model):
             move.invoice_payment_term_id = move.sale_type_id.payment_term_id
         return res
 
-    @api.depends("sale_type_id")
-    def _compute_journal_id(self):
-        res = super()._compute_journal_id()
-        for move in self.filtered("sale_type_id.journal_id"):
-            move.journal_id = move.sale_type_id.journal_id
+    def _search_default_journal(self):
+        res = super()._search_default_journal()
+        sale_type_journal_id = self.sale_type_id.journal_id
+        if sale_type_journal_id and sale_type_journal_id in self.suitable_journal_ids:
+            return sale_type_journal_id
         return res
+
+    @api.onchange("sale_type_id")
+    def _onchange_sale_type_id(self):
+        sale_type_journal_id = self.sale_type_id.journal_id
+        if sale_type_journal_id and sale_type_journal_id in self.suitable_journal_ids:
+            self.journal_id = sale_type_journal_id.id


rov-adhoc:
This PR resolve the same error of the PR https://github.com/OCA/sale-workflow/pull/3159

rov-adhoc (approved):

maq-adhoc:
Hi @rousseldenis I see that you analyzed the problem that @rov-adhoc  raised in the PR.
This is a similar problem and also fixes that issue in a more elegant way.

rov-adhoc (approved):

rov-adhoc:
@rousseldenis Hi Denis! I would like to ask you if you can please take a look on this PR. 
Thanks in advance! Best regards!

sale_order_type/models/account_move.py
```diff
@@ -48,9 +48,15 @@ def _compute_invoice_payment_term_id(self):
...
-    def _compute_journal_id(self):
-        res = super()._compute_journal_id()
-        for move in self.filtered("sale_type_id.journal_id"):
-            move.journal_id = move.sale_type_id.journal_id
+    def _search_default_journal(self):
+        res = super()._search_default_journal()
+        sale_type_journal_id = self.sale_type_id.journal_id
+        if sale_type_journal_id and sale_type_journal_id in self.suitable_journal_ids:
+            return sale_type_journal_id
         return res
```
alexey-pelykh, line 56: This assigns an integer (`.id`) to a `Many2one` field. In Odoo's onchange context, it is more correct and consistent to assign the recordset directly:
```python
self.journal_id = sale_type_journal_id
```
This also makes the code consistent with `_search_default_journal` above, which returns a recordset.

sale_order_type/models/account_move.py
```diff
@@ -48,9 +48,15 @@ def _compute_invoice_payment_term_id(self):
             move.invoice_payment_term_id = move.sale_type_id.payment_term_id
         return res
 
-    @api.depends("sale_type_id")
-    def _compute_journal_id(self):
-        res = super()._compute_journal_id()
-        for move in self.filtered("sale_type_id.journal_id"):
```
alexey-pelykh, line 54: The base `_compute_journal_id` in Odoo 17 only calls `_search_default_journal()` when the current journal type is **not** in the valid journal types:
```python
for record in self.filtered(lambda r: r.journal_id.type not in r._get_valid_journal_types()):
    record.journal_id = record._search_default_journal()
```
This means if the record already has a journal with a valid type (the common case for invoices), this override will never execute, and the sale type's journal preference will be silently ignored for programmatic creation and imports. Only the `@api.onchange` covers the UI path. This is a behavioral regression compared to the original `_compute_journal_id` override which applied in all contexts.

alexey-pelykh (changes requested):
## Review: `sale_order_type` -- Incompatibility with `l10n_latam_check`

Thank you for this contribution addressing the `l10n_latam_check` incompatibility. The root cause analysis in the PR description is clear and well-explained. Moving the journal assignment from a `@api.depends` on `_compute_journal_id` to `_search_default_journal` is a reasonable direction to avoid the unwanted recomputation cascade affecting `account.payment`. However, there are several issues that should be addressed before merging.

### 1. Assignment of `.id` (integer) instead of recordset in onchange

In `_onchange_sale_type_id`, the journal is assigned as:
```python
self.journal_id = sale_type_journal_id.id
```
This assigns an integer to a `Many2one` field. While this may work in some onchange contexts, it is inconsistent with Odoo conventions and with how the same logic works in `_search_default_journal` (which returns a recordset). It should be:
```python
self.journal_id = sale_type_journal_id
```

### 2. Loss of programmatic journal assignment

The original `_compute_journal_id` override with `@api.depends("sale_type_id")` ensured that the journal was set from the sale type in **all contexts**: UI, programmatic creation, and imports. The new approach splits this into:
- `_search_default_journal`: only called by `_compute_journal_id` when the current journal type is not in the valid journal types (i.e., effectively only on record creation or move type change)
- `@api.onchange`: only fires in the web UI form context

This means that if an invoice is created programmatically (e.g., from a sale order confirmation, or via `create()` with a `sale_type_id` value), and the default journal already has a valid type, `_search_default_journal` will not be called -- the sale type's journal preference will be silently ignored. This is a **behavioral regression** for non-UI flows.

Consider whether a lighter-touch `_compute_journal_id` override could be retained that avoids the `l10n_latam_check` conflict while still covering programmatic flows, or document clearly that this is an intentional trade-off.

### 3. No tests

This change modifies the journal selection mechanism, which is core to invoice creation correctness. Tests should cover at minimum:
- Invoice created via UI with a sale type that has a journal -- journal is applied
- Invoice created programmatically with a sale type that has a journal -- journal is applied
- Journal from sale type is validated against `suitable_journal_ids`
- Compatibility scenario: creating a payment does not trigger unwanted recomputation of `l10n_latam_check` fields

### 4. Minor: `_onchange` vs `@api.onchange` in Odoo 17

Odoo 17 is progressively moving from `@api.onchange` to computed fields with `store=False` or `@api.depends` for reactivity. While `@api.onchange` still works, the direction of the framework is moving away from it. This is not blocking, but worth considering for future compatibility.

---

PR #3554, opened 2025-01-29
[18.0][MIG] sale_order_product_assortment

bosd:
Standard migration using odoo-module-migrator /ocabot migration sale_order_product_assortment .
depens on: https://github.com/OCA/product-attribute/pull/1864

and: https://github.com/OCA/sale-workflow/pull/3559

PR #3559, opened 2025-02-02
[FIX][18.0] Fix CI tests sale_order_type: Run test post_install only

bosd:


StefanRijnhart (commented):
For my information, how does this solve the error in #3554? Isn't `move_type` the field on `account.move.line` rather than `product.product`?

```
   File "/__w/sale-workflow/sale-workflow/sale_order_product_assortment/models/sale_order.py", line 35, in _compute_product_assortment_ids
    products = self.env["product.product"].search(product_domain)
  File "/opt/odoo/odoo/models.py", line 1717, in search
    return self.search_fetch(domain, [], offset=offset, limit=limit, order=order)
  File "/opt/odoo/odoo/models.py", line 1741, in search_fetch
    query = self._search(domain, offset=offset, limit=limit, order=order or self._order)
  File "/opt/odoo/addons/product/models/product_product.py", line 481, in _search
    return super()._search(domain, offset, limit, order)
  File "/opt/odoo/odoo/models.py", line 5754, in _search
    query = self._where_calc(domain)
  File "/opt/odoo/odoo/models.py", line 5505, in _where_calc
    return expression.expression(domain, self).query
  File "/opt/odoo/odoo/osv/expression.py", line 781, in __init__
    self.expression = domain_combine_anies(domain, model)
  File "/opt/odoo/odoo/osv/expression.py", line 599, in domain_combine_anies
    domain_any = _anyfy_leaves(domain, model)
  File "/opt/odoo/odoo/osv/expression.py", line 374, in _anyfy_leaves
    raise ValueError(f"Invalid field {model._name}.{path[0]} in leaf {item}")
ValueError: Invalid field product.product.move_type in leaf ('move_type', '=', 'out_refund')
```

bosd:
@StefanRijnhart I'm actually not sure if this would solve the error from #3554.
I've tested #3554 extensively on local machine in different config. Could not reproduce the errors that occured here on github tests.
Then I noticed the error was not coming from `sale_order_product_assortment ` but from `sale_order_type`
Likely some compatability issue.

(Did not run test locally with both modules installed. As were not using `sale_order_type`)
Should have done that. 🙄 


```
2025-02-02 11:00:20,766 338 ERROR odoo odoo.addons.sale_order_type.tests.test_sale_order_type: ERROR: TestSaleOrderType.test_sequence_default
  File "/__w/sale-workflow/sale-workflow/sale_order_type/tests/test_sale_order_type.py", line 260, in test_sequence_default
    order = self.create_sale_order()
  File "/__w/sale-workflow/sale-workflow/sale_order_type/tests/test_sale_order_type.py", line 135, in create_sale_order
    sale_form.partner_id = partner or self.partner
  File "/opt/odoo/odoo/tests/form.py", line 352, in __setattr__
    self[field_name] = value
```

Assumed that it would good practice to run tests only post install..

StefanRijnhart:
Mmm, if `sale_order_product_assortment`'s `_compute_product_assortment_ids` raises an exception, even if it is somehow related to sale_order_type, then it's probably something to solve in your migration of `sale_order_product_assortment` unless it can be reproduced without `sale_order_product_assortment` as well.

bosd:
I don't know the solution for the sale_order_type module.
Tested it locally. Sometimes tests fail.. Sometimes they pass...
Anyway, the conflicting issue with the other module dissapeared. So closing this one.

PR #1902, opened 2022-02-07
[14.0][FIX] sale_order_line_date: Fix for same behaviour as v13

BernatPForgeFlow:
This fix pretends to follow with the same behaviour of the module as in v13.  Actual changes are the following:
- Add the correct propagation of the scheduled date for stock moves in SO confimation.
- Change test as it was outdated.

After this PR, another one will be created to improve some parts. If we are using sale_order_line_date to determine dates for sale_order it should be computed based on sale_order_line_dates and shipping policy.

PR #3565, opened 2025-02-03
[18.0][FIX] sale_order_line_date: fix setting empty commitment date on stock moves

### LINKED ISSUE #3363: [16.0] sale_order_line_date : removing the date after sale has been confirmed raises TypeError

StefanRijnhart:
Port of https://github.com/OCA/sale-workflow/pull/3364

Fixes #3363:
```
TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType'
```
in stock.move's _set_date_deadline

### DIFF
--- a/sale_order_line_date/models/sale_order_line.py
+++ b/sale_order_line_date/models/sale_order_line.py
@@ -30,9 +30,19 @@ class SaleOrderLine(models.Model):
         return vals
 
     def write(self, vals):
+        # Propagate a new commitment date to pending stock moves
         res = super().write(vals)
         if "commitment_date" in vals:
-            moves = self.move_ids.filtered(lambda m: m.state not in ["cancel", "done"])
-            if moves:
-                moves.write({"date_deadline": vals["commitment_date"]})
+            if vals.get("commitment_date"):
+                self.move_ids.filtered(
+                    lambda sm: sm.state not in ["cancel", "done"]
+                ).write({"date_deadline": vals.get("commitment_date")})
+            else:
+                for line in self:
+                    date_deadline = (
+                        line.order_id.commitment_date or line._expected_date()
+                    )
+                    line.move_ids.filtered(
+                        lambda sm: sm.state not in ["cancel", "done"]
+                    ).write({"date_deadline": date_deadline})
         return res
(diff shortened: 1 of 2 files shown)


sale_order_line_date/models/sale_order_line.py
```diff
@@ -30,9 +30,19 @@ def _prepare_procurement_values(self, group_id=False):
         return vals
 
     def write(self, vals):
+        """Propagate a new commitment date to pending stock moves"""
```
sebalix: It's not encouraged to override the docstring of such method, as it hides the original docstring in some IDE.
Better to put this as comment.

sebalix (approved):
LG, small comment

sale_order_line_date/models/sale_order_line.py
StefanRijnhart: Thanks, fixed!

sebalix:
BTW I spotted also this one done in 14.0: https://github.com/OCA/sale-workflow/pull/1902
It also updates the schedule date, didn't check into the details.

For now we can merge this one as it's coming from 16.0!

/ocabot merge patch

StefanRijnhart:
@sebalix That one looks a lot like https://github.com/OCA/sale-workflow/commit/fd06e8fb8da014e7772916cd30ebc63b68b9d45a which is included in 18.0.

PR #2637, opened 2023-08-11
[16.0][MIG] sale_order_end_user

acsonefho:
Migration of `sale_order_end_user`.
Based on: https://github.com/OCA/sale-workflow/pull/763

Depends on https://github.com/OCA/partner-contact/pull/1561

PR #3567, opened 2025-02-05
[16.0][ADD] sale_end_customer

AungKoKoLin1997:
This module introduces the `end_partner_id `field in both sales and invoices, ensuring its propagation from sales orders to invoices. Additionally, this field enables grouping in sales and invoice analysis.

@qrtl QT5158


nobuQuartile (approved):
I have completed the functional test and read the code.
LGTM

nobuQuartile (approved):

rousseldenis:
@AungKoKoLin1997 @nobuQuartile There was already a module for this functionality

https://github.com/OCA/sale-workflow/pull/2637

AungKoKoLin1997:
> There was already a module for this functionality

@rousseldenis 
As explained by @yostashiro , coexistence of both modules would be acceptable if the need arises, so I will keep this PR open.
https://github.com/OCA/sale-workflow/pull/2637#issuecomment-3248232444

sale_end_customer/reports/account_invoice_report.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "account.invoice.report"
+
+    partner_end_customer_id = fields.Many2one("res.partner", string="End Customer")
+
+    def _select(self):
+        select_str = super()._select()
+        select_str += """
+            , move.partner_end_customer_id as partner_end_customer_id
+            """
+        return select_str
```
alexey-pelykh, line 17: The `_select()` is overridden to add `partner_end_customer_id` to the SELECT clause, but there is no corresponding `_group_by()` override. Compare with `sale_report.py` which correctly overrides `_group_by_sale()`.

Without this, grouping by "End Customer" in the Invoices Analysis report will produce a SQL error (`partner_end_customer_id` in SELECT but not in GROUP BY).

Please add a `_group_by()` override, e.g.:

```python
def _group_by(self):
    res = super()._group_by()
    res += ", move.partner_end_customer_id"
    return res
```

sale_end_customer/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_sale_end_partner
```
alexey-pelykh, line 1: Naming inconsistency: the module is `sale_end_customer` but the test file is `test_sale_end_partner`. For consistency, consider renaming to `test_sale_end_customer` / `TestSaleEndCustomer`.

alexey-pelykh (changes requested):
## Review: `sale_end_customer` module

Thanks for the contribution. The module addresses a real use case (dealer/reseller flows), and the overall structure follows OCA conventions well. However, there are a few issues that should be addressed before merge.

### Issues

**1. Missing `_group_by` override in `account_invoice_report.py` (bug)**

The `_select()` method is overridden to include `partner_end_customer_id` in the SELECT clause, but there is no corresponding `_group_by()` override. The `sale_report.py` correctly overrides `_group_by_sale()`, but the same treatment is missing for the invoice report. This will cause SQL errors when users try to group by "End Customer" in the invoice analysis report, since the column is in SELECT but not in GROUP BY.

**2. Naming inconsistency: test file**

The module is named `sale_end_customer`, but the test file is `test_sale_end_partner.py` with class `TestSaleEndPartner`. For consistency with the module name, this should be `test_sale_end_customer.py` / `TestSaleEndCustomer`.

### Suggestions

- Consider adding a test for the case where `partner_end_customer_id` is empty on the SO, to confirm `False` (or `0`) is propagated cleanly to the invoice.
- Consider adding a test that validates the report grouping actually works (a quick read on `sale.report` and `account.invoice.report` with the end customer group-by).
- The `.pot` file under `i18n/` is absent. OCA modules typically include a generated `.pot` file. This can be generated with `oca-gen-addon-readme` / the OCA maintainer tools.
- The field does not set `copy=False`. When duplicating a sale order, the end customer will be copied. This may be intentional, but it is worth confirming.

Overall the module is clean and well-structured. Fixing the `_group_by` issue in the invoice report is the main blocker.

PR #3568, opened 2025-02-05
[17.0][MIG] sale_readonly_security: Migration to 17.0

victoralmau:
Migration to 17.0

Please @carlos-lopez-tecnativa and @chienandalu can you review it?

@Tecnativa TT54932


carlos-lopez-tecnativa (approved):
LGTM

sale_readonly_security/views/sale_order_views.xml
```diff
@@ -5,6 +5,12 @@
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
+            <field name="partner_id" position="after">
```
pedrobaeza: Seeing this change, this module is going to be very fragile if other fields in extra modules are added to the header and they are the same used in other places, so the strategy should switch to hide the header, but not remove it.

pedrobaeza (changes requested):

sale_readonly_security/views/sale_order_views.xml
victoralmau: Ok, changed to apply invisible when necessary.

pedrobaeza (approved):
Code and functional review

/ocabot merge nobump

PR #3569, opened 2025-02-05
[FIX] sale_order_type: Fix item search in account move view

stevTresCloud:
- In case there is a modification of the view, with studio or a custom development by adding a new tab before the main invoice lines tab, an error occurs with this search because it doesn't find the item in the xpath expression. With this modification, we ensure that the item to search for is in the main invoice lines tab.

### DIFF
--- a/sale_order_type/views/account_move_views.xml
+++ b/sale_order_type/views/account_move_views.xml
@@ -11,7 +11,10 @@
                     readonly="state != 'draft'"
                 />
             </label>
-            <xpath expr="//notebook//list//field[@name='sequence']" position="before">
+            <xpath
+                expr="//notebook//page[@id='invoice_tab']//list//field[@name='sequence']"
+                position="before"
+            >
                 <field name="account_id" column_invisible="True" />
             </xpath>
         </field>


sale_order_type/views/account_move_views.xml
```diff
@@ -11,7 +11,10 @@
                     readonly="state != 'draft'"
                 />
             </label>
-            <xpath expr="//notebook//list//field[@name='sequence']" position="before">
+            <xpath
+                expr="//notebook//page[@id='invoice_tab']//list//field[@name='sequence']"
+                position="before"
+            >
```
celm1990, line 17: It can be more specific and simplified to avoid ambiguity, as follows:
```suggestion
            <xpath
                expr="//field[@name='invoice_line_ids']//list//field[@name='sequence']"
                position="before"
            >
```
I think this field might be unnecessary according to this PR: https://github.com/odoo/odoo/pull/137031. The fields used in other fields/attributes are automatically added as invisible.

@pedrobaeza @StefanRijnhart What do you think?

celm1990 (changes requested):

sale_order_type/views/account_move_views.xml
pedrobaeza: Yeah, that seems so, so you can remove it directly.

alexey-pelykh (commented):
## Review: `sale_order_type` -- XPath fix for account move view

**Module:** `sale_order_type` (v18.0.1.1.1)
**Change:** Makes the XPath expression in `account_move_views.xml` more specific by adding `page[@id='invoice_tab']` to the path, preventing errors when other modules/Studio customizations insert tabs before the invoice lines tab.

### Analysis

The problem statement is valid: the current XPath `//notebook//list//field[@name='sequence']` can match the wrong element if another module adds a notebook page with a list containing a `sequence` field before the invoice lines tab. Adding specificity to the XPath is the correct approach.

However, I agree with the feedback already provided by @celm1990 and @pedrobaeza:

**1. XPath specificity:** The expression `//field[@name='invoice_line_ids']//list//field[@name='sequence']` (as suggested by @celm1990) is more robust than `//page[@id='invoice_tab']//list//field[@name='sequence']` because it targets the exact field container (`invoice_line_ids`) rather than relying on the page `id` attribute. If the page `id` ever changes upstream, the field-based expression would still work.

**2. Necessity of the entire XPath block:** Since Odoo 17.3+ (see [odoo/odoo#137031](https://github.com/odoo/odoo/pull/137031)), fields referenced in Python expressions (`invisible`, `column_invisible`, `readonly`, `required`, `context`, `domain`) are automatically added as invisible/readonly to the view. In the Odoo 18.0 `account.view_move_form`, `account_id` is already referenced in `required`, `domain`, and `context` attributes on the invoice lines. This means the explicit `<field name="account_id" column_invisible="True" />` injection via XPath is redundant in 18.0 and the entire XPath block can likely be removed.

As @pedrobaeza confirmed: the block can be removed directly.

### Verdict

The fix direction is correct, but the existing reviewer feedback (from Feb 2025) remains unaddressed. The PR should either:
- **Option A:** Remove the entire XPath block (preferred, since the `account_id` field auto-add makes it unnecessary in 18.0), or
- **Option B:** At minimum, adopt the more specific XPath suggested by @celm1990: `//field[@name='invoice_line_ids']//list//field[@name='sequence']`

Please address the outstanding review comments before this can move forward.

---

PR #3570, opened 2025-02-06
[MIG] sale_partner_address_restrict : Migration to 18.0

rousseldenis:
Comes from : https://github.com/OCA/partner-contact/pull/1925

Improvements:

- Allow to configure the restriction activation (per company)
- Add a domain computed field

PR #3574, opened 2025-02-07
[17.0][MIG] sale_partner_address_restrict

JordiMForgeFlow:
Standard migration moving the module from partner-contact to sale-workflow

Includes improvements from https://github.com/OCA/sale-workflow/pull/3570

Improvements:

- Allow to configure the restriction activation (per company)
- Add a domain computed field


JordiMForgeFlow:
@rousseldenis ready to review :)

AlexPForgeFlow (approved):
Functional and code review :+1:

AlexPForgeFlow:
@JordiMForgeFlow as an improvement, could we consider to restrict also not confirmed sales orders if setting it's enabled, or maybe is it too much restrictive?

JordiMForgeFlow:
@AlexPForgeFlow the restriction is already covering non confirmed orders with the constrain

AlexPForgeFlow:
@JordiMForgeFlow The constrain does not cover that if we have a draft sale order created on the system, and then we enable the setting, not being able to confirm it. I have been able to confirm sale order in that scenario.

JordiMForgeFlow:
@AlexPForgeFlow ah I see. I don't think we should bother about previously created SOs. If they are updated then they will get the error.

AaronHForgeFlow (approved):
Code review LGTM!

rousseldenis (approved):

PR #3576, opened 2025-02-07
[18.0][MIG] sale_order_product_availability_inline: Migration to version 18.0

carlos-lopez-tecnativa:
TT54284
@Tecnativa @pedrobaeza @sergio-teruel @SodexisTeam Could you please review this?


sergio-teruel (changes requested):
Sorry @carlos-lopez-tecnativa 

the available cuantities is not shown!!

image

image

carlos-lopez-tecnativa:
> Sorry @carlos-lopez-tecnativa
> 
> the available cuantities is not shown!!
> 

@sergio-teruel Please take into account that in `sale.order.line`, we have two fields: `product_id` and `product_template_id`. This module only affects `product_id`. If you add this field to the list view, the quantity is displayed.

I think we need to add support for the other field. What do you think?
image

sergio-teruel:
H.. @carlos-lopez-tecnativa you are right...

sergio-teruel (approved):

youring:
Tested in runboat. 
Open existing demo sale order, product variant not selected in order line, the free_qty will show up beside product_template_id.
image

If create a new quotation, select product variant, select some products, free_qty is still displayed after saving the sale order. 
image
Isn't it meant to be displayed only in the dropdown list? In Odoo 16, it will disappear after saving.

carlos-lopez-tecnativa:
> Tested in runboat. Open existing demo sale order, product variant not selected in order line, the free_qty will show up beside product_template_id. image
> 

@youring sorry, but I don't understand.

I noticed in your image that the **product is a variant**, not a **product template**, because the name includes **attributes like "Custom, White."** Please check again.

In an existing sale order, with the product variant hidden and the product template displayed, the quantity is not shown.

V16
image

V17
image



> If create a new quotation, select product variant, select some products, free_qty is still displayed after saving the sale order. image Isn't it meant to be displayed only in the dropdown list? In Odoo 16, it will disappear after saving.

Regarding your second point:

In Odoo 16, the quantity disappears after saving, which matches your observation.
However, in Odoo 17, the quantity remains displayed after saving.

Since this PR is migrating the module to V18 (using V17 as the base branch), the behavior follows the V17 standard.
If this change in behavior is due to an Odoo update, that would be a separate issue outside the scope of this migration.

youring:
Sorry I didn't test Odoo 17. It seems we should fix in v17 first to keep the functionality from v16.

pedrobaeza:
I don't think the correct behavior is the 16 one. I prefer this "new" one.

youring:
Maybe we could change the name to `sale_order_product_available_search` or `sale_order_product_available_dropdown`, and make a "new" one `sale_order_line_available_inline`.
Anyway, I think the product column is too crowded to carry available quantity inline.

According to the original purpose of the module since v13, I think we should go this way as `stock_picking_product_availability_inline` https://github.com/OCA/stock-logistics-workflow/pull/1847

pedrobaeza:
@sergio-teruel what's your take on this?

sergio-teruel:
Ok.. I think that this module is correct... 
The main difference is that in v18 by default the product_tmpl_id is shown 
image
but this module does not cover that case, at least not without more budget.

If user change to product variant the module runs as expected... 

So I approve this migration

And the units remain the same as in the previous version.
image

pedrobaeza (approved):

dreispt:
@carlos-lopez-tecnativa can you help understanding the test error on PR #3581 please?

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa can you help understanding the test error on PR #3581 please?

@dreispt  I confirm that this commit https://github.com/odoo/odoo/pull/209141 changed the behavior and caused the test to fail.
I tested before this commit and the test worked fine.
Later, I tested again after updating the Odoo code with this commit, and the test failed.

Before this commit, the `name` was displayed as:
image
After this commit, the `name` is displayed as:
image

So, I'm not sure what the best approach is — whether to update the test (change `assertFalse` to `assertTrue`) or report this behavior to Odoo. WDYT?

pedrobaeza:
Section and notes should be excluded from putting such text.

PR #2861, opened 2023-12-18
[14.0][MIG] sale_order_line_cancel backport from 16

TDu:
Related to this 16.0 PR #2357

PR #3584, opened 2025-02-14
[14.0] backport sale_order_line_cancel

mmequignon:
Supersedes https://github.com/OCA/sale-workflow/pull/2861


mt-software-de (approved):
LGTM

mt-software-de:
@jbaudoux could you review and merge?

jbaudoux (approved):

mt-software-de:
@rousseldenis could you merge this one?

rousseldenis:
@mmequignon Could you take this ? https://github.com/OCA/sale-workflow/pull/3666

mt-software-de:
@rousseldenis could we move forward with this?
I would create then backports for the new commits of v16.

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -0,0 +1,78 @@
...
+            moves_to_cancel = line._get_moves_to_cancel()
+            line._check_moves_to_cancel(moves_to_cancel)
+            moves_to_cancel._action_cancel()
+            line.order_id.message_post(
+                body=_(
+                    "<b>%(product)s</b>: The order line has been canceled",
+                    product=line.product_id.display_name,
+                )
+            )
+            return True
```
alexey-pelykh, line 78: Bug: `return True` is inside the `for` loop, so only the first line in the filtered recordset will ever be processed. The remaining lines are silently skipped.

The `return True` should be dedented one level to sit outside the loop:

```python
def cancel_remaining_qty(self):
    lines = self.filtered(lambda l: l.can_cancel_remaining_qty)
    for line in lines:
        moves_to_cancel = line._get_moves_to_cancel()
        line._check_moves_to_cancel(moves_to_cancel)
        moves_to_cancel._action_cancel()
        line.order_id.message_post(
            body=_(
                "<b>%(product)s</b>: The order line has been canceled",
                product=line.product_id.display_name,
            )
        )
    return True
```

While the wizard currently calls this on single lines, this method is public API on `sale.order.line` and should handle recordsets correctly.

sale_order_line_cancel/wizards/sale_order_line_cancel.py
```diff
@@ -0,0 +1,27 @@
...
+class SaleOrderLineCancel(models.TransientModel):
+    _name = "sale.order.line.cancel"
+    _description = "Cancel Remaining Wizard"
+
+    def _get_sale_order_line(self):
+        active_id = self.env.context.get("active_id")
+        active_model = self.env.context.get("active_model")
+        if not active_id or active_model != "sale.order.line":
+            raise UserError(_("No sale order line ID found"))
+        return self.env[active_model].browse(active_id)
```
alexey-pelykh, line 18: Dead code: `_get_moves_to_cancel` is defined here but never called. The wizard's `cancel_remaining_qty` delegates to `line.cancel_remaining_qty()` which uses the model's own `_get_moves_to_cancel()`. Consider removing this unused method.

sale_order_line_cancel/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "summary": """Sale cancel remaining""",
+    "depends": ["sale_stock"],
+    "data": [
+        "security/sale_order_line_cancel.xml",
+        "wizards/sale_order_line_cancel.xml",
+        "views/sale_order.xml",
+        "views/sale_order_line.xml",
+    ],
+    "website": "https://github.com/OCA/sale-workflow",
+}
```
alexey-pelykh, line 21: Minor: OCA convention for 14.0 modules is to include `"installable": True` explicitly in the manifest.

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+
+    @api.depends(
+        "product_uom_qty",
+        "qty_delivered",
+        "product_qty_canceled",
+    )
+    def _compute_product_qty_remains_to_deliver(self):
+        for line in self:
+            qty_remaining = line.qty_to_deliver - line.product_qty_canceled
```
alexey-pelykh, line 48: The stored computation uses `line.qty_to_deliver` which is itself computed (depends on `product_uom_qty`, `qty_delivered`, and `display_type`). Since `display_type` is not listed in `@api.depends`, the stored value could become stale if a line's display type changes. Consider either:
- Adding `display_type` to the depends list, or
- Computing directly from raw fields: `product_uom_qty - qty_delivered - product_qty_canceled`

alexey-pelykh (commented):
## OCA PR Review: `sale_order_line_cancel` backport to 14.0

**Note: Branch 14.0 is end-of-life (EOL).** Odoo 14.0 has reached end of life and is no longer maintained by Odoo SA or OCA. This PR targets an EOL branch. New module backports to EOL branches are unlikely to be merged per OCA policy. Contributors should consider targeting a supported branch (16.0+) instead.

---

That said, here is a technical review of the code for reference:

### Bug: `cancel_remaining_qty` only processes the first line

In `sale_order_line_cancel/models/sale_order_line.py`, the `cancel_remaining_qty` method has a `return True` inside the `for` loop body, which means only the first matching line will ever be processed:

```python
def cancel_remaining_qty(self):
    lines = self.filtered(lambda l: l.can_cancel_remaining_qty)
    for line in lines:
        moves_to_cancel = line._get_moves_to_cancel()
        line._check_moves_to_cancel(moves_to_cancel)
        moves_to_cancel._action_cancel()
        line.order_id.message_post(
            body=_(
                "<b>%(product)s</b>: The order line has been canceled",
                product=line.product_id.display_name,
            )
        )
        return True  # <-- exits after first iteration
```

The `return True` should be moved outside the `for` loop (dedented one level). While the current wizard calls this on a single line at a time, the method is a public API on `sale.order.line` and could be called on a recordset. This is a correctness issue that should be fixed.

### Dead code in wizard

The wizard class `SaleOrderLineCancel` in `wizards/sale_order_line_cancel.py` defines `_get_moves_to_cancel(self, line)` (line ~10-11), but this method is never called. The wizard's `cancel_remaining_qty` delegates directly to `line.cancel_remaining_qty()`, which uses the model's own `_get_moves_to_cancel()`. This dead code should be removed to avoid confusion.

### Missing `installable` key in manifest

The `__manifest__.py` does not include `"installable": True`. While it defaults to `True`, OCA convention for 14.0 modules is to include it explicitly.

### i18n: Version mismatch in .pot/.po headers

Both `i18n/sale_order_line_cancel.pot` and `i18n/it.po` have `Project-Id-Version: Odoo Server 16.0` in their headers. Since this is a 14.0 backport, the version should read `Odoo Server 14.0`. This is minor but indicates the translation templates were copied from the 16.0 version without updating the header.

### Stored compute field dependency completeness

`_compute_product_qty_remains_to_deliver` is a stored computed field that depends on `product_uom_qty`, `qty_delivered`, and `product_qty_canceled`. However, the computation uses `line.qty_to_deliver` which is itself a computed field that also depends on `display_type`. If `display_type` changes (e.g., section/note lines), the stored value might not recompute. Consider adding `qty_to_deliver` or its underlying dependencies to `@api.depends` to be safe, or compute from the raw fields directly (`product_uom_qty - qty_delivered - product_qty_canceled`).

### XML indentation inconsistency

In `views/sale_order.xml`, the indentation style is inconsistent -- some blocks use 2-space indentation while others (inherited from copy-paste) use larger indentation. This is cosmetic but worth normalizing for readability.

---

### Summary

| Finding | Severity |
|---------|----------|
| EOL branch (14.0) | Informational |
| `return True` inside loop -- only first line processed | Medium (bug) |
| Dead `_get_moves_to_cancel` in wizard | Low (dead code) |
| Missing `installable` in manifest | Low |
| .pot/.po version header mismatch (16.0 vs 14.0) | Low |
| Stored compute dependency completeness | Low-Medium |
| XML indentation inconsistency | Cosmetic |

The module is already approved by two reviewers and labeled as such. The primary concern remains that **14.0 is EOL**.

PR #3585, opened 2025-02-19
[17.0][FIX] sale_order_secondary_unit: fix test

pilarvargas-tecnativa:
The behaviour of odoo in v17 when modifying an existing order line is to recalculate the price based on the price set on the product and not on the line. This module itself had an error in its tests as it was checking a price based on the modified price of the line (1000) which was not correct as when modifying any data of the order line the price was recalculated based on the price of the product which by default is 1 if it has not been set.

cc @tecnativa TT54893

@juancarlosonate-tecnativa @pedrobaeza please review

### DIFF
--- a/sale_order_secondary_unit/tests/test_sale_order.py
+++ b/sale_order_secondary_unit/tests/test_sale_order.py
@@ -26,6 +26,7 @@ class TestSaleOrder(TransactionCase):
                 "name": "test",
                 "uom_id": cls.product_uom_kg.id,
                 "uom_po_id": cls.product_uom_kg.id,
+                "list_price": 1000,
             }
         )
         # Set secondary uom on product template
@@ -54,7 +55,6 @@ class TestSaleOrder(TransactionCase):
             with order_form.order_line.new() as line_form:
                 line_form.product_id = cls.product
                 line_form.product_uom_qty = 1
-                line_form.price_unit = 1000.00
         cls.order = order_form.save()
 
     def test_onchange_secondary_uom(self):


juancarlosonate-tecnativa (approved):

pedrobaeza:
I don't get the problem seeing the code. You are not exposing the traceback/failed test before the change, but this also seems to me that is a flight forward putting the sales price in the product for getting always that price in case of recomputation, which doesn't feel right.

pilarvargas-tecnativa:
@pedrobaeza 

The module tests fail when run in isolation:
```
2025-02-19 11:46:21,324 1 ERROR devel odoo.addons.sale_order_secondary_unit.tests.test_sale_order: FAIL: TestSaleOrder.test_secondary_uom_unit_price
  File "/opt/odoo/auto/addons/sale_order_secondary_unit/tests/test_sale_order.py", line 111, in test_secondary_uom_unit_price
    self.assertEqual(self.order.order_line.secondary_uom_unit_price, 500)
AssertionError: 0.5 != 500
```
The main reason for this failure is that the test ‘test_secondary_uom_unit_price’ is not creating a new order line, it is modifying a line that already exists in the order created. 
The product that has been created for the tests does not have a defined price, so its price will be 1 by default.
The price_unit field in the order line is a calculated field, so modifications to the existing order line cause this data to be recalculated:
https://github.com/odoo/odoo/blob/17.0/addons/sale/models/sale_order_line.py#L463
Hence the result is erroneous as it takes ‘1’ as the price and not ‘1000’ as set in the order line. In this way it calculates that the secondary price is ‘0.5’ and not ‘500’. 
The calculation of the price when modifying the line is the basic behaviour of odoo and this module does not modify that behaviour, it only adds the secondary units to an order.
For this reason one of the solutions is to set the price of the product, another solution would be to re-set the price of the line when it is modified in the test ‘test_secondary_uom_unit_price’.

Example in an odoo runbot:


https://github.com/user-attachments/assets/17f70927-3cf6-4a03-929b-cd0a091c6d2b

pedrobaeza (approved):
OK, maybe it's better to explicitly put that price on the affected test, as this is still relying on the fallback using product list price, but let's continue this way for now.

/ocabot merge patch

PR #2493, opened 2023-04-26
14.0 fix sale automatic workflow

ChrisOForgeFlow:
Reference: https://github.com/OCA/sale-workflow/pull/1514

PR #3587, opened 2025-02-21
[14.0] sale_automatic_workflow: fixes and adaptations

AlexPForgeFlow:
Reference: https://github.com/OCA/sale-workflow/pull/2493

### DIFF
--- a/sale_automatic_workflow/__manifest__.py
+++ b/sale_automatic_workflow/__manifest__.py
@@ -19,6 +19,7 @@
         "security/ir.model.access.csv",
         "views/sale_view.xml",
         "views/sale_workflow_process_view.xml",
+        "views/res_partner_view.xml",
         "data/automatic_workflow_data.xml",
     ],
 }
--- a/sale_automatic_workflow/models/__init__.py
+++ b/sale_automatic_workflow/models/__init__.py
@@ -4,3 +4,4 @@ from . import sale_order
 from . import sale_workflow_process
 from . import stock_move
 from . import stock_picking
+from . import res_partner
--- a/sale_automatic_workflow/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow/models/automatic_workflow_job.py
@@ -1,6 +1,6 @@
 # Copyright 2011 Akretion Sébastien BEAU <sebastien.beau@akretion.com>
 # Copyright 2013 Camptocamp SA (author: Guewen Baconnier)
-# Copyright 2016 Sodexis
+# Copyright 2016-21 Sodexis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
@@ -13,16 +13,19 @@ _logger = logging.getLogger(__name__)
 
 
 @contextmanager
-def savepoint(cr):
+def savepoint(cr, autocommit=False):
     """Open a savepoint on the cursor, then yield.
 
     Warning: using this method, the exceptions are logged then discarded.
     """
     try:
-        with cr.savepoint():
-            yield
+        yield
     except Exception:
         _logger.exception("Error during an automatic workflow action.")
+        cr.rollback()
+    else:
+        if autocommit:
+            cr.commit()
 
 
 class AutomaticWorkflowJob(models.Model):
@@ -50,12 +53,12 @@ class AutomaticWorkflowJob(models.Model):
         sale._send_order_confirmation_mail()
 
     @api.model
-    def _validate_sale_orders(self, order_filter):
+    def _validate_sale_orders(self, order_filter, auto_commit=False, limit=None):
         sale_obj = self.env["sale.order"]
-        sales = sale_obj.search(order_filter)
+        sales = sale_obj.search(order_filter, limit=limit)
         _logger.debug("Sale Orders to validate: %s", sales.ids)
         for sale in sales:
-            with savepoint(self.env.cr):
+            with savepoint(self.env.cr, auto_commit):
                 self._do_validate_sale_order(
                     sale.with_company(sale.company_id),
                     order_filter,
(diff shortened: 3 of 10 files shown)


rousseldenis:
@AlexPForgeFlow Thanks.

IMHO, as commits have not been merged in other PR's, commit messages should be amended to respect conventions like `[IMP] sale_automatic_workflow: (...)`.

rousseldenis:
@sebastienbeau @gurneyalex @grindtildeath Could you review ?

AlexPForgeFlow:
@rousseldenis Dou you think we can keep this two commits as are modifying different modules, or you want me to squash them too?

sale_automatic_workflow/__manifest__.py
```diff
@@ -5,20 +5,21 @@
...
+    "version": "14.0.1.2.0",
     "category": "Sales Management",
     "license": "AGPL-3",
     "author": "Akretion, "
     "Camptocamp, "
     "Sodexis, "
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
-    "depends": ["sale_stock", "sales_team"],
+    "depends": ["sale_stock", "sales_team", "account_payment_partner"],
```
grindtildeath: IMO That's a no go. We definitely don't want to pull account_payment_partner as a dependency of sale automatic workflow.
There' the module sale_automatic_workflow_payment_mode that can be used for such things.

grindtildeath (changes requested):

sale_automatic_workflow/__manifest__.py
rousseldenis: That's why I pinged 😅 

@AlexPForgeFlow could you adapt this to limit the modifications and remove new dependency?

AlexPForgeFlow: @rousseldenis @grindtildeath You were right was not necessary dependency

sale_automatic_workflow/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Sale Automatic Workflow",
-    "version": "14.0.1.4.0",
+    "version": "14.0.1.2.0",
```
rousseldenis: You should not modify this (if you don't provide migration scripts) as it will be updated by merge bot.

AlexPForgeFlow: @rousseldenis done!

alexey-pelykh (commented):
## OCA Review -- `sale_automatic_workflow` fixes and adaptations

> **Note:** This PR targets the **14.0** branch, which has reached End of Life. New feature work should target a supported branch (16.0+). Posting as COMMENT rather than REQUEST_CHANGES given the EOL status.

This PR backports several enhancements from PR #2493 into 14.0:
- `auto_commit` and `search_limit` fields on workflow process
- `default` flag on workflow process with `default_get` on sale order
- `workflow_process_id` on `res.partner` with `onchange` propagation to sale orders
- Savepoint context manager rewrite (removes `cr.savepoint()`, adds manual `cr.rollback()`/`cr.commit()`)

---

### Critical Issues

**1. `cr.commit()` / `cr.rollback()` bypasses ORM transaction management**

The `savepoint()` context manager was rewritten to replace `cr.savepoint()` with bare `cr.rollback()` and `cr.commit()`. This is a significant change in transaction semantics:

- The original used `cr.savepoint()`, which creates a SQL SAVEPOINT -- on error only the work inside that savepoint is rolled back, and the outer transaction continues cleanly.
- The new code calls `cr.rollback()` on exception, which rolls back the **entire transaction** (not just the current record's work), potentially undoing work from previous loop iterations that had already succeeded.
- `cr.commit()` inside a loop is generally discouraged in Odoo because it breaks atomicity and can leave the database in an inconsistent state if the cron job is interrupted mid-way. If the cron crashes after committing 50 of 100 records, a re-run will process a different set (since committed records changed state), which may cause subtle issues.
- The `auto_commit` flag makes this opt-in, which mitigates the risk somewhat, but the `cr.rollback()` path runs unconditionally (even when `auto_commit=False`), meaning errors now roll back the entire transaction rather than just a savepoint.

This fundamentally changes error isolation behavior and should be reconsidered.

**2. Version downgrade: `14.0.1.5.0` -> `14.0.1.4.0`**

As already noted by @rousseldenis, the version in `__manifest__.py` is being downgraded from `14.0.1.5.0` (current base) to `14.0.1.4.0`. The OCA merge bot handles version bumps. The PR author acknowledged this but the diff still shows the downgrade on the current head commit (`561677785d`). This must be fixed before merge.

---

### Moderate Issues

**3. `search_limit` defaults to `20` -- may silently skip records**

The new `search_limit` field defaults to `20`. This means the automated workflow cron will, by default, only process 20 records per run. If a backlog builds up (e.g., 200 orders to validate), it will take 10 cron runs to clear. Worse, users may not realize records are being silently deferred. Consider:
- Defaulting to `0` (no limit) to preserve current behavior
- Or at minimum documenting this behavior change prominently

**4. `onchange_partner_id` override pattern**

The `onchange_partner_id` method on `SaleOrder` calls `super().onchange_partner_id()` and returns `res`. In Odoo 14, the `onchange` on `partner_id` in the base `sale.order` is decorated with `@api.onchange` and does not return a value (returns `None` implicitly). Storing and returning `res` is harmless but unnecessary. More importantly, overriding `onchange_partner_id` by name is fragile -- if other modules also override this `@api.onchange`, method resolution depends on module load order.

**5. XPath `expr="//page[2]"` is fragile**

In `res_partner_view.xml`, the view inherits `base.view_partner_form` and uses `xpath expr="//page[2]"` to position the new tab. This is position-dependent and will break if other modules add or remove pages before position 2. A named page reference (e.g., `expr="//page[@name='sales_purchases']"`) would be more robust.

**6. No migration script for new fields**

The PR adds new fields (`default`, `auto_commit`, `search_limit` on `sale.workflow.process` and `workflow_process_id` on `res.partner`). The `external_dependencies` already includes `openupgradelib`, suggesting migrations are expected. Since the version is being changed, a migration script should be provided -- or the version should not be changed (let the merge bot handle it).

**7. Missing tests for new functionality**

No test files are added or modified. The new features (default workflow, partner workflow propagation, auto_commit, search_limit) have no test coverage. OCA guidelines expect test coverage for new features.

---

### Minor / Informational

**8. Copyright year `2016-21`** -- should be `2016-2021` (4-digit year in range).

**9. `default_get` override vs `default` field attribute** -- using `default_get` to search for a workflow with `default=True` is a valid approach, but it runs a search query on every sale order form open. For high-volume usage, a computed default or `ir.default` might be more efficient.

**10. HTML description changes** -- the `index.html` changes appear to be a regeneration of the RST-to-HTML output (CSS updates, formatting). These are fine but unrelated to the functional changes.

---

### Summary

The PR brings useful features (batch limiting, auto-commit control, partner-level workflow defaults) but has significant concerns around transaction management semantics and missing test coverage. The version downgrade must be fixed. Given that 14.0 is EOL, contributors may want to focus effort on forward-porting to a supported branch instead.

PR #3589, opened 2025-02-21
[17.0][MIG] partner_contact_sale_info_propagation: Migration to 17.0

juancarlosonate-tecnativa:
This PR replaces https://github.com/OCA/sale-workflow/pull/3127 due to inactivity.

The onchange and get_view methods are no longer necessary because the propagation logic is now handled directly in the create and compute methods in base https://github.com/odoo/odoo/blob/643180fc73658ca2425dc4bc6a9f242978efe9fc/odoo/addons/base/models/res_partner.py#L379 The previous approach relied on onchange to update fields when the parent_id changed and modified the view context dynamically, but this is redundant with the current implementation, which ensures propagation at the model level.

@Tecnativa TT52337
@sergio-teruel @pedrobaeza can you check? thanks!


sergio-teruel (approved):
Functional review OK

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -18,67 +15,43 @@ def _check_propagation_allowed(self):
         )
 
     def write(self, vals):
-        """Propagate Salesperson and Sales Channel change in the partner to the
-        child contacts."""
+        """
```
pedrobaeza: The correct docstring format according PEP257 is the previous one: https://peps.python.org/pep-0257/#multi-line-docstrings

partner_contact_sale_info_propagation/models/res_partner.py
```diff
@@ -18,67 +15,43 @@ def _check_propagation_allowed(self):
...
+        child contacts, and ensure team_id is inherited from parent_id.
+        """
         if not self._check_propagation_allowed():
             return super().write(vals)
         for record in self:
             if "user_id" in vals:
                 childs = record.mapped("child_ids").filtered(
-                    lambda r: not r.user_id or r.user_id == record.user_id
+                    lambda r, user_id=record.user_id: not r.user_id
+                    or r.user_id == user_id
```
pedrobaeza: The variable should be named after what represents, and it's a record, not an ID (the only exception here is the fields, which suffixes `_id` for noting that they are m2o records):

```suggestion
                    lambda r, user=record.user_id: not r.user_id
                    or r.user_id == user
```

Do the same for the rest of the occurrences.

partner_contact_sale_info_propagation/tests/test_partner_contact_sale_info_propagation.py
```diff
@@ -1,67 +1,69 @@
...
         self.partner_model = self.env["res.partner"].with_context(test_propagation=True)
-        self.parent_company = self.partner_model.create(
+        self.salesperson = self.env["res.users"].create(
+            {
+                "name": "Test Salesperson",
+                "login": "sales@test.com",
+            }
+        )
+        self.team = self.env["crm.team"].create({"name": "Sales Team A"})
+        self.company = self.partner_model.create(
```
pedrobaeza: Don't name it as company, as that's a confusing terminology, thinking that is a `res.company` record. Call it `parent_partner` for example.

partner_contact_sale_info_propagation/models/res_partner.py
juancarlosonate-tecnativa: done, review it again! thanks!

partner_contact_sale_info_propagation/models/res_partner.py
pedrobaeza: Being strict, the closing `"""` should be put in a new line (not the opening one, which is where I commented), but for this time, it's OK to leave it this way.

pedrobaeza (approved):
/ocabot merge nobump

partner_contact_sale_info_propagation/models/res_partner.py
juancarlosonate-tecnativa: next time, i will nail it!

PR #3595, opened 2025-02-25
[16.0][FIX] sale_payment_sheet + sale_order_type + sale_force_invoiced + sale_order_invoicing_finished_task + sale_order_line_sequence: Post-install test + fallback to load CoA

victoralmau:
Locked by:
- [ ] https://github.com/OCA/sale-workflow/pull/2885

Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

`odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale`

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/sale_order_invoicing_finished_task/tests/test_sale_order_invoicing_finished_task.py
+++ b/sale_order_invoicing_finished_task/tests/test_sale_order_invoicing_finished_task.py
@@ -2,13 +2,23 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo.exceptions import ValidationError
-from odoo.tests import TransactionCase
+from odoo.tests import TransactionCase, tagged
 
 
+@tagged("-at_install", "post_install")
 class TestInvoicefinishedTask(TransactionCase):
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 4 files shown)


victoralmau:
Addons incompatibility related to https://github.com/OCA/account-invoicing/pull/1840

juancarlosonate-tecnativa:
should a pr be opened for each of the mentioned modules and then cherry-pick?

juancarlosonate-tecnativa:
I think the commit for sale_force_invoiced can be removed https://github.com/OCA/sale-workflow/pull/3600

victoralmau:
Finally the CI is already green, so this could already be merged. Ping @pedrobaeza

sergio-teruel (approved):

PR #3592, opened 2025-02-25
[16.0][ADD]sale_order_warehouse_location: Set warehouse in sale order based on delivery country and state

manuelregidor:
New module to automatically set the warehouse in a sale order depending on the country and state of the delivery address.

An error in tests seems to be caused by these changes in account-invoicing repository: https://github.com/OCA/account-invoicing/pull/1840

@HaraldPanten @ValentinVinagre @luis-ron 

T-7589

PR #3598, opened 2025-02-25
[FIX] sale_fixed_discount, sale_triple_discount: modules are incompatibles and can not work together by design.

legalsylvain:
similar to account_invoice_fixed_discount and account_invoice_triple_discount

**Rational:** 

- account_invoice_fixed_discount and account_invoice_triple_discount have been incompatible since their creation. (by design) 
- sale_fixed_discount and sale_triple_discount have exactly the same incompatibility.
- However, modules has never been marked as incompatible. (not in the copier answer / CI and neither in the manifest).
- during @grindtildeath refactoring in https://github.com/OCA/account-invoicing/pull/1638 and https://github.com/OCA/account-invoicing/pull/1840, it was mandatory to mark the two account modules as rebel.
- so in the same way, it is now mandatory to mark the two sale modules as rebel.

Fixes problem in 
- https://github.com/OCA/sale-workflow/pull/3595#issuecomment-2681996180 (CC : @victoralmau)
- https://github.com/OCA/sale-workflow/pull/3592#issue-2877691549 CC : (@manuelregidor)

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -18,6 +18,8 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - sale_packaging_default,sale_order_product_recommendation,sale_order_product_recommendation_packaging_default,sale_order_product_recommendation_elaboration
+- sale_triple_discount
+- sale_fixed_discount
 repo_description: 'TODO: add repo description.'
 repo_name: sale-workflow
 repo_slug: sale-workflow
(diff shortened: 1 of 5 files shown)


legalsylvain:
@grindtildeath : The CI is red because, I guess, https://github.com/OCA/sale-workflow/pull/2885 is not merged. don't you think ? 
Could you rebase #2885 ? I guess you have to cherry pick my commit.

Thanks !

manuelregidor:
@legalsylvain Any idea when this will be solved? Thank you!

legalsylvain:
@manuelregidor please see that PR https://github.com/OCA/sale-workflow/pull/2885

PR #2612, opened 2023-07-20
[16.0][mig] sale_order_note_template

aktiv-heli-kantawala:

PR #3605, opened 2025-02-27
[16.0][MIG] sale_order_note_template

IsabelAForgeFlow:
Supersede: https://github.com/OCA/sale-workflow/pull/2612


AaronHForgeFlow (commented):
There are some things that are not correct IMHO:

- There is no menu for the T&C Template: you can only create them from the SO. But you can only edit them when the SO is confirmed (because the many2one field becomes not editable and you can go inside to edit it)
- The archive wizard is not good: 
image

AaronHForgeFlow (approved):
LGTM! Code + Functional

sebalix (approved):

dreispt (approved):

PR #3606, opened 2025-02-28
[16.0][WIP] sale_payment_sheet: Define @freeze_time decorator in tests and not in TestSaleInvoicePayment to avoid side effects

victoralmau:
Define @freeze_time decorator in tests and not in TestSaleInvoicePayment to avoid side effects

If a module (sale_payment_sheet_financial_risk for example) has a test https://github.com/OCA/credit-control/blob/120b39c8d92e856159d79ca084e69dd0d9c6badd/sale_payment_sheet_financial_risk/tests/test_sale_payment_sheet_financial_risk.py#L9 that extends it, a crash occurs in post tests

@Tecnativa


victoralmau:
Added in https://github.com/OCA/sale-workflow/pull/3595

PR #3609, opened 2025-03-03
[18.0] sale_restricted_qty: Migration to 18.0

samirGuesmi:


sale_restricted_qty/tests/test_sale.py
```diff
@@ -0,0 +1,188 @@
...
+        line_values = {"product_id": self.product.id, "product_uom_qty": 15.0}
+        self.product.manual_sale_min_qty = 10
+        self.product.manual_sale_multiple_qty = 10
+        # Create sale order line with Qty not multiple Qty
+        with self.assertRaises(ValidationError):
+            sale_order = self.sale_order_model.create(
+                {"partner_id": self.partner.id, "order_line": [(0, 0, line_values)]}
+            )
+            sale_values = self.refrech_sale_values(sale_order)
+            self.sale_order_model.create(sale_values)
```
rousseldenis: @samirGuesmi Could you check why tests are not reaching this line? I think I don't see any `refrech_sale_values` method.

samirGuesmi: @rousseldenis I deleted these lines, they were most propably forgotten during a previous update.

bosd:
Can you please recreate a runboat?

marielejeune (approved):
LGTM (code review)

sale_restricted_qty/models/sale.py
```diff
@@ -0,0 +1,188 @@
...
+                )
+            )
+
+        if invaild_min_lines:
+            msg += _(
+                "Check minimum order quantity for this products: * \n"
+            ) + "\n ".join(invaild_min_lines)
+            msg += _(
+                "\n* If you want sell quantity less than Min Quantity"
+                ',Check "force min quatity" on product'
```
bosd, line 72:
```suggestion
                ',Check "force min quantity" on product'
```

sale_restricted_qty/models/sale.py
```diff
@@ -0,0 +1,188 @@
...
+                )
+            )
+
+        if invaild_max_lines:
+            msg += _(
+                "Check maximum order quantity for this products: * \n"
+            ) + "\n ".join(invaild_max_lines)
+            msg += _(
+                "\n* If you want sell quantity bigger than max Quantity"
+                ',Check "force max quatity" on product'
```
bosd, line 93:
```suggestion
                ',Check "force max quantity" on product'
```

sale_restricted_qty/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,10 @@
+This module allows to define min, max and multiple order quantity on
+product. Min and Max order quantity is either required or recommanded.
```
bosd, line 2:
```suggestion
product. Min and Max order quantity is either required or recommended.
```

bosd (changes requested):
Thanks for your work on this, but the Functional test made me very confused:

Set a min quantity on a product of 100.
Add 50 pieces of the product to an so.
Confirm the SO. 
**Result**
There was no warning, that my quantity is below minimum. SO got confirmed.
**Expected result**
A warning that I am selling below the MOQ. Unable to confirm the SO.

Then I Tried to copy that Saleorder.
Now there is an error message that my MOQ is below the minimum :confused: :grey_question: 
The message:
```
If you want sell quantity less than Min Quantity,Check "force min quatity" on product
```
qua**n**tity

So if I understand this correctly, I have to enable the enforce min quantity  to be able to use less then the minimum quantity? :brain: :question:

bosd:
Seems like most of my comments are actually already been addressed in: https://github.com/OCA/sale-workflow/pull/2997 by @alexey-pelykh 
Is that still a WIP or ready to be included here?

alexey-pelykh:
@bosd already used in production, I wanted to merge migration first and then refactored one - but it's ready

alexey-pelykh:
Idea is that #2997 completely remade the UI as current one less than optimal let's say, so migration to 18.0 should've included that. However as #2996 has been in review for _one year 7 months_ (completely normal phenomenon for a functioning organization), obviously this migration does not include it

sale_restricted_qty/views/sale_views.xml
```diff
@@ -0,0 +1,93 @@
...
+                <field name="force_sale_min_qty" invisible="1" />
+                <field name="sale_max_qty" />
+                <field name="is_qty_bigger_max_qty" invisible="1" />
+                <field name="force_sale_max_qty" invisible="1" />
+                <field name="is_qty_not_multiple_qty" invisible="1" />
+                <label
+                    colspan="2"
+                    for="sale_min_qty"
+                    string="Min Quantity Recommended"
+                    attrs="{'invisible': ['|',  ('is_qty_less_min_qty','=', False),
```
alexey-pelykh, line 25: **Critical (migration blocker):** The `attrs` dict syntax (`attrs="{'invisible': [...]}"`) was removed in Odoo 17.0/18.0. All `attrs` in this file must be migrated to the new inline domain expression syntax.

For example, this should become:
```xml
<label
    colspan="2"
    for="sale_min_qty"
    string="Min Quantity Recommended"
    invisible="is_qty_less_min_qty == False or (is_qty_less_min_qty == True and force_sale_min_qty == False)"
    style="background-color:yellow"
/>
```

All 5 `attrs` usages in this file need the same treatment.

sale_restricted_qty/views/sale_views.xml
```diff
@@ -0,0 +1,93 @@
...
+                    string="Max Quantity Recommended"
+                    attrs="{'invisible': ['|',  ('is_qty_bigger_max_qty','=', False),
+                     '&amp;', ('is_qty_bigger_max_qty','=', True), ('force_sale_max_qty','=', False,)]}"
+                    style="background-color:yellow"
+                />
+                <label
+                    colspan="2"
+                    for="sale_min_qty"
+                    string="Max Quantity Exceeded"
+                    attrs="{'invisible': ['|',  ('is_qty_bigger_max_qty','=', True),
```
alexey-pelykh, line 56: **Bug:** The invisible domain here is inverted. The first `|` operand is `('is_qty_bigger_max_qty','=', True)` which makes this label invisible when max IS exceeded -- the opposite of the intended behavior.

Compare with the working "Min Quantity Required" label pattern (line 33) which uses `('is_qty_less_min_qty','=', False)` as the first operand.

This should be:
```
invisible: ['|', ('is_qty_bigger_max_qty','=', False),
 '&', ('is_qty_bigger_max_qty','=', True), ('force_sale_max_qty','=', True)]
```
(invisible when max not exceeded OR when max exceeded but force is enabled, i.e. only visible when exceeded AND required).

sale_restricted_qty/views/sale_views.xml
```diff
@@ -0,0 +1,93 @@
...
+                <field name="force_sale_min_qty" column_invisible="1" />
+                <field name="sale_max_qty" optional="hide" />
+                <field name="is_qty_bigger_max_qty" column_invisible="1" />
+                <field name="force_sale_max_qty" column_invisible="1" />
+                <field name="is_qty_not_multiple_qty" column_invisible="1" />
+            </xpath>
+
+            <xpath expr="//page/field[@name='order_line']/list" position="attributes">
+                <attribute
+                    name="decoration-danger"
```
alexey-pelykh, line 76: **Bug:** Three `<attribute name="decoration-danger">` elements are applied to the same `<list>`. In XML, only the last attribute value is kept, so:
- Min qty danger decoration (line 67-68) -- **silently lost**
- Max qty danger decoration (line 73-74) -- **silently lost**
- Multiple qty danger decoration (line 79-80) -- the only one that actually works

These should be combined into a single expression:
```xml
<attribute name="decoration-danger">
    (is_qty_less_min_qty == True and force_sale_min_qty == False) or
    (is_qty_bigger_max_qty == True and force_sale_max_qty == False) or
    is_qty_not_multiple_qty == True
</attribute>
```

Same issue applies to `decoration-warning` if both min and max warnings are needed.

sale_restricted_qty/i18n/sale_order_min_qty.pot
```diff
@@ -0,0 +1,82 @@
+# Translation of Odoo Server.
```
alexey-pelykh, line 1: This `.pot` file belongs to the old module name (`sale_order_min_qty`) and should not be included in the 18.0 migration of `sale_restricted_qty`. Please remove it.

sale_restricted_qty/views/product_category_views.xml
```diff
@@ -0,0 +1,58 @@
...
+                        />
+                        <field name="force_sale_max_qty" />
+                        <label for="manual_force_sale_max_qty" class="oe_edit_only" />
+                        <field
+                            name="manual_force_sale_max_qty"
+                            class="oe_edit_only"
+                            nolabel="1"
+                        />
+                    </group>
+                    <group name="sale_multiple_qty_gr" string="Sale Miltiple qty">
```
alexey-pelykh, line 45: Typo: "Sale Miltiple qty" should be "Sale Multiple qty". This same typo is also present in `product_product_views.xml` and `product_template_views.xml`.

sale_restricted_qty/tests/test_sale.py
```diff
@@ -0,0 +1,184 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import odoo.tests.common as common
+from odoo.exceptions import ValidationError
+from odoo.tests import tagged
+
+
+@tagged("post_install", "-at_install")
+class TestSaleOrderLineMinQty(common.TransactionCase):
+    def setUp(self):
```
alexey-pelykh, line 13: OCA best practice for 18.0: consider using `@classmethod` + `setUpClass(cls)` instead of `setUp(self)`. This runs the setup once per test class instead of once per test method, improving test suite performance.

```python
@classmethod
def setUpClass(cls):
    super().setUpClass()
    # Create models
    cls.sale_order_model = cls.env["sale.order"]
    ...
```

sale_restricted_qty/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Akretion (<http://www.akretion.com>)
+# Copyright 2023 Simone Rubino - Aion Tech
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale order min quantity",
+    "version": "18.0.1.0.0",
+    "category": "Sales Management",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "contributors": ["Ashish Hirpara"],
```
alexey-pelykh, line 10: The `contributors` key is not a standard OCA manifest key and is not recognized by OCA tooling. Contributors are already properly listed in `readme/CONTRIBUTORS.md`. Consider removing this key.

sale_restricted_qty/models/sale.py
```diff
@@ -0,0 +1,188 @@
...
+    is_qty_not_multiple_qty = fields.Boolean(
+        string="Not Multiple Qty", compute="_compute_is_qty_not_multiple_qty"
+    )
+
+    @api.constrains(
+        "product_uom_qty", "sale_min_qty", "sale_max_qty", "sale_multiple_qty"
+    )
+    def check_constraint_restricted_qty(self):
+        msg = ""
+        invaild_min_lines = []
```
alexey-pelykh, line 53: Minor typo: `invaild_min_lines` should be `invalid_min_lines`. The same typo (`invaild`) is repeated for `invaild_max_lines` and `invaild_multiple_lines` throughout this method.

alexey-pelykh (changes requested):
## Review: [18.0] sale_restricted_qty: Migration to 18.0

Thank you for working on this migration. The module structure and business logic are solid overall. However, I found several issues that need to be addressed before this can be merged -- most critically, the view XML uses syntax that is incompatible with Odoo 18.0.

### Critical

1. **`attrs` dict syntax must be migrated to Odoo 18.0 inline domain expressions** (`sale_views.xml`). In Odoo 18.0 (since 17.0), the `attrs="{'invisible': [...]}"` dict syntax was removed. All `attrs` must be replaced with the new inline syntax, e.g. `invisible="[('field','=',value)]"`. This is a migration blocker.

2. **Duplicate `decoration-danger` attributes** (`sale_views.xml`, lines 67-80). Three `<attribute name="decoration-danger">` elements are set on the same tree view. Only the last one takes effect, so min-qty and max-qty danger decorations are silently lost. These need to be combined into a single expression using `or`.

3. **Inverted invisible logic on "Max Quantity Exceeded" label** (`sale_views.xml`, line 56). The first condition is `('is_qty_bigger_max_qty','=', True)` which makes the label invisible when the max IS exceeded and visible when it is NOT -- the exact opposite of intended behavior. Compare with the working "Min Quantity Required" pattern which correctly uses `('is_qty_less_min_qty','=', False)` as the first `|` operand.

### Cleanup

4. **Stale `sale_order_min_qty.pot` file should be removed.** This `.pot` file belongs to the old module name and should not be shipped in the 18.0 migration.

5. **Typo: "Sale Miltiple qty"** should be "Sale Multiple qty" -- present in all three product view files and propagated to `.po`/`.pot` files.

6. **`oe_edit_only` CSS class** -- this class was deprecated/removed in Odoo 17.0+. In 18.0, the replacement is `readonly="1"` on the field or conditional `invisible` expressions. Please verify this still works in 18.0 and migrate if needed.

### Best Practices

7. **Tests use `setUp` instead of `setUpClass`** (`test_sale.py`). OCA best practice for 18.0 is `@classmethod` + `setUpClass(cls)` for better test performance since setup runs once per class rather than once per test method.

8. **`contributors` is not a standard OCA manifest key** (`__manifest__.py`). Contributors should be listed in `readme/CONTRIBUTORS.md` (which already exists). The `contributors` key in `__manifest__.py` is not recognized by OCA tooling.

### Minor

9. **Typo `invaild` in variable names** (`sale.py`): `invaild_min_lines`, `invaild_max_lines`, `invaild_multiple_lines` should be `invalid_*`.

10. **Translation file headers are outdated**: `sale_restricted_qty.pot` references "Odoo Server 16.0", `fr.po` references 14.0, `it.po` references 12.0. These should be regenerated for 18.0.

---

marielejeune:
@samirGuesmi I think you could close this PR as it is superseeded by https://github.com/OCA/sale-workflow/pull/3959

PR #3612, opened 2025-03-05
[MIG] sale_delivery_split_date: Migration to 18.0

jaydeep32:


rousseldenis:
@jaydeep32 Could you rebase ?

sale_delivery_split_date/views/stock_picking.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2018 Alex Comba - Agile Business Group
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <record id="vpicktree" model="ir.ui.view">
+        <field name="name">stock.picking.tree</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.vpicktree" />
```
alexey-pelykh, line 8: This `vpicktree` override is **no longer needed** in Odoo 18.0. The base `stock` module already defines `scheduled_date` as visible:

```xml
<field name="scheduled_date" optional="show" widget="remaining_days" invisible="state in ('done', 'cancel')" readonly="state in ['cancel', 'done']"/>
```

Setting `invisible="0"` here would conflict with the conditional visibility and remove the `remaining_days` widget behavior. Please remove this entire `<record id="vpicktree">` block.

The search view override below (`view_picking_internal_search`) is still valid and should be kept.

sale_delivery_split_date/readme/ROADMAP.md
```diff
@@ -0,0 +1,2 @@
+- Incompatible with
+  [sale_procurement_group_by_commitment_date](https://github.com/OCA/sale-workflow/tree/12.0/sale_procurement_group_by_commitment_date)
```
alexey-pelykh, line 2: This links to the **12.0** branch of `sale_procurement_group_by_commitment_date`, which no longer exists on 18.0. Please either:
- Update to a branch where it was last available, or
- Remove this entry if the incompatibility no longer applies in 18.0

Note that the generated `README.rst` and `index.html` will also need regeneration after this fix.

sale_delivery_split_date/models/sale_order_line.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.commitment_date:
+            comm_date = self._get_security_lead_time_commitment_date()
+            vals.update({"date_planned": comm_date})
+        return vals
+
+    def _get_security_lead_time_commitment_date(self):
+        """Return the commitment date with security lead time"""
+        return self.commitment_date - timedelta(
+            days=self.order_id.company_id.security_lead
+        )
```
alexey-pelykh, line 46: The model code is identical to 16.0. I've verified that the core Odoo 18.0 `sale_stock` module still uses the same method signatures:
- `_prepare_procurement_values(self, group_id=False)` -- unchanged
- `_prepare_procurement_group_vals()` -- provided by `sale_procurement_group_by_line`
- `_get_procurement_group_key()` -- provided by `sale_procurement_group_by_line`

The `date_planned` key in `_prepare_procurement_values` is also still used in 18.0. So the logic is compatible -- just needs CI validation via rebase.

alexey-pelykh (changes requested):
## Review: `sale_delivery_split_date` migration to 18.0

Thanks for working on this migration. The module structure is solid and tests are comprehensive. However, there are a few issues that need to be addressed for Odoo 18.0 compatibility.

### Issues Found

#### 1. Unnecessary view override for `vpicktree` (stock.picking tree view)

In Odoo 18.0, `scheduled_date` is **already visible** in the `vpicktree` list view with `optional="show"` and `widget="remaining_days"`. The view override that sets `invisible="0"` is no longer needed and may conflict with the existing attribute definitions (since 18.0 uses `invisible="state in ('done', 'cancel')"` rather than a simple boolean). This override should be removed entirely.

#### 2. `date_planned` key may need verification in `_prepare_procurement_values`

The method `_prepare_procurement_values` sets `vals["date_planned"]`. In Odoo 18.0 `sale_stock`, this key is indeed still used (confirmed by checking the base method), so this is correct. No change needed here -- just flagging that it was verified.

#### 3. Roadmap references outdated branch (12.0)

The `readme/ROADMAP.md` (and the generated `README.rst`) references `sale_procurement_group_by_commitment_date` with a link to the **12.0** branch. This module does not exist on 18.0. The roadmap should either:
- Update the link to a relevant branch where it last existed, or
- Remove the entry entirely if the incompatibility is no longer relevant

#### 4. Translation files reference old Odoo Server version

The `.po` files in `i18n/` reference `Project-Id-Version: Odoo Server 12.0` (ca, es, pt_PT) or include stale `POT-Creation-Date` from 12.0 era. While these will still work, the OCA tooling would normally regenerate them. The `.pot` file correctly references 18.0. This is a minor issue -- the translations will be picked up by Weblate after merge.

#### 5. Model code is identical to 16.0 -- verify API compatibility

The `models/sale_order_line.py` is byte-for-byte identical to the 16.0 version. While the core methods (`_prepare_procurement_values`, `_get_procurement_group_key`, `_prepare_procurement_group_vals`) still exist with the same signatures in Odoo 18.0, **the PR needs to pass CI** to confirm full runtime compatibility. The PR currently has `mergeable_state: dirty` (needs rebase), so CI cannot run.

#### 6. Missing `pyproject.toml` `installable` handling

The `pyproject.toml` is correct and uses `whool` as the build backend. The `__manifest__.py` includes `"installable": True` which is fine.

### Action Required

1. **Remove the `vpicktree` view override** from `views/stock_picking.xml` (the `<record id="vpicktree">` block) since `scheduled_date` is already visible in Odoo 18.0. Keep the search view addition (`view_picking_internal_search`), as the search field is not present by default.
2. **Update `readme/ROADMAP.md`** to fix the stale 12.0 branch link for `sale_procurement_group_by_commitment_date`, or remove the entry if it's no longer relevant for 18.0.
3. **Rebase onto the current 18.0 branch** as requested by the maintainer.
4. Regenerate `README.rst` and `static/description/index.html` via `oca-gen-addon-readme` after the above changes.

---

PR #3617, opened 2025-03-10
[18.0][MIG] sale_order_qty_change_no_recompute: Migration to 18.0

psugne:


sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.tests import Form
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderQtyChange(TransactionCase):
```
CRogos, line 7: Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

CRogos (changes requested):
@psugne are you still working on this?

Some of the commits can be squashed and the pre-commit commit is not correctly named:
image

Also some unit tests are failing

pedrobaeza:
It seems the module is no longer needed: https://github.com/OCA/OpenUpgrade/pull/4870

CRogos:
Totally agree with @pedrobaeza 
Just followed these instructions and the price was not reset without this module.

```

1.  Create a new sale order.
2.  Add product line and set custom unit price.
3.  Save sale order.
4.  Edit sale order and change quantity (custom unit price not been
    reset).
```

=> close

psugne:
Closing PR as sale_order_qty_change_no_recompute module is no longer needed.

PR #3618, opened 2025-03-10
[18.0][MIG] product_price_category

bealdav:


bealdav:
Just tested, fix done by @florian-dacosta works nicely

I approve it

florian-dacosta (approved):

florian-dacosta:
FYI as last contributor / reviewers, it may be of your interest
@sbejaoui 
@rousseldenis
@marielejeune
@flotho
@JordiMForgeFlow

product_price_category/models/product_pricelist_item.py
```diff
@@ -13,6 +13,10 @@ class ProductPricelistItem(models.Model):
         selection_add=[("2b_product_price_category", "Price Category")],
         ondelete={"2b_product_price_category": "set default"},
     )
+    display_applied_on = fields.Selection(
```
rousseldenis, line 16: @bealdav No need of migration script for this ?

florian-dacosta: Hum, true, it is for a new project, and I did not even think about it...
I'll provide one

florian-dacosta: Thanks for your vigilance @rousseldenis 
I've added a migration script.I could not test it on a real migrated database though, but I did test that it runs without crashing

product_price_category/models/product_pricelist_item.py
```diff
@@ -24,20 +28,21 @@ class ProductPricelistItem(models.Model):
         readonly=False,
     )
 
-    def _compute_name_and_price(self):
-        result = super()._compute_name_and_price()
+    def _compute_name(self):
```
JordiMForgeFlow, line 32: api.depends should be maintained from the original method

florian-dacosta: I did the change.
I hardly believed it was not necessary although it does not hurt.

rousseldenis: @JordiMForgeFlow For `depends()`, Odoo will append new dependencies in further modules that decorates same function. That's why you don't need to put original fields in it. @florian-dacosta This is why it worked without. This is not necessary.

I'm wondering if `price_category_id` should not be in depends ?

bealdav: I've just rebased this old PR.
@rousseldenis , concerning if price_category_id should trigger this method, on my side I think not.
I've tested here

Image

Image

On save computed name is refreshed when you change the price category

It seems to me we have answered all questions.

A merge can be done ?

bealdav: Please @rousseldenis are these screens anwser to your questions ?
Could you merge if it's ok ?

sebastienbeau: @florian-dacosta can you remove the useless depends field added 
I confirm that it's useless and also **it propagate a wrong believe that you need to add all the depends**
The code in V18 here : https://github.com/odoo/odoo/blob/dc9dc90dd99f1f7dfb4a781ad2bec7378548d95b/odoo/fields.py#L585 will aggregate the depends. It's the case since version 13 at least and maybe even before (but I do not remember)

Regarding the fact of add "price_category_id" it's needed. The rule is simple if the result of computation of the field depend on "price_category_id", you have to add it. 

Right now the onchange is broken (the name only change when you save it). I am pretty sure that if you do an import (an update) it will not update the name if there is not depends on "price_category_id"

So please add it

bealdav: Done @sebastienbeau

bealdav: @rousseldenis we finally found you're right about depends on price_category_id.

florian-dacosta (approved):
Functional test + new review
All seems good, thanks @bealdav

product_price_category/models/product_pricelist_item.py
sebastienbeau: OK for me

sebastienbeau (approved):
LGTM (code review)

PR #3620, opened 2025-03-12
[16.0][FIX] sale_blanket_order: SO from Blanket Order disable adding lines

LauraCForgeFlow:
Before this fix, SO lines could never be added to SOs from Blanket Orders, no matter if the user belonged to the _Disable adding more lines to SOs from Blanket Orders_ or not. It seems like `<t groups...` in an `attributes` `<xpath>` was not working as expected.

With this commit, that problem is solved.


AaronHForgeFlow (approved):
Code review LGTM

AlexPForgeFlow (approved):
Code LGTM!

MeritxellAForgeFlow (approved):
LGTM! :)

PR #3621, opened 2025-03-13
[17.0][FIX] sale_force_whole_invoiceability: button visibility

jbernausVunkers:
Since the code will throw an error if the button is used when it is not a confirmed sale order or if we already have an invoice, we should hide the button in those cases.

### DIFF
--- a/sale_force_whole_invoiceability/views/sale_order_views.xml
+++ b/sale_force_whole_invoiceability/views/sale_order_views.xml
@@ -11,7 +11,7 @@
                     name="force_lines_to_invoice_policy_order"
                     string="Force Invoice Policy"
                     type="object"
-                    invisible="state not in ['sale'] and invoice_count > 0"
+                    invisible="state not in ['sale'] or invoice_count > 0"
                     groups="sales_team.group_sale_manager"
                 />
             </button>


pedrobaeza (approved):

alexey-pelykh (approved):
Code review: the fix is correct.

## Analysis

The button `force_lines_to_invoice_policy_order` should only be visible when the sale order is in `sale` state **and** has no invoices yet. The Python method (`sale_order.py`) enforces exactly this, raising `UserError` when:

1. `self.state != "sale"` -- order not confirmed
2. `self.invoice_count` -- order already invoiced

The **old** domain used `and`:
```xml
invisible="state not in ['sale'] and invoice_count > 0"
```
This hides the button only when **both** conditions are true simultaneously (not in sale AND has invoices). In practice the button was visible on draft orders (no invoices, so the `and` short-circuits) and on confirmed orders with invoices (state is `sale`, so the first condition is false).

The **fix** changes to `or`:
```xml
invisible="state not in ['sale'] or invoice_count > 0"
```
Now the button is hidden whenever **either** condition is true, which matches the Python guards exactly.

## Correctness

| State | Invoices | Old (and) | New (or) | Python allows? |
|-------|----------|-----------|----------|----------------|
| draft | 0 | visible | **hidden** | No |
| sale | 0 | visible | visible | Yes |
| draft | >0 | hidden | hidden | No |
| sale | >0 | visible | **hidden** | No |

All four cases now align with the server-side validation.

## Tests

Existing tests (`test_sale_force_whole.py`) cover all three `UserError` paths (draft state, wrong user group, already invoiced) and the happy path. The fix is a view-only change that makes the UI consistent with already-tested server logic. No new tests are needed.

LGTM -- clean one-line fix with clear correctness argument.

pedrobaeza (approved):
/ocabot merge patch

PR #3623, opened 2025-03-13
[16.0][IMP] sale_elaboration: Add multilines and new report

AlvaroRM11:
@BinhexTeam

### DIFF
--- a/sale_elaboration/__manifest__.py
+++ b/sale_elaboration/__manifest__.py
@@ -24,6 +24,7 @@
         "views/stock_move_line_views.xml",
         "views/stock_picking_views.xml",
         "reports/report_base.xml",
+        "reports/report_saleorder.xml",
         "reports/report_deliveryslip.xml",
         "reports/report_picking_operations.xml",
     ],
--- a/sale_elaboration/models/product_elaboration_mixin.py
+++ b/sale_elaboration/models/product_elaboration_mixin.py
@@ -12,7 +12,7 @@ class ProductElaborationMixin(models.AbstractModel):
         comodel_name="product.elaboration",
         string="Elaborations",
     )
-    elaboration_note = fields.Char(
+    elaboration_note = fields.Text(
         store=True,
     )
     is_elaboration = fields.Boolean(
--- a/sale_elaboration/models/res_config_settings.py
+++ b/sale_elaboration/models/res_config_settings.py
@@ -11,6 +11,10 @@ class ResConfigSettings(models.TransientModel):
         "Display Elaboration notes on Delivery Slips",
         implied_group="sale_elaboration.group_elaboration_note_on_delivery_slip",
     )
+    group_elaboration_note_on_order_report = fields.Boolean(
+        "Display Elaboration notes on Order Reports",
+        implied_group="sale_elaboration.group_elaboration_note_on_order_report",
+    )
     group_elaboration_note_on_picking_operations = fields.Boolean(
         "Display Elaboration notes on Picking Operations",
         implied_group="sale_elaboration.group_elaboration_note_on_picking_operations",
--- a/sale_elaboration/models/stock_move.py
+++ b/sale_elaboration/models/stock_move.py
@@ -20,4 +20,4 @@ class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
     elaboration_ids = fields.Many2many(related="move_id.elaboration_ids")
-    elaboration_note = fields.Char(related="move_id.elaboration_note")
+    elaboration_note = fields.Text(related="move_id.elaboration_note")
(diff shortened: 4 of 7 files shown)


sale_elaboration/views/sale_order_views.xml
```diff
@@ -19,6 +19,7 @@
                     />
                     <field
                         name="elaboration_note"
+                        widget="text"
```
Shide: Why not change the original field from Char to Text?
All reports that use it will be changed and no widget is needed.

Shide (changes requested):
When you doesn't click the line, the text is displayed inline, but when you click the line, it's in multiline format (that's ok).

But on the reports, is always displayed inline. We want to use the minimum space for this text, so... 
- Why do you want multiline?
- Why show the field to salesman in a different look than the end customer sees it?

image

Do you have any example of a real multiline elaboration? Our customers usually uses few words to indicate a preference in preparation (for the inventory user).

Thanks!

sale_elaboration/views/sale_order_views.xml
AlvaroRM11: Thank you for your review! I was concerned that switching from "Char" to "Text" might delete current users' database information, but after testing, I confirmed that everything is working fine.

AlvaroRM11:
> When you doesn't click the line, the text is displayed inline, but when you click the line, it's in multiline format (that's ok).
> 
> But on the reports, is always displayed inline. We want to use the minimum space for this text, so...
> 
>     * Why do you want multiline?
> 
>     * Why show the field to salesman in a different look than the end customer sees it?
> 
> 
> image
> 
> Do you have any example of a real multiline elaboration? Our customers usually uses few words to indicate a preference in preparation (for the inventory user). The
> 
> Thanks!

Thank you for your review! I have a customer who requires two separate lines. One for packaging and one for format. I've taken your suggestions into account.

christian-ramos-tecnativa (approved):
LGTM

christian-ramos-tecnativa:
Any new thoughs @Shide?

Shide (approved):

rafaelbn:
Please @AlvaroRM11 , review test failing 🔴 and a rebase to check it again. Thanks!

alexey-pelykh (commented):
## Review: `sale_elaboration` - Add multilines and new report

Thanks for the contribution. The PR adds a useful feature (elaboration notes on sale order reports) and addresses the reviewer feedback about changing `Char` to `Text`. A few observations below.

### Key Findings

| 1 | Moderate | `reports/report_saleorder.xml` | The `<hr class="mt-2" />` is rendered **unconditionally** for every sale order line, regardless of whether the line has elaboration notes. When the `group_elaboration_note_on_order_report` group is disabled or the line has no elaboration, the user still sees an extra `<hr>` after every line. The `<hr>` should be inside the `t-if` / `groups` guard, or placed conditionally. Compare with `report_deliveryslip.xml` and `report_picking_operations.xml` which do not add extra visual separators. |
| 2 | Moderate | `reports/report_saleorder.xml` | The `elaboration_notes` template comment documents that `record` can be `stock.move` or `stock.move.line`, but here it is called with a `sale.order.line`. While it works (both have `elaboration_ids` and `elaboration_note` via the mixin), it would be good to update the comment in `report_base.xml` to also mention `sale.order.line` for clarity and maintainability. |
| 3 | Minor | `reports/report_saleorder.xml` | Missing XML declaration wrapper: the file uses `<data>` as root element, whereas the existing delivery slip report uses `<odoo>` as the root element. OCA convention for 16.0 typically uses `<odoo>` as the outer wrapper for data files. The `report_base.xml` and `report_picking_operations.xml` use `<data>`, so there is some inconsistency in the existing module, but matching `<odoo>` from the delivery slip report would be more standard. |
| 4 | Minor | `reports/report_saleorder.xml` | Missing copyright header. The other report XML files include copyright and license headers (e.g., `report_base.xml` and `report_deliveryslip.xml` have `<!-- Copyright ... -->` blocks). This new file should follow the same pattern. |
| 5 | Note | `__manifest__.py` | The version is not bumped. Since this PR adds a new feature (new security group, new config setting, new report, field type change), the version should be bumped per OCA guidelines. |
| 6 | Note | CI | Tests are currently failing (red) per the maintainer's comment. The author has been asked to investigate and rebase. |
| 7 | Note | `reports/report_saleorder.xml` | The xpath targets `//t[@t-foreach='lines_to_report']/tr` which depends on the structure of `sale.report_saleorder_document`. If upstream Odoo changes the report structure, this xpath may break. This is a standard risk for report inheritance and acceptable, just noting for awareness. |

### Summary

The core approach is sound: reusing the existing `elaboration_notes` template, adding a proper security group with a config toggle, and changing `Char` to `Text` as suggested by the prior reviewer. The main actionable issue is **finding #1** -- the unconditional `<hr>` that will appear for every sale order line even when no elaboration notes are present. This should be addressed before merge.

I'm posting as COMMENT rather than REQUEST_CHANGES since this PR already has two approvals and the maintainer has asked the author to fix the failing tests and rebase. The unconditional `<hr>` issue is worth addressing in that same pass.

PR #3628, opened 2025-03-17
[17.0][FIX] sale_blanket_order: SO from Blanket Order disable adding lines

LauraCForgeFlow:
Forward-port of #3620 to version 17.0

### DIFF
--- a/sale_blanket_order/models/sale_orders.py
+++ b/sale_blanket_order/models/sale_orders.py
@@ -15,6 +15,9 @@ class SaleOrder(models.Model):
         string="Origin blanket order",
         related="order_line.blanket_order_line.order_id",
     )
+    disable_adding_lines = fields.Boolean(
+        compute="_compute_disable_adding_lines",
+    )
 
     @api.model
     def _check_exchausted_blanket_order_line(self):
@@ -46,6 +49,16 @@ class SaleOrder(models.Model):
                             "blanket order lines customer"
                         )
                     )
+
+    @api.depends("blanket_order_id")
+    @api.depends_context("uid")
+    def _compute_disable_adding_lines(self):
+        self.disable_adding_lines = False
+        if self.env.user.has_group(
+            "sale_blanket_order.blanket_orders_disable_adding_lines"
+        ):
+            for order in self:
+                order.disable_adding_lines = order.blanket_order_id
 
 
 class SaleOrderLine(models.Model):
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

alexey-pelykh (approved):
## Review: sale_blanket_order - SO from Blanket Order disable adding lines

Forward-port of #3620 (16.0 → 17.0). The original bug: `<t groups="...">` inside an `attributes` xpath didn't work, so the "Disable adding more lines" group restriction was broken — lines could **never** be added to SOs from Blanket Orders regardless of group membership.

### Approach Assessment

The 17.0 forward-port correctly adapts the 16.0 fix to Odoo 17's OWL-based view system:

1. **Computed `disable_adding_lines` boolean** — Moves group-check logic from the view layer into a Python compute method. This is the right call: `<t groups="...">` within attribute xpaths was unreliable in 16.0 and remains so in 17.0.

2. **`options={'create': [('disable_adding_lines', '=', False)]}`** — Uses the Odoo 17 `options` dict with a domain-like condition on the x2many field instead of the 16.0 approach of `<attribute name="create">not disable_adding_lines</attribute>` on the tree element. This is the correct adaptation for Odoo 17's form view handling of x2many create permissions.

### Correctness

- **`@api.depends("blanket_order_id")`** — Correct dependency. `blanket_order_id` is itself computed from `order_line.blanket_order_line.order_id`, so the compute chain is properly established.
- **`@api.depends_context("uid")`** — Necessary since the result depends on the current user's group membership via `has_group()`. Good.
- **Default assignment pattern** — `self.disable_adding_lines = False` before the conditional loop is correct Odoo compute method practice (set default for all records, then override selectively).
- **Boolean coercion** — `order.disable_adding_lines = order.blanket_order_id` assigns a Many2one (truthy/falsy) to a Boolean field. Odoo handles this coercion correctly.
- **Invisible helper field** — `<field name="disable_adding_lines" invisible="1" />` placed before the `order_line` field ensures the value is available in the view context for the `options` evaluation.

### Security

- Group check via `has_group()` is the standard Odoo pattern — no concerns.
- No new security groups or access rules introduced; existing `blanket_orders_disable_adding_lines` group is reused.

### Performance

- The `has_group()` call is cached per user session, so the early-exit when user lacks the group avoids unnecessary iteration. Efficient.
- The compute is lightweight (boolean assignment from a Many2one truthiness check).

### Notes

- **No test changes**: The PR doesn't add tests for this fix. The original 16.0 PR (#3620) also lacked tests. Ideally a test would verify that users in the group cannot add lines to SOs linked to a blanket order, but this is a pre-existing gap, not introduced by this PR.
- **No manifest version bump**: The version remains `17.0.2.1.0`. For OCA modules, the version bump is typically handled by the maintainer/merge bot, so this is acceptable.

Clean forward-port. LGTM.

PR #3631, opened 2025-03-19
[18.0][ADD] sale_product_attribute_custom_value_variant

rousseldenis:
Depends on:

- https://github.com/OCA/product-attribute/pull/1934


sale_product_attribute_custom_value_variant/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": """
+        This module allows to use product custom attribute
+        and create the corresponding variant when using it on sale order line
+    """,
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Aion Tech,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": ["sale", "product_attribute_custom_value_variant"],
+}
```
alexey-pelykh, line 16: OCA convention: consider adding `"installable": True` and `"development_status": "Beta"` (the README badge already shows Beta).

sale_product_attribute_custom_value_variant/models/sale_order_line.py
```diff
@@ -0,0 +1,23 @@
...
+            ) as new_variant:
+                line.product_id = new_variant
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        custom_value_lines = lines.filtered("product_custom_attribute_value_ids")
+        if custom_value_lines:
+            custom_value_lines._assign_new_custom_variant()
+        return lines
```
alexey-pelykh, line 23: The `create` override handles variant creation nicely. Two observations:

1. **Double write on `product_id`**: After `super().create()`, assigning `line.product_id = new_variant` triggers a second write, which may cause dependent fields (price, taxes, etc.) to recompute twice. An alternative approach would be to resolve the variant *before* `super().create()` by modifying `vals_list` entries.

2. **No `write` override**: If a user edits an existing SO line to change custom attribute values, the variant won't be updated. This may be by design (the product configurator typically sets custom values at creation time), but it would be good to document this limitation or handle it.

sale_product_attribute_custom_value_variant/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
alexey-pelykh, line 1: Minor: Missing AGPL-3 license header (other `.py` files in the module include it).

alexey-pelykh (commented):
## OCA Review: `sale_product_attribute_custom_value_variant`

Thank you for this contribution! The module is well-structured and the core logic is clean. Below are my findings.

### Summary

| 1 | minor | `__manifest__.py` | Missing `"installable": True` (OCA convention) |
| 2 | minor | `__manifest__.py` | Missing `"development_status": "Beta"` (README badge says Beta but manifest omits it) |
| 3 | suggestion | `models/sale_order_line.py` | No `write` override -- if a user edits a SO line to add/change custom attribute values after creation, no variant will be created. This may be intentional, but worth documenting the limitation. |
| 4 | nitpick | `__init__.py`, `models/__init__.py` | Missing AGPL-3 license/copyright headers (all other `.py` files in the module have them) |
| 5 | note | `test-requirements.txt` | Dependency on unmerged PR (`OCA/product-attribute#1934`) -- this is expected and documented in PR description, just needs to be resolved before merge. |

### Detailed Notes

**On the `create` override (models/sale_order_line.py:17-23):**
The pattern of calling `super().create()` and then reassigning `line.product_id` works but means the product is written twice -- once during create (with the original product) and again via `_assign_new_custom_variant`. This could trigger recomputation of dependent fields twice. If performance becomes a concern, consider creating the variant before `super().create()` and injecting the correct `product_id` into `vals_list` directly. Not blocking, just something to keep in mind.

**On test coverage (tests/test_sale.py):**
The test covers the happy path well. Consider adding a test for the case where no custom attribute values are provided (to verify the module doesn't interfere with normal SO line creation).

Overall this is a clean, focused module. Nice work! 👍

sale_product_attribute_custom_value_variant/__manifest__.py
rousseldenis: That are the default values, no need to add them

sale_product_attribute_custom_value_variant/models/sale_order_line.py
sbejaoui: > The `create` override handles variant creation nicely. Two observations:
> 
> 1. **Double write on `product_id`**

I don't think this is relevant here, while assigning product_id after create() results in an additional write, it doesn't necessarily mean that all dependent fields are recomputed twice, as odoo handles recomputations lazily

also, for code readability, it's easier to handle this case on a recordset rather than vals_list, so I'd prefer to keep the current flow unless we identify an actual issue caused by the additional write

sbejaoui (approved):

PR #3632, opened 2025-03-19
[MIG] sale_order_line_remove: Migration to 18.0

JochenDeBie:


BhaveshHeliconia (approved):
Funtional test LG.

matteotognini (approved):
LGTM

ivs-cetmix (commented):
Hi @JochenDeBie , please squash administrative commits (if any) with the previous commit for reducing commit noise. Check [the migration wiki](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0) for details, ensue that you have followed all the steps of that document.
Please prefix your pr with the version tag as well: `[18.0][MIG]....`

ivs-cetmix (changes requested):
Thank you!

ivs-cetmix:
Hi @JochenDeBie, commits are good now, please fix the tests.

JochenDeBie:
@ivs-cetmix Not really sure what to do with this one. The test that is failing now was added in another module to specifically check that confirmed SO lines can't be removed, which is the purpose of this module:

https://github.com/OCA/sale-workflow/blame/43f33c42b51eff3b3e3b88f571aec541ce72530f/sale_order_product_recommendation/tests/test_recommendation.py#L207

ivs-cetmix:
> @ivs-cetmix Not really sure what to do with this one. The test that is failing now was added in another module to specifically check that confirmed SO lines can't be removed, which is the purpose of this module:
> 
> https://github.com/OCA/sale-workflow/blame/43f33c42b51eff3b3e3b88f571aec541ce72530f/sale_order_product_recommendation/tests/test_recommendation.py#L207

Looks like we need to use `rebel modules` here. I'm not very familiar with this flow, so probably we need to ask @dreispt or @pedrobaeza  how to do this right.

ArnauCForgeFlow:
Hello, what's the state of this PR? @ivs-cetmix @JochenDeBie I can help if needed.

ivs-cetmix (approved):
Code review LGTM

ivs-cetmix:
> Hello, what's the state of this PR? @ivs-cetmix @JochenDeBie I can help if needed.

Hi @ArnauCForgeFlow , thank you! Could you please check how we can solve an issue with those tests falling?

ArnauCForgeFlow:
@ivs-cetmix @JochenDeBie I think rebel modules are needed here. I don’t have much experience with rebel modules, but I think you need to update the copier answers and include this module as a rebel module.

ivs-cetmix:
That's the

> @ivs-cetmix @JochenDeBie I think rebel modules are needed here. I don’t have much experience with rebel modules, but I think you need to update the copier answers and include this module as a rebel module.

Unfortunately I don't have much experience with them either. Would appreciate if @rousseldenis @sbidoul or @dreispt could give a helper hand here.

sale_order_line_remove/i18n/sale_order_line_remove.pot
```diff
@@ -0,0 +1,31 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* sale_order_line_remove
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 15.0\n"
+"Report-Msgid-Bugs-To: \n"
```
alexey-pelykh, line 8: **Minor**: The `.pot` header references `Odoo Server 15.0`. This should be regenerated for 18.0 to accurately reflect the target version.
```
"Project-Id-Version: Odoo Server 15.0\n"
```
Should be:
```
"Project-Id-Version: Odoo Server 18.0\n"
```

sale_order_line_remove/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Sale Order Line Remove",
+    "version": "18.0.1.0.0",
+    "category": "Sales",
+    "summary": "Allows removal of sale order lines from confirmed "
+    "orders if not invoiced or received",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["sale_stock"],
+    "data": [],
```
alexey-pelykh, line 13: **Nitpick**: OCA convention is to omit the `"data"` key entirely when there are no data files, rather than including an empty list.
```python
"data": [],  # Consider removing this line
```

alexey-pelykh (commented):
## Code Review: `sale_order_line_remove` Migration to 18.0

Overall, this is a clean migration from 16.0 with minimal code changes (only removing the explicit `SaleOrderLine` in `super()` calls, which is correct for Python 3 / Odoo 18). The module structure, manifest, tests, i18n, and readme are all properly set up.

However, I have several observations that should be considered before merging.

### Key Findings

#### 1. `.pot` file references Odoo Server 15.0 (Minor)
The `.pot` header says `Project-Id-Version: Odoo Server 15.0`. For an 18.0 module, this should be regenerated to reference `Odoo Server 18.0`. While not blocking, it shows the `.pot` was not regenerated after migration.

#### 2. `_check_line_unlink` includes `"done"` state -- may be unnecessary in 18.0 (Informational)
In Odoo 18, the base `sale.order.line._check_line_unlink()` only blocks deletion for `state == 'sale'`. Lines in `"done"` state are not blocked by the base method. This module's `removable_lines` filter checks `line.state in ("sale", "done")` -- the `"done"` part may not match any lines from `non_removable_lines` (since the base doesn't include them). This is not a bug (the UserError checks still correctly guard against delivered/invoiced lines in `"done"` state), but it's worth being aware of.

#### 3. CI tests are failing due to cross-module interaction (Blocking for merge)
As discussed in the PR comments, the `sale_order_product_recommendation` module has a test that expects confirmed SO lines to be non-removable (which is the standard Odoo behavior that this module intentionally changes). This requires a "rebel module" configuration in the CI setup. This is a known issue but remains unresolved.

#### 4. `unlink()` override calls `_check_line_unlink()` directly, diverging from Odoo 18 base pattern (Medium)
In Odoo 18's base `sale.order.line.unlink()`, the method calls `_check_line_unlink()` and raises a `UserError` if any non-removable lines are returned. This module's `unlink()` override calls `_check_line_unlink()` but uses the return value differently (to identify which lines need stock move cleanup). Since `_check_line_unlink()` already raises `UserError` for invoiced/delivered lines, the flow is correct, but the pattern is slightly unusual -- the `unlink()` method stores the return value of `_check_line_unlink()` in `non_removable_lines`, but those lines should never actually exist in `self` at that point (because the module's override removes them from the non-removable set). If there ARE truly non-removable lines (e.g., from other modules extending `_check_line_unlink`), the `super().unlink()` call at the end would include them, which would trigger the base Odoo error. This is consistent with the 16.0 version, so not a regression.

#### 5. Empty `"data": []` in manifest (Minor)
The manifest has `"data": []`. OCA convention is to omit the key entirely if empty rather than include an empty list.

#### 6. Test coverage is solid
The tests cover: line unlink check, successful unlink, delivered lines blocking, invoiced lines blocking, and empty picking cleanup. Good coverage for the module's functionality.

### Summary

The code migration itself is correct and minimal. The primary blocker is the CI test failure from `sale_order_product_recommendation`, which requires rebel module configuration. The code-level findings are minor.

BhaveshHeliconia (commented):
This module is already merged: https://github.com/OCA/sale-workflow/tree/18.0/sale_order_line_remove. Could you please close this PR?

PR #3635, opened 2025-03-21
[17.0][IMP] sale_order_carrier_auto_assign: Add delivery_auto_refresh changes

victoralmau:
Add `delivery_auto_refresh` changes

Please @chienandalu can you review it?

@Tecnativa TT52316

### DIFF
--- a/sale_order_carrier_auto_assign/models/res_config_settings.py
+++ b/sale_order_carrier_auto_assign/models/res_config_settings.py
@@ -10,5 +10,8 @@ class ResConfigSettings(models.TransientModel):
     carrier_auto_assign = fields.Boolean(
         related="company_id.carrier_auto_assign",
         readonly=False,
-        help="Enable carrier auto assign on sale order confirmation.",
+    )
+    carrier_on_create = fields.Boolean(
+        related="company_id.carrier_on_create",
+        readonly=False,
     )
--- a/sale_order_carrier_auto_assign/models/sale_order.py
+++ b/sale_order_carrier_auto_assign/models/sale_order.py
@@ -1,20 +1,54 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from odoo import models
+from odoo import api, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    @api.onchange("partner_id", "partner_shipping_id")
+    def _add_delivery_carrier_on_partner_change(self):
+        partner = self.partner_shipping_id or self.partner_id
+        if not partner:
+            return
+        if self.company_id.carrier_on_create:
+            self._set_delivery_carrier(
+                set_delivery_line=False,
+                preserve_order_carrier=False,
+            )
+
+    def _is_auto_set_carrier_on_create(self):
+        self.ensure_one()
+        if self.state not in ("draft", "sent"):
+            return False
+        return self.company_id.carrier_on_create
+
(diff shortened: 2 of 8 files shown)


chienandalu (approved):
Tested along https://github.com/OCA/delivery-carrier/pull/1002 :+1:

sale_order_carrier_auto_assign/models/res_config_settings.py
```diff
@@ -12,3 +12,8 @@ class ResConfigSettings(models.TransientModel):
         readonly=False,
         help="Enable carrier auto assign on sale order confirmation.",
     )
+    carrier_on_create = fields.Boolean(
+        related="company_id.carrier_on_create",
+        readonly=False,
+        help="On the sales quotation, add the shipping method on creation.",
```
rousseldenis: Shouldn't the help put on company level field ?

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -1,31 +1,77 @@
...
+        partner = self.partner_shipping_id or self.partner_id
+        if not partner:
+            return
+        if self.company_id.carrier_on_create:
+            self._set_delivery_carrier(
+                set_delivery_line=False,
+                preserve_order_carrier=False,
+            )
+
+    def _is_auto_set_carrier_on_create(self):
```
rousseldenis, line 20: You can maybe change it to `property`

rousseldenis (commented):
Some remarks

sale_order_carrier_auto_assign/models/sale_order.py
victoralmau: Sorry, I don't understand, you mean adding a `property_x` + `company_dependent=True` style field in `sale.order`? I
 don't understand the need for that, plus in some cases (in some order states) you will always want to return False.

sale_order_carrier_auto_assign/models/res_config_settings.py
victoralmau: I had added the help here as was the other field https://github.com/OCA/sale-workflow/blob/d748de97e455344a7d87c3461e4863af685753a9/sale_order_carrier_auto_assign/models/res_config_settings.py#L13 but ok, no problem, I move it to `res.company`.

sale_order_carrier_auto_assign/models/sale_order.py
pedrobaeza: He means to use Python decorator `@property` for accessing this method like a regular "variable". Example:

```python
@property
def carrier_on_create(self):
```

so you do `if order.carrier_on_create:` to query it. Anyway, let's not change it now for moving this on.

pedrobaeza (approved):
/ocabot merge minor

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -1,31 +1,77 @@
...
+        return orders
+
     def action_confirm(self):
         for rec in self:
             if not rec.company_id.carrier_auto_assign:
                 continue
-            rec._add_delivery_carrier_on_confirmation()
+            rec._set_delivery_carrier(
+                set_delivery_line=True,
+                preserve_order_carrier=False,
```
jbaudoux, line 40: This is wrong, should be only for auto_refresh.
The test `test_sale_order_carrier_auto_assign_carrier_already_set` should fail but setup is not correct, so you don't see it.

I'm pushing a fix for this

PR #3000, opened 2024-03-07
[IMP] backport pr 2592 sale_discount_display_amount

sergiocorato:
Baceport of #2592

PR #3646, opened 2025-03-28
[FIX] sale_discount_display_amount: remove duplicated naming in fields.

legalsylvain:
The module introduces price_total_no_discount and price_subtotal_no_discount fields in 'sale.order' and 'sale.order.line' models. However, the string is the same, so it generates warning in odoo logs, and impossibility to distinguish the fields for end-users

Fixes: #3000

CC original developper / reviewers :  @sergiocorato, @dreispt,  @rousseldenis


rousseldenis (approved):
Trivial. Thanks !

dreispt (approved):

PR #3647, opened 2025-03-28
[REF] sale_discount_display_amount: reduce migration duration.

legalsylvain:
Rational: for the time being, computation is done for all the line of an order, if the order contains a discounted line. Now, it is done only for discounted lines

draft PR.

### DIFF
--- a/sale_discount_display_amount/migrations/12.0.1.2.0/post-compute-subtotal-columns.py
+++ b/sale_discount_display_amount/migrations/12.0.1.2.0/post-compute-subtotal-columns.py
@@ -7,21 +7,20 @@ _logger = logging.getLogger(__name__)
 
 
 def migrate(cr, version):
-    _logger.info("Compute discount columns")
     env = Environment(cr, SUPERUSER_ID, {})
 
+    _logger.info("Compute discount columns sale_order_line.price_subtotal_no_discount")
     query = """
-    update sale_order_line
-    set
-        price_subtotal_no_discount = price_subtotal
-    where discount = 0.0
+    UPDATE sale_order_line
+    SET price_subtotal_no_discount = price_subtotal
+    WHERE discount = 0.0
     """
     cr.execute(query)
 
+    _logger.info("Compute discount columns sale_order.price_subtotal_no_discount")
     query = """
-    update sale_order
-    set
-        price_subtotal_no_discount = amount_untaxed
+    UPDATE sale_order
+    SET price_subtotal_no_discount = amount_untaxed
     """
     cr.execute(query)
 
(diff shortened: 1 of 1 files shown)


alexey-pelykh (commented):
## OCA Review: `sale_discount_display_amount` migration optimization

**Note:** This PR targets the **12.0** branch, which is **end-of-life**. New contributions to EOL branches are generally not accepted by OCA. Additionally, this PR is marked as **draft**.

That said, here are review observations on the change itself:

### Potential correctness concern

The original code computes `_compute_discount()` on **all order lines** for orders that have at least one discounted line:

```python
orders.mapped("order_line")._compute_discount()
```

The new code filters to only discounted lines per order:

```python
lines = order.mapped("order_line").filtered(lambda x: x.discount)
lines._compute_discount()
```

If `_compute_discount()` has any side effect on the **order-level** aggregation (e.g., updating `price_subtotal_no_discount` on the parent `sale.order`), processing only the discounted lines might yield different results than processing all lines at once. The non-discounted lines already got their `price_subtotal_no_discount = price_subtotal` via the SQL UPDATE, so the line-level values are likely fine -- but please confirm that the order-level `price_subtotal_no_discount` is correctly recomputed as the sum of all lines (not just the subset passed to `_compute_discount()`).

### Performance trade-off: loop vs batch

The original code processes all order lines in a **single batch** ORM call. The new code introduces a **per-order loop**:

```python
for i, order in enumerate(orders, start=1):
    lines = order.mapped("order_line").filtered(lambda x: x.discount)
    lines._compute_discount()
```

For databases with many orders containing discounted lines, this loop introduces N separate ORM calls (one per order) instead of one bulk call. The progress logging is valuable for observability, but the per-order iteration could actually be **slower** than the original batch approach due to ORM overhead per iteration (cache invalidation, prefetch resets, etc.).

A middle-ground approach could be to keep the batch operation but filter to only discounted lines:

```python
all_lines = orders.mapped("order_line").filtered(lambda x: x.discount)
all_lines._compute_discount()
```

This preserves the batch efficiency while still reducing the set of lines processed.

### Logging: prefer lazy formatting over f-strings

The new code uses f-strings in logging calls:

```python
_logger.info(f"Found {len(orders)} orders.")
_logger.info(f"{i} / {len(orders)}. order #{order.id}...")
```

Odoo and Python logging best practice is to use lazy `%`-style formatting so the string interpolation is skipped entirely when the log level is disabled:

```python
_logger.info("Found %d orders.", len(orders))
_logger.info("%d / %d. order #%d computing order lines data for %d lines.", i, len(orders), order.id, len(lines))
```

### SQL style changes

The keyword casing normalization (lowercase to UPPERCASE SQL keywords) is fine as a cosmetic improvement.

---

rousseldenis:
@legalsylvain Still draft ?

PR #3648, opened 2025-03-29
[18.0][MIG][REF] sale_triple_discount: Consolidate discount in std field

sbejaoui:
This PR includes a refactoring inspired by https://github.com/OCA/account-invoicing/pull/1638

Introducing a new field, discount1, and using the standard discount field to hold the result of multiple applied discounts makes more sense and reduces code complexity.


legalsylvain:
Thanks for porting this module ! 

> This PR includes a refactoring inspired by https://github.com/OCA/account-invoicing/pull/1638

Did you took a look on this unfinished PR ? https://github.com/OCA/sale-workflow/pull/2885

rousseldenis:
@sbejaoui

sbejaoui:
> Thanks for porting this module !
> 
> > This PR includes a refactoring inspired by [OCA/account-invoicing#1638](https://github.com/OCA/account-invoicing/pull/1638)
> 
> Did you took a look on this unfinished PR ? #2885

No, but I'll have a look to see if there's anything I should take from it

lef-adhoc:
@sbejaoui @legalsylvain 

When there's a pricelist with a discount (visible in the UI) for the selected product on the line, it’s not being reflected in discount1. Also, regarding this, what happens if some discounts are already set and the product is changed? Ideally, it should check if any discount applies from the pricelist and update discount1 accordingly.

Can you check this module? With this PR, they both end up following the same approach.
https://github.com/ingadhoc/sale/blob/18.0/sale_three_discounts/models/sale_order_line.py
In the discount wizard we also added 3 discount fields

sbejaoui:
> @sbejaoui @legalsylvain
> 
> When there's a pricelist with a discount (visible in the UI) for the selected product on the line, it’s not being reflected in discount1. Also, regarding this, what happens if some discounts are already set and the product is changed? Ideally, it should check if any discount applies from the pricelist and update discount1 accordingly.
> 
> Can you check this module? With this PR, they both end up following the same approach. https://github.com/ingadhoc/sale/blob/18.0/sale_three_discounts/models/sale_order_line.py In the discount wizard we also added 3 discount fields

I had the idea to add an inverse to the standard discount field, which resets the three discounts and sets discount1 with the value coming from the inverse. this way, any standard process like pricelists will integrate with this logic out of the box, without extra customization

I will test and get back to you

legalsylvain:
> I had the idea to add an inverse to the standard discount field, which resets the three discounts and sets discount1 with the value coming from the inverse. this way, any standard process like pricelists will integrate with this logic out of the box, without extra customization


Very good idéa ! It should be in the mixin right ?

Let me know if there is something to review.

sbejaoui:
> > I had the idea to add an inverse to the standard discount field, which resets the three discounts and sets discount1 with the value coming from the inverse. this way, any standard process like pricelists will integrate with this logic out of the box, without extra customization
> 
> Very good idéa ! It should be in the mixin right ?
> 
> Let me know if there is something to review.

you can review my last commit

lef-adhoc:
@sbejaoui @legalsylvain @rousseldenis 
The issue is that when any of the `@depends` fields are changed, the inverse method doesn't run — so the discount isn't recalculated or updated in discount1.
If the product, product_uom, or product_uom_qty is modified, the discount should be recalculated and overwrite the previous value if applicable.
image

sbejaoui:
> @sbejaoui @legalsylvain @rousseldenis The issue is that when any of the `@depends` fields are changed, the inverse method doesn't run — so the discount isn't recalculated or updated in discount1. If the product, product_uom, or product_uom_qty is modified, the discount should be recalculated and overwrite the previous value if applicable. 

I get what you mean. I added the compute method and ensured that the discount is recomputed when the pricelist discount applies and any of its dependencies change, check the unit test:

https://github.com/OCA/sale-workflow/blob/dfb33eb6d9231daf3db139a3847dc97bbb49e03e/sale_triple_discount/tests/test_sale_triple_discount.py#L263-L307

sbejaoui:
@legalsylvain , @rousseldenis 

how to fix:

`odoo.exceptions.UserError: Modules "Account Invoice Triple Discount" and "Account Fixed Discount" are incompatible.`

matteotognini:
Hi @sbejaoui,  could you please replace the unreleased dependencies with https://github.com/OCA/account-invoicing/pull/2019. I think that should fix the error you mentioned, as you can see in the PR conversation. Thanks!

rousseldenis:
> @legalsylvain , @rousseldenis
> 
> how to fix:
> 
> `odoo.exceptions.UserError: Modules "Account Invoice Triple Discount" and "Account Fixed Discount" are incompatible.

Could you add a commit to put `sale_triple_discount` and `sale_fixed_discount` in rebel modules  (with copier)?

sbejaoui:
> > @legalsylvain , @rousseldenis
> > how to fix:
> > `odoo.exceptions.UserError: Modules "Account Invoice Triple Discount" and "Account Fixed Discount" are incompatible.
> 
> Could you add a commit to put `sale_triple_discount` and `sale_fixed_discount` in rebel modules (with copier)?

thanks, 

now it fails at sale_order_product_availability_inline

rousseldenis:
> > > @legalsylvain , @rousseldenis
> > > how to fix:
> > > `odoo.exceptions.UserError: Modules "Account Invoice Triple Discount" and "Account Fixed Discount" are incompatible.
> > 
> > 
> > Could you add a commit to put `sale_triple_discount` and `sale_fixed_discount` in rebel modules (with copier)?
> 
> thanks,
> 
> now it fails at sale_order_product_availability_inline

@sbejaoui Please see : https://github.com/OCA/sale-workflow/pull/3765

lef-adhoc:
@sbejaoui There is an issue when changing the product quantity and a price rule is applied. The expected behavior is to set discounts 2 and 3 to 0 (which works correctly), but discount 1 should be set to the value defined by the price rule (this is how Odoo works natively—when a rule is applied, it overrides discount with the value from the pricelist).

Video:
https://drive.google.com/file/d/1tiaL8jZSlo7OboAQf0ecx3swjXG7OOrK/view

legalsylvain:
@lef-adhoc thanks a lot for your detailed comment.

sergioM-S73:
@sbejaoui pls resolve confilcts

sbejaoui:
> @sbejaoui There is an issue when changing the product quantity and a price rule is applied. The expected behavior is to set discounts 2 and 3 to 0 (which works correctly), but discount 1 should be set to the value defined by the price rule (this is how Odoo works natively—when a rule is applied, it overrides discount with the value from the pricelist).
> 
> Video: https://drive.google.com/file/d/1tiaL8jZSlo7OboAQf0ecx3swjXG7OOrK/view

Thanks for the test and good catch — that was a tricky bug.
I’ve pushed a fix, could you please test it again and let me know if it works as expected?


https://github.com/user-attachments/assets/9c2f416d-0a20-4bf3-a3cc-92779a7e786e

lef-adhoc:
@sbejaoui 
Hi! I've been testing the feature and it works great, nice job! 👍

I wanted to propose an enhancement: what do you think about integrating the three-discount calculation into the discount wizard? I believe it would improve the user experience.

We have actually developed a working implementation of this concept in our own repository. Here is the code for the wizard, which could serve as a reference:
https://github.com/ingadhoc/sale/blob/18.0/sale_three_discounts/wizards/sale_order_discount.py

Let me know what you think!
image

rousseldenis:
> @sbejaoui Hi! I've been testing the feature and it works great, nice job! 👍
> 
> I wanted to propose an enhancement: what do you think about integrating the three-discount calculation into the discount wizard? I believe it would improve the user experience.
> 
> We have actually developed a working implementation of this concept in our own repository. Here is the code for the wizard, which could serve as a reference: https://github.com/ingadhoc/sale/blob/18.0/sale_three_discounts/wizards/sale_order_discount.py
> 
> Let me know what you think! image

@lef-adhoc Interesting, could you maybe make a PR on this branch in acsone repository with the improvement?

sbejaoui:
> @sbejaoui Hi! I've been testing the feature and it works great, nice job! 👍
> 
> I wanted to propose an enhancement: what do you think about integrating the three-discount calculation into the discount wizard? I believe it would improve the user experience.
> 
> We have actually developed a working implementation of this concept in our own repository. Here is the code for the wizard, which could serve as a reference: https://github.com/ingadhoc/sale/blob/18.0/sale_three_discounts/wizards/sale_order_discount.py
> 
> Let me know what you think! ![image](https://private-user-images.githubusercontent.com/148369234/463788333-

Thanks for the feedback, glad it’s working well!
The enhancement sounds great, but since it’s not an immediate customer requirement for us, it’s not planned for the near term
I suggest we merge this as is and open a separate PR for the improvement you proposed

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,158 @@
...
+            rec.discount = rec._get_final_discount()
+        return res
+
+    def _inverse_discount(self):
+        for rec in self:
+            rec.update({"discount1": rec.discount, "discount2": 0, "discount3": 0})
+
+    @api.depends("discount")
+    def _compute_discounts(self):
+        super()._compute_discount()
```
ferran-S73, line 128: You're calling a super for a different method, is this intended?

sbejaoui: yes, I will add a comment to explain why, when the discount is changed directly or via its compute method, it triggers the compute methods for the three discount fields. At this point, we need to make sure the cache is coherent before calling the inverse. calling _compute_discount() directly would alter the discount value by recomputing it from the three discounts, which may still be outdated, calling super() on the other hand avoids that

ferran-S73:
@sbejaoui I've encountered a case in which the computed discount is lost.

Steps to reproduce:
 - Configure a pricelist and a pricelist item with a disocunt.
 - Set the pricelist in a partner.
 - Import via xlsx a sale order for said partner and pricelist, with a product that is included in our pricelist item.

The new sale order line created is created without any discount. I've done some code debugging and this is due to `_compute_discounts` not being triggered when sale module sets the discount from the pricelist. Thus reseting every discount to in `_get_final_discount`.

This also happens when creating a sale line from the web store, or anytime a line is created directly from calling the create method without setting any specific discount (you can try it from a server action for example).

sbejaoui:
> @sbejaoui I've encountered a case in which the computed discount is lost.
> 
> Steps to reproduce:
> 
> * Configure a pricelist and a pricelist item with a disocunt.
> * Set the pricelist in a partner.
> * Import via xlsx a sale order for said partner and pricelist, with a product that is included in our pricelist item.
> 
> The new sale order line created is created without any discount. I've done some code debugging and this is due to `_compute_discounts` not being triggered when sale module sets the discount from the pricelist. Thus reseting every discount to in `_get_final_discount`.
> 
> This also happens when creating a sale line from the web store, or anytime a line is created directly from calling the create method without setting any specific discount (you can try it from a server action for example).

Thank you for the test, could you please share the excel you used for the import

ferran-S73:
> > @sbejaoui I've encountered a case in which the computed discount is lost.
> > Steps to reproduce:
> > 
> > * Configure a pricelist and a pricelist item with a disocunt.
> > * Set the pricelist in a partner.
> > * Import via xlsx a sale order for said partner and pricelist, with a product that is included in our pricelist item.
> > 
> > The new sale order line created is created without any discount. I've done some code debugging and this is due to `_compute_discounts` not being triggered when sale module sets the discount from the pricelist. Thus reseting every discount to in `_get_final_discount`.
> > This also happens when creating a sale line from the web store, or anytime a line is created directly from calling the create method without setting any specific discount (you can try it from a server action for example).
> 
> Thank you for the test, could you please share the excel you used for the import

Sure! Please keep in mind that I'm using some external ids and the current configuration in the runboat but that may be lost if you do another push as it is regenarated

[Sales Order (sale.order)(2).xlsx](https://github.com/user-attachments/files/21141085/Sales.Order.sale.order.2.xlsx)

ferran-S73:
> > > @sbejaoui I've encountered a case in which the computed discount is lost.
> > > Steps to reproduce:
> > > 
> > > * Configure a pricelist and a pricelist item with a disocunt.
> > > * Set the pricelist in a partner.
> > > * Import via xlsx a sale order for said partner and pricelist, with a product that is included in our pricelist item.
> > > 
> > > The new sale order line created is created without any discount. I've done some code debugging and this is due to `_compute_discounts` not being triggered when sale module sets the discount from the pricelist. Thus reseting every discount to in `_get_final_discount`.
> > > This also happens when creating a sale line from the web store, or anytime a line is created directly from calling the create method without setting any specific discount (you can try it from a server action for example).
> > 
> > 
> > Thank you for the test, could you please share the excel you used for the import
> 
> Sure! Please keep in mind that I'm using some external ids and the current configuration in the runboat but that may be lost if you do another push as it is regenarated
> 
> [Sales Order (sale.order)(2).xlsx](https://github.com/user-attachments/files/21141085/Sales.Order.sale.order.2.xlsx)

Hi @sbejaoui , any news on this issue? We're currently using this pr in some clients and would like to know if you plan on giving a solution soon. It works perfectly aside from that btw.
Let me know if there's something I can do to help you solve this.

lef-adhoc:
Hi @sbejaoui , any progress on this ? If there is anything I can help just let me know.

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,163 @@
...
+            "discount2_limit",
+            "CHECK (discount2 <= 100.0)",
+            "Discount 2 must be lower or equal than 100%.",
+        ),
+        (
+            "discount3_limit",
+            "CHECK (discount3 <= 100.0)",
+            "Discount 3 must be lower or equal than 100%.",
+        ),
+    ]
```
akash-aipl, line 148: I was wondering if we could consider replacing the current _sql_constraints with validation logic in create and write methods.

The reason is that using SQL constraints can cause issues during module installation or migration if there are already records with discount values above 100%. PostgreSQL would block the constraint from being added in such cases.

An alternative could be to move the validation to a helper method and call it in create and write, something like:

```

    def _check_discount_limits(self, vals):
        for field in self._discount_fields():
            if field in vals and vals[field] > 100.0:
                raise ValidationError(
                    _(
                        f"{field.replace('discount', 'Discount ').title()} must be lower than 100%."
                    )
                )

    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            self._check_discount_limits(vals)

    def write(self, vals):
        self._check_discount_limits(vals)
        return super().write(vals)

```

This would avoid migration issues while still ensuring the validation for new or updated records.
Would this approach work ?

lef-adhoc:
@sbejaoui Hi, , are you going to be able to continue with the PR?

lef-adhoc:
@sbejaoui Hi, any news on this?

pablo-cort-s73:
Superseded by #3889

rousseldenis:
@sbejaoui

PR #3468, opened 2024-12-06
[18.0][MIG] sale_stock_picking_note: Migration to 18.0

Naraka:
Migration to 18.0

PR #3651, opened 2025-03-31
[18.0][MIG] sale_stock_picking_note: Migration to 18.0

victoralmau:
Supersedes https://github.com/OCA/sale-workflow/pull/3468

Migration to 18.0

Please @chienandalu and @carlosdauden can you review it?

@Tecnativa TT54496


JordiMForgeFlow (approved):

sale_stock_picking_note/models/sale_order.py
```diff
@@ -25,7 +25,7 @@ class SaleOrder(models.Model):
     @api.depends("partner_id", "partner_shipping_id")
     def _compute_picking_notes(self):
         for order in self:
-            if order.state not in {"sale", "done", "cancel"}:
+            if order.state not in {"sale", "cancel"}:
```
rousseldenis, line 28: Shouldn't we take into account the `locked` field ?

victoralmau: IMO it does not change the condition, i.e., if it is `locked` it will already be in the `sale` state.

carlosdauden (approved):

PR #3653, opened 2025-03-31
[IMP] sale_validity_auto_cancel: add warnings for sale orders to be auto cancelled

AlexPForgeFlow:
Adds possibility to define a default validity warning of sale orders.
A warning will be displied if sale order is modified and was going to be auto cancelled in X days.
It is introduced a setting called 'Default Validity Warning of Sale Orders' to define the threshold of number of days for the warning.

### DIFF
--- a/sale_validity_auto_cancel/models/res_company.py
+++ b/sale_validity_auto_cancel/models/res_company.py
@@ -1,7 +1,8 @@
 # Copyright 2023 ForgeFlow S.L.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResCompany(models.Model):
@@ -13,6 +14,16 @@ class ResCompany(models.Model):
         help="Quotations will be cancelled after the specified number of"
         " days since the expiration date.",
     )
+    sale_validity_warning_days = fields.Integer(
+        string="Default Validity Warning of Sale Orders",
+        default=0,
+        help="By default, the validity date warning of sale orders will be "
+        "the number of days defined on this field before the date of the "
+        "validity sale order. If the value of this field is 0, the sale orders "
+        "will not have a validity warning by default.",
+    )
+
+    sale_validity_warning_enabled = fields.Boolean(default=False)
 
     _sql_constraints = [
         (
@@ -21,4 +32,25 @@ class ResCompany(models.Model):
             "The value of the field 'Auto-cancel expired quotations after' "
             "must be positive or 0.",
         ),
+        (
+            "sale_order_validity_warning_days_positive",
+            "CHECK (sale_validity_warning_days >= 0)",
+            "The value of the field 'Default Validity Warning of Sale Orders' "
+            "must be positive or 0.",
+        ),
     ]
+
+    @api.constrains("sale_validity_warning_days", "sale_validity_auto_cancel_days")
+    def _check_warning_vs_auto_cancel(self):
+        for rec in self:
+            if (
+                rec.sale_validity_warning_days
+                and rec.sale_validity_auto_cancel_days
+                and rec.sale_validity_warning_days >= rec.sale_validity_auto_cancel_days
+            ):
+                raise ValidationError(
+                    _(
+                        "The 'Default Validity Warning of Sale Orders' must be less than "
+                        "'Auto-cancel expired quotations after (days)'."
+                    )
+                )
--- a/sale_validity_auto_cancel/models/sale_order.py
+++ b/sale_validity_auto_cancel/models/sale_order.py
@@ -4,13 +4,21 @@ import logging
 
 from dateutil.relativedelta import relativedelta
 
-from odoo import fields, models
+from odoo import _, api, fields, models
 
 _logger = logging.getLogger(__name__)
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
+
+    validity_date_warning_message = fields.Text(
+        compute="_compute_validity_date_warning_message",
+    )
+
+    sale_validity_warning_enabled = fields.Boolean(
+        related="company_id.sale_validity_warning_enabled", readonly=False,
+    )
 
     def _get_expired_order_states(self):
         # Can be inherited to exclude/include order states
@@ -31,3 +39,33 @@ class SaleOrder(models.Model):
                     order.with_context(company_id=company.id).action_cancel()
                 except Exception as e:
                     _logger.error("Failed to auto-cancel %s: %s" % (order.name, str(e)))
+
+    @api.depends("validity_date", "company_id")
+    def _compute_validity_date_warning_message(self):
+        for order in self:
+            order.validity_date_warning_message = False
+            company = order.company_id or self.env["res.company"]._company_default_get(
+                "sale.order"
+            )
+            if company.sale_validity_warning_days:
+                today = fields.Date.today()
+                expired_states = order._get_expired_order_states()
+                auto_cancel_date = order.validity_date + relativedelta(
+                    days=company.sale_validity_auto_cancel_days
+                )
+                warning_date = auto_cancel_date - relativedelta(
+                    days=company.sale_validity_warning_days
+                )
+                if (
+                    order.state in expired_states
+                    and auto_cancel_date > today > warning_date
+                ):
+                    days = int((auto_cancel_date - today).days)
+                    order.validity_date_warning_message = (
+                        _(
+                            "This Quotation will be automatically cancelled in %s days.\n"
+                            "If needed, the expiration date (found on the 'Other info' tab) "
+                            "can be updated to a future date."
+                        )
+                        % days
+                    )
--- a/sale_validity_auto_cancel/readme/DESCRIPTION.rst
+++ b/sale_validity_auto_cancel/readme/DESCRIPTION.rst
@@ -1,3 +1,5 @@
 This module adds a scheduled action that automatically cancels quotations after their expiration date.
+Also adds the possibility to diplay a warning before quotation will be automatically canceled.
 
 A company setting can be modified to decide how many days after the expiration date the quotations are automatically cancelled.
+There is also one company setting to enable the warnings, and define how many days before automatic cancellation the warning will be shown.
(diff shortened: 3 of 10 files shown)


sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -4,14 +4,22 @@
...
-from odoo import fields, models
+from odoo import fields, models, api, _
 
 _logger = logging.getLogger(__name__)
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    warning_validity_date = fields.Date(
```
JordiMForgeFlow: no need to store this one, you can evaluate it in a method

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -4,14 +4,22 @@
...
+        compute="_compute_warning_validity_date",
+    )
+
+    warning_message_validity_date = fields.Text(
```
JordiMForgeFlow:
```suggestion
    validity_date_warning_message = fields.Text(
```

sale_validity_auto_cancel/models/res_company.py
```diff
@@ -13,6 +14,13 @@ class ResCompany(models.Model):
         help="Quotations will be cancelled after the specified number of"
         " days since the expiration date.",
     )
+    default_sale_order_validity_warning_days = fields.Integer(
```
JordiMForgeFlow:
```suggestion
    sale_order_validity_warning_days = fields.Integer(
```

As it will be a generic value it is not really a default because you will not be able to change it anywhere else

sale_validity_auto_cancel/views/res_config_settings.xml
```diff
@@ -25,6 +25,19 @@
                         </div>
                     </div>
                 </div>
+                <div class="col-12 col-lg-6 o_setting_box">
+                    <div class="o_setting_left_pane" />
+                    <div class="o_setting_right_pane">
+                        <label for="sale_validity_warning_days" />
+                        <div class="text-muted mt4 mb4">
+                            Warning will be shown for quotations that will be automatically cancelled in the expiration date minus of days specified.
```
JordiMForgeFlow:
```suggestion
                            Number of days before automatic cancellation to display a warning on quotations.
```

sale_validity_auto_cancel/views/sale_order.xml
```diff
@@ -0,0 +1,30 @@
...
+<odoo>
+    <record model="ir.ui.view" id="view_order_form_warning">
+        <field name="name">sale.order.form (validity_warn)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <header position="after">
+                <field
+                    name="sale_validity_warning_enabled"
+                    attrs="{'invisible': True}"
```
JordiMForgeFlow:
```suggestion
                    invisible="1"
```

JordiMForgeFlow (commented):
minor comment, the rest LGTM

JordiMForgeFlow (approved):

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -31,3 +39,33 @@ def cron_sale_validity_auto_cancel(self):
...
+                    and auto_cancel_date > today > warning_date
+                ):
+                    days = int((auto_cancel_date - today).days)
+                    order.validity_date_warning_message = (
+                        _(
+                            "This Quotation will be automatically cancelled in %s days.\n"
+                            "If needed, the expiration date (found on the 'Other info' tab) "
+                            "can be updated to a future date."
+                        )
+                        % days
```
rousseldenis, line 70: Please use translation function arguments instead.

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -31,3 +39,33 @@ def cron_sale_validity_auto_cancel(self):
...
+
+    @api.depends("validity_date", "company_id")
+    def _compute_validity_date_warning_message(self):
+        for order in self:
+            order.validity_date_warning_message = False
+            company = order.company_id or self.env["res.company"]._company_default_get(
+                "sale.order"
+            )
+            if company.sale_validity_warning_days:
+                today = fields.Date.today()
```
alexey-pelykh, line 51: **Critical:** `validity_date` can be `False` (it is not a required field on `sale.order`). When it is `False`, the expression `order.validity_date + relativedelta(days=...)` will raise a `TypeError`.

Please add a guard:
```python
if not order.validity_date:
    continue
```
before accessing `order.validity_date`.

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -31,3 +39,33 @@ def cron_sale_validity_auto_cancel(self):
                     order.with_context(company_id=company.id).action_cancel()
                 except Exception as e:
                     _logger.error("Failed to auto-cancel %s: %s" % (order.name, str(e)))
+
```
alexey-pelykh, line 42: **Missing dependency:** The compute method checks `order.state in expired_states` (line 57), but `state` is not listed in `@api.depends`. This means the warning message will not recompute when the order state changes (e.g., from `draft` to `sale`).

Should be:
```python
@api.depends("validity_date", "company_id", "state")
```

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -31,3 +39,33 @@ def cron_sale_validity_auto_cancel(self):
...
+                    days=company.sale_validity_auto_cancel_days
+                )
+                warning_date = auto_cancel_date - relativedelta(
+                    days=company.sale_validity_warning_days
+                )
+                if (
+                    order.state in expired_states
+                    and auto_cancel_date > today > warning_date
+                ):
+                    days = int((auto_cancel_date - today).days)
```
alexey-pelykh, line 63: As @rousseldenis noted, please use named arguments for the translation function instead of `%` string formatting. This is the OCA standard and avoids issues with translator tools:

```python
order.validity_date_warning_message = _(
    "This Quotation will be automatically cancelled in %(days)s days.\n"
    "If needed, the expiration date (found on the 'Other info' tab) "
    "can be updated to a future date.",
    days=days,
)
```

Note: On Odoo 13.0, `_()` does not support keyword arguments natively. The alternative for v13 is:
```python
_("...%s days...") % days
```
which is what you have. However, for future portability, consider using `.format()` or `%()` style. Either way, this review comment from @rousseldenis is still open and should be addressed or discussed.

sale_validity_auto_cancel/models/sale_order.py
```diff
@@ -4,14 +4,22 @@
...
 _logger = logging.getLogger(__name__)
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    validity_date_warning_message = fields.Text(
+        compute="_compute_validity_date_warning_message",
+    )
+
```
alexey-pelykh, line 18: **Design question:** Having `sale_validity_warning_enabled` as a `related` field with `readonly=False` on `sale.order` means any user editing a sale order can toggle the company-level setting. This is likely unintentional -- the field should be `readonly=True` (or simply omit `readonly=False` since `True` is the default for related fields) so it's only used to read the company setting for the view's `attrs` logic.

sale_validity_auto_cancel/views/res_config_settings.xml
```diff
@@ -25,6 +25,44 @@
                         </div>
                     </div>
                 </div>
+                <div
+                    class="col-12 col-lg-6 o_setting_box"
```
alexey-pelykh, line 29: The `id` attribute `sale_config_online_confirmation_sign` appears to be copy-pasted from another setting block. It is misleading and could conflict with other modules. Please rename to something meaningful like `sale_validity_warning_setting`.

sale_validity_auto_cancel/tests/test_sale_validity_auto_cancel.py
```diff
@@ -33,3 +33,16 @@ def create_so(self):
         }
         so = self.env["sale.order"].create(vals)
         return so
+
+    def test_sale_validity_auto_cancel_warning(self):
```
alexey-pelykh, line 37: The test never sets `company.sale_validity_warning_enabled = True`. While the current compute method does not check this boolean (it only checks `sale_validity_warning_days`), this means the test does not validate the full user-facing flow where the toggle must be enabled. Please either:
1. Add `company.sale_validity_warning_enabled = True` to align with the UI flow, or
2. Add a separate test asserting the warning is empty when the toggle is disabled.

sale_validity_auto_cancel/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,5 @@
 This module adds a scheduled action that automatically cancels quotations after their expiration date.
+Also adds the possibility to diplay a warning before quotation will be automatically canceled.
```
alexey-pelykh, line 2: Typo: `diplay` should be `display`.

alexey-pelykh (changes requested):
## OCA Community Review

Thank you for adding the warning feature to `sale_validity_auto_cancel`. The concept is useful -- notifying users before auto-cancellation is a good UX improvement. However, there are several issues that should be addressed before this can be merged.

**Note:** This PR targets the **13.0** branch, which has reached End-of-Life. Consider whether a forward-port to a supported branch (16.0+) would be more appropriate.

### Summary of Findings

| 1 | Critical | `models/sale_order.py` | `validity_date` can be `False`; arithmetic on it will raise `TypeError` |
| 2 | Major | `models/sale_order.py` | `state` is used in compute but missing from `@api.depends` |
| 3 | Major | `models/sale_order.py` | Unresolved review: use `_()` named arguments instead of `%` formatting |
| 4 | Moderate | `models/sale_order.py` | `sale_validity_warning_enabled` related field with `readonly=False` on sale order is unusual |
| 5 | Moderate | `views/sale_order.xml` | Misleading XML `id` copied from another setting (`sale_config_online_confirmation_sign`) |
| 6 | Minor | `models/res_company.py` | License header is LGPL-3.0 but new view file uses AGPL-3.0 -- should be consistent |
| 7 | Minor | `tests/test_sale_validity_auto_cancel.py` | Test does not set `sale_validity_warning_enabled = True` |
| 8 | Minor | `readme/DESCRIPTION.rst` | Typo: "diplay" should be "display" |

PR #3655, opened 2025-04-01
[17.0][FIX] sale_product_multi_add: avoid description overwrite

MateuGForgeFlow:
Current behavior of the wizard overwrites the Description of the sale order line (removing internal reference)

image

After fix
image


LoisRForgeFlow (approved):
Keeping aside the fact that this module is overlapping with the standard "Catalog" feature and focusing on just the fix: LGTM

legalsylvain (approved):

PR #3658, opened 2025-04-01
[16.0][IMP] sale_order_product_recommendation: create sale lines with write

Tisho99:
This is intended for also triggering inheritances of the sale order's write method, for example: https://github.com/OCA/l10n-spain/blob/c5a84391e3e2ff87cbbc99f606f386798ea4cfad/l10n_es_sigaus_sale/models/sale_order.py#L48-L76


pedrobaeza:
The patch should be done in the other module, as it should cover `create` method as well.

Tisho99:
Okay @pedrobaeza, I'll do it in the other module

Thank you

PR #3664, opened 2025-04-03
[ADD] telegram_notification: Add module for sending Telegram notifica…

AnmollGarg:
This PR introduces the **telegram_notification** module, which allows sending Telegram notifications for sales order events such as confirmation, quotation sending, and cancellation. Key features include:

- Configurable Telegram bots with customizable message templates.
- Event-based notifications for sales orders.
- Dynamic placeholders for personalized messages.
- Multi-chat support for broadcasting notifications.

This module is compatible with Odoo 16.0.


telegram_notification/models/telegram_notification.py
```diff
@@ -0,0 +1,136 @@
+import requests
+import logging
+from odoo import models, fields, api  # noqa
+
+logger = logging.getLogger(__name__)
+
+
+class TelegramNotification(models.Model):
+    _name = "telegram.notification"
```
rousseldenis, line 9: IMHO, a base module should be done in order to be able to reuse behavior on other flows.

rousseldenis:
@AnmollGarg What do you think of ?

pedrobaeza:
Isn't there some overlapping with `mail_gateway_telegram` here?

telegram_notification/models/telegram_notification.py
syntax-sensei: Just to confirm, are you suggesting an abstract model for logic reuse??

syntax-sensei: > IMHO, a base module should be done in order to be able to reuse behavior on other flows.

Just to confirm, are you suggesting an abstract model for logic reuse??

syntax-sensei:
> Isn't there some overlapping with `mail_gateway_telegram` here?

`mail_gateway_telegram` handles chat with customers, while this sends outbound alerts to internal teams on sales events.

telegram_notification/models/telegram_notification.py
rousseldenis: Not an abstract model but putting this model in a separate module in https://github.com/OCA/social/ should be better

telegram_notification/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
...
+                return f"🎉 Sales Order {self.name} has been confirmed!"
+            elif event_type == "quotation_sent":
+                return f"📄 Quotation {self.name} has been sent"
+            elif event_type == "sale_canceled":
+                return f"❌ Sales Order {self.name} has been canceled"
+            return f"Notification about Sales Order {self.name}"
+
+        # Replace placeholders like ${order.name} with actual values
+        def replace_placeholder(match):
+            expr = match.group(1)
```
alexey-pelykh, line 96: **Security: Arbitrary attribute traversal via `getattr()` chain.**

This pattern allows a user who can edit message templates to traverse the entire Odoo ORM graph via `getattr()`. For example, `${order.env.cr}` or `${order.env['ir.config_parameter'].sudo().get_param('database.secret')}` -- while the regex restricts to dotted attribute access (no method calls with parentheses), it still exposes any readable attribute on any reachable object.

Use Odoo's built-in `mail.render.mixin` / QWeb templating with sandboxed evaluation, or at minimum whitelist a set of allowed field paths.

telegram_notification/models/sale_order_notification.py
```diff
@@ -0,0 +1,53 @@
+from odoo import models, fields, api
```
alexey-pelykh, line 1: **This entire file is a duplicate/conflicting notification mechanism.**

The `sale.order.notification.log` model uses `ir.config_parameter` for bot token and chat ID, while `telegram.notification` model stores them as model fields. Having two independent implementations of the same feature in one module is confusing and error-prone.

Additionally, this model has no access rules in `ir.model.access.csv`, which will cause `AccessError` at runtime.

Remove this file and consolidate into the `telegram.notification` approach.

telegram_notification/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
+from odoo import models, fields, api, _  # noqa
```
alexey-pelykh, line 1: **Import ordering violates OCA standards.**

Odoo/OCA coding standards require: (1) stdlib, (2) third-party, (3) odoo imports.

Should be:
```python
import logging
import re

from odoo import _, api, fields, models
```

Also, `api` and `fields` are imported but not used in this file. Remove unused imports instead of using `# noqa`.

telegram_notification/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
...
+        res = super(SaleOrder, self).action_quotation_send()
+        self._send_telegram_notification("quotation_sent")
+        return res
+
+    def _send_telegram_notification(self, event_type):
+        """Send a notification to Telegram with better logging"""
+        _logger = logging.getLogger(__name__)
+
+        _logger.info(
+            "Preparing to send %s notification for order %s", event_type, self.name
```
alexey-pelykh, line 33: **Redundant logger redefinition.** `_logger` is already defined at module level (line 5). This local redefinition is unnecessary and shadows the module-level logger.

telegram_notification/models/sale_order_notification.py
```diff
@@ -0,0 +1,53 @@
...
+from odoo.tools.translate import _
+
+
+logger = logging.getLogger(__name__)
+
+
+class SaleOrderNotificationLog(models.Model):
+    _name = "sale.order.notification.log"
+    _description = _("Log for Sale Order Telegram Notifications")
+
```
alexey-pelykh, line 14: **Do not use `_()` in `_description` or field `string` parameters.**

Odoo automatically extracts these for translation. Wrapping them in `_()` is incorrect and can cause issues with the translation extraction tooling.

```python
# Wrong:
_description = _("Log for Sale Order Telegram Notifications")
string=_("Sale Order")

# Correct:
_description = "Log for Sale Order Telegram Notifications"
string="Sale Order"
```

telegram_notification/models/telegram_notification.py
```diff
@@ -0,0 +1,136 @@
+import requests
+import logging
+from odoo import models, fields, api  # noqa
```
alexey-pelykh, line 3: **Import ordering and unused import.**

Stdlib (`logging`) should come before third-party (`requests`) which should come before odoo imports. Also, `api` is imported but never used -- remove it.

```python
import logging

import requests

from odoo import fields, models
```

telegram_notification/views/telegram_notification_views.xml
```diff
@@ -0,0 +1,89 @@
...
+                                <field name="sale_confirmed_template"
+                                    placeholder="Template for confirmed sales orders"
+                                    widget="text_markdown"
+                                    attrs="{'required': [('enabled_event_sale_confirmed', '=', True)]}" />
+                                <field name="quotation_sent_template"
+                                    placeholder="Template for sent quotations"
+                                    widget="text_markdown"
+                                    attrs="{'required': [('enabled_event_quotation_sent', '=', True)]}" />
+                                <field name="sale_canceled_template"
+                                    placeholder="Template for canceled sales orders"
```
alexey-pelykh, line 37: **`widget="text_markdown"` does not exist in standard Odoo 16.**

This will cause a client-side error. Use `widget="text"` or remove the widget attribute entirely for a plain textarea.

telegram_notification/views/telegram_notification_views.xml
```diff
@@ -0,0 +1,89 @@
...
+    <record id="action_telegram_notification" model="ir.actions.act_window">
+        <field name="name">Telegram Notifications</field>
+        <field name="res_model">telegram.notification</field>
+        <field name="view_mode">list,form</field>
+        <field name="help" type="html">
+            <p class="o_view_nocontent_smiling_face">
+                Create your first Telegram notification configuration
+            </p>
+        </field>
+    </record>
```
alexey-pelykh, line 83: **Root-level menu item is not appropriate.**

OCA modules should not create top-level menu entries. This should be placed under an existing menu, such as the Sales configuration menu (`sale.sale_menu_root`) or Settings.

telegram_notification/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
...
+                return f"📄 Quotation {self.name} has been sent"
+            elif event_type == "sale_canceled":
+                return f"❌ Sales Order {self.name} has been canceled"
+            return f"Notification about Sales Order {self.name}"
+
+        # Replace placeholders like ${order.name} with actual values
+        def replace_placeholder(match):
+            expr = match.group(1)
+            try:
+                # Start with self (the order) as the base object
```
alexey-pelykh, line 98: **Use `%s` formatting in logger calls, not f-strings.**

Logger calls should use lazy `%s` formatting so the string interpolation is skipped if the log level is not active:

```python
# Wrong:
_logger.error(f"Error processing placeholder ${{{expr}}}: {str(e)}")

# Correct:
_logger.error("Error processing placeholder ${%s}: %s", expr, e)
```

telegram_notification/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
...
+        res = super(SaleOrder, self).action_cancel()
+        # Send notification after cancellation is complete
+        self._send_telegram_notification("sale_canceled")
+        return res
+
+    def action_quotation_send(self):
+        res = super(SaleOrder, self).action_quotation_send()
+        self._send_telegram_notification("quotation_sent")
+        return res
+
```
alexey-pelykh, line 27: **`action_quotation_send()` returns a wizard action.**

In Odoo 16, this method opens a mail compose wizard. The notification is sent here before the user actually sends the quotation -- if the user cancels the wizard, a false notification will have been sent.

The notification should be triggered from the actual send action (e.g., override `action_send_mail` on the wizard or use a mail post-hook), not when the wizard opens.

telegram_notification/tests/test_telegram_notification.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @patch("requests.post")
+    def test_send_message_success(self, mock_post):
+        mock_post.return_value.status_code = 200
+        result = self.notification.send_message("Test message")
+        self.assertTrue(result)
+        mock_post.assert_called_once()
+
+    @patch("requests.post")
+    def test_send_message_failure(self, mock_post):
```
alexey-pelykh, line 25: **Test assertion will fail.**

The `send_message` code tries HTML first (one `requests.post` call). On 400 status, it retries with plain text (second `requests.post` call). So `mock_post` will be called **twice**, not once. `assert_called_once()` will fail.

Also: missing `tests/__init__.py` means this test file will never be discovered by Odoo's test runner.

telegram_notification/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
```
alexey-pelykh, line 1: **Module naming concern.**

The name `telegram_notification` is too generic for a sale-specific module. If the base Telegram functionality is split to OCA/social (as recommended by the maintainer), this sale-specific glue module should be named `sale_telegram_notification` or `sale_order_telegram`.

Also: the `"demo": []` key is unnecessary when empty -- just omit it.

alexey-pelykh (changes requested):
Thank you for the contribution! The idea of Telegram notifications for sales events is interesting. However, there are several architectural, security, code quality, and OCA compliance issues that need to be addressed before this can be merged.

## Architecture

As @rousseldenis already noted, the base Telegram integration model (`telegram.notification`) should live in a separate module in **OCA/social**, not in sale-workflow. The sale-workflow repo should only contain the sale-specific glue module that depends on a generic `base_telegram` (or similar) module. This is a fundamental OCA design principle: generic functionality should be reusable across domains.

## Critical Issues

### 1. Security: Arbitrary attribute traversal (RCE-adjacent)
The `_generate_notification_message` method in `sale_order.py` uses `getattr()` in a loop to traverse arbitrary object attributes from user-supplied template placeholders (`${order.partner_id.name}`). While the regex restricts to `order.` prefix, this still allows traversal to sensitive model attributes (e.g., `${order.env.cr}`, `${order.company_id.partner_id.bank_ids}`). This is a significant security risk. Use Odoo's built-in `mail.template` / `jinja` sandboxed rendering, or use `qweb` templates, or at minimum whitelist allowed attributes.

### 2. Duplicate/conflicting notification mechanisms
`sale_order_notification.py` defines `sale.order.notification.log` which is an entirely separate notification mechanism (using `ir.config_parameter` for bot token/chat ID) that conflicts with `telegram.notification` model. The module has two parallel implementations of the same feature. Remove one.

### 3. Missing access rules for `sale.order.notification.log`
The `ir.model.access.csv` only defines access for `telegram.notification`. The `sale.order.notification.log` model has no access rules, which will cause `AccessError` at runtime.

### 4. Missing `tests/__init__.py`
The test file exists but there is no `tests/__init__.py` to import it, so tests will never be discovered by Odoo's test runner.

### 5. Bot token stored in plain text
The `token` field is a plain `Char` field. While the view uses `password="True"` for display masking, the token is stored unencrypted in the database and will appear in logs, exports, and database dumps. Consider using `ir.config_parameter` with restricted access or at minimum ensure the field is not exported/logged.

## OCA Compliance Issues

### 6. Import ordering
OCA/Odoo coding standards require: stdlib first, then third-party, then odoo. In `sale_order.py`: `from odoo import models, fields, api, _` then `import re` then `import logging` -- stdlib imports must come first. Same issue in `telegram_notification.py` and `sale_order_notification.py`.

### 7. `_()` used at class level / module level
In `sale_order_notification.py`, `_()` is used in class-level field definitions (`string=_("Sale Order")`, `_description = _("Log for...")`) and in Selection tuples. The `_()` function should NOT be used in field `string` parameters (Odoo extracts these automatically for i18n) or in `_description`. For Selection values, use `_()` only when the string is used in runtime logic, not in field definitions.

### 8. `noqa` comments masking unused imports
In `__init__.py` and model files, `# noqa` is used. In `telegram_notification.py`, `api` is imported but never used -- remove the unused import rather than suppressing the linter.

### 9. Redundant `_logger` redefinition
In `sale_order.py`, `_logger = logging.getLogger(__name__)` is defined at module level AND again inside `_send_telegram_notification()`. The local redefinition shadows the module-level logger and is unnecessary.

### 10. Module naming
The module name `telegram_notification` is too generic for a sale-specific module in sale-workflow. If kept here (though it should be split per point above), it should be named `sale_telegram_notification` or similar. The generic `telegram_notification` name should be reserved for the base module in OCA/social.

### 11. Menu placement
The module creates a root-level menu item (`menu_telegram_notification_root`) which is not appropriate. It should be placed under the Sales menu or Settings, not as a top-level menu.

### 12. `widget="text_markdown"` does not exist in Odoo 16
The form view uses `widget="text_markdown"` for template fields. This widget does not exist in standard Odoo 16. Use `widget="text"` or no widget (plain textarea).

## Minor Issues

### 13. f-string in logger call
In `sale_order.py` line ~98: `_logger.error(f"Error processing placeholder ${{{expr}}}: {str(e)}")` -- use `%s` formatting for logger calls (lazy evaluation), not f-strings.

### 14. Bare `requests.post()` without `requests.Session`
Multiple calls to `requests.post()` without connection pooling. For sending to multiple chat IDs, a `requests.Session` would be more efficient.

### 15. `README.rst` has duplicate sections
The README contains duplicate `Usage` and `Contributors` sections (once from the template structure, once from the fragment files).

### 16. Missing newline at end of file
Several files are missing a trailing newline: `ir.model.access.csv`, `telegram_notification_security.xml`, `telegram_notification_views.xml`, `DESCRIPTION.rst`.

### 17. `action_quotation_send` override may break wizard flow
The `action_quotation_send()` method in Odoo 16 returns a wizard action. The notification is sent before the user actually sends the quotation (the wizard may be cancelled). The notification should be sent after the actual send, not when the wizard opens.

### 18. Tests are minimal
The `test_send_message_failure` test mocks a 400 status but the actual code retries with plain text on 400, so the mock needs two return values. The test for `send_message_failure` will actually pass (the second plain-text attempt also gets 400) but it calls `mock_post` twice, not once -- the `assert_called_once()` assertion will fail.

## Summary

The module needs significant rework:
1. **Split into base module (OCA/social) + sale glue module (OCA/sale-workflow)** per maintainer feedback
2. **Fix the security vulnerability** in template placeholder resolution
3. **Remove the duplicate notification mechanism** (`sale_order_notification.py`)
4. **Add missing access rules and test init**
5. **Fix OCA coding standard violations** (imports, i18n, naming)
6. **Fix the non-existent widget** and menu placement
7. **Fix test assertions** to match actual code behavior

AnmollGarg:
Hello @rousseldenis, @pedrobaeza, and @alexey-pelykh,

Thank you all for the detailed feedback and guidance on the architecture and security of this contribution.

Following your suggestions, I have addressed the issues by extracting the core logic into a standalone base module in OCA/social for version 18.0. Maintainers, please review the new base PR here: [OCA/social#1824](https://github.com/OCA/social/pull/1824).

I will follow up with the sales-specific glue module once this base is approved. Thank you

PR #3668, opened 2025-04-07
[17.0][IMP] sale_elaboration: Add multilines and new sales order report

AlvaroRM11:
@BinhexTeam

This implementation adds support for multi-line elaboration notes and introduces a new report for the sales order.

### DIFF
--- a/sale_elaboration/reports/report_saleorder.xml
+++ b/sale_elaboration/reports/report_saleorder.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<data>
+    <template id="report_saleorder" inherit_id="sale.report_saleorder_document">
+        <xpath expr="//t[@t-foreach='lines_to_report']/tr" position="after">
+        <tr>
+            <td colspan="99">
+                <t
+                        t-if="line.elaboration_note"
+                        t-call="sale_elaboration.elaboration_notes"
+                        groups="sale_elaboration.group_elaboration_note_on_order_report"
+                    >
+                    <t t-set="record" t-value="line" />
+                </t>
+                <hr class="mt-2" />
+            </td>
+        </tr>
+        </xpath>
+    </template>
+</data>
--- a/sale_elaboration/views/res_config_settings_views.xml
+++ b/sale_elaboration/views/res_config_settings_views.xml
@@ -10,15 +10,36 @@
                     id="group_elaboration_note_on_delivery_slip"
                 >
                         <field name="group_elaboration_note_on_delivery_slip" />
-                        <label for="group_elaboration_note_on_delivery_slip" />
                 </setting>
                 <setting
                     help="Elaboration notes will appear on the picking operations report"
                     id="group_elaboration_note_on_picking_operations"
                 >
                     <field name="group_elaboration_note_on_picking_operations" />
-                    <label for="group_elaboration_note_on_picking_operations" />
                 </setting>
+            </xpath>
+        </field>
+    </record>
+
+    <record id="res_config_settings_sale_view" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="sale.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//setting[@id='standardized_offers_setting']" position="after">
+                <div
+                    class="col-12 col-lg-6 o_setting_box"
+                    id="group_elaboration_note_on_order_report"
+                >
+                    <div class="o_setting_left_pane">
+                        <field name="group_elaboration_note_on_order_report" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label for="group_elaboration_note_on_order_report" />
+                        <div class="text-muted">
+                            Elaboration notes will appear on the order report
+                        </div>
+                    </div>
+                </div>
             </xpath>
         </field>
     </record>
--- a/sale_elaboration/__manifest__.py
+++ b/sale_elaboration/__manifest__.py
@@ -25,6 +25,7 @@
         "views/stock_picking_views.xml",
         "views/stock_move_views.xml",
         "reports/report_base.xml",
+        "reports/report_saleorder.xml",
         "reports/report_deliveryslip.xml",
         "reports/report_picking_operations.xml",
     ],
--- a/sale_elaboration/models/product_elaboration_mixin.py
+++ b/sale_elaboration/models/product_elaboration_mixin.py
@@ -12,7 +12,7 @@ class ProductElaborationMixin(models.AbstractModel):
         comodel_name="product.elaboration",
         string="Elaborations",
     )
-    elaboration_note = fields.Char(
+    elaboration_note = fields.Text(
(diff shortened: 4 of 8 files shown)


Yoana-RB:
Functional review LGTM! :)

Yoana-RB (approved):

lauradiaz22 (approved):
Functional review and this LGTM 👍

rrebollo (approved):
Code Review. LGTM!

pedrobaeza:
@sergio-teruel as this change impacts your module, can you please check?

ivs-cetmix (changes requested):
Hi @AlvaroRM11 , thank you for your contribution!

Please complete your commit message and add a description so it complies with the [OCA Coding Guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message).
It also makes it easier to follow the change flow later.
I think that the PR description can be used as well:
`This implementation adds support for multi-line elaboration notes and introduces a new report for the sales order.`

rousseldenis:
@sergio-teruel

edescalona:
Hi @sergio-teruel , if you could check the changes, thanks.

ivs-cetmix (approved):

ivs-cetmix:
Ok, let's wait for a couple of days for @sergio-teruel and if there is no response then let's have it merged.

sale_elaboration/reports/report_saleorder.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<data>
+    <template id="report_saleorder" inherit_id="sale.report_saleorder_document">
+        <xpath expr="//t[@t-foreach='lines_to_report']/tr" position="after">
+        <tr>
+            <td colspan="99">
+                <t
+                        t-if="line.elaboration_note"
+                        t-call="sale_elaboration.elaboration_notes"
+                        groups="sale_elaboration.group_elaboration_note_on_order_report"
```
alexey-pelykh, line 10: **Bug**: The `elaboration_notes` base template (in `report_base.xml`) already has `groups="sale_elaboration.group_elaboration_note_on_delivery_slip"` hardcoded on the inner `<div>`. By adding `groups="sale_elaboration.group_elaboration_note_on_order_report"` here, you create a double gate: the user must belong to **both** groups for the notes to appear on the sale order report.

This needs to be refactored so each caller controls its own visibility independently. One approach: remove the `groups` attribute from the base `elaboration_notes` template and let each caller (delivery slip, picking operations, sale order) wrap the `t-call` with their own `groups`.

sale_elaboration/reports/report_saleorder.xml
```diff
@@ -0,0 +1,19 @@
...
+                    >
+                    <t t-set="record" t-value="line" />
+                </t>
+                <hr class="mt-2" />
```
alexey-pelykh, line 14: **Bug**: This `<hr>` is outside the `t-if` condition. It will render after **every** sale order line, even lines without elaboration notes. It should either be moved inside the `t-if` block, or the entire `<tr>` should be conditionally rendered.

Suggestion:
```xml
<tr t-if="line.elaboration_note">
    <td colspan="99">
        <t t-call="sale_elaboration.elaboration_notes" groups="...">
            <t t-set="record" t-value="line" />
        </t>
        <hr class="mt-2" />
    </td>
</tr>
```

sale_elaboration/reports/report_saleorder.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
alexey-pelykh, line 1: Minor: Please use `<odoo>` as the root element instead of `<data>`, for consistency with other XML files in this module (e.g., `report_deliveryslip.xml` uses `<odoo>`).

sale_elaboration/views/res_config_settings_views.xml
```diff
@@ -10,16 +10,37 @@
...
-                        <label for="group_elaboration_note_on_delivery_slip" />
                 </setting>
                 <setting
                     help="Elaboration notes will appear on the picking operations report"
                     id="group_elaboration_note_on_picking_operations"
                 >
                     <field name="group_elaboration_note_on_picking_operations" />
-                    <label for="group_elaboration_note_on_picking_operations" />
                 </setting>
             </xpath>
```
alexey-pelykh, line 20: Why were the `<label>` elements removed from the existing delivery slip and picking operations settings? This changes the visual rendering of these settings. If not intentional, please revert these removals to avoid visual regressions.

alexey-pelykh (changes requested):
Thank you for this contribution! The idea of supporting multi-line elaboration notes and adding a sale order report is valuable. However, I found a few issues that should be addressed before merging.

## Summary

### 1. Double `groups` gating on the sale order report (bug)

The reusable `elaboration_notes` template in `report_base.xml` has a hardcoded `groups="sale_elaboration.group_elaboration_note_on_delivery_slip"`. In the new `report_saleorder.xml`, you call this template with an additional `groups="sale_elaboration.group_elaboration_note_on_order_report"`. This creates a double gate: the elaboration notes on the sale order will only render when **both** the delivery slip group **and** the order report group are enabled for the user. This is almost certainly unintended.

To fix this properly, the `elaboration_notes` base template should either:
- Have its `groups` attribute removed (letting each caller control visibility), or
- Be parameterized so each caller passes its own group.

This also means the existing delivery slip and picking operations callers would need to be updated accordingly.

### 2. Unconditional `<hr>` rendered for every line

The `<hr class="mt-2" />` in `report_saleorder.xml` is placed **outside** the `t-if` block. It renders on every sale order line regardless of whether elaboration notes exist, adding an unwanted horizontal rule after every line. It should be inside the `t-if` condition or the whole `<tr>` should be conditional.

### 3. Missing `<odoo>` root element

`report_saleorder.xml` uses `<data>` as root element. Other report files in this module (e.g., `report_deliveryslip.xml`, `report_picking_operations.xml`) use `<odoo>`. Please use `<odoo>` for consistency, which is also the standard for Odoo 17.0.

### 4. Removed `<label>` elements from existing settings

The labels for the delivery slip and picking operations settings were removed from `res_config_settings_views.xml`. This changes the visual appearance of those existing settings. Unless this is intentional and tested, these removals should be reverted to avoid regressions.

### 5. Inconsistent settings view structure

The new order report setting is placed in `sale.res_config_settings_view_form` using a manual `div`-based layout, while the existing elaboration settings live in `stock.res_config_settings_view_form` using the `<setting>` helper. This creates two different locations and two different visual patterns for related settings. Consider placing all elaboration settings together.

### 6. No tests

No tests were added for the new report or the new configuration setting. At minimum, a test verifying the sale order report renders with elaboration notes when the group is enabled would strengthen this contribution.

---

The `Char` to `Text` field type change is fine and appropriate for multi-line support. The unrelated comment punctuation change in `product_template.py` is trivial and acceptable.

rousseldenis:
Closing this as very old. Author seems no more active. Please feel free to reopen it if needed

PR #3671, opened 2025-04-10
[17.0][FIX] product_supplierinfo_for_customer_sale: ensure partner context

Qlasta:
partner context passed from view not always present.

**Description**
If sale order Pricelist type `partner`,  `price_unit` of sale order line is calculated using partner context passed in the view (https://github.com/odoo/odoo/blob/17.0/addons/sale/views/sale_order_views.xml#L520-L526).
However context is lost if price computation is triggered again from backend (e.g. when updating quantities) and it sets `list price`, but not `product.customerinfo` price.

**Current Behaviour**
When pricelist type is "Partner Prices on the product form", unit price in sale order line sometimes is set from list price not from customer info. Example by updating product quantity a few times in sale order line.

**Expected Behavior**
When pricelist type is "Partner Prices on the product form", unit price in sale order line is always set from customer info.

**Proposed implementation**
extend `_get_product_price_context` and update context by adding `partner_id` from sale order.

### DIFF
--- a/product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
+++ b/product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
@@ -14,6 +14,8 @@ class TestProductSupplierinfoForCustomerSale(TransactionCase):
         cls.customerinfo_model = cls.env["product.customerinfo"]
         cls.pricelist_item_model = cls.env["product.pricelist.item"]
         cls.pricelist_model = cls.env["product.pricelist"]
+        cls.sale_order_model = cls.env["sale.order"]
+        cls.sale_order_line_model = cls.env["sale.order.line"]
         cls.customer = cls._create_customer("customer1")
         cls.product = cls.env.ref("product.product_product_4")
         cls.product_variant_1 = cls.env.ref("product.product_product_4b")
@@ -41,6 +43,27 @@ class TestProductSupplierinfoForCustomerSale(TransactionCase):
         )
         cls.pricelist_template = cls._create_pricelist(
             "Test Pricelist Template", cls.product_template.product_variant_ids[:1]
+        )
+        cls.pricelist_customer = cls.pricelist_model.create(
+            {
+                "name": "Test Pricelist Customer",
+                "currency_id": cls.env.ref("base.USD").id,
+            }
+        )
+        cls.pricelist_item_customer = cls.pricelist_item_model.create(
+            {
+                "applied_on": "3_global",
+                "base": "partner",
+                "name": "Test Pricelist Item",
+                "pricelist_id": cls.pricelist_customer.id,
+                "compute_price": "formula",
+            }
+        )
+        cls.sale_order_customer = cls.sale_order_model.create(
+            {
+                "partner_id": cls.customer.id,
+                "pricelist_id": cls.pricelist_customer.id,
+            }
         )
         cls.env.user.groups_id |= cls.env.ref("product.group_product_pricelist")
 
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
LGTM

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
```diff
@@ -30,6 +32,7 @@ def setUpClass(cls):
         cls._create_partnerinfo(
             "customer", cls.customer, cls.product_variant_2, empty_variant=True
         )
+        cls._create_partnerinfo("customer", cls.customer, cls.product)
         cls.product_template = cls.env["product.template"].create(
```
alexey-pelykh, line 36: This line creates a second `product.customerinfo` for the same partner (`cls.customer`) and product (`cls.product`) that was already created on line 22 of the existing code:
```python
cls.customerinfo = cls._create_partnerinfo("customer", cls.customer, cls.product)
```
Since `_select_customerinfo` returns the first match and both records have the same price (100.0), this works but is redundant. Consider removing this duplicate to keep the test data clean.

alexey-pelykh (commented):
Thanks for this fix @Qlasta! The problem analysis is spot-on: when `_compute_price_unit` is triggered from the backend (e.g., quantity update), the view context carrying `partner_id` is lost, so `_price_compute` with `price_type='partner'` falls back to `list_price` instead of looking up `product.customerinfo`.

The approach of extending `_get_product_price_context` is clean and targets the right hook. A few observations:

**Overall**: The fix is correct and well-tested. The two new tests cover both the "context present" and "context absent" scenarios. I'm leaving as COMMENT rather than APPROVE since I have a minor suggestion and a question.

---

**Minor observations:**

1. **Duplicate customerinfo in test setUp**: The existing `setUpClass` already creates `cls.customerinfo = cls._create_partnerinfo("customer", cls.customer, cls.product)`. The PR adds another `cls._create_partnerinfo("customer", cls.customer, cls.product)` (without assigning the result). This creates a second `product.customerinfo` record for the same partner/product pair. While it works because `_select_customerinfo` returns the first match and both have `price=100.0`, it's unnecessary duplication. Consider removing this extra call since the existing `cls.customerinfo` already covers it.

2. **Test could also verify the recomputation scenario**: The PR description mentions the bug surfaces specifically when "updating quantities a few times" triggers recomputation. The current tests verify initial price at line creation with/without context, which is good. An additional test that creates a line, then updates `product_uom_qty` and verifies `price_unit` remains correct would more directly reproduce the described bug. This is optional but would strengthen confidence.

3. **Rebase needed**: The PR branch appears to be based on an older revision of 17.0 (manifest shows `17.0.1.0.0` while current 17.0 has `17.0.1.0.1`). A rebase onto current 17.0 would resolve this.

Good contribution! 👍

Qlasta:
@alexey-pelykh thank you for comments, updated my commit as suggested and rebased :)

rousseldenis:
@Qlasta Could you attract more reviewers as this is very old? Thanks

Qlasta:
> @Qlasta Could you attract more reviewers as this is very old? Thanks

Sure, will try to attract.
@AaronHForgeFlow , @LoisRForgeFlow , @GuillemCForgeFlow could you please review, when you have a chance? I saw you have reviewed the  similar [PR](https://github.com/OCA/sale-workflow/pull/3976) on v18.0

AaronHForgeFlow (commented):
Code LGTM. I think with only one new test for the context is passed would have been sufficient, and maybe the inline comments in tests are not really useful and you can just delete them.

product_supplierinfo_for_customer_sale/tests/test_product_supplierinfo_for_customer_sale.py
```diff
@@ -166,3 +189,57 @@ def test_product_supplierinfo_for_customer_sale_variant_wo_template(self):
             customerinfo.product_code,
             "Error: Customer product code was not passed to sale order line",
         )
+
+    def test_product_supplierinfo_sale_price_with_context_passed(self):
+        """Test normal behaviour when context is passed from view."""
+        # GIVEN / WHEN
```
rousseldenis, line 195: @Qlasta Could you remove those comments ?

PR #3676, opened 2025-04-14
[UPD] Copier + add rebel discount modules

rousseldenis:


legalsylvain (approved):
Hi @rousseldenis. thanks for taking care of this repo.

suggestion : I think you should mark the module in "excludes" manifest key, like in https://github.com/OCA/sale-workflow/pull/3598/files.

if you want a review, please ping me when it's ready.

thansk !

rousseldenis:
> Hi @rousseldenis. thanks for taking care of this repo.
> 
> suggestion : I think you should mark the module in "excludes" manifest key, like in https://github.com/OCA/sale-workflow/pull/3598/files.
> 
> if you want a review, please ping me when it's ready.
> 
> thansk !

It seems that the problem is the test: https://github.com/OCA/sale-workflow/actions/runs/14444314984/job/40501255461?pr=3676#step:8:45

setup/sale_triple_discount/setup.py
```diff
@@ -2,5 +2,9 @@
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
```
rousseldenis, line 5: @legalsylvain @grindtildeath Ok, I've found a solution to limit the version of account module here.

I'll merge to fix CI builds in v16

sbidoul (approved):

setup/sale_triple_discount/setup.py
legalsylvain: Nice !

PR #3683, opened 2025-04-17
[15.0][FIX] sales_team_invoiced_target_domain: Remove CRM dependency

juancarlosonate-tecnativa:
Fix view setting dependency

@tecnativa TT55088
@pedrobaeza @sergio-teruel

### DIFF
--- a/sales_team_invoiced_target_domain/model/res_company.py
+++ b/sales_team_invoiced_target_domain/model/res_company.py
@@ -4,7 +4,7 @@ from odoo import fields, models
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    crm_team_invoiced_domain = fields.Char(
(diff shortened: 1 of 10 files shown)


sales_team_invoiced_target_domain/model/res_company.py
```diff
@@ -4,7 +4,7 @@
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    crm_team_invoiced_domain = fields.Char(
+    sales_team_invoiced_domain = fields.Char(
```
pedrobaeza, line 7: This requires a migration script renaming the field.

juancarlosonate-tecnativa: done

pedrobaeza (approved):

sergio-teruel (approved):

PR #3684, opened 2025-04-17
mig: set start_date and end_date to datetime to better integrate with…

wlin-kencove:
Standard sale_start_end_dates is incompatible with enterprise due to conflicting  field names and incompatible field types:

In sale-workflow->sale_start_end_dates: start_date is a Date
In enterprise->sale_renting: start_date is a Datetime

This causes errors in computed fields due to the mismatching type. 
Testing setting date field to Datetime, but can see that changing the field name to avoid the name conflict with a separate module might be better to keep the calculations separate.

### DIFF
--- a/sale_start_end_dates/__manifest__.py
+++ b/sale_start_end_dates/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "Sale Start End Dates",
-    "version": "16.0.1.0.1",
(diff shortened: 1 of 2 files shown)


wlin-kencove:
Unneeded, main compatiblity issue was with sale_renting module which we could just uninstall.

PR #3686, opened 2025-04-22
[16.0][IMP] sale_order_line_input: add smartbutton to view sol

trisdoan:
### This change
- Add a button to view SO lines

- `sol_section_and_note` is to be able to view section and note in list view of SO lines (Before it's only available for embedded o2m field)

### DIFF
--- a/sale_order_line_input/models/res_config_settings.py
+++ b/sale_order_line_input/models/res_config_settings.py
@@ -1,0 +1,32 @@
+# Copyright 2025 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    show_view_sale_order_line = fields.Boolean(
+        string="Enable Smart Button to view SO line right on SO Form",
+        config_parameter="sale_order_line_input.show_view_sale_order_line ",
+    )
+
+    def set_values(self):
+        res = super().set_values()
+        group = self.env.ref("sale_order_line_input.sale_orderline_view_group")
+        if self.show_view_sale_order_line:
+            users = self.env["res.users"].search(
+                [
+                    (
+                        "groups_id",
+                        "in",
+                        self.env.ref("sales_team.group_sale_salesman").id,
+                    )
+                ]
+            )
+            users.write({"groups_id": [(4, group.id)]})
+        else:
+            users = self.env["res.users"].search([("groups_id", "in", [group.id])])
+            users.write({"groups_id": [(3, group.id)]})
+        return res
--- a/sale_order_line_input/models/sale_order.py
+++ b/sale_order_line_input/models/sale_order.py
@@ -1,57 +1,33 @@
-# Copyright 2018 Tecnativa - Carlos Dauden
-# Copyright 2023 Tecnativa - Carolina Fernandez
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2025 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, fields, models
+from odoo import _, fields, models
 
 
-class SaleOrderLine(models.Model):
-    _inherit = "sale.order.line"
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
 
-    pricelist_id = fields.Many2one(related="order_id.pricelist_id")
-    force_company_id = fields.Many2one(
-        comodel_name="res.company",
-        string="Forced company",
-        compute="_compute_force_company_id",
-        readonly=False,
-        help="Technical field to force company or get it "
-        "from env user if order don't exist.",
-    )
+    sol_count = fields.Integer(compute="_compute_sol_count")
 
-    @api.depends("order_id")
-    def _compute_force_company_id(self):
-        """Related company is not computed already when we click create new line"""
-        for line in self:
-            line.force_company_id = line.order_id.company_id or self.env.company
+    def _compute_sol_count(self):
+        for rec in self:
+            rec.sol_count = len(rec.order_line)
 
-    def _compute_name(self):
-        # A NewId is needed to set the product for the parent method to set
-        # the language correctly. Empty id leads to error in ‘_get_lang’.
-        for line in self:
-            if not line.order_id and line.product_id:
-                SaleOrder = self.env["sale.order"]
-                line.order_id = SaleOrder.new({})
-        return super()._compute_name()
-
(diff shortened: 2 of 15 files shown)


santostelmo (approved):
Tested ! LG

A couple of improvements that could be done:
- Possibility to add new sections and notes on the list view. It seems we are only able to read existing lines.
- Add the handle sequence in the view to order lines

sale_order_line_input/models/res_config_settings.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2025 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    show_view_sale_order_line = fields.Boolean(
```
rousseldenis, line 10: @trisdoan  I'm wondering why not using `group_...` field with `implied_group` instead of this (to avoid the `set_value()` stuff ?

rousseldenis, line 10: @trisdoan  I'm wondering why not using `group_...` field with `implied_group` instead of this (to avoid the `set_value()` stuff ?

rousseldenis: @trisdoan

cyrilmanuel:
hi @trisdoan could you answer / make the modification please ? thanks a lot for your help

cyrilmanuel (approved):

sale_order_line_input/models/res_config_settings.py
```diff
@@ -0,0 +1,32 @@
...
+        string="Enable Smart Button to view SO line right on SO Form",
+        config_parameter="sale_order_line_input.show_view_sale_order_line ",
```
alexey-pelykh, line 12: Bug: Trailing space in `config_parameter` value — `"sale_order_line_input.show_view_sale_order_line "` has a trailing space at the end. This will store the system parameter with a trailing space in its key, causing lookup issues.

Also, as @rousseldenis suggested, the idiomatic Odoo approach is to use `implied_group` on a `group_` prefixed Boolean field, which handles group toggling automatically without needing `set_values()`/`get_values()` overrides:

```python
group_sale_orderline_view = fields.Boolean(
    string="Enable Smart Button to view SO line right on SO Form",
    implied_group="sale_order_line_input.sale_orderline_view_group",
)
```

This would replace the entire `set_values()` method and the `show_view_sale_order_line` field.

sale_order_line_input/models/res_config_settings.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2025 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
```
alexey-pelykh, line 3: Copyright says `Camptocamp` but the contributor is from Trobz. Please use the correct copyright holder.

sale_order_line_input/security/sale_order_line_view_group.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
```
alexey-pelykh, line 2: `noupdate="1"` is generally not appropriate for security group definitions. If the group name or category needs updating on module upgrade, this flag will prevent it. Consider removing `noupdate="1"` (just use `<odoo>`).

sale_order_line_input/models/sale_order.py
```diff
@@ -1,57 +1,33 @@
-# Copyright 2018 Tecnativa - Carlos Dauden
-# Copyright 2023 Tecnativa - Carolina Fernandez
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2025 Camptocamp
```
alexey-pelykh, line 1: Copyright says `Camptocamp` but the contributor is from Trobz.

alexey-pelykh (changes requested):
Thank you for the contribution. I've done a thorough review and found several issues that should be addressed before merging.

## Summary

The smartbutton feature and the section/note list rendering are useful additions. However, there are a bug, architectural concerns, missing tests, and style issues that need attention.

## Key Findings

### Bug: Trailing space in `config_parameter`

`res_config_settings.py:12` — The `config_parameter` string has a trailing space:
```python
config_parameter="sale_order_line_input.show_view_sale_order_line ",
```
This will cause the system parameter to be stored with a trailing space in its key, which may lead to subtle lookup failures.

### Architecture: Use `implied_group` instead of manual `set_values()`

As @rousseldenis already noted, the standard Odoo pattern for toggling group membership from settings is to use a `group_` prefixed Boolean with the `implied_group` attribute. This eliminates the need for the custom `set_values()` override entirely:

```python
group_sale_orderline_view = fields.Boolean(
    string="Enable Smart Button to view SO line right on SO Form",
    implied_group="sale_order_line_input.sale_orderline_view_group",
)
```

The current manual approach has a secondary problem: there's no `get_values()` override, so reopening the Settings form won't reflect the current state correctly (the `config_parameter` stores a bool, but the actual group membership is managed separately and can drift).

### Missing version bump

The `__manifest__.py` version remains `16.0.1.0.1`. Since this adds a new feature, the minor version should be bumped (e.g., `16.0.1.1.0`).

### Missing tests

No tests are included for the new functionality. OCA modules are expected to have test coverage. At minimum:
- Test that the smartbutton action returns the correct domain/view
- Test the `sol_count` computed field
- Test the settings toggle (group assignment)

### `noupdate="1"` on security group

In `security/sale_order_line_view_group.xml`, the group record is wrapped in `<odoo noupdate="1">`. Security groups should generally **not** use `noupdate` — this prevents the record from being updated on module upgrade if the definition changes.

### Copyright header inconsistency

New files (`res_config_settings.py`, `sale_order.py`) have `Copyright 2025 Camptocamp` but the contributor is from Trobz. The copyright should match the contributing organization.

### XML formatting inconsistencies

Several XML files have mixed indentation styles (2-space in some places, 4-space or irregular in others). For example, `res_config_settings.xml` uses 2-space indentation for the outer structure but has irregularly indented inner content. `sale_order_line_view.xml` has inconsistent closing tag indentation in the new `view_sales_order_line_input_tree_inherit` record. Please align with the existing module's style.

### Minor: `sol_count` computation

`len(rec.order_line)` triggers a full recordset load. For display-only computed fields on potentially large orders, `self.env['sale.order.line'].search_count([('order_id', '=', rec.id)])` could be more efficient, though the practical impact is small for typical order sizes.

cyrilmanuel:
can be close in favor of https://github.com/OCA/sale-workflow/pull/4195 (superseed)

rousseldenis:
Closing in favor of #4195

Issue #3638, opened 2025-03-24
[16.0] build broken due to incompatible modules


simahawk:
https://github.com/OCA/sale-workflow/actions/runs/13944990367/job/39030049143

```
2025-03-19 11:06:18,492 549 INFO odoo odoo.addons.base.models.ir_module: ALLOW access to module.button_install on ['Sales', 'Invoicing', 'CRM', 'Manufacturing', 'Calendar', 'Account Global Discount', 'Account Fixed Discount', 'Account Invoice Start End Dates', 'Account Invoice Triple Discount', 'Account Payment Sale', 'Exception Rule', 'Base Revision (abstract)', 'Base Sub State', 'Base Tier Validation', 'Extended view inheritance', 'Delivery Costs', 'HR Holidays Public', 'Mail Message Destiny Link Template', 'Payment Engine', 'Portal Accounting Personal Data Only', 'Product Assortment', 'Product Packaging Container Deposit', 'Product Packaging level salable', 'Product set', 'Product Supplierinfo for Customers', 'Job Queue', 'Sales', 'Sale Loyalty', 'Sale Order Product Recommendation', 'Sales Teams', 'Sales and Warehouse Management', 'Sales Timesheet', 'Stock Picking Invoice Link', 'Stock picking lines with sequence number', 'Stock Procurement Customer', 'Stock Restrict Lot', 'Units of measure', 'web_action_conditionable', 'Web Widget Numeric Step'] to user __system__ #1 via n/a 
2025-03-19 11:06:18,632 549 WARNING odoo odoo.modules.loading: Transient module states were reset 
2025-03-19 11:06:18,633 549 ERROR odoo odoo.modules.registry: Failed to load registry 
Traceback (most recent call last):
  File "/opt/odoo/odoo/modules/registry.py", line 87, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/opt/odoo/odoo/modules/loading.py", line 457, in load_modules
    modules.button_install()
  File "<decorator-gen-76>", line 2, in button_install
  File "/opt/odoo/odoo/addons/base/models/ir_module.py", line 74, in check_and_log
    return method(self, *args, **kwargs)
  File "/opt/odoo/odoo/addons/base/models/ir_module.py", line 434, in button_install
    raise UserError(msg % (module.shortdesc, exclusion.exclusion_id.shortdesc))
[...]

PR #3687, opened 2025-04-22
[16.0][IMP] sale_order_product_recommendation_quick_add: add return keyword

trisdoan:
### Note
- CI is red due to some incompatible modules, which is currently discussed in https://github.com/OCA/sale-workflow/issues/3638

### This change
- Add 'return' keyword allows continuing function chain, e.g: viewing SO lines after clicking 'Quick Add' https://github.com/OCA/sale-workflow/pull/3686


santostelmo (approved):

rousseldenis (approved):
LGTM. Trivial.

Thanks!

PR #3689, opened 2025-04-24
[17.0][IMP] sale_automatic_workflow: add option for default payment journal

kevinkhao:


sale_automatic_workflow/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Sale Automatic Workflow",
-    "version": "17.0.1.1.0",
+    "version": "17.0.1.2.0",
```
rousseldenis: If you don't have migration scripts, don't change this. Will be done at merge

rousseldenis (approved):
Minor comment

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -202,7 +208,9 @@ def run_with_workflow(self, sale_workflow):
             )
 
         if sale_workflow.register_payment:
-            self._register_payments(
+            self.with_context(
+                property_payment_journal_id=sale_workflow.property_payment_journal_id
```
StefanRijnhart: Is it perhaps the case that at this point, the process is still company agnostic, and you should fetch the value of this company dependent field at a later stage (with the order's company if not already set)?

rousseldenis: @kevinkhao

kevinkhao:
@StefanRijnhart right! Fixed

StefanRijnhart (approved):

PR #3701, opened 2025-05-06
[17.0][MIG] sales_team_invoiced_target_domain: Migration to 17.0

juancarlosonate-tecnativa:
@tecnativa TT55849
@sergio-teruel @carlos-lopez-tecnativa


sales_team_invoiced_target_domain/model/crm_team.py
```diff
@@ -0,0 +1,30 @@
...
+class CrmTeam(models.Model):
+    _inherit = "crm.team"
+
+    def _compute_invoiced(self):
+        if not self:
+            return
+        sales_team_invoiced_domain = self.env.company.sales_team_invoiced_domain
+        if not sales_team_invoiced_domain:
+            return super()._compute_invoiced()
+        today = fields.Date.today()
```
carlos-lopez-tecnativa: Please use context_today to prevent errors in countries where the timezone is negative.

```suggestion
        today = fields.Date.context_today(self)
```

sales_team_invoiced_target_domain/model/crm_team.py
```diff
@@ -0,0 +1,30 @@
...
+        invoiced_domain = [
+            ("move_type", "in", ["out_invoice", "out_refund", "out_receipt"]),
+            ("team_id", "in", self.ids),
+            ("date", ">=", fields.Date.to_string(today.replace(day=1))),
+            ("date", "<=", fields.Date.to_string(today)),
```
carlos-lopez-tecnativa: Converting to string is unnecessary; it's better to use native date types.

```suggestion
            ("date", ">=", today.replace(day=1)),
            ("date", "<=", today),
```

sales_team_invoiced_target_domain/model/crm_team.py
```diff
@@ -0,0 +1,30 @@
...
+        ]
+        domain_list = ast.literal_eval(sales_team_invoiced_domain)
+        invoiced_domain = expression.AND([invoiced_domain, domain_list])
+        for team in self:
+            team.invoiced = 0.0
+            moves = self.env["account.move"].search(invoiced_domain)
+            for move in moves:
+                if move.team_id == team:
+                    team.invoiced += move.amount_untaxed_signed
```
carlos-lopez-tecnativa: I think this can be improved using `_read_group`; it's more efficient and can retrieve all data grouped by `team_id` in a single query. Could you update the code, please? Or is there any limitation preventing the use of `_read_group`?

sales_team_invoiced_target_domain/model/res_company.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    sales_team_invoiced_domain = fields.Char(
+        default="['|', '|', "
+        "['payment_state', '=', 'in_payment'], "
+        "['payment_state', '=', 'paid'], "
+        "['payment_state', '=', 'reversed']]"
```
carlos-lopez-tecnativa: I think this can be improved using the `in` operator.

```suggestion
        default="[('payment_state', 'in', ['in_payment', 'paid', 'partial', 'reversed'])]"
```

image

Why didn't you include the `partial` option?
image

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,84 @@
+from odoo import fields
+from odoo.tests.common import TransactionCase
+
+
+class TestCrmTeamInvoiced(TransactionCase):
```
carlos-lopez-tecnativa: Could you inherit from [BaseCommon](https://github.com/odoo/odoo/blob/28cd13dba2a79db9d71f2250d840e93231e6b53d/odoo/addons/base/tests/common.py#L21) to reduce the overhead from tracking and improve the performance of the test suite?

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,84 @@
...
+    def setUp(self):
+        super().setUp()
```
carlos-lopez-tecnativa: Why is this not done in `setUpClass` instead of `setUp`?

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,84 @@
...
+        self.team = self.env["crm.team"].create({"name": "Test Team"})
+        self.team2 = self.env["crm.team"].create({"name": "Test Team 2"})
+        self.partner_id = self.env["res.partner"].create({"name": "Test Partner"})
+
+    def create_account_move(self, team_id, payment_state_code, amount):
+        move = self.env["account.move"].create(
+            {
+                "move_type": "out_invoice",
+                "team_id": team_id.id,
+                "date": fields.Date.today(),
```
carlos-lopez-tecnativa:
```suggestion
                "date": fields.Date.context_today(self.env.user),
```

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,84 @@
...
+                    ("team_id", "=", self.team2.id),
+                    (
+                        "date",
+                        ">=",
+                        fields.Date.to_string(fields.Date.today().replace(day=1)),
+                    ),
+                    ("date", "<=", fields.Date.to_string(fields.Date.today())),
+                ]
+            )
+            .mapped("amount_untaxed_signed")
```
carlos-lopez-tecnativa: I see duplicated code that can be improved.
Create a new function in `crm.team`, such as 
`def _prepare_invoice_domain()`
and reuse this function in both `_compute_invoiced` and here in the test.

carlos-lopez-tecnativa (changes requested):
Code review

sales_team_invoiced_target_domain/model/crm_team.py
juancarlosonate-tecnativa: But then, if a team isn’t in the group because of the domain, we’ll still need to compute its invoiced value even if it’s 0 so the performance gain from using read_group is lost when we have to look up the team and assign its invoiced, right?

sales_team_invoiced_target_domain/model/res_company.py
juancarlosonate-tecnativa: I’m not adding it because that’s how it’s done in the core. https://github.com/odoo/odoo/blob/18.0/addons/sale/models/crm_team.py#L70

sales_team_invoiced_target_domain/model/crm_team.py
juancarlosonate-tecnativa: Please check the other changes, this is the only thing left to resolve. Thx!

carlos-lopez-tecnativa: Please try something like this:

```suggestion
        domain_list = ast.literal_eval(self.env.company.sales_team_invoiced_domain)
        if not domain_list:
            return super()._compute_invoiced()
        invoiced_domain = expression.AND([self._prepare_invoice_domain(), domain_list])
        team_data = self.env["account.move"]._read_group(
            invoiced_domain,
            groupby=["team_id"],
            aggregates=["amount_untaxed_signed:sum"],
            
        )
        team_dict = dict(team_data)
        for team in self:
            team.invoiced = team_dict.get(team) or 0.0
```

As you're iterating over `self`, if a team has no data in `team_dict`, the default value is 0.0. This covers your case.
Please let me know if this works for you, and if so, kindly update the test as well.

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,64 @@
...
+                "date": fields.Date.context_today(self.env.user),
+                "amount_untaxed_signed": amount,
+                "partner_id": self.partner_id.id,
+                "invoice_line_ids": [
+                    (0, 0, {"name": "Test Line", "quantity": 1, "price_unit": amount})
+                ],
+            }
+        )
+        move.action_post()
+        move.payment_state = payment_state_code
```
carlos-lopez-tecnativa, line 30: I'm not sure if this field should be set manually. If it's a computed field, why not add a `account.payment` instead?

carlos-lopez-tecnativa (changes requested):

sales_team_invoiced_target_domain/model/crm_team.py
juancarlosonate-tecnativa: Yep, works fine, thx!

sales_team_invoiced_target_domain/tests/test_crm_team.py
juancarlosonate-tecnativa: You're right, but that would overcomplicate these tests unnecessarily, since the goal here is only to verify that the domain filtering on payment_state works correctly.

sales_team_invoiced_target_domain/tests/test_crm_team.py
```diff
@@ -0,0 +1,66 @@
...
+            .search(self.team._prepare_invoice_domain())
+            .mapped("amount_untaxed_signed")
+        )
+        expected_team2_invoiced = sum(
+            self.env["account.move"]
+            .search(self.team2._prepare_invoice_domain())
+            .mapped("amount_untaxed_signed")
+        )
+        self.assertEqual(self.team.invoiced, expected_team_invoiced)
+        self.assertEqual(self.team2.invoiced, expected_team2_invoiced)
```
carlos-lopez-tecnativa: This test is not consistent. You’re not really testing anything, because you're just comparing the amount with the same data already present in the system.
It would be better to compare the team’s invoiced value with a specific expected value (not with the `sum` from `account.move`)

```suggestion
        self.assertEqual(self.team.invoiced, 100) # the specific expected value  
        self.assertEqual(self.team2.invoiced, 200) # the specific expected value  
```

carlos-lopez-tecnativa (approved):
Tested and working. 
Just a minor comment regarding the test (non-blocking from my side) but I think the test is not realistic. What do you think?
cc @pedrobaeza

sales_team_invoiced_target_domain/tests/test_crm_team.py
pedrobaeza: Yes, tests should be deterministic with fixed data, not to replicate the business code. For that, you have to create in setup a deterministic dataset.

juancarlosonate-tecnativa: you're absolutely right, also, the move with not_paid status should result in 0 because it doesn't follow the core filter

pedrobaeza (approved):
/ocabot merge nobump

PR #3702, opened 2025-05-07
[15.0][FIX] sale_order_product_assortment: Add compute_sudo in has_allowed_products

mariancuadranetkia:
We detect problem creating sales with non manager users. In this moment, allowed assortments are not being calculated correctly.

So, we've added compute_sudo in has_allowed_products too like in allowed_product_ids and tested again.

@sergio-teruel how do you see this fix? I'm not sure if this field has not compute_sudo for any reason or not.

Thanks!

### DIFF
--- a/sale_order_product_assortment/models/sale_order.py
+++ b/sale_order_product_assortment/models/sale_order.py
@@ -16,7 +16,10 @@ class SaleOrder(models.Model):
         compute="_compute_product_assortment_ids",
         compute_sudo=True,
     )
-    has_allowed_products = fields.Boolean(compute="_compute_product_assortment_ids")
+    has_allowed_products = fields.Boolean(
+        compute="_compute_product_assortment_ids",
+        compute_sudo=True,
+    )
 
     @api.depends("partner_id", "partner_shipping_id", "partner_invoice_id")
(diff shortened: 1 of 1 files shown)


javierobesonetkia (approved):
Tested that now it works as expected

sergio-teruel:
Hi @mariancuadranetkia what field do you need access in this method?
All employee can access to partners, products and filters (Own filters and with a False user)

mariancuadranetkia:
Hi @sergio-teruel 
I was testing, and I saw that a non manager user can't see: partner.applied_assortment_id. 
If this kind of user, creates a sale order, and without save chooses a partner and adds different lines, allowed_product_ids is not correctly calculated because this field is empty. 

I doubted about why first field works with compute_sudo and second one not using the same compute method... For this reason I ask and mention you too, I hope you don't mind.

Thank you very much for your help!

sergio-teruel (approved):

sergio-teruel:
Oks!! 
Anyway, it doesn't hurt to make this change, so perfect.

pedrobaeza (approved):
It's OK also for consistency, as `allowed_product_ids` has already that option for the same compute method,.

/ocabot merge patch

PR #3703, opened 2025-05-07
[16.0][FIX] sale_blanket_order: SO blanket_order_id computation

LauraCForgeFlow:
Before this commit, the `blanket_order_id` was a related field, meaning that if a SO had multiple lines and the first line was not linked to a BO line, this field remained unset, even if other lines were linked to a BO.

With these changes, we assign the first BO that we can find, regardless of its position in the order lines, and avoid this problem. Also added some test cases.

### DIFF
--- a/sale_blanket_order/models/sale_orders.py
+++ b/sale_blanket_order/models/sale_orders.py
@@ -13,7 +13,7 @@ class SaleOrder(models.Model):
     blanket_order_id = fields.Many2one(
         "sale.blanket.order",
         string="Origin blanket order",
-        related="order_line.blanket_order_line.order_id",
+        compute="_compute_blanket_order_id",
     )
     disable_adding_lines = fields.Boolean(
         compute="_compute_disable_adding_lines",
@@ -49,6 +49,12 @@ class SaleOrder(models.Model):
                             "blanket order lines customer"
                         )
                     )
+
+    @api.depends("order_line.blanket_order_line.order_id")
+    def _compute_blanket_order_id(self):
+        for order in self:
+            blanket_order = order.order_line.mapped("blanket_order_line.order_id")
+            order.blanket_order_id = blanket_order[:1]
 
     @api.depends("blanket_order_id")
     @api.depends_context("uid")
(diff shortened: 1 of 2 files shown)


sale_blanket_order/models/sale_orders.py
```diff
@@ -47,6 +47,16 @@ def check_partner_id(self):
...
 
+    @api.depends("order_line.blanket_order_line.order_id")
+    def _compute_blanket_order_id(self):
+        for order in self:
+            blanket_order = False
+            for line in order.order_line:
+                if line.blanket_order_line:
+                    blanket_order = line.blanket_order_line.order_id
+                    break
+            order.blanket_order_id = blanket_order
```
JordiMForgeFlow: blanket_order = order.order_line.mapped("blanket_order_line.order_id")
order.blanket_order_id = blanket_order[0] if blanket_order else False

LoisRForgeFlow (approved):
LGTM :+1:

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
Code review LGTM!

sale_blanket_order/models/sale_orders.py
```diff
@@ -50,6 +50,12 @@ def check_partner_id(self):
                         )
                     )
 
+    @api.depends("order_line.blanket_order_line.order_id")
+    def _compute_blanket_order_id(self):
+        for order in self:
+            blanket_order = order.order_line.mapped("blanket_order_line.order_id")
+            order.blanket_order_id = blanket_order[0] if blanket_order else False
```
rousseldenis: @LauraCForgeFlow You can use `first()`

MiquelRForgeFlow: or `blanket_order[:1]`

LauraCForgeFlow: @rousseldenis @MiquelRForgeFlow done, thanks! :)

rousseldenis (approved):

PR #3705, opened 2025-05-07
[16.0][ADD] sale_order_line_price_lock_by_pricelist: New module

chienandalu:
cc @moduon @rafaelbn @Shide @fcvalgar MT-9860

### DIFF
--- a/sale_order_line_price_lock_by_pricelist/models/product_pricelist.py
+++ b/sale_order_line_price_lock_by_pricelist/models/product_pricelist.py
@@ -1,0 +1,33 @@
+# Copyright 2025 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import fields, models
+
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    lock_product_prices_applied_on = fields.Selection(
+        string="Lock price when rule is applied on",
+        selection=[
+            ("3_global", "All products"),
+            ("2_product_category", "Product Category"),
+            ("1_product", "Product"),
+            ("0_product_variant", "Product Variant"),
+        ],
+        help="Avoid that salesmen change prices based on this scope. Rules applied on "
+        "base pricelist take into account the items of those pricelists with the rule "
+        "of the pricelist applied on the sale order",
+    )
+
+    def _get_base_product_rule(self, product, quantity, uom=None, date=False, **kwargs):
+        """Recurrent method to pull the pricelist item from the pricelist chain of
+        inheritance"""
+        self.ensure_one()
+        item = self.env["product.pricelist.item"].browse(
+            self._get_product_rule(product, quantity, uom, date, **kwargs)
+        )
+        if item.base and item.base == "pricelist" and item.base_pricelist_id:
+            return item.base_pricelist_id._get_base_product_rule(
+                product, quantity, uom, date, **kwargs
+            )
+        return item.id
--- a/sale_order_line_price_lock_by_pricelist/models/sale_order.py
+++ b/sale_order_line_price_lock_by_pricelist/models/sale_order.py
@@ -1,0 +1,61 @@
+# Copyright 2025 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order.line"
+
+    origin_pricelist_item_id = fields.Many2one(
+        comodel_name="product.pricelist.item",
+        compute="_compute_origin_pricelist_item_id",
+        help="Technical field to get the base pricelist item that has the origin "
+        "price calculation",
+    )
+    price_locked = fields.Boolean(
+        compute="_compute_price_locked",
+        help="This flag will make the price_unit and discount readonly based on the "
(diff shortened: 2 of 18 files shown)


sale_order_line_price_lock_by_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+    @api.depends("pricelist_item_id")
+    def _compute_origin_pricelist_item_id(self):
+        for line in self:
+            # Just compute those items which price is pulled from another pricelist
+            if (
+                line.pricelist_item_id.base
+                and line.pricelist_item_id.base == "pricelist"
+                and line.pricelist_item_id.base_pricelist_id
```
Shide, line 28: Why not check directly base_pricelist_id?

Shide (commented):
I'm thinking of why not have a default config setting that applies for all pricelists (maybe adding an extra option to selector -> default).
If you don't want to apply or change the default setting, you can set it to None, or change the selection.

And why you return the base pricelist? What happen if a pricelist is configured like this?

```
Base pricelist
    Pricelist 1 (with lock)
        Pricelist 2
```

The base pricelist will be the "Base pricelist" and the Pricelist 1 has a lock.
Will be locked in this case?

Maybe with a video or something like this will help us to understand how it works

chienandalu:
> I'm thinking of why not have a default config setting that applies for all pricelists (maybe adding an extra option to selector -> default). If you don't want to apply or change the default setting, you can set it to None, or change the selection.

The default value is to be unset. The sales manager should configure those values in the pricelists they want to lock.

> 
> And why you return the base pricelist? What happen if a pricelist is configured like this?
> 
> ```
> Base pricelist
>     Pricelist 1 (with lock)
>         Pricelist 2
> ```
> 
> The base pricelist will be the "Base pricelist" and the Pricelist 1 has a lock. Will be locked in this case?

It won't be locked. The idea:

- When the sales order has a pricelist with a locking scope setting, that's the config ruling all the possible pulled items from other pricelists.
- Otherwise, the prices won't be locked even if a pricelist in the middle is configured like that.

This gives IMO much more simplicity to the scheme, as you don't have to figure out if some pricelist in the chain is going to lock your price editing, you just have to check the top one.

sale_order_line_price_lock_by_pricelist/models/sale_order.py
chienandalu: It could happen that the base is different (i.e.: malformed data). You can see that Odoo ensure that check as well:

https://github.com/odoo/odoo/blob/9e22dbb7b6fb581d2f11bf0ec48b230047686050/addons/product/models/product_pricelist_item.py#L424-L425

rafaelbn (approved):
Functional test in Staging 👍🏼

Shide (approved):

Gelojr (approved):
Nice job @chienandalu

fcvalgar (approved):
Functional review, 

LGTM, thank you @chienandalu

sale_order_line_price_lock_by_pricelist/models/product_pricelist.py
```diff
@@ -0,0 +1,33 @@
...
+        help="Avoid that salesmen change prices based on this scope. Rules applied on "
+        "base pricelist take into account the items of those pricelists with the rule "
+        "of the pricelist applied on the sale order",
+    )
+
+    def _get_base_product_rule(self, product, quantity, uom=None, date=False, **kwargs):
+        """Recurrent method to pull the pricelist item from the pricelist chain of
+        inheritance"""
+        self.ensure_one()
+        item = self.env["product.pricelist.item"].browse(
```
rousseldenis, line 26: @chienandalu Could you maybe add a test for this code part ?

Gelojr (approved):
Congrats on the contribution @chienandalu  great work!
The following tests were executed and passed:

- Test 1: With pricelist scope Product and a 15% discount rule for [FURN_6666] Acoustic Bloc Screens, a salesperson saw price_unit read-only while discount remained editable (changed to 10%) as expected.
- Test 2: For a product without any rule in the pricelist, both price_unit and discount were editable by the salesperson.
- Test 3: After switching to a pricelist without a rule for [FURN_6666], price_unit became editable again; discount stayed editable.
- Test 4: A Sales Manager user could edit both price_unit and discount even when the salesperson view was locked.
- Test 5: With a rule coming from a base pricelist (scope Product), the line’s price_unit was locked based on the origin rule; discount remained editable.
- Test 6: With pricelist scope Product category and a category rule (formula ≈25% discount), products in that category had price_unit locked and discount editable; products outside the category stayed fully editable.
- Test 7: With scope All products and a global rule (formula ≈30% discount), price_unit was locked on all lines while discount remained editable.

Gelojr:
Here’s a short explainer video covering the module’s basics: configuring “Lock price when rule is applied on” with the condition set to Product, and how that behavior works in a sales order. You’ll see the unit price locked for salespeople while the discount remains editable.
https://www.loom.com/share/297a4fdde26c408fa6f903d8ed633067?sid=e49d1798-1228-431e-8136-bd72611ed817

rafaelbn (approved):
/ocabot merge patch

rafaelbn:
Thank you for the work and review @chienandalu @Shide @fcvalgar @rousseldenis ! ❤️ ! Very usefull the short video @Gelojr 😄 .

PR #3710, opened 2025-05-08
[17.0][FIX] sale_order_carrier_auto_assign: preserve carrier on confirm

jbaudoux:
Fix Add delivery_auto_refresh changes from https://github.com/OCA/sale-workflow/pull/3635 :

* The carrier should be preserved on action confirm
* The carrier should not be set if all service

Also recover missing copyrights


sale_order_carrier_auto_assign/tests/test_sale_order_carrier_auto_assign.py
```diff
@@ -1,67 +1,62 @@
...
-class TestSaleOrderCarrierAutoAssign(BaseCommon):
+# Common class for OnCreate and OnConfirm
+class TestSaleOrderCarrierAutoAssignCommon(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.settings = cls.env["res.config.settings"].create({})
-        cls.partner = cls.env["res.partner"].create({"name": "Test partner"})
-        cls.product = cls.env.ref("product.product_product_9")
+        cls.partner = cls.env.ref("base.res_partner_2")
```
pedrobaeza: Don't use data outside the test. Create the needed data in the tests.

sale_order_carrier_auto_assign/__manifest__.py
```diff
@@ -7,7 +7,8 @@
     "development_status": "Beta",
     "category": "Operations/Inventory/Delivery",
     "website": "https://github.com/OCA/sale-workflow",
-    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
```
pedrobaeza, line 10: I'm curious. Where did you enter that copyrights? I doubt my colleague removes them intentionally.

jbaudoux: I designed this module with @grindtildeath in July 2020 on v13.0 and I'm following its evolution
Nobody removed it, it was lacking from the beginning.

Then there was this PR that was closed https://github.com/OCA/sale-workflow/pull/3081 and later partially integrated

I'm recovering the missing bits

sale_order_carrier_auto_assign/tests/test_sale_order_carrier_auto_assign.py
jbaudoux: sure :) you were too quick in the review. Thanks

jbaudoux:
@victoralmau Why you set a carrier when the SO is service only ? It doesn't make sense in my opinion

* This in v17.0 : https://github.com/OCA/sale-workflow/commit/85583359060d35fcbbf19b430d612577bc5d864a#diff-c0db59164377871a4b3083527e320de3d150ad86441abdf24da743f7edc55f27R71
* Compared to v16.0 : https://github.com/OCA/sale-workflow/blob/16.0/sale_order_carrier_auto_assign/models/sale_order.py#L21

I'm recovering the missing test https://github.com/OCA/sale-workflow/blob/16.0/sale_order_carrier_auto_assign/tests/test_sale_order_carrier_auto_assign.py#L75
I need to change the code or the test

cc @pedrobaeza 

**Edit:**
I think it's a mistake in regards of this https://github.com/OCA/delivery-carrier/blob/16.0/delivery_auto_refresh/models/sale_order.py#L145
The carrier should not be set

victoralmau (approved):
Code review OK.

I have checked and there seems to be no side effects from these changes in `delivery_auto_refresh`.

sale_order_carrier_auto_assign/__manifest__.py
rousseldenis: The man from the shadows :sweat_smile:

rousseldenis (approved):

rousseldenis:
@jbaudoux Do you consider this as a patch or minor change ?

jbaudoux:
> @jbaudoux Do you consider this as a patch or minor change ?

@rousseldenis Patch

PR #3448, opened 2024-12-03
[18.0][MIG] sale_stock_delivery_state: Migration to 18.0

thienvh332:
Context:
- This is a glue module for **`sale_delivery_state`** and **`sale_stock`**

Changes:
- Added **`checklog-odoo.cfg`** because **`sale_delivery_state`** can operate independently from **`sale_stock`** so it is not possible to determine dependencies or installation order. Leading to having to ignore the **`"overrides existing selection" warning`**

Depends on:
- #3395

PR #3285, opened 2024-08-28
[14.0][IMP] sale_delivery_state: migration script

manuelregidor:
Create column skip_sale_delivery_state in sale_order_line table when upgrading the module to avoid memory error.

T-6522

PR #3713, opened 2025-05-08
[18.0][FWD][IMP] sale_delivery_state: Skip sale order lines for delivery state + misc fixes

simahawk:
FWD port #2864 + #3285.

NOTE: the migrate script has been used also for pre init hook to avoid impacting existing dbs where this module is not installed yet (this was missing in the work done on v14).
I also moved the post init script from https://github.com/OCA/sale-workflow/pull/3448 to here.

Additional fixes:
* compat of the status field w/ sale_stock
* for the sol view, as the 'started' state was not considered

### DIFF
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -14,6 +14,7 @@
     "depends": ["sale"],
     "data": [
         "views/sale_order_views.xml",
+        "views/res_config_settings_views.xml",
     ],
     "demo": [
         "demo/sale_demo.xml",
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Sale delivery State",
     "summary": "Show the delivery state on the sale order",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
     "category": "Product",
     "website": "https://github.com/OCA/sale-workflow",
     "author": "Akretion, Odoo Community Association (OCA)",
@@ -19,4 +19,5 @@
     "demo": [
         "demo/sale_demo.xml",
     ],
+    "pre_init_hook": "pre_init_hook",
 }
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -20,4 +20,5 @@
         "demo/sale_demo.xml",
     ],
     "pre_init_hook": "pre_init_hook",
+    "post_init_hook": "post_init_hook",
 }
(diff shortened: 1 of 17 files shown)


sale_delivery_state/models/sale_order.py
```diff
@@ -63,16 +79,23 @@ def _partially_delivered(self):
...
-    @api.depends("order_line.qty_delivered", "state", "force_delivery_state")
+    @api.depends(
+        "order_line.qty_delivered",
+        "order_line.skip_sale_delivery_state",
+        "state",
+        "force_delivery_state",
+    )
     def _compute_oca_delivery_status(self):
+        # Loose dep on sale_stock
+        has_pickings = "picking_ids" in self._fields
```
simahawk: Well, I should find a better solution for this since the field is stored. Probably we need a sale_stock glue module if we want to support the "started" state. I'll add a TODO.

sale_delivery_state/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,11 @@
...
+# @author: Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import SUPERUSER_ID, api
+
+from odoo.addons.sale_delivery_state.hooks import _setup_new_sol_column
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _setup_new_sol_column(env)
```
RobbeDynapps: The function `_setup_new_sol_column` is expecting a Cursor, not an environment

sale_delivery_state/models/sale_order.py
```diff
@@ -15,6 +17,16 @@ class SaleOrder(models.Model):
...
+        # Respect the same order as in sale_stock
+        # Including the 'started' state
+        # that is not used here but we compute it
+        # if pickings are available, to be compatible.
+        selection=[
+            ("pending", "Not Delivered"),
+            ("started", "Started"),
+            ("partial", "Partially Delivered"),
+            ("full", "Fully Delivered"),
+        ],
```
sbiosca-s73, line 29:
```suggestion
```

This is not necessary, the problem is in the dependence with sale, must be dependence with sale_stock.

sale_delivery_state/__manifest__.py
```diff
@@ -14,8 +14,10 @@
     "depends": ["sale"],
```
sbiosca-s73, line 14:
```suggestion
    "depends": ["sale_stock"],
```

sale_delivery_state/models/sale_order.py
simahawk: That's what I don't want to do :wink:

RabbitJon-S73: What's is the point of doing like this?

simahawk: We already have a glue module for that https://github.com/OCA/sale-workflow/pull/3448 and this was done back in v16 https://github.com/OCA/sale-workflow/pull/2770.

rousseldenis (approved):

sebalix (approved):

PR #3722, opened 2025-05-14
[16.0][ADD] sale_order_safe_commitment_date: New module

chienandalu:
When salesmen set a commitment date there's nothing disallowing them to set it before the expected date computed according to the order lead times. That can cause great issues with deliveries schedules, as those lead times are set to ensure that the orders are delivered on time.

image

cc @moduon @rafaelbn @Shide @Gelojr MT-9790


sale_order_safe_commitment_date/views/sale_order_views.xml
```diff
@@ -0,0 +1,56 @@
...
+            <field name="show_update_pricelist" position="before">
+                <label for="commitment_date" string="Delivery Date" />
+                <div name="commitment_date_div" class="o_row">
+                    <field
+                        name="commitment_date"
+                        attrs="{'invisible': [('commitment_date', '=', False), ('state', 'not in', ('draft', 'sent'))]}"
+                    />
+                    <span name="expected_date_span" class="text-muted"><i
+                            class="fa fa-warning text-warning"
+                            attrs="{'invisible': [('is_commitment_date_unsafe', '=', False)]}"
```
Shide, line 48: **suggestion:** Hide if expected date has not been calculated yet

```suggestion
                            attrs="{'invisible': ['|', ('is_commitment_date_unsafe', '=', False), ('expected_date', '=', False)]}"
```

sale_order_safe_commitment_date/views/sale_order_views.xml
```diff
@@ -0,0 +1,56 @@
...
+                </div>
+            </field>
+        </field>
+    </record>
+    <data noupdate="1">
+        <!-- TODO: Enable via res.config.settings -->
+        <record id="view_order_form_commitment_date_ux" model="ir.ui.view">
+            <field name="active" eval="False" />
+        </record>
+    </data>
```
Shide, line 64: Why is this necessary?

Shide (changes requested):

sale_order_safe_commitment_date/views/sale_order_views.xml
chienandalu: Simply making it optionable

sale_order_safe_commitment_date/views/sale_order_views.xml
chienandalu: In that case the `is_commitment_date_unsafe` is always `False`

fcvalgar:
Can you launch the runboat again @chienandalu? Thank you

image

chienandalu:
> Can you launch the runboat again @chienandalu? Thank you

looks like a runboat quirk... you can launch in debug with assets mode (🦍  gorilla mode ;)) and it works

fcvalgar:
> > Can you launch the runboat again @chienandalu? Thank you
> 
> looks like a runboat quirk... you can launch in debug with assets mode (🦍 gorilla mode ;)) and it works

I have tried both modes and clearing my browser cache but the result has been the same. 

Thank you.

image

Gelojr (approved):

fcvalgar (approved):
Functional review,

LGTM, thank you @chienandalu.

image

image

sale_order_safe_commitment_date/views/sale_order_views.xml
Shide: I just add `expected_date` to the invisible domain. 
When there is no lines, hide the expected date label too.

sale_order_safe_commitment_date/views/sale_order_views.xml
Shide: Just go to the views and unarchive it?
How can be enabled via res.config.settings?
Can we add CONFIGURE.md to handle this situation?

Shide (approved):

sale_order_safe_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+    def _compute_is_commitment_date_unsafe(self):
+        """A commitment date is considered unsafe if it is before the expected date as
+        the products won't be delivered on time."""
+        self.is_commitment_date_unsafe = False
+        for order in self.filtered(
+            lambda x: x.commitment_date
+            and x.expected_date
+            and x.state in {"draft", "sent"}
+        ):
+            order.is_commitment_date_unsafe = (
```
rousseldenis: Putting the condition in `filtered` will allow to do just one `__set__` on the field for `True` values.

chienandalu: Done, thanks :)

rousseldenis (approved):

PR #3725, opened 2025-05-16
[18.0][FIX] sale_stock_picking_blocking: Fixed issue when duplicating a sales order.

rf-ife:
Removed @api.returns("self", lambda value: value.id) from the copy() method because it must return a recordset, not an ID. Returning an ID causes an error when duplicating records like sale.order, since the ORM expects a record, not a primitive value.

### DIFF
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -46,7 +46,6 @@ class SaleOrder(models.Model):
         order_to_unblock.order_line._action_launch_stock_rule()
         return True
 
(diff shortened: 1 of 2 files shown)


fb-ife (approved):

rousseldenis (approved):
Indeed. This is not needed anymore.

rousseldenis:
A little test for copy() should be great

rf-ife:
> A little test for copy() should be great
@rousseldenis a test has been added for copy() function

dreispt (approved):

PR #3729, opened 2025-05-22
[17.0] [FIX] sale_substate: Only Show Substates Pertaining to Main State

Nikul-OSI:


rousseldenis:
@Nikul-OSI Thanks for this. Could you include module name in commit message ?

Like `[IMP] sale_substate: (...)`

dreispt (changes requested):
Please include module name in commit message

sale_substate/views/sale_views.xml
```diff
@@ -12,6 +12,7 @@
                         name="substate_id"
                         widget="statusbar"
                         options="{'clickable': '1',}"
+                        domain="[('target_state_value_id.target_state_value', '=', state)]"
```
alexey-pelykh, line 15: This domain overrides the Python-level domain on `substate_id` (which filters by `model == self._name`), so the model filter is lost here.

Please change to:
```xml
domain="[('model', '=', 'sale.order'),('target_state_value_id.target_state_value', '=', state)]"
```

This matches the pattern used in `purchase_substate` and `purchase_request_substate` in `OCA/purchase-workflow`.

alexey-pelykh (changes requested):
Thanks for working on this -- the idea of filtering the statusbar to show only substates matching the current `state` is exactly the right fix. The `purchase_substate` module already does this, so there is a clear reference implementation.

However, the domain in this PR is **missing the model filter**. In Odoo, a `domain` attribute set on a `<field>` element in an XML view **overrides** the Python-level domain defined on the field. The `substate_id` field in `base.substate.mixin` already defines:

```python
domain=lambda self: [("model", "=", self._name)],
```

By adding only `[("target_state_value_id.target_state_value", "=", state)]` in the view, the Python domain is overridden, which means the model filter is lost. This could show substates from other models (e.g., `purchase.order` substates) if `base_substate` is used across multiple models in the same database.

The correct domain should include both filters, matching what `purchase_substate` does:

```xml
domain="[('model', '=', 'sale.order'),('target_state_value_id.target_state_value', '=', state)]"
```

Please update the domain to include both conditions.

sale_substate/views/sale_views.xml
alexey-pelykh, line 15: The domain is correct and well-formed. It properly filters `base.substate` records to only those whose `target_state_value_id.target_state_value` matches the current sale order's `state`.

This composes with the model-level domain `[('model', '=', self._name)]` already defined on the `substate_id` field in `base.substate.mixin`, so the combined effect filters by both model AND state -- exactly what's needed.

One minor style note: for consistency with other OCA modules, you might consider whether it would be cleaner to also include the model filter here explicitly (e.g. `[('model', '=', 'sale.order'), ('target_state_value_id.target_state_value', '=', state)]`), but since the model-level domain already handles it, this is fine as-is.

alexey-pelykh (commented):
## Review by CorporateHub

**Overall**: The fix is logically correct and addresses a real UX issue -- the substate statusbar currently shows all substates for the model regardless of the current main state, which is confusing and can lead to validation errors (the constraint `check_substate_id_value` in the mixin already prevents selecting mismatched substates, but the statusbar shouldn't offer them in the first place).

The domain expression `[('target_state_value_id.target_state_value', '=', state)]` correctly follows the relation chain: `base.substate` -> `target.state.value` -> `target_state_value` (char field holding e.g. `'draft'`, `'sale'`), and `state` refers to the current record's `state` field. This composes properly with the existing model-level domain `[('model', '=', self._name)]` defined on the `substate_id` field in `base.substate.mixin`.

### Issues found

| # | Severity | Finding |
|---|----------|--------|
| 1 | Minor (non-blocking) | **Commit message format**: OCA convention requires the module name prefix. Current: `[FIX] Only Show Substates Pertaining to Main State`. Expected: `[FIX] sale_substate: Only Show Substates Pertaining to Main State`. This was already flagged by @rousseldenis and @dreispt. |
| 2 | Note | **No tests for the domain filter**: The existing test suite doesn't verify that the statusbar filters substates by state. While testing XML domain attributes in isolation is tricky and the constraint already provides server-side safety, a test that verifies `_get_default_substate_domain` returns the correct domain for a given state would strengthen confidence. Not blocking. |

### Verification

- The `target_state_value_id.target_state_value` traversal is correct per the model definitions in `base_substate`.
- The `state` variable in the XML domain is resolved by the Odoo web client from the current record's `state` field, which is standard Odoo behavior for view domains.
- The domain dynamically updates when the main state changes (e.g., confirming a quotation), so the statusbar will correctly show only the substates for the new state.
- The `sale.order` model inherits both `sale.order` and `base.substate.mixin` (via `sale_substate/models/sale_order.py`), so `state` and `substate_id` are both available on the form.

**Verdict**: The code change itself is correct and ready. Please fix the commit message as requested by the maintainers, then this should be good to go.

rousseldenis:
@Nikul-OSI @dreispt What's the status of this ?

alexey-pelykh:
Correcting my earlier follow-up: I'd noted the view domain composes with the field's model-level domain — it doesn't. A `domain` on a `<field>` replaces the field's Python-level domain rather than AND-composing with it.

`substate_id` on `base.substate.mixin` defines `domain=lambda self: [("model", "=", self._name)]`. This view sets only `[('target_state_value_id.target_state_value', '=', state)]`, dropping the model term — so substates defined on other models that use the mixin can surface in the sale.order dropdown. `purchase_substate` re-includes it:

    domain="[('model', '=', 'purchase.order'), ('target_state_value_id.target_state_value', '=', state)]"

`sale_substate` should do the same with `('model', '=', 'sale.order')`.

rousseldenis:

PR #3730, opened 2025-05-22
[IMP][16.0] sale_order_invoicing_picking_filter: inheritance for included stock moves in invoice

manuelregidor:
Before this IMP, the way stock moves to be included in an invoice was part of the _create_invoices_from_pickings method in sale.order. That makes it difficult to modify the stock moves that are considered, in case some moves need to be added or removed.

It might happend, for example, if sale order lines have specific routes whose stock moves need a different processing when added to an invoice.

T-8291

### DIFF
--- a/sale_order_invoicing_picking_filter/models/sale_order.py
+++ b/sale_order_invoicing_picking_filter/models/sale_order.py
@@ -77,13 +77,7 @@ class SaleOrder(models.Model):
         incoming_invoice_vals_list = []
         for order in self:
             for code in ["outgoing", "incoming"]:
-                move_ids = (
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
Trivial

manuelregidor:
@Tisho99 @ValentinVinagre Could you review, please?

ValentinVinagre (approved):
I don't know if it would be more convenient to do it in 1 search function, but it's not badly done either.

Jaimermaccione (approved):
@manuelregidor LGTM! Review on local environment.

PR #3732, opened 2025-05-27
[17.0][ADD] sale_partner_utm_source

hailangvn:
This module provides a way to classify partners using Odoo’s native utm.source model. Each partner can be assigned a Source, which is added just below the partner's name on the form view. When creating a Quotation, the Source field of the partner is automatically copied to the Quotation. Users can manually adjust the Source in the Quotation if needed.

Depend on:

* https://github.com/OCA/partner-contact/pull/1886


sale_partner_utm_source/tests/test_sale_order.py
```diff
@@ -0,0 +1,67 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestSaleOrder(common.TransactionCase):
+    def setUp(self):
+        super().setUp()
```
rousseldenis: Please use classmethod instead for performance reasons.

hailangvn: Thanks for your review. The method has been updated.

rousseldenis (approved):

javierizaca (approved):
Functional review completed LGTM

PR #3733, opened 2025-05-27
[16.0][ADD] sale_order_line_no_print: new module

chienandalu:
cc @moduon MT-9255

@rafaelbn @yajo 

image


sale_order_line_no_print/models/sale_order.py
```diff
@@ -0,0 +1,16 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    hide_in_report = fields.Boolean(
+        help="This line will be hidden in quotations for customers"
+    )
```
yajo: suggestion: please [think positive when using booleans](https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/):

```suggestion
    display_in_report = fields.Boolean(
        default=True,
        help="Display this line in quotations for customers."
    )
```

sale_order_line_no_print/models/account_move.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    hide_in_report = fields.Boolean(help="This line will be hidden in invoice reports")
```
yajo: suggestion: same as the other comment.

yajo (changes requested):

sale_order_line_no_print/models/sale_order.py
chienandalu: oks, changed to fit the opposite

ypapouin:
@chienandalu , you also need to swap the display logic: `add="and line.display_in_report"`

chienandalu:
> @chienandalu , you also need to swap the display logic: `add="and line.display_in_report"`

Sure :man_facepalming:

sale_order_line_no_print/readme/USAGE.md
```diff
@@ -0,0 +1,6 @@
+To hide sale or account lines from the report:
+
+- Go to a sales order/quotation/invoice.
+- In the sale lines, click on the column selector to reveal the 🖶 column.
```
yajo: question: that weird character is intended?

```suggestion
- In the sale lines, click on the column selector to reveal the column.
```

sale_order_line_no_print/views/account_move_views.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="priority">99</field>
+        <field name="arch" type="xml">
+            <!-- `display_in_report` must be on both `invoice_line_ids` and `line_ids`
+                because move._move_autocomplete_invoice_lines_write() from write() call
+                can remove `invoice_line_ids` from write vals -->
+            <xpath expr="//field[@name='invoice_line_ids']/tree" position="inside">
+                <field
+                    string="🖶"
```
yajo: issue: this is how it looks in my computer:

image

Can't you use a fontawesome icon instead?

sale_order_line_no_print/views/report_invoice_document.xml
```diff
@@ -0,0 +1,17 @@
...
+        id="report_invoice_document"
+        inherit_id="account.report_invoice_document"
+        priority="99"
+    >
+        <xpath
+            expr="//t[@name='account_invoice_line_accountable']"
+            position="attributes"
+        >
+            <attribute name="t-if" separator=" " add="and line.display_in_report" />
+        </xpath>
```
yajo: idea: I don't know if you considered wrapping it. What you did is good, so this is just to let you consider it in case you prefer it:

```suggestion
        <xpath
            expr="//t[@name='account_invoice_line_accountable']"
            position="replace"
        >
            <t t-if="line.display_in_report">$0</t>
        </xpath>
```

The same applies in other views in case you like this way better. Otherwise it's OK as it is.

sale_order_line_no_print/views/sale_order_views.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="priority">99</field>
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/tree//field[@name='price_subtotal']"
+                position="after"
+            >
+                <field
+                    string="🖶"
```
yajo: issue: I only see a small square:

image

yajo (changes requested):

sale_order_line_no_print/views/account_move_views.xml
chienandalu: Oh, that's a shame... As far as I could see, there's now way to render the fa icon in the header instead :shrimp:

sale_order_line_no_print/views/report_invoice_document.xml
chienandalu: Cool one. Changed

sale_order_line_no_print/views/sale_order_views.xml
chienandalu: changed

yajo (approved):
Now it all looks good. However, I'm wondering... Shouldn't we require that you can only hide lines without amount? 🤔  Otherwise it's easy to shoot your own foot, hiding something with value and then the reports make no sense.

Well, one could then add another line that compensates the first one, so the final result is still 0. But that is weird 😵 

WDYT @rafaelbn?

chienandalu:
> Shouldn't we require that you can only hide lines without amount? 

I agree. Changed

sale_order_line_no_print/views/sale_order_views.xml
```diff
@@ -0,0 +1,26 @@
...
+                expr="//field[@name='order_line']/tree//field[@name='price_subtotal']"
+                position="after"
+            >
+                <field name="price_total" invisible="1" />
+                <field
+                    name="display_in_report"
+                    widget="boolean_toggle"
+                    options="{'autosave': False}"
+                    optional="hide"
+                    attrs="{'readonly': [('price_total', '!=', 0)], 'invisible': [('price_total', '!=', 0)]}"
```
yajo, line 21: thought: I'm not sure if this is enough... I could set total to zero, hide line, and change total. Maybe we should have a constraint in the model itself.

yajo (approved):

Jaimermaccione (commented):
@chienandalu Functional Review on Runbot environment. The module works as expected, but it does not allow modifying the report display status field when the document is a confirmed invoice.

While this behavior is understood to be intentional, it would be advisable to display a warning message to inform the user that the field cannot be edited in this state, in order to avoid confusion.

Thank you!

rousseldenis:
@chienandalu Could you explain the difference with:

https://github.com/OCA/sale-reporting/tree/14.0/sale_layout_category_hide_detail

sale_order_line_no_print/views/sale_order_views.xml
chienandalu: I'd be afraid of performance for such case...

chienandalu:
> Could you explain the difference with

@rousseldenis this goes on the line level without the need of defining sections. It's also meant to be passed to the stock.picking so it's possible to automatically hide some of the lines automatically (there isn't the concept of sections there)

chienandalu:
> it does not allow modifying the report display status field when the document is a confirmed invoice

Indeed, we're discussing internally whether to let hiding anything at all in the invoice as it could be problematic

yajo:
Please check https://github.com/OCA/sale-reporting/pull/329, which complements this PR.

rafaelbn:
Please @chienandalu rebase. And check test please

@rousseldenis we study deeply but it doesn't fit or expand it would be complicated as @chienandalu said.

chienandalu:
Ready for review

yajo (approved):

sale_order_line_no_print/views/report_saleorder_document.xml
```diff
@@ -0,0 +1,33 @@
...
+    <template
+        id="sale_order_portal_content"
+        inherit_id="sale.sale_order_portal_content"
+        priority="99"
+    >
+        <!-- Ensure compatibility if other modules extend the conditions matching
+         just the beggining of the condition-->
+        <xpath
+            expr="//t[starts-with(@t-if, 'not line.display_type')]"
+            position="replace"
```
rousseldenis, line 28: @chienandalu Don't you have any solution without `replace`?

chienandalu: Well, this is not really replacing the node, but wrapping it inside a `<t>` tag. So it's equivalent to:

```xml
<xpath expr="//t[starts-with(@t-if, 'not line.display_type')]" position="before">
  <t t-if="line.display_in_report" />
</xpath>
<xpath expr="//t[@t-if='line.display_in_report']" position="inside">
  <xpath expr="//t[starts-with(@t-if, 'not line.display_type')]" position="move">
</xpath>
```

yajo (approved):

rousseldenis (approved):

Gelojr (approved):
Congratulations on the work @chienandalu great job!
The following tests were executed and all passed:
Test 1: On a quotation for “Chiquito de la Calzada SA”, hid a 0.00€ info line via Display in report; line did not appear on PDF/portal and totals unchanged — OK.
Test 2: On a quotation for “Lola Flores SL”, attempted to hide a 120.00€ line; toggle was not editable/shown and the line appeared on the report — OK.
Test 3: On a quotation for “Alaska y Dinarama”, hid a 0.00€ line and then set price to 10.00€; the line became visible automatically on PDF/portal — OK.
Test 4: From SO for “Mortadelo y Filemón S.A.” with a hidden 0.00€ line, created an invoice; the hidden line did not appear on the invoice PDF — OK.
Test 5: On a customer invoice for “David Bisbal”, hid a 0.00€ line via Display in report; the line did not appear on the invoice PDF and totals remained the same — OK.

PR #3735, opened 2025-05-28
16.0 ref sale triple discount second try

legalsylvain:
Do not review for the time being.

### DIFF
--- a/sale_fixed_discount/__manifest__.py
+++ b/sale_fixed_discount/__manifest__.py
@@ -11,6 +11,7 @@
     "application": False,
     "installable": True,
     "depends": ["sale", "account_invoice_fixed_discount"],
+    "excludes": ["sale_triple_discount"],
     "data": [
         "security/res_groups.xml",
         "reports/report_sale_order.xml",
--- a/sale_fixed_discount/static/description/index.html
+++ b/sale_fixed_discount/static/description/index.html
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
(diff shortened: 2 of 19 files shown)


alexey-pelykh (commented):
This PR is currently marked as **Draft** and the description states "Do not review for the time being." Noting this for the record -- will revisit once the PR is marked as ready for review.

In the meantime, a few observations from a preliminary scan that may be useful when the refactoring is finalized:

**Architecture (major refactoring)**

The core change introduces a `discount1` field to replace the direct use of `discount`, making `discount` a computed (consolidated) read-only field. This is a significant architectural change that eliminates the `_aggregated_discount` context manager and `_convert_to_tax_base_line_dict` override -- a cleaner approach overall. The version bump to `16.0.2.0.0` with pre/post migration scripts is appropriate for this scope.

**Items to consider before marking ready:**

1. **`hooks.py` -- deprecated `pkg_resources`**: `pkg_resources.parse_version` is deprecated in favor of `packaging.version.Version`. While Odoo still ships `pkg_resources` via setuptools, it would be more forward-compatible to use `from packaging.version import Version`.

2. **`hooks.py` -- bare `Exception`**: The `raise Exception(msg)` should use a more specific exception type (e.g., `ImportError` or a custom exception). Bare `Exception` is harder to catch selectively.

3. **`_compute_discount` -- missing `return` / missing `continue` after early assignment**: In the new `_compute_discount` method (sale_order_line.py), when `line.display_type` is truthy, `discount1` is set to 0.0 but execution continues into the pricelist check. A `continue` is needed after the `line.discount1 = 0.0` assignment inside `if not line.product_id or line.display_type`.

4. **`_calc_discount_from_pricelist` -- implicit `None` return**: When `base_price == 0` or the discount condition is not met, the method returns `None` implicitly. This `None` will be assigned to `line.discount1`, which may cause issues. Consider returning `0.0` explicitly in those cases.

5. **`_prepare_invoice_line` -- additive branch**: In the `else` (additive) branch, only `discount1` is set to `self.discount` (the consolidated value). This means `discount2` and `discount3` values are lost on the invoice line for additive discounting. Is this intentional?

6. **`sale_order_general_discount` compatibility**: The `if "discount1" in self._fields` check in `sale_order_general_discount/models/sale_order_line.py` is a runtime field detection pattern. This works but couples the module to the field name. Consider whether the `excludes` mechanism (already used between `sale_fixed_discount` and `sale_triple_discount`) would be more appropriate here.

7. **SQL constraint**: The new `discount1_limit` constraint (`CHECK (discount1 <= 100.0)`) does not check for `>= -100.0`, which is inconsistent if negative discounts (surcharges) should also be bounded. The existing `discount2_limit` and `discount3_limit` have the same pattern, so this may be intentional -- but worth verifying.

8. **Migration -- `openupgradelib` dependency**: The migration scripts depend on `openupgradelib`. This should be documented as a dependency or handled gracefully if not installed.

9. **Test coverage**: Good to see new tests for `discount1`, `discount2`, and `discount3` individually (`test_01_2`, `test_01_3`). The consolidated `discount` field assertions throughout the existing tests strengthen confidence in the refactoring.

rousseldenis:
@legalsylvain This has already been fixed. Closing

PR #2276, opened 2022-11-23
[14.0][FIX] sale_procurement_group_by_line

MiquelRForgeFlow:
Same as https://github.com/OCA/sale-workflow/pull/2255 due to bad migration.

PR #2255, opened 2022-11-08
[15.0][FIX] sale_procurement_group_by_line

AaronHForgeFlow:
Use sale line procurement group when launching the stock rule again for the same sale order line.

Steps to reproduce:

1. Create a SO with at least one sale order line
2. Confirm it
3. Increase the quantity in the sale order line
    3.1 Current behavior: A new procurement group is used and assigned to the sales order, new delivery is created
    3.2 Desired behavior: The procurement group of the sale order line is used and same delivery is used.

cc @ForgeFlow

PR #3737, opened 2025-05-30
[15.0][FIX] sale_stock_delivery_address: remove obsolete code

RicardCForgeFlow:
Same as: https://github.com/OCA/sale-workflow/pull/2276 and https://github.com/OCA/sale-workflow/pull/2255

We should use _prepare_procurement_group_vals from the sale lines instead of _prepare_procurement_group_by_line from the order


sale_stock_delivery_address/models/sale_order_line.py
```diff
@@ -27,3 +27,12 @@ def _get_procurement_group_key(self):
         if self.dest_address_id:
             return (priority, self.dest_address_id)
         return (priority, key)
+
+    def _prepare_procurement_group_vals(self):
+        vals = super(SaleOrderLine, self)._prepare_procurement_group_vals()
+        if self._get_procurement_group_key()[0] == 16 and self.dest_address_id:
+            name_extension = (
```
rousseldenis: @RicardCForgeFlow Could you add a test to cover that code ?

RicardCForgeFlow: Hello @rousseldenis , do you think the condition `if self._get_procurement_group_key()[0] == 16` is necessary? I don't understand why it was added :disappointed:

StefanRijnhart: @LoisRForgeFlow do you remember why you checked for priority 16 in https://github.com/OCA/sale-workflow/pull/1068/files#diff-b98a7947befb3af9a4bd8af3543b652674adced1bb03541ada92c50be71c30aaR14? Is there a module in another repo that returns this in its override of `_get_procurement_group_key`?

LoisRForgeFlow: @StefanRijnhart I don't remember to be honest. I have reviewed in different versions codebases (I don't even have v11 in my computer anymore) and I have found no use for this. 

Let's assume it was a customization that is death at this point. You can remove it.

StefanRijnhart: @LoisRForgeFlow Great, thanks for coming back on this so quickly!

StefanRijnhart: @RicardCForgeFlow You have now removed the clause that checks for priority 16, but I interpret Lois' comments differently. If the customization that returns priority 16 is obsolete, then the alternate behaviour that needs to happen in this case is also obsolete, and should not be upgraded to the default behaviour. I think `_prepare_procurement_group_vals` can be removed entirely now.

RicardCForgeFlow: Hi @StefanRijnhart and @LoisRForgeFlow , just to confirm
We don't want to differentiate the name of the procurement groups?
image
If we delete `_prepare_procurement_group_vals`:
image

StefanRijnhart: Indeed. Judging from the code, the bottom version is what is the current behaviour (for lack of a module that returns priority 16). While it may be nice to distinguish procurement groups by a different name, addresses can be long and things could get cluttered quickly if we include them in the procurement group name.

RicardCForgeFlow: I agree, by deleting the method we continue with the current behaviour. :+1:

StefanRijnhart (approved):

sale_stock_delivery_address/models/sale_order_line.py
LoisRForgeFlow: Exactly, thanks everybody :+1:

LoisRForgeFlow (approved):
:+1:

PR #3738, opened 2025-05-30
[18.0][MIG]sale_triple_discount: Migration to v18

matteotognini:
PR based on #3648 by @sbejaoui. As suggested by @lef-adhoc, I've added some code from their repository to improve the  managements of pricelists. Previously, the discount from the pricelist was not being applied to the sale order line.


sale_triple_discount/models/sale_order_line.py
```diff
@@ -131,3 +131,36 @@ def _prepare_invoice_line(self, **kwargs):
...
+                discount_factor = 1.0
+                for discount in [line.discount1, line.discount2, line.discount3]:
+                    discount_factor *= (100.0 - discount) / 100.0
+                line.discount = 100.0 - (discount_factor * 100.0)
+
+    @api.depends("discount")
+    def _compute_discounts(self):
+        for line in self:
+            if line.check_compute_discounts():
+                line.discount1 = line.discount
```
rousseldenis, line 150: @matteonext You can't do that !  These fields are not computed. So, this is not syntaxically correct.

rousseldenis (changes requested):
We'll see to get a better solution than this on https://github.com/OCA/sale-workflow/pull/3648

sale_triple_discount/models/sale_order_line.py
matteotognini: Hi @rousseldenis, Thanks for the review! I'll close this Pr, as you told me, here #3648 there is a better solution!

PR #3742, opened 2025-06-03
[15.0][ADD] sale_invoice_negative_quantities: New module sale_invoice_negative_quantities

BernatObrador:
This module allows invoicing negative quantities on sales order, when product policy is set to delivered quantities.

cc https://github.com/APSL 11077

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review


peluko00 (approved):
LGTM!

sale_invoice_negative_quantities/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+    def button_validate(self):
+        res = super().button_validate()
+        self._update_negative_deliveries()
+        return res
+
+    def _update_negative_deliveries(self):
+        for record in self:
+            for move in record.move_lines:
+                so_line = move.sale_line_id
+                if so_line and so_line.product_uom_qty < 0:
```
rousseldenis: As `product_uom_qty` uses digit precision, you should use `float_compare`.

See: https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order_line.py#L257

sale_invoice_negative_quantities/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows invoicing negative quantities on sales order, when product policy is set to delivered quantities.
```
rousseldenis: @BernatObrador This could be enhanced by adding a CONTEXT to explain when it is useful.

rousseldenis (changes requested):

BernatObrador:
@rousseldenis Thanks for your feedback!
Could you review again please?

mpascuall (approved):
LGTM

sale_invoice_negative_quantities/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Moduon Team S.L.
```
rousseldenis: @BernatObrador Is it a copy paste ?

sale_invoice_negative_quantities/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
+from odoo import models
```
rousseldenis, line 3: You should include licensing here and in test file

sale_invoice_negative_quantities/tests/test_sale_invoive_negative_qty.py
```diff
@@ -0,0 +1,69 @@
...
+            {
+                "partner_id": cls.partner.id,
+            }
+        )
+
+        cls.sale_order_line = cls.env["sale.order.line"].create(
+            {
+                "order_id": cls.sale_order.id,
+                "product_id": cls.product.id,
+                "product_uom_qty": -2.0,
```
rousseldenis, line 34: IMHO, this does not correspond to your use case.

BernatObrador: Hello @rousseldenis, just added more tests. Could you review again?

Thanks!

sale_invoice_negative_quantities/models/stock_picking.py
```diff
@@ -0,0 +1,29 @@
...
+                    so_line
+                    and float_compare(
+                        so_line.product_uom_qty,
+                        0.0,
+                        precision_rounding=so_line.product_uom.rounding,
+                    )
+                    < 0
+                ):
+                    delivered = so_line.qty_delivered or 0.0
+                    so_line.qty_delivered = -(delivered + move.quantity_done)
```
alexey-pelykh, line 29: **Bug: incorrect accumulation on partial deliveries.**

The formula `-(delivered + move.quantity_done)` breaks when there are multiple partial deliveries for the same SO line.

Example scenario:
- SO line qty = -2, first partial delivery of 1 unit
  - `delivered` = 0, `quantity_done` = 1 --> `qty_delivered = -(0 + 1) = -1` (correct)
- Second partial delivery of 1 unit
  - `delivered` = -1, `quantity_done` = 1 --> `qty_delivered = -((-1) + 1) = 0` (wrong, expected -2)

Suggested fix:
```python
so_line.qty_delivered = delivered - move.quantity_done
```
This correctly subtracts each new delivery from the running total:
- First: `0 - 1 = -1`
- Second: `-1 - 1 = -2`

sale_invoice_negative_quantities/models/stock_picking.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo.tools.float_utils import float_compare
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def button_validate(self):
+        res = super().button_validate()
+        self._update_negative_deliveries()
+        return res
```
alexey-pelykh, line 13: `super().button_validate()` can return a wizard action dict (e.g., the immediate transfer or backorder wizards) instead of `True`. When a wizard is returned, the picking is **not yet validated** -- the user must interact with the wizard first.

Calling `_update_negative_deliveries()` here would prematurely update `qty_delivered` before the actual validation completes.

Consider either:
1. Checking `if not isinstance(res, dict):` before calling the update, or
2. Hooking into `_action_done()` instead of `button_validate()`, which is called only when the picking is truly validated.

sale_invoice_negative_quantities/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025 APSL-Nagarro Bernat Obrador <borbador@apsl.net>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+{
+    "name": "Sale Invoice Negative Quantities",
+    "summary": "Enbales negative quantities in delivery quantity on sale invoice",
```
alexey-pelykh, line 6: Typo: "Enbales" should be "Enables".

sale_invoice_negative_quantities/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,16 @@
+This module allows invoicing negative quantities on sales orders, when the product’s invoicing policy is set to **delivered quantities**.  
+It is particularly useful for businesses that work with **returnable containers** or **returnable products** in general.
+
+### Example use case
```
alexey-pelykh, line 4: This file is reStructuredText but uses Markdown heading syntax (`### Example use case`). In the generated `index.html`, this renders as literal text rather than as a heading. Please use RST syntax instead, e.g.:

```rst
Example use case
~~~~~~~~~~~~~~~~
```

alexey-pelykh (changes requested):
## Review: sale_invoice_negative_quantities

Thank you for this module. The use case for returnable containers is legitimate and the overall approach of overriding `button_validate` on `stock.picking` to adjust `qty_delivered` for negative SO lines is reasonable. However, I found several issues that should be addressed before merging.

### Key Findings

| 1 | Critical | `models/stock_picking.py:28-29` | **Incorrect delivered quantity calculation.** The formula `-(delivered + move.quantity_done)` accumulates incorrectly on subsequent deliveries. If a partial delivery of 1 is done first (`qty_delivered` becomes -1), then a second delivery of 1 is done, the formula computes `-((-1) + 1) = 0` instead of the expected `-2`. The logic should be `so_line.qty_delivered = delivered - move.quantity_done` (i.e., subtract the newly done quantity from the current delivered total). |
| 2 | Moderate | `models/stock_picking.py` | **Missing `stock` dependency.** The module extends `stock.picking` but only declares `sale_management` as a dependency. While `sale_management` transitively depends on `stock`, the OCA convention is to explicitly declare direct dependencies on modules whose models you extend. Add `"stock"` to the `depends` list in `__manifest__.py`. |
| 3 | Moderate | `models/stock_picking.py:10-13` | **`button_validate` override may skip update on wizard return.** `super().button_validate()` can return a wizard action dict (e.g., immediate transfer wizard) instead of completing the validation. In that case, `_update_negative_deliveries()` would run prematurely before the picking is actually validated. The method should check whether `res` is a dict (wizard action) and skip the update if so, or alternatively hook into `_action_done` instead. |
| 4 | Moderate | `models/stock_picking.py:17` | **`move_lines` is deprecated in Odoo 15+.** The field was renamed to `move_ids` (with `move_lines` kept as a deprecated alias). While it still works in 15.0, using the canonical name `move_ids` is preferred for forward compatibility and to follow OCA linting standards. |
| 5 | Moderate | `__manifest__.py:6` | **Typo in summary.** "Enbales" should be "Enables". |
| 6 | Minor | `tests/test_sale_invoive_negative_qty.py` | **Typo in test filename.** "invoive" should be "invoice" -- `test_sale_invoice_negative_qty.py`. Also update the import in `tests/__init__.py`. |
| 7 | Minor | `readme/DESCRIPTION.rst` | **Markdown syntax in RST file.** The description uses `### Example use case` (Markdown heading) and backtick-based inline code, but OCA README fragments are reStructuredText. This renders incorrectly in the generated `index.html` (visible as literal `### Example use case` text). Use RST heading syntax instead. |
| 8 | Minor | `__manifest__.py:1` | **Inconsistent email in copyright.** The copyright header says `<borbador@apsl.net>` but the CONTRIBUTORS.rst says `<bobrador@apsl.net>`. One of these is a typo. |
| 9 | Minor | `tests/test_sale_invoive_negative_qty.py:107-108` | **Fragile index-based access to invoice lines.** `invoice_lines[0]` and `invoice_lines[1]` assume a specific ordering. If the order changes, the test breaks silently or gives a misleading failure. Filter by `price_unit` or `quantity` sign to identify negative vs positive lines explicitly. |

### Additional Observations

- The `_update_negative_deliveries` method directly writes to `qty_delivered` which is a computed/stored field on `sale.order.line`. Bypassing the standard computation can cause inconsistencies if other modules also compute this field. Consider whether overriding `_get_delivered_qty()` on `sale.order.line` would be a more robust integration point -- it would make the negative quantity logic participate in the standard delivery quantity computation chain rather than overwriting it after the fact.
- The test `test_negative_and_positive_delivery_update` adds a new SO line after `action_confirm()`. This triggers procurement and creates new pickings, which is the scenario being tested. The assertions on picking counts (2) and move line counts (1 each) are good but could be more descriptive with error messages.

Overall, finding #1 (incorrect accumulation logic) is a correctness bug that will manifest with partial deliveries. Finding #3 (premature update when wizard is returned) could cause silent data corruption. Both should be addressed.

sale_invoice_negative_quantities/models/stock_picking.py
rousseldenis: @BernatObrador Could you add a test for several partial deliveries as this seems legit ?

PR #3743, opened 2025-06-03
[18.0][FIX] sale_exception: update rule due to product type change

matiasperalta1:


matiasperalta1:
@pedrobaeza Hi, could you merge this change? This is because the type of product "product" no longer exist and it was replaced by "consu". So, this rule did not work.

lav-adhoc (approved):

matiasperalta1:
@rousseldenis @StefanRijnhart Hi! Could you merge this change? This is because of the type of product "product" no longer exist and it was replaced by "consu". So, this rule did not work.

lef-adhoc (approved):

sale_exception/data/sale_exception_data.xml
```diff
@@ -26,7 +26,7 @@
         <field name="model">sale.order.line</field>
         <field
             name="code"
-        >failed=self.product_id and self.product_id.type == 'product' and self.virtual_available_at_date &lt; self.product_uom_qty</field>
+        >failed=self.product_id and self.product_id.type == 'consu' and self.virtual_available_at_date &lt; self.product_uom_qty</field>
```
StefanRijnhart: The equivalent of checking product type 'product' in 18 would be to check 'is_storable'. Or is there a reason that 'consu' is fine here too?

rousseldenis: @matiasperalta1

rousseldenis: @StefanRijnhart In fact, the `consu` products can have their available quantity evaluated. But to track quants in stock, you need to set it `is_storable`. So, in this case, I think the change is correct.

StefanRijnhart: Oh, I always thought that the point of consumables is that you can sell regardless of inventory. So as a user, I would be surprised to find that an issue is raised on the order for such a product because of insufficient forecasted quantity.

StefanRijnhart:
@rousseldenis Isn't it more like a bug than an enhancement?

matiasperalta1:
@StefanRijnhart @rousseldenis I think it's the right change. Could we merge it?

sale_exception/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Sale Exception",
     "summary": "Custom exceptions on sale order",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
StefanRijnhart: Don't bump the version number manually. Ocabot will take care of that during the merge.

StefanRijnhart:
@sebastienbeau I believe you are the original author, do you have an opinion on the consu/is_stockable issue?

sale_exception/__manifest__.py
StefanRijnhart: Unless you want to add a migration script, which actually makes sense, because the modified rule is noupdate!

matiasperalta1:
@StefanRijnhart Everything seems ok. But we can wait the author opinion

matiasperalta1:
@rousseldenis Should we continue with the merge?

sale_exception/data/sale_exception_data.xml
matiasperalta1: Thanks @StefanRijnhart  and @rousseldenis. We had a real case where a consumable product (not tracked in inventory) triggered the constraint unexpectedly.

To avoid confusion for users, we refined the condition so that it only applies when the product is both consu and is_storable = True.
This way, consumables that aren't tracked won't block the order.

StefanRijnhart (approved):
Looks like the test failure is unrelated. Hope it gets fixed soon!

sale_exception/data/sale_exception_data.xml
sebastienbeau: Hi sorry for the delay.

Looking at the code the field "is_storable" is define in the module stock (so not in the dependency, same with virtual_available_at_date). **BUT** as this rule is not activated by default, and it only make sense to activate it when you use stock, it's ok to me to have this "none explicit" dependency. (Creating a glue module for that is overkill)

Thanks for you work !

sebastienbeau (approved):
LGTM (code review)

PR #3744, opened 2025-06-05
[16.0][IMP] sale_pricelist_from_commitment_date force date on compute price rule

santostelmo:
…
This improvement allows `_compute_price_rule` to obtain the price and rule on a specific date passed by context

### DIFF
--- a/sale_pricelist_from_commitment_date/models/product_pricelist.py
+++ b/sale_pricelist_from_commitment_date/models/product_pricelist.py
@@ -13,3 +13,11 @@ class ProductPricelist(models.Model):
         if force_pricelist_date:
             date = force_pricelist_date
         return super()._get_product_rule(product, quantity, uom, date, **kwargs)
+
+    def _compute_price_rule(self, products, quantity, uom=None, date=False, **kwargs):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super()._compute_price_rule(
+            products, quantity, uom=uom, date=date, **kwargs
+        )
--- a/sale_pricelist_from_commitment_date/tests/test_pricelist_from_commitment_date.py
+++ b/sale_pricelist_from_commitment_date/tests/test_pricelist_from_commitment_date.py
@@ -100,6 +100,16 @@ class PricelistFromCommitmentDate(TransactionCase):
         self.assertEqual(order_line.price_unit, product.list_price)
         # Test with commitment date
         sale.commitment_date = "2020-03-08"
+        self.assertEqual(
+            sale.pricelist_id.with_context(
+                force_pricelist_date=sale.commitment_date
+            )._compute_price_rule(product, 1.0, date=order_line.order_id.date_order)[
+                product.id
+            ][
+                0
+            ],
+            10,
+        )
         self.assertEqual(order_line.price_unit, 10)
         sale.commitment_date = "2020-03-12"
         self.assertEqual(order_line.price_unit, 20)


gurneyalex (approved):

wouitmil (approved):

cyrilmanuel:
@OCA/sale-blanket-maintainers someone can merge this PR please ?

cyrilmanuel:
hi @sbidoul , could you merge this pr please ? or let me know who can help me :)

alexey-pelykh (approved):
## Review: sale_pricelist_from_commitment_date - force date on _compute_price_rule

**Target**: 16.0 (supported)
**Scope**: +16 lines, 2 files, 1 commit

### Summary

This PR adds an override of `_compute_price_rule` on `product.pricelist` so that the `force_pricelist_date` context key (already honored by `_get_product_rule`) is also respected when computing price rules. This is a logical completeness fix -- without it, code paths going through `_compute_price_rule` would ignore the commitment date.

### Findings

| Area | Status | Notes |
|------|--------|-------|
| Correctness | OK | Override follows the exact same pattern as the existing `_get_product_rule` override. Method signature matches Odoo 16 base (positional args align). |
| Odoo conventions | OK | Properly uses `super()` delegation, context-based date override is the established pattern in this module. |
| Security | OK | No new input vectors. Context key usage is standard Odoo practice. |
| Performance | OK | Single context dict lookup, no additional queries. |
| Tests | OK | New test assertion directly exercises `_compute_price_rule` with `force_pricelist_date` context and verifies correct price is returned for the commitment date range. |
| Manifest | OK | No version bump needed -- this is an improvement to existing functionality, and the manifest was not changed. A version bump to `16.0.1.1.0` would be *nice-to-have* but is not blocking. |
| i18n | OK | No user-facing strings added. |
| License headers | OK | Existing files, no new files added. |

### Minor observation (non-blocking)

The parameter name `quantity` in the override differs from the base Odoo 16 signature which uses `qty`:
```python
# Base Odoo 16:
def _compute_price_rule(self, products, qty, uom=None, date=False, **kwargs):

# This PR:
def _compute_price_rule(self, products, quantity, uom=None, date=False, **kwargs):
```
This works correctly since the argument is passed positionally, but matching the base parameter name (`qty`) would be more consistent and avoid potential confusion if someone tries to call with `qty=` as a keyword argument. Not blocking.

APPROVED -- clean, minimal, well-tested improvement.

cyrilmanuel:
hi @pedrobaeza  could you merge this PR please ?

sale_pricelist_from_commitment_date/models/product_pricelist.py
```diff
@@ -13,3 +13,9 @@ def _get_product_rule(self, product, quantity, uom=None, date=False, **kwargs):
         if force_pricelist_date:
             date = force_pricelist_date
         return super()._get_product_rule(product, quantity, uom, date, **kwargs)
+
+    def _compute_price_rule(self, products, quantity, uom=None, date=False, **kwargs):
+        force_pricelist_date = self.env.context.get("force_pricelist_date")
+        if force_pricelist_date:
+            date = force_pricelist_date
+        return super()._compute_price_rule(products, quantity, uom, date, **kwargs)
```
pedrobaeza: Shouldn't be this for not altering positional and keyword arguments?

```suggestion
        return super()._compute_price_rule(products, quantity, uom=uom, date=date, **kwargs)
```

santostelmo: Thanks @pedrobaeza. I updated the code to pass the same arguments as it is in the super method.

PR #3746, opened 2025-06-06
[18.0][IMP] sale_invoice_policy: automatic invoice

jbaudoux:
Mark the SO invoicing policy on order when the SO lines are forced to order by the automatic invoicing.

To better support this odoo feature:
* https://github.com/odoo/odoo/blob/57b4056798b9a380027fd3da1c4f3965249a4509/addons/sale/models/sale_order.py#L1806

Also added openupgradelib in external dependencies as it is called by pre-init-hook

cc @marielejeune @rousseldenis

### DIFF
--- a/sale_invoice_policy/tests/test_sale_invoice_policy.py
+++ b/sale_invoice_policy/tests/test_sale_invoice_policy.py
@@ -1,4 +1,5 @@
-# © 2017 Acsone SA/NV (http://www.acsone.eu)
+# Copyright 2017 Acsone SA/NV (http://www.acsone.eu)
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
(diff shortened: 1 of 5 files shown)


sale_invoice_policy/tests/test_sale_invoice_policy.py
```diff
@@ -167,3 +168,21 @@ def test_context_manager_exception(self):
             "the same invoice policy",
             exc.exception.args[0],
         )
+
+    def _force_lines_to_invoice_policy_order(self):
```
marielejeune: Is the test run? I thought only methods prefixed with `test_` are run inside test files.

jbaudoux: oh good catch!

marielejeune (approved):
LGTM (code review)

rousseldenis (approved):
LGTM

jbaudoux:
@OCA/crm-sales-marketing-maintainers Could you merge this improvement ?

PR #3747, opened 2025-06-06
[IMP] sales_security_team: revert changes to recover esencial functionality

matiasperalta1:


legalsylvain:
Hi. Could you change message to explain the changes ?

Thanks !

sales_team_security/models/ir_rule.py
```diff
@@ -26,19 +26,34 @@ def _compute_domain(self, model_name, mode="read"):
         """
         res = super()._compute_domain(model_name, mode=mode)
         user = self.env.user
-        group_my_records = "sales_team.group_sale_salesman"
-        group_all_records = "sales_team.group_sale_salesman_all_leads"
+        group1 = "sales_team.group_sale_salesman"
```
rousseldenis, line 31: @matiasperalta1 IMHO, changing variables names from explicit to not explicit (groupX) is a drawback (and is not needed).

sales_team_security/models/ir_rule.py
```diff
@@ -26,19 +26,34 @@ def _compute_domain(self, model_name, mode="read"):
...
-                domain_followers = [
+            if user.has_group(group1) and not user.has_group(group3):
+                extra_domain = [
                     "|",
                     ("message_partner_ids", "in", user.partner_id.ids),
+                    "|",
                     ("id", "=", user.partner_id.id),
                 ]
-                domain_user = [("user_id", "in", [user.id, False])]
-                extra_domain = expression.OR([domain_followers, domain_user])
```
rousseldenis, line 41: IMHO, using `AND()` and `OR()` is better in terms of readability.

sales_team_security/models/ir_rule.py
```diff
@@ -26,19 +26,34 @@ def _compute_domain(self, model_name, mode="read"):
...
         if model_name == "res.partner" and not self.env.su:
-            if user.has_group(group_my_records) and not user.has_group(
-                group_all_records
-            ):
-                domain_followers = [
+            if user.has_group(group1) and not user.has_group(group3):
+                extra_domain = [
                     "|",
                     ("message_partner_ids", "in", user.partner_id.ids),
+                    "|",
```
rousseldenis, line 43: I think this is not useful as there is only one argument after this 'OR'

sales_team_security/models/ir_rule.py
```diff
@@ -26,19 +26,34 @@ def _compute_domain(self, model_name, mode="read"):
...
+                    extra_domain += [
+                        "|",
+                        ("user_id", "=", user.id),
+                        "&",
+                        ("user_id", "=", False),
+                        "|",
+                        ("team_id", "=", False),
+                        ("team_id", "=", user.sale_team_id.id),
+                    ]
+
```
rousseldenis, line 59: If I summarize the change here:

- If user is a manager, can see all contacts of a team (even with user assigned)
- either, user can see only its contacts or contacts with no assignation and (no team or same team).


This should be reflected in documentation

sales_team_security/models/res_partner.py
```diff
@@ -11,6 +13,64 @@ class ResPartner(models.Model):
 
     # add indexes for better performance on record rules
     user_id = fields.Many2one(index=True)
+    team_id = fields.Many2one(
```
rousseldenis: You introduce a new functionality here. Maybe this should be done in another module.

sales_team_security/models/res_partner.py
```diff
@@ -11,6 +13,64 @@ class ResPartner(models.Model):
...
+
+    @api.onchange("parent_id")
+    def _onchange_parent_id_sales_team_security(self):
+        """If assigning a parent partner and the contact doesn't have
+        team or salesman, we put the parent's one (if any).
+        """
+        if self.parent_id and self.parent_id.team_id and not self.team_id:
+            self.team_id = self.parent_id.team_id.id
+        if self.parent_id and self.parent_id.user_id and not self.user_id:
+            self.user_id = self.parent_id.user_id.id
```
rousseldenis, line 57: IMHO, this is totally out of scope. Moreover this is already done by: https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/res_partner.py#L397

Those changes should be done in computes.

sales_team_security/models/res_partner.py
```diff
@@ -22,8 +82,8 @@ def _remove_key_followers(self, partner):
 
     def _add_followers_from_salesmen(self):
         """Sync followers in commercial partner + delivery/invoice contacts."""
-        for record in self.commercial_partner_id:
-            followers = (record.child_ids + record).user_id.partner_id
+        for record in self.mapped("commercial_partner_id"):
```
rousseldenis: Those changes are not necessary as Odoo already include a `mapped()` when you access a relation through a multi recordset

rousseldenis (changes requested):

nicolascol:
Hi!

First, in OCA’s runboat, the sales_team_security module is not working. You can test it as follows:
1. Create a sales team with both "admin" and "demo" in the same team.
2. Assign a res.partner to "admin" as the "salesperson".
3. Assign the "demo" user to the sales group "see documents of own team only".
4. Log in as "demo".
5. The user is not able to see the corresponding res.partner, although they should.

Second, we have reintroduced the sales team field in res.partner because, in cases where a salesperson belongs to multiple sales teams, we don’t want the res.partner to be visible across all teams, only in the one it's assigned to. (as it is v17). One thing to keep in mind is that Odoo deprecated the sales team in res.partner because they didn't use it.

I’m sharing a video in Spanish, but you can use YouTube’s automatic translator.
https://www.youtube.com/watch?v=N2bgoG_EXUE
At the end of the video, you can see a small bug in this PR, which we're already working on.

sales_team_security



@rousseldenis @pedrobaeza @matiasperalta1 @carlos-lopez-tecnativa

sales_team_security/models/res_partner.py
carlos-lopez-tecnativa: As @matiasperalta1 mentioned, and as I commented in the migration PR, Odoo removed this field in this commit https://github.com/odoo/odoo/commit/d11691c0312470c8eeae0f3061953adc8be926c4 
I'm not sure if reintroducing this field is good for the user experience.
For example, if I have the following settings in the sales teams:
image
image
With this new field, I can encounter inconsistencies like in the following image: the sales team is selected independently of the user, and the user is not a member of that team.
image

carlos-lopez-tecnativa (commented):
Please, check the CI

nicolascol:
In previous versions, it was possible to select a different team than the salesperson's team in res.partner.
I understand that Odoo removed the field because it wasn't used much (it was a field visible only in debug mode) and it caused confusion for them in identifying which team a salesperson belonged to. On the other hand, with the "sales_team_security" module, our goal is to identify the customer.

Regarding the development, we have two options:
1. We can make the team of the salesperson compute automatically in res.partner. If the salesperson belongs to more than one team, we take the one with the lowest sequence.
2. We fix the display for the "Team Documents Only" group. We do not add team_id in res.partner and clarify in the README that the module, starting from version 18, may be non-functional if "multi-teams" is enabled.

We would prefer to go with option 1 because we believe it reflects the essence of the module, and several of our clients would be affected if we went with option 2.

pedrobaeza:
IMO, `team_id` field shouldn't be restored in `res.partner`, but the security rules work on `res.user~team_ids`.

alexey-pelykh (changes requested):
## Review: sales_team_security improvements

Thank you for working on fixing the `res.partner` visibility issue for team members. The core problem (team members in `group_sale_team_manager` not seeing their team's contacts) is a real gap. However, this PR has several issues that need to be addressed before it can be merged.

### Critical Issues

**1. Extensive commented-out code throughout the diff**

The PR contains large blocks of commented-out code in multiple files (`res_partner.py`, `ir_rule.py`). OCA coding standards require that dead/commented code be removed, not left in. If the `team_id` field approach has been abandoned (as suggested by maintainers @pedrobaeza and @rousseldenis), the commented blocks should be deleted entirely. If it is still being considered, it should be in a separate PR.

Affected locations:
- `res_partner.py`: Commented-out `team_id` field definition (lines ~13-24), `_onchange_user_id_sales_team_security` (lines ~57-59), `_compute_team_id` (lines ~61-68), `_sync_team_id_to_children` (lines ~113-120), and references in `create`/`write`
- `ir_rule.py`: Commented-out original condition (lines ~34-36)

**2. Domain construction in `ir_rule.py` is incorrect**

The new `extra_domain` construction manually builds a Polish-notation domain by concatenating lists with `+=`. This is fragile and hard to verify. The original code used `expression.OR()` which is both safer and more readable.

Additionally, as @rousseldenis noted, the second `"|"` operator on line 43 (`"|", ("id", "=", user.partner_id.id)`) only has one leaf after it before the `+=` appends more items. The final domain shape depends on runtime concatenation, making it very difficult to reason about correctness. The `expression.OR()` and `expression.AND()` helpers should be used instead of manual prefix-notation construction.

**3. Variable names `group1`, `group2`, `group3` are non-descriptive**

This was already flagged by @rousseldenis and has not been addressed. The original names (`group_my_records`, `group_all_records`) were self-documenting. The new names provide no hint about what each group represents. Please restore descriptive names.

**4. `__manifest__.py` does not declare the new view data file**

The PR adds `views/res_partner_view.xml` but the manifest `data` key still only lists `["security/sales_team_security.xml"]`. The new view file will not be loaded at install/upgrade, making it dead code.

**5. Tests reference `team_id` field and functionality that is commented out**

Several new tests (`test_onchange_parent_id`, `test_onchange_user_id`, `test_assign_contacts_team`, `test_partner_fields_view_get`) reference `team_id` on `res.partner`, but the field definition is commented out in `res_partner.py`. These tests will fail because `team_id` is not defined by this module (it was removed by Odoo in v18). Similarly, `test_assign_contacts_team` imports `post_init_hook` and expects it to sync `team_id` to children, but `post_init_hook` does not do that.

The `_check_permission` method in `common.py` now unconditionally writes `team_id`, which will raise an error if the field does not exist on the model being tested.

**6. `get_view` override uses fragile string manipulation**

The `get_view` override uses `str.replace("{", "{'default_user_id': user_id,", 1)` to inject context into the XML arch. This is brittle -- if the existing context already contains a `default_user_id` key, it will produce invalid Python dict syntax. The standard Odoo approach is to use view inheritance XML to set context on the field, not runtime arch manipulation.

**7. The `get_view` injects `default_team_id` but the test checks for it while the field is commented out**

`test_partner_fields_view_get` asserts `"default_team_id" in xml_fields[0].get("context", "")` but the current `get_view` code only injects `default_user_id` (the `team_id` part was removed). The test expects functionality that is no longer in the code.

### Moderate Issues

**8. PR title and description need improvement**

The PR title contains a typo ("esencial" should be "essential") and the body is empty (`null`). As requested by @legalsylvain, a clear description of what the PR changes and why is needed for OCA review.

**9. Unnecessary `mapped()` change was reverted but inconsistently**

The diff shows `(record.child_ids + record).mapped("user_id.partner_id")` replacing `(record.child_ids + record).user_id.partner_id`. As @rousseldenis noted, accessing a relational field on a multi-recordset already does the equivalent of `mapped()`. This change is unnecessary.

**10. Spanish comments in the code**

The inline comments in `ir_rule.py` (`# Ver todos los contactos de su equipo`, `# Ver solo propios y sin asignar`) are in Spanish. OCA coding standards require English for all code comments.

### Recommendation

Based on the maintainer feedback from @pedrobaeza ("team_id field shouldn't be restored in res.partner, but the security rules work on res.user~team_ids"), I would suggest:

1. Remove all commented-out code
2. Remove the `team_id` field reintroduction and all related `get_view`/onchange/compute logic
3. Focus the PR on fixing the actual security rule to use `user.sale_team_id` (or `user.crm_team_ids`) for the team manager group check, without adding `team_id` to `res.partner`
4. Fix the domain construction to use `expression.OR()`/`expression.AND()`
5. Restore descriptive variable names
6. Update the manifest if new data files are kept
7. Add a proper PR description
8. Translate comments to English

This would make the PR much cleaner and aligned with maintainer expectations.

rousseldenis:

PR #3748, opened 2025-06-09
[sale_company_currency] FIX: Use currency._convert for proper currenc…

straga:
…y conversion

Replace manual currency_rate multiplication with standard Odoo currency._convert method for accurate currency conversion:

- Fixed _compute_amount_company to use currency_id._convert()
- Added proper conversion date handling (date_order or today)
- Updated dependencies to include currency_id, company_id, date_order
- Enhanced tests to validate conversion using _convert method
- Added test for reversed currency conversion (USD to EUR)

This ensures proper handling of currency rates, rounding rules, and date-specific exchange rates as per Odoo standards.

### DIFF
--- a/sale_company_currency/models/sale_order.py
+++ b/sale_company_currency/models/sale_order.py
@@ -23,11 +23,21 @@ class SaleOrder(models.Model):
         store=True,
     )
 
-    @api.depends("amount_total", "currency_rate")
+    @api.depends("amount_total", "currency_id", "company_id", "date_order")
     def _compute_amount_company(self):
         for order in self:
-            if order.currency_id.id == order.company_id.currency_id.id:
-                to_amount = order.amount_total
+            # order.currency_id == order.company_id.currency_id: 
+            if order.currency_id == order.company_id.currency_id:
+                order.amount_total_curr = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
-            order.amount_total_curr = to_amount
+                # Convert order amount to company currency
+                conversion_date = (
+                    order.date_order and order.date_order.date()
+                    or fields.Date.context_today(self)
+                )
+                order.amount_total_curr = order.currency_id._convert(
+                    order.amount_total,
+                    order.company_id.currency_id,
+                    order.company_id,
+                    conversion_date,
+                )
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -1,4 +1,5 @@
 from odoo.tests.common import TransactionCase
+from odoo import fields
 
 
 class TestSaleOrder(TransactionCase):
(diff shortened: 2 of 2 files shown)


sale_company_currency/models/sale_order.py
```diff
@@ -23,11 +23,21 @@ class SaleOrder(models.Model):
...
+                order.amount_total_curr = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
-            order.amount_total_curr = to_amount
+                # Convert order amount to company currency
+                conversion_date = (
+                    order.date_order and order.date_order.date()
+                    or fields.Date.context_today(self)
+                )
+                order.amount_total_curr = order.currency_id._convert(
```
rousseldenis, line 38: @straga Why writing twice the field ?

straga: @rousseldenis I don't see any duplication of the amount_total_curr field—it is only defined once. Perhaps you meant something else? Could you clarify where exactly
  you see the duplication?

rousseldenis: You're right, didn'st see the `else` :sweat_smile:

sale_company_currency/models/sale_order.py
```diff
@@ -23,11 +23,21 @@ class SaleOrder(models.Model):
...
 
-    @api.depends("amount_total", "currency_rate")
+    @api.depends("amount_total", "currency_id", "company_id", "date_order")
     def _compute_amount_company(self):
         for order in self:
-            if order.currency_id.id == order.company_id.currency_id.id:
-                to_amount = order.amount_total
+            # order.currency_id == order.company_id.currency_id: 
+            if order.currency_id == order.company_id.currency_id:
+                order.amount_total_curr = order.amount_total
```
alexey-pelykh, line 31: This comment just restates the `if` condition below and has a trailing space. Please remove it -- the code is self-explanatory.

sale_company_currency/models/sale_order.py
```diff
@@ -23,11 +23,21 @@ class SaleOrder(models.Model):
...
             else:
-                to_amount = order.amount_total * order.currency_rate
-            order.amount_total_curr = to_amount
+                # Convert order amount to company currency
+                conversion_date = (
+                    order.date_order and order.date_order.date()
+                    or fields.Date.context_today(self)
```
alexey-pelykh, line 36: Inside the `for order in self` loop, `self` refers to the full recordset, not the current record. While `context_today` typically works on any record (it reads timezone from context), using `order` instead of `self` would be more idiomatic:
```python
or fields.Date.context_today(order)
```

alexey-pelykh (changes requested):
Thanks for the PR @straga -- switching from manual `amount_total * currency_rate` to `currency_id._convert()` is the correct Odoo-standard approach for currency conversion. It properly handles rounding rules, date-specific exchange rates, and multi-company scenarios.

However, there are a few items that should be addressed before merge:

---

### Must fix

**1. Missing version bump in `__manifest__.py`**

Per OCA guidelines, bugfixes require a patch-level version bump. The version should be updated from `18.0.1.0.0` to `18.0.1.0.1`.

---

### Should fix

**2. Redundant/leftover comment (line 31 of `models/sale_order.py`)**

```python
# order.currency_id == order.company_id.currency_id: 
```

This comment just restates the `if` condition on the next line and has a trailing space. It should be removed -- the code is self-explanatory.

**3. `fields.Date.context_today(self)` should be `fields.Date.context_today(order)`**

Inside the `for order in self` loop, `self` is the full recordset. While this typically works (context_today reads timezone from context, not from the record), using `order` is more idiomatic and consistent. This applies to both the model code and the test code.

---

### Suggestion (non-blocking)

**4. Tautological test pattern**

Both `test_02` and `test_03` compute the expected value using the same `currency_id._convert()` call as the implementation itself. This verifies that the code uses `_convert` but doesn't validate correctness against a known expected value.

A stronger test would set up a known exchange rate (e.g., 1 EUR = 1.5 USD) and assert a concrete numeric result:

```python
# Example: set a known rate
self.env['res.currency.rate'].create({
    'currency_id': self.currency_eur.id,
    'rate': 0.85,
    'name': conversion_date,
    'company_id': self.company.id,
})
# Then assert against a calculated expected value
expected = round(self.sale_order.amount_total / 0.85, 2)
self.assertAlmostEqual(self.sale_order.amount_total_curr, expected, places=2)
```

This would catch regressions more effectively. However, the current tests are still an improvement over the previous ones, so this is non-blocking.

---

Overall the fix is sound and well-motivated. Please address items 1-3 and this should be ready for another round of review.

alexey-pelykh (approved):
Reclassifying this as an approval rather than a change request — apologies for the over-strict verdict. The fix is correct (currency_id._convert() is the right approach). My only "must-fix" was the __manifest__.py version bump, but OCA's oca-github-bot handles that at merge, so it shouldn't gate the PR. The rest — the leftover comment on line 31, context_today(order), and a stronger test with a known rate — are non-blocking nits.

rousseldenis (approved):

rousseldenis:
@straga Could you fix pre-commit ?

rousseldenis:
Closing this as no answer. Please feel free to reopen it if needed

PR #3752, opened 2025-06-10
[18.0][MIG] sale_order_lot_generator

hparfr:


sale_order_lot_generator/tests/test_sale_order_lot_generator.py
```diff
@@ -0,0 +1,67 @@
...
+                "order_id": self.order1.id,
+                "product_id": self.prd_acoustic.id,
+                "product_uom_qty": 1,
+            }
+        )
+        lot_number = "%s-%03d" % (self.order1.name, 3)
+        self.assertEqual(self.sol3.lot_id.name, lot_number)
+        for line in self.order1.picking_ids.move_line_ids:
+            if line.product_id.id == self.prd_flipover.id:
+                self.assertEqual(line.lot_id, self.sol1.lot_id)
```
rousseldenis: @hparfr Could you check why some tests lines are not accessed?

hparfr: ... so tests are not covered by tests !

hparfr: ... and that's good. Because,  I didn't pay enough attention. There is no move_lines_ids at this time. So the code inside this for loop is never executing.

rousseldenis: Of course. So, could you check why those lines are necessary ? :sweat_smile:

hparfr: Thanks for this good catch @rousseldenis 
I've just added a fix.

sale_order_lot_generator/tests/test_sale_order_lot_generator.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+        lot_number = "%s-%03d" % (self.order1.name, 3)
+        self.assertEqual(self.sol3.lot_id.name, lot_number)
+        for line in self.order1.picking_ids.move_ids:
+            if line.product_id.id == self.prd_flipover.id:
+                self.assertEqual(line.restrict_lot_id, self.sol1.lot_id)
+            if line.product_id.id == self.prd_desk.id:
+                self.assertEqual(line.restrict_lot_id, self.sol2.lot_id)
+            if line.product_id.id == self.prd_acoustic.id:
+                self.assertEqual(line.restrict_lot_id, self.sol3.lot_id)
```
alexey-pelykh: This `for` loop over `self.order1.picking_ids.move_ids` never executes because there are no stock moves with quants for freshly created lots (as discussed in earlier review comments). These dead assertions (lines 61-67) should be removed entirely -- they provide false confidence that stock move lot assignment is being tested when it is not.

The latest commit message says "fix tests" but the code still contains this unreachable block.

sale_order_lot_generator/models/sale_order_line.py
```diff
@@ -0,0 +1,41 @@
...
+    _inherit = "sale.order.line"
+
+    def _prepare_vals_lot_number(self, index_lot):
+        # Prepare values before creating a lot number
+        self.ensure_one()
+        lot_number = "%s-%03d" % (self.order_id.name, index_lot)
+        return {
+            "name": lot_number,
+            "product_id": self.product_id.id,
+            "product_qty": self.product_uom_qty,
```
alexey-pelykh: `product_qty` on `stock.lot` is a computed field (derived from quants) in Odoo 18, not a stored/writable field. Passing it in `create()` vals will be silently ignored or could cause issues. This key should be removed from the vals dict.

sale_order_lot_generator/models/product_template.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    auto_generate_prodlot = fields.Boolean(
+        string="Auto Generate Lot",
+        help="Forces to specifiy a Serial Number for all "
```
alexey-pelykh: Minor: typo `"specifiy"` should be `"specify"`.

alexey-pelykh (changes requested):
## Review: [18.0][MIG] sale_order_lot_generator

Thanks for migrating this module to 18.0, @hparfr. The module is generally well-structured and the core logic is clean. However, there are a few issues that need to be addressed before this can be merged.

### Issues requiring changes

#### 1. Test dead code: `restrict_lot_id` assertions never execute

As already discussed in previous comments with @rousseldenis, the `for` loop at the end of the test iterating over `self.order1.picking_ids.move_ids` never executes because there are no stock moves with quants for freshly created lots. The latest commit message says "fix tests" but the code on the branch still contains these dead assertions (lines 61-67 of the test file).

These lines should be removed entirely since they provide false confidence -- they appear to test stock move lot assignment but never actually run.

#### 2. `product_qty` field in `_prepare_vals_lot_number`

In `sale_order_line.py`, the `_prepare_vals_lot_number` method sets `product_qty` on the `stock.lot` creation vals:

```python
"product_qty": self.product_uom_qty,
```

`product_qty` on `stock.lot` is a computed field (computed from quants) in Odoo 18, not a stored/writable field. Passing it in `create()` vals will either be silently ignored or could raise an error depending on the Odoo version. This line should be removed.

#### 3. POT file references Odoo Server 17.0

The `sale_order_lot_generator.pot` file header says:
```
Project-Id-Version: Odoo Server 17.0
```

This should be regenerated for 18.0 (typically handled by `oca-gen-addon-readme` / the OCA tooling, but worth noting).

### Suggestions (non-blocking)

- **Test class**: `SingleTransactionCase` is used, which means all tests share a single transaction. This is fine for the current single test method, but if more tests are added in the future, `TransactionCase` would provide better isolation.

- **Test hardcoded XML IDs**: The test relies on demo product XML IDs (`product.product_product_20`, etc.) which may change across Odoo versions. Consider creating test products in `setUpClass` instead for better resilience.

- **View `name` attribute**: The view record `view_template_property_form` is missing the `<field name="name">` element. While Odoo can infer the name, it is best practice in OCA modules to explicitly set it.

- **Minor typo in help string**: `"Forces to specifiy"` should be `"Forces to specify"` (in `product_template.py` and propagated to all `.po`/`.pot` files).

Overall the migration looks straightforward and the business logic is sound. Please address the test dead code and the `product_qty` computed field issue, and this should be good to go.

sale_order_lot_generator/tests/test_sale_order_lot_generator.py
metaminux: As stated by @rousseldenis and @alexey-pelykh `sale_order_lot_generator` is not responsible for propagating lots set on sale order lines to pickings or stock moves.

I removed the loop.

sale_order_lot_generator/tests/test_sale_order_lot_generator.py
metaminux: The block is reachable as stock moves exist, but sale_order_lot_generator is only responsible of the lot generation on sale order lines, so this part of test is not relevant here.

I removed the entire block

sale_order_lot_generator/models/sale_order_line.py
metaminux: OK, done ✅

sale_order_lot_generator/models/product_template.py
metaminux: OK, done ✅

metaminux:
Hello @hparfr @rousseldenis @alexey-pelykh @mourad-ehm @Kev-Roche 

Can you review this PR again now that all comments have been treated and that all tests are green ?
(At least those for this module, `sale_partner_primeship` seems broken due to timezone issue...)

Regards

rousseldenis (approved):

sale_order_lot_generator/tests/test_sale_order_lot_generator.py
hparfr: >The block is reachable as stock moves exist, but sale_order_lot_generator is only responsible of the lot generation on sale order lines, so this part of test is not relevant here.

I don't get why you remove this test. The goal of this module is to have lots on stock moves.

metaminux: Nope, the goal of this module is to generate lots on each sale order line as stated in the DESCRIPTION.md

It depends on `sale_order_lot_selection` which is responsible to set the `restrict_lot_id` on stock moves.

hparfr:
@alexey-pelykh your request for change is blocking this pr

alexey-pelykh:
Will look into it a bit later today

alexey-pelykh (approved):
Re-reviewing after the updates — all blocking issues from my previous review are resolved:

1. **Test dead code** — ✅ Dead assertions in the `for` loop over `move_ids` have been removed.
2. **`product_qty` in lot creation vals** — ✅ Removed. `_prepare_vals_lot_number` now only passes `name`, `product_id`, `company_id`.
3. **POT version** — Still references `Odoo Server 17.0` (minor, can be fixed at merge or by maintainer tooling).

Non-blocking suggestions also addressed:
- Test class switched to `TransactionCase` ✅
- View `name` field added ✅
- Typo "specifiy" → "specify" ✅

CI note: The `test with Odoo` failure is from `sale_partner_primeship` (timezone-related test failures), not from `sale_order_lot_generator`. Tests for this module pass with both OCB and Odoo.

BhaveshHeliconia (approved):
Functional review LGTM!

PR #3756, opened 2025-06-12
[ADD] sale_order_line_discount_preserve: New module sale_order_line_discount_preserve

miquelpascual:
This module preserves the discount set on a sales order line, even if values on that line are changed (for example, a product attribute or the quantity).

cc https://github.com/APSL HT02127
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


peluko00 (approved):
LGTM! Tested in runboat and code review

lbarry-apsl (approved):
Code and Runboat reviewed. LGTM

dreispt:
Sorry, tests are failing.

mpascuall:
> Sorry, tests are failing.

Hi, thanks for your comment! The tests are passing successfully on my local environment. Do you have any idea of how could i fix that? Many thanks in advance!

alexey-pelykh (changes requested):
## Review: sale_order_line_discount_preserve

Thank you for the contribution. The idea of preserving manually-set discounts is genuinely useful. However, the current implementation has several critical issues that need to be addressed before this can be merged.

### Critical Issues

**1. Inverted logic condition (breaks discount computation entirely)**

The base Odoo 18 `_compute_discount` method has:
```python
if not (line.order_id.pricelist_id and discount_enabled):
    continue
```
This skips pricelist discount computation when there is no pricelist OR discounts are disabled.

The override changes this to:
```python
if not (line.order_id.pricelist_id and not discount_enabled):
    continue
```
The `not discount_enabled` inverts the condition: this now skips discount computation **when discounts ARE enabled**, which is the exact opposite of the intended behavior. This is almost certainly the root cause of the CI test failures.

**2. The override does not actually preserve discounts**

The module's stated purpose is to preserve manually-set discounts when other line values change. However, the `_compute_discount` override completely replaces the base method and simply re-computes the discount from the pricelist every time -- there is no mechanism to detect whether a discount was manually set by the user vs. computed from a pricelist rule, and no logic to skip recomputation in that case.

To actually preserve a manually-set discount, the module would need to either:
- Store whether the discount was manually set (e.g., a boolean field `discount_manual` or similar) and skip recomputation when it is, or
- Compare the current discount against the pricelist-computed discount before overwriting, and retain it if they differ (though this approach has edge cases).

**3. Missing `combo_item_id` handling**

The base Odoo 18 method includes:
```python
if line.combo_item_id:
    line.discount = line._get_linked_line().discount
    continue
```
This block is completely absent from the override. This will break combo item discount computation for anyone using that feature.

### Moderate Issues

**4. Missing `data` key in `__manifest__.py`**

The manifest has no `data` key. While there are no XML/CSV data files in the module, this is a minor point -- just noting it for completeness since OCA modules typically include this key even if empty.

**5. Missing newline at end of file**

Several readme files (`CONFIGURE.md`, `CONTRIBUTORS.md`, `DESCRIPTION.md`, `USAGE.md`) are missing the trailing newline, which will cause `pre-commit` warnings in many configurations.

**6. Test does not actually validate the module's purpose**

The test creates a line with `discount=20.0`, then changes quantity, and asserts discount is still 20.0. However, this test would also pass on vanilla Odoo if there is no pricelist rule that would change the discount on quantity change. The test needs a pricelist with a quantity-based discount rule so that the base Odoo behavior would change the discount, and the module's override should prevent that change. Without such a setup, the test does not prove the module works.

### CI Status

Both `test with OCB` and `test with Odoo` are **failing**. This aligns with the inverted logic condition identified above.

### Suggestions

The core approach of overriding `_compute_discount` to preserve manual discounts is architecturally reasonable, but the implementation needs a fundamentally different strategy:

1. Fix the inverted condition (`not discount_enabled` -> `discount_enabled`)
2. Add the missing `combo_item_id` handling
3. Add actual discount preservation logic (e.g., a `discount_manually_set` boolean field that gets set via an `onchange` or `write` override, and skip `_compute_discount` when it is `True`)
4. Write tests that set up a pricelist with quantity breaks so that changing quantity would normally alter the discount, then verify the manually-set discount is preserved
5. Fix missing trailing newlines in readme files

I would be happy to re-review once these issues are addressed.

sale_order_line_discount_preserve/models/sale_order_line.py
```diff
@@ -0,0 +1,33 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.depends("product_id", "product_uom", "product_uom_qty")
+    def _compute_discount(self):
```
rousseldenis, line 8: @mpascuall You should find another way to do this as this is quite insane to copy the whole Odoo core method here.

IMHO, you should add a new parameter for this prevent feature.

Then, you should rewrite this by:

- keeping the initial discount values
- calling `super()`
- then, reassign the initial value if changed.

lmignon (changes requested):
This module will break all the installations where others addons override the `_compute_discount` method. Moreover, it redefine the the whole _compute_discount method on sale_order_line without calling super. The problem if you want to block the discount is that you must ensure that your override of the `_compute_discount` method is always at the top of the call stack whatever the installed addons in your odoo instance. By default it's not possible to achieve such a behavior without explicitly declaring all addons overriding the _compute_discount method as a dependency of your addons....
Another approach also exists but is more subject to controversy discussion.  To ensure that your method is always at the top of the call stack, you could also implement a `_register_hook() ` method into the `sale.order.line` model to patch the `_compute_discount` method. This method is indeed called at the end of the registry initialization. That means that if you patch the method from `self.env['sale.order.line'].__class__`  you path the method at the top of the call stack whatever the installed addons.... This is a kind of hack ....

rousseldenis:
Closing this as no answer. Please feel free to reopen it if needed

PR #3757, opened 2025-06-13
[16.0][IMP] sale_order_line_cancel: Allow to decrease product_uom_qty on line cancel

mt-software-de:
This PR adds a new setting field to allow to decrease the order line qty `product_uom_qty` instead of increasing `product_qty_canceled`. 

cc @jbaudoux @lmignon @rousseldenis

### DIFF
--- a/sale_order_line_cancel/models/res_company.py
+++ b/sale_order_line_cancel/models/res_company.py
@@ -1,0 +1,13 @@
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    on_sale_line_cancel_decrease_line_qty = fields.Boolean(
+        "On sale order line cancel decrease line quantity",
+        help="On canceling the remaining qty to deliver of an order line "
+        "it decreases the initial quantity",
+    )
--- a/sale_order_line_cancel/models/sale_order_line.py
+++ b/sale_order_line_cancel/models/sale_order_line.py
@@ -1,6 +1,7 @@
 # Copyright 2018 Okia SPRL
 # Copyright 2018 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # Copyright 2020 ACSONE SA/NV
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
@@ -41,7 +42,7 @@ class SaleOrderLine(models.Model):
     @api.depends("qty_to_deliver", "product_qty_canceled")
     def _compute_product_qty_remains_to_deliver(self):
         for line in self:
-            qty_remaining = line.qty_to_deliver - line.product_qty_canceled
+            qty_remaining = max(0, line.qty_to_deliver - line.product_qty_canceled)
             line.product_qty_remains_to_deliver = qty_remaining
 
     def _get_moves_to_cancel(self):
@@ -57,7 +58,14 @@ class SaleOrderLine(models.Model):
         for line in self:
             if line._get_moves_to_cancel():
                 continue
-            line.product_qty_canceled = line.qty_to_deliver
+            qty_to_deliver = line.qty_to_deliver
+            vals = {"product_qty_canceled": qty_to_deliver}
+            if (
+                line.state == "sale"
+                and line.company_id.on_sale_line_cancel_decrease_line_qty
+            ):
+                vals["product_uom_qty"] = line.qty_delivered
+            line.write(vals)
 
     def cancel_remaining_qty(self):
         lines = self.filtered(lambda l: l.can_cancel_remaining_qty)
--- a/sale_order_line_cancel/tests/test_sale_order_line_cancel.py
+++ b/sale_order_line_cancel/tests/test_sale_order_line_cancel.py
@@ -1,4 +1,5 @@
 # Copyright 2023 ACSONE SA/NV
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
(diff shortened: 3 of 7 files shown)


sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -57,7 +58,15 @@ def _update_qty_canceled(self):
         for line in self:
             if line._get_moves_to_cancel():
                 continue
-            line.product_qty_canceled = line.qty_to_deliver
+            if (
+                line.state == "sale"
+                and line.company_id.on_sale_line_cancel_decrease_line_qty
+            ):
+                line.product_uom_qty = max(
```
lmignon: @mt-software-de IFAIK, if you update the product_uom_qty on a confirmed SO, this will trigger procurement rules with a negative qty for the product....

mt-software-de: It shouldn't because the outgoing qty is the same AS the product_uom_qty https://github.com/odoo/odoo/blob/16.0/addons%2Fsale_stock%2Fmodels%2Fsale_order_line.py#L329

sale_order_line_cancel/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                        class="col-12 col-lg-6 o_setting_box"
+                        id="line_cancel_decrease_qty_configuration"
+                    >
+                        <div class="o_setting_left_pane">
+                            <field name="on_sale_line_cancel_decrease_line_qty" />
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="on_sale_line_cancel_decrease_line_qty" />
+                            <div class="text-muted">
+                                On cancelling the remaining qty to deliver of a Order Line it decrease the initial Quantity
```
jbaudoux:
```suggestion
                                On canceling the remaining qty to deliver of an order line, it decreases the initial quantity
```

sale_order_line_cancel/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    on_sale_line_cancel_decrease_line_qty = fields.Boolean(
+        "On sale order line cancel decrease line quantity",
+        help="On cancelling the remaining qty to deliver of a Order Line "
```
jbaudoux:
```suggestion
        help="On canceling the remaining qty to deliver of an order line, "
```

sale_order_line_cancel/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+        "it decrease the initial Quantity",
```
jbaudoux:
```suggestion
        "it decreases the initial quantity",
```

jbaudoux (commented):
What is the use case ?

The idea was to keep the ordered quantity and show how much is delivered and how much is canceled.
Changing the ordered quantity will make the paid amount bigger than the order amount if you have a payment transaction attached to the SO.

mt-software-de:
> What is the use case ?
> 
> The idea was to keep the ordered quantity and show how much is delivered and how much is canceled. Changing the ordered quantity will make the paid amount bigger than the order amount if you have a payment transaction attached to the SO.

The purpose was to achieve a change of the `order amount` to be able to send out an updated version of the confirmation. Maybe it would be nice to still track which amount was canceled.

mt-software-de:
> > What is the use case ?
> > The idea was to keep the ordered quantity and show how much is delivered and how much is canceled. Changing the ordered quantity will make the paid amount bigger than the order amount if you have a payment transaction attached to the SO.
> 
> The purpose was to achieve a change of the `order amount` to be able to send out an updated version of the confirmation. Maybe it would be nice to still track which amount was canceled.

I now extended it, the canceled qty is still saved and the line's quantity will be decreased.

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -57,7 +58,16 @@ def _update_qty_canceled(self):
...
-            line.product_qty_canceled = line.qty_to_deliver
+            qty_to_deliver = line.qty_to_deliver
+            vals = {"product_qty_canceled": qty_to_deliver}
+            if (
+                line.state == "sale"
+                and line.company_id.on_sale_line_cancel_decrease_line_qty
+            ):
+                vals["product_uom_qty"] = max(
+                    line.qty_delivered, line.product_uom_qty - qty_to_deliver
+                )
```
jbaudoux:
```suggestion
                )
                line = line.with_context(skip_procurement=True)
```

sale_order_line_cancel/tests/test_sale_order_line_cancel.py
```diff
@@ -125,3 +126,25 @@ def test_draft_sale_order_with_picking_cancel(self):
...
+        ship.move_ids.move_line_ids.qty_done = 4
+        ship.with_context(cancel_backorder=False)._action_done()
+        self.wiz.with_context(
+            active_id=line.id, active_model="sale.order.line"
+        ).cancel_remaining_qty()
+        self.assertEqual(line.product_qty_canceled, 6)
+        self.assertEqual(line.product_qty_remains_to_deliver, 0)
+        self.assertEqual(line.qty_to_deliver, 0)
+        self.assertEqual(line.qty_delivered, 4)
+        self.assertEqual(line.product_uom_qty, 4)
```
jbaudoux, line 168: you need to check you don't have a return move as you decreased the SO line quantity

jbaudoux (changes requested):
I think you are generating an unwanted return move

sale_order_line_cancel/models/sale_order_line.py
jbaudoux: it's not mrp safe, you still need to skip procurement

mt-software-de: Why?
Scenario 1: Nothing is done. Sale line qty delivered = 0
All moves will be cancelled
product_uom_qty will be set to 0 
https://github.com/odoo/odoo/blob/16.0/addons/sale_mrp/models/sale_order_line.py#L115 will return 0
https://github.com/odoo/odoo/blob/a800803f8108d81548919e494e35a1679beb4622/addons/sale_stock/models/sale_order_line.py#L329 is true nothing will happen

Scenario 2: Partial done just a component is done qty delivered = 0
All not done moves will be canceled
product_uom_qty will be still set to 0 because qty_delivered is also still 0
https://github.com/odoo/odoo/blob/16.0/addons/sale_mrp/models/sale_order_line.py#L115 will return 0
https://github.com/odoo/odoo/blob/a800803f8108d81548919e494e35a1679beb4622/addons/sale_stock/models/sale_order_line.py#L329 is true nothing will happen

Scenario 3: Partially done both components partially done qty delivered = 1 (out of 2)
All not done moves will be canceled
product_uom_qty will be set to the delivered qty 1
https://github.com/odoo/odoo/blob/16.0/addons/sale_mrp/models/sale_order_line.py#L115 will return 1
https://github.com/odoo/odoo/blob/a800803f8108d81548919e494e35a1679beb4622/addons/sale_stock/models/sale_order_line.py#L329 is still true nothing will happen

jbaudoux: So it's always working because all moves are already canceled. Then the procurement quantity is always computed as 0.

jbaudoux (approved):
Would still be good to extend the test to check there are no return move generated

mt-software-de:
@lmignon Could you update your review, this would be now ready for merging.

lmignon (approved):

raumschmiede-joshuaL (approved):

rousseldenis (approved):

PR #3760, opened 2025-06-16
[18.0][ADD] sale_order_add_byproduct

bosd:
This module is designed for contract manufacturing scenarios where the
client provides raw materials or a base product for processing. Unlike
standard manufacturing where the output quantity of finished goods is
precisely known, this use case involves a process where the exact yield
of **by-products** is uncertain until after the manufacturing operation
is complete.

Scenario
--------

A contract manufacturer processes materials supplied by a client. During
this process, in addition to any primary output, a variable quantity of
by-products is inevitably generated. The client is obligated to purchase
these by-products from the contract manufacturer.


sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        Adds produced by-products to the related Sale Order.
+        If a Sale Order is found, it will either create new order lines
+        or update existing ones for the by-products.
+        """
+        self.ensure_one()
+        sale_order = self.procurement_group_id.sale_id
+        if not sale_order:
+            # Try to find the sale order by origin
+            if self.origin:
```
rousseldenis, line 38: Why doing such queries as you have this direct field?

https://github.com/odoo/odoo/blob/18.0/addons/sale_mrp/models/mrp_production.py#L14C5-L14C17

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                self.origin,
+                self.procurement_group_id,
+            )
+            return
+
+        for byproduct_move in self.move_byproduct_ids:
+            # In Odoo 18, the quantity field might be named differently
+            # Use quantity_done if available, otherwise fallback to quantity
+            # If neither is available, use the move line quantities
+            if (
```
rousseldenis: This should be removed as in 18.0, `quantity_done` field does not exist anymore.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                hasattr(byproduct_move, "quantity_done")
+                and byproduct_move.quantity_done
+            ):
+                quantity_produced = byproduct_move.quantity_done
+            elif hasattr(byproduct_move, "quantity") and byproduct_move.quantity:
```
rousseldenis: stock move has `quantity` field, so the `else` case is not relevant anymore.

rousseldenis: In fact, if no move lines, the quantity done == 0.0. But, if you look for a done quantity, you should take the move lines with `picked` field as `True`

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                    _logger.warning(
+                        "MRP Production %s: By-product move %s has no move lines, "
+                        "cannot determine produced quantity.",
+                        self.name,
+                        byproduct_move.name,
+                    )
+                    continue
+                quantity_produced = sum(byproduct_move.move_line_ids.mapped("quantity"))
+
+            if not byproduct_move.product_id.sale_ok:
```
rousseldenis, line 65: This has to be put at the beginning of the loop to avoid too much conditions evaluations.

rousseldenis: Moreover, I would have put this in a separate function like `_get_byproduct_moves_to_add()` or somethng like that.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                continue
+
+            if quantity_produced > 0:
```
rousseldenis: Use `float_compare()`

sale_order_add_byproduct/models/res_config_settings.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    byproduct_note_template = fields.Char(
+        string="By-product Note Template",
+        help="Template for the note on sale order lines created for by-products. "
+        "You can use placeholders like {product_name} and {mo_name}.",
```
rousseldenis, line 13: Use `config_parameter="sale_order_add_byproduct.byproduct_note_template"` to avoid `get` and `set` methods

rousseldenis: I'm wondering if a per company parameter shoulde be better or not

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                        "MRP Production %s: Updated Sale Order Line for %s "
+                        "(Qty: %s) on SO %s.",
+                        self.name,
+                        product_to_add.name,
+                        quantity_produced,
+                        sale_order.name,
+                    )
+                else:
+                    # Get the template from the config parameter
+                    config_settings = self.env["res.config.settings"].create({})
```
rousseldenis: Don't do that. Move the `get_` method (in res.config.settings - see my other comment) to an api.model one here.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                    note_template = config_settings.get_byproduct_note_template()
+                    if note_template:
+                        note = note_template.format(
+                            product_name=product_to_add.name, mo_name=self.name
+                        )
+                    else:
+                        note = product_to_add.name
+                    self.env["sale.order.line"].create(
```
rousseldenis: You should build create values (a list) in the loop and after the loop, just call one `create()` with that list.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,131 @@
...
+                            "order_id": sale_order.id,
+                            "product_id": product_to_add.id,
+                            "product_uom_qty": quantity_produced,
+                            "product_uom": product_to_add.uom_id.id,
+                            "price_unit": product_to_add.lst_price,
+                            "name": note,
+                            "is_mrp_byproduct_line": True,
+                        }
+                    )
+                    _logger.info(
```
rousseldenis: Are all these info lines should stay as this will pollute the log file? I would have choose debug instead.

sale_order_add_byproduct/models/sale_order_line.py
```diff
@@ -0,0 +1,45 @@
...
+        return super().write(values)
+
+    def create(self, values):
+        # Create the record using standard behavior
+        record = super().create(values)
+        
+        # For regular lines on confirmed orders, ensure procurement is triggered
+        # This ensures the procurement group is properly maintained
+        if record.order_id and record.order_id.state == 'sale':
+            # Only trigger for regular lines, not by-product lines
```
rousseldenis: I'm not sure about the code here after as this can lead to unwanted behaviours.

rousseldenis (changes requested):

bosd:
Thanks for the review. 
CI is still red, but that seems to come from a different module.

Dranyel-Bosd (approved):
LGTM

sale_order_add_byproduct/models/sale_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+        "new procurements.",
+    )
+
+    def _action_launch_stock_rule(self, previous_product_uom_qty=False):
+        # Filter out by-product lines
+        lines_not_to_procure = self.filtered(lambda line: line.is_mrp_byproduct_line)
+
+        # Call super only on lines that should be procured
+        if lines_not_to_procure:
+            return True
```
alexey-pelykh, line 21: **Bug (critical):** This logic is inverted for mixed recordsets. If `self` contains 5 lines and 1 is a byproduct, this returns `True` immediately and **skips procurement for the other 4 lines**.

The correct pattern is to exclude byproduct lines and call `super()` on the remaining lines:

```python
def _action_launch_stock_rule(self, previous_product_uom_qty=False):
    lines_to_procure = self.filtered(lambda line: not line.is_mrp_byproduct_line)
    if not lines_to_procure:
        return True
    return super(SaleOrderLine, lines_to_procure)._action_launch_stock_rule(
        previous_product_uom_qty=previous_product_uom_qty
    )
```

sale_order_add_byproduct/models/sale_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        # In case of line addition on a confirmed SO, we need to trigger the
+        # procurement group creation.
+        for line in lines:
+            if line.order_id.state == "sale" and not line.is_mrp_byproduct_line:
+                line._action_launch_stock_rule()
+        return lines
```
alexey-pelykh, line 35: **Bug:** Overriding `create` to manually call `_action_launch_stock_rule()` for non-byproduct lines is risky. Odoo's standard `sale.order.line` already triggers procurement on line creation for confirmed SOs via its own mechanism. This override may cause **double procurement** (two MOs created for one SO line).

I believe the `_action_launch_stock_rule` override above (once fixed) is sufficient to prevent byproduct lines from triggering procurement. This entire `create` override should likely be removed.

If it IS needed, please add a test that demonstrates the scenario it solves -- currently no test covers the `create`-triggered procurement path specifically.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,200 @@
...
+            if self.origin:
+                # First try direct search by origin
+                sale_order = self.env["sale.order"].search(
+                    [("name", "=", self.origin)], limit=1
+                )
+                # If not found and origin starts with "S", try with origin
+                if not sale_order and self.origin.startswith("S"):
+                    sale_order = self.env["sale.order"].search(
+                        [("name", "=", self.origin)], limit=1
+                    )
```
alexey-pelykh, line 47: **Dead code:** This `if` branch (`self.origin.startswith("S")`) executes the exact same search as lines 38-40 above:
```python
sale_order = self.env["sale.order"].search(
    [("name", "=", self.origin)], limit=1
)
```
This duplicate search can never find anything the first one missed. Remove this branch entirely.

sale_order_add_byproduct/models/mrp_production.py
```diff
@@ -0,0 +1,200 @@
...
+                            .sudo()
+                            .get_param(
+                                "sale_order_add_byproduct.byproduct_note_template",
+                                "{product_name} (By-product from MO: {mo_name})",
+                            )
+                        )
+                    if config_param:
+                        note = config_param.format(
+                            product_name=product_to_add.name, mo_name=self.name
+                        )
```
alexey-pelykh, line 131: **Security concern:** Using `str.format()` with a user-controlled template string is a known Python injection vector. A malicious admin could craft a template like `{product_name.__class__.__mro__}` to leak internal Python object information.

Consider using `safe_substitute` from `string.Template`, or restrict the format to known placeholders via a simple `.replace()` approach:
```python
note = config_param.replace("{product_name}", product_to_add.name).replace("{mo_name}", self.name)
```

sale_order_add_byproduct/models/sale_order_line.py
```diff
@@ -0,0 +1,35 @@
+from odoo import fields, models
```
alexey-pelykh, line 1: **OCA convention:** Missing license header. All Python files in OCA modules must have the copyright and license header:
```python
# Copyright 2025 OBS Solutions BV
```

sale_order_add_byproduct/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "version": "18.0.2.0.0",
+    "category": "Sales Management",
+    "summary": "Add produced by-products from Manufacturing Order "
+    "to the originating Sales Order as invoiceable lines.",
+    "author": "Odoo Community Association (OCA), OBS Solutions BV",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["mrp", "sale_management", "sale_stock"],
+    "data": ["data/config_parameter_data.xml", "views/res_config_settings_views.xml"],
+    # "demo": ["demo/demo_data.xml"],
```
alexey-pelykh, line 12: **Cleanup:** Remove the commented-out `"demo"` line. OCA guidelines discourage committed commented-out code.

sale_order_add_byproduct/models/res_config_settings.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    byproduct_note_template = fields.Char(
+        string="By-product Note Template",
+        help="Template for the note on sale order lines created for by-products. "
```
alexey-pelykh, line 12: As `rousseldenis` already noted: since you now have a `res.company` field, the settings field should use `related="company_id.byproduct_note_template"` rather than manual `_compute` / `_inverse` methods. This is the standard OCA pattern for company-level settings and removes ~15 lines of code.

alexey-pelykh (changes requested):
## Review: REQUEST_CHANGES

Thank you for the contribution -- the business case is well-documented and the test coverage is solid. However, there are several correctness issues and OCA convention gaps that should be addressed before merge.

### Critical Findings

| # | File | Severity | Finding |
|---|------|----------|---------|
| 1 | `sale_order_line.py` | Bug | `_action_launch_stock_rule` override breaks mixed recordsets -- returns `True` for the entire batch when ANY line is a byproduct, silently skipping procurement for legitimate lines |
| 2 | `sale_order_line.py` | Bug | `create` override may double-trigger procurement for non-byproduct lines added to confirmed SOs (Odoo already handles this) |
| 3 | `mrp_production.py` | Dead code | Duplicate SO search at L42-45 (`origin.startswith("S")` branch does the exact same query as L38-40) |
| 4 | `mrp_production.py` | Security | Note template uses Python `str.format()` with user-controlled template string from `ir.config_parameter` -- could expose object attributes via `{0.__class__}` style injection |
| 5 | `sale_order_line.py` | OCA | Missing license header |
| 6 | `res_config_settings.py` | Convention | Reviewer `rousseldenis` already flagged: prefer `config_parameter=` attribute or a proper company-level field with `related=` instead of manual compute/inverse |

### Suggestions

| # | File | Finding |
|---|------|---------|
| 7 | `__manifest__.py` | Remove commented-out `"demo"` key |
| 8 | Tests | Excessive `invalidate_all()` + re-browse suggests misunderstanding of ORM cache -- consider simplifying |
| 9 | `mrp_production.py` | `bom_param` argument in helper is accepted but never used -- remove or implement |

rousseldenis:
@bosd Could you improve your commit messages and squash some of them if needed. Thanks

bosd:
@rousseldenis Done, squashed the commits.

flotho:
Hi @bosd ,
Any chance to rebase ?

bosd:
> Hi @bosd ,\nAny chance to rebase ?

Im on holiday.. will try..

/ocabot rebase

alexey-pelykh (approved):
Thanks for the thorough rework, @bosd — I went back through the current code and everything from my earlier review is addressed:

- `_action_launch_stock_rule` now excludes byproduct lines and calls `super()` on the remainder (`lines_to_procure`), so mixed recordsets procure correctly.
- The `create` override is gone — byproduct lines are kept out of procurement purely via the flag + stock-rule override.
- SO resolution goes through `sale_line_id.order_id` / `procurement_group_id.sale_id` with a single `origin` fallback (`_get_sale_order_for_byproducts`), so the duplicate search is dropped.
- The note template uses `.replace()` on known placeholders instead of `str.format()`.
- License headers are in place, and the settings field is now `related="company_id.byproduct_note_template"` on a real `res.company` field.

Nicely factored too — `_get_byproduct_moves_to_add` (with `picked` move lines + `float_compare`), the batched `create()`, and `_logger.debug`. Test coverage is solid.

Non-blocking: codecov is still red on the patch — worth a glance, but the functional suite passes and I don't treat it as a blocker.

LGTM.

rousseldenis (approved):

bosd:
@rousseldenis Can you check the bot?

PR #3763, opened 2025-06-19
[17.0][MIG] sale_resource_booking: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT51945


sale_resource_booking/models/resource_booking.py
```diff
@@ -41,11 +44,14 @@ def _compute_state(self):
...
         result = super()._compute_state()
-        # False means "no sale order"
-        confirmable_states = {False, "sale", "done"}
         for one in self:
             # Only affect confirmed bookings related to unconfirmed quotations
-            if one.state != "confirmed" or one.sale_order_state in confirmable_states:
+            if (
+                one.state != "confirmed"
+                or not one.sale_order_line_id
+                or one.sale_order_line_id.order_id.locked
```
pedrobaeza: You can only lock being on `sale` state, so no need of this extra condition.

pedrobaeza (approved):

sale_resource_booking/models/resource_booking.py
victoralmau: I think it is important to keep this condition to make it similar to v15. If it is `locked` it will be in the exited state, but it is also possible that it is NOT locked (because that feature is not used) and the state is `sale`

pedrobaeza: so no need of the condition, because the next `or` will match.

sale_resource_booking/wizards/resource_booking_sale.py
```diff
@@ -0,0 +1,46 @@
...
+
+    type_id = fields.Many2one(
+        "resource.booking.type",
+        string="Booking type",
+        index=True,
+        ondelete="cascade",
+        required=True,
+    )
+    partner_id = fields.Many2one(
+        "res.partner", string="Customer", index=True, ondelete="cascade", required=True
```
carlos-lopez-tecnativa: Why add an index to a TransientModel? I think this doesn’t provide any benefit, since users never search on this field in the model.

```suggestion
        "res.partner", string="Customer", ondelete="cascade", required=True
```

sale_resource_booking/wizards/resource_booking_sale.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    product_id = fields.Many2one(
+        "product.product",
+        "Product",
+        domain="[('resource_booking_type_id', '=', type_id)]",
+        index=True,
```
carlos-lopez-tecnativa: The same here
```suggestion
```

sale_resource_booking/wizards/sale_order_booking_confirm.py
```diff
@@ -0,0 +1,42 @@
...
+                    default_partner_ids=[(4, booking.partner_id.id, 0)],
+                )
+            )
+            share = share_f.save()
+            share.action_send_mail()
+
+    def action_noop(self):
+        # At this point, the record was already created with the required
+        # changes; nothing left to do
+        return
```
carlos-lopez-tecnativa: When I confirm a sale order, the record is not reloaded or refreshed, so the user might feel that the order wasn't confirmed.
I think it's necessary to add an action to force the record to refresh.
Please try something like:
```suggestion
        return {"type": "ir.actions.act_window_close"}
```

carlos-lopez-tecnativa (changes requested):

sale_resource_booking/wizards/sale_order_booking_confirm.py
victoralmau: Ok, thanks, actually the right thing to do (in the other method as well) is to use `return {"type": "ir.actions.client", "tag": "reload"}` so that it actually reload and you see the change that the order is confirmed.

carlos-lopez-tecnativa: I suggested using `{"type": "ir.actions.act_window_close"}`, which is used when a wizard is closed and refreshes the current record.
I noticed that the `reload` tag reloads the whole window and causes the breadcrumbs to be lost, but anyway, LGTM.

carlos-lopez-tecnativa (approved):

sale_resource_booking/wizards/sale_order_booking_confirm.py
victoralmau: Previously I had used `{"type": "ir.actions.act_window_close"}` but it does not work properly, the same happened in the  `action_invite()` method.

pedrobaeza: What about the following?: on module `web_refresher`, create a client action with tag `soft_reload` or similar. On this module, make it depend on it, and call that action?

victoralmau: IMO, I think adding web_refresher as a dependency is not a good idea, basically because the use case (refresh) is already covered, and we usually don't like it when new dependencies are added to modules.

rousseldenis (approved):

PR #3765, opened 2025-06-21
[18.0][FIX] sale_order_lot_selection/sale_order_product_availability_inline: Don't use demo data for tests / fix display

rousseldenis:
Fixes:

- [x] sale_order_lot_selection: Don't rely on demo data for tests
- [x] sale_order_product_availability_inline: Since Odoo's change here: https://github.com/odoo/odoo/commit/c84b67e5dca1a74e1a5862e754455944015f4105

### DIFF
--- a/sale_order_product_availability_inline/models/sale.py
+++ b/sale_order_product_availability_inline/models/sale.py
@@ -12,3 +12,8 @@ class SaleOrderLine(models.Model):
         if self.env.context.get("so_product_stock_inline"):
             self = self.with_context(so_product_stock_inline=False)
         return super()._compute_name()
+
+    def _compute_translated_product_name(self):
+        if self.env.context.get("so_product_stock_inline"):
+            self = self.with_context(so_product_stock_inline=False)
+        return super()._compute_translated_product_name()
--- a/sale_order_product_availability_inline/views/sale_views.xml
+++ b/sale_order_product_availability_inline/views/sale_views.xml
@@ -15,6 +15,16 @@
                     "warehouse": parent.warehouse_id}
                 </attribute>
             </xpath>
+
+            <xpath
+                expr="//field[@name='order_line']/list//field[@name='product_template_id']"
+                position="attributes"
+            >
+                <attribute name="context" operation="update">
+                    {"so_product_stock_inline": True,
+                    "warehouse": parent.warehouse_id}
+                </attribute>
+            </xpath>
             <xpath
                 expr="//field[@name='order_line']/form//field[@name='product_id']"
                 position="attributes"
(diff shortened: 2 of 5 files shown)


rousseldenis:
@bosd @astirpe This fixes tests when mrp is installed.

bosd (approved):
Thanks! Let's get this branch :green_circle: again

bosd:
Fix as discovered broken tests in #3667

astirpe:
Thank you @rousseldenis !

astirpe (approved):

rousseldenis:
@ernestotejeda

rousseldenis:
@pedrobaeza As this involves one of yours

rousseldenis:
@bosd @astirpe Could you functionaly review `sale_order_product_availability_inline` too on runboat (that one failed too)?

sale_order_product_availability_inline/models/sale.py
```diff
@@ -12,3 +12,8 @@ def _compute_name(self):
         if self.env.context.get("so_product_stock_inline"):
             self = self.with_context(so_product_stock_inline=False)
         return super()._compute_name()
+
+    def _compute_translated_product_name(self):
+        if self.env.context.get("so_product_stock_inline"):
```
pedrobaeza, line 17: Can you add any code comment here explaining why this?

sale_order_product_availability_inline/views/sale_views.xml
```diff
@@ -15,6 +15,16 @@
...
                 </attribute>
             </xpath>
+
+            <xpath
+                expr="//field[@name='order_line']/list//field[@name='product_template_id']"
+                position="attributes"
+            >
+                <attribute name="context" operation="update">
+                    {"so_product_stock_inline": True,
+                    "warehouse": parent.warehouse_id}
```
pedrobaeza, line 25: Isn't there here other way for not overriding the context and being fragile if Odoo changes this context upstream (or in new versions)?

rousseldenis: It shouldn't in stable version. `update` is quite safe as they wouldn't remove context at all

rousseldenis: Note: I try here to fix tests in 18.0 branch. Maybe enhancements could be done in a new PR

pedrobaeza: Well, we both know that these things happen in stable, but if you don't have any other way to fix it, OK.

pedrobaeza (approved):
/ocabot merge patch

PR #3766, opened 2025-06-23
[18.0][FIX] sale_automatic_workflow: Fix SO domains

mmequignon:
Since 18.0, there's no `done` state anymore on sale orders.
Both former `done` and `sale` states are now `sale`.
Whethere a SO is done or not is represented by the `locked` boolean state.
Applied this on the `sale_automatic_workflow` domains.

### DIFF
--- a/sale_automatic_workflow/data/automatic_workflow_data.xml
+++ b/sale_automatic_workflow/data/automatic_workflow_data.xml
@@ -15,7 +15,7 @@
         <field name="model_id">sale.order</field>
         <field
             name="domain"
-        >[('state','=','sale'),('invoice_status','=','to invoice')]</field>
+        >[('state','=','sale'), ('locked', '=', False), ('invoice_status','=','to invoice')]</field>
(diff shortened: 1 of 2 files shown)


sebalix:
@mmequignon tests are :red_circle: , can you check? :pray:

rousseldenis:
@mmequignon Could you rebase too as the other module error is fixed on 18.0 ?

mmequignon:
@sebalix @rousseldenis This is fixed now, thanks!

rousseldenis (approved):

rousseldenis:
@sebalix Could you review ?

sebalix (approved):

sebalix:
Should we write a migration script for existing installations to update such `noupdate` filters?

rousseldenis:
> Should we write a migration script for existing installations to update such `noupdate` filters?

@sebalix Yes because the domain won't be updated on existing installations.

dreispt (approved):

PR #3767, opened 2025-06-23
[FIX] sale_exception: add a register_hook to prevent other action_con…

lav-adhoc:
…firm being executed if there is an exception rule.

For example, if a coupon was used, a point would be added, and then upon confirming the sale, it would be added again.
The issue is in the sale_loyalty module, where the loyalty program is computed before the sale is confirmed in this [line](https://github.com/odoo/odoo/blob/812d8a5c0117346b7fa40689213a0b0fda61f9ad/addons/sale_loyalty/models/sale_order.py#L136)

This video demonstrates the issue (version 16), because in the version 16 there is the same problem. We decided to apply the fix in version 18, since version 16 is a stable.
https://drive.google.com/file/d/1qeEYlVVqXb8MOJq7BybVe4EaYtiDcIPS/view

Before this PR, if a sale was not confirmed due to an exception, the loyalty program would still be applied.

Although the problem occurs with coupons, it's possible that new problems might also arise with payment integrations or electronic invoicing.

### DIFF
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -54,6 +54,25 @@ class SaleOrder(models.Model):
             return self._popup_exceptions()
         return super().action_confirm()
 
+    def _register_hook(self):
+        ModelClass = self.env.registry["sale.order"]
+
+        original_action_confirm = ModelClass.action_confirm
+
+        def patched_action_confirm(self):
+            if self.detect_exceptions():
+                if not self.env.company.sale_exception_show_popup:
+                    return
+                return self._popup_exceptions()
+            original_func = patched_action_confirm.origin
+            return original_func(self)
+
+        patched_action_confirm.origin = original_action_confirm
+
+        ModelClass.action_confirm = patched_action_confirm
+
+        return super()._register_hook()
+
     def action_draft(self):
         res = super().action_draft()
         orders = self.filtered("ignore_exception")
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -54,6 +54,34 @@ class SaleOrder(models.Model):
         self.detect_exceptions()
         return super().action_confirm()
 
+    def _register_hook(self):
+        # Exceptions must be detected before any other module runs its own
+        # `action_confirm` logic. Standard inheritance cannot grant that:
+        # modules with no dependency between them are ordered by installation
+        # order, so an override that acts before calling `super()` may run
+        # first. `sale_loyalty` is one case: it adds the loyalty points before
+        # the order is confirmed, so the points were granted even when an
+        # exception blocked the confirmation. The same happens with
+        # `sale_order_lot_generator`, and it may happen with payment or
+        # e-invoicing integrations, whose side effects live outside the
+        # transaction and are not undone by a rollback.
+        # Patching the resolved `action_confirm` on the registry class puts the
+        # detection outermost, whatever the MRO is.
+        ModelClass = self.env.registry["sale.order"]
+
+        original_action_confirm = ModelClass.action_confirm
+
+        def patched_action_confirm(self):
+            self.detect_exceptions()
+            original_func = patched_action_confirm.origin
+            return original_func(self)
+
+        patched_action_confirm.origin = original_action_confirm
+
+        ModelClass.action_confirm = patched_action_confirm
+
+        return super()._register_hook()
+
     def action_draft(self):
         res = super().action_draft()
         orders = self.filtered("ignore_exception")
--- a/sale_exception/tests/test_sale_exception.py
+++ b/sale_exception/tests/test_sale_exception.py
@@ -13,6 +13,8 @@ class TestSaleException(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env["sale.order"]._register_hook()
+        cls.env["exception.rule"].search([]).write({"active": False})
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
         cls.default_pl = cls.env["product.pricelist"].create(
             {
--- a/sale_exception/tests/test_sale_exception.py
+++ b/sale_exception/tests/test_sale_exception.py
@@ -29,6 +29,8 @@ class TestSaleException(TransactionCase):
                 cls.env.context, test_base_exception=True, tracking_disable=True
             )
         )
+        cls.env["sale.order"]._register_hook()
+        cls.env["exception.rule"].search([]).write({"active": False})
         cls.default_pl = cls.env["product.pricelist"].create(
             {
                 "name": "Public Pricelist",


lef-adhoc (approved):

rousseldenis:
@lav-adhoc Could you explain again the use case you have and what actions that are not protected by transactions are concerned ?

lav-adhoc:
@rousseldenis 
The problem is the same that in 16
For example, if a coupon was used, a point would be added, and then upon confirming the sale, it would be added again.
The issue is in the sale_loyalty module, where the loyalty program is computed before the sale is confirmed in this [line](https://github.com/odoo/odoo/blob/812d8a5c0117346b7fa40689213a0b0fda61f9ad/addons/sale_loyalty/models/sale_order.py#L136)

This video demonstrates the issue (version 16)
https://drive.google.com/file/d/1qeEYlVVqXb8MOJq7BybVe4EaYtiDcIPS/view

Before this PR, if a sale was not confirmed due to an exception, the loyalty program would still be applied.

Although the problem occurs with coupons, it's possible that new problems might also arise with payment integrations or electronic invoicing.

lav-adhoc:
@rousseldenis 
Since these are modules that do not depend on each other, the action_confirm method will be executed first from the module that was installed earlier.

We don't consider the solution of creating bridge modules to be viable, as it would require developing an additional module for each combination of independent modules that perform a specific action (such as assigning points in sale_loyalty).

We believe this is the best solution, as it decouples the behavior from the installation order of the modules. It would also work for any case, not just loyalty.

Lastly, the tests are failing, but the issue is unrelated to this implementation and comes from other modules.

rousseldenis:
Ok, I have a concern about the method used (like before).

IMHO, in fact, that's because your functional need is: if there are exceptions, keep them but don't add coupons.
That's why I recommend instead a glue module that will add an option on exception rules that say 'don't add coupon in case of this exception'.

Then, in `_get_point_changes()`, don't retrieve items if sale has exceptions that verify the preceding option.

What do you think ?

maq-adhoc:
@rousseldenis , you're right that we could add a mechanism to prevent coupons from being created. Our primary concern, however, lies with the design of the module.

From our perspective, if an exception occurs, everything executed before the super() call should be rolled back. This means if sale_exception halts a method's flow, the Sale Order should remain unmodified. The current behavior is inconsistent with how Odoo typically handles exceptions.

We've explored various possibilities, such as creating a new exception type in the ORM that triggers a rollback and write exception data via another database pointer. However, all these solutions would require a significant refactor of the module.

While @lav-adhoc 's proposal is somewhat of a "hack," it currently appears to be the least invasive way to address this module's issue.

lav-adhoc:
Hi @rousseldenis what do you think about @maq-adhoc 's comment?
We believe this is a good way to solve it without making major changes and ensuring that the problem with coupons does not occur with other methods.

grindtildeath:
@lav-adhoc Thanks for the PR

I stumbled on the same issue while having sale_exception and OdooEE sale_subscription module installed, but I'm not sure if this "hack" is the right solution, because we can't rely on proper MRO between modules having no dependencies between each other.

IMO as @maq-adhoc mentioned, an exception must be raised in the `action_confirm` of this module to rollback anything that could happen in the `action_confirm` from another module. I tried to go this way in https://github.com/OCA/sale-workflow/pull/3885 but there's still a UI issue and the test to fix :disappointed: 

Anyway, I would suggest to move forward and merge this PR for the time being until we can find a better solution, but could you please do the following:
 - Rebase on 18.0 now that https://github.com/OCA/sale-workflow/pull/3837 is merged
 - Add roadmap entry explaining the issue
 - Write a big comment on the register hook explaining the issue

Eventually you could mention my PR #3885 as a possible way to solve this, but as expressed, I guess the refactor will have to be more consequent :sweat: 

EDIT: Actually, this didn't fix my issue with https://github.com/OCA/sale-workflow/pull/3837 integrated into the patched function. I'll get back to it next week

florian-dacosta:
Hello,
I also have this issue since always, did not attend because not so critical in my case, but you have a similar issue with https://github.com/OCA/sale-workflow/blob/16.0/sale_order_lot_generator
The lot is generated even if the sale has exception and is not confirmed.

I agree this is a general issue of this module and the glue module stuff for me is not good enough. When a module override the _action_confirm, it can only expects its action are either rollbacked in case of exception, either commit if the sale order is confirmed, but commited without the SO validation, can only leads to weird cases.

Between the multiple PR around this, this one seems to do the job. I would also be in favor on merging this until https://github.com/OCA/sale-workflow/pull/3885 is completed.
Although I also think it deserves a comment in the _register_hook and also a new test to how the issue.

sale_exception/models/sale_order.py
```diff
@@ -47,10 +47,22 @@ def sale_check_exception(self):
...
-    def action_confirm(self):
-        if self.detect_exceptions():
-            return self._popup_exceptions()
-        return super().action_confirm()
+    def _register_hook(self):
+        ModelClass = self.env.registry["sale.order"]
+
+        original_action_confirm = ModelClass.action_confirm
+
+        def patched_action_confirm(self):
```
alexey-pelykh, line 62: The `sale_exception_show_popup` company setting check from the original `action_confirm` is missing here. The base branch has:

```python
if not self.env.company.sale_exception_show_popup:
    return
```

This must be preserved to avoid a behavioral regression for users who have disabled the popup. The patched function should be:

```python
def patched_action_confirm(self):
    if self.detect_exceptions():
        if not self.env.company.sale_exception_show_popup:
            return
    original_func = patched_action_confirm.origin
    return original_func(self)
```

sale_exception/tests/test_sale_exception.py
```diff
@@ -13,6 +13,8 @@ class TestSaleException(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env["sale.order"]._register_hook()
+        cls.env["exception.rule"].search([]).write({"active": False})
```
alexey-pelykh, line 17: Minor: The `_register_hook()` call and `exception.rule` deactivation run before `tracking_disable` is set in the context (line below). Consider reordering so the `tracking_disable` context is established first, to avoid unnecessary mail/tracking side effects during this setup:

```python
cls.env["sale.order"]._register_hook()
cls.env["exception.rule"].search([]).write({"active": False})
```

alexey-pelykh (changes requested):
Thank you for the contribution! I understand the problem you are solving: when `sale_loyalty` (or similar modules) run logic inside `action_confirm` before `super()` reaches `sale_exception`'s override, side effects can occur even when an exception should have blocked confirmation.

However, I have significant concerns with the current approach.

### Behavioral regression: `sale_exception_show_popup` check dropped

The current `action_confirm` on the base branch respects the `sale_exception_show_popup` company setting:

```python
def action_confirm(self):
    if self.detect_exceptions():
        if not self.env.company.sale_exception_show_popup:
            return
```

The patched version in this PR always calls `self._popup_exceptions()` when exceptions are detected, ignoring the company setting. Users who have explicitly disabled the popup will experience a regression -- they will start seeing popups (or lose the silent-block behavior).

### Monkey-patching via `_register_hook` is fragile for this use case

`_register_hook` with runtime method replacement bypasses Odoo's standard MRO-based inheritance. This creates several risks:

- **Ordering fragility**: Other modules that also patch `action_confirm` (whether via standard inheritance or similar hooks) may conflict with this patch in unpredictable ways. The manually wired `.origin` chain is a reimplementation of what Odoo's MRO already provides.
- **Debuggability**: Stack traces through monkey-patched closures are harder to follow than standard `super()` chains.
- **Maintenance burden**: Future contributors may not expect a `_register_hook` approach for what looks like a straightforward method override.

### Suggested alternative approach

Have you considered overriding the specific methods that cause the premature side effects instead? For the `sale_loyalty` case, a targeted override of the loyalty computation that checks whether exceptions exist before proceeding might be more surgical.

Alternatively, if the goal is truly to ensure exception checking happens first regardless of MRO, a cleaner pattern might be to use a `@api.constrains` or a `write`-triggered check on state transition, which would fire independently of the `action_confirm` call chain.

If `_register_hook` is the chosen path, at minimum please:
1. Restore the `sale_exception_show_popup` check in the patched function.
2. Add a comment explaining why the standard inheritance approach is insufficient.
3. Consider the interaction with other modules that may also patch `action_confirm`.

alexey-pelykh (approved):
The blocking issue from my previous review is resolved — the `sale_exception_show_popup` company setting check is now correctly preserved in the patched `action_confirm`.

CI is fully green.

Minor: in `setUpClass`, `_register_hook()` and `exception.rule` deactivation still run before `tracking_disable` is set. Reordering would avoid unnecessary mail side effects during setup, but not blocking.

simahawk:
fix for v18 is getting merged https://github.com/OCA/server-tools/pull/3476
companion PR for sale https://github.com/OCA/sale-workflow/pull/4255

PR #3538, opened 2025-01-15
[18.0][MIG] sale_order_secondary_unit: Migration to 18.0

matiasperalta1:

PR #3773, opened 2025-06-25
[18.0][MIG] sale_order_secondary_unit: Migration to version 18.0

pilarvargas-tecnativa:
Supersedes https://github.com/OCA/sale-workflow/pull/3538

This PR respects the authorship of the migration of the PR it supersedes.

@Tecnativa TT54291

@carlos-lopez-tecnativa @carlosdauden please review


sale_order_secondary_unit/views/product_secondary_unit_views.xml
```diff
@@ -0,0 +1,58 @@
...
+        </field>
+    </record>
+
+    <record id="product_secondary_unit_view_tree" model="ir.ui.view">
+        <field name="name">product.secondary.unit.sale.view.list</field>
+        <field name="model">product.secondary.unit</field>
+        <field name="arch" type="xml">
+            <list editable="top" multi_edit="1">
+                <field name="product_tmpl_id" />
+                <field name="product_id" />
```
carlos-lopez-tecnativa: The most common use case is adding secondary units from the product or product template. However, if a user adds a record from this menu, the variant can be selected freely and is not related to the template. Please add a domain to handle this.
image

sale_order_secondary_unit/models/product_template.py
```diff
@@ -0,0 +1,72 @@
...
+    _inherit = "product.template"
+
+    sale_secondary_uom_id = fields.Many2one(
+        comodel_name="product.secondary.unit",
+        string="Default secondary unit for sales",
+        compute="_compute_sale_secondary_uom_id",
+        inverse="_inverse_sale_secondary_uom_id",
+        help="In order to set a value, please first add at least one record"
+        " in 'Secondary Unit of Measure'",
+        domain="[('product_tmpl_id', '=', id), ('product_id', '=', False)]",
```
carlos-lopez-tecnativa, line 16: Why doesn't this field show the expected records? I have three records, two without variants, so I expected those to be selectable, but only one is displayed. However, if I `search more`, I can see the two records. This seems inconsistent to me, could you please check it?
image

image

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -0,0 +1,64 @@
...
+    _secondary_unit_fields = {
+        "qty_field": "product_uom_qty",
+        "uom_field": "product_uom",
+    }
+
+    secondary_uom_unit_price = fields.Float(
+        string="2nd unit price",
+        digits="Product Price",
+        compute="_compute_secondary_uom_unit_price",
+    )
```
carlos-lopez-tecnativa, line 18: What is the purpose of this field? I don’t see it in the views or the report, why is that?

sale_order_secondary_unit/report/sale_report_templates.xml
```diff
@@ -0,0 +1,26 @@
...
+<odoo>
+    <template
+        id="report_saleorder_document"
+        inherit_id="sale.report_saleorder_document"
+    >
+        <xpath
+            expr="//table[hasclass('o_has_total_table','table','o_main_table')]//th[@name='th_quantity']"
+            position="before"
+        >
+            <th name="th_secondary_uom_id" class="text-right" groups="uom.group_uom">
```
carlos-lopez-tecnativa:
```suggestion
            <th name="th_secondary_uom_id" class="text-end" groups="uom.group_uom">
```

sale_order_secondary_unit/report/sale_report_templates.xml
```diff
@@ -0,0 +1,26 @@
...
+                <span>Second. Qty</span>
+            </th>
+        </xpath>
+        <xpath expr="//span[@t-field='line.product_uom_qty']/.." position="before">
+            <td name="td_secondary_uom_id" class="text-right" groups="uom.group_uom">
```
carlos-lopez-tecnativa:
```suggestion
            <td name="td_secondary_uom_id" class="text-end" groups="uom.group_uom">
```

carlos-lopez-tecnativa (changes requested):

sale_order_secondary_unit/models/product_template.py
pilarvargas-tecnativa: The result is as expected: if a record is specific to the variant, it cannot be selected in the product template, but it can be selected in the variant. This is how the field domain defines it, and it is the expected behaviour. To select the default unit of measure for the product template, that secondary unit must not be specific to the variant.

sale_order_secondary_unit/models/sale_order.py
pilarvargas-tecnativa: All done, thanks!

sale_order_secondary_unit/models/product_template.py
carlos-lopez-tecnativa: > The result is as expected: if a record is specific to the variant, it cannot be selected in the product template, but it can be selected in the variant. This is how the field domain defines it, and it is the expected behaviour. To select the default unit of measure for the product template, that secondary unit must not be specific to the variant.

Please check the image and the previous comment. I said, "I have three records, **two without variants**." I expected these two records to be selectable.

carlos-lopez-tecnativa: Please check this comment. In this case, I created three secondary units without a variant, but only the first SU is displayed. However, when I search further, all three records are available. I don't understand why.
image

Additionally, in the sale order, only one SU is displayed, but I expected to see all three records.
image

sale_order_secondary_unit/models/sale_order.py
carlos-lopez-tecnativa: Sorry, but I don't understand what you changed. To me, this field is unset. Could you please clarify the purpose?

sale_order_secondary_unit/views/product_secondary_unit_views.xml
```diff
@@ -0,0 +1,61 @@
...
+                <group expand="1" string="Group By">
+                    <filter
+                        string="Product Template"
+                        name="product_tmpl_id"
+                        context="{'group_by':'product_tmpl_id'}"
+                    />
+                    <filter
+                        string="Product Variant"
+                        name="product_id"
+                        context="{'group_by':'product_tmpl_id'}"
```
carlos-lopez-tecnativa: This filter has an incorrect field; please update it. Additionally, please add the corresponding security group.

```suggestion
                        context="{'group_by':'product_id'}"
                        groups="product.group_product_variant"
```

sale_order_secondary_unit/views/product_secondary_unit_views.xml
```diff
@@ -0,0 +1,61 @@
...
+    <record id="product_secondary_unit_view_tree" model="ir.ui.view">
+        <field name="name">product.secondary.unit.sale.view.list</field>
+        <field name="model">product.secondary.unit</field>
+        <field name="arch" type="xml">
+            <list editable="top" multi_edit="1">
+                <field name="product_tmpl_id" />
+                <field
+                    name="product_id"
+                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"
+                />
```
carlos-lopez-tecnativa, line 42: Please add the corresponding security group.
```suggestion
                <field
                    name="product_id"
                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"
                    groups="product.group_product_variant"
                />
```

carlos-lopez-tecnativa (changes requested):

sale_order_secondary_unit/models/product_template.py
pilarvargas-tecnativa: In runboat I don't replicate the problem you mention.

image

image

sale_order_secondary_unit/models/sale_order.py
pilarvargas-tecnativa: Hi @Kev-Roche this field was removed from the view in v15. 

https://github.com/OCA/sale-workflow/commit/1710a9d1ff18eb915cf47dd1953d2631e45cb667#diff-f206d07f7cdb9ec1f638ab75260d3e525b36c7f7dc4590562c6930540f51beb3L28

Is it still needed in the module? Should it be restored to the view again?

rousseldenis (approved):

marcelsavegnago (approved):
LGTM

sale_order_secondary_unit/models/sale_order.py
carlos-lopez-tecnativa: This field is unused (and it's not stored, so it can be removed later if necessary). Anyway, let's go ahead and merge.

carlos-lopez-tecnativa (approved):
LGTM

PR #3774, opened 2025-06-25
[17.0][IMP] sale_order_secondary_unit: context for skipping default sec. unit

carlos-lopez-tecnativa:
FWP of https://github.com/OCA/sale-workflow/pull/3588
In some cases we might want to skip the default secondary unit to be set. This change brings that possibility.

TT55850
@Tecnativa @pedrobaeza @pilarvargas-tecnativa could you please review this?


pedrobaeza (approved):
Adapted fw-port, so fast-tracking it:

/ocabot merge patch

PR #3775, opened 2025-06-26
[ADD] sale_order_compute_commitment_date: New module sale_order_compute_commitment_date

miquelpascual:
This module automatically computes sale orders commitment date based on the product template sale_delay
and product attribute variant lead time fields.

cc https://github.com/APSL 14210
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


sale_order_compute_commitment_date/models/product_template.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 APSL Nagarro
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    extend_lead_time = fields.Boolean(
```
rousseldenis, line 10: IMHO, this field name should reflect the `attribute` term as the name is too generic.

Something like `attribute_extend_lead_time`.

sale_order_compute_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+        compute="_compute_commitment_date",
+        store=True,
+        readonly=False,
+        help="This is the delivery date promised to the customer. "
+        "If set, the delivery order will be "
+        "scheduled based on this date rather than product lead times.",
+    )
+
+    @api.depends(
+        "order_line.product_id",
```
rousseldenis, line 25: IMHO, only `order_line.product_id` should be taken into account here.

You should manage the fact changing the delay, the product variant lead time or the extend configuration will trigger a compute on every existing sale order line that uses that product. And that is unwanted behavior

rousseldenis (commented):
@mpascuall

sale_order_compute_commitment_date/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Sale Order Compute Commitment Date",
+    "version": "18.0.1.0.0",
+    "category": "Sales Management",
+    "summary": "Automatically computes the commitment date based on product"
+    "lead times and attributes",
+    "author": "APSL - Nagarro, Odoo Community Association (OCA)",
+    "maintainers": ["mpascuall"],
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "LGPL-3",
```
alexey-pelykh, line 13: License mismatch: the manifest declares `LGPL-3` but all Python file headers use `AGPL-3.0 or later`. These must be consistent. For OCA modules, `AGPL-3` is the standard -- please update the manifest to `"license": "AGPL-3"` (or change all headers to LGPL-3 if that's intended).

Also, `"base"` should be removed from `depends` -- it's always loaded by Odoo and OCA linters will flag it.

sale_order_compute_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+        for order in self:
+            max_total_lead_time = 0
+
+            for line in order.order_line:
+                product_template = line.product_id.product_tmpl_id
+
+                current_line_lead_time = (
+                    product_template.sale_delay
+                    if product_template.sale_delay is not None
+                    else 0
```
alexey-pelykh, line 44: Strongly agree with @rousseldenis here. The depends list should only include `"order_line.product_id"` and `"date_order"`.

Including `sale_delay`, `extend_lead_time`, attribute `lead_time`, `state`, and company calendar attendance/leave fields means:

1. **Performance**: Changing `sale_delay` on a popular product template will trigger recomputation on potentially thousands of historical sale orders.
2. **Correctness**: `state` is never read in the method body, so it shouldn't be a dependency.
3. **Side effects**: Editing the company's working calendar would recompute commitment dates on all sale orders for that company.

If the intent is to recompute on confirmation, consider overriding `action_confirm()` to call the compute manually.

sale_order_compute_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+        "company_id.resource_calendar_id.leave_ids",
+    )
+    def _compute_commitment_date(self):
+        for order in self:
+            max_total_lead_time = 0
+
+            for line in order.order_line:
+                product_template = line.product_id.product_tmpl_id
+
+                current_line_lead_time = (
```
alexey-pelykh, line 41: Redefining `commitment_date` as a computed field changes its behavior globally for all sale orders. If a user manually sets a commitment date on an order, it will be overwritten on the next recompute trigger (e.g. adding/changing an order line).

Consider either:
- Only computing when the field is empty (check `if not order.commitment_date` before assigning), or
- Adding a Boolean like `is_commitment_date_computed` to track whether it was auto-set vs. manually entered, or
- Using an `@api.onchange` instead of a stored compute, so it suggests a value but doesn't overwrite manual entries after save.

sale_order_compute_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,91 @@
...
+                    product_template.sale_delay
+                    if product_template.sale_delay is not None
+                    else 0
+                )
```
alexey-pelykh, line 45:
```suggestion
                current_line_lead_time = product_template.sale_delay or 0
```

Odoo integer fields return `0` (not `None`) when unset, so `is not None` is always `True` and the ternary is misleading.

alexey-pelykh (changes requested):
Thanks for the contribution. This module addresses a real need -- automatically computing commitment dates from product lead times and variant attributes is a useful feature.

However, I've identified several issues that should be addressed before this can be merged. I'm largely aligned with @rousseldenis's earlier feedback on the `@api.depends` scope, and I've found a few additional concerns.

## Key Findings

| 1 | Critical | `__manifest__.py` / headers | **License mismatch**: manifest declares `LGPL-3` but all file headers say `AGPL-3.0`. These must be consistent. OCA modules typically use `AGPL-3`. |
| 2 | Critical | `sale_order.py` | **Over-broad `@api.depends`**: Depending on `sale_delay`, `extend_lead_time`, attribute `lead_time`, `state`, and company calendar fields means changing any of these will trigger recomputation on *every* existing sale order referencing those records. This is both a performance risk and a functional problem (users don't expect historical orders to change). Only `order_line.product_id` and `date_order` should be in depends (as @rousseldenis noted). |
| 3 | Critical | `sale_order.py` | **`state` in depends but never read**: The compute method never references `order.state`, so `state` should not be in `@api.depends`. If the goal is to recompute on confirmation, override `action_confirm()` instead. |
| 4 | Major | `sale_order.py` | **Field redefinition changes upstream behavior**: Redefining `commitment_date` as `compute/store/readonly=False` changes the field's behavior globally -- even for orders where no product has `sale_delay` set. A user who manually sets a commitment date will have it overwritten on the next recompute trigger. Consider adding a Boolean flag (e.g. `is_commitment_date_manual`) or only computing when the field is empty/falsy. |
| 5 | Minor | `sale_order.py:43-45` | **`sale_delay is not None` is misleading**: Odoo integer fields are never Python `None`; they return `0` when unset. The `is not None` check is always `True`. Simplify to `product_template.sale_delay or 0` or just `product_template.sale_delay`. |
| 6 | Minor | `__manifest__.py` | **`base` in depends**: `base` is always loaded by Odoo and should not be listed as an explicit dependency. OCA linters (`oca-checks`) will flag this. |
| 7 | Minor | `l10n/es.po` | **Translation directory**: OCA convention is `i18n/`, not `l10n/`. The `oca-gen-addon-readme` tooling and CI expect `i18n/`. |
| 8 | Minor | `readme/*.md` | **Missing trailing newline**: `CONFIGURE.md`, `CONTRIBUTORS.md`, and `DESCRIPTION.md` are all missing a final newline character. |
| 9 | Suggestion | `product_template.py` | **Field naming**: Agree with @rousseldenis that `extend_lead_time` is too generic. Something like `attribute_extend_lead_time` makes the purpose clearer. |
| 10 | Suggestion | `sale_order.py` | **Calendar fallback**: When no resource calendar is configured, the code falls back to a simple `timedelta(days=...)` which counts calendar days. This behavioral difference (working days vs. calendar days) depending on calendar presence could be surprising. Consider documenting this or always using calendar days as fallback. |

rousseldenis:
@mpascuall What's the status of this ?

rousseldenis:
@miquelpascual This is quite old. Do you plan to fix it or shall we close it ?

miquelpascual:
> @miquelpascual This is quite old. Do you plan to fix it or shall we close it ?

Hi, sorry for the late reply. I will fix it!

sale_order_compute_commitment_date/__manifest__.py
miquelpascual: Many thanks for your comments, i fixed it!

sale_order_compute_commitment_date/models/sale_order.py
miquelpascual: Done! Many thanks.

PR #3779, opened 2025-06-27
[18.0][ADD] sale_action_paid

jbaudoux:
Adds a button "Paid" on a sent sales order. This is especially convenient for
wire transfer, to mark the SO as paid. It can also serve for other payment
modes in case the postprocessing failed but the payment went through.

The action will confirm the SO. An email confirmation will be generated to
notify the client that the order has been confirmed.

When there is pending payment transaction (for instance for webshop orders),
the transaction will be confirmed.

The SO invoicing policy will be forced to on order to be able to generated the
invoice for any non invoiced line.

According to odoo standard settings, the invoice will be posted and the invoice
will be sent.


sale_action_paid/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+                # recover invoice creation to allow manual payment reconciliation
+                if auto_invoice:
+                    # Invoice the sales orders of confirmed transactions
+                    # instead of only confirmed orders to create the invoice
+                    # even if only a partial payment was made.
+                    tx._invoice_sale_orders()
+                    tx.invoice_ids.filtered(
+                        lambda inv: inv.state == "draft"
+                    ).action_post()
+                tx.is_post_processed = True
```
alexey-pelykh: This line is redundant. `_post_process()` (called above on line 27) already sets `self.is_post_processed = True` in the base `payment` module. Can be removed.

sale_action_paid/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+                lambda t: t.state in ("draft", "pending")
+            )
+            tx = txs._get_last()
+            if not tx:
+                tx = txs.sorted()[:1]
+            if tx:
+                tx._set_done()
+                # Prevent to generate a new payment for marking the invoice as paid.
+                # This will disable the send_payment_succeeded_for_order_mail,
+                # auto invoice, invoice sending.
```
alexey-pelykh: Changing `tx.operation` to `"validation"` persists on the transaction record. Any downstream logic that inspects the operation type will now see `"validation"` instead of the original value. This is a somewhat invasive side effect. Consider at least saving and restoring the original value:

```python
original_operation = tx.operation
tx.operation = "validation"
tx._post_process()
tx.operation = original_operation
```

Or, if the change is intentional, please document why `"validation"` is the correct permanent state.

sale_action_paid/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+            self.env["ir.config_parameter"].sudo().get_param("sale.automatic_invoice")
+        )
+        for order in self.filtered(lambda so: so.state in ("draft", "sent")):
+            txs = order.sudo().transaction_ids.filtered(
+                lambda t: t.state in ("draft", "pending")
+            )
+            tx = txs._get_last()
+            if not tx:
+                tx = txs.sorted()[:1]
+            if tx:
```
alexey-pelykh: If no draft/pending transaction exists (e.g. the SO was created manually without a payment link), `tx` will be falsy and `action_paid()` silently does nothing. The user clicks "Paid" with no feedback.

Consider raising a `UserError` when no eligible transaction is found:
```python
...
if not tx:
    raise UserError(_("No pending payment transaction found for this order."))
```

sale_action_paid/tests/test_action_paid.py
```diff
@@ -0,0 +1,43 @@
...
+            {
+                "partner_id": cls.partner.id,
+                "order_line": [
+                    Command.create({"product_id": cls.product.id, "product_uom_qty": 1})
+                ],
+            }
+        )
+
+    def test_action_paid(self):
+        """Test that the communication displayed is the sale order reference."""
```
alexey-pelykh: The docstring says *"Test that the communication displayed is the sale order reference"* but this test validates the `action_paid` flow. Looks like a copy-paste error -- please update the docstring to match the actual test.

sale_action_paid/views/sale_order_views.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <button name="action_cancel" position="after">
+                <button
+                    name="action_paid"
+                    type="object"
+                    class="oe_highlight"
+                    string="Paid"
+                    invisible="state not in ('draft', 'sent')"
```
alexey-pelykh, line 14: Consider restricting the button visibility to `state == 'sent'` only (instead of `draft` and `sent`). A draft quotation typically has no payment transaction, so showing "Paid" on it is likely confusing.

Alternatively, add a computed boolean field (e.g. `has_pending_transaction`) and use it in the `invisible` condition.

sale_action_paid/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Jacques-Etienne Baudoux (BCIM) \<<je.bcim.be>\>
```
alexey-pelykh: The email address appears malformed: `<<je.bcim.be>>` should be `<je@bcim.be>` (missing `@` symbol).

alexey-pelykh (changes requested):
## Review: sale_action_paid

Thank you for this module -- the use case (marking wire-transfer orders as paid without going through ecommerce postprocessing) is practical and well-motivated.

I do have several concerns that I believe should be addressed before merging.

---

### Unresolved FIXME in PR description

The PR body states:

> **FIXME**: i think a payment should be created to mark the invoice "in payment" until it is reconciled

This is an open design question. The invoice created by this action will go straight to `posted` without an associated payment journal entry. Until reconciliation happens, the invoice will not reflect "in payment" status. This could lead to incorrect aging reports and cash-flow visibility. This should be resolved or explicitly documented as a known limitation before merging.

### Core logic concerns in `action_paid()`

1. **Silent no-op when no transaction exists**: If the SO has no draft/pending transaction, the method silently does nothing. The user clicks "Paid" and nothing happens -- no confirmation, no error, no feedback. This is poor UX. At minimum, raise a `UserError` so the user knows why the action did not proceed.

2. **Redundant `is_post_processed` assignment**: Line 38 (`tx.is_post_processed = True`) is redundant -- `_post_process()` (called on line 27) already sets this flag in the base `payment` module. This line can be removed.

3. **Transaction `operation` permanently changed to `"validation"`**: Setting `tx.operation = "validation"` (line 26) is a side effect that persists on the transaction record. Any downstream code or reporting that inspects the operation type will now see `"validation"` instead of the original value (e.g. `"online_redirect"`, `"offline"`). Consider saving and restoring the original operation, or document this as intentional.

4. **Confusing `_get_last()` + fallback logic**: The code filters transactions to `draft`/`pending`, then calls `_get_last()` which internally filters OUT `draft` state. If `_get_last()` returns empty, the fallback `txs.sorted()[:1]` picks the first draft. This two-step approach is hard to follow. Consider simplifying: prefer `pending` transactions first, fall back to `draft`.

5. **Missing `sudo()` on `_set_done()` and `_post_process()`**: The transactions are fetched via `order.sudo().transaction_ids`, but `_set_done()` and `_post_process()` are called without sudo. These internal methods may write to payment provider records or send notifications that require elevated privileges. This could fail for users without payment admin rights.

### Button visibility

The button is visible on both `draft` and `sent` states. A draft quotation (not yet sent to the customer) typically has no payment transaction. Showing "Paid" on draft quotations could be confusing. Consider restricting visibility to `sent` state only, or at least adding a condition checking that a pending/draft transaction exists (e.g. via a computed `Boolean` field).

### Test coverage

- **Docstring mismatch**: The test docstring says *"Test that the communication displayed is the sale order reference"* but the test actually validates the `action_paid` flow. This appears to be a copy-paste error.

- **Missing edge cases**: Only the happy path (pending tx + auto_invoice=True) is tested. Consider adding tests for:
  - No transaction on the order (should raise `UserError` if the above suggestion is adopted)
  - `auto_invoice` = False (verify no invoice is created)
  - Order already confirmed (`sale` state)
  - Multiple transactions on the order

### Minor issues

- **Contributor email**: In `readme/CONTRIBUTORS.md`, the email appears as `<<je.bcim.be>>` -- the `@` is missing. Should be `<je@bcim.be>`.

- **Line length**: `test_action_paid.py` line 23 (`Command.create(...)`) exceeds the typical 79/88 char limit for OCA modules.

---

**Verdict**: REQUEST_CHANGES -- the silent no-op, FIXME in description, and limited test coverage are the main blockers.

sale_action_paid/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,12 @@
+Adds a button "Paid" on a sent sales order.
+
+The action will confirm the SO and, in case of website order, it will also mark
+as done the pending payment transaction. For manually created order, a payment
+transaction will be created.
```
hbrunn: I don't really understand why you create a payment transaction in this case. What do you need that for?

sale_action_paid/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,12 @@
...
+Note that marking as done the payment transaction will change the invoicing
+policy to on order (odoo standard). This can also enable standard automatic
+invoicing.
+
+This is especially convenient for wire transfer, to mark the SO as paid.
+It can also serve for other payment mode in case the postprocessing failed but
```
hbrunn:
```suggestion
It can also serve for other payment modes in case the postprocessing failed but
```

sale_action_paid/models/sale_order.py
```diff
@@ -0,0 +1,68 @@
...
+        transaction_vals = dict(default_values, **values)
+        if not transaction_vals.get("payment_method_id"):
+            # default to Wire transfer
+            method = self.env.ref(
+                "payment_custom.payment_method_wire_transfer", raise_if_not_found=False
+            )
+            if not method:
+                raise UserError(
+                    self.env._(
+                        "Cannot mark as paid as there are no pending transaction. "
```
hbrunn:
```suggestion
                        "Cannot mark as paid as there are no pending transactions. "
```

hbrunn (commented):
I wonder if you can't move setting the transaction to done into `account.move#_invoice_paid_hook`?

sale_action_paid/readme/DESCRIPTION.md
jbaudoux: Thanks for the review. Indeed I think it's overkill and better to mimic what post_process is doing if there is no transaction.

I changed it in this commit:
* https://github.com/OCA/sale-workflow/pull/3779/changes/0a24ea46b63c8f0b882a0c7a31e472e58318897c

jbaudoux:
> I wonder if you can't move setting the transaction to done into `account.move#_invoice_paid_hook`?

I want to rely on odoo standard transaction processing to let odoo confirm the order, send email, generate invoice and send invoice.

alexey-pelykh (commented):
Good progress — the test coverage is now substantially better (4 test cases covering webshop, manual, confirmed, and downpayment scenarios), and the docstring/contributor email are fixed.

A few items from my previous review remain. None are blocking individually, but addressing them would strengthen the module:

1. **`tx.is_post_processed = True` is still redundant** — `_post_process()` already sets this.
2. **`tx.operation = "validation"` persists permanently** — consider save/restore or a comment explaining why.
3. **`_set_done()` and `_post_process()` called without `sudo()`** — transactions are fetched via `self.sudo().transaction_ids`, but subsequent calls drop to user rights. Could fail for non-payment-admin users.
4. **Button still visible on `draft` quotations** — consider restricting to `state == 'sent'` or adding a computed boolean condition.

sale_action_paid/models/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+    def _action_paid_get_transaction(self):
+        self.ensure_one()
+        txs = self.sudo().transaction_ids.filtered(
+            lambda t: t.state in ("draft", "pending")
+        )
+        if not txs:
+            return
+        tx = txs._get_last()
+        if not tx:
+            tx = txs.sorted()[:1]
```
rousseldenis, line 19: @jbaudoux Be cautious with slice as Odoo does not preserve prefetch (before 19.1) in that case (`next(iter())` is better in 18.0)

rousseldenis (approved):

rousseldenis:
@jbaudoux

rousseldenis:
@hbrunn Could oou review again ?

PR #3782, opened 2025-07-01
[17.0][FIX] sale_order_secondary_unit: fix tests when run alongside sale_stock

david-banon-tecnativa:
When tests ran in a database with sale_stock installed, tests were failing.
I still have no clue of the root cause that makes tests pass when standalone, but fail when sale_stock is installed, but this fixes the issue.

@tecnativa @pedrobaeza @carlos-lopez-tecnativa 
TT52379


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
Code review

PR #3784, opened 2025-07-02
[16.0][IMP] sale_order_safe_commitment_date: unsafe delivery date as expected

chienandalu:
Users who set the delivery date expect it to be present in order to be able to filter by it. In practice it doesn't make much difference having it or not.

cc @moduon MT-9790

please review @Shide @Gelojr @rafaelbn


Shide (approved):
code lgtm

rafaelbn (approved):
👍🏼 

/ocabot merge minor

PR #3786, opened 2025-07-02
[18.0][MIG] sale_order_priority: Migration to 18.0

apami0111:
- Migrated from 17.0 to 18.0.
- Updated xpath expression in the view to reflect Odoo 18.0 changes (//list instead of //tree).
- Test class now uses Command.create to define order_line values, in line with Odoo 18.0 conventions.


sale_order_priority/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Sale Order Priority",
     "summary": "Define priority on sale orders",
-    "version": "17.0.1.0.1",
+    "version": "18.0.1.0.1",
```
rousseldenis: Please reset to `18.0.1.0.0`

sale_order_priority/models/sale.py
```diff
@@ -0,0 +1,47 @@
...
+    def _inverse_priority(self):
+        for order in self:
+            priority = order.priority
+            for line in order.order_line.filtered(
+                lambda x, pr=priority: x.priority != pr
+            ):
+                line.priority = priority
+
+    def action_confirm(self):
+        return super(
```
rousseldenis: This is not correct as `self` can be multi recordset.

Moreover, I think the transmission of the priority should be done through procurement if possible as context should be avoided.

rousseldenis (changes requested):

apami0111:
Thank you very much for your comment. I will be working on the suggested
changes and will update the PR accordingly.

apami0111:
@rousseldenis Thanks for your feedback! I've addressed the issue with the multi-recordset by adjusting the logic accordingly. I’ve also reworked the way priority is transmitted, avoiding the use of context as suggested.

The following structural changes were made as well:
- Renamed: sale_order_priority/models/{sale.py => sale_order.py}
- Created: sale_order_priority/models/sale_order_line.py
- Created: sale_order_priority/models/stock_move.py
- Removed: sale_order_priority/models/stock_picking.py

Feel free to review the updated implementation. Let me know if anything else needs to be improved.

rousseldenis (approved):

celm1990 (approved):
LGTM

marcelsavegnago (approved):
LGTM

BhaveshHeliconia (approved):
Functional review LGTM!

PR #3789, opened 2025-07-03
[18.0] [IMP] sale_exception: Confirm directly the sale order when ignoring exceptions

ivantodorovich:
When there are exceptions blocking the confirmation of an order, the user normally finds out when clicking on the "Confirm" button.

At that time, the wizard is popup and the user -if he has enough permissions- can ignore the exceptions and click confirm.

Before this commit, the wizard would only flag the order to ignore the exceptions, but the order would not be confirmed. The user would have to click again on the "Confirm" button to do so.

This is counter-intuitive since the user intention was clearly to confirm the order in the first place, and it just forces the user to click again.

After this commit, the order is confirmed directly.

Additionally, the wizard will now hide the "Confirm" button if the exceptions haven't been ignored.

### DIFF
--- a/sale_exception/tests/test_sale_exception.py
+++ b/sale_exception/tests/test_sale_exception.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import Command
-from odoo.exceptions import ValidationError
+from odoo.exceptions import UserError, ValidationError
 from odoo.tests import Form, TransactionCase
 
 
@@ -126,7 +126,7 @@ class TestSaleException(TransactionCase):
             }
         )
         exception.active = False
-        so1.action_cancel()
+        so1.with_context(disable_cancel_warning=True).action_cancel()
         so1.action_draft()
         self.assertTrue(not so1.ignore_exception)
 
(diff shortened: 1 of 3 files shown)


rousseldenis:
> When there are exceptions blocking the confirmation of an order, the user normally finds out when clicking on the "Confirm" button.
> 
> At that time, the wizard is popup and the user -if he has enough permissions- can ignore the exceptions and click confirm.
> 
> Before this commit, the wizard would only flag the order to ignore the exceptions, but the order would not be confirmed. The user would have to click again on the "Confirm" button to do so.
> 
> This is counter-intuitive since the user intention was clearly to confirm the order in the first place, and it just forces the user to click again.
> 
> After this commit, the order is confirmed directly.
> 
> Additionally, the wizard will now hide the "Confirm" button if the exceptions haven't been ignored.

@ivantodorovich I think this is the case from the beginning of this module.

@mourad-ehm @rvalyi @yvaucher Do you have an advice on this ?

yvaucher:
I remember that this module forces you do twice the action and I don't see any reason to have this behavior. Always though it was an annoying limitation.

It's great to make it more intuitive.

yvaucher (approved):
LGTM

TDu (approved):

sebalix (approved):

PR #3724, opened 2025-05-16
[16.0][IMP] sale_order_line_cancel: Do not update canceled qty in state draft

mt-software-de:
To make this addon compatible with sale_stock_prebook,
the product_qty_canceled should only be updated in a not draft status.
@i-vyshnevska @raumschmiede-joshuaL @jbaudoux @mmequignon @rousseldenis

PR #3790, opened 2025-07-03
[16.0][IMP] sale_order_line_cancel: Do not update canceled qty in state sent

mt-software-de:
In order to make #3724 complete we should also check for state `sent`
@i-vyshnevska @raumschmiede-joshuaL @jbaudoux @mmequignon @rousseldenis


jbaudoux (approved):
idd

rousseldenis (approved):

PR #3795, opened 2025-07-07
[18.0][MIG] sale_order_line_input: Migration to 18.0

victoralmau:
Supersedes https://github.com/OCA/sale-workflow/pull/3458

Changes done:
- [x] Purge translations
- [x] Squash administrative commits
- [x] https://github.com/OCA/sale-workflow/pull/3458#pullrequestreview-2814429503
- [x] Use `BaseCommon` in tests
- [x] Remove invisible fields (according to the migration guide they are not needed, odoo will auto-add them if needed).

Migration to 18.0

Please @carlosdauden and @pedrobaeza can you review it?

@Tecnativa TT54282


sale_order_line_input/models/sale_order.py
```diff
@@ -51,7 +51,3 @@ def _onchange_order_partner_id(self):
             onchange_method(new_so)
         order_vals = new_so._convert_to_write(new_so._cache)
         self.order_id = SaleOrder.create(order_vals)
-
-    def action_sale_order_form(self):
```
pedrobaeza: We have lost from 15.0 the ability to open the sales order form, so please restore it.

victoralmau: This change has been done at the suggestion of https://github.com/OCA/sale-workflow/pull/3458#pullrequestreview-2814429503

Yes, it is possible to access the order

ejemplo-1

ejemplo

What problem do you see with this behavior?

pedrobaeza: OK then.

pedrobaeza (approved):

sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -0,0 +1,239 @@
...
+                    context="{'res_partner_search_mode': 'customer', 'default_is_company': True, 'default_customer_rank': 1}"
+                    options='{"always_reload": True}'
+                    required="1"
+                    readonly="order_partner_id and order_id"
+                />
+                <field name="write_date" widget="date" optional="hide" />
+                <field
+                    name="product_id"
+                    readonly="not product_updatable"
+                    required="not display_type"
```
carlosdauden: This is something I have to comment on every migration...
The objective of this module is to manage the order lines from a tree view, but keeping as much as possible the behaviour they have in the normal order view.
To do this, what you have to do in the migration of this module is to go to compare the order view and put the same fields with the same attributes (except the conditions in which ‘parent.’ is used).

This is just an example of the attributes that have not been checked, but there are others like with, string, etc. that are not seen here either:
https://github.com/odoo/odoo/blob/c84b67e5dca1a74e1a5862e754455944015f4105/addons/sale/views/sale_order_views.xml#L531

carlosdauden (changes requested):

sale_order_line_input/views/sale_order_line_view.xml
victoralmau: Ok, “_copied_” the sale view in this module; I don't like this technique because for example many things will not work the same (sections, notes, sequences) in addition to the changes that other modules make in the sales order lines in the form view (add an extra column for example). It could also be the case (with other OCA modules that extend sales) that in order to create new lines, fields that are not in the view are required and/or need to be defined.

The `sol_product_many2one` widget is not supported outside the order form view, therefore the `product_id` field is always shown and the `product_template_id` field is hidden.

2 extra fields have been added:
- `company_price_amount`: They replace the `price_subtotal` + `price_total` fields because it is not possible to make column_invisible with conditional and also because it will not always be the same value (if we are showing lines from different companies and each one has a defined configuration). https://github.com/odoo/odoo/blob/a7a29ed691db4f1607c0d12929c56845681164fa/addons/sale/views/sale_order_views.xml#L657
- `company_price_tax_incl_excl`: Replaces the fields (Optional amounts) `price_total` + `price_subtotal` for the same reason as above. https://github.com/odoo/odoo/blob/a7a29ed691db4f1607c0d12929c56845681164fa/addons/sale/views/sale_order_views.xml#L666

victoralmau:
Ping @CarlosRoca13

carlosdauden (changes requested):
Great work.

I think it is better not to create those additional fields to show the amounts with and without tax.

On the one hand we can cause confusion to the user when viewing records of several companies with different configuration and in the same column some amounts are with taxes and others without taxes.
On the other hand, we increase the complexity of the module in an ‘unnecessary’ way, as I think that one option would be to add all the fields in the view so that the user can manage them as he/she sees fit.

We can put `optional=‘show’` for the fields without taxes and `optional=‘hide’` for the fields with taxes and this way we are coherent with the default values that Odoo brings regarding this configuration.

carlosdauden (approved):
Sorry, I understand that the use of the additional fields does not cause confusion

PR #3802, opened 2025-07-09
[16.0][ADD] sale_product_email : Sending a mail on order confirmation for a given product

Honeyxilia:
This module was proposed for inclusion in the 12.0 branch by @fcayre with the following pull request : 
https://github.com/OCA/sale-workflow/pull/1891

It introduces as a feature to be able to send a mail for a given product (product.template) once a sale order has been confirmed.

There were mainly talks about the method to send mails, which functions through the sale order line by line, but they hadn't progressed since 2022, and we introduced the module in our code base.
Since we're migrating our Odoo instance to v16.0, we migrated this module to 16.0, and we wish to introduce it to the OCA once more.


sale_product_email/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module adds an optional email model to products that will be sent
```
rousseldenis: @Honeyxilia @fcayre IMHO, you should add a context to explain the use case you have. Because this can potentially send a lot of emails to the customer at sale confirmation.

Thanks

Honeyxilia: I've reworded the description file, as well as adding a `USAGE.rst` file with a basic setup for the feature c:

mihien:
Lgtm !

Honeyxilia:
I updated the description to explain the difference between it and `product_email_template`, which boil down to the email sending timing (invoice posting vs. sale order confirm) c:

Honeyxilia:
We spoke IRL with @LaetitiaElabore , and since demo data should be installed when running tests, as is the case for the CI jobs, this is not a particuliar issue c:

LaetitiaElabore (approved):

rousseldenis (approved):

PR #3808, opened 2025-07-10
[16.0][ADD] sale_warehouse_rule

Kev-Roche:
This module allows to force the source warehouse on sale order line from the product, depending of some warehouse rules.

cc @chafique-delli , @sebastienbeau 

depends on : https://github.com/OCA/product-attribute/pull/2022


sale_warehouse_rule/models/product.py
```diff
@@ -0,0 +1,33 @@
...
+    @api.depends(
+        "warehouse_rule_ids",
+        "warehouse_rule_ids.product_id",
+        "warehouse_rule_ids.attribute_value_ids",
+        "warehouse_rule_ids.warehouse_id",
+    )
+    @api.depends_context("allowed_company_ids", "force_company_id")
+    def _compute_variant_warehouse_id(self):
+        for product in self:
+            company_id = self.env.context.get("force_company_id") or self.env.company.id
```
rousseldenis, line 25: `force_company_id` is not used anymore. Replaced by changing env with `with_company()`

rousseldenis:
@Kev-Roche Maybe adding the dependency in `test-requirements` would help to review functionaly.

rousseldenis (approved):

sale_warehouse_rule/models/product.py
```diff
@@ -0,0 +1,33 @@
...
+
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    variant_warehouse_id = fields.Many2one(
+        comodel_name="stock.warehouse",
+        string="Entrepôt par défaut",
```
alexey-pelykh, line 12: Field string is in French (`"Entrepôt par défaut"`). OCA convention requires English for all user-facing strings; translations belong in `.po` files.

Suggestion:
```python
string="Default Warehouse",
```

sale_warehouse_rule/models/product.py
alexey-pelykh, line 25: +1 to @rousseldenis's earlier comment. `force_company_id` is deprecated in Odoo 16. Use `self.env.company` (which respects `with_company()`).

The `@api.depends_context` should then only list `"allowed_company_ids"` (drop `"force_company_id"`).

sale_warehouse_rule/models/sale_order.py
```diff
@@ -0,0 +1,87 @@
...
+                    product
+                    for product in rec.order_line.product_id
+                    if not product.variant_warehouse_id
+                )
+                and rec.warehouse_id not in warehouse_ids
+            ):
+                warehouses = warehouse_ids + rec.warehouse_id
+                rec.warehouse_rule_info = _(
+                    "The delivery will be sent from multiple warehouses: "
+                    + ", ".join(warehouses.mapped("name"))
```
alexey-pelykh, line 54: The translatable string is built by concatenating `_()` with the warehouse names:
```python
_("The delivery will be sent from multiple warehouses: " + ", ".join(...))
```
This means the full string (including warehouse names) is passed to `_()`, so it will **never** match a `.po` entry.

Use a format placeholder instead:
```python
_("The delivery will be sent from multiple warehouses: %s") % ", ".join(warehouses.mapped("name"))
```

sale_warehouse_rule/models/sale_order.py
```diff
@@ -0,0 +1,87 @@
...
+                )
+
+    def action_confirm(self):
+        for rec in self:
+            if rec.warehouse_rule_need_change:
+                warehouse_id = rec.order_line.mapped("product_id.variant_warehouse_id")
+                rec.warehouse_id = warehouse_id
```
alexey-pelykh, line 61: When `warehouse_rule_need_change` is truthy, `mapped("product_id.variant_warehouse_id")` can return **multiple** warehouse records (e.g. if only some products have the same warehouse rule). Assigning a multi-record recordset to `rec.warehouse_id` (a `Many2one`) will raise a `ValueError`.

This code path is only correct when all lines share a single warehouse, but the earlier condition in `_compute_warehouse_rule_message` that sets `warehouse_rule_need_change` already verifies `len(warehouse_ids) == 1`. However, `action_confirm` re-fetches the warehouses without the same guard. If the order lines changed between compute and confirm, this could crash.

Suggestion: add a safety check, e.g.:
```python
warehouse_ids = rec.order_line.mapped("product_id.variant_warehouse_id")
if len(warehouse_ids) == 1:
    rec.warehouse_id = warehouse_ids
```

sale_warehouse_rule/models/sale_warehouse_rule.py
```diff
@@ -0,0 +1,66 @@
...
+        for rule in self:
+            if rule.product_id:
+                rule.applied_on = "0_product"
+            elif rule.attribute_value_ids:
+                rule.applied_on = "1_attribute"
+            else:
+                rule.applied_on = "2_template"
+
+    @api.constrains("product_id", "attribute_value_ids", "warehouse_id")
+    def _check_warehouse_rule_uniqueness(self):
```
alexey-pelykh, line 42: The uniqueness constraint logic mutates the `domain` list by index (`domain[1] = ...`, `domain[2] = ...`), which is fragile and hard to follow.

More importantly, the domain does not exclude the current record (`("id", "!=", rule.id)`) -- when updating an existing rule, `search_count` will find the rule itself and raise a false `ValidationError`.

Consider:
1. Adding `("id", "!=", rule.id)` to the base domain.
2. Building separate, explicit domains for each case instead of mutating one shared list.

sale_warehouse_rule/views/sale_order.xml
```diff
@@ -0,0 +1,45 @@
...
+            </xpath>
+
+            <xpath
+                expr="//field[@name='order_line']/tree/field[@name='warehouse_id']"
+                position="attributes"
+            >
+                <attribute name="invisible">0</attribute>
+                <attribute
+                    name="attrs"
+                >{'readonly': [('parent.state', '!=', 'draft')], 'column_invisible': [('parent.show_sale_line_warehouse_column', '=', 'False')]}</attribute>
```
alexey-pelykh, line 40:
```xml
'column_invisible': [('parent.show_sale_line_warehouse_column', '=', 'False')]
```

`show_sale_line_warehouse_column` is a `Boolean` field, but it is compared to the **string** `'False'`. In Odoo 16 this will not evaluate correctly.

Should be:
```xml
'column_invisible': [('parent.show_sale_line_warehouse_column', '=', False)]
```

sale_warehouse_rule/tests/test_sale_warehouse_rule.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2023 Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests import Form, SavepointCase
+
+
+class TestSaleWarehouseRule(SavepointCase):
```
alexey-pelykh, line 8: `SavepointCase` has been deprecated since Odoo 15. For 16.0, use `TransactionCase`:

```python
from odoo.tests import Form, TransactionCase

class TestSaleWarehouseRule(TransactionCase):
```

sale_warehouse_rule/views/res_config_settings.xml
```diff
@@ -0,0 +1,24 @@
...
+            >
+                <div class="col-xs-12 col-md-6 o_setting_box">
+                    <div class="o_setting_left_pane">
+                        <field name="show_sale_line_warehouse_column" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label
+                            for="show_sale_line_warehouse_column"
+                            string="Afficher la colonne entrepôt sur les lignes de commande"
+                        />
```
alexey-pelykh, line 18: The label text is in French:
```xml
string="Afficher la colonne entrepôt sur les lignes de commande"
```

OCA modules must use English for all view strings. Use:
```xml
string="Show Warehouse Column on Sale Order Lines"
```
and add the French translation to `fr.po`.

alexey-pelykh (changes requested):
Thank you for this module -- the concept of per-product warehouse routing rules is useful. I found several issues that should be addressed before merging.

### Summary of findings

| # | File | Severity | Finding |
|---|------|----------|----------|
| 1 | `models/product.py` | ⛔ blocker | Hardcoded French string `"Entrepôt par défaut"` as field label; OCA requires English |
| 2 | `models/product.py:25` | ⛔ blocker | Deprecated `force_company_id` context key (already flagged by @rousseldenis) |
| 3 | `models/sale_order.py:77` | ⛔ blocker | `_()` wraps a dynamically-built string (`+ ", ".join(...)`) -- the full string is never in `.po`, so translation is impossible |
| 4 | `views/sale_order.xml:35` | ⛔ blocker | `column_invisible` compares Boolean to string `'False'` instead of `False` -- column visibility will not work correctly |
| 5 | `models/sale_order.py:83-84` | ⛔ blocker | `action_confirm` assigns a multi-record recordset to `rec.warehouse_id` (Many2one) when `warehouse_rule_need_change` is truthy but lines have mixed warehouse rules -- will crash |
| 6 | `models/sale_warehouse_rule.py:57-82` | ⚠️ major | Uniqueness constraint mutates domain by index and does not exclude `self` (`("id", "!=", rule.id)` is missing), risking false positives |
| 7 | `tests/test_sale_warehouse_rule.py:6` | ⚠️ minor | `SavepointCase` is deprecated since Odoo 15; use `TransactionCase` for 16.0 |
| 8 | `views/res_config_settings.xml:14` | ⚠️ minor | Settings label is in French; should be English with translation in `.po` |
| 9 | `i18n/` | ⚠️ minor | Missing `.pot` template file (`i18n/sale_warehouse_rule.pot`) |
| 10 | `models/sale_order.py:72-74` | 💡 nit | `variant_warehouse_id` is a Many2one (always 0 or 1 record) -- `.filtered()` on it is technically correct but misleading |

See inline comments for details.

sale_warehouse_rule/models/sale_order.py
rousseldenis: You should use translation function arguments to pass parameters

metaminux:
I updated OCA/product-attribute#2022 and renamed the model `attribute.value.dependant.mixin` to `attribute.value.dependent.mixin`

So I added a fixup commit to change the dependency here.

I also added a `[DON'T MERGE] test-requirements.txt` commit to make the dependency on unreleased PR OCA/product-attribute#2022

metaminux:
Hello,

As OCA/product-attribute#2022 has been merged, I remove the "temporary dependency on unreleased PR" commit

Waiting for other checks to end...

sale_warehouse_rule/models/product.py
metaminux: Done ✅

sale_warehouse_rule/models/product.py
metaminux: Done ✅

sale_warehouse_rule/models/sale_order.py
metaminux: Done ✅

sale_warehouse_rule/models/sale_order.py
metaminux: Done ✅

sale_sourced_by_line_product_rule/models/sale_line_product_rule.py
```diff
@@ -0,0 +1,66 @@
...
+        for rule in self:
+            if rule.product_id:
+                rule.applied_on = "0_product"
+            elif rule.attribute_value_ids:
+                rule.applied_on = "1_attribute"
+            else:
+                rule.applied_on = "2_template"
+
+    @api.constrains("product_id", "attribute_value_ids", "warehouse_id")
+    def _check_warehouse_rule_uniqueness(self):
```
metaminux: Done ✅

sale_warehouse_rule/views/sale_order.xml
metaminux: Done ✅

sale_warehouse_rule/tests/test_sale_warehouse_rule.py
metaminux: Done ✅

sale_sourced_by_line_product_rule/views/res_config_settings.xml
```diff
@@ -0,0 +1,24 @@
...
+            >
+                <div class="col-xs-12 col-md-6 o_setting_box">
+                    <div class="o_setting_left_pane">
+                        <field name="show_sale_line_warehouse_column" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label
+                            for="show_sale_line_warehouse_column"
+                            string="Afficher la colonne entrepôt sur les lignes de commande"
+                        />
```
metaminux: Done ✅

sale_warehouse_rule/models/product.py
metaminux: Done ✅

metaminux:
Just added a fixup commit to take into account what @rousseldenis and @alexey-pelykh suggested

Now, one test is failing in `sale_sourced_by_line` because `sale_warehouse_rule` is computing warehouse_id on the second line and as no rules are defined, it resets the warehouse to the one defined on the first line, so only one delivery order is created when `sale_sourced_by_line` is expecting 2...

Don't know what to do there...

Open for your reviews and advices.

rousseldenis:
In fact, this is a derivation of `sale_sourced_by_line`. The computation is different but the aim is the same.

rousseldenis:
So, maybe depending on that one and renaming this `sale_sourced_by_line_product_rule` or ...

metaminux:
> So, maybe depending on that one and renaming this `sale_sourced_by_line_product_rule` or ...

Thanks @rousseldenis I renamed the module `sale_sourced_by_line_product_rule` by adding a `pre_init_hook`, it now depends on `sale_sourced_by_line` and few other changes.
CI is now fully green, if everyone's OK with that, I'll squash the 3 commits and we can merge...

metaminux (approved):

alexey-pelykh (approved):
Thanks for the substantial rework and the rename to `sale_sourced_by_line_product_rule`. Re-reviewed at `cbd4c83` — all of my earlier blocking points are resolved:

- English field/label strings — `models/product.py:12`, `views/res_config_settings.xml:17`
- `force_company_id` dropped for `self.env.company` + `@api.depends_context("allowed_company_ids")` — `models/product.py:22-25`
- Translatable message no longer concatenated — `models/sale_order.py:52-55` uses a named `%(warehouses)s` placeholder
- `column_invisible` now compares to `False`, not `'False'` — `views/sale_order.xml:40`
- `action_confirm` guards the Many2one assignment with `len(warehouse_ids) == 1` — `models/sale_order.py:57-62`
- Uniqueness constraint excludes self (`("id","!=",rule.id)`) and builds explicit per-case domains — `models/sale_line_product_rule.py:44-49`
- Test migrated to `TransactionCase` — `tests/test_sale_line_product_rule.py:8`

CI is green. LGTM. (Nit, non-blocking: the `.pot` still carries the old `sale_warehouse_rule` name after the rename — the OCA export will regenerate it.)

PR #3809, opened 2025-07-10
[16.0][IMP] sale_elaboration: set products profile from its categories

Andrii9090-tecnativa:
I added new functionality to propagate the elaboration profile from the category to the products. When you set the elaboration profile in the category, all its products will have this profile.

MT-10786 @moduon @yajo @rafaelbn @Shide @chienandalu @u0f please review if you want :smile:


sale_elaboration/models/product_elaboration_profile.py
```diff
@@ -18,3 +18,8 @@ class ProductElaborationProfile(models.Model):
         column1="profile_id",
         column2="elaboration_id",
     )
+    product_category_ids = fields.One2many(
+        "product.category",
+        inverse_name="elaboration_profile_id",
+        string="Product Category",
+    )
```
chienandalu: No need to overload the model with this relation
```suggestion
```

sale_elaboration/models/product_template.py
```diff
@@ -14,13 +14,20 @@ class ProductTemplate(models.Model):
...
-                template.product_variant_ids.elaboration_profile_id
-            )
+            if not template.elaboration_profile_id and template.categ_id:
+                template.elaboration_profile_id = (
+                    template.categ_id.elaboration_profile_id
+                )
+            else:
+                template.elaboration_profile_id = (
+                    template.product_variant_ids.elaboration_profile_id
+                )
```
chienandalu: I feel this shouldn't be necessary at all. This should be triggered whenever the variant profile is set.

sale_elaboration/models/product_template.py
```diff
@@ -33,7 +40,7 @@ def _inverse_elaboration_profile_id(self):
 
     @api.model_create_multi
     def create(self, vals_list):
-        templates = super(ProductTemplate, self).create(vals_list)
+        templates = super().create(vals_list)
```
chienandalu: Although it's correct. Let's minimize the diff to the feature changes
```suggestion
        templates = super(ProductTemplate, self).create(vals_list)
```

sale_elaboration/tests/test_sale_elaboration.py
```diff
@@ -11,9 +11,25 @@ class TestSaleElaboration(AccountTestInvoicingCommon):
...
         )
+        cls.product_2 = cls.env["product.product"].create(
+            {"name": "test 2", "tracking": "none", "list_price": 1000}
+        )
+        cls.product_3 = cls.env["product.product"].create(
+            {"name": "test 2", "tracking": "none", "list_price": 1000}
+        )
+        cls.product.categ_id = cls.child_category
+        cls.product_2.categ_id = cls.child_category
+        cls.product_3.categ_id = cls.parent_category
```
chienandalu: I think you can play with `cls.product_elaboration_A`  and `cls.product_elaboration_B` so you don't need to declare that many records.

Then, in your test cases, you can assign them to the categories to avoid modifiying the existing cases behavior.

sale_elaboration/tests/test_sale_elaboration.py
```diff
@@ -197,3 +226,63 @@ def test_propagation_from_sale_order_to_stock_move(self):
...
+                "elaboration_profile_id": self.elaboration_profile_a.id,
+            }
+        )
+        self.assertNotEqual(
+            product.elaboration_profile_id.id,
+            self.child_category.elaboration_profile_id.id,
+        )
+        self.assertEqual(
+            product.elaboration_profile_id.id, self.elaboration_profile_a.id
+        )
```
chienandalu: Overall, document the workflow cases we're testing with either docstrings, comments or assert fail strings.

A comment like this helps a lot to understand what's being tested:
```
# We're setting the profile in category A: only products in that category will be affected

File: sale_elaboration/views/product_category_views.xml [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+        model="ir.ui.view"
+        id="sale_elaboration_product_category_view_list_inherit_elaboration_id"
+    >
+        <field name="name">product.category.view.list.sale.elaboration</field>
+        <field name="model">product.category</field>
+        <field name="inherit_id" ref="product.product_category_list_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//tree" position="attributes">
+                <attribute name="multi_edit">1</attribute>
+            </xpath>
```
chienandalu: Add a comment on why you're doing this

sale_elaboration/models/product_category.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    elaboration_profile_id = fields.Many2one(
+        comodel_name="product.elaboration.profile", string="Elaboration Profile"
+    )
+
+    def write(self, vals):
+        """Override update to set the elaboration profile on products."""
```
chienandalu: Don't use docstrings on overriden methods as the docstring is overriden as well. Then when you inspect the method you'll get the last overriden string.

Use a regular comment instead and be concise avoiding the most obvious information.
```suggestion
        # Set the elaboration profile on the category products it apply just to the products with the exact category"""
```

sale_elaboration/models/product_category.py
```diff
@@ -0,0 +1,22 @@
...
+        result = super().write(vals)
+        if "elaboration_profile_id" in vals:
+            products = self.env["product.product"].search(
+                [("categ_id", "in", self.ids)]
+            )
+            products.elaboration_profile_id = vals["elaboration_profile_id"]
```
chienandalu: Indeed, you already have it in your recordset as `super` is already run, so this should be fine as well:
```suggestion
            products.elaboration_profile_id = self.elaboration_profile_id
```

chienandalu (changes requested):
Thanks @Andrii9090 !  Some comments below:

- The title of the PR should be: [16.0][IMP] sale_elaboration: add new functionality to propagate the elaboration profile from the category to the product
- You can use a more concise title anyway: [16.0][IMP] sale_elaboration: set products profile from its categories
- Check tests fails.

sale_elaboration/models/product_elaboration_profile.py
Andrii9090-tecnativa: All atended

sale_elaboration/models/product_template.py
Andrii9090-tecnativa: This is required to propagate the elaboration profile from the category to the product when the category is changed. It works as expected when the elaboration profile hasn’t been set yet

sale_elaboration/models/product_template.py
Andrii9090-tecnativa: All atended

sale_elaboration/tests/test_sale_elaboration.py
Andrii9090-tecnativa: All atended.

sale_elaboration/tests/test_sale_elaboration.py
Andrii9090-tecnativa: I have added comments

sale_elaboration/views/product_category_views.xml
Andrii9090-tecnativa: I have added comment

sale_elaboration/models/product_category.py
Andrii9090-tecnativa: All atended.

sale_elaboration/models/product_category.py
Andrii9090-tecnativa: All atended

sale_elaboration/models/product_category.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    elaboration_profile_id = fields.Many2one(
+        comodel_name="product.elaboration.profile", string="Elaboration Profile"
+    )
+
+    def write(self, vals):
```
chienandalu: You should deal with the create as well

sale_elaboration/readme/CONFIGURE.rst
```diff
@@ -8,3 +8,12 @@ To configure this module you need to:
...
    elaborations on Picking Operations.
+
+To configure elaboration profile for all products in a category:
+
+#. Go to *Inventory > Configuration > Product Categories* and select one or you 
+can add a column of Elaboration Profiles in the Product Categories list view.
+#. Set the elaboration profile you want to use for all products in this category. 
+It is used in the case if the category hasn't child category. When you 
+create/update a product, the elaboration profile is set from the product's 
+category if the elaboration profile isn't yet setted.
```
chienandalu: Keep some idention so it's easier to read it in code:
```suggestion
#. Go to *Inventory > Configuration > Product Categories* and select one or you 
   can add a column of Elaboration Profiles in the Product Categories list view.
#. Set the elaboration profile you want to use for all products in this category. 
   It is used in the case if the category hasn't child category. When you 
   create/update a product, the elaboration profile is set from the product's 
   category if the elaboration profile isn't yet setted.
```

sale_elaboration/readme/CONFIGURE.rst
```diff
@@ -8,3 +8,12 @@ To configure this module you need to:
    notes on Delivery Slips* if you want to show elaborations on Delivery Slips
    or *Display Elaboration notes on Picking Operations* if you want to show
    elaborations on Picking Operations.
+
+To configure elaboration profile for all products in a category:
+
+#. Go to *Inventory > Configuration > Product Categories* and select one or you 
+can add a column of Elaboration Profiles in the Product Categories list view.
```
chienandalu: Add that in an extra comment after this block like:  You can also use multi-edit in the list view to set the profile in multiple categories at once. Just unhide it from the dots in the right.
```suggestion
#. Go to *Inventory > Configuration > Product Categories* and select one.
```

sale_elaboration/readme/CONFIGURE.rst
chienandalu: Set this aside:

When you create product or update its category , the elaboration profile is set from the product's 
category. Anyway, it's possible to change it afterwards.


```suggestion
The profiles is just set in the category products and it won't be propagated to the child category products.
```

sale_elaboration/models/product_template.py
```diff
@@ -14,13 +14,20 @@ class ProductTemplate(models.Model):
...
-            template.elaboration_profile_id = (
-                template.product_variant_ids.elaboration_profile_id
-            )
+            if not template.elaboration_profile_id and template.categ_id:
+                template.elaboration_profile_id = (
+                    template.categ_id.elaboration_profile_id
+                )
+            else:
+                template.elaboration_profile_id = (
+                    template.product_variant_ids.elaboration_profile_id
```
chienandalu: careful here: there could be many variants. You should ensure you don't try to set multiple elaboration profiles

chienandalu (changes requested):
Some more small changes ;)

sale_elaboration/models/product_template.py
chienandalu: My comment isn't correct, sorry :)

Andrii9090-tecnativa:
This short video highlights the newest improvements to the module. Thank you @yajo :smile: 


https://github.com/user-attachments/assets/0f164bfc-a05f-48bc-9c9f-c8874a69081b

chienandalu (approved):
I can't see the logic. Maybe you scrambled your commits? :sweat_smile:

sale_elaboration/views/product_views.xml
```diff
@@ -14,7 +14,6 @@
             </field>
         </field>
     </record>
-
```
yajo: **issue:** Please disable the automatic formatter that you probably have in your IDE for XML files, and that removes these lines. Then, leave them as they were.

These lines are not wrong. Just let `prettier` (the one in pre-commit) do the formatting. This way we avoid distractions when reviewing these unneeded stylistic changes.

yajo (changes requested):

chienandalu (commented):
@Andrii9090 I changed the orientation of your changes for the most simple logic that gets to the original goal.

In a summary

- The module behaves as it was until now.
- When we set a elaboration profile in the category and no profile is set in the product, that will be the default. Otherwise, no profile is applied and all the elaborations are available.
- Sadly, it's hard to test this with the `test.Form` tools (I'll do some research) as the domain is evaluated in the view and it seems that it doesn't care about it at that level... O left some of your original test case declarations... although the go with no effect, so maybe I remove them for the sake of commit clarity...

@rafaelbn @sergio-teruel all yours if you want to review :slightly_smiling_face:

rafaelbn (approved):
Tested functionaly 👍🏼 

2 minutes video with a full explanation

https://www.loom.com/share/d0570c9362ce47e6abd853135aac6807?sid=8ae7d7f6-80f9-4df0-8cc7-85999139cedf

fcvalgar (approved):
Functional review, 

LGTM, thank you @Andrii9090

rafaelbn:
Please @chienandalu port to 17 and 18 in the same task 😄 ❤️ Thank you!

PR #3813, opened 2025-07-14
[18.0][ADD]sale_order_client_order_ref_mandatory

DantePereyra:
This module, makes `client_order_ref` field required and reposition it,for both form view and Sale reports.Also makes this field searchable and add an optional column on both quotation and sale orders lists.
SALE ORDER BEFORE
client_sale_order_ref_before_image
SALE ORDER AFTER
client_sale_order_ref_after_image
SALE REPORT TEMPLATE BEFORE
client_sale_order_ref_template_before_image
SALE REPORT TEMPLATE AFTER
client_sale_order_ref_template_after_image


rousseldenis:
@DantePereyra Could you check tests ?

DantePereyra:
> @DantePereyra Could you check tests ?

@rousseldenis, tests are failing because when the `client_order_ref` field was made mandatory in the view, it caused conflicts in addons whose tests called the` sale.order` form without that field being completed. That’s why I added the code snippet that detects if the environment is in test mode and auto-fills the field.

However, a new error has now appeared in the test:
```bash
2025-07-15 08:11:27,282 333 ERROR odoo odoo.addons.sale_require_po_doc.tests.test_sale_require_po_doc: FAIL: TestSaleRequirePODoc.test_require_customer_need_po
  File "/__w/sale-workflow/sale-workflow/sale_require_po_doc/tests/test_sale_require_po_doc.py", line 39, in test_require_customer_need_po
    with self.assertRaises(ValidationError, msg=messsage):
  File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/opt/odoo/odoo/tests/common.py", line 484, in _assertRaises
    with ExitStack() as inner:
  File "/usr/lib/python3.10/contextlib.py", line 576, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.10/contextlib.py", line 561, in __exit__
    if cb(*exc_details):
AssertionError: ValidationError not raised : You can not confirm sale order without Customer reference.
```
The validation is not triggered because the field is already filled in. I'm a bit stuck at this point—if you have any suggestions on how to move forward, it would be helpful. Thanks!

sale_order_client_order_ref_mandatory/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    client_order_ref = fields.Char(
+        string="Customer Reference",
+        copy=False,
+        default=lambda self: self._default_test_dummy_name(),
+    )
+
+    def _default_test_dummy_name(self):
```
rousseldenis: @DantePereyra Making code depending on test behaviour is not recommended

sale_order_client_order_ref_mandatory/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    client_order_ref = fields.Char(
+        string="Customer Reference",
+        copy=False,
+        default="/",
+    )
```
alexey-pelykh, line 13: Several issues with this field redefinition:

1. **`default="/"`** -- This is problematic. Setting a default of `"/"` means every new sale order will have `"/"` as the customer reference. This defeats the entire purpose of making the field mandatory, because the default value satisfies the `required` constraint in the view. Users can confirm orders without ever changing the reference. If the goal is to force users to provide a *real* customer reference, the default should be removed and the `required` should be enforced in the Python model.

2. **The field is not actually required in the model.** The module name says "mandatory" but `required=True` is not set here -- it is only set in the XML view (`required="1"`). This means the constraint is only enforced in the UI form view. Sale orders created via API, imports, automated actions, or other code paths will bypass this entirely. To make it truly mandatory, add `required=True` in the field definition here.

3. **Redundant attributes.** `string="Customer Reference"` and `copy=False` are already defined on `client_order_ref` in the base `sale.order` model. Re-declaring them here is unnecessary and creates a maintenance burden if the base ever changes.

sale_order_client_order_ref_mandatory/views/sale_order_view.xml
```diff
@@ -0,0 +1,72 @@
...
+    <record id="view_order_form" model="ir.ui.view">
+        <field
+            name="name"
+        >sale.order.form (in sale_order_client_order_ref_mandatory)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='client_order_ref']" position="attributes">
+                <attribute name="invisible">1</attribute>
+                <attribute name="required">1</attribute>
```
alexey-pelykh, line 11: Setting `required="1"` on a field that is simultaneously made `invisible="1"` is contradictory. The original field position is hidden, so the `required` attribute here has no practical effect (the field won't be visible for user input). The `required` enforcement should come from the Python model definition (`required=True`) rather than from the view. You can remove `<attribute name="required">1</attribute>` from this xpath since the field is invisible and the requirement should be in the model layer.

sale_order_client_order_ref_mandatory/views/sale_order_view.xml
```diff
@@ -0,0 +1,72 @@
...
+            </xpath>
+            <xpath expr="//group[@name='sale_header']" position="before">
+                <group>
+                    <h2>
+                        <field
+                            name="client_order_ref"
+                            required="1"
+                            nolabel="1"
```
alexey-pelykh, line 19: The new field placement looks good for the UX, but note that `required="1"` here only protects the form view. API/import creation of sale orders will bypass this. The requirement should be enforced at the model level.

sale_order_client_order_ref_mandatory/report/sale_order_report_templates.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
+    <template
+        id="report_saleorder_document"
+        inherit_id="sale.report_saleorder_document"
+    >
+        <xpath expr="//div[@id='informations']" position="before">
+            <h3>
+                <span t-field="doc.client_order_ref" />
```
alexey-pelykh, line 8: If `client_order_ref` has the default value `"/"` (as defined in the model), this will render `"/"` as a visible `<h3>` heading on the printed sale order report, which is not meaningful to the customer. Consider adding a `t-if` condition like `t-if="doc.client_order_ref and doc.client_order_ref != '/'"` -- or better yet, remove the `default="/"` from the model entirely so this is not needed.

sale_order_client_order_ref_mandatory/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "version": "18.0.1.0.0",
+    "category": "Sale",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": ["sale"],
+    "data": [
+        "views/sale_order_view.xml",
+        "report/sale_order_report_templates.xml",
+    ],
+    "installable": True,
+}
```
alexey-pelykh, line 19: The module is missing a `tests` directory. OCA guidelines expect modules to include tests. At a minimum, please add:
- A test that creating a sale order without `client_order_ref` raises a validation error (to verify the mandatory constraint works)
- A test that the field is properly displayed in the report

This is especially important for a module whose entire purpose is enforcing a constraint.

alexey-pelykh (changes requested):
Hi, thanks for the contribution! I reviewed the module and have several concerns that I think should be addressed before this can move forward.

**Summary of findings:**

| # | Severity | Finding |
|---|----------|---------||
| 1 | Critical | Field is not truly mandatory -- `required` is only enforced at the view level, not in the model |
| 2 | Critical | `default="/"` defeats the mandatory purpose and will leave meaningless values on orders |
| 3 | Major | No tests included |
| 4 | Moderate | Redundant field attribute re-declarations (`copy=False`, `string`) |
| 5 | Moderate | Setting `required=1` on the original field that is simultaneously made `invisible=1` |

Please see inline comments for details.

rousseldenis:
@DantePereyra What's the status of this ?

DantePereyra:
Hi @rousseldenis, it's on stale as there were problems with other addons and test, I have to check it again.

PR #3818, opened 2025-07-16
[ADD][16.0] sale_delivery_split_free_product

Shide:
Create new procurement to split free products into separate delivery orders

MT-10882 @moduon @rafaelbn @chienandalu @yajo @fcvalgar @Gelojr please review if you want 😄


sale_delivery_split_free_product/readme/CONTEXT.md
```diff
@@ -0,0 +1,5 @@
+This module was developed because sometimes we don't want our stock movements to be merged if free products are sold in separate order lines.
```
rousseldenis, line 1: @Shide Maybe a little explanation on the `why`. Because, it seems to me weird that you have with this additional effort in warehouse operations to treat (by the operator) the different lines for the same product. No ?

Shide:
@rousseldenis this pr was made to allow manage free product lines separately because in this PR: https://github.com/OCA/stock-logistics-workflow/pull/2014 I didn't see a way to manage returns (or lowering que quantity on the SO line). 
Thanks to @EmilioPascual, this module is not necessary right now.

I'm going to close this PR in favor of https://github.com/OCA/stock-logistics-workflow/pull/2014

PR #3810, opened 2025-07-10
[18.0][MIG] sale_payment_sheet: Migration to 18.0

victoralmau:
Migration to 18.0

Please @sergio-teruel and @pilarvargas-tecnativa can you review it?

@Tecnativa TT54295

PR #3821, opened 2025-07-17
[18.0][MIG] sale_planner_calendar: Migration to 18.0

victoralmau:
Migration to 18.0

Locked by:
- [x] sale_payment_sheet: https://github.com/OCA/sale-workflow/pull/3810

Please @CarlosRoca13 and @pilarvargas-tecnativa can you review it?

@Tecnativa TT54296


sale_planner_calendar/views/sale_planner_calendar_event_view.xml
```diff
@@ -0,0 +1,628 @@
...
+                                            />
+                                            <button
+                                                name="action_open_rating"
+                                                type="object"
+                                                invisible="sale_planner_state=='cancel'"
+                                                title="Change calendar event status by adding a previous rating"
+                                                t-att-class="'btn p-0 px-1 btn-lg text-capitalize ' + (record.sale_planner_state.raw_value == 'done' and 'btn-success' or 'btn-info')"
+                                            >
+                                                <span
+                                                    class="badge pb-2"
```
pilarvargas-tecnativa:
```suggestion
                                                    class="badge pt-2"
```

sale_planner_calendar/views/sale_planner_calendar_event_view.xml
```diff
@@ -0,0 +1,628 @@
...
+                                                    t-esc="record.sale_planner_state.raw_value"
+                                                />
+                                                <span
+                                                    class="h3"
+                                                    t-esc="record.sale_planner_rating.value"
```
pilarvargas-tecnativa, line 397:
```suggestion
   t-if="record.sale_planner_rating.value" t-esc="record.sale_planner_rating.value"
```

pilarvargas-tecnativa (changes requested):

pilarvargas-tecnativa (approved):

alexey-pelykh (commented):
## Review: [18.0][MIG] sale_planner_calendar

The migration structure looks solid overall — `__manifest__.py` version is correctly set to `18.0.1.0.0`, the pre-migration script properly uses `openupgradelib` to load `noupdate_changes.xml`, and the OWL components follow the `.esm.js` pattern expected for Odoo 18.0.

Below are findings from the review, ordered by severity.

---

### Bug: `_inverse_hour` uses `self` instead of `rec` in loop

**File:** `sale_planner_calendar/models/calendar_event.py`, method `_inverse_hour`

```python
def _inverse_hour(self):
    for rec in self:
        duration = rec.duration
        date = self._get_hour_tz_offset()  # <-- should be rec._get_hour_tz_offset()
```

The `_compute_hour` method correctly uses `rec._get_hour_tz_offset()`, but the inverse uses `self._get_hour_tz_offset()`. Since `_get_hour_tz_offset()` reads `self.start`, this will always use the first record's `start` for all records in a multi-record write. Should be `rec._get_hour_tz_offset()` for consistency and correctness.

---

### Bug: ValidationError message formatting

**File:** `sale_planner_calendar/models/sale_planner_calendar_summary.py`, `_check_date_user_id_type` constraint

```python
raise ValidationError(
    self.env._(
        "Already exists a summary with same user, date and event type)"
        "Access with 'Sale planner calendar summary' menu option"
    )
)
```

Two issues here:
1. Stray `)` at the end of the first string literal (inside the message text).
2. No separator between the two implicitly concatenated strings — the resulting message reads `...event type)Access with...` with no space or newline.

Suggested fix:
```python
raise ValidationError(
    self.env._(
        "Already exists a summary with same user, date and event type.\n"
        "Access with 'Sale planner calendar summary' menu option"
    )
)
```

---

### License inconsistency in `res_config_settings.py`

**File:** `sale_planner_calendar/models/res_config_settings.py`

This file has an LGPL-3.0 header while all other module files (and the manifest) use AGPL-3.0. Should be aligned to AGPL-3.0.

---

### Commented-out code

Two instances of commented-out code that should be removed for cleanliness:

1. **`calendar_event.py`:** `# dynamic_end_date = fields.Date(compute="_compute_dynamic_end_date")`
2. **`sale_planner_calendar_summary.py`**, `action_done` method: `# "comment": 'Done',`

---

### Code organization: `ResCompany` in `res_config_settings.py`

**File:** `sale_planner_calendar/models/res_config_settings.py`

This file defines both `ResConfigSettings` and `ResCompany` classes. OCA convention generally prefers one model per file. Consider moving `ResCompany` to a separate `res_company.py` (and updating `__init__.py`).

---

### Translation file headers reference Odoo 15.0

**Files:** `sale_planner_calendar/i18n/*.po`

The `.po` file headers contain `Odoo Server 15.0` — these should reference 18.0 for the migration. This is typically auto-generated, but worth updating.

---

### Observations (non-blocking)

- **Pre-existing typos** in field names (`susbscriptions_backward_days`, `unsuscribe_old_salesperson`) — these appear to be carried from prior versions. Fixing them would require additional migration scripts, so understandable to keep as-is for the migration PR. Worth tracking for a follow-up.
- **`action_mass_deletion`** is referenced in the JS controllers but not defined in this module's Python code — I assume this is provided by Odoo core's `calendar` module. If so, no issue; just noting for visibility.
- **`test-requirements.txt`** references PR #3810 for `sale_payment_sheet` — expected while the dependency PR is still open.

---

Overall the migration is well-structured. The `_inverse_hour` self/rec issue and the ValidationError message are the main items I'd suggest addressing. The rest are minor improvements.

Reviewed-by: CorporateHub

CarlosRoca13 (approved):

PR #3824, opened 2025-07-22
[16.0][IMP] sale_blanket_order: lines with price 0

LauraCForgeFlow:
In some business cases, it makes sense to have some blanket order lines with price 0 (maybe they are gifted articles, provided as part of a service package...). We remove the restriction to be able to do so.

### DIFF
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -660,9 +660,6 @@ class BlanketOrderLine(models.Model):
         try:
             for line in self:
                 assert (
-                    not line.display_type and line.price_unit > 0.0
-                ) or line.display_type, _("Price must be greater than zero")
-                assert (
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
I agree with you!

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
Fair enough

AlexPForgeFlow (approved):
LGTM!

sale_blanket_order/models/blanket_orders.py
```diff
@@ -659,9 +659,6 @@ def _compute_quantities(self):
     def _validate(self):
         try:
             for line in self:
-                assert (
```
rousseldenis, line 662: Ok, even if in an Odoo context, using derived from 'UserError' exceptions is better.

rousseldenis (approved):

PR #3827, opened 2025-07-23
[15.0][ADD] sale_semaphore: New module to be alerted of unit prices set on sale order

CarlosRoca13:
cc @Tecnativa TT57153


sale_semaphore/models/product_category.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    discount_success = fields.Float()
```
rousseldenis:
```suggestion
    semaphore_discount_success = fields.Float()
```

and so on.

sale_semaphore/models/semaphore_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SemaphoreMixin(models.AbstractModel):
+    _name = "semaphore.mixin"
+    _description = "Semaphore mixin"
+
+    max_price_success = fields.Float(compute="_compute_semaphore_max_prices")
```
rousseldenis:
```suggestion
    semaphore_max_price_success = fields.Float(compute="_compute_semaphore_max_prices")
```

sale_semaphore/models/semaphore_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+            ("dark", "Dark"),
+            ("success", "Success"),
+            ("warning", "Warning"),
+            ("danger", "danger"),
+        ],
+        compute="_compute_semaphore",
+        store=True,
+    )
+
+    def _get_uom_field(self):
```
rousseldenis:
```suggestion
    @api.model
    def _get_uom_field(self):
```

sale_semaphore/models/semaphore_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+        return "product_uom" if self._name == "sale.order.line" else "product_uom_id"
+
+    def _get_semaphore_dependencies(self):
```
rousseldenis:
```suggestion
    @api.model
    def _get_semaphore_dependencies(self):
```

rousseldenis (commented):
Interesting module.

Wouldn't have it be great to externalize semaphore mixin in a `price_semaphore` (as it is oriented to prices ones)?

CarlosRoca13:
Hi @rousseldenis, thank you for the feedback :smile: . We're still working through this development and will take your comment into account. I'm going to mark the PR as a draft since it's not finished yet.

sale_semaphore/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module adds a commercial semaphore to provide visual feedback on the
+prices being offered. It also adds a widget to allow selection of the
```
rousseldenis: It should be great to explain in which manner as it seems not clear to me at this point.

sale_semaphore/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["sale"],
+    "data": [
+        "views/account_invoice_report_views.xml",
+        "views/account_move_views.xml",
+        "views/product_category_views.xml",
+        "views/product_views.xml",
+        "views/sale_order_views.xml",
```
alexey-pelykh, line 18: This PR targets the `15.0` branch, but Odoo 15.0 is end-of-life. OCA generally does not merge new modules into EOL branches. Consider porting to 16.0 or 17.0.

sale_semaphore/models/sale_order.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+from odoo.tools import float_compare
+
+
+class SaleOderLine(models.Model):
```
alexey-pelykh, line 9: Typo: `SaleOderLine` should be `SaleOrder` (this class inherits `sale.order`, not `sale.order.line`, and "Oder" is missing the 'r').

sale_semaphore/models/sale_order_line.py
```diff
@@ -0,0 +1,136 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools import float_compare
+
+
+class SaleOderLine(models.Model):
+    _inherit = "sale.order.line"
```
alexey-pelykh, line 9: Typo: `SaleOderLine` -- missing 'r' in "Order".

sale_semaphore/models/sale_order_line.py
```diff
@@ -0,0 +1,136 @@
...
+                float_compare(
+                    record.semaphore_max_price_danger,
+                    record.price_reduce,
+                    precision_digits=dp,
+                )
+                > 0
+            )
+
+    def write(self, vals):
+        res = super().write(vals)
```
alexey-pelykh, line 108: Bug: `super().write(vals)` is called on line 108, but the semaphore price validation happens after (lines 109-136). If the validation fails, the write has already been applied. The validation check should occur *before* the `super().write()` call, or the entire block should be restructured so the error prevents persistence.

```suggestion
    def write(self, vals):
        dp = self.env["decimal.precision"].precision_get("Product Price")
        if ("price_unit" in vals or "discount" in vals) and not self.env.user.has_group(
            "sales_team.group_sale_manager"
        ):
            res = super().write(vals)
            lines_below_semaphore = self.filtered(
                lambda line: line.price_below_semaphore
                and line.state in ["sale", "done"]
            )
            if lines_below_semaphore:
                lines_price_below_pricelist = lines_below_semaphore.filtered(
                    lambda l: float_compare(
                        l.price_reduce,
                        l.order_id.pricelist_id.get_product_price(
                            l.product_id, l.product_uom_qty, l.order_id.partner_id
                        ),
                        precision_digits=dp,
                    )
                    < 0
                )
                if lines_price_below_pricelist:
                    raise UserError(
                        _(
                            "There's a line with price below semaphore's accepted.\n\n"
                            "Please set the prices in a way that they are accepted by the "
                            "semaphore, or contact the purchasing administrators."
                        )
                    )
            return res
        return super().write(vals)
```

Note: Even with this restructuring, the `raise UserError` after `super().write()` will still roll back the transaction in Odoo since it is within the same cursor transaction. However, logically it is cleaner to validate before persisting when possible.

sale_semaphore/static/src/semaphore/semaphore.esm.js
```diff
@@ -0,0 +1,220 @@
...
+                    method: "get_reset_price_unit",
+                    args: [this.record.res_id],
+                });
+                this.trigger_up("field_changed", {
+                    dataPointID: this.record.id,
+                    changes: {price_unit: price_unit_val, discount: 0},
+                    onSuccess: () => {
+                        const $tr = this.$el.closest("tr");
+                        $tr.find("td[name='price_unit']").text(
+                            formatMonetary(this.recordData.price_unit_val, {
```
alexey-pelykh, line 192: Bug: `this.recordData.price_unit_val` does not exist. The RPC result is stored in local variable `price_unit_val` (line 181). This should be:
```js
formatMonetary(price_unit_val, {
```

sale_semaphore/models/sale_order_line.py
```diff
@@ -0,0 +1,136 @@
...
+        vals = super()._prepare_invoice_line(**optional_values)
+        vals["semaphore"] = self.semaphore
+        return vals
+
+    @api.depends("price_reduce")
+    def _compute_price_below_semaphore(self):
+        dp = self.env["decimal.precision"].precision_get("Product Price")
+        self.price_below_semaphore = False
+        for record in self.filtered(lambda line: line.product_id.semaphore_active):
+            record.price_below_semaphore = (
```
alexey-pelykh, line 98: Inconsistency: This filters only on `line.product_id.semaphore_active`, but `_get_semaphore_data()` in `product.py` also considers `self.categ_id.semaphore_active`. Lines where the semaphore is active at category level (but not product level) will never be flagged as `price_below_semaphore`.

Consider using `line.semaphore_active` instead, which is computed from `_compute_semaphore` and correctly accounts for both product and category activation.

alexey-pelykh (changes requested):
Thank you for this module. The concept of a price semaphore for commercial alerting is useful. However, I found several issues that need to be addressed before this can be merged.

**Critical issues:**

1. **Target branch is 15.0** -- Odoo 15.0 has reached end-of-life. New modules should target a supported version (e.g., 16.0 or 17.0). OCA typically no longer merges new features into EOL branches.

2. **Bug in `write()` validation** (`sale_order_line.py`): `super().write(vals)` is called before the price validation check. If the price is below the semaphore threshold, the `UserError` is raised *after* the write has already been applied to the database. The validation should happen *before* calling `super().write()`, or alternatively, the write should be wrapped so the error prevents the commit.

3. **JS bug in reset handler** (`semaphore.esm.js`): In the `reset_value` click handler's `onSuccess` callback, `this.recordData.price_unit_val` is referenced, but `price_unit_val` is a local variable from the `rpc.query` result, not a property of `recordData`. This will display `undefined` instead of the reset price.

4. **No tests**: This module contains substantial business logic (semaphore computation, blocking order confirmation, price thresholds per product/category) but has zero test files. OCA guidelines require tests for new modules.

**Other findings:**

5. **Typo in class names**: Both `sale_order.py` and `sale_order_line.py` define the class as `SaleOderLine` (missing 'r' in "Order"). While this does not affect Odoo ORM behavior (it uses `_inherit`), it is confusing -- especially in `sale_order.py` where the class inherits `sale.order`, not `sale.order.line`.

6. **Inconsistency in `_compute_price_below_semaphore`**: The filter uses `line.product_id.semaphore_active` but `_get_semaphore_data()` also checks `self.categ_id.semaphore_active`. This means lines where the semaphore is configured at the *category* level but not the *product* level will never be flagged as below semaphore.

7. **Previous reviewer feedback unaddressed**: @rousseldenis suggested extracting a semaphore mixin and improving the description. Those comments appear to still be open.

I would recommend porting this to a supported Odoo version and addressing the bugs before requesting another review.

rousseldenis:
@CarlosRoca13 What's the status of this ?

CarlosRoca13:
I have added tests and improved the readme. I think it is ready to be merged :smile: 

carlosdauden (approved):

sergio-teruel (approved):

PR #3828, opened 2025-07-23
[FIX] sale_partner_selectable_option: Don't allow to create or edit orders when a partner not selectable is added.

CarlosRoca13:
With this change we avoid creating orders with partners that are not selectable. Example: default_partner_id is set on context

cc @Tecnativa TT56709

ping @victoralmau @sergio-teruel

### DIFF
--- a/sale_partner_selectable_option/models/sale_order.py
+++ b/sale_partner_selectable_option/models/sale_order.py
@@ -3,11 +3,22 @@
 
 from lxml import etree
 
-from odoo import api, models
+from odoo import _, api, models
+from odoo.exceptions import UserError
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
+
+    @api.constrains("partner_id")
+    def _check_partner_id(self):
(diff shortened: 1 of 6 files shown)


sale_partner_selectable_option/models/sale_order.py
```diff
@@ -3,12 +3,21 @@
...
+from odoo.exceptions import UserError
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    @api.constrains("partner_id")
+    def _check_partner_id(self):
+        for order in self:
+            if not order.partner_id.sale_selectable:
```
sergio-teruel: Mybe you can add a skip key context to avoid this restriction in some background process like as import data..

CarlosRoca13: Done @sergio-teruel

victoralmau (approved):

sergio-teruel (approved):

PR #3820, opened 2025-07-17
[16.0][IMP] sale_order_safe_commitment_date: fill delivery date

chienandalu:
The delivery date is computed by default from the expected date. We can still edit it, but if can't be delivered, it will be force on confirmation.

cc @moduon @rafaelbn @Gelojr  @Shide   MT-9790

PR #3831, opened 2025-07-25
[16.0][FIX] sale_order_safe_commitment_date: Validation without checking time

EmilioPascual:
- In validation to show warning doesn't check time, only date, because by this way you have a safe commitment date.
- Remove deactivation of view by default, the field commitment date should be always in header because the module update the field and it's important to module behavior.

https://www.loom.com/share/0d04a268a5ce4559a406cc2bee79e451?sid=dee2d17d-1f81-496d-9244-c7ed54baaf40

@chienandalu @rafaelbn @yajo please review
    
@moduon MT-9790


rafaelbn:
Hello dear,

Problems detected:

- [x] The view with the Delivery Date (commitment_date) in the main form (header) is archived. We should by default activate it
- [x] The salesman set the Delivery Date in the first moment because is one of the most important fields to fill. So we cannot calculated automatically this fields, as it overwrite the salesman's data.

We can meet to think about it 😄 

With love

chienandalu (commented):
Is this intended?:

image

EmilioPascual:
> Is this intended?:
> 
> image

Yes, it's intended to resolve an error in the commit https://github.com/OCA/sale-workflow/pull/3820/commits/eca8ba391f7f29036494c406f522679ff4240041 that was merged too hastily.

rafaelbn:
@EmilioPascual is this Pr still WIP? Or maybe we should close it?

PR #3832, opened 2025-07-25
[17.0][FIX] sale_advance_payment: posting many invoices did not reconcile advance payments

dreispt:
Previous logic only supported a one by one invoice posting.
With this change a set of invoices can be posted at the same time, and
the advance payments will be appplied properly.

### DIFF
--- a/sale_advance_payment/models/account_move.py
+++ b/sale_advance_payment/models/account_move.py
@@ -7,15 +7,19 @@ from odoo import models
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    def action_post(self):
+    def _post(self, soft=True):
         # Automatic reconciliation of payment when invoice confirmed.
-        res = super().action_post()
-        sale_order = self.mapped("line_ids.sale_line_ids.order_id")
-        if sale_order and self.invoice_outstanding_credits_debits_widget is not False:
-            json_invoice_outstanding_data = (
-                self.invoice_outstanding_credits_debits_widget.get("content", [])
+        res = super()._post(soft=soft)
+        for invoice in self:
+            # Get Advance Payment Account Moves
+            sale_orders = invoice.mapped("line_ids.sale_line_ids.order_id")
+            advance_payment_moves = sale_orders.account_payment_ids.move_id
+            # Get reconcilable payments JSON data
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (approved):
code LGTM

sale_advance_payment/models/account_move.py
```diff
@@ -7,15 +7,19 @@
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    def action_post(self):
+    def _post(self, soft=True):
```
alexey-pelykh, line 10: The `__manifest__.py` file needs a version bump for this bugfix. The current version on the `17.0` branch is `17.0.1.0.5`, so this should be bumped to `17.0.1.0.6`.

Also, a minor note: it would be valuable to add a test covering the multi-invoice posting scenario (posting a recordset of multiple invoices at once) to validate this fix and prevent regressions.

alexey-pelykh (changes requested):
Thank you for this fix! The core logic change -- iterating per-invoice and overriding `_post` instead of `action_post` -- is correct and well-reasoned. In Odoo 17, `_post` is the proper hook point for batch posting scenarios.

However, the PR is missing a version bump in `__manifest__.py`. Per OCA guidelines, bugfix PRs should increment the patch segment (current base is `17.0.1.0.5`, so this should become `17.0.1.0.6`).

**Summary of changes:**
- Overrides `_post(soft=True)` instead of `action_post()` -- correct, since `_post` is the lower-level method called during batch posting
- Iterates per-invoice (`for invoice in self:`) instead of operating on the full recordset -- fixes the multi-record bug since `invoice_outstanding_credits_debits_widget` is a per-record computed field
- Pre-filters reconcilable lines using `filter()` to only advance payment moves -- cleaner logic, equivalent behavior

**Suggestion:** A test covering the multi-invoice posting scenario would help prevent regressions, though I understand CI is currently green with the existing test suite.

rousseldenis (approved):

sale_advance_payment/models/account_move.py
rousseldenis: @alexey-pelykh This is not a correct review as ocabot will bump version.

PR #3836, opened 2025-07-28
[FIX] sale_packaging_default: Don't use modulo operator to check if qty is a multiple

EmilioPascual:
We do not use the modulo operator to check if qty is a mltiple of q. Indeed the quantity per package might be a float, leading to incorrect results. For example:
  - 8 % 1.6 = 1.5999999999999996
  - 5.4 % 1.8 = 2.220446049250313e-16 https://github.com/odoo/odoo/blob/16.0/addons/product/models/product_packaging.py#L47

@yajo @rafaelbn please review

@moduon MT-10950


yajo (approved):
Amazing, thanks!

Functional & code reviews. Fixes the problem!

**suggestion:** add a test please.

EmilioPascual:
> **suggestion:** add a test please.

Add new test. Thank you

yajo (approved):
Excellent! 🏆 

/ocabot merge patch

PR #3837, opened 2025-07-28
[18.0][IMP] sale_exception: sale_exception_show_popup configurable

rlimaeco:
This PR includes changes to the company sale configuration, allowing the exception wizard to be optionally shown to the user. This way, the user doesn't have to confirm multiple times and can rely on the warning blocking message displayed on the sale order form.

### DIFF
--- a/sale_exception/tests/test_sale_exception.py
+++ b/sale_exception/tests/test_sale_exception.py
@@ -155,7 +155,28 @@ class TestSaleException(TransactionCase):
         sale_order.action_confirm()
         partner.sale_warn = "warning"
         sale_order2 = sale_order.copy()
-        sale_order2.detect_exceptions()
+        self.env.company.sale_exception_show_popup = True
+        result = sale_order2.action_confirm()
+        self.assertEqual(
+            result.get("xml_id"), "sale_exception.action_sale_exception_confirm"
+        )
+        self.assertEqual(sale_order2.state, "draft")
+        self.assertTrue(sale_order2.exception_ids.filtered(lambda x: x == exception))
+
+    def test_exception_partner_sale_warning_no_popup(self):
+        exception = self.env.ref("sale_exception.exception_partner_sale_warning")
+        exception.active = True
+        partner = self.env.ref("base.res_partner_1")
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        partner.sale_warn = "warning"
+        sale_order2 = sale_order.copy()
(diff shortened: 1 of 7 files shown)


sale_exception/views/res_config_settings.xml
```diff
@@ -0,0 +1,21 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="res_config_settings_view_form_sale_exception" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="sale.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <setting id="order_warnings" position="after">
+                <setting
+                    id="sale_exception_show_popup"
+                    help="Display popup at confirm or simply rely on the red alert already visible on top of the order."
```
simahawk, line 12: is this help msg repetition necessary?

sale_exception/wizard/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Camptocamp SA
```
simahawk: this normally goes to `models` instead of `wizard`

simahawk (changes requested):
LG. Minor remarks.
Would be nice to have a test (you can take an existing one as an example).

twalter-c2c (approved):
LG in general, remarks from @simahawk to be addressed. Pre-approving.

sale_exception/views/res_config_settings.xml
rlimaeco: I will let only in this xml file this help string so

sale_exception/tests/test_sale_exception.py
```diff
@@ -157,6 +157,29 @@ def test_exception_partner_sale_warning(self):
...
+        sale_order = self._create_sale_order(
+            partner=partner, product=self.env.ref("product.product_product_6")
+        )
+        sale_order.action_confirm()
+        partner.sale_warn = "warning"
+        sale_order2 = sale_order.copy()
+        sale_order2.detect_exceptions()
+        self.assertTrue(sale_order2.exception_ids.filtered(lambda x: x == exception))
+        self.env.company.sale_exception_show_popup = False
+        result = sale_order2.action_confirm()
```
simahawk, line 177: in both tests instead of calling `detect_exceptions` we can directly call `action_confirm`.
Not a blocker.

simahawk (approved):
LG

leemannd (approved):

rousseldenis (approved):

PR #3838, opened 2025-07-30
[18.0][MIG] sale_order_invoicing_finished_task: Migration to 18.0

victoralmau:
Migration to 18.0

Please @carlos-lopez-tecnativa and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT57321


sale_order_invoicing_finished_task/models/project.py
```diff
@@ -0,0 +1,68 @@
...
+    def toggle_invoiceable(self):
+        self._check_sale_line_state()
+        for task in self:
+            task.invoiceable = not task.invoiceable
+
+    def write(self, vals):
+        if "sale_line_id" in vals:
+            self._check_sale_line_state(vals["sale_line_id"])
+        res = super().write(vals)
+        return res
```
carlos-lopez-tecnativa:
```suggestion
        return super().write(vals)
```

sale_order_invoicing_finished_task/models/project.py
```diff
@@ -0,0 +1,68 @@
...
+            if (
+                sale_line.state == "cancel"
+                or sale_line.invoice_status == "invoiced"
+                or sale_line.order_id.locked
+            ):
+                raise ValidationError(
+                    _(
+                        "You cannot create/modify a task related with a "
+                        "invoiced, locked or cancel sale order line "
+                    )
```
carlos-lopez-tecnativa:
```suggestion
                    self.env._(
                        "You cannot create/modify a task related with a "
                        "invoiced, locked or cancel sale order line "
                    )
```

sale_order_invoicing_finished_task/views/product_view.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field
+            name="inherit_id"
+            ref="sale_project.product_template_form_view_invoice_policy_inherit_sale_project"
+        />
+        <field name="arch" type="xml">
+            <field name="service_tracking" position="after">
+                <field
+                    name="invoicing_finished_task"
+                    invisible="(type != 'service') or (service_tracking not in ['task_global_project', 'task_in_project'])"
+                />
```
carlos-lopez-tecnativa:
```suggestion
                    invisible="type != 'service' or service_tracking not in ['task_global_project', 'task_in_project']"
                />
```

sale_order_invoicing_finished_task/views/project_view.xml
```diff
@@ -0,0 +1,43 @@
...
+        </field>
+    </record>
+    <record id="edit_project_task_track" model="ir.ui.view">
+        <field name="name">project.task.form.track</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <xpath expr="/form" position="inside">
+                <field name="invoicing_finished_task" invisible="1" />
+            </xpath>
```
carlos-lopez-tecnativa: I think in V18 it is not necessary to add the fields as invisible; Odoo adds this automatically.
```suggestion
```

sale_order_invoicing_finished_task/views/project_view.xml
```diff
@@ -0,0 +1,43 @@
...
+            <div name="button_box" position="inside">
+                <field name="invoiceable" invisible="1" />
```
carlos-lopez-tecnativa, line 22: The same here
```suggestion
```

sale_order_invoicing_finished_task/views/project_view.xml
```diff
@@ -0,0 +1,43 @@
...
+                <button
+                    name="toggle_invoiceable"
+                    type="object"
+                    invisible="not 'invoicing_finished_task'"
```
carlos-lopez-tecnativa: This expression is incorrect; you are evaluating it as a string instead of using the field value. Please review.
```suggestion
                    invisible="not invoicing_finished_task"
```

sale_order_invoicing_finished_task/models/project.py
```diff
@@ -0,0 +1,68 @@
...
+    @api.onchange("stage_id")
+    def _onchange_stage_id(self):
+        tasks = self.filtered(
+            lambda t: (
+                t.invoicing_finished_task
+                and t.stage_id.invoiceable
+                and not t.invoiceable
+            )
+        )
+        tasks.toggle_invoiceable()
```
carlos-lopez-tecnativa, line 33: I'm not sure if my comment is applicable or if I misunderstood the behavior.
I set a stage as invoiceable — for example, "Done". When I move the task to that stage, it is marked as invoiceable, which is correct.
However, if I move the task to another stage that is not invoiceable, the task remains marked as invoiceable. I expected the task to be updated and marked as not invoiceable.
Could you please check and confirm if this is the expected behavior or if it's a mistake?

image

carlos-lopez-tecnativa (changes requested):

sale_order_invoicing_finished_task/views/project_view.xml
victoralmau: Specifically, this field should not be removed from view to avoid incorrect behavior in tests with `Form()`. I have added the comment to prevent it from being accidentally deleted in the future.

sale_order_invoicing_finished_task/models/project.py
victoralmau: I understand what you're saying, but it's intentional.
The idea is that the stage change applies “towards billable,” not from billable to something “non-billable.”
Imagine that you have already changed the stage to "_Invoiceable_", the order status has changed, etc., and then you change the stage back to “_In progress_” because you are still working on it (for a few days, for example), but that should not change the `invoiceable` field again and the implications that would have on the sales order and billing status.

carlos-lopez-tecnativa: Ok, I understand. Thanks for explaining it to me.

carlos-lopez-tecnativa (approved):
LGTM

juancarlosonate-tecnativa (approved):
Everything looks fine, the only thing is that the usage images might need to be updated or removed

PR #3839, opened 2025-07-30
[16.0][FIX] sale_order_invoicing_finished_task: Change methods that do not exist and/or should not be used

victoralmau:
Change methods that do not exist and/or should not be used

Changes done:
- [x] Change the `_get_to_invoice_qty()` method (no longer exists) to `_compute_qty_to_invoice()`
- [x] Do not use the `_get_invoiced()` method; replace it with `_compute_invoice_status()`
- [x] Make tests more readable + use `Form()` to avoid manually calling onchange methods

Although the `_get_invoiced()` method exists, the name is confusing; it is actually a compute. The appropriate and most readable approach is to use the appropriate `invoice_status` value in each line of the order, and the order status will update itself.

Similar to https://github.com/OCA/sale-workflow/pull/3838

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT57321

### DIFF
--- a/sale_order_invoicing_finished_task/models/sale_order.py
+++ b/sale_order_invoicing_finished_task/models/sale_order.py
@@ -1,26 +1,9 @@
 # Copyright 2017 Tecnativa - Sergio Teruel
 # Copyright 2017 Tecnativa - Carlos Dauden
+# Copyright 2025 Tecnativa - Víctor Martínez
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

rousseldenis (approved):
LGTM

PR #3772, opened 2025-06-25
[17.0][IMP] sale_delivery_state: add delivery status field to search views

IriaAlonso:
With this improvement a filter for pending deliveries is added. A group by this field is added as well.

PR #3843, opened 2025-07-31
[18.0][IMP] sale_delivery_state: add delivery status field to search views

IriaAlonso:
With this improvement a filter for pending deliveries is added. A group by this field is added as well.
FWP of https://github.com/OCA/sale-workflow/pull/3772

### DIFF
--- a/sale_delivery_state/views/sale_order_views.xml
+++ b/sale_delivery_state/views/sale_order_views.xml
@@ -46,4 +46,30 @@
             </field>
         </field>
     </record>
+
+    <record
+        id="sale_order_view_search_inherit_sale_inherit_delivery_status"
+        model="ir.ui.view"
+    >
+        <field name="name">sale.order.search</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.sale_order_view_search_inherit_sale" />
+        <field name="arch" type="xml">
+            <filter name="order_date" position="after">
+                <separator />
+                <filter
+                    string="Pending Deliveries"
+                    name="filter_delivery_status_pending"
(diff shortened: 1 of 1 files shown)


IriaAlonso:
@astirpe @rousseldenis @yostashiro Can you review it, please?

sale_delivery_state/views/sale_order_views.xml
```diff
@@ -46,4 +46,30 @@
...
+        <field name="name">sale.order.search</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.sale_order_view_search_inherit_sale" />
+        <field name="arch" type="xml">
+            <filter name="order_date" position="after">
+                <separator />
+                <filter
+                    string="Pending Deliveries"
+                    name="filter_delivery_status_pending"
+                    domain="[('delivery_status','in',['pending', 'partial'])]"
```
yostashiro: We might as well add `started` here? See https://github.com/OCA/sale-workflow/commit/3836d15d0d245f7f9dc5cb3912b0fea6b3e4a83f.

```suggestion
                    domain="[('delivery_status','in',['pending', 'started', 'partial'])]"
```

IriaAlonso: Should it be added even though that value never appears, for now?

yostashiro: I think so. IIUC, the delivery status will become `started` if a picking is done in a multi-step scenario. In any case, it won't do any harm to add it.

rousseldenis (approved):

rousseldenis:
@yostashiro

yostashiro (approved):
LGTM. 👍

PR #3844, opened 2025-07-31
[18.0][IMP] sell_only_by_packaging: warning when qty not pack multiple

rlimaeco:
The validation that checks whether the quantity on a sales order line is a multiple of the package quantity was present in v14. This PR restores this functionality to raise a warning on sale order.

### DIFF
--- a/sell_only_by_packaging/models/sale_order_line.py
+++ b/sell_only_by_packaging/models/sale_order_line.py
@@ -3,7 +3,7 @@
 
 from odoo import api, models
 from odoo.exceptions import ValidationError
-from odoo.tools import float_compare
+from odoo.tools import float_compare, float_round
 
 
 class SaleOrderLine(models.Model):
@@ -16,6 +16,9 @@ class SaleOrderLine(models.Model):
         errors = []
         for line in self:
             if not line.product_uom_qty:
+                continue
+            if error_message_qty_multiple := line._check_pkg_qty_multiple():
+                errors.append(error_message_qty_multiple)
                 continue
             if not line.product_id.sell_only_by_packaging:
                 continue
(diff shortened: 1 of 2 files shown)


mmequignon:
I'm not sure this should be the responsibility of this module to ensure this.

sell_only_by_packaging/models/sale_order_line.py
```diff
@@ -13,26 +13,71 @@ class SaleOrderLine(models.Model):
...
+                continue
+            if not line.product_id.sell_only_by_packaging:
+                continue
+            if error_message_min_qty := line._check_min_qty_packaging():
+                errors.append(error_message_min_qty)
+        if errors:
+            raise ValidationError(
+                self.env._(
+                    "The following lines has some issues with "
+                    + "packaging quantities:\n    - %s",
```
twalter-c2c, line 31:
```suggestion
                    "The following lines have some issues with "
                    "packaging quantities:\n    - %s",
```

sell_only_by_packaging/models/sale_order_line.py
```diff
@@ -13,26 +13,71 @@ class SaleOrderLine(models.Model):
...
+                    "\n    - ".join(errors),
```
twalter-c2c, line 32: Isn't it duplicated in the line above?

twalter-c2c (commented):
Technically, looks good. I have not tested, though. I also agree that this probably should not be included in this module.

sell_only_by_packaging/models/sale_order_line.py
rlimaeco: the line above is only for the first item in the list, for the others the "-" using join will do the job

mmequignon:
@simahawk is this ok, or should we made a sale exception out of this ?

rlimaeco:
Closed in favor of:

- https://github.com/OCA/product-attribute/pull/2038

PR #2759, opened 2023-11-02
[16.0][MIG] sale_stock_reconcile_valuation_kit: Migration to 16.0

PauBForgeFlow:

PR #3845, opened 2025-08-01
[18.0][MIG] sale_stock_reconcile_valuation_kit: Migration to 18.0

Kimkhoi3010:
Migrate `sale_stock_reconcile_valuation_kit` to 18.0
## Depend on:
- https://github.com/OCA/sale-workflow/pull/2759

## Change:
In Odoo 17.0, the field quantity_done has been renamed to quantity on the stock.move model.
You can see it in [this commit](https://github.com/odoo/odoo/commit/7dda6bb92715ea25b2818a62fec5e646f3678b81#diff-55c6314416a6a400da6acd5018d161a55eeeb0e3008fec8828121e3dd12be0ebL157) for reference


sale_stock_reconcile_valuation_kit/i18n/sale_stock_reconcile_valuation_kit.pot
```diff
@@ -0,0 +1,34 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* sale_stock_reconcile_valuation_kit
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
+"Report-Msgid-Bugs-To: \n"
+"Last-Translator: \n"
```
alexey-pelykh, line 9: Minor: This `.pot` file appears to be copied from 14.0 without regeneration -- `Project-Id-Version` still says `Odoo Server 14.0`, and the `__last_update` field entry (line 33-34) references a field removed in Odoo 17. Consider regenerating with `oca-gen-addon-readme` / the standard OCA i18n export tooling.

sale_stock_reconcile_valuation_kit/models/account_move.py
```diff
@@ -0,0 +1,64 @@
...
+                        sl_product_account_moves += product_stock_moves.mapped(
+                            "account_move_ids.line_ids"
+                        ).filtered(
+                            lambda line,
+                            account=product_interim_account: line.account_id == account
+                            and not line.reconciled
+                        )
+                        # Reconcile.
+                        sl_product_account_moves.reconcile()
+        return
```
alexey-pelykh, line 64: Nit: This bare `return` is unnecessary -- the method returns `None` implicitly. Removing it would be slightly cleaner, but not blocking.

alexey-pelykh (approved):
Thank you for the migration to 18.0!

The core logic looks correct -- the lambda closure variable binding improvements are good practice, and the test correctly uses `quantity` instead of the removed `quantity_done` field. CI passes.

A few observations:

**Minor items:**

1. The `.pot` file still says `Project-Id-Version: Odoo Server 14.0` and references `__last_update` which was removed in Odoo 17. This should ideally be regenerated for 18.0, though OCA CI tooling may handle it.

2. The bare `return` at the end of `_stock_account_anglo_saxon_reconcile_valuation` (line 64 of `account_move.py`) is unnecessary -- the method returns `None` implicitly. Very minor, not blocking.

3. The copyright header in `__manifest__.py` says `LGPL-3.0` but the `license` field is `AGPL-3`. This is inherited from 14.0 and pre-existing, but worth noting for eventual cleanup.

Overall the migration is straightforward and well done. Tests are a welcome addition (the 14.0 version had none).

PR #3846, opened 2025-08-01
[17.0][FIX] sale_product_multi_add: not pull price from pricelists

khaihoan123:
price_unit field already have auto compute, if we set it with list_price, compute wouldn't be triggered

### DIFF
--- a/sale_product_multi_add/__manifest__.py
+++ b/sale_product_multi_add/__manifest__.py
@@ -7,7 +7,7 @@
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "category": "Sale Management",
-    "version": "17.0.1.0.1",
+    "version": "17.0.1.0.2",
     "license": "AGPL-3",
     "depends": ["sale"],
     "data": [
--- a/sale_product_multi_add/wizards/sale_import_products.py
+++ b/sale_product_multi_add/wizards/sale_import_products.py
@@ -41,7 +41,6 @@ class SaleImportProducts(models.TransientModel):
                 "product_id": item.product_id.id,
                 "product_uom_qty": item.quantity,
                 "product_uom": item.product_id.uom_id.id,
-                "price_unit": item.product_id.list_price,
             }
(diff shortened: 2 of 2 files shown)


rousseldenis (approved):
LGTM

khaihoan123:
> LGTM

@rousseldenis Could you merge this pull, please?

alexey-pelykh (commented):
Good catch. In Odoo 17, setting `price_unit` explicitly to `list_price` bypasses the `_compute_price_unit` mechanism on `sale.order.line`, which means pricelist rules, partner-specific pricing, and discounts configured on the sale order are ignored. Removing this line allows the standard compute to run and resolve the correct price from the pricelist.

The fix is correct and minimal.

One thing to address: the `__manifest__.py` version should be bumped from `17.0.1.0.1` to `17.0.1.0.2` per OCA guidelines (patch-level bump for a bug fix).

khaihoan123:
> Good catch. In Odoo 17, setting `price_unit` explicitly to `list_price` bypasses the `_compute_price_unit` mechanism on `sale.order.line`, which means pricelist rules, partner-specific pricing, and discounts configured on the sale order are ignored. Removing this line allows the standard compute to run and resolve the correct price from the pricelist.
> 
> The fix is correct and minimal.
> 
> One thing to address: the `__manifest__.py` version should be bumped from `17.0.1.0.1` to `17.0.1.0.2` per OCA guidelines (patch-level bump for a bug fix).

[alexey-pelykh](https://github.com/alexey-pelykh) i updated code, could you help me merge this

sale_product_multi_add/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "category": "Sale Management",
-    "version": "17.0.1.0.1",
+    "version": "17.0.1.0.2",
```
rousseldenis, line 10: @khaihoan123 Could you juste remove this as ocabot will update it itself ?

PR #3850, opened 2025-08-04
[18.0][IMP] sale_product_set: improve product set search behavior

MohamedOsman7:
This PR improves the user experience when selecting product sets in forms by enhancing the name_search behavior. Product sets are now easier to find through the standard search field.

Before:
Searching for a product set by keywords did not always return relevant results unless "Search more..." was used.

After:
The name_search method is now extended to improve matching, making relevant product sets appear directly in the autocomplete suggestions.

### DIFF
--- a/sale_product_set/README.rst
+++ b/sale_product_set/README.rst
@@ -46,21 +46,20 @@ updated or removed as any other sales order lines.
 Usage
 =====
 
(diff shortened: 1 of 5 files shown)


MohamedOsman7:
@juancarlosonate-tecnativa @CarlosRoca13 Could you please have a look at this PR?

PR #3851, opened 2025-08-06
[16.0][IMP] sale_order_safe_commitment_date: delivery date helper

chienandalu:
When the lead time has decimals (for example, 1.5 = 36 hours) a lot of false positives are raised for the delivery exceptions. as the default hour for the delivery date is the current one. For example: in that case if the current hour is 01:00, the expected date will be 13:00 for the next day, but when we set the delivery date for tomorrow, the default hour will be tomorrow at 1:00, which is before that expected day. The users tend to ignore the warnings after stumbling upon these cases over and over.

So now:

- There's a new delivery date helper field to simply set the date (no time) that sets the hour to 23.59.59 for the delivery date (commitment_date)
- The delivery date can be still changed manually whenever that info is important as usal in the standard.

cc @moduon MT-9790

please review @rafaelbn @EmilioPascual

### DIFF
--- a/sale_order_safe_commitment_date/models/sale_order.py
+++ b/sale_order_safe_commitment_date/models/sale_order.py
@@ -1,26 +1,61 @@
 # Copyright 2025 Moduon Team S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from datetime import datetime, time
+
+import pytz
+
 from odoo import api, fields, models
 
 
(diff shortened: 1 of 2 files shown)


EmilioPascual (approved):
@chienandalu good job, it seems to fix the unusual behavior.

Functional and code review

rafaelbn (approved):
/ocabot merge minor

Let see test

Funcional Ok, we need to re-think for Odoo 18

PR #3857, opened 2025-08-11
[18.0][IMP] sale_elaboration: elaboration profile from product category

chienandalu:
fw of

- #3809 

Elaboration profiles can now be set from the category so if no profile is set in the product it defaults to that one.

TODO:

- `test.Form` doesn't constrain the view domain for profiles... so the only way to test the flow would be either to use a helper domain field or a tour... Some tooling left in the test case anyway.

cc @moduon MT-10786

please review if you feel like it :slightly_smiling_face:  @rafaelbn @sergio-teruel @Andrii9090


rafaelbn (approved):
Tested in runbot 👍🏼 , tested as 16 (already merged) and 17

fcvalgar (approved):
Functional review, 
LGTM, thank you @chienandalu .

PR #3861, opened 2025-08-13
[14.0][IMP] sale_exception: SO line inherit change to base.exception

raumschmiede-joshuaL:
Hi.

Because of my other PR https://github.com/OCA/server-tools/pull/3346 I changed `sale_exception` to use more of `base_exception`. Some fields are added again to `sale.order.line` eventhough they already exist on `base.exception`.

With these changes the note "(Is blocking)" now also appears in the exceptions summary on the SO line. Previously not as `sale.order.line` defined its own computation for this field.
And with the new inherit it has now the field `main_exception_id`.

Here I did more changes than in other modules that depend on `base_exception` to remove the duplicated stuff

### DIFF
--- a/sale_exception/migrations/14.0.1.2.0/pre-migration.py
+++ b/sale_exception/migrations/14.0.1.2.0/pre-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2025 Raumschmiede GmbH
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tools.sql import column_exists, create_column
+
+
+def migrate(cr, version):
+    if column_exists(cr, "sale_order_line", "main_exception_id"):
+        return
+
+    create_column(cr, "sale_order_line", "main_exception_id", "int4")
--- a/sale_exception/models/sale_order_line.py
+++ b/sale_exception/models/sale_order_line.py
@@ -1,50 +1,16 @@
 # © 2019 Akretion
+# Copyright 2025 Raumschmiede GmbH
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-import html
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class SaleOrderLine(models.Model):
-    _inherit = ["sale.order.line", "base.exception.method"]
+    _inherit = ["sale.order.line", "base.exception"]
     _name = "sale.order.line"
 
-    exception_ids = fields.Many2many(
-        "exception.rule", string="Exceptions", copy=False, readonly=True
-    )
-    exceptions_summary = fields.Html(
-        readonly=True, compute="_compute_exceptions_summary"
-    )
-    ignore_exception = fields.Boolean(
-        related="order_id.ignore_exception", store=True, string="Ignore Exceptions"
-    )
-
-    @api.depends("exception_ids", "ignore_exception")
-    def _compute_exceptions_summary(self):
-        for rec in self:
-            if rec.exception_ids and not rec.ignore_exception:
-                rec.exceptions_summary = rec._get_exception_summary()
-            else:
-                rec.exceptions_summary = False
-
-    def _get_exception_summary(self):
-        return "<ul>%s</ul>" % "".join(
-            [
-                "<li>%s: <i>%s</i></li>"
-                % tuple(map(html.escape, (e.name, e.description)))
-                for e in self.exception_ids
-            ]
-        )
+    ignore_exception = fields.Boolean(related="order_id.ignore_exception")
 
     def _get_main_records(self):
         return self.mapped("order_id")
-
-    def _detect_exceptions(self, rule):
-        records = super()._detect_exceptions(rule)
-        # Thanks to the new flush of odoo 13.0, queries will be optimized
-        # together at the end even if we update the exception_ids many times.
-        # On previous versions, this could be unoptimized.
-        (self - records).exception_ids = [(3, rule.id)]
-        records.exception_ids = [(4, rule.id)]
-        return records.mapped("order_id")
--- a/sale_exception/models/sale_order_line.py
+++ b/sale_exception/models/sale_order_line.py
@@ -40,10 +40,6 @@ class SaleOrderLine(models.Model):
     def _get_main_records(self):
         return self.mapped("order_id")
 
-    @api.model
-    def _reverse_field(self):
-        return "sale_ids"
-
     def _detect_exceptions(self, rule):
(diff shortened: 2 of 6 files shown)


raumschmiede-joshuaL:
@mt-software-de, @jbaudoux, @i-vyshnevska

sale_exception/models/sale_order_line.py
```diff
@@ -1,41 +1,16 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-import html
 
 from odoo import api, fields, models
 
 
 class SaleOrderLine(models.Model):
-    _inherit = ["sale.order.line", "base.exception.method"]
+    _inherit = ["sale.order.line", "base.exception"]
```
mt-software-de, line 10: By changing the `_inherit` to `base.exception` is it than still needed to have `_detect_exceptions` in `sale.order.line`?

mt-software-de (changes requested):
Thx. One minor remark. 
Could you please add the open PR as temporary reference as it is described here. https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

sale_exception/models/sale_order_line.py
raumschmiede-joshuaL: Yes, still needed.
`detect_exceptions` on the SO line is still triggered by the SO. Now with the new `_get_sub_exception_field_names`. But because SO line returns its SO in `_get_main_records` and `_reverse_field`, it will assign the exception that was detected to the SO.
`_detect_exceptions` here assigns the exceptions to the SO line.
Something that should be changed, too. But in general that something like `sale_ids` is required on the exception rule is something that could be reworked. Or how `detect_exceptions` works. Too much for here

raumschmiede-joshuaL: With PR https://github.com/OCA/server-tools/pull/3365 the `_detect_exceptions` override can be removed. I will update this here soon

raumschmiede-joshuaL:
@rousseldenis, can you check this here? This relates to the other PRs I created for repo server-tools.

The 2nd commit with the most changes could be separated by the PRs I mentioned in the commit message. But I didn't do it because the differences would only be in the unit tests to check for the blocking text

raumschmiede-joshuaL:
On a database with a lot of sale order lines the computation of `main_exception_id` will take a long time.  I will add a migration script to add this field directly to the table without computation. The exceptions are already on `exception_ids`, and the field is not displayed. No need to set it in the script

sale_exception/models/sale_order_line.py
```diff
@@ -1,54 +1,16 @@
...
-
-    def _get_exception_summary(self):
-        return "<ul>%s</ul>" % "".join(
-            [
-                "<li>%s: <i>%s</i></li>"
-                % tuple(map(html.escape, (e.name, e.description)))
-                for e in self.exception_ids
-            ]
-        )
+    ignore_exception = fields.Boolean(related="order_id.ignore_exception")
```
alexey-pelykh, line 14: Minor note: the original `ignore_exception` field had `store=True` explicitly set:
```python
ignore_exception = fields.Boolean(
    related="order_id.ignore_exception", store=True, string="Ignore Exceptions"
)
```

The new definition defaults to `store=False` in Odoo 14 for related fields. This means the `ignore_exception` column on `sale_order_line` table will effectively be dropped after this migration.

This is likely intentional (the value is always derived from the parent order, and the tree view `decoration-danger` evaluates per-record so it still works), but worth being explicit about since it's a schema change that could affect custom search domains or reports filtering on `sale.order.line` `ignore_exception` directly.

Consider either:
- Adding `store=True` back if backward compatibility is important, or
- Documenting this as an intentional schema change in the PR description

sale_exception/migrations/14.0.1.2.0/pre-migration.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tools.sql import column_exists, create_column
+
+
+def migrate(cr, version):
+    if column_exists(cr, "sale_order_line", "main_exception_id"):
+        return
+
+    create_column(cr, "sale_order_line", "main_exception_id", "int4")
```
alexey-pelykh, line 11: Good approach -- creating the column in pre-migration avoids the ORM trying to compute the field for all existing records during module update. The `column_exists` guard is a nice touch for idempotency.

alexey-pelykh (commented):
## Review by alexey-pelykh (CorporateHub)

Good improvement -- removing duplication by leveraging `base.exception` inheritance directly on `sale.order.line` instead of the lighter `base.exception.method` is the right direction. The migration script is properly structured and the test additions verify the blocking exception summary.

### Overall Assessment

The code changes themselves are sound. The main concern is the dependency chain and one behavioral change worth confirming.

### Findings

| # | Severity | Finding |
|---|----------|----------|
| 1 | Note | **Dependency chain**: This PR depends on OCA/server-tools#3346 (still open) and references #3365. The `test-requirements.txt` correctly uses the temporary PR reference per OCA conventions. Just flagging that this cannot be merged until the upstream dependency lands. |
| 2 | Minor | **`ignore_exception` field lost `store=True`**: The original field had `store=True` on the related field. The new definition `fields.Boolean(related="order_id.ignore_exception")` defaults to `store=False` in Odoo 14. While `base.exception` defines `ignore_exception` as a regular stored field, the `sale.order.line` override makes it a non-stored related field. This works functionally (the tree view's `decoration-danger` evaluates per-record), but it changes the DB schema -- the column will be dropped. If any custom search domains or SQL queries rely on this column on `sale_order_line`, they would break. This is likely intentional (since the value is always derived from the parent order), but worth confirming explicitly. |
| 3 | Note | **`_sale_get_lines` removal**: The method `_sale_get_lines` was removed from `sale.order`. If any downstream module or customization calls this method, it would break. Appears to be unused by the base module itself, but worth noting for dependent modules. |
| 4 | Note | **`sale_ids` M2M field removal from `exception.rule`**: This reverse relation is removed because the new `base_exception` handles sub-exception detection differently (via `_get_sub_exception_field_names`). Downstream modules that query `exception_rule.sale_ids` would break. Again, this is expected given the upstream refactoring. |

All actual test runs (Odoo and OCB) pass. Pre-commit passes. The "Detect unreleased dependencies" CI failure is expected due to the temporary git reference in `test-requirements.txt`.

Looks good overall. The changes are clean and well-structured.

rousseldenis:

PR #3863, opened 2025-08-14
[16.0][FIX] sale_advance_payment: fix invoice paid amount

JordiMForgeFlow:
- Filter out credit notes and cancelled invoices
- Use invoice currency amounts instead of company currency amounts

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -50,6 +50,8 @@ class SaleOrder(models.Model):
         "account_payment_ids.move_id.line_ids.credit",
         "account_payment_ids.move_id.line_ids.currency_id",
         "account_payment_ids.move_id.line_ids.amount_currency",
+        "invoice_ids",
+        "invoice_ids.state",
         "invoice_ids.amount_residual",
     )
     def _compute_advance_payment(self):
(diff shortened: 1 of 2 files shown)


sale_advance_payment/models/sale.py
```diff
@@ -81,10 +81,9 @@ def _compute_advance_payment(self):
                     advance_amount += line_amount
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
-            for inv in order.invoice_ids:
-                invoice_paid_amount += (
-                    inv.amount_total_signed - inv.amount_residual_signed
-                )
+            for inv in order.invoice_ids.filtered(lambda x: x.state != "cancel"):
+                paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual
```
LoisRForgeFlow, line 87: isn't it going to be a problem if there are several invoices in different currencies? maybe you should convert from the company currency to the currency of the SO?

JordiMForgeFlow: @LoisRForgeFlow you are right, I am now converting to SO currency when needed, thanks!

rousseldenis:
@JordiMForgeFlow Maybe a little test with a cancelled invoice ?

JordiMForgeFlow:
@rousseldenis I just made the check in an already existing test by testing cancelling the invoice at the end

corredato:
LGTM

sale_advance_payment/models/sale.py
```diff
@@ -81,10 +83,16 @@ def _compute_advance_payment(self):
...
+                if inv.currency_id != order.currency_id:
+                    paid_amount = inv.currency_id._convert(
+                        paid_amount,
+                        order.currency_id,
+                        order.company_id,
+                        inv.invoice_date or inv.date,
+                    )
```
rousseldenis, line 94: To avoid computing things if not necessary
```suggestion
                    )
                else:
                    paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual
```

JordiMForgeFlow: @rousseldenis I don't get your proposal, why should we have to add the 'else' if we have already set the paid amount before the 'if'?

rousseldenis: Because potentially, you can do this :

`paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual` even if `inv.currency_id != order.currency_id`

JordiMForgeFlow: but you need to do `paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual` anyway in order to do the conversion when `inv.currency_id != order.currency_id`. If the currencies are not equal then you do the conversion, but if they are the same you already have the correct paid_amount. Not sure if I am missing something.

rousseldenis: Ooooh yes, didn't see, ok

rousseldenis (approved):

JordiMForgeFlow:
@rousseldenis could you merge this one?

LoisRForgeFlow:
@JordiMForgeFlow does this need fwport?

PR #3385, opened 2024-10-31
[17.0][MIG] sale_order_secondary_unit: Migration to 17.0

mmrondon:

PR #3873, opened 2025-08-26
[18.0][FIX] sale_order_secondary_unit: quantity resets to 0 with secondary UoM

lef-adhoc:
This PR fixes a bug where the product_uom_qty on a sale order line would reset to 0 when adding a product that uses a secondary unit of measure.

The logic that prevented this in previous versions was altered during the migration to v17 (https://github.com/OCA/sale-workflow/pull/3385)

### DIFF
--- a/sale_order_secondary_unit/models/sale_order.py
+++ b/sale_order_secondary_unit/models/sale_order.py
@@ -19,19 +19,13 @@ class SaleOrderLine(models.Model):
 
     product_uom_qty = fields.Float(copy=True)
 
-    @api.depends(
-        "display_type",
-        "product_id",
-        "product_packaging_qty",
+    @api.onchange(
         "secondary_uom_qty",
         "secondary_uom_id",
-        "product_uom_qty",
     )
-    def _compute_product_uom_qty(self):
-        res = super()._compute_product_uom_qty()
+    def _compute_target_field_qty(self):
         for line in self:
             line._compute_helper_target_field_qty()
-        return res
 
     @api.depends("product_id")
     def _compute_product_uom(self):
@@ -42,12 +36,22 @@ class SaleOrderLine(models.Model):
 
     @api.onchange("product_id")
     def _onchange_product_id_warning(self):
+        """
+        If default sales secondary unit set on product, put on secondary
+        quantity 1 for being the default quantity. We override this method,
+        that is the one that sets by default 1 on the other quantity with that
+        purpose.
+        """
         res = super()._onchange_product_id_warning()
         if self.product_id and not self.env.context.get("skip_secondary_uom_default"):
+            line_uom_qty = self.product_uom_qty
             self.secondary_uom_id = self.product_id.sale_secondary_uom_id
-            if self.product_uom_qty == 1.0:
-                self.secondary_uom_qty = 1.0
-                self._onchange_helper_product_uom_for_secondary()
+            if self.secondary_uom_id:
+                if line_uom_qty == 1.0:
+                    self.secondary_uom_qty = 1.0
+                    self._compute_helper_target_field_qty()
+                else:
+                    self.product_uom_qty = line_uom_qty
         return res
 
     @api.depends("secondary_uom_qty", "product_uom_qty", "price_unit")


lef-adhoc:
Hi @pedrobaeza, can you check this PR?

pedrobaeza:
@sergio-teruel is more suitable than me for checking this.

lef-adhoc:
Hi @sergio-teruel could you check this?

nicolascol (approved):
Funcitonal LGTM

lav-adhoc (approved):
LGTM

bosd (approved):

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -42,12 +42,22 @@ def _compute_product_uom(self):
...
         if self.product_id and not self.env.context.get("skip_secondary_uom_default"):
+            line_uom_qty = self.product_uom_qty
             self.secondary_uom_id = self.product_id.sale_secondary_uom_id
-            if self.product_uom_qty == 1.0:
-                self.secondary_uom_qty = 1.0
-                self._onchange_helper_product_uom_for_secondary()
+            if self.secondary_uom_id:
+                if line_uom_qty == 1.0:
+                    self.secondary_uom_qty = 1.0
+                    self._onchange_helper_product_uom_for_secondary()
```
mav-adhoc: instead of _onchange_helper_product_uom_for_secondary(), you should call the _compute_helper_target_field_qty() method, otherwise when the product qty is 1 and makes the comparison (in the if) to insert 1 unit in the secondary unit, after introducing 1 unit in the secondary qty, it introduces again 1 unit in the product original qty and makes the convertion.

rousseldenis:
@sergio-teruel

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -19,19 +19,13 @@ class SaleOrderLine(models.Model):
...
 
-    @api.depends(
-        "display_type",
-        "product_id",
-        "product_packaging_qty",
+    @api.onchange(
         "secondary_uom_qty",
         "secondary_uom_id",
-        "product_uom_qty",
     )
```
alexey-pelykh, line 25: This changes from `@api.depends` (computed, fires on any write) to `@api.onchange` (fires only in form views). Programmatic updates to `secondary_uom_qty` or `secondary_uom_id` via `write()`, `create()`, server actions, RPC, or data imports will no longer trigger a recomputation of `product_uom_qty`.

Also, the existing test `test_onchange_secondary_uom` calls `_compute_product_uom_qty()` directly, which no longer contains secondary unit logic after this change. That test may pass incidentally but is no longer exercising the intended path.

Could the fix be achieved while keeping the computed override of `_compute_product_uom_qty`?

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -42,12 +36,22 @@ def _compute_product_uom(self):
...
+        """
         res = super()._onchange_product_id_warning()
         if self.product_id and not self.env.context.get("skip_secondary_uom_default"):
+            line_uom_qty = self.product_uom_qty
             self.secondary_uom_id = self.product_id.sale_secondary_uom_id
-            if self.product_uom_qty == 1.0:
-                self.secondary_uom_qty = 1.0
-                self._onchange_helper_product_uom_for_secondary()
+            if self.secondary_uom_id:
+                if line_uom_qty == 1.0:
```
alexey-pelykh, line 50: The logic here to preserve `product_uom_qty` around the `secondary_uom_id` assignment looks correct and addresses the reported bug. However, when `secondary_uom_id` is falsy (product has no default secondary UoM), the code still assigns `self.secondary_uom_id = False` but skips the restoration of `product_uom_qty`. If that assignment triggers any recomputation in the chain, `product_uom_qty` could still be affected. A minor point, but worth verifying.

alexey-pelykh (changes requested):
Thank you for addressing this bug. The fix for preserving `product_uom_qty` in `_onchange_product_id_warning` looks like the right approach. However, I have concerns about the change from `@api.depends` computed override to `@api.onchange`, and the lack of a regression test.

**Key concern: `@api.depends` to `@api.onchange` migration**

The original code overrode `_compute_product_uom_qty` as an `@api.depends` computed method. The new `_compute_target_field_qty` is an `@api.onchange` method. This is a behavioral change:

- `@api.depends` (computed): fires on any write, programmatic or UI.
- `@api.onchange`: fires **only** in form views (client-side). Programmatic `write()` / `create()` calls, server actions, RPC, and data imports will **not** trigger the onchange.

This means that updating `secondary_uom_qty` or `secondary_uom_id` via code will no longer recompute `product_uom_qty`. This is a regression for any workflow that modifies secondary unit fields server-side.

**Test coverage**

1. The existing test `test_onchange_secondary_uom` (line 57) still calls `self.order.order_line._compute_product_uom_qty()`, which now falls through to the base Odoo method without secondary unit logic. The test may pass incidentally but no longer validates the secondary unit computation path.
2. No new test was added to cover the specific bug (quantity resetting to 0 when adding a product with a secondary UoM).

**CI status**: The "test with OCB" check is failing (logs expired, so root cause is unclear).

**Suggestion**: Could the fix in `_onchange_product_id_warning` (saving/restoring `product_uom_qty`) be applied while keeping the `_compute_product_uom_qty` override as a computed method? That would preserve server-side computation behavior. If the computed override itself causes the reset-to-zero bug, it would be helpful to understand the exact mechanism and document it.

Bilbonet (commented):
This PR solution works well.
I think the improvements mentioned by @alexey-pelykh are good ones to implement.

cgarcia-solvos (commented):
The FIX direction looks good but it should implement the fixes mentioned by @alexey-pelykh

rousseldenis:
@lef-adhoc Doesn't this : https://github.com/OCA/product-attribute/pull/2398 solve your problem ?

PR #10, opened 2014-08-07
7.0 pep8 fix

eLBati:

PR #3874, opened 2025-08-27
[18.0][ADD] sale_product_catalog_extended: New module to use catalog as a product_picker

CarlosRoca13:
cc @Tecnativa TT52109


pedrobaeza:
Reading the description, this should be called `sale_product_catalog_inline_edit`, isn't it? Or do you plan to add more things?

CarlosRoca13:
> Reading the description, this should be called `sale_product_catalog_inline_edit`, isn't it? Or do you plan to add more things?

More things are added, I'm improving the description

pedrobaeza:
OK, anyway, can we avoid the suffix `_extended`, as it's very generic?

sale_product_catalog_extended/controllers/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+        """Open sale order line on a given order for a given product.
+
+        :param int order_id: The order id.
+        :param int product_id: The product, as a `product.product` id.
+        :return: The id of the open sale order line.
+        :rtype: int
+        """
+        order = request.env["sale.order"].browse(order_id)
+        return order.with_company(order.company_id)._get_catalog_order_line(
+            product_id,
```
alexey-pelykh, line 20: **Security concern**: `order_id` is accepted directly from the client. There is no verification that the current user has access to this sale order. A logged-in user could pass any order ID and retrieve line information for orders they should not see. Consider adding `order.check_access_rights('read')` and `order.check_access_rule('read')`, or use `_search` to limit to accessible records.

sale_product_catalog_extended/controllers/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+    @route("/product/catalog/sale/get_order_line_data", auth="user", type="json")
+    def product_catalog_get_order_line_data(self, order_line_ids, **kwargs):
+        """Open sale order line on a given order for a given product.
+
+        :param list order_line_ids: The order lines to update the record.
+        :return: The data of the record that is being updated.
+        :rtype: dict
+        """
+        order_lines = request.env["sale.order.line"].browse(order_line_ids)
+        return order_lines._get_product_catalog_lines_data(**kwargs)
```
alexey-pelykh, line 33: **Security concern**: Same issue as above -- `order_line_ids` are accepted directly from the client without access verification. A user could pass arbitrary sale order line IDs to retrieve data they should not see.

sale_product_catalog_extended/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+    def search_fetch(self, domain, field_names, offset=0, limit=None, order=None):
+        catalog_orig_data_bool_list = [
+            "catalog_origin_data" in subdomain for subdomain in domain
+        ]
+        if any(catalog_orig_data_bool_list):
+            # Get value of catalog_origin_data from domain to know which method throw.
+            # Then browse the ids to respect the order.
+            product_ids = getattr(
+                self,
+                f"_get_product_picker_data_{domain[catalog_orig_data_bool_list.index(True)][2]}",
```
alexey-pelykh, line 27: This domain inspection is fragile. `domain` can contain string operators (`'&'`, `'|'`, `'!'`) which will also be iterated, and `"catalog_origin_data" in subdomain` will match if the string appears anywhere in the element (including as a substring). Consider using `odoo.osv.expression` utilities to safely extract leaves by field name.

Also, this bypasses `offset`, `limit`, and `order` parameters entirely when the special domain is detected -- it returns a browsed recordset without respecting pagination. This could cause performance issues with large result sets.

sale_product_catalog_extended/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+        months = 6
+        start = fields.Datetime.now() - timedelta(days=months * 30)
+        start = fields.Datetime.to_string(start)
+        catalog_partner_id = self.env.context.get("product_catalog_partner_id", False)
+        catalog_order_id = self.env.context.get("product_catalog_order_id", False)
+        # Search with sudo for get sale order from other commercials users
+        other_sales = (
+            self.env["sale.order"]
+            # .sudo()
+            ._search(
```
alexey-pelykh, line 51: Commented-out `.sudo()` -- please either remove or, if sudo access is intentionally needed, uncomment with a justification comment explaining why elevated privileges are required.

sale_product_catalog_extended/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+        ]
+        return domain
+
+    @api.model
+    def _get_product_picker_data_sale_order(self):
+        # Specific limit to allow show all products sold in recent orders
+        limit = int(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("sale_order_product_picker.product_picker_last_order_limit", "0")
```
alexey-pelykh, line 72: The config parameter key `sale_order_product_picker.product_picker_last_order_limit` does not match the module name `sale_product_catalog_extended`. This appears to be a leftover from a different/internal module. Consider renaming to `sale_product_catalog_extended.product_picker_last_order_limit` for consistency.

sale_product_catalog_extended/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+        catalog_orig_data_bool_list = [
+            "catalog_origin_data" in subdomain for subdomain in domain
+        ]
+        if any(catalog_orig_data_bool_list):
+            # Get value of catalog_origin_data from domain to know which method throw.
+            # Then browse the ids to respect the order.
+            product_ids = getattr(
+                self,
+                f"_get_product_picker_data_{domain[catalog_orig_data_bool_list.index(True)][2]}",
+            )()
```
alexey-pelykh, line 28: **Security risk**: This constructs a method name from a domain value (`domain[...][2]`) using `getattr` + f-string. While currently limited to the `catalog_origin_data` selection field, this pattern allows invoking any method named `_get_product_picker_data_*` on the model. A safer approach would be to use an explicit mapping dict:
```python
method_map = {"sale_order": self._get_product_picker_data_sale_order}
method = method_map.get(value)
```

sale_product_catalog_extended/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        self.ensure_one()
+        return self.order_line.filtered_domain(
+            self._get_catalog_order_line_filter_domain(product_id, **kwargs)
+        ).ids
+
+    def _get_action_add_from_catalog_extra_context(self):
+        # TODO: partner_id or shipping_partner_id?
+        return {
+            **super()._get_action_add_from_catalog_extra_context(),
```
alexey-pelykh, line 26: TODO comment suggests the design is not finalized. `partner_id` vs `partner_shipping_id` affects which customer's past orders are shown. This should be resolved before the PR is ready for review.

sale_product_catalog_extended/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_product_catalog_lines_data(self, **kwargs):
+        res = super()._get_product_catalog_lines_data(**kwargs)
+        if len(self) > 1:
+            res["readOnly"] = (
+                self.order_id._is_readonly()
+                or self.product_id.sale_line_warn == "block"
```
alexey-pelykh, line 39: `self.order_id._is_readonly()` and `self.product_id.sale_line_warn` are called on a recordset that can contain multiple lines (the `len(self) > 1` branch). If lines belong to different orders, `self.order_id` is a multi-record recordset, and `_is_readonly()` may raise or behave unexpectedly. Consider iterating or using `self.order_id[:1]._is_readonly()` if all lines are guaranteed to share the same order.

sale_product_catalog_extended/static/src/product_catalog/order_line/order_line.xml
```diff
@@ -0,0 +1,41 @@
...
+            expr="//div[hasclass('o_product_catalog_buttons')]/button[@t-on-click.stop='() => this.env.addProduct()']"
+            position="before"
+        >
+            <button
+                t-if="isInOrder() &amp;&amp; displayEdit()"
+                t-on-click.stop="() => this.env.openOrderLine()"
+                class="btn btn-light border w-100"
+            >
+                <i class="fa fa-pencil-square-o" />
+                Edit
```
alexey-pelykh, line 14: The strings "Edit", "Edit Lines", and "Total quantity: " are hardcoded in English. In Odoo OWL templates, user-facing text should be translatable. Wrap them appropriately so they can be extracted for i18n.

alexey-pelykh (commented):
## Review: sale_product_catalog_extended

This PR is currently marked as **Draft**. The comments below are intended as early feedback to help guide development.

### 1. Security: Missing access control on controller endpoints

Both new JSON-RPC endpoints (`/product/catalog/sale/open_order_line` and `/product/catalog/sale/get_order_line_data`) accept `order_id` / `order_line_ids` directly from the client but do not perform any access rights verification beyond `auth="user"`. A logged-in user could pass arbitrary IDs to read order line data belonging to other users/companies. Consider:
- Adding explicit `check_access_rights` / `check_access_rule` calls, or
- Using `_search` with the current user's context to filter accessible records.

### 2. No security files (`ir.model.access.csv`)

The module has no `security/` directory and no access rules declared in `__manifest__.py`. Even if the module only inherits existing models, OCA guidelines typically expect an explicit security directory if new fields or controllers are introduced.

### 3. No tests

There is no `tests/` directory. OCA modules require test coverage. At minimum, tests should cover:
- The new controller endpoints (access control, expected return values)
- The `search_fetch` override behavior
- The `_get_product_catalog_lines_data` logic for single and multi-line cases

### 4. Commented-out `.sudo()` in `product_product.py`

```python
# .sudo()
```

This looks like leftover debug code (line ~231 in the diff). It should either be removed or, if sudo is intentionally needed, uncommented with a justification comment.

### 5. Fragile `search_fetch` override

The override inspects domain elements by checking `"catalog_origin_data" in subdomain`, which treats each domain leaf as iterable/string-containable. This is fragile because:
- It will break on non-tuple domain elements (operators like `'&'`, `'|'`, `'!'`).
- It does not handle nested domains or normalized domain formats.

Consider using `odoo.osv.expression` utilities to safely extract the relevant leaf.

### 6. Config parameter key mismatch

The `ir.config_parameter` key is `sale_order_product_picker.product_picker_last_order_limit` but the module is named `sale_product_catalog_extended`. This looks like a leftover from a different/internal module. Should be renamed to match this module's technical name.

### 7. `_get_product_catalog_lines_data` multi-record safety

In `sale_order.py`, `_get_product_catalog_lines_data` accesses `self.order_id` and `self.product_id` directly on the recordset without iteration. When `len(self) > 1`, `self.order_id` returns a recordset of all related orders (or raises if `ensure_one` is expected internally by `_is_readonly`). The `.order_id._is_readonly()` call may not behave correctly for multiple distinct orders.

### 8. Hardcoded English strings in XML templates

The strings "Edit", "Edit Lines", and "Total quantity: " in `order_line/order_line.xml` are not wrapped for translation. In Odoo OWL templates, user-facing strings should use the translation mechanism.

### 9. TODO comments

There are `TODO` comments in:
- `sale_order.py` line ~291: `# TODO: partner_id or shipping_partner_id?`
- `sale_order_line_views.xml`: `<!-- TODO: Change to target new...`

These suggest the implementation is not finalized.

### 10. `getattr` with `f-string` for dynamic dispatch

In `product_product.py`:
```python
getattr(self, f"_get_product_picker_data_{domain[...][2]}")
```

This constructs a method name from user-controllable domain values. While domain values come from the search panel (limited selection), this pattern is risky. If the value is ever extended or manipulated, it could invoke unintended methods. Consider a mapping dict instead.

### Summary

Since this is a draft PR, these are early observations. The main blockers for OCA merge-readiness would be:
- **Security**: Access control on endpoints and the dynamic dispatch pattern
- **Tests**: Required by OCA
- **Code cleanup**: Commented code, TODOs, config key naming

---

CarlosRoca13:
I think it is ready to be merged @carlosdauden @sergio-teruel :smile:

sale_product_catalog_extended/models/sale_order.py
```diff
@@ -0,0 +1,202 @@
...
+    # Not stored, only used to know which partner the catalog history is
+    # matched against. Its default can be set per user/company through
+    # Default Values (ir.default).
+    use_delivery_address = fields.Boolean(store=False, default=False)
+
+    def _get_use_delivery_address_default_value(self):
+        """Return the user/company default value set for ``use_delivery_address``
+        through Default Values (ir.default).
+        """
+        field = self.env["ir.model.fields"].search(
```
eduezerouali-tecnativa:
```suggestion
        field = self.env["ir.model.fields"].sudo().search(
```
user might not be allowed to access it

eduezerouali-tecnativa (changes requested):
Tested on runboat.

carlosdauden:
Went through this in depth and checked each point against the current code (some had already moved since this review, and it looks like eduezerouali's suggestion is already applied too). Status:

**Fixed** (pushing shortly):
- `#4` (commented-out `.sudo()`): was a real bug, not leftover debug code — the comment right above it describes exactly why sudo is needed ("search sale order from other commercial users"), but with it commented out the *Last sales* history was silently scoped to the current user's own orders. Uncommented, with a clearer justification comment.
- `#6` (config parameter key mismatch): `sale_order_product_picker.product_picker_last_order_limit` was indeed a leftover from a different module and wasn't documented anywhere. Renamed to `sale_product_catalog_extended.catalog_last_order_limit` and added to `CONFIGURE.md`.
- `#3` (no tests): added coverage for the catalog origin resolution (`search_fetch`/`_get_catalog_origin_product_ids`, including the safe no-op on a stale origin value), the *Last sales* exclusion lifecycle, the renamed limit parameter, and the four controller endpoints. Verified with a full install + test run on a clean DB.

**Already resolved, from later commits**:
- `#9` (TODO comments): gone, no `TODO` left in the module.
- `#5` (fragile domain inspection): `_get_catalog_origin_product_ids` now checks `isinstance(leaf, list | tuple) and len(leaf) == 3 and leaf[0] == "catalog_origin_data"`, so operators like `'&'`/`'|'`/`'!'` are safely skipped.
- `#8` (hardcoded strings): "Edit"/"New line" are already extracted in the `.pot`/`.po` and translated — QWeb template text nodes get picked up by Odoo's own `--i18n-export` the same way `ir.ui.view` XML does, no explicit `_t()`/wrapping needed for plain text nodes.
- eduezerouali's suggestion (`.sudo()` on the `ir.model.fields`/`ir.default` lookup): already applied.

**Not actual bugs, left as is**:
- `#1`/`#2` (controller access on `order_id`/`order_line_ids`): checked this against `product/controllers/catalog.py` in core — `order = request.env[res_model].browse(order_id)` with no explicit `check_access_rights`/`check_access_rule` is exactly the pattern core's own catalog controller uses. The non-sudo `env` enforces `ir.rule`/`ir.model.access` on the first field read (e.g. `order.company_id`), so this isn't weaker than the base catalog.
- `#7` (`_is_readonly()` on a multi-record recordset): the module doesn't override `_get_product_catalog_lines_data` anymore, it's calling core's `sale.order.line` method directly, which already guards the multi-record branch with `self.product_id.ensure_one()` before touching a single line's `order_id`.
- `#10` (`getattr` + f-string dispatch): still there, but guarded by `hasattr(self, method_name)` before the call, and it's the intended extension point ("other modules may add further origins" by defining more `_get_product_picker_data_*` methods) — a mapping dict would just be the same lookup with extra indirection, not a safety improvement.

carlosdauden (approved):

carlosdauden:
@alexey-pelykh @eduezerouali-tecnativa Does that seem right to you now?

eduezerouali-tecnativa (approved):

pedrobaeza:
@OCA/pypi-support to reserve name

PR #3875, opened 2025-08-27
[FIX] Fix packing change

maq-adhoc:


luc-adhoc (approved):
functional review

rousseldenis:
@maq-adhoc Why this change as compute should replace onchanges?

maq-adhoc:
@rousseldenis 

The issue with depends comes up when using packages. If I change the number of packages, the secondary unit doesn't update.

The PR is in draft because I don't like the solution of changing it to an onchange. I tried several approaches, but I couldn't avoid the recursive computation between packages and units.

This week I'll continue looking for other solutions, but I'm open to suggestions.

rousseldenis:
> If I change the number of packages,

`product_packaging_qty` ?

alexey-pelykh (commented):
## OCA Review: `sale_order_secondary_unit` - Draft PR Observation

> **Note:** This PR is marked as **draft**. The author has acknowledged the solution is not ideal and is exploring alternatives. This review documents the technical concerns to assist with that exploration.

### Summary of Change

The PR replaces the `_compute_product_uom_qty` computed method override (with `@api.depends`) with a new `_compute_target_field_qty` method using `@api.onchange`. The stated motivation is fixing a recursive computation issue when changing packaging quantities.

### Concerns

#### 1. Regression: `super()._compute_product_uom_qty()` is no longer called

The original code overrides `_compute_product_uom_qty` and calls `super()` to ensure the base `sale.order.line` compute logic runs before applying secondary unit adjustments. The new code:
- Defines a completely new method `_compute_target_field_qty` (not an override)
- Never calls `super()._compute_product_uom_qty()`
- Does not remove or replace the `@api.depends` decorator on the original method

This means the base `_compute_product_uom_qty` from `sale.order.line` will run without the secondary unit post-processing, and the new `@api.onchange` method runs separately. This could lead to the compute overwriting the onchange result, since Odoo computes run after onchanges in many scenarios.

#### 2. `@api.onchange` vs `@api.depends` — architectural regression

Odoo 18 continues the trend of replacing onchanges with computed fields. The mixin documentation (`product.secondary.unit.mixin`) explicitly states:

> *"you must convert the field to computed writable (computed, stored and readonly=False), and you have to define the compute method adding `secondary_uom_id` and `secondary_uom_qty` fields as dependencies and calling inside to `self._compute_helper_target_field_qty()`"*

Switching from `@api.depends` to `@api.onchange` goes against this documented pattern and the Odoo 18 architectural direction. Onchanges do not fire during programmatic writes (e.g., `line.write({...})` or imports), so secondary unit conversion would silently stop working in non-UI contexts.

#### 3. Missing return value

The original `_compute_product_uom_qty` returns `res` from `super()`. While Odoo compute methods typically don't rely on return values, some upstream code or other modules in the inheritance chain may depend on it. The new method returns `None` implicitly.

#### 4. No version bump in `__manifest__.py`

If this change were to be merged, it would need a version bump since it changes observable behavior.

#### 5. No tests updated

The module has test files (`test_sale_order.py`, `test_product_template.py`) but this PR does not add or update tests to validate the new behavior or demonstrate the packaging issue being fixed.

### Suggestion

The root cause appears to be a circular dependency between `product_packaging_qty` and `product_uom_qty` computes when secondary units are in the dependency graph. Rather than switching to onchange, consider:

1. **Guarding against recursion** in the compute method (e.g., using a context flag like `_secondary_unit_no_recompute`)
2. **Narrowing the `@api.depends` triggers** — removing `product_packaging_qty` from the depends list and instead handling the packaging-to-secondary-unit flow through a separate compute or onchange specifically for the packaging field
3. **Using `env.remove_to_compute()`** (as the mixin already does in `_compute_helper_target_field_qty`) to break the cycle at the right point

The discussion between @rousseldenis and @maq-adhoc in the comments indicates this is still being investigated, which aligns with the draft status.

---

rousseldenis:
@maq-adhoc Do you plan to finish this as this is quite old? We should close it.

maq-adhoc:
@rousseldenis I'll go over it with the team and either move forward or close the pr

maq-adhoc:
I'm closing this because we decided on a different approach: https://github.com/OCA/sale-workflow/pull/4435

PR #3880, opened 2025-09-01
[18.0][IMP] sale_order_line_sequence: change visible_sequence from int to char

OKT96:
This PR introduces the following changes:

- Field type of `visible_sequence` (sale.order.line) changed from integer to char. This allows for more flexible numbering schemes e.g: 1.1, 1.2, 1.3 ...
- Reordered the `related_so_sequence` in the view, so that the `handle` widget is shown first (as in sale order lines):
  before
  
  after

### DIFF
--- a/sale_order_line_sequence/models/sale_order_line.py
+++ b/sale_order_line_sequence/models/sale_order_line.py
@@ -14,7 +14,7 @@ class SaleOrderLine(models.Model):
         default=9999,
     )
 
-    visible_sequence = fields.Integer(
+    visible_sequence = fields.Char(
         "Line Number",
         help="Displays the sequence of the line in the sale order.",
         compute="_compute_visible_sequence",
@@ -29,5 +29,5 @@ class SaleOrderLine(models.Model):
                 lambda order_line: not order_line.display_type
             )
             for line in sorted(order_lines, key=lambda order_line: order_line.sequence):
-                line.visible_sequence = sequence
+                line.visible_sequence = str(sequence)
                 sequence += 1
--- a/sale_order_line_sequence/__manifest__.py
+++ b/sale_order_line_sequence/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Sale Order Line Sequence",
     "summary": "Propagates SO line sequence to invoices and stock picking.",
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.2.0",
     "author": "ForgeFlow, Serpent CS, Odoo Community Association (OCA)",
     "category": "Sales",
     "website": "https://github.com/OCA/sale-workflow",
--- a/sale_order_line_sequence/migrations/18.0.1.2.0/pre-migrate.py
+++ b/sale_order_line_sequence/migrations/18.0.1.2.0/pre-migrate.py
@@ -1,0 +1,9 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        "ALTER TABLE sale_order_line ALTER COLUMN visible_sequence TYPE varchar",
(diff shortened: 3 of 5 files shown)


sale_order_line_sequence/models/sale_order_line.py
```diff
@@ -14,7 +14,7 @@ class SaleOrderLine(models.Model):
         default=9999,
     )
 
-    visible_sequence = fields.Integer(
+    visible_sequence = fields.Char(
```
alexey-pelykh, line 17: Since `visible_sequence` is a stored field (`store=True`), changing its type from `Integer` to `Char` requires a database column type change for existing installations.

Please:
1. Bump the version in `__manifest__.py` (e.g., to `18.0.1.2.0`)
2. Add a migration script at `migrations/18.0.1.2.0/pre-migrate.py` to handle the column conversion. For example:

```python
from openupgradelib import openupgrade

@openupgrade.migrate()
def migrate(env, version):
    openupgrade.logged_query(
        env.cr,
        "ALTER TABLE sale_order_line ALTER COLUMN visible_sequence TYPE varchar",
    )
```

Also, consider assigning `str(sequence)` in `_compute_visible_sequence` (line 34) instead of relying on implicit int-to-string coercion, to be consistent with the new field type and the updated test assertions.

alexey-pelykh (changes requested):
Thanks for the PR! The change from Integer to Char for `visible_sequence` is a reasonable improvement that unlocks flexible numbering schemes.

However, I see a couple of issues that should be addressed before merging:

**1. Missing migration script and version bump**

`visible_sequence` is a **stored** computed field (`store=True`). Changing its type from Integer to Char requires a database column type change. While PostgreSQL can handle the implicit cast from `integer` to `varchar`, OCA best practice is to:
- Bump the version in `__manifest__.py` (e.g., `18.0.1.2.0`)
- Add a `migrations/18.0.1.2.0/pre-migrate.py` script using `openupgradelib` to handle the column type conversion cleanly

Without a version bump and migration, the upgrade path for existing installations is not explicitly handled.

**2. Compute method still assigns integer values**

The `_compute_visible_sequence` method (line 34) still assigns `line.visible_sequence = sequence` where `sequence` is an `int`. While Odoo will implicitly convert this to a string for a Char field, it would be more explicit and consistent with the field type change to assign `str(sequence)`. This also makes the intent clearer for future developers and is consistent with the test assertions (which now correctly compare against string values like `"1"`).

OKT96:
@alexey-pelykh Hi, thanks for the review and the helpful pointers! I've updated the PR with all the requested changes

rousseldenis (changes requested):
This is not correct. As `sequence` field is still an Integer, `visible_sequence` should not be another type.

@OKT96

alexey-pelykh (approved):
Thanks @OKT96 — both items from my previous review are addressed:

- Version bumped to `18.0.1.2.0` with a `migrations/18.0.1.2.0/pre-migrate.py` that runs the `ALTER COLUMN visible_sequence TYPE varchar` conversion for the stored field.
- `_compute_visible_sequence` now assigns `str(sequence)` explicitly.

CI is green. LGTM.

alexey-pelykh (commented):
Withdrawing my approval here — apologies for the churn. I'd checked off the mechanical items (version bump, `pre-migrate.py`, `str(sequence)`), but those don't address @rousseldenis's substantive point, which is the design one: while `sequence` stays an `Integer`, changing `visible_sequence` to `Char` introduces a type divergence that should be justified first. I don't have grounds to approve over that — deferring to @rousseldenis. The type question is the thing to settle before this moves.

OKT96:
@rousseldenis I would like to propose this type change to allow more flexibility... Could you tell me your thoughts on why `visible_sequence` should stay the same type as `sequence`?

Issue #2667, opened 2023-08-31
[16]sale_exception: Subscription behavior when confirming an upsell


jok-adhoc:
module: sale_exception
version: 16.0

**Steps to reproduce**
- Create a subscription with a partner who makes an exception arise
- Confirm the subscription ignoring exception and create the first invoice
- Create an upsell
- Try to confirm it but do not ignore the exception

**Current behavior**

Even though the exception is not ignored, the subscription's sale order lines are updated. This is because the sequence action_confirm is taking. 

sale_subscription's action_confirm:
![image](https://github.com/OCA/sale-workflow/assets/109111399/58d462a2-f6f3-4bc8-9f6e-9ddfa51ebd16)
The exception is being consider when res is defined but it continues and the lines are updated bellow.

Furthermore, if the exception is ignored, they are updated too. And then, if it is confirmed, the subscription's lines are updated again.
I did not evaluate the behavior with renewals, but judging the code, it may also generate some issues.

**Expected behavior**

The subscription's order lines are expected to be updated just once, when the upsell is confirmed.

grindtildeath:
@OCA/crm-sales-marketing-maintainers Can we please reopen this issue please?  🙏

simahawk:
fix for v18 is getting merged https://github.com/OCA/server-tools/pull/3476

PR #3885, opened 2025-09-05
[18.0][FIX] sale_exception: Rollback confirmation upon exception

grindtildeath:


grindtildeath:
FTR: this is not new https://github.com/OCA/sale-workflow/issues/2667 :sob:

lef-adhoc:
Hi @grindtildeath Is there any progress on this?

Could you check this https://github.com/OCA/sale-workflow/compare/16.0...adhoc-dev:sale-workflow:16.0-h-89839-lav2?expand=1 ?

grindtildeath:
@lef-adhoc Sorry for the lack of answer, I had something in backlog to get back to this issue but did not have much time to work on it before the last few days, thanks to the holiday season.

I did work on a POC that would rollback any transaction where an exception rule is detected, but this had to be done in base_exception module: https://github.com/OCA/server-tools/pull/3476

It still needs to be improved and we'll need tests, but AFAICT first results are promising.

sale_exception/models/sale_order.py
```diff
@@ -2,6 +2,8 @@
 # Copyright 2018 Akretion
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.api import Environment
+from odoo.modules.registry import Registry
 
 from odoo import api, models
```
alexey-pelykh, line 9: `breakpoint()` must be removed -- this will halt execution in production.

sale_exception/models/sale_order.py
```diff
@@ -48,10 +50,18 @@ def sale_check_exception(self):
```
alexey-pelykh, line 47: Opening a new cursor via `Registry(...).cursor()` and committing inside it bypasses Odoo's transaction management. The context manager will also commit on exit, leading to a double-commit. Consider using `cr.savepoint()` instead for exception detection within the current transaction.

Additionally, any writes committed on `new_cr` become permanently visible even if the original transaction rolls back, which can leave orphaned data.

sale_exception/models/sale_order.py
```diff
@@ -48,10 +50,18 @@ def sale_check_exception(self):
...
 
     def action_confirm(self):
-        if self.detect_exceptions():
-            if not self.env.company.sale_exception_show_popup:
-                return
-            return self._popup_exceptions()
+        breakpoint()
+        with Registry(self.env.cr.dbname).cursor() as new_cr:
+            new_env = Environment(new_cr, self.env.uid, self.env.context)
+            exception_ids = self.with_env(new_env).detect_exceptions()
```
alexey-pelykh, line 56: The original `_popup_exceptions()` returned a wizard action for user review. Replacing it with `_check_exception()` (which raises `ValidationError`) removes the ability for users to acknowledge/override exceptions. The FIXME here captures this concern -- it is indeed a UX regression that should be addressed.

test_sale_exception_confirm/tests/test_sale_exception_confirm.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo.fields import Command
+from odoo.tests import TransactionCase
+
+
+class TestSaleExceptionConfirm(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.no_free_exception = cls.env.ref("sale_exception.excep_no_free")
+        # cls.no_free_exception.active = True
```
alexey-pelykh, line 12: Commented-out code: `# cls.no_free_exception.active = True`. If the demo data (which sets `active=True` on the exception rule) handles this, this line should be removed entirely rather than left commented out.

alexey-pelykh (commented):
## Review: Draft / Work-in-Progress

This PR is currently marked as **draft**. The author has noted in comments that the approach is evolving, with a more complete solution being developed in [OCA/server-tools#3476](https://github.com/OCA/server-tools/pull/3476) (at the `base_exception` level). The observations below are intended to help guide that continued development.

---

### Critical: `breakpoint()` left in code

`sale_exception/models/sale_order.py` line 53 contains a `breakpoint()` call. This is a Python debugger statement that will halt execution in any non-interactive context (e.g., production Odoo workers). Must be removed before any merge.

### Critical: New cursor pattern is risky

The approach of opening a separate cursor via `Registry(self.env.cr.dbname).cursor()` and committing inside it has several concerns:

1. **Manual `new_cr.commit()` inside the `with` block** -- the context manager for `Registry.cursor()` will also commit on successful exit, leading to a double-commit. More importantly, manual commits inside Odoo business logic circumvent the ORM's savepoint and transaction management, which is strongly discouraged.

2. **Data visibility** -- exception detection results written in the new cursor's transaction become visible to other concurrent transactions immediately after commit, even if the calling transaction (the original `self.env.cr`) ultimately rolls back. This can leave orphaned exception records in the database.

3. **Environment mismatch** -- `self.with_env(new_env)` rebinds the recordset to a different cursor. Any cache populated on `new_env` is invisible to the original environment, which can lead to subtle stale-cache bugs.

Idiomatic Odoo approaches for this kind of problem typically use `cr.savepoint()` to detect exceptions within the current transaction and roll back to the savepoint if needed, rather than opening a new cursor.

### Concern: Removed popup/wizard UX flow

The original code called `self._popup_exceptions()` which returns a wizard action allowing the user to review and optionally ignore exceptions. The new code replaces this with `self._check_exception()` which raises a `ValidationError` directly. This is a UX regression -- users lose the ability to acknowledge and override exceptions. The FIXME comment in the code acknowledges this issue.

### Minor: Potentially unbound `exception_ids`

If an unexpected error occurs inside the `with` block before `exception_ids = self.with_env(new_env).detect_exceptions()` executes, the subsequent `if exception_ids:` would raise `UnboundLocalError`. Consider initializing `exception_ids` before the `with` block.

### Minor: Commented-out code in test

`test_sale_exception_confirm/tests/test_sale_exception_confirm.py` line 12 has `# cls.no_free_exception.active = True` commented out. This should either be removed or uncommented with an explanation.

### Minor: Test manifest missing `installable` key

The `test_sale_exception_confirm/__manifest__.py` does not explicitly set `"installable": True`. While Odoo defaults to `True`, OCA convention is to be explicit. Additionally, test-only modules in OCA typically follow the pattern of not being installable in production.

### Observation: No PR description

The PR body is empty. A description explaining the problem being solved, the approach taken, and known limitations would help reviewers -- especially since this is addressing a long-standing issue ([#2667](https://github.com/OCA/sale-workflow/issues/2667)).

---

**Summary**: This is clearly still a work-in-progress exploration. The core challenge (rolling back `action_confirm` side effects when exceptions are detected) is a real and important problem. The new-cursor approach has fundamental transaction safety issues, but the direction of moving this logic to `base_exception` (as mentioned in the comments) seems more promising. Looking forward to seeing the evolution of this work.

rousseldenis:
@grindtildeath

grindtildeath:
Closed in favour of https://github.com/OCA/sale-workflow/pull/4255

PR #3886, opened 2025-09-08
[18.0][ADD] sale_product_identification

edescalona:
@BinhexTeam

Sometimes, when selling a product, it is necessary to require documentation (authorization, ID, etc.) from the customer to justify the sale.
Examples:

- Sale of a corrosive product: A document authorizing its handling is required. To control this process, we configure a Corrosive category in the product listing.
- Sale of alcohol: In this case, sales to minors are not permitted. To control this process, we configure a Minor category in the product listing.

This module allows you to control these processes through identifications in the product listing.


sale_product_identification/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module allows you to add required IDs to products and validate those identifications when creating a sales order.
```
rrebollo:
```suggestion
This module allows you to add required IDs to products and validate those identifications when confirming a sales order.
```
Validation does not occur during creation, is that correct?

sale_product_identification/readme/DESCRIPTION.md
rrebollo, line 1: Please note that this functionality depends on the `partner_identification` module. For detailed setup and configuration, please refer to that addon's documentation.

sale_product_identification/models/res_partner_id_number.py
```diff
@@ -0,0 +1,58 @@
...
+import pytz
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class ResPartnerIdNumber(models.Model):
+    _inherit = "res.partner.id_number"
+
+    def validate_identification(self, **params):
```
rrebollo:
```suggestion
    @api.model
    def validate_identification(self, **params):
```

Additionally, adding a brief docstring here would be helpful for clarity and future maintenance.

sale_product_identification/models/sale_order.py
```diff
@@ -0,0 +1,45 @@
...
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _validate_identification(self):
+        products_identification_ids = (
```
rrebollo, line 90:
```suggestion
        self.ensure_one()
        products_identification_ids = (
```

sale_product_identification/models/res_partner_id_number.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2025 Binhex <https://www.binhex.cloud>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from datetime import datetime
+
+import pytz
+
+from odoo import _, models
```
rrebollo:
```suggestion
from odoo import _, api, models
```

sale_product_identification/tests/sale_product_identification_common.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import Command
+from odoo.tests import common
```
rrebollo:
```suggestion
from odoo.addons.base.tests.common import BaseCommon
```

sale_product_identification/tests/sale_product_identification_common.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class TestSaleOrderIdentificationCommon(common.TransactionCase):
```
rrebollo:
```suggestion
class TestSaleOrderIdentificationCommon(BaseCommon):
```

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

edescalona:
Hi @rrebollo thanks for your comments, ready for review.

epieters:
The naming of the checkboxes is confusing:
- Required Identification
- Optional Required Identification
In addition, you currently can select both, which doesn't make sense to me.

Suggestion:
- Identification
- Required YES/NO

edescalona:
Hello @epieters , thank you for your comments. The Optional Required Identification option indicates that for that product, it is not mandatory for the partner to have the associated categories, but rather requires user action to validate it when confirming the sales order.

Example:

When selling whiskey, the user asks the customer for their ID to determine if they are of legal age and then proceeds to validate the order.

epieters:
Hi @edescalona, from a linguistic viewpoint, optional and required are mutual exclusive. 
Either something is optional OR it is required, ... there is no such thing as an optional required identification. :-)

Other question I have is suppose you have multiple identifications, some are mandatory, and some are optional.
It looks like this is not possible in the current implementation.

Hence I suggest that required is a property of the identification. i.e. ssn is optional, but 18+ is required.

edescalona:
Hi @epieters , I understand what you're saying. I think the solution might be to change the name of the optional field to "Validation Required" or something like that? What do you think? Because it could be that today they're required, but tomorrow they're optional, and then it's just a matter of checking that validation field and that's it.

The thing is, the IDs aren't optional; it's more the product itself. That is, if you check the product as an optional ID, then they're all optional.

epieters:
Hi @edescalona, yes indeed, renaming the 2nd checkbox would indeed help.
My suggestion:
- Identification (checkbox)
- Required (Boolean)

I also see great value in having some Identifications mandatory and some optional.

edescalona:
Hi @epieters , thanks for your suggestions. I'll keep them in mind and let you know for review.

Based on your last comment, if the required field is marked on the product and one of the selected IDs is optional, would the mandatory field take priority?

rrebollo (approved):
Regarding my suggestions, I can confirm that you have addressed all of them. Thank you. Nevertheless, it's import to evaluate @epieters comments.

edescalona:
Hi @epieters , thanks for your suggestions. I'll keep them in mind and let you know for review.

sale_product_identification/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,4 @@
+This module allows you to add required IDs to products and validate those identifications when confirming a sales order.
```
rousseldenis: This should a little bit explained (which IDs ?).

A CONTEXT in readme should be great to explain it. Thanks

edescalona:
Hi @epieters @rousseldenis @rrebollo @arielbarreiros96, I made some changes to the module to allow categories to be mandatory or not, and to add a category description for each product if necessary.

image

rousseldenis (approved):
LGTM

sale_product_identification/models/res_partner_id_number.py
```diff
@@ -0,0 +1,99 @@
...
+            ("valid_from", "!=", False),
+            ("valid_until", "!=", False),
+            ("valid_from", "<=", now_dt),
+            ("valid_until", ">=", now_dt),
+            "&",
+            "&",
+            ("valid_from", "=", False),
+            ("valid_until", "!=", False),
+            ("valid_until", ">=", now_dt),
+        ]
```
christian-ramos-tecnativa: IMO this should be inside a method to make it easier to extend

sale_product_identification/models/product_template_id_category.py
```diff
@@ -0,0 +1,20 @@
...
+
+class ProductTemplateIdcategory(models.Model):
+    _name = "product.template.id_category"
+    _description = "Product Template Identification Category"
+
+    product_tmpl_id = fields.Many2one("product.template")
+    category_id = fields.Many2one("res.partner.id_category")
+    is_required = fields.Boolean(
+        default=True, help="Defines whether identification is required."
+    )
```
christian-ramos-tecnativa: In order for the nomenclature be more clear I propose changing this to "is_mandatory"

christian-ramos-tecnativa (changes requested):

edescalona:
Hello @Christian-RB  thanks for your comments, ready for review.

arielbarreiros96 (approved):
@edescalona tested after latest changes, and it seems quite good, I tested all scenarios I could think of and so far everything seems covered.

christian-ramos-tecnativa (approved):

epieters (commented):
All good from my side. This will work for the use cases I can think off.

PR #3887, opened 2025-09-09
[17.0][FIX] - Error quick search in sale order form view

DucTruongKomit:
**Current Behavior**
In SO form view, press Ctrtl + K, the error will appear

NOTE:
Make sure that base.substate does not have any record. If there is a record, the error will not appear.

### DIFF
--- a/sale_substate/models/sale_order.py
+++ b/sale_substate/models/sale_order.py
@@ -15,3 +15,9 @@ class BaseSubstateType(models.Model):
 class SaleOrder(models.Model):
     _inherit = ["sale.order", "base.substate.mixin"]
     _name = "sale.order"
+
+    has_base_substate = fields.Boolean(compute="_compute_has_base_substate")
+
+    def _compute_has_base_substate(self):
+        for rec in self:
+            rec.has_base_substate = bool(self.env["base.substate"].search_count([]))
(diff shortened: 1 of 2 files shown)


sale_substate/models/sale_order.py
```diff
@@ -15,3 +15,9 @@ class BaseSubstateType(models.Model):
 class SaleOrder(models.Model):
     _inherit = ["sale.order", "base.substate.mixin"]
     _name = "sale.order"
+
+    has_base_substate = fields.Boolean(compute="_compute_has_base_substate")
```
rousseldenis, line 19: @DucTruongKomit And why not having done it in base module ?

sale_substate/models/sale_order.py
```diff
@@ -15,3 +15,9 @@ class BaseSubstateType(models.Model):
...
+
+    def _compute_has_base_substate(self):
+        for rec in self:
+            rec.has_base_substate = bool(self.env["base.substate"].search_count([]))
```
rousseldenis, line 23: You should at least add `model == 'sale.order'` because you will possibly retrieve substates from other models

sale_substate/models/sale_order.py
DucTruongKomit: Hi thank you your review. I think you're right.  I will move it to base.substate.mixin.

DucTruongKomit: I have a better idea that we can make this field becomes a count field.

DucTruongKomit: I will have a new commit asap.

rousseldenis:
@DucTruongKomit

DucTruongKomit:
I'm still on it.

DucTruongKomit:
I did the commit wrong repo, I will close for now.

PR #3888, opened 2025-09-09
[18.0][MIG] sale_order_line_remove

ArnauCForgeFlow:
Migration to version 18.0 + add sale_order_line_remove as rebel module as mentioned in the comments from the original pr: https://github.com/OCA/sale-workflow/pull/3632#issuecomment-3068578707.

Supersedes: https://github.com/OCA/sale-workflow/pull/3632


LoisRForgeFlow (approved):
Functional test, works as expected :+1:

BhaveshHeliconia (approved):
Functional review LGTM!

.copier-answers.yml
```diff
@@ -19,6 +19,7 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - sale_packaging_default
+- sale_order_line_remove
```
rousseldenis: @ArnauCForgeFlow @LoisRForgeFlow And what about introducing an activation parameter that allows to not break standard behaviour? That allows to activate/deactivate behaviour ... and preserve planet with less CI jobs :smile:

LoisRForgeFlow: @rousseldenis If you install this module you want the behavior, the parameter is redundant in my opinion. So I guess it is a trade-off between code simplicity and CI efficiency.

rousseldenis: That is the eternal debate. But having a way to deactivate/activate a behavior change in the interface is a good practice IMHO

LoisRForgeFlow: @rousseldenis I let you the final call as PSC of the repo. Do you want us to add that parameter?

rousseldenis: Should be great

LoisRForgeFlow: Sure!

@ArnauCForgeFlow could you take care of this?

ArnauCForgeFlow:
@rousseldenis done!

rousseldenis (approved):

Issue #3649, opened 2025-03-29
sale_triple_discount: Additive discount type causes mismatch between sale order and invoice


sbejaoui:
During the migration of `sale_triple_discount` (!3648), I noticed a major compatibility issue between this module and `account_invoice_triple_discount`.

In `sale_triple_discount`, there is a `discount_type` field that allows selecting between `multiplicative` and `additive` discounts. However, in the account_ module, only `multiplicative` discounts are supported. This causes discrepancies in the discount, total, and other amounts between the order and the invoice when the additive discount type is selected.

After investigating, I found that this issue has existed for a while (at least since v16). This makes me question whether the additive discount type is actually being used in practice.

To address this issue, I propose hiding the field for now and preventing users from selecting the additive discount type. If someone is interested in properly fixing this in the future—ideally by adding discount_type to account.move.line—they can revert my commit and implement the necessary changes.

screenshots from v16
![Image](https://github.com/user-attachments/assets/dd332c54-2c48-4b69-835d-932d28105206)

![Image](https://github.com/user-attachments/assets/1f6ffb42-d7f1-4649-900c-151fd3f9c007)

PR #3889, opened 2025-09-09
[18.0][MIG] sale_triple_discount: Migration to version 18.0 #2093

pablo-cort-s73:
Migration sale_triple_discount to v18, adapt changes https://github.com/OCA/sale-workflow/pull/3929
Migration sale_order_general_discount_triple to v18 with correct changes.

Supersedes #3648 

- [x] depends #4446


ferran-S73 (approved):

sale_triple_discount/models/sale_order_line.py
```diff
@@ -161,3 +161,13 @@ def _prepare_invoice_line(self, **kwargs):
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        order_lines = super().create(vals_list)
+        lines_to_discount = order_lines
+        for line in order_lines:
+            if line.discount != 0.0 or line.discount1 != 0.0:
+                lines_to_discount -= line
+        lines_to_discount._compute_discounts()
+        return order_lines
```
akash-aipl, line 150: ⚠️ Potential issue:
This create() override calls `_compute_discounts()` on all lines where
both `discount` and `discount1` appear to be 0.0 immediately after create.
However, at this stage in the ORM lifecycle, computed fields such as
`discount` and `discount1` may not yet have been fully recomputed or reflect
the values provided in `vals_list`. As a result, lines that actually had a
user-defined discount (e.g. discount1 = 0 or a nonzero value passed in vals)
can be incorrectly included in `lines_to_discount`, causing
`_compute_discounts()` to run and `_inverse_discount()` to overwrite
the user-provided values.

This leads to the UX issue where setting `discount1 = 0` requires doing it
twice, since the first assignment gets replaced by the computed discount.

ferran-S73: Hi @akash-aipl thanks so much for your review!

I'm not sure I fully understand in which case is the discount overridden, could you please specify the configuration needed for that case? Maybe reproduce it in this PR's runbot? Thanks!

akash-aipl: > Hi @akash-aipl thanks so much for your review!
> 
> I'm not sure I fully understand in which case is the discount overridden, could you please specify the configuration needed for that case? Maybe reproduce it in this PR's runbot? Thanks!


https://github.com/user-attachments/assets/d5bf71db-bc82-46c4-9339-bd02c48b508b

Hello @ferran-S73!

Here's the video of case, please check.

ferran-S73: Hi again @akash-aipl 

I added another check inside the create method to check if discount1 is received in vals. Thus asserting that:
 - For a manually created line the discount is saved either if it was calculated by Odoo or set by the user.
 - Any other line through code will calculate the discount if it's no set specified or save the discount that is passed to the create method.

I think this should cover for every case. Let me know if there's anything else you think should be review again.

ferran-S73:
@lef-adhoc @rousseldenis could you review our approach to solve the issues mentioned in the original pr please?

lef-adhoc (approved):
LGTM @ferran-S73 @pablo-cort-s73

nicolascol (approved):
Functional review LGTM
Could you merge please?
@rousseldenis

lef-adhoc:
Could you correct the tests?

ferran-S73:
> Could you correct the tests?

@lef-adhoc it appears that sale_order_general_discount has been merged since the opening of the original pr for this module. So now the sale_order_general_discount_triple may be needed for compatibility between the 2 modules. I'll try to get to migrating it as soon as I have some free time :sweat_smile:

lef-adhoc:
@ferran-S73 @pablo-cort-s73 Could you check this migration? https://github.com/OCA/sale-workflow/pull/3965

sbejaoui (approved):
Functional test

ferran-S73:
@lef-adhoc tests passed. We included the sale_order_general_discount_triple in this PR.

@hildickethan @DantePereyra your reviews would be greatly appreciated.

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,147 @@
...
+    @api.depends(
+        lambda self: self._get_multiple_discount_field_names()
+        + ["product_id", "product_uom", "product_uom_qty"]
+    )
+    def _compute_discount(self):
+        # Base Odoo just continues instead of assigning to 0 in this case
+        # but we depend on the super() value resetting to a discount unpolluted
+        # by the extra fields before taking them into account
+        for line in self:
+            if not (line.pricelist_item_id and line.pricelist_item_id._show_discount()):
```
hildickethan:
```suggestion
        discount_enabled = self.env['product.pricelist.item']._is_discount_feature_enabled()
        for line in self:
            if not (line.order_id.pricelist_id and discount_enabled):
```
Base 18.0 `_compute_discount` looks a bit different to 17.0
https://github.com/odoo/odoo/blob/322c6d0468bf79e9d29e1375c49aa13d4a7b7a67/addons/sale/models/sale_order_line.py#L758-L777

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        order_lines = super().create(vals_list)
+        lines_to_discount = self.env["sale.order.line"]
+        for line, vals in zip(order_lines, vals_list, strict=True):
+            if "discount" in vals and vals["discount"] == 0:
+                lines_to_discount |= line
+        lines_to_discount.write({"discount1": 0.0, "discount2": 0.0, "discount3": 0.0})
+        return order_lines
```
hildickethan, line 150: this shouldn't be necessary as dragging `discount` values to `discount1` in create/write is handled in the `triple.discount.mixin` https://github.com/OCA/account-invoicing/blob/11911fe8c88b61f6b37851394bd69faab4dbf330/account_invoice_triple_discount/models/triple_discount_mixin.py#L77-L91

Unfortunately it looks like the 18.0 version is a bit outdated, I changed it a little to help with handling `sale_fixed_discount` but it only really changes the `write` logic a little 
https://github.com/OCA/account-invoicing/blob/e2fe4512114ae614470e4efa6b53d766cddd93f6/account_invoice_triple_discount/models/triple_discount_mixin.py#L84-L109

.copier-answers.yml
```diff
@@ -19,6 +19,8 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - sale_packaging_default
+- sale_fixed_discount
+- sale_triple_discount,sale_order_general_discount_triple
```
hildickethan, line 23: you can avoid this by adding ANOTHER glue module.... like I did in https://github.com/OCA/sale-workflow/pull/3907 but it hasn't really gathered much attention so it doesn't matter for now seperating is also ok

hildickethan (commented):
the rest seems ok, if it works

sale_triple_discount/models/sale_order_line.py
sbiosca-s73: Regarding your comments, I have modified the create function for cases where the `discount` is forced to 0 in other modules, as it is intended that those lines have no `discount` regardless of the pricelist. I see that what you mentioned about the function in `triple.discount.mixin` in v17 also breaks the case I described, since I am detecting that situation in v17 and cannot force the `total discount` to 0 in any case. In v18, since it is outdated as you mentioned, I have that case solved with this create. However, it does not work for me in v17, since I see that in the `triple.discount.mixin` function, if the `total discount` is passed as 0, it is assigned to `discount1`, but if `discount2` or 3 have values, they remain. I do not think this is correct, as the total `discount` has been forced to 0.

hildickethan: The logic I followed when I changed it was that the code is necessary for base Odoo tests and other OCA modules that expect to be able to set a value to `discount`, which with `sale_triple_discount` becomes `discount1`, so it passes the value over. But if a personal or another OCA module is aware of `sale_triple_discount` and is actively using the `discount2` and `discount3` fields, then IMO it's up to the module in question to set the values to the appropriate fields
Since it's just a matter of opinion I don't expect you to change it, as long as it works for those that use the module and the tests pass it's fine as is

sale_triple_discount/models/sale_order_line.py
hildickethan: @sbiosca-s73 I'd say this change is necessary to avoid future headaches of possible edge cases where the 17.0 `if` passes and the `18.0` one doesn't or vice versa

hildickethan (approved):

laiaespinosa:
Hello. I have tested this module in an Odoo 18.0 Enterprise environment (Spanish localization).
Installation: OK
Functional behavior: OK (Triple discounts are calculated correctly in Sale Orders and Invoices).

lef-adhoc:
@laiaespinosa Hi, can you check the message from Oca Bot?

laiaespinosa (approved):
**Tested and working on Odoo 18.0 EE. Ready for migration.**

Hi! Sorry for the confusion with the bot. I am not a maintainer of this repo, so I cannot execute the / o c a b o t m e r g e command (and that's why it failed with the bumpversion message).

I have already submitted an official 'Approve' through the 'Review changes' tab to confirm that the module works perfectly in Odoo 18.0 EE. A maintainer should be able to trigger the merge now. Thanks!

laiaespinosa:
I've edited my previous comment to avoid confusion with the bot. 
As I confirmed, the module is working fine on Odoo 18.0 EE.
Maybe @pedrobaeza or @carlosdauden can take a look when they have a chance? 
Thanks!

rousseldenis:
Hi all.

I'm diving into this.

.copier-answers.yml
rousseldenis: @pablo-cort-s73 Have you considered this ?

sale_triple_discount/hooks.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024-Today - Sylvain Le GAL (GRAP)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(env):
```
rousseldenis: @pablo-cort-s73 @legalsylvain I have an issue with this.

This comes too late at install on existing database. All existing discounts are lost.

I will check if putting in pre_init_hook is better.

legalsylvain: Indeed ! putting it in pre_init_hook will do the job.

pablo-cort-s73: Hook updated :+1:

hitrosol:
The runboat doesn't produce the demo db, but it works on my local. LGTM.

alexey-pelykh (commented):
Thanks for this migration work -- the overall structure is solid and the test coverage is good.

A few findings from reviewing the diff:

### Bug: String format key mismatch in `_get_final_discount()`

In `sale_triple_discount/models/sale_order_line.py`, the `_get_final_discount()` method has a string formatting bug that would cause a `KeyError` at runtime:

```python
raise ValidationError(
    _("Sale order line %(name)s has unknown discounting type %(dic_type)s")
    % {"name": self.name, "disc_type": self.discounting_type}
)
```

The format string uses `%(dic_type)s` but the dict key is `disc_type` (missing the `c` swap). This should be either `%(disc_type)s` in the format string or `dic_type` in the dict key. The same typo is propagated to all i18n `.po` files.

### Question: `post_init_hook` vs `pre_init_hook`

I see that @rousseldenis and @legalsylvain previously flagged that the hook should be a `pre_init_hook` rather than `post_init_hook` to avoid losing existing discount values on install to an existing database. @pablo-cort-s73 replied "Hook updated" but the manifest still declares `"post_init_hook": "post_init_hook"` and `hooks.py` still defines `def post_init_hook(env):`.

Is this intentionally kept as `post_init_hook`, or was the update not pushed yet? Given that there's also a `pre-migrate.py` that handles the `discount -> discount1` copy, I wonder if the hook is only needed for fresh installs (where pre-migrate wouldn't run), in which case `post_init_hook` is fine. Could you clarify?

### Note: Additive discount migration formula

In `sale_triple_discount/migrations/18.0.1.0.0/pre-migrate.py`, the additive discount consolidation formula:

```sql
SET discount = 100 * (
    1 - (
        (100 - COALESCE(discount1, 0.0)) / 100
        + (100 - COALESCE(discount2, 0.0)) / 100
        + (100 - COALESCE(discount3, 0.0)) / 100
    )
)
WHERE discounting_type='additive';
```

For additive discounts (d1=20, d2=20, d3=20), this computes `100*(1-(0.8+0.8+0.8))` = `-140`, which is incorrect. The additive formula should simply sum the discounts: `COALESCE(discount1,0) + COALESCE(discount2,0) + COALESCE(discount3,0)`.

I understand additive mode is currently blocked by a constraint and tests are skipped (FIXME #3649), so this is a latent issue -- but worth fixing now to avoid surprises when additive mode is re-enabled.

### Positive observations

- Good test coverage (327 lines) with multiplicative, invoicing, wizard, and pricelist scenarios
- Proper use of `openupgradelib` for migration
- Rebel module CI configuration is correct (grouped with `sale_order_general_discount_triple`)
- `_compute_discount1` using `env.protecting` to avoid side effects on `discount` field is a clean approach
- Migration script correctly handles `discount -> discount1` copy before consolidation

Codecov failures are expected for new module additions and non-blocking.

dreispt:
> ### Note: Additive discount migration formula
> 
> In `sale_triple_discount/migrations/18.0.1.0.0/pre-migrate.py`, the additive discount consolidation formula:
> 
> ```sql
> SET discount = 100 * (
>     1 - (
>         (100 - COALESCE(discount1, 0.0)) / 100
>         + (100 - COALESCE(discount2, 0.0)) / 100
>         + (100 - COALESCE(discount3, 0.0)) / 100
>     )
> )
> WHERE discounting_type='additive';
> ```

I can't find this code in the diff. 
Am i looking at the wrong place?

adriancarmona-aromasdelcampo:
Hi, I have a question: 

How do migrations work? 

I've seen several migrated modules approved, but the merge isn't happening. 

What are the requirements?

rousseldenis:
@pablo-cort-s73 Could you change last commit using `copier update` instead:

- When answering copier questions, put the rebel modules there.
- Commit message is better like : `[UPD] Copier + change discount modules in rebel groups`

Thanks

Bilbonet (approved):
LGTM Thanks

antoniocanovas (changes requested):
It doesn't work for  me, the problem is in the _prepare_invoice_line method in sale_triple_discount/models/sale_order_line.py. It's not correctly passing the main discount as discount1 to the invoice line. The account.move.line model expects discount1, discount2, and discount3, but it's receiving discount, discount2, and discount3.

Perhaps in sale.order.line:
res = super()._prepare_invoice_line(**kwargs)
         res["discount1"] = res.pop("discount")
         res.update({"discount2": self.discount2, "discount3": self.discount3})
         return res

pablo-cort-s73:
> It doesn't work for me, the problem is in the _prepare_invoice_line method in sale_triple_discount/models/sale_order_line.py. It's not correctly passing the main discount as discount1 to the invoice line. The account.move.line model expects discount1, discount2, and discount3, but it's receiving discount, discount2, and discount3.
> 
> Perhaps in sale.order.line: res = super()._prepare_invoice_line(**kwargs) res["discount1"] = res.pop("discount") res.update({"discount2": self.discount2, "discount3": self.discount3}) return res

"Thanks for the feedback. However, we cannot use res["discount1"] = res.pop("discount") because in Odoo 18.0, the discount field is a stored computed field that represents the total accumulated discount (not the first discount).
If we do that, we would pass the total accumulated discount as discount1, and when combined again with discount2 and discount3 in the invoice line, it would apply a double discount.
For example, with discounts of 5% + 9% + 13%:
The SO line computes a total discount of 24.79%.
With your suggestion, the invoice line would receive discount1 = 24.79%, discount2 = 9%, and discount3 = 13%.
The invoice line would then recalculate a total discount of ~40.95% (instead of the correct 24.79%).
The current code correctly pops discount and passes discount1 (which is 5%), discount2 (9%), and discount3 (13%), letting the invoice line compute the 24.79% discount cleanly."

Reyes4711-S73 (approved):
LGTM

rousseldenis:
@pablo-cort-s73 Why the depends on `sell_only_by_packaging` module ?

.pylintrc
```diff
@@ -104,7 +104,11 @@ enable=anomalous-backslash-in-string,
     website-manifest-key-not-valid-uri,
     external-request-timeout,
     missing-manifest-dependency,
+<<<<<<< HEAD
```
rousseldenis: @pablo-cort-s73 Could you fix this ?

Reyes4711-S73:
@rousseldenis Done

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo-test-helper
+odoo-addon-sell_only_by_packaging @ git+https://github.com/OCA/sale-workflow.git@refs/pull/4446/head#subdirectory=sell_only_by_packaging
```
rousseldenis: @pablo-cort-s73 You can also remove the commit for this

Reyes4711-S73: @rousseldenis Commit removed

rousseldenis (approved):

rousseldenis (changes requested):
It seems no tests are running...

@pablo-cort-s73

pablo-cort-s73:
I’ll have a look at it now, @rousseldenis . Thanks.

pablo-cort-s73:
@rousseldenis tests ok again

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,150 @@
...
+    # the computation of the subtotal exact.
+    # For exemple, if discounts are 05%, 09% and 13%
+    # the main discount is 24.7885 % (and not 24.79)
+    discount = fields.Float(
+        string="Total Disc (%)",
+        store=True,
+        compute="_compute_discount",
+        compute_sudo=True,
+        precompute=True,
+    )
```
stenext, line 28:
```suggestion
        digits=None,
    )
```
I found out that `triple.discount.mixin` declares `digits=None`, but it doesn't "win": with
`_inherit = ["sale.order.line", "triple.discount.mixin"]`, core's `sale.order.line.discount (digits="Discount")` precedes the mixin in the MRO.
On a running registry:
`env["sale.order.line"]._fields["discount"]._digits   # 'Discount'   (expected: None)`

Since `price_subtotal` is derived from discount, the aggregate being rounded makes the line total wrong.

In my real case: 

--- | this PR | with digits=None
--- | --- | ---
_fields['discount']._digits | 'Discount' | None
discount1 / 2 / 3 | 5 / 9 / 13 | 5 / 9 / 13
discount (aggregated) | **25.0** | 24.7885
price_unit | 1200.00 | 1200.00
price_subtotal | **900.00** | 902.54

Restoring `_digits is None` (before commit ff42532fe2), `discount1/2/3` keep `digits="Discount"`, so discounts are still displayed with the configured precision.

rousseldenis: @pablo-cort-s73 Do you have insight here ?

PR #3891, opened 2025-09-10
[17.0][FIX] sale_force_invoiced: Don't hide the field

pedrobaeza:
You may want to predefine in draft status to force invoiceable status before confirming the quotation, and in the migration to 17 (#2961), such option was restricted.

@Tecnativa TT57973


pedrobaeza:
cc @siemenv @StefanRijnhart

pedrobaeza:
Uhm, I'm seeing that there's no regression, because the combination with `states` in the field makes that before the field was shown, but not editable, so it's similar. Let's close it then.

PR #3370, opened 2024-10-22
[18.0][MIG] sale_blanket_order: Migration to 18.0

BertVGroenendael:

PR #3892, opened 2025-09-11
[18.0] [MIG] sale_blanket_order: migration to 18.0

mdurepos:
- Replace `analytic_account_id` on `blanket.order.line` by using the new
      `analitic.mixin` from Odoo 18.0.
    - In `_get_display_price`, fall back to simply using price calculation from
      `product.pricelist`, triggered by the fact that `product.pricelist` no longer
       has a `discount_policy` field.
    - Removed `_get_real_price_currency` method as it is now dead code.
    - Removed logic in report template for conditionally including/excluding
      taxes from line subtotals
      (`account.group_show_line_subtotals_tax_excluded` no longer exists).


sbidoul:
Coming here from the post on discord. See also #3370 which seems to do the same, as well the oca/sale-blanket repo which is probably where the energy will be going forward.

mdurepos:
> Coming here from the post on discord. See also #3370 which seems to do the same, as well the oca/sale-blanket repo which is probably where the energy will be going forward.

:( thanks, Stéphane. Don't know how I missed that PR when I searched for it before doing this. Will leave this here for now as I haven't had time to look at the other PR or sale-blanket, but all good if this gets closed.

rousseldenis:
So, closing this in favor of https://github.com/OCA/sale-workflow/pull/3370

PR #3894, opened 2025-09-11
[16.0][IMP] sale_order_line_cancel: Refactor and create base addon

mt-software-de:
This PR moves the base functionality of `sale_order_line_cancel` into an own addon `sale_order_line_cancel_base`.
This allows it, to use it also without stock modules. 

Within this refactoring i changed the logic, when the update of `product_qty_canceled`, before it was triggered by canceling a stock move. Now it is directly triggered when someone cancels the remaining qty. 
This leads to the improvement i wanted, you can cancel the remaining qty of sale.order.lines without a stock.move. 
This happens when a sale.order.line creates a not confirmed purchase.order. 

`sale_order_line_cancel` will be now only adding the functionality to cancel the stock.moves when the cancel on the sale.order.lines is triggered or the other way around. 

IMO: the naming of the addons should be `sale_order_line_cancel` and `sale_order_line_cancel_stock`, but i don't know if this would be good since `sale_order_line_cancel` is already in use. 

A next task would be to create a addon called `sale_order_line_cancel_purchase`.


mt-software-de:
ping @jbaudoux @rousseldenis @lmignon @raumschmiede-joshuaL @i-vyshnevska

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -4,79 +4,27 @@
...
-            "Product Unit of Measure"
-        )
-        for rec in self:
-            rec.can_cancel_remaining_qty = (
-                float_compare(
-                    rec.product_qty_remains_to_deliver, 0, precision_digits=precision
-                )
-                == 1
-                and rec.state in ("sale", "done")
-                and rec.qty_delivered_method == "stock_move"
```
mt-software-de, line 39: This line is now gone, imo it is not needed to check for `qty_delivered_method`.

sale_order_line_cancel/models/stock_move.py
```diff
@@ -29,18 +29,17 @@ def _action_done(self, cancel_backorder=False):
     def _get_sale_lines_to_update_qty_canceled(self):
         sale_lines = self.env["sale.order.line"]
         for move in self:
-            if (
-                move.sale_line_id
```
mt-software-de, line 33: This check was removed it is already in `_is_move_to_take_into_account_for_qty_canceled`

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -4,79 +4,27 @@
...
 
     def _check_moves_to_cancel(self, moves):
         """Override this method to add checks before cancel"""
-        self.ensure_one()
-
-    def _update_qty_canceled(self):
-        """Update SO line qty canceled only when all remaining moves are canceled"""
-        for line in self:
-            if line._get_moves_to_cancel():
-                continue
```
mt-software-de, line 60: By moving this method i removed also the check, when the method is called, it will be updating `product_qty_canceled`. The check if it shouldn't updated in case of multiple moves on a line, was moved to `stock.move`. Because only then it is necessary to check. If `_update_qty_canceled` is called from `cancel_remaining_qty` it is not needed to check, because we are canceling all open moves.

sale_order_line_cancel_base/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Sylvain Van Hoof <sylvain@okia.be>
+* Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+* Souheil Bejaoui <souheil.bejaoui@acsone.eu.com>
```
lmignon:
```suggestion
* Souheil Bejaoui <souheil.bejaoui@acsone.eu.com>
* Laurent Mignon <laurent.mignon@acsone.eu>
```

sale_order_line_cancel/models/stock_move.py
```diff
@@ -29,18 +29,17 @@ def _action_done(self, cancel_backorder=False):
...
         self.ensure_one()
+        sale_line = self.sale_line_id
         return (
-            self.state == "cancel"
-            and self.sale_line_id
-            and self.sale_line_id.state not in ["draft", "sent"]
+            sale_line
+            and self.state == "cancel"
             and self.picking_type_id.code == "outgoing"
+            and sale_line.state not in ["draft", "sent", "cancel"]
```
mt-software-de, line 41: This ensures that i can not be called for a canceled order line.

sale_order_line_cancel/models/stock_move.py
```diff
@@ -29,18 +29,17 @@ def _action_done(self, cancel_backorder=False):
...
+            and not sale_line._get_moves_to_cancel()
```
mt-software-de, line 42: This check ensures that there are no other open move lines, like in a mrp scenario.

sale_order_line_cancel_base/readme/CONTRIBUTORS.rst
mt-software-de: i will add you as well to `sale_order_line_cancel`

sale_order_line_cancel/__manifest__.py
```diff
@@ -7,18 +7,12 @@
 {
     "name": "Sale Order Line Cancel",
     "version": "16.0.1.3.1",
```
vuwnevska, line 9:
```suggestion
    "version": "16.0.2.0.0",
```

vuwnevska (approved):

sale_order_line_cancel/__manifest__.py
lmignon: @vuwnevska This change is done at merge time...

sale_order_line_cancel/models/sale_order_line.py
lmignon: @mt-software-de Why do you replace everywhere the use of `line.qty_to_deliver` by `line.product_uom_qty - line.qty_delivered` ???

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -4,79 +4,27 @@
...
-            qty_remaining = max(0, line.qty_to_deliver - line.product_qty_canceled)
-            line.product_qty_remains_to_deliver = qty_remaining
-
     def _get_moves_to_cancel(self):
-        lines = self.filtered(lambda l: l.qty_delivered_method == "stock_move")
+        lines = self.filtered(
+            lambda l: l.qty_delivered_method == "stock_move"
+            and l.can_cancel_remaining_qty
+        )
         return lines.move_ids.filtered(lambda m: m.state not in ("done", "cancel"))
```
lmignon: On a performance point of view, this code iterates 2 times on the lines.... We should do it on 1 loop.

lmignon (commented):
The changes seem consistent overall. I have a quick question about the reason for no longer using the ‘qty_to_deliver’ attribute...
As for the module names, I find them counterintuitive. I would prefer to keep the basic functionality in `sale_order_line_cancel` and the code related to stock in `sale_stock_sale_order_line_cancel`. This seems to me to follow the logic adopted by Odoo.

sale_order_line_cancel/models/sale_order_line.py
mt-software-de: Yes of course. I just copied the whole Module. And did Theo necessary Changes. 👍🏼

sale_order_line_cancel/models/sale_order_line.py
mt-software-de: Because it is not available with just sale as dependency.

mt-software-de:
> The changes seem consistent overall. I have a quick question about the reason for no longer using the ‘qty_to_deliver’ attribute...

Because by changing the dependency to only sale in the base addon, it is not available anymore.

mt-software-de:
> As for the module names, I find them counterintuitive. I would prefer to keep the basic functionality in `sale_order_line_cancel` and the code related to stock in `sale_stock_sale_order_line_cancel`. This seems to me to follow the logic adopted by Odoo.

Yes thats what i have written in the description of this PR. The only concern i have about this, whats about other user which are using this addon already? If we are changing the name the initial functionality isn't available anymore only if they know to install the renamed addon.

sale_order_line_cancel/models/sale_order_line.py
mt-software-de: Sorry no. It doesn't, or? :-)
At first it iterates on lines and then on the moves.

rousseldenis: @lmignon Not sure, as at `return lines.move_ids...` lines are recordset. No more iteration

rousseldenis:
> > As for the module names, I find them counterintuitive. I would prefer to keep the basic functionality in `sale_order_line_cancel` and the code related to stock in `sale_stock_sale_order_line_cancel`. This seems to me to follow the logic adopted by Odoo.
> 
> Yes thats what i have written in the description of this PR. The only concern i have about this, whats about other user which are using this addon already? If we are changing the name the initial functionality isn't available anymore only if they know to install the renamed addon.

Agree with @lmignon  on that.

mt-software-de:
> > > As for the module names, I find them counterintuitive. I would prefer to keep the basic functionality in `sale_order_line_cancel` and the code related to stock in `sale_stock_sale_order_line_cancel`. This seems to me to follow the logic adopted by Odoo.
> > 
> > Yes thats what i have written in the description of this PR. The only concern i have about this, whats about other user which are using this addon already? If we are changing the name the initial functionality isn't available anymore only if they know to install the renamed addon.
> 
> Agree with @lmignon  on that.

OK perfect then i will rename the modules

rousseldenis (approved):

rousseldenis:
@lmignon

lmignon (approved):

PR #3897, opened 2025-09-12
[IMP] sale_triple_discount: compute untaxed_amount_to_invoice considering multiple discounts

Chionne27:


HekkiMelody (approved):
Code review, LGTM

alexey-pelykh (commented):
## Review: `sale_triple_discount` -- untaxed_amount_to_invoice with multiple discounts

**PR title:** [IMP] sale_triple_discount: compute untaxed_amount_to_invoice considering multiple discounts

**CI Status:** All green (pre-commit, test with OCB/Odoo, codecov patch/project, unreleased dependency check).

**Existing review:** One APPROVED by HekkiMelody (2025-09-23).

---

### Summary

This PR fixes a real bug where Odoo 14.0's `_compute_untaxed_amount_to_invoice` hardcodes `line.discount` to compute the reduced price (`price_reduce = line.price_unit * (1 - (line.discount or 0.0) / 100.0)`), ignoring `discount2` and `discount3`. The approach correctly reuses the established `triple_discount_preprocess`/`triple_discount_postprocess` pattern already used by `_compute_amount` and `_get_price_reduce`, which temporarily collapses all three discounts into the `discount` field so the super method produces the right value.

The version bump to `14.0.3.0.1` is appropriate (major segment increment for a behavioral change). The `post_init_hook` + migration script ensure existing data is recomputed on install and upgrade.

---

### Findings

#### 1. Missing `discounting_type` in `@api.depends` (informational)

The new method uses:
```python
@api.depends(lambda self: self._discount_fields())
def _compute_untaxed_amount_to_invoice(self):
```

This resolves to `["discount", "discount2", "discount3"]` and does not include `discounting_type`. When a user changes the discounting type from multiplicative to additive (or vice versa), the `untaxed_amount_to_invoice` field will not be recomputed.

However, this is **consistent with `_get_price_reduce`** which uses the same `@api.depends(lambda self: self._discount_fields())` without `discounting_type`. Only `_compute_amount` adds `discounting_type` explicitly. So this is a pre-existing gap in the module, not introduced by this PR. Flagging for awareness only.

#### 2. No test coverage for `untaxed_amount_to_invoice` (suggestion)

The existing test suite validates `price_subtotal`, `amount_untaxed`, `amount_tax`, and `amount_total`, but does not test `untaxed_amount_to_invoice` on confirmed sale orders with triple discounts. A test that confirms a sale order with triple discounts and checks `so_line.untaxed_amount_to_invoice` would guard against regressions. This is an improvement suggestion, not a blocker, especially since CI passes with codecov coverage success.

#### 3. Migration script: `registry` parameter receives `env` (minor, harmless)

In `migrations/14.0.3.0.1/post-migrate.py`:
```python
def migrate(env, version):
    post_init_hook(env.cr, env)
```

The hook signature is `post_init_hook(cr, registry)` but `registry` receives the ORM `env`. This works because the hook never uses `registry` -- it builds its own `env` from `cr`. This is a common OCA pattern and is functionally correct, though slightly imprecise semantically.

#### 4. `readme/DESCRIPTION.rst` missing trailing newline (minor)

The file ends without a trailing newline (`\ No newline at end of file`). POSIX convention expects a trailing newline. Minor nit.

---

### Verdict: APPROVE

The approach is sound, consistent with existing module patterns, version bump is correct, migration is handled, and CI is fully green. The core logic (preprocess/super/postprocess) is the proven pattern in this module. The missing `discounting_type` dependency is a pre-existing gap, and the absent test for `untaxed_amount_to_invoice` is an improvement opportunity but not a blocker.

Good contribution.

PR #3898, opened 2025-09-15
[17.0][IMP] sale_order_product_recommendation: use Product Price decimal precision for price_unit

andrel-exo:
Before this change, the wizard field `price_unit` was defined as a Monetary field, which always respects the currency's decimal places (typically 2).  
This caused inconsistencies when the database has a higher decimal precision defined for product prices.

With this change, `price_unit` is now a Float field using `digits="Product Price"`.  
This makes the wizard consistent with other places in Odoo (e.g. `sale.order.line.price_unit`) that already rely on the Product Price decimal precision.


gurneyalex (approved):

sergio-teruel (approved):

yajo (approved):
/ocabot merge minor

PR #3903, opened 2025-09-18
[18.0] [MIG] sale_invoice_split_payment: Migration to 18.0

luisDIXMIT:
Standard migration.


sale_invoice_split_payment/tests/test_sale_order.py
```diff
@@ -28,10 +28,7 @@ def setUp(self):
                     (
                         0,
                         0,
-                        {
```
rousseldenis: @luisDIXMIT Why removing this ?

luisDIXMIT: Because, I got many errors due to the changes of those fields and their values on version 18.0, and I considered that these values aren't necessary for the execution of the test. The important thing is to create 2 different payment_terms and split invoices when invoicing sales with these payment_terms. The content of the payment_terms isn't relevant, IMHO.

etobella (approved):
LGTM

@rousseldenis Fixed the tests to minimize changes 😉

rousseldenis (approved):

PR #3906, opened 2025-09-24
[16.0][FIX] sale_stock_picking_blocking: Do not compute blocking_id on create()

### LINKED ISSUE #3815: [18.0] sale_stock_picking_blocking saving the SO the first time removes blocking reason
## Module
sale_stock_picking_blocking

## Describe the bug
Since  autosave was introduced in V16.0, when either manually saving an SO for the first time or when releasing it directly without saving it before, the blocking reason is removed during saving (and in case of directly confirming, the Delivery is created). Only when editing a previously saved SO, the block reason stays when saving again.

## To Reproduce
16.0-18.0:

1.Create SO, add partner, product and sale delivery blocking reason
2.Click on save

**Expected behavior**
The block reason should stay when saving the first time.

ntsirintanis:
This tackles this bug:

https://github.com/OCA/sale-workflow/issues/3815

for version 16.0

### DIFF
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -14,6 +14,7 @@ class SaleOrder(models.Model):
         string="Delivery Block Reason",
         compute="_compute_delivery_block_id",
         store=True,
+        precompute=True,
         states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
     )
 


NL66278 (approved):
👍 LGTM

sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -45,6 +45,17 @@ def action_remove_delivery_block(self):
             order_to_unblock.action_done()
         return True
 
+    @api.model_create_multi
+    def create(self, vals_list):
```
rousseldenis: @ntsirintanis Could you try adding `precompute=True` to the field instead ?

ntsirintanis: @rousseldenis thank you very very much, it works. But why?

rousseldenis: Because... it precomputes field value before creating record in DB. :sweat_smile: 

https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L232

NL66278:
@ntsirintanis Please squash commits, as the second completely undoes the first and then adds the concise solution.

rousseldenis (approved):

PR #3907, opened 2025-09-26
[16.0][REF] sale_triple_discount,sale_fixed_discount,sale_order_general_discount: Compatiblity

hildickethan:
- Depends on https://github.com/OCA/account-invoicing/pull/2109

Supersede/followup of the changes in https://github.com/OCA/sale-workflow/pull/2885
Based on the same changes I proposed for the invoice modules in https://github.com/OCA/account-invoicing/pull/2109 and talked about in this issue https://github.com/OCA/sale-workflow/issues/3638

The main objective is to finish the refactor process that was started by @grindtildeath that got stuck halfway between merged in `account-invoicing` and a pending PR for `sale-workflow` leading to the repo being stuck in limbo for the past months.

Things this PR accomplishes that I tested and they seem to be working better:
 - Discount1 field for `sale_triple_discount` instead of using the base `discount` directly, helping with compatibility between the modules and simplifying the code.
- Create a new glue module between `sale_triple_discount` and `sale_fixed_discount` called `sale_fixed_triple_discount`
   - Control the partial incompatibility between `discount_fixed` and triple discount fields, as of right now they don't work together if you use both on the same line and will warn the user with a yellow label rather than silently just not work in the background. However using them on different lines even if it's the same order should now work correctly
 - Adapt the `sale_pricelist_triple_discount`, `sale_order_general_discount` and `sale_order_general_discount_triple` modules to use the new field
 - Adapt the way some reports change things so they don't step on each others fingers

All this is mostly to clean up the current state of the repo and be able to merge the 17.0 and 18.0 migration PRs that are blocked by the refactor still pending in 16.0, but after seeing how many modules there are related to discounts (triple, general, global and fixed) I would like to propose an even bigger refactor that consolidates some of the modules, gets rid of some redundancy and restructures them so we don't have to make a web of glue modules between each combination you have installed like we do now. This of course would way out of scope for "stable" versions but now that we're still in early stages of 19.0 I think it should be seriously considered for 19.0+.

I would also encourage anyone that actively uses the modules to check any of these changes to see if there would be any missing migration scripts, as I don't currently use this version to test all the fields properly on real data


hildickethan:
~~it seems the pre-commit changes are just javascript errors presumably due to the copier template update, not sure what do about those~~ omitted the `.eslintrc` changes from the template update

lmignon (commented):
Thank you @hildickethan for all this hard work. I functionally tested sale_triple_discount and reviewed the code and everything LGTM. Unfortunately, I've no time to fully review the PR.

hildickethan:
I've also been using the 17.0 version of this PR since its creation and haven't had any issues

grindtildeath:
Thanks for taking this over @hildickethan :hugs:

alexey-pelykh (commented):
## Review of `sale_triple_discount` / `sale_fixed_discount` compatibility refactor

Thanks for taking on this substantial refactor, @hildickethan. The overall direction is solid: introducing `discount1` via the `triple.discount.mixin`, creating the `sale_fixed_triple_discount` glue module, and cleaning up the three separate `_compute_discount*` methods in `sale_order_general_discount_triple` into a single `_compute_discount1`. Below is a structured review.

---

### Critical: Merge conflict markers in `README.md`

The root `README.md` contains multiple unresolved git merge conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`). These appear at several locations in the module table. This must be resolved before merge -- the file is currently broken.

Examples:
```
<<<<<<< HEAD
[account_invoice_pricelist_technical](...)
=======
>>>>>>> ad04a0d4 ([ADD] sale_fixed_triple_discount: Glue module)
```

---

### Critical: `test-requirements.txt` pins to a PR head ref

The two new entries in `test-requirements.txt` pull dependencies from an unmerged PR branch:
```
odoo-addon-account_invoice_triple_discount @ git+https://github.com/OCA/account-invoicing.git@refs/pull/2109/head#...
odoo-addon-account_invoice_fixed_discount @ git+https://github.com/OCA/account-invoicing.git@refs/pull/2109/head#...
```

This is acceptable during development/CI testing but should not be merged to `16.0` until OCA/account-invoicing#2109 is merged first, since `refs/pull/2109/head` is unstable (force pushes change the ref). The PR description already notes this dependency -- just confirming it should be tracked as a merge-order gate.

---

### Moderate: Variable naming `move` in `sale_fixed_triple_discount/models/sale_order.py`

In `_compute_show_discount_warning_label`, the loop variable is named `move`:
```python
for move in self:
    move.show_discount_warning_label = any(...)
```

This is a `sale.order` model, not `account.move`. The conventional name here would be `order`. This is a minor readability issue but worth correcting since it could confuse future contributors.

---

### Moderate: `hooks.py` uses deprecated `pkg_resources.parse_version`

`pkg_resources` is deprecated in favor of `importlib.metadata` / `packaging.version.Version`. While this is not blocking for Odoo 16.0 (which still ships with setuptools), it would be good to use `packaging.version.Version` instead for forward compatibility:
```python
from packaging.version import Version
if not Version(manifest["version"]) >= Version("16.0.2.0.0"):
```

---

### Moderate: `_get_lines_to_compute_discount` filtering logic

In `sale_fixed_triple_discount/models/sale_order_line.py`:
```python
def _get_lines_to_compute_discount(self):
    lines = super()._get_lines_to_compute_discount()
    return lines.filtered(
        lambda line: float_is_zero(
            line.discount_fixed, precision_rounding=line.currency_id.rounding
        )
    )
```

This filters OUT lines that have a non-zero `discount_fixed`, which means those lines skip the triple discount computation. The intent is correct (fixed discount takes priority), but the method name `_get_lines_to_compute_discount` is inherited from the parent and its contract is not immediately obvious. A brief docstring explaining the filtering rationale would help.

---

### Minor: `_compute_discount1` uses `env.protecting` pattern

In `sale_triple_discount/models/sale_order_line.py`, `_compute_discount1` uses:
```python
with self.env.protecting(
    [self.env["sale.order.line"]._fields["discount"]], self
):
    self.with_context(skip_triple_discount=True)._compute_discount()
    for line in self:
        line.discount1 = line.discount
```

This is a clever approach to drag the base discount into `discount1` without persisting the intermediate value. The inline comments explain the reasoning well. The `FIXME` comment in the test about `waiting_approval` state is also appreciated.

---

### Minor: `_compute_discount` reset logic

In `sale_triple_discount/models/sale_order_line.py`:
```python
for line in self:
    if not (
        line.order_id.pricelist_id
        and line.order_id.pricelist_id.discount_policy == "without_discount"
    ):
        line.discount = 0
```

The comment explains this is needed because base Odoo `continue`s instead of resetting to 0. This is a valid workaround but somewhat fragile -- if base Odoo changes this behavior in a patch, it could cause double-zeroing. Consider adding a version-specific comment noting which Odoo behavior this compensates for.

---

### CI Status Notes

- **pre-commit**: PASS
- **test with Odoo / OCB** (multiple matrices): PASS
- **Detect unreleased dependencies**: FAIL (expected, since this PR depends on OCA/account-invoicing#2109 which is not yet released)
- **codecov/patch**: FAIL (coverage delta, non-blocking for OCA)
- **runboat/build**: PASS

The CI failures are expected given the cross-repo dependency. Once OCA/account-invoicing#2109 is merged, the unreleased dependency check should pass.

---

### Summary

| Category | Finding |
|----------|---------|
| **Must fix** | Resolve merge conflict markers in `README.md` |
| **Must fix (merge order)** | Ensure OCA/account-invoicing#2109 merges first |
| **Should fix** | Rename `move` -> `order` in `sale_fixed_triple_discount/models/sale_order.py` |
| **Should fix** | Replace deprecated `pkg_resources.parse_version` with `packaging.version.Version` |
| **Nice to have** | Add docstring to `_get_lines_to_compute_discount` override |
| **Nice to have** | Add Odoo version note to the `discount = 0` reset workaround |

Overall the refactor is well-structured, the new `sale_fixed_triple_discount` glue module is a clean approach, the migration script correctly copies `discount` to `discount1`, and the test coverage for the new module and the updated tests for existing modules are thorough. Good work.

lmignon:
`sale_triple_discount` works like a charm in production. Thank you @hildickethan !

rousseldenis:
@grindtildeath Could you review this ?

PR #3909, opened 2025-09-28
[14.0][FIX] sale_manual_delivery: use _for_xml_id instead of env.ref for safer XML ID resolution

GSLabIt:
Fix issue related to action reference


rousseldenis (approved):

rousseldenis:
Merging this as trivial and for old version.

/ocabot merge patch

PR #3915, opened 2025-09-29
[18.0][FIX] sale_order_carrier_auto_assign: set carrier on create - create empty SO then add product line

jbaudoux:
When a blank SO is created and then a product line is added, the carrier should also be set when `carrier_on_create` is enabled. This is because the carrier is not set on create when there are no product line. So while there is no carrier on the SO and  `carrier_on_create` is enabled, try to set a carrier on write.

cc @pedrobaeza @victoralmau @grindtildeath @sebalix

### DIFF
--- a/sale_order_carrier_auto_assign/models/sale_order.py
+++ b/sale_order_carrier_auto_assign/models/sale_order.py
@@ -28,11 +28,26 @@ class SaleOrder(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        orders = super().create(vals_list)
-        for order in orders:
-            if not order.carrier_id and order._is_auto_set_carrier_on_create():
-                order._set_delivery_carrier()
+        ctx_carrier_on_create = self.env.context.get("carrier_on_create")
+        orders = super(SaleOrder, self.with_context(carrier_on_create=True)).create(
+            vals_list
+        )
+        orders = orders.with_context(carrier_on_create=ctx_carrier_on_create)
+        orders._set_carrier_on_create()
         return orders
+
+    def _set_carrier_on_create(self):
+        if self.env.context.get("carrier_on_create"):
+            return
+        for order in self:
+            if not order.carrier_id and order._is_auto_set_carrier_on_create():
+                order.with_context(carrier_on_create=True)._set_delivery_carrier()
+
+    def write(self, vals):
+        # When product lines are added, set the carrier
+        res = super(SaleOrder, self.with_context(carrier_on_create=True)).write(vals)
+        self._set_carrier_on_create()
+        return res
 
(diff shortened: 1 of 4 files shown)


sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -29,11 +29,22 @@ def _is_auto_set_carrier_on_create(self):
...
+    def _set_carrier_on_create(self):
+        if self.env.context.get("carrier_on_create"):
+            return
+        for order in self:
+            if not order.carrier_id and order._is_auto_set_carrier_on_create():
+                order.with_context(carrier_on_create=True)._set_delivery_carrier()
+
+    def write(self, vals):
+        # When product lines are added, set the carrier
+        res = super(SaleOrder, self.with_context(carrier_on_create=True)).write(vals)
```
pedrobaeza, line 48: Maybe we should do this on `sale.order.line` instead?

twalter-c2c (approved):
LGTM

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: fixed

sebalix (approved):

sale_order_carrier_auto_assign/models/sale_order.py
pedrobaeza: But now is on both and it seems redundant.

jbaudoux: It's on purpose. If you access SO, the carrier computation occurs when all lines are added and not when the first SO line is added

rousseldenis (approved):

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -28,12 +28,27 @@ def _is_auto_set_carrier_on_create(self):
...
+        self._set_carrier_on_create()
+        return res
```
hildickethan, line 50: should this not check for `order_line` similar to  the `if vals.get("product_id"):` in `sale.order.line`?

jbaudoux: indeed, I cherry-picked this fix on 17.0
https://github.com/OCA/sale-workflow/pull/3961

HaraldPanten:
@jbaudoux Do you think this one is ready to be merged?

THX

jbaudoux:
> @jbaudoux Do you think this one is ready to be merged?
> 
> THX

For me yes

HaraldPanten:
@pedrobaeza could we merge this one, please?

PR #3916, opened 2025-09-30
[15.0][ADD] sale_stock_warehouse_partner: New module

eduezerouali-tecnativa:
cc @Tecnativa TT58152


sale_stock_warehouse_partner/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+            elif order.partner_id.warehouse_id:
+                order.warehouse_id = order.partner_id.warehouse_id
+            else:
+                order.warehouse_id = order.user_id._get_default_warehouse_id()
+
+    def check_sale_order_type(self):
+        sale_order_type_module = self.env["ir.module.module"].search(
+            [("name", "=", "sale_order_type")]
+        )
+        if sale_order_type_module or sale_order_type_module.state == "installed":
```
carlosdauden: Use registry instead of search to minimize DB calls.
```suggestion
        if "sale_order_type" in self.env.registry._init_modules and "sale_stock_warehouse_order_type" not in self.env.registry._init_modules:
```

carlosdauden (changes requested):

sale_stock_warehouse_partner/models/sale_order.py
eduezerouali-tecnativa: No longer that need it, but good to know.

sale_stock_warehouse_partner/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    warehouse_id = fields.Many2one(
```
rousseldenis: This should be company dependent IMHO

sergio-teruel: @eduezerouali-tecnativa Can you attend this comment?? Thanks..

eduezerouali-tecnativa: @rousseldenis You're right, makes it more flexible for multi-company scenario. Changes done. Thanks!

sale_stock_warehouse_partner/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+        comodel_name="stock.warehouse",
+        company_dependent=True,
+        string="Warehouse",
+        help="Set deafult warehouse for Sale Order",
```
rousseldenis:
```suggestion
        help="Set default warehouse for Sale Order",
```

sale_stock_warehouse_partner/models/res_partner.py
rousseldenis: One more thing, could you change the field name like `sale_warehouse_id` ? Thanks

rousseldenis: Here, not on sale.order level :sweat_smile:

eduezerouali-tecnativa:
@rousseldenis changes done, could you review please.

rousseldenis (approved):
LGTM

eduezerouali-tecnativa:

sergio-teruel (approved):

carlosdauden (approved):

Issue #3864, opened 2025-08-18
[18.0] Smart button on sale order with link to purchase orders is not shown


vince-dynapps:
## Module
sale_procurement_group_by_line

## Describe the bug
If you confirm a sales order with MTO (make to order) products, Odoo will automatically create a purchase order and show a smart button on the sales order with al ink to the created purchase order(s). This button disappears if you install the sale_procurement_group_by_line module.

The following function computes the purchase orders to show in the smart button:
```
    def _get_purchase_orders(self):
        self.ensure_one()
        linked_po = self.procurement_group_id.stock_move_ids.created_purchase_line_ids.order_id \
                  | self.env['stock.move'].browse(self.procurement_group_id.stock_move_ids._rollup_move_origs()).purchase_line_id.order_id
        group_po = self.procurement_group_id.purchase_line_ids.order_id

        return linked_po | group_po
```
If I install the sale_procurement_group_by_line module, the procurement_group_id will be stored on the individual sale order lines so the function above needs to be rewritten to take this into account.

## To Reproduce
18.0

1. Configure a product with the MTO/Buy routes enabled. Make sure a vendor is configured so Odoo knows for which vendor it needs to create a purchase order
2. Make a sales order to sell the configured product
3. Confirm the sales order and verify a purchase order is generated by Odoo. The smart button should show the created purchase orders but it doesn't.

**Expected behavior**
I would expect that the standard Odoo functionality keeps working. The smart button with a link to purchase orders should still be visible on the sales order.

**Additional context**

Niharika062:
Hello, I’m really interested in contributing to this project. Please assign me this

PR #3919, opened 2025-09-30
[18.0][FIX] sale_procurement_group_by_line: Fix smart button on sale order with link to purchase orders is not shown.

Jay2163:
**Problem:**

When the module sale_procurement_group_by_line is installed, the procurement group is stored at the sale order line level instead of directly on the sale order.
However, the smart button that shows related purchase orders on the sales order form relies on order_id.procurement_group_id.
As a result, the smart button disappears or shows no purchase orders, even though purchase orders are generated correctly.

**Solution:**

Added the following line in _action_launch_stock_rule:

**line.order_id.procurement_group_id = group_id**

This ensures that the procurement group assigned to the line is also propagated to the parent sale order.
Now, the smart button correctly retrieves and displays purchase orders when sale_procurement_group_by_line is active.

**Impact:**

Restores expected behavior of the Purchase Orders smart button on sales orders.

Fixes issue [#3864](https://github.com/OCA/sale-workflow/issues/3864?utm_source=chatgpt.com).

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -67,6 +67,7 @@ class SaleOrderLine(models.Model):
             if not group_id:
                 vals = line._prepare_procurement_group_vals()
                 group_id = self.env["procurement.group"].create(vals)
+                line.order_id.procurement_group_id = group_id
             else:
                 # In case the procurement group is already created and the
                 # order was cancelled, we need to update certain values


sale_procurement_group_by_line/model/sale.py
```diff
@@ -67,6 +67,7 @@ def _action_launch_stock_rule(self, previous_product_uom_qty=False):
             if not group_id:
                 vals = line._prepare_procurement_group_vals()
                 group_id = self.env["procurement.group"].create(vals)
+                line.order_id.procurement_group_id = group_id
```
rousseldenis, line 70: @Jay2163 IMHO, that's not correct as lines in the same order can have a different procurement group.

The way domain is computed on button action should be checked instead

rousseldenis (changes requested):

sale_procurement_group_by_line/model/sale.py
alexey-pelykh, line 70: This assignment is the core problem. When a sale order has multiple lines, this loop iterates over each line. Each iteration overwrites `order.procurement_group_id` with the current line's group. Only the last line's procurement group survives on the order header.

Additionally, this only fires inside the `if not group_id:` branch (new group creation). If an existing group is reused via the `groups` dict lookup above, this line is never reached — making the fix partial.

The correct approach, as @rousseldenis noted, is to fix the smart button's domain or the method that computes purchase orders so it gathers procurement groups from all order lines, rather than writing a single group back to the order header.

alexey-pelykh (commented):
## Review of sale_procurement_group_by_line smart button fix

### Summary

I agree with @rousseldenis's earlier feedback. The current approach of writing `line.order_id.procurement_group_id = group_id` at line 70 is semantically incorrect and introduces a data integrity issue. The whole purpose of this module is to store procurement groups **per sale order line**, yet the proposed fix back-propagates one line's group to the order header, which can only hold a single value.

### Detailed Analysis

**Root cause of the issue (#3864):**

In standard Odoo 18, the smart button computes purchase orders via:

```python
def _get_purchase_orders(self):
    return self.order_line.purchase_line_ids.order_id
```

This traverses `sale.order` -> `sale.order.line` -> `purchase.order.line` -> `purchase.order` through the `purchase_line_ids` M2M relationship. If the `sale_purchase` module's `_get_purchase_orders` method works this way in the version being targeted, the smart button should already work — the issue may lie in a different code path or a domain filter on the action that references `procurement_group_id` at the order level. The exact path causing the breakage needs investigation.

**Why the proposed fix is problematic:**

1. **Data corruption for multi-line orders:** Consider a sale order with 3 lines, each receiving its own procurement group (PG-A, PG-B, PG-C). The code iterates lines sequentially, so `order.procurement_group_id` would be overwritten with each line's group. Only the last line's group survives on the order header. This silently loses the association to the other groups.

2. **Contradicts the module's core design:** The module exists specifically so that `procurement_group_id` lives on lines, not the order. Writing it back to the order defeats this design and could cause subtle breakage in other modules that inspect `order.procurement_group_id`.

3. **Only covers the `create` branch:** The assignment is inside the `if not group_id:` block (new group creation), so it does not execute when an existing group is reused. This means the fix is partial even within its own logic.

### Recommended Approach

As @rousseldenis suggested, the correct fix is to override the method or action domain that computes the purchase order smart button, so it collects procurement groups from **all order lines** rather than relying solely on the order-level `procurement_group_id`.

A possible direction (to be validated against the actual Odoo 18 code paths):

```python
# In a sale.order model extension within this module:
    _inherit = "sale.order"

    def _get_purchase_orders(self):
        # Collect POs from line-level procurement groups
        return self.order_line.purchase_line_ids.order_id
```

Or if the issue is in the action domain rather than `_get_purchase_orders`, the domain should be adjusted to search purchase orders linked through the line-level `procurement_group_id` fields:

```python
groups = self.order_line.mapped('procurement_group_id')
# Use groups to build a proper domain
```

The contributor should trace exactly which code path the smart button uses to identify the correct fix location.

### CI Status

All checks pass (pre-commit, test with Odoo, test with OCB, codecov). However, the existing test suite does not cover the purchase order smart button scenario when `sale_purchase` is also installed, so passing CI does not validate correctness here.

### Verdict

The approach needs reworking per the maintainer's feedback before it can be merged.

rousseldenis:
@Jay2163

PR #3921, opened 2025-10-01
[14.0][ADD] sale_order_tax_breakdown: show tax breakdown in sale order form

juancarlosonate-tecnativa:
@tecnativa TT56706
@sergio-teruel @carlosdauden

image


carlosdauden (approved):

sergio-teruel (approved):

sale_order_tax_breakdown/models/sale.py
```diff
@@ -0,0 +1,50 @@
...
+            for line in order.order_line:
+                price_reduce = line.price_unit * (1.0 - line.discount / 100.0)
+                taxes = line.tax_id.compute_all(
+                    price_reduce,
+                    quantity=line.product_uom_qty,
+                    product=line.product_id,
+                    partner=order.partner_shipping_id,
+                )["taxes"]
+                for tax in line.tax_id:
+                    tax_name = tax.name or tax.tax_group_id.name
```
alexey-pelykh, line 28: **Grouping key difference from core:**

In Odoo core, the grouping key is `tax.tax_group_id` (the tax group *object*), and the display name comes from `group.name`. Here, you use `tax.name or tax.tax_group_id.name` as both the grouping key and the display name.

This means if two *different* taxes happen to have the same `name` string (e.g., two taxes both named "VAT 21%" but with different configurations), their amounts would be incorrectly merged. This is arguably an edge case, but grouping by `tax.id` (or the tax record itself) and then using `tax.name` only for display would be safer.

Also, `tax.name` is a required field in `account.tax`, so the `or tax.tax_group_id.name` fallback should never actually trigger. Consider whether the fallback adds value or just adds confusion.

alexey-pelykh (commented):
## Review: sale_order_tax_breakdown

Thank you for this module. The idea of showing a per-tax breakdown (rather than per-tax-group) on the sale order form is useful. I have a few observations and one concern worth discussing.

### Key observations

**1. Behavior change vs. Odoo core:** The original `_amount_by_group` groups amounts by `tax.tax_group_id` (the tax group object), while this override groups by individual `tax.name` (or falls back to `tax.tax_group_id.name`). This means two taxes in the same tax group will now appear as separate rows instead of being aggregated into one. This is the intended "breakdown" behavior, but it is worth calling out explicitly in the module description so users understand the difference.

**2. Target branch is 14.0:** This PR targets the `14.0` branch, which has been end-of-life for some time. Is there a reason this is not targeting a more recent branch (e.g., 16.0 or 18.0)? Note that in Odoo 15+, `amount_by_group` was replaced with `tax_totals_json` / `tax_totals`, so this approach would need significant rework for newer versions anyway.

**3. Missing `account` in dependencies:** The view uses `widget="tax-group-custom-field"` which is registered by the `account` module. While `account` is available transitively (`sale` -> `payment` -> `account`), relying on transitive dependencies is fragile. If Odoo ever refactors the dependency chain, this module would break. Consider adding `"account"` to `depends` explicitly, or at minimum documenting the assumption.

**4. No `pre_init_hook` / `post_init_hook` concerns:** Since this is a pure compute override with no stored fields or data migration, there are no migration concerns. Good.

**5. Tests cover the core scenario** of two different taxes producing separate breakdown entries, which is appropriate.

Overall the module is small, focused, and the code closely mirrors the original Odoo implementation with a targeted change. The two existing OCA member approvals and passing CI (tests + pre-commit) are positive signals.

sale_order_tax_breakdown/models/sale.py
```diff
@@ -0,0 +1,50 @@
...
+                    res.setdefault(
+                        tax_name, {"amount": 0.0, "base": 0.0, "sequence": tax.sequence}
```
alexey-pelykh, line 30: Minor: the `sequence` stored here is `tax.sequence`, which is the individual tax's sequence. The original Odoo code sorts by `tax_group_id.sequence`. This is consistent with your design choice of breaking down by individual tax, but it is a subtle behavioral difference worth mentioning in the module description.

sale_order_tax_breakdown/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025 Juan Carlos Oñate - Tecnativa <juancarlos.onate@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sale Order Tax Breakdown",
+    "summary": "Show tax breakdown by tax group on sale orders and quotations.",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": ["sale"],
```
alexey-pelykh, line 10: Consider adding `"account"` to the `depends` list. The view XML uses `widget="tax-group-custom-field"` which is defined in the `account` module's JS assets. Currently `account` is only available transitively through `sale` -> `payment` -> `account`, which could break if Odoo refactors its dependency chain.

sale_order_tax_breakdown/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025 Juan Carlos Oñate - Tecnativa <juancarlos.onate@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sale Order Tax Breakdown",
+    "summary": "Show tax breakdown by tax group on sale orders and quotations.",
+    "version": "14.0.1.0.0",
```
alexey-pelykh, line 6: The version `14.0.1.0.0` is correct for the 14.0 branch. However, note that the 14.0 branch is EOL. If there is interest in a more recent version, this approach would need rework since `amount_by_group` was replaced in Odoo 15+ with `tax_totals_json`.

sale_order_tax_breakdown/tests/test_sale_order_tax_breakdown.py
```diff
@@ -0,0 +1,62 @@
...
+                "product_id": self.product.id,
+                "product_uom_qty": 1,
+                "price_unit": 100.0,
+                "tax_id": [(6, 0, [self.tax_10.id])],
+            }
+        )
+
+    def test_amount_by_group_splits_by_tax_name(self):
+        self.order._amount_by_group()
+        groups = {g[0]: g for g in self.order.amount_by_group}
```
alexey-pelykh, line 56: Good coverage of the basic two-tax scenario. A few additional test cases that would strengthen confidence:

1. **Same tax on multiple lines** -- verify the amounts are correctly aggregated (e.g., two lines both with 21% tax).
2. **Child taxes** -- the code handles `tax.children_tax_ids`, but there is no test for a parent tax with children (group-of-taxes scenario).
3. **Discount** -- verify the `price_reduce` calculation works correctly when `line.discount > 0`.
4. **Zero-tax order** -- verify `amount_by_group` is empty or handles gracefully when no taxes are applied.

PR #3922, opened 2025-10-02
[17.0][FIX] - Error quick search in sale order form view

DucTruongKomit:
**Current Behavior**
In SO form view, press Ctrtl + K, the error will appear

NOTE:
Make sure that base.substate does not have any record. If there is a record, the error will not appear.


cuongnmtm:
Odoo fixes the issue here https://github.com/odoo/odoo/commit/714cb8b19ef05aff7fdefaf623c859117ed4a35f

@DucTruongKomit  please close the PR.

PR #3406, opened 2024-11-12
[17.0][MIG] sale_triple_discount

dalonsod:
Standard migration.

Supersedes #3157 due to inactivity.

Pending tasks:

- [ ] Depends on https://github.com/OCA/account-invoicing/pull/1626
- [ ] Add https://github.com/OCA/sale-workflow/pull/2885 when merged in v16
- [ ] Squash adminstrative commits

PR #2114, opened 2022-07-06
[15.0][MIG] sale_order_line_input: Migration to 15.0

cesar-tecnativa:
@Tecnativa TT36521

PR #3929, opened 2025-10-06
[17.0][MIG] sale_triple_discount

hildickethan:
- Supersede of https://github.com/OCA/sale-workflow/pull/3406 to include the finished refactors in https://github.com/OCA/sale-workflow/pull/3907
- Depends on https://github.com/OCA/account-invoicing/pull/2114, ideally the 16.0 versions would get merged first

Notably this PR adds the `discount1` field compared to the previous migration PR which would require migrating data but it doesn't make sense to include here in code, so you would have to do that manually ([like this](https://github.com/hildickethan/sale-workflow/blob/268d6f37e554fb06263fdb6c291b6295db5ee607/sale_triple_discount/migrations/16.0.2.0.0/pre-migrate.py#L6))


sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,110 @@
...
+    _name = "sale.order.line"
+    _inherit = ["sale.order.line", "triple.discount.mixin"]
+
+    discount = fields.Float(
+        string="Total discount",
+        store=True,
+        compute="_compute_discount",
+        compute_sudo=True,
+        precompute=True,
+        readonly=True,
```
DantePereyra: I think we should avoid this and define it on view if possible as it's recommended [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0):
 - _If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead._

DantePereyra (changes requested):
Tested on local environment and works perfect! Only minor comment.

sale_triple_discount/models/sale_order_line.py
hildickethan: Seems reasonable, changed. Tests failing atm due to aws outage

DantePereyra (approved):

ferran-S73 (approved):

adrip-s73:
@hildickethan We have detected a problem with the discount calculation; when setting certain discounts and then changing the "discounting_type" field from "multiplicative" to "additive", the discount is not recalculated.
image

hildickethan:
> @hildickethan We have detected a problem with the discount calculation; when setting certain discounts and then changing the "discounting_type" field from "multiplicative" to "additive", the discount is not recalculated. 

I think the additive type is a bit abandoned right now since no one seems to actually use it anymore, talked about more in depth in https://github.com/OCA/sale-workflow/issues/3649. That issue proposes just hiding it for now which I think is done in the 18.0 PR so I will probably copy that if a simple change doesn't fix the issue

alexey-pelykh (commented):
## Review: `sale_triple_discount` migration to 17.0

Overall this is a solid migration. The module correctly inherits the `triple.discount.mixin`, adapts the discount compute logic for the pricelist-aware `_compute_discount` in Odoo 17, and includes good test coverage. A few items worth addressing:

### Bug: String formatting typo in `_get_final_discount`

**`sale_triple_discount/models/sale_order_line.py` line 84 (in the ValidationError)**

The format dict uses the key `"disc_type"` but the translatable string uses `%(dic_type)s`. This will raise a `KeyError` at runtime instead of the intended `ValidationError`:

```python
_("Sale order line %(name)s has unknown discounting type %(dic_type)s")
% {"name": self.name, "disc_type": self.discounting_type}
```

The placeholder is `%(dic_type)s` but the dict key is `"disc_type"`. One of them needs to change to match the other. Since the `.pot` and `.po` files already contain `%(dic_type)s`, the fix should be in the dict key:

```python
% {"name": self.name, "dic_type": self.discounting_type}
```

### Observation: `_additive_discount` returns 0 for negative discount sums

The method caps the result: `if discount <= 0: return 0`. This means if a user enters three negative discounts (surcharges) under additive mode, the net surcharge is silently clamped to zero. Multiplicative mode does not have this limitation. This may be intentional design but is worth documenting or reconsidering, since the USAGE.md explicitly shows negative discounts as a supported use case.

### Minor: `discounting_type` field shown with trailing `%` in form view

In `sale_order_view.xml`, the `discounting_type` selection field div includes a trailing ` %` text node (line 67 of the view XML), which is copy-paste from the discount field divs above. A selection field ("Additive" / "Multiplicative") should not display a percent sign.

### Positive observations

- The `_compute_discount1` approach using `env.protecting` to temporarily compute the pricelist discount and drag it into `discount1` is a clever adaptation for the 17.0 compute model.
- The `skip_triple_discount` context key prevents infinite recursion between `_compute_discount` and `_compute_discount1`.
- Test coverage is thorough: single discount on each field, multiplicative, additive, switching types, invoicing propagation, round-globally, and zero-discount edge case.
- Dependency on `account_invoice_triple_discount` (PR #2114) is already merged.
- All CI checks pass (pre-commit, tests with Odoo, tests with OCB, codecov).
- The `_prepare_invoice_line` correctly maps additive discounts to `discount1` only (since the individual discount fields lose meaning under additive mode).

### Summary

The string format key mismatch is a real bug that will surface at runtime if an invalid `discounting_type` is ever encountered. The trailing `%` on the selection field is cosmetic. The additive negative-discount clamping is a design question for the maintainers.

hildickethan:
> ## Review: `sale_triple_discount` migration to 17.0
thank you chatgpt, at least it did catch one bug that was pretty much impossible to see without messing with the db but a bug nonetheless

alexey-pelykh:
@hildickethan Thanks for addressing the review points and for catching the discount calculation bug. The migration looks solid.

v-garcies:
Hi, it is possible to merge this PR ?

rousseldenis (approved):

hildickethan:
> @rousseldenis your merge command was aborted due to failed check(s), which you can inspect on [this commit of 17.0-ocabot-merge-pr-3929-by-rousseldenis-bump-nobump](https://github.com/OCA/sale-workflow/commits/2d03e169c45fd179a39ba1e4c74a9edd840a9f3f).
> 
> After fixing the problem, you can re-issue a merge command. Please refrain from merging manually as it will most probably make the target branch red.

seems like something happened that made it start failing on the 29th/30th since the same test error is in the latest merged translation commit https://github.com/OCA/sale-workflow/actions/runs/28438717070/job/84538002045, I guess something external with Odoo or another repo?

rousseldenis:
@hildickethan I'm on it. sale_default_packaging module is failing

PR #3057, opened 2024-04-04
[16.0][IMP][sale_order_lot_selection] allow or not change of lot_id 

Kev-Roche:
A lot / serial number can be selected on a sale order line and is propagated to related stock moves when the order is confirmed (through procurements)
But this lot / serial number can be modified after confirmation without propagation to stock moves.

We think we should make the lot / serial number field read-only after confirmation OR propagate the modification to related stock moves (if not already done)

The current PR add a config parameter to choose between these two strategies.

The "propagation" strategy implies another PR for the stock_restrict_lot module : https://github.com/OCA/stock-logistics-workflow/pull/1560

Both @metaminux and me worked on this PR.

PR #3934, opened 2025-10-08
[16.0][FIX] sale_order_lot_selection: Fixed, that lot_id only gets propagat…

baguenth:
…ed to stock moves, if lot_id is actually filled

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -41,7 +41,7 @@ class SaleOrderLine(models.Model):
     def write(self, vals):
         res = super().write(vals)
         allow_to_change_lot = self.env.company.allow_to_change_lot_on_confirmed_so
-        if "lot_id" in vals and (
+        if vals.get("lot_id") and (
             allow_to_change_lot or self.order_id.state not in ["sale", "done"]
         ):
             self.move_ids.write({"restrict_lot_id": vals.get("lot_id")})
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -39,12 +39,41 @@ class SaleOrderLine(models.Model):
             )
 
     def write(self, vals):
+        """
+        Override of write to manage propagation of lot changes to stock moves.
+
+        Behavior
+        - If vals does not include lot_id, this method behaves exactly as the
+          standard write and returns.
+        - If lot_id is present, we compare the original lot per line with the
+          new lot after super().write(vals):
+          - If the lot did not actually change, the call is a no-op regarding
+            stock moves (nothing is propagated and no error is raised).
+          - If the order is in draft/quotation (not in sale/done), we propagate
+            the change to related stock moves by writing move.restrict_lot_id
+            to the new value (which can be a lot id or False).
+          - If the order is in sale/done:
+            - When the company setting allow_to_change_lot_on_confirmed_so is
+              True, we propagate the change to the related stock moves.
+            - When the company setting is False, we prevent changing the lot
+              and raise a UserError.
+        """
+        # Capture original lot ids to detect actual changes per record
+        original_lots = {rec.id: rec.lot_id.id for rec in self}
         res = super().write(vals)
+        if "lot_id" not in vals:
+            return res
         allow_to_change_lot = self.env.company.allow_to_change_lot_on_confirmed_so
-        if vals.get("lot_id") and (
-            allow_to_change_lot or self.order_id.state not in ["sale", "done"]
-        ):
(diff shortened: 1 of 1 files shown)


baguenth:
Hey @Kev-Roche,

I have discovered some issues with this module after this pull request https://github.com/OCA/sale-workflow/pull/3057 was integrated. 

My scenario is: 

- Sale order line doesn't have a lot id
- Related stock moves get a restrict_lot_id independently from the sale order line
- Sale order is changed and saved, lot_id was not changed but is in the vals for write function (it's still false)

Since the sale order line propagates the lot_id, no matter if it is filled or not, we try to set restrict_lot_id = False on stock moves which have a restrict_lot_id and are may already done. This is the moment, where the validation in stock_restrict_lot kicks in. 

https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_restrict_lot/models/stock_move.py#L147

I am not exactly sure if my fix is what we want and if it is complete. 

I think in general we should question:

- Should we propagate empty lot_id to move ids
- Should we propagate lot_id to move ids, regardless of the move ids state

@bodedra Maybe also interesting for you.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -41,7 +41,7 @@ def _compute_lot_id_readonly(self):
     def write(self, vals):
         res = super().write(vals)
         allow_to_change_lot = self.env.company.allow_to_change_lot_on_confirmed_so
-        if "lot_id" in vals and (
+        if vals.get("lot_id") in vals and (
```
rousseldenis: @baguenth Your fix is not correct.

In `vals.get("lot_id")` you'll get an... id. But you want to compare it with a `dict`.

rousseldenis (changes requested):

sale_order_lot_selection/models/sale_order_line.py
baguenth: @rousseldenis Sorry, my bad. 🙈

rousseldenis:
@baguenth Could you add a test to avoid regression ? Thanks

sale_order_lot_selection/models/sale_order_line.py
baguenth: Even though this will fix the issue, I am wondering if we need a more sophisticated check like:

```python
    def write(self, vals):
        # Capture original lot ids to detect actual changes per record
        original_lots = {rec.id: rec.lot_id.id for rec in self}
        res = super().write(vals)
        if "lot_id" not in vals:
            return res
        allow_to_change_lot = self.env.company.allow_to_change_lot_on_confirmed_so
        for line in self:
            old_lot = original_lots.get(line.id)
            new_lot = line.lot_id.id
            # Only act if there is an actual change
            if new_lot == old_lot:
                continue
            if allow_to_change_lot or line.order_id.state not in ["sale", "done"]:
                # Propagate the new lot restriction to related stock moves
                line.move_ids.write({"restrict_lot_id": new_lot})
            else:
                # Disallow changing the lot on confirmed SO if company setting forbids it
                raise UserError(_("You can't change the lot on confirmed sale order."))
        return res
``` 

With this we only do something, if the lot has actually changed for a sale order line.

baguenth:
@rousseldenis I have update the pr with a more general solution. 

We now check, if the lot has actually changed, before propagating it to the stock moves. This doesn't lead to a functional change and avoids unnecessary propagations of lots to related stock moves.

baguenth:
Hey @rousseldenis and @bodedra,

kindly asking for (another) review. 

Thanks :)

sebinthomasst (approved):

nggopakumar (approved):

sarangvbeo (approved):

QuatschSalat (approved):

njojic-ametras (approved):

baguenth:
Someone? :) @rousseldenis @bodedra @Kev-Roche

baguenth:
Hey @pedrobaeza, 

I saw you are also active on this module. Can I may ask you for a review / merge? My PR is stuck.

pedrobaeza:
@rousseldenis should confirm this due to his "Requested changes". You should also squash the commits into one.

baguenth:
@pedrobaeza I unfortunately cannot squash the commits, since the source branch is a production branch. I would need to recreate the pull request from a new source branch.

pedrobaeza:
Not really. You can rebase interactively and do the fixup, but if not able to, just open another PR.

baguenth:
FYI @rousseldenis @pedrobaeza

Closing the PR in favor of https://github.com/OCA/sale-workflow/pull/4029. 

There all commits have been squashed.

PR #3935, opened 2025-10-09
[18.0][IMP] Eficent -> ForgeFlow

MiquelRForgeFlow:
.

### DIFF
--- a/sale_commercial_partner/tests/test_sale_commercial_partner.py
+++ b/sale_commercial_partner/tests/test_sale_commercial_partner.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 Eficent Business and IT Consulting Services S.L.
+# Copyright (C) 2018 ForgeFlow, S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.tests import Form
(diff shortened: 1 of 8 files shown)


MiquelRForgeFlow:
@rousseldenis

rousseldenis:
@MiquelRForgeFlow Could you do a commit per module and add the module name in commit message ? Thanks

MiquelRForgeFlow:
I will do it when repo CI is green 🤷‍♂️

rvalyi (approved):

pedrobaeza:
As the change is a logical set, I don't think it should be split in one commit per module.

MiquelRForgeFlow:
> As the change is a logical set, I don't think it should be split in one commit per module.

Nice. Then nothing to do from my part. In fact, you could merge manually as the error in CI is unrelated.

pedrobaeza:
Not sure if the bot will work correctly on README regeneration merging directly.

rousseldenis:
> Not sure if the bot will work correctly on README regeneration merging directly.

@pedrobaeza Opened this : https://github.com/OCA/sale-workflow/issues/3936

pedrobaeza:
@rousseldenis yeah, I'm checking this.

PR #3938, opened 2025-10-10
[18.0][FIX] sale_automatic_workflow_job email sending

gurneyalex:
Issue: in `sale_automatic_workflow`, when the workflow was saying to send the email after the validation of the sale order, the email would not be sent.

Multiple causes involved:
* not preserving the context keys in the jobs to allow controlling the sending of the email
* not preserving the context keys in the jobs to allow creating a job for the sending of the email from the confirmation job
* the code in `sale_automatic_workflow` would create a job for sending the email just after creating the job for confirming the sale order, that job would run before the order was confirmed and not send the email (because of the state of the order)
* job identity would prevent the job confirming the order to create another job after the order was confirmed

Solution:

* make sure that the code in `super()._do_validate_sale_order()` will never try to send the email by setting the context key `send_order_confirmation` to `False`, and memorizing its original value in a different key
* forward the context keys that we need in the jobs
* check the value of the alternative key in the `_do_validate_sale_order` job to create the job that will send the email

### DIFF
--- a/sale_automatic_workflow_job/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow_job/models/automatic_workflow_job.py
@@ -1,13 +1,26 @@
 # Copyright 2020 Camptocamp (https://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
 
-from odoo import models
+from odoo import api, models
 
 from odoo.addons.queue_job.job import identity_exact
+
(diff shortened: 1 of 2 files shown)


simahawk (approved):

sebalix (approved):

sebalix:
@gurneyalex tests are failing, can you check? :)

gurneyalex:
@sebalix CI issue is fixed.

PR #3939, opened 2025-10-13
[18.0][MIG] ``pricelist_cache_rest``

SilvioC2C:


sebalix (approved):

pricelist_cache_rest/wizards/res_config_settings.xml
```diff
@@ -6,27 +6,23 @@
...
                 <h2>Pricelist Cache API keys</h2>
                 <div
                     class="row mt16 o_settings_container"
                     id="pricelist_cache_api_settings"
                 >
                     <div class="col-12 col-lg-6 o_setting_box">
-                        <div class="o_setting_left_pane" modifiers="{}" />
-                        <div class="o_setting_right_pane" modifiers="{}">
-                            <b modifiers="{}">Authorized Pricelist Cache API keys</b>
+                        <div class="o_setting_left_pane" />
```
sebalix, line 19: I have some doubts about this config panel, are you sure it's still declared that way? I'm thinking about `block` and `setting` tags

SilvioC2C: I checked an existing example from another OCA module, but I guess it can be improved if needed

pricelist_cache_rest/tests/test_pricelist_cache_rest.py
```diff
@@ -1,65 +1,31 @@
...
-        {"id": 17, "price": 45.0},
-        {"id": 18, "price": 99.0},
-        {"id": 19, "price": 120.0},
-        {"id": 20, "price": 67.0},
-    ],
-}
-
 
-class TestPricelistCache(TestPricelistCacheCommon):
+# Usa a date that's consistent w/ LIST_PRICES_MAPPING: it uses the prices  defined in
```
simahawk, line 18: Freezing on the class can give unexpected results. We should this date where needed w/ the ctx manager or method decorator.

pricelist_cache_rest/wizards/res_config_settings.xml
simahawk: have you tested in a browser?

simahawk (changes requested):

pricelist_cache_rest/wizards/res_config_settings.xml
SilvioC2C: On an empty DB:

Screenshot from 2025-10-14 17-18-53

pricelist_cache_rest/tests/test_pricelist_cache_rest.py
SilvioC2C: I just mimicked what I found in `pricelist_cache` tests:
https://github.com/OCA/sale-workflow/blob/18.0/pricelist_cache/tests/test_pricelist_cache.py#L9-L10

simahawk (approved):

Issue #3769, opened 2025-06-23
[17.0] sale_fixed_discount: PDF report omits printing fixed discount column and online preview omits displaying discounted unit price


kwoychesko:
I'm happy to work on this issue myself, but need some input/confirmation from the devs/maintainers.

## Module
sale_fixed_discount

## Describe the bug
When printing a PDF report of a quotation/order that includes a Fixed Discount, the fixed discount column is not included:

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/33623049-b4b0-4ae0-bc1b-9692d77654ff" />

Also, the online Preview does not show the discounted unit price. I have included this in the same issue because they both deal with the display of a fixed discount, but if you like I can create a separate issue. But perhaps this is by-design/a known limitation?

<img width="870" alt="Image" src="https://github.com/user-attachments/assets/a51af533-f3c8-4abc-8dd4-2c59c53dc4ea" />

## To Reproduce
**Affected versions**: 17.0

1. In a runboat.odoo-community.org instance for sale-workflow, login and enable Discounts in the Sales general settings.
2. With sale_fixed_discount installed, create a new quotation.
3. Add a product (e.g. Acoustic Bloc Screens with a unit price of $295).
4. In the line’s Disc. Fixed Amount column, enter a discount (e.g. $100).
5. Click Preview (the portal/online preview view). Notice that the regular unit price is red with strikethrough but the discounted unit price is not displayed (by design?)
6. Return to the backend view of the quotation/order and print it to create a PDF. Notice that the Disc. Fixed Amount column is not displayed.

**Expected behavior**

- Online preview: ideally should behave like the Disc. % functionality and show the discounted unit price beneath the strikethrough.
- PDF report: should include a “Disc. Fixed Amount” column showing the fixed discount (e.g. “$100.00”).

kwoychesko:
The second part of this issue (i.e. the online preview and the discounted price not being shown) is fixed by PR #3945 

After that PR is approved I will address the PDF printout (because I will know it's not "by design" as I questioned above).

Issue #3768, opened 2025-06-23
[17.0] sale_fixed_discount: Online Preview doesn't show discounted unit price


kwoychesko:
I'm happy to work on this issue myself (love this module! Thank you for your work including the big refactor!) But I just wanted to be sure nobody was working on it already.

## Module
sale_fixed_discount

## Describe the bug
If sale_fixed_discount is not installed and a % discount is applied to a product the online Preview (e.g. using the Preview button on the sale order) shows the regular unit price in red with strikethrough and the new discounted price shown below in black. If sale_fixed_discount is installed, the "new discounted price" is just the old regular price.

<img width="872" alt="Image" src="https://github.com/user-attachments/assets/227f0b8f-eb8b-4a96-b4b8-84e5ee220b34" />

## To Reproduce
**Affected versions**: 17.0

1. In a runboat.odoo-community.org instance for sale-workflow, login and enable Discounts in the Sales general settings.
2. With sale_fixed_discount installed, create a new quotation.
3. Add a product (e.g. Acoustic Bloc Screens with a unit price of $295).
4. In the line’s Disc.% column, enter 50.
5. Click **Preview** (the portal/online preview view).
**Expected behavior**
The strikethrough price should be $295.00 and the replacement price should be the discounted unit price, $147.50—just as in the preview when sale_fixed_discount is not installed. Instead, the non-discounted unit price is shown.

kwoychesko:
This issue is addressed by PR #3945

PR #3945, opened 2025-10-15
[17.0][FIX] `sale_fixed_discount`: fix portal template

SilvioC2C:
### SO lines info
From the SO form view's `order_line` tree view (using quantity as sequence to ease comparison w/ the template reports below):
Screenshot from 2025-10-15 09-43-08

NB: the fact that the "Disc. %" on the second line is 0 even though the "Discount (Fixed)" is 0.10 is not an error: since field `discount` is computed but not readonly, there are cases in which field `discount` could be assigned a value which is different than the one assigned by the `compute` method (for example, when importing the SO through the Odoo file importer).

### Portal report before the fix
Screenshot from 2025-10-15 09-43-17

:red_circle: line 1: the discounted unit price is 1.00 instead of 0.90 (10% applied on 1.00)
:red_circle: line 2: the discounted unit price is not shown at all
:green_circle: line 3: the discounted unit price is shown correctly (3.00 w/ 0.30 or 10% discount = 2.70)
:green_circle: line 4: no discount applied

### Portal report after the fix
Screenshot from 2025-10-15 09-43-24

:green_circle: line 1: the discounted unit price is shown correctly (1.00 w/ 10% discount = 0.90)
:green_circle: line 2: the discounted unit price is shown correctly (2.00 w/ 0.10 discount = 1.90)
:green_circle: line 3: the discounted unit price is shown correctly (3.00 w/ 0.30 or 10% discount = 2.70)
:green_circle: line 4: no discount applied

### DIFF
--- a/sale_fixed_discount/views/sale_portal_templates.xml
+++ b/sale_fixed_discount/views/sale_portal_templates.xml
@@ -35,8 +35,13 @@
                 name="t-att-class"
             >((line.discount or line.discount_fixed) and 'text-danger' or '') + ' text-right'</attribute>
         </xpath>
-        <xpath expr="//div[@t-if='line.discount']/t" position="attributes">
-            <attribute name="t-out">line.price_unit - line.discount_fixed</attribute>
+        <xpath expr="//div[@t-if='line.discount']" position="after">
+            <div t-elif="line.discount_fixed">
+                <t
+                    t-out="line.price_unit - line.discount_fixed"
+                    t-options='{"widget": "float", "decimal_precision": "Product Price"}'
+                />
+            </div>
         </xpath>
         <xpath
             expr="//section[@id='details']//table/tbody//t[@t-foreach='lines_to_report']/tr//t[@t-if='not line.display_type']//td[@t-if='display_discount']"


kwoychesko:
Great job @rousseldenis and thanks for this PR! I've tested it and can confirm it fixes the following:

- Completely solves the issue I reported in #3768. 
- Solves the "online" portion of the issue I reported in #3769 (the printed PDF error lies in a different xml file).

kwoychesko:
@rousseldenis Hiya! Okay to get this merged now? I'd like to continue with fixing the PDF printout issue that I raised in Issue #3769  Thanks!

alexey-pelykh (approved):
Clean and well-targeted fix for the portal template discount rendering.

**What was broken:** The old xpath `//div[@t-if='line.discount']/t` replaced the base percentage discount formula inside the `line.discount` div with `line.price_unit - line.discount_fixed`. This caused two defects:
1. Lines with percentage discount (`line.discount > 0`): the formula was overwritten, losing the percentage-based calculation.
2. Lines with only fixed discount (`discount_fixed > 0`, `discount == 0`): the parent `t-if="line.discount"` evaluated to false, so the discounted price was never rendered.

**How the fix works:** Instead of overriding the base discount div's inner element, the fix adds a sibling `<div t-elif="line.discount_fixed">` after the base `<div t-if="line.discount">`. This correctly:
- Leaves the Odoo base percentage-discount path intact.
- Creates a separate rendering path for fixed-discount-only lines.
- Uses proper `t-options` with `decimal_precision: "Product Price"`, consistent with the field definition.

The `t-elif` structure is appropriate since `_check_discounts` enforces consistency when both fields are set, and when both are present `line.discount > 0` will be true (handled by the first branch).

CI is green (pre-commit, tests with Odoo and OCB, codecov all passing). The screenshots in the PR description clearly demonstrate the before/after behavior.

LGTM.

rousseldenis (approved):

PR #3946, opened 2025-10-15
[19.0][MIG] sale_order_line_date Migrate to 19.0

Tatider:
* Added support of combo products added (maybe should be backported to 18?)
* An update of tests was required because in Odoo 19.0, the database by default doesn't have Demo Data.


eslint.config.cjs
```diff
@@ -22,7 +22,7 @@ const config = [{
         },
 
         ecmaVersion: 2024,
-        sourceType: "script",
+        sourceType: "module",
```
rousseldenis: You should not include this in migration commit.

Moreover this should be first done at copier template level. So, you should do a PR there : https://github.com/OCA/oca-addons-repo-template

Tatider: 1. It would be a bit challenging for me. I have removed the change from this MR, but I have no experience making changes to the repository you have mentioned so I have no idea if I am doing it in a right way... But here it is https://github.com/OCA/oca-addons-repo-template/pull/325

2. I don't understand how a change in the template will affect this (existing) repository with an existing branch for 19.0.  I am concerned that now my MR is failing, and I don't understand the process of making it valid or how could I affect things moving forward for it...

sale_order_line_date/static/src/js/sale_order_line_field/sale_order_line_field.esm.js
```diff
@@ -0,0 +1,12 @@
...
+import {SaleOrderLineListRenderer} from "@sale/js/sale_order_line_field/sale_order_line_field";
+import {patch} from "@web/core/utils/patch";
+
+patch(SaleOrderLineListRenderer.prototype, {
+    get comboColumns() {
+        const columns = super.comboColumns;
+        columns.push("commitment_date");
+        return columns;
+    },
+});
```
yankinmax, line 12: As @pedrobaeza suggested here:
https://github.com/OCA/oca-addons-repo-template/pull/325#issuecomment-3451645003
You can try to rename `sale_order_line_field.js` to `sale_order_line_field.esm.js`

sale_order_line_date/README.rst
```diff
@@ -85,13 +86,17 @@ Contributors
 - Moaad Bourhim <moaad.bourhim@gmail.com>
 - Bernat Puig <bernat.puig@forgeflow.com>
 - Nhan Tran <nhant@trobz.com>
+- Tatiana Deribina <tatiana.deribina@sprintit.fi>
 
 Other credits
 -------------
 
 The migration of this module from 17.0 to 18.0 was financially supported
 by Camptocamp
```
yankinmax: You can remove the credits from previous migrations.

> Remove references in CREDITS.rst of past financed migrations by other companies.

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

yankinmax (approved):
Technical migration, thanks

ivantodorovich (approved):

sale_order_line_date/static/src/js/sale_order_line_field/sale_order_line_field.esm.js
Tatider: Thank you a lot, I didn't understand what he meant by his comment. It makes sense now. Updated.

PR #3947, opened 2025-10-15
[19.0][MIG] sale_procurement_group_by_line: Migrate to 19.0

Tatider:
- Major changes were done to the module because "procurement.group" model was converted to "stock.reference" by Odoo SA (more details [here](https://github.com/odoo/odoo/commit/2713876dbc70d3984e584a9037a2206dcda4e84a#diff-624e618ed121333ed277c864cb002690b71ac2fc05e1a0e94305c70d4f6e21ce)).
    - Removed Update of "partner_id" and "move_type" - no fields like these exist in the "stock.reference"
    - Added overload of `_prepare_procurement_values` because "group_id" is no longer an arguments
    - Removed `origin` calculation  - no longer an argument of `_create_procurements` function
    - Removed check of `stock.move:name` - move no longer has name/description field

P.S. An update of tests was required because in Odoo 19.0, the database by default doesn't have Demo Data


sale_procurement_group_by_line/__manifest__.py
```diff
@@ -4,13 +4,18 @@
...
-    "version": "18.0.1.0.1",
-    "author": "Camptocamp,"
-    "ForgeFlow,"
-    "Serpent Consulting Services Pvt. Ltd.,"
-    "Odoo Community Association (OCA)",
+    "name": "Sale Stock Reference by Line",
+    "summary": (
+        "Base module for definition of stock references creation rules for Sale order"
+    ),
+    "version": "19.0.1.0.1",
```
ivantodorovich:
```suggestion
    "version": "19.0.1.0.0",
```

sale_procurement_group_by_line/migrations/19.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,6 @@
+def migrate(cr, version):
+    """
+    TODO recalculate `stock_reference_id` once Odoo migration scripts are ready
```
ivantodorovich: Is this still pending, or already taken care of by the `pre-migrate` script?

JordiMForgeFlow (commented):
Overall looks good, just some comments/doubts:

- In the `_action_launch_stock_rule` when you call super() at the end I think you are missing to pass all *kwargs
- I understand the renaming of the module README name, but I wonder if we should also rename all the module name. Seems weird to have a discrepancy between the module name and what is shown in the README.
- I don't think we should put migration scripts in the module, it should be already taken care by the migration service (either Odoo enterprise or Openupgrade scripts).

sale_procurement_group_by_line/migrations/19.0.1.0.0/post-migrate.py
StefanRijnhart: In any case, from what I can tell from a 19.0 migrated database is that they just renamed the table, so no additional migration should be required on that part.

Tatider:
> Overall looks good, just some comments/doubts:
> 
> * In the `_action_launch_stock_rule` when you call super() at the end I think you are missing to pass all *kwargs
> * I understand the renaming of the module README name, but I wonder if we should also rename all the module name. Seems weird to have a discrepancy between the module name and what is shown in the README.
> * I don't think we should put migration scripts in the module, it should be already taken care by the migration service (either Odoo enterprise or Openupgrade scripts).

I have addressed everything except the module technical name change. In my experience, changing of the module technical name usually causes a lot of extra work during migration. This module doesn't have a own maintainer, so it is probably repository key maintainer should decide on it...

yankinmax:
@Tatider @JordiMForgeFlow 
I agree to get @OCA/community-maintainers opinion on that topic, but it's not actually a difficult task to write a pre-script and update existing models/fields/references in database tables using the power of openupgradelib:
https://github.com/OCA/openupgradelib/blob/189837ed289892720e5040551c275f1330c470d2/openupgradelib/openupgrade.py#L188-L191
It can be safe enough for the migration methods to be called if in pre-script we check for the old name module to be installed.

sale_procurement_group_by_line/model/sale.py
```diff
@@ -10,30 +10,35 @@
...
         procurement groups
 
         """
         return 8, self.order_id.id
 
-    def _action_launch_stock_rule(self, previous_product_uom_qty=False):
+    def _prepare_procurement_values(self):
+        """
+        Overload core method to modify stock references to the line specific
+        """
```
yankinmax:
```
# Add stock reference to procurement values
```
If you add a comment, you don't redefine the original docstring. This is considered a good practice.

sale_stock_reference_by_line/tests/test_sale_stock_reference_by_line.py
```diff
@@ -71,55 +71,60 @@ def _create_sale_order(cls):
...
-        self.assertEqual(self.line1.move_ids.name, self.line1.product_id.display_name)
         self.assertEqual(len(self.line2.move_ids), 1)
-        self.assertEqual(self.line2.move_ids.name, self.line2.product_id.display_name)
 
     def test_03_action_launch_procurement_rule_2(self):
-        group_id = self.proc_group_model.create(
-            {"move_type": "one", "sale_id": self.sale.id, "name": self.sale.name}
+        stock_ref = self.proc_group_model.create(
+            {
+                "sale_ids": [(6, 0, self.sale.ids)],
```
yankinmax, line 107:
```
from odoo.fields import Command
...
"sale_ids": [Command.set(self.sale.ids)],
```

yankinmax (approved):
Technical review, thanks

StefanRijnhart:
Agree to rename the module. The name is still acceptable now, but in Odoo 20 the young folks are not going to know what a procurement group was and by then a rename will be unavoidable. Might as well do it now.

Tatider:
@StefanRijnhart @JordiMForgeFlow 
Should the new module name be "sale_stock_reference_by_line"?

StefanRijnhart:
Sounds reasonable!

sale_stock_reference_by_line/tests/__init__.py
```diff
@@ -1,3 +1,3 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from . import test_sale_procurement_group_by_line
+from . import test_sale_stock_reference_by_line
```
StefanRijnhart, line 3: Please also rename the file to fix the tests.

sale_stock_reference_by_line/tests/test_sale_procurement_group_by_line.py
```diff
@@ -4,48 +4,48 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo.tests import Form
-from odoo.tests.common import TransactionCase
 
+from odoo.addons.sale_stock.tests.common import TestSaleStockCommon
 
-class TestSaleProcurementGroupByLine(TransactionCase):
+
+class TestSaleProcurementGroupByLine(TestSaleStockCommon):
```
StefanRijnhart: Please rename test class.

sale_stock_reference_by_line/README.rst
```diff
@@ -21,55 +21,39 @@ Sale Order General Discount
...
     :target: https://runboat.odoo-community.org/builds?repo=OCA/sale-workflow&target_branch=19.0
     :alt: Try me on Runboat
 
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
-This module allows to set a general discount in a sales order. This
-general discount is set to each line order in the standard discount
-field.
+This module was written to extend the functionality of stock references
+(aka procurement groups) created from a sale order.
```
StefanRijnhart:
```suggestion
(previously: procurement groups) created from a sale order.
```

sale_stock_reference_by_line/README.rst
```diff
@@ -21,55 +21,39 @@ Sale Order General Discount
...
 
-You can configure:
+On itself, this module does nothing it is a requirement for modules
+which needs to create stock references (aka procurement groups) per sale
+order line basis.
```
StefanRijnhart:
```suggestion
On itself, this module does nothing. It is a requirement for modules that
need to create a stock reference for each indivual sale order line.
```

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
@JordiMForgeFlow @ivantodorovich How is it looking to you at this point?

JordiMForgeFlow:
LGTM, but in other cases I had to rename a module, the best practice is to rename the entire commit history. Here is how to do it:

`git filter-branch --tree-filter 'if [ -d sale_procurement_group_by_line ]; then mv sale_procurement_group_by_line sale_stock_reference_by_line; fi' HEAD
git rebase origin/18.0
git push -f`

StefanRijnhart:
@JordiMForgeFlow That is certainly an option, but it can also be said that the rename is a legitimate part of the module history that takes place now, and is reflected in the module's commit history. Also, the rename of the module directory is only part of the work, and there are field names, file names and external references to consider that cannot easily be rewritten in each commit and will be done in a specific commit. You could consider commits with a rewritten directory name misleading, as they make the history of a module less clear.

@pedrobaeza What is considered the best practice when renaming modules? Is it to rewrite the name throughout the git history or do it in a separate renaming commit?

JordiMForgeFlow:
@StefanRijnhart actually it was @pedrobaeza who explained the procedure to me 😆

pedrobaeza:
When renaming, if the whole commit history is not renamed, then on next version, the rescue procedure will only catch from the migration commit, losing part of the commit history. Thus, the whole renaming is needed IMO.

pedrobaeza:
Anyway, you will need an extra commit with the rest of the work of the renaming (translations, complete XML-IDs where needed, etc).

StefanRijnhart:
OK, thanks for clarifying! @Tatider can you apply the git history rewrite?

Tatider:
> Anyway, you will need an extra commit with the rest of the work of the renaming (translations, complete XML-IDs where needed, etc).
I have already updated the translations files in main migration commit, no XML ID updated needed. Would it be a problem?

Tatider:
> OK, thanks for clarifying! @Tatider can you apply the git history rewrite?

I am trying, but it is not exactly working with just a copy-paste command...
Screenshot 2025-12-19 at 12 47 55

@JordiMForgeFlow are you also sure it should be on top of 18.0 and not 19.0?

JordiMForgeFlow:
@Tatider yes it should be on top of 19.0, I just copied the template command but you should adapt it to what you need.

Tatider:
@StefanRijnhart @JordiMForgeFlow the history is now rewritten.

StefanRijnhart (approved):
Thanks for all the work! Looking real good.

JordiMForgeFlow (approved):

dreispt (approved):

StefanRijnhart:
Rename added to openupgrade 19 apriori in https://github.com/hbrunn/OpenUpgrade/pull/20

PR #3951, opened 2025-10-17
ADD sale_project_milestone

eLBati:


alexey-pelykh (commented):
## Review: `sale_project_milestone` (new module)

Nice module concept — linking milestones to sale order lines with flexible project assignment is a useful extension. CI is fully green and test coverage is solid (884 lines, 20+ test cases). A few items to address before merge:

---

### 1. `_generate_milestone` signature change (medium)

**File:** `models/sale_order_line.py`, lines ~100-130

The base Odoo `_generate_milestone()` method on `sale.order.line` takes **no arguments** (only `self`). This override changes the signature to `_generate_milestone(self, project=None)`, which is fine for direct callers but breaks any third-party module or Odoo base code that calls `_generate_milestone()` without the `project` keyword — the base method's callers will still work since `project` defaults to `None`, but other modules overriding `_generate_milestone` may not expect the extra parameter if they call `super()`.

More importantly, this override **unconditionally creates a milestone** when a project is passed (skipping the `super()` call), which bypasses the base logic that checks `service_policy == 'delivered_milestones'` before creating milestones. This means milestone records are created even when the service policy is `ordered_prepaid` or `delivered_manual`. Consider whether this is intentional — if so, document it; if not, add the policy check.

### 2. `_compute_is_product_milestone` override widens filter (low)

**File:** `models/sale_order.py`, lines ~12-18

The base method filters products by `service_policy == 'delivered_milestones'`. The override adds `or p.service_tracking == 'milestone_project'`. This means the milestone smart button on the sale order will now appear for products that use `milestone_project` tracking even when they have `ordered_prepaid` or `delivered_manual` policy. This may be intentional (the button is useful for navigation), but the docstring says "show milestone button also for milestone_project tracking" — good. Just be aware that this changes visible UI for products that would otherwise not show the button.

Also, the override doesn't call `super()`. If another module extends `_compute_is_product_milestone`, that extension will be lost. Consider calling `super()` first, then adding the extra condition.

### 3. `_register_analytic_lines_for_existing_invoices` — direct analytic line creation on posted moves (medium)

**File:** `models/sale_order_line.py`, lines ~190-250

For posted invoices, this method directly creates `account.analytic.line` records and modifies `analytic_distribution` on posted `account.move.line` records. In Odoo 16, modifying posted journal entries normally requires either canceling/reposting or using specific reconciliation mechanisms. Directly writing `analytic_distribution` on a posted move line bypasses the accounting integrity checks (`_check_balanced`, journal lock dates, etc.).

The analytic line creation itself also doesn't go through the standard Odoo analytic line generation pipeline (which normally happens via `_create_analytic_lines()` on `account.move.line`). This could lead to:
- Duplicate analytic lines if Odoo later reconciles/reposts the move
- Missing fields that the standard pipeline would populate

Consider using `move.button_draft()` / re-post pattern, or at minimum document this as a known trade-off.

### 4. `existing_milestone_id` domain doesn't filter by company (low)

**File:** `models/sale_order_line.py`, line ~30

The `existing_project_id` field has `domain="[('company_id', '=', company_id)]"` which is good. However, `existing_milestone_id` only filters by `project_id` and `sale_line_id == False`. In multi-company environments, the milestone domain implicitly inherits company filtering from the project filter, but only if a project is selected first. If somehow the domain is evaluated without `existing_project_id`, it could show milestones from all companies.

### 5. `hr_timesheet` in dependencies — is it needed? (low)

**File:** `__manifest__.py`, line ~16

The module depends on `sale_project` and `project`, but also on `hr_timesheet`. Reviewing the code, the module doesn't use any timesheet-specific models or fields. The dependency likely exists because `_timesheet_service_generation` is defined in `sale_timesheet` (which depends on `hr_timesheet`). However, since `sale_project` already depends on `sale_timesheet` in Odoo 16, the explicit `hr_timesheet` dependency may be redundant. Verify whether it can be removed to keep the dependency tree minimal.

### 6. View: `attrs` dict uses legacy format (low)

**File:** `views/sale_order_views.xml`, `views/product_template_views.xml`

The `attrs` dictionaries use the legacy Odoo format (string-encoded Python dicts). This is standard for Odoo 16 but worth noting for forward-migration to 17+ where the new `invisible`/`readonly` attribute syntax is preferred. Not blocking for 16.0.

### 7. Test hygiene: mutating shared class data (low)

**File:** `tests/test_sale_project_milestone.py`, multiple tests

Several tests (05, 06, 07, 08, 09) mutate `cls.product_no_tracking.service_tracking` during the test and reset it at the end. Since `TransactionCase` rolls back after each test, the reset is technically unnecessary and the mutation is safe. However, if any test fails before the reset line, subsequent tests in a `SavepointCase` subclass would see corrupted state. Since these are `TransactionCase`, this is fine — but adding a comment explaining the pattern would help future maintainers.

### 8. Missing `test_12` and `test_20` in test numbering (trivial)

Test numbers skip from `test_11` to `test_13` and from `test_19` to `test_21`. Not functional, but suggests removed tests. No action needed.

---

### Summary

| # | Finding | Severity |
|---|---------|----------|
| 1 | `_generate_milestone` signature change + unconditional creation | Medium |
| 2 | `_compute_is_product_milestone` doesn't call super | Medium |
| 3 | Direct modification of posted move analytic distribution | Medium |
| 4 | Milestone domain missing company filter fallback | Low |
| 5 | Possibly redundant `hr_timesheet` dependency | Low |
| 6 | Legacy `attrs` format (acceptable for 16.0) | Info |
| 7 | Mutating shared test fixtures | Low |
| 8 | Test numbering gaps | Info |

Overall: well-structured module with good test coverage and clear documentation. The medium items (1-3) would benefit from clarification or adjustment before merge.

sale_project_milestone/models/product_template.py
```diff
@@ -0,0 +1,74 @@
...
+        """Extend constraints for milestone_project tracking."""
+        res = super()._check_project_and_template()
+        for product in self:
+            if product.service_tracking == "milestone_project" and product.project_id:
+                raise ValidationError(
+                    _(
+                        "The product %(product)s should not have a global project "
+                        "since it will generate a project."
+                    )
+                    % {"product": product.name}
```
rousseldenis, line 54: Use translation function arguments instead.

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @api.depends("product_id.service_tracking")
+    def _compute_show_project_milestone_field(self):
+        for line in self:
+            line.show_project_milestone_field = (
+                line.product_id.service_tracking == "milestone_project"
+            )
+
+    @api.onchange("existing_project_id")
+    def _onchange_existing_project_id(self):
```
rousseldenis, line 37: Please avoid onchanges for new fields. Use `compute/store=True/readonly=False` fields instead.

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+                "project_id": project.id,
+                "sale_line_id": self.id,
+                "quantity_percentage": 1.0,
+            }
+        )
+
+        # Post message on sale order
+        msg_body = _(
+            "Milestone Created (%(product)s) in project %(project)s: %(milestone)s"
+        ) % {
```
rousseldenis, line 114: Please use translation function arguments instead

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+        self.ensure_one()
+        if not project:
+            project = self.project_id or self.order_id.project_id
+
+        if not project:
+            return super()._generate_milestone()
+
+        # Always create milestone when service_tracking is milestone_project
+        milestone = self.env["project.milestone"].create(
+            {
```
rousseldenis, line 103: Maybe a method to prepare data ?

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+        if not self.project_id:
+            self.project_id = self.existing_milestone_id.project_id
+
+        # Register analytic lines for existing invoices if any
+        self._register_analytic_lines_for_existing_invoices()
+
+        # Post message on sale order chatter
+        msg_body = _(
+            "Line (%(product)s) linked to project %(project)s: milestone %(milestone)s"
+        ) % {
```
rousseldenis, line 173: Same as above

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+                values["analytic_distribution"] = {
+                    self.project_id.analytic_account_id.id: 100
+                }
+
+        return values
+
+    def action_link_existing_milestone(self):
+        """Action to link an existing milestone to confirmed sale order lines"""
+        self.ensure_one()
+        if not self.existing_milestone_id:
```
rousseldenis, line 143: Maybe a `_check_...` method in order to group the exceptions ?

sale_project_milestone/models/sale_order_line.py
```diff
@@ -0,0 +1,243 @@
...
+        for inv_line in posted_invoice_lines:
+            # Update analytic_distribution on the invoice line
+            if (
+                not inv_line.analytic_distribution
+                or analytic_account.id not in inv_line.analytic_distribution
+            ):
+                inv_line.analytic_distribution = {analytic_account.id: 100}
+
+            # Check if analytic lines already exist for this account
+            existing_lines = self.env["account.analytic.line"].search(
```
rousseldenis, line 220: Maybe a search before the loop ? Moreover, I would have used `read_group` to do it

rousseldenis (changes requested):

rousseldenis:
@eLBati

PR #3952, opened 2025-10-17
[18.0][FIX] sale_exception: handle exception rollback in action_confirm method

lef-adhoc:
It's the same problem as #3672  https://github.com/OCA/sale-workflow/pull/3767 https://github.com/OCA/sale-workflow/pull/3885

### DIFF
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -49,6 +49,11 @@ class SaleOrder(models.Model):
 
     def action_confirm(self):
         if self.detect_exceptions():
+            exception_map = {sale.id: sale.exception_ids.ids for sale in self}
+            self.env.cr.rollback()
+            for sale in self:
+                exception_ids = exception_map.get(sale.id, [])
+                sale.write({"exception_ids": [(6, 0, exception_ids)]})
             if not self.env.company.sale_exception_show_popup:
                 return
             return self._popup_exceptions()


sale_exception/models/sale_order.py
```diff
@@ -49,6 +49,11 @@ def sale_check_exception(self):
 
     def action_confirm(self):
         if self.detect_exceptions():
+            exception_map = {sale.id: sale.exception_ids.ids for sale in self}
+            self.env.cr.rollback()
```
alexey-pelykh, line 53: `self.env.cr.rollback()` rolls back the **entire** current transaction -- not just the side effects from `detect_exceptions()`. In a web request, this undoes everything since the RPC call began. If `action_confirm` was called after other writes in the same request (e.g., from a wizard flow), those writes are also lost.

The idiomatic Odoo approach is to use a **savepoint**:
```python
with self.env.cr.savepoint():
    ...
```
A savepoint creates a nested transaction boundary. Raising inside it rolls back only to the savepoint, preserving the outer transaction.

sale_exception/models/sale_order.py
```diff
@@ -49,6 +49,11 @@ def sale_check_exception(self):
...
+            for sale in self:
+                exception_ids = exception_map.get(sale.id, [])
```
alexey-pelykh, line 55: After `cr.rollback()`, the ORM cache is stale -- it reflects data that no longer exists in the database. Calling `sale.write()` here operates on corrupted cache state.

At minimum, `self.env.invalidate_all()` must be called after rollback. But even with cache invalidation, the recordset `self` may reference records that were rolled back (if created in the same transaction).

This is why savepoints are preferred -- Odoo's savepoint context manager handles cache invalidation automatically when the savepoint is rolled back.

alexey-pelykh (changes requested):
## Review: `sale_exception` rollback handling in `action_confirm`

Thanks for working on this issue. The underlying problem is real and well-documented across #3672, #3767, and #3885: side effects from other modules (e.g. `sale_loyalty`) executing during `super().action_confirm()` persist even when `sale_exception` detects exceptions and aborts the confirmation.

However, the approach used here -- calling `self.env.cr.rollback()` directly -- raises several concerns that I believe need to be addressed before this can be merged.

### Critical Issues

**1. `cr.rollback()` rolls back the entire transaction, not just `action_confirm` side effects.**

This rolls back everything since the last commit/savepoint in the current transaction, which may include writes performed by the caller, by UI form saves, or by other business logic that triggered `action_confirm`. In a web request context, Odoo wraps the entire RPC call in a transaction, so this rollback could undo more than intended.

The idiomatic Odoo pattern for partial rollback is **savepoints**:
```python
with self.env.cr.savepoint():
    # operations here are rolled back if an exception is raised
```

**2. ORM cache is not invalidated after rollback.**

After `self.env.cr.rollback()`, the ORM cache (registry cache, record cache) is stale -- it reflects the pre-rollback database state that no longer exists. The subsequent `sale.write({"exception_ids": [(6, 0, exception_ids)]})` operates on a potentially corrupted cache. At minimum, `self.env.invalidate_all()` must be called after rollback to avoid silent data corruption or `MissingError` exceptions.

**3. `self` recordset references may be invalid after rollback.**

If any of the sale orders in `self` were created in the same transaction (e.g., from a wizard that creates and immediately confirms), `cr.rollback()` would undo the creation itself. The subsequent iteration over `self` and `sale.write()` would fail or produce undefined behavior.

**4. No test coverage for the rollback behavior.**

The PR does not add or modify tests. PR #3885 (which addresses the same issue) includes a dedicated `test_sale_exception_confirm` module with a test that verifies side effects from other `action_confirm` overrides are properly rolled back. This kind of behavioral change absolutely needs test coverage.

### Suggestion

Consider using a **savepoint** pattern instead of raw `cr.rollback()`:

```python
def action_confirm(self):
    try:
        with self.env.cr.savepoint():
            exceptions = self.detect_exceptions()
            if exceptions:
                raise UserError("Exceptions detected")  # triggers savepoint rollback
    except UserError:
        pass
    # After savepoint rollback, re-detect and write exceptions cleanly
    self.detect_exceptions()
    if not self.env.company.sale_exception_show_popup:
        return
```

Alternatively, consider the `_register_hook` approach from PR #3767, which avoids the rollback problem entirely by intercepting before any `super()` chain executes.

### CI Status

The module-specific tests pass (pre-commit, codecov, test with Odoo/OCB for the changed module). The failing jobs appear to be full-suite runs with failures likely unrelated to this change (logs are expired/unavailable for verification). `runboat/build` is still pending.

### Verdict

The problem being solved is valid, but the `cr.rollback()` approach is unsafe for production use. Requesting changes.

simahawk:
Fix for v18 is getting merged https://github.com/OCA/server-tools/pull/3590
Companion fix for sale https://github.com/OCA/sale-workflow/pull/4255

PR #3953, opened 2025-10-17
[18.0][ADD] sale_confirm_group: Add module

SilvioC2C:
This module allows configuring a list of groups per-company who are granted permission to confirm sale orders:

1. button "Confirm" in sale views is always hidden for users not in those groups
2. if users outside those groups try to confirm a SO, an error is raised


sale_confirm_security_group/models/sale_order.py
```diff
@@ -0,0 +1,74 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def _get_groups_can_confirm_xmlids(self) -> list[str]:
+        """Returns a list of groups that are allowed to confirm a SO
+
+        Hook method, can be overridden by inheriting modules
+        """
+        return ["sale_confirm_security_group.group_can_confirm_sale"]
```
rousseldenis: And why not having  a x2many on `res.company` that can allow more flexible configurations instead ?

SilvioC2C: Oh, you mean a `Many2many` field to assign specific groups to each company? Yeah, that could work, I'll check if it's easy to implement and add tests for that, thanks for the suggestion!

henrybackman (approved):

SilvioC2C:
Tests are failing because of a 503 error, possibly due to AWS outage

sale_confirm_groups/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    user_can_confirm = fields.Boolean(
+        compute="_compute_user_can_confirm", compute_sudo=True
+    )
+
+    @api.depends("company_id.can_confirm_sales_groups_ids.users")
```
simahawk: this field is not stored, I would remove the depends to avoid unnecessary computations

sale_confirm_groups/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _skip_check_user_can_confirm(self) -> bool:
+        """Checks whether the user permission to confirm should be checked
+
+        If "skip_check_user_can_confirm" is found in the context, its value is returned.
+        Returns True by default when in test mode, else False.
+        """
+        if "skip_check_user_can_confirm" in self.env.context:
+            return bool(self.env.context["skip_check_user_can_confirm"])
+        return bool(
```
simahawk: why do we need this?

sale_confirm_groups/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+        if self._skip_check_user_can_confirm():
+            return arch, view
+
+        # We want to hide all the ``action_confirm()`` buttons related to ``sale.order``
+        # records, so we find all of them, traverse their parent nodes top-down until
+        # we reach the node's direct parent, and if we encounter fields along the way
+        # (which can only be ``X2Many`` fields) we update the node's model, to make sure
+        # we hide *only* the ``action_confirm()`` buttons related to ``sale.order``
+        # records, not buttons with the same name for other models
+        for node in arch.xpath("//button[@name='action_confirm']"):
```
simahawk: I don't understand this. This an override on the model `sale.order` why should you have any btn w/ the same action name on other models? 🤔

simahawk (changes requested):

simahawk:
The name should be `sale_confirm_group`

sale_confirm_groups/models/sale_order.py
simahawk: as we said: we need a flag to enable the feature (off by default) then you can get rid of the test check

SilvioC2C:
CI is red due to `odoo_test_helper.FakeModelLoader._check_wrong_import()` that thinks models were already loaded, if someone can give an hint I'd be happy to check it

sale_confirm_group/tests/test_sale_confirm_group_views.py
```diff
@@ -0,0 +1,71 @@
...
+        # Update context env to be able to use the view we'll create in tests
+        cls.env = cls.env(context=dict(cls.env.context, load_all_views=True))
+
+        # For button-invisibility testing we add:
+        # - a dummy M2M field from ``sale.order`` to itself
+        # - a dummy M2M field from ``sale.order`` to ``res.users`` and its inverse
+        # - a dummy function ``action_confirm()`` on ``res.users``
+        # so we can add nested list views w/ other ``action_confirm`` buttons
+        cls.loader = FakeModelLoader(cls.env, cls.__module__)
+        cls.loader.backup_registry()
```
simahawk: you miss the restore_registry call

sale_confirm_group/tests/test_sale_confirm_group_views.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TestSaleConfirmGroupViews(TestSaleConfirmGroupCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Update context env to be able to use the view we'll create in tests
+        cls.env = cls.env(context=dict(cls.env.context, load_all_views=True))
+
+        # For button-invisibility testing we add:
+        # - a dummy M2M field from ``sale.order`` to itself
```
simahawk: this complexity comes from the complexity on `get_view` that we discussed and should be removed

simahawk (changes requested):

simahawk:
> The name should be `sale_confirm_group`

reminder

simahawk:
I've cleaned up the complexity

simahawk (approved):

sale_confirm_group/models/sale_order.py
```diff
@@ -0,0 +1,88 @@
...
+        "company_id.use_sale_confirmation_groups",
+        "company_id.sale_confirmation_group_ids.users",
+    )
+    @api.depends_context("uid")
+    def _compute_user_can_confirm(self):
+        user = self.env.user
+        for company, sales in self.grouped("company_id").items():
+            if not company.use_sale_confirmation_groups:
+                sales.user_can_confirm = True
+            elif not (groups := company.sale_confirmation_group_ids):
```
sebalix, line 31: I'm not a big fan of using the result of walrus operator of 2nd condition in the 3rd one, not clear enough, but it's nitpicking.

sale_confirm_group/models/sale_order.py
```diff
@@ -0,0 +1,88 @@
...
+                sales.user_can_confirm = True
+            else:
+                sales.user_can_confirm = user in groups.users
```
sebalix: We could have an archived user processing orders (e.g. OdooBot with `sale_automatic_workflow` modue), and `user in groups.users` check won't work in that case:
```suggestion
                sales.user_can_confirm = user in groups.with_context(active_test=False).users
```

sebalix (changes requested):

sale_confirm_group/models/sale_order.py
sebalix: I adressed this issue in the 2nd commit.

hildickethan (approved):

PR #3954, opened 2025-10-17
[IMP][14.0] sale_exception: optional="show" on list views

alexis-via:


rousseldenis (approved):

alexey-pelykh (approved):
Looks good to me. Clean and minimal UX improvement.

The `optional="show"` attribute on `main_exception_id` in both the sale order and quotation tree views is a welcome addition -- it allows users to toggle the exception column visibility while keeping it shown by default.

- `optional` attribute is supported since Odoo 13.0, compatible with the 14.0 target branch.
- Both inherited views (`view_order_tree` and `view_quotation_tree`) are updated consistently.
- All CI checks pass.

APPROVED.

PR #3819, opened 2025-07-17
[18.0][MIG] sale_order_invoice_amount: Migration to 18.0

sbiosca-s73:
## What this module does in 18.0

This module adds the **amount_invoiced** and **amount_to_invoice** fields to sale order views:

- In the **form view**, both fields are displayed in the tax totals summary, above the total order amount.
- In the **list view**, both fields are available as optional user fields.

## Version history note

In version 17 and earlier, this module introduced **amount_invoiced** and **amount_to_invoice** as new computed fields on sale orders.  
Since version 18, these fields are already provided by Odoo core. The module now focuses on **UI visibility only**, exposing those existing fields in sale order views.

PR #3955, opened 2025-10-20
[18.0][MIG] sale_order_invoice_amount

DavidJForgeFlow:
Supersede #3819 

Odoo has changed how it computes the invoiced and uninvoiced amounts. Now it works as it was doing the OCA module before, so I removed all the logic that we were doing and only maintain the visuals and tax totals. 
https://github.com/odoo/odoo/commit/eb3d01c853ad84391ddf84393137defb74940522


rousseldenis:
@DavidJForgeFlow Thanks for this.

Usually, it is fair to pick already done commits by sbiosca-s73

DavidJForgeFlow:
> @DavidJForgeFlow Thanks for this.
> 
> Usually, it is fair to pick already done commits by sbiosca-s73

As it is contains almost complete different changes I didn't think of taking the sbiosca-s73 commit. His was a really good migration previous to the 17.0 changes and the Odoo ones in v18.0!

PR #3957, opened 2025-10-21
[18.0][IMP] sale_order_lot_selection: Product document lot compatibility

victoralmau:
Product document lot compatibility

Changes done:
- Added the Document smart button to the lot form view
- Added the Lot field to Product documents
- Added support for displaying lot in documents and taking it into account when uploading documents linked to a lot

Please @pilarvargas-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT57112


victoralmau:
Ping @pilarvargas-tecnativa

christian-ramos-tecnativa (commented):
I think this is a great feature, but I'm not sure it makes sense to include it in this module. IMO, it would fit better in a separate one, maybe called `product_lot_document` or something similar. WDYT

victoralmau:
> I think this is a great feature, but I'm not sure it makes sense to include it in this module. IMO, it would fit better in a separate one, maybe called `product_lot_document` or something similar. WDYT

What do you think @pedrobaeza ?

It is important to clarify other features that will be added and/or will exist:
- Limit product document per lot
- Indicate sale price in lot (each lot has a different sale price because each item sold is completely different and “unique,” but you do NOT want to create a product  variant for each lot). Configurable by company in res.config.settings.
- Display the selector of available lot in the product page in e-commerce. Configurable by company. This will be a new IMO module. Check that this lot is only used in one order when confirmed.
- Change the price when selecting a lot in the product page in e-commerce.
- Only display the images of the selected lot in the product page in e-commerce.

christian-ramos-tecnativa:
Are you going to include all those features in this module? If thats the case you will need to add the website_sale dependency in the manifest changing a lot the module scope. IMHO the modules relation should look like:
`website_sale_lot_selection` or similar (With the features you just mention)
dependences:
- `sale_order_lot_selection`
- `product_lot_document` (with this PR features)

The separation is mainly because both modules bring value by them self, avoiding installing unwanted modules. 
i.e: Would be great to add the documents by lots feature to an intance with just inventory or purchases installed,

victoralmau:
> Are you going to include all those features in this module? If thats the case you will need to add the website_sale dependency in the manifest changing a lot the module scope. IMHO the modules relation should look like: `website_sale_lot_selection` or similar (With the features you just mention) dependences:
> 
>     * `sale_order_lot_selection`
> 
>     * `product_lot_document` (with this PR features)
> 
> 
> The separation is mainly because both modules bring value by them self, avoiding installing unwanted modules. i.e: Would be great to add the documents by lots feature to an intance with just inventory or purchases installed,

No, I don't want to add all those features to this module. I was sure that `website_sale_lot_selection` would need to be created, but I'm not so sure about `product_lot_document` (because I would need another module for the website). 
In any case, it's just an explanation to understand the overall idea so we can decide the best place for each feature.

victoralmau:
Moved to https://github.com/OCA/stock-logistics-warehouse/pull/2449

PR #3959, opened 2025-10-21
[18.0][MIG][IMP] sale_restricted_qty

bosd:
Maybe this is usefull for somebody. It contains the improvements from 17.0 (#2997 ) and the original migration to 18.0.

 - Complete redesign of the inheritance model using an own/inherited/computed pattern. Each quantity setting (min, max, multiple-of)
  now has three layers: sale_own_* (set directly), sale_inherited_* (from parent), and sale_* (effective combined value). This replaces
   the old flat manual_* fields.
  - Mixin refactored with _sale_restricted_qty_parent_field class attribute — a single attribute on child models (parent_id, categ_id,
  product_tmpl_id) drives all 12 inherited field lookups via callable @api.depends, eliminating 24+ repetitive method overrides per
  child model.
  - Mixin renamed from product.restricted.qty.mixin to sale.product.restricted.qty.mixin to reflect its sale-specific scope.
  - Merged duplicated model inheritance into a single clean structure.

Blocking/Warning restriction mode

  - Replaced the old force/not_force/use_parent selection with a cleaner Blocking/Warning (1/0) selection per constraint type (min,
  max, multiple-of), each with own/inherited/effective layers.

 Bug fixes

  - Skip quantity restriction checks for confirmed sale order lines — prevents false validation errors on historical data.
  - Float precision — use float_is_zero with UoM rounding for multiple-of quantity modulo checks to avoid floating-point artifacts.
  - Logic bugs in compute methods fixed and covered by tests.

  Usability

  - Auto-populate quantity with the minimum when a product with an enforced min qty is selected on a sale order line.
  - Order name included in validation error messages for easier identification.
  - Improved field help texts explaining Blocking vs Warning behavior.

Was :men_wrestling: with git, so the improvement commit author is assigned to me instead of alex.


bosd:
Still needs some work..

https://github.com/user-attachments/assets/8f23e0eb-7ca0-4591-a71c-ad231b7219ea

Dranyel-Bosd (approved):
LGTM :100:

bofiltd (approved):
Functional test + Quick code review :100:

sale_restricted_qty/models/product_category.py
```diff
@@ -1,65 +1,142 @@
...
-            rec.update(rec._get_sale_restricted_qty())
+    @api.depends("parent_id.is_sale_min_qty_set")
+    def _compute_is_sale_inherited_min_qty_set(self):
+        return super()._compute_is_sale_inherited_min_qty_set()
+
+    def _get_is_sale_inherited_min_qty_set(self):
+        if not self.parent_id:
+            return super()._get_is_sale_inherited_min_qty_set()
+
+        self.ensure_one()
```
rousseldenis: Why here and not at the function beginning ?

sale_restricted_qty/models/product_restricted_qty_mixin.py
```diff
@@ -1,117 +1,552 @@
...
+RESTRICTION_DISABLED = "0"
+RESTRICTION_SELECTION = [
+    (RESTRICTION_ENABLED, "Yes"),
+    (RESTRICTION_DISABLED, "No"),
+]
 
-class ProductMinMultipleMixin(models.AbstractModel):
+
+class ProductRestrictedQtyMixin(models.AbstractModel):
     _name = "product.restricted.qty.mixin"
```
rousseldenis: As this is 'sale' oriented, I would have renamed "sale.product..."

sale_restricted_qty/models/product_template.py
```diff
@@ -9,41 +10,134 @@ class ProductTemplate(models.Model):
...
+        return super()._compute_sale_inherited_min_qty()
+
+    def _get_sale_inherited_min_qty(self):
+        if not self.categ_id:
+            return super()._get_sale_inherited_min_qty()
+
+        self.ensure_one()
+        return self.categ_id.sale_min_qty
+
+    @api.depends("categ_id.is_sale_restrict_min_qty_set")
```
rousseldenis: To avoid all these overrides (for only depends()) in products and in categories, I would have added a private field in mixin like `_sale_restricted_qty_field` that would have been called in mixin depends.

bosd:
@rousseldenis Thanks for the review, the items hass been adressed. Is it ok now? Can we push this one forward?

rousseldenis (approved):
Seems lighter. LGTM

bosd:
@rousseldenis Can you please merge?

sale_restricted_qty/models/sale.py
```diff
@@ -0,0 +1,209 @@
...
+        "product_uom_qty",
+        "is_min_qty_set",
+        "min_qty",
+        "restrict_min_qty",
+        "is_max_qty_set",
+        "max_qty",
+        "restrict_max_qty",
+        "is_multiple_of_qty_set",
+        "multiple_of_qty",
+        "restrict_multiple_of_qty",
```
alexey-pelykh, line 141: Float modulo (`qty % line.multiple_of_qty`) can produce precision artifacts with floats. Consider `float_is_zero(qty % line.multiple_of_qty, precision_rounding=...)` from `odoo.tools.float_utils` for robustness. Minor concern for typical order quantities.

sale_restricted_qty/migrations/18.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def migrate_selection_and_flags(cr, table):
+    """
+    Map legacy 'manual_force_*' selections to new Blocking/Warning selections
+    and set the new 'is_sale_own_*_set' flags.
+    """
+    for field in ["min_qty", "max_qty"]:
+        old_col = f"manual_force_sale_{field}"
+        new_col = f"sale_own_restrict_{field}"
```
alexey-pelykh, line 33: This loop handles `min_qty` and `max_qty` restrict fields. Was there a `manual_force_sale_multiple_qty` column in the previous version? If so, it would need handling here too.

alexey-pelykh (approved):
Solid migration and redesign. The own/inherited/computed pattern for hierarchical quantity constraints is well-structured and the test coverage is thorough.

A few observations:

1. **Typo in .pot**: `"Sale Miltiple qty"` should be `"Sale Multiple qty"` -- this propagates to all .po files. Likely an auto-extraction artifact, but worth regenerating.

2. **`is_sale_own_*_set` fields not stored on mixin**: `is_sale_own_min_qty_set`, `is_sale_own_max_qty_set`, `is_sale_own_multiple_of_qty_set` are plain `fields.Boolean()` without `store=True` on the mixin definition. This is fine since they default to stored, just noting for consistency with the computed booleans that have explicit `store=True`.

3. **`_compute_restricted_qty_constraints` not stored**: The `is_below_min_qty`, `is_above_max_qty`, `is_not_multiple_of_qty` fields on `sale.order.line` are computed without `store=True`. This is intentional (avoids stale data on qty changes), but means every form/list render triggers recomputation. Acceptable for sale order line volumes.

4. **Float modulo in multiple-of check**: `qty % line.multiple_of_qty != 0` uses float modulo which can have precision issues. Consider using `float_compare` or `float_round` from `odoo.tools` for robustness, though for typical order quantities this is unlikely to cause real issues.

5. **Migration handles `multiple_of` restrict**: The migration script handles `min_qty` and `max_qty` restrict fields but not `multiple_of_qty` restrict (the `for field in ["min_qty", "max_qty"]` loop in `migrate_selection_and_flags`). If the previous version had a `manual_force_sale_multiple_qty` column, it would be missed. If it didn't exist, this is fine.

6. **Legacy .pot file**: `sale_order_min_qty.pot` is included alongside `sale_restricted_qty.pot`. Should this be removed if the module was renamed?

Overall LGTM. The architecture is clean, tests cover the inheritance chain well including UoM conversion edge cases, and the migration script handles the data transition correctly.

marielejeune:
@rousseldenis I think the failing tests are not related to this PR. Moreover with 4 approved reviews, it may be time to merge it! Thanks

rousseldenis:
> @rousseldenis I think the failing tests are not related to this PR. Moreover with 4 approved reviews, it may be time to merge it! Thanks

Merge will not pass if tests are red. Usually, someone (a good will) will fix them in order to merge :smile:

marielejeune:
@bosd can you rebase your branch?

BhaveshHeliconia (approved):
Functional review LGTM!

marielejeune:
Hi all,

I'm a bit surprised the way this PR was done. 
I initially thought it just superseeded the original PR of @samirGuesmi (straightforward migration), with some new merged PRs in v17.0

However, updating the sources in our project, I see that the UX completely changed.
I agree this is better and a good improvement, however I notice these improvements come from a 17.0 non merged (and not fully reviewed) PR: https://github.com/OCA/sale-workflow/pull/2997

I find it a bit strange to include an old open PR in the migration process of a higher version, without explicitly mentioning it in the description. 

In my opinion before merging this 18.0 PR we should finish the 17.0 UX improvements PR from @alexey-pelykh so that we're sure it doesn't change, merge it, and then merge this 18.0 PR.

What do you think?

sale_restricted_qty/models/product_restricted_qty_mixin.py
```diff
@@ -7,8 +7,8 @@
 RESTRICTION_ENABLED = "1"
 RESTRICTION_DISABLED = "0"
 RESTRICTION_SELECTION = [
-    (RESTRICTION_ENABLED, "Yes"),
-    (RESTRICTION_DISABLED, "No"),
+    (RESTRICTION_ENABLED, "Blocking"),
+    (RESTRICTION_DISABLED, "Warning"),
```
marielejeune, line 11: Just repeating the comment I've added on the 17.0 PR: 

Would be better to have real chars as selection values, it eases the code comprehension:

I would suggest something like RESTRICTION_ENABLED = "enabled", "restricted" or "blocking", and RESTRICTION_DISABLED = "disabled", "unrestricted" or "warning".

rousseldenis: @bosd

rousseldenis:
@bosd Could you put in PR description the different improvements you've added as this is not only a migration ? Thanks

sergiocorato:
> Hi all,
> 
> I'm a bit surprised the way this PR was done. I initially thought it just superseeded the original PR of @samirGuesmi (straightforward migration), with some new merged PRs in v17.0
> 
> However, updating the sources in our project, I see that the UX completely changed. I agree this is better and a good improvement, however I notice these improvements come from a 17.0 non merged (and not fully reviewed) PR: #2997
> 
> I find it a bit strange to include an old open PR in the migration process of a higher version, without explicitly mentioning it in the description.
> 
> In my opinion before merging this 18.0 PR we should finish the 17.0 UX improvements PR from @alexey-pelykh so that we're sure it doesn't change, merge it, and then merge this 18.0 PR.
> 
> What do you think?

It's possible to do these improvements here, if there is no one interested in v. 17.0 (I wouldn't personally check that version too).

Anyway, other then explain it, it could lead to a slower process of the migration, so it is not always a good idea.

In every case, commits of the improvements have to be squashed in 1 commit only.

sergiocorato (changes requested):
Squash improvement commits and explain their logic in the first comment of the PR at least, no code check.

sale_restricted_qty/models/product_restricted_qty_mixin.py
bosd: @marielejeune Good point!
I would propose to do that in a follow-up pr. 
We already using this module in production. 

Introducing a version bump in this pr is undesirable, as well as adding a complex migration script for changing it withing the same version, would increase complexity for this nice to have. 
IMO, better to handle it in a different pr.

bosd:
Done, commits squashed.

The work on this pr, and the one prior to it on version 17 have been open way too long.
Can we move forward with this?

sergiocorato:
> Done, commits squashed.
> 
> The work on this pr, and the one prior to it on version 17 have been open way too long. Can we move forward with this?

I don't see that the commits had been squashed, anyway I'm not a maintainer of this repo, so you should ping them.
Maybe a faster approach is to do here the simple migration and the improvements in a different PR.

Screenshot(226)

bosd:
There where more commits :)
IMO hiding improvement commits in migration commits is'nt really usefull.
It pollutes the migration commit and makes backporting harder.

Ideally we would have a clean migration pr and a improvement pr as a follow up. 
Then the improvements would land in their own commits respectively.

Since it took almost 2 years for the V17 migration to get  merged https://github.com/OCA/sale-workflow/pull/2996#issuecomment-3957573034 ,I combined the migration and improvements in this one pr.

(Sorry my customer cannot wait that long, we need to take it to production. )

rousseldenis:
> In every case, commits of the improvements have to be squashed in 1 commit only.

@sergiocorato Please, don't ask such things. It forbids forward/backports of those improvements as they are mixed in migration commit. The migration commit should contain only migration code.

sergiocorato:
> > In every case, commits of the improvements have to be squashed in 1 commit only.
> 
> @sergiocorato Please, don't ask such things. It forbids forward/backports of those improvements as they are mixed in migration commit. The migration commit should contain only migration code.

You're right, thanks for the correction.

bofiltd:
@rousseldenis Can you please merge? :pray:

PR #3960, opened 2025-10-21
[19.0][MIG] sales_team_security: Migration to 19.0

lef-adhoc:


matiasperalta1 (approved):
LGTM

lef-adhoc:
@pedrobaeza @ivantodorovich Hi, could you review this PR?

ivantodorovich (approved):
LG. Thanks!

lef-adhoc:
@chienandalu @carlos-lopez-tecnativa Hi, could you review this PR?

chienandalu (approved):
Just code review 👍

sales_team_security/models/ir_rule.py
```diff
@@ -41,5 +42,15 @@ def _compute_domain(self, model_name, mode="read"):
...
                 extra_domain = domain_followers | domain_user
+
+                if user.has_group(group_team_manager):
+                    user_teams = user.crm_team_member_ids.mapped("crm_team_id")
+                    if user_teams:
+                        team_user_ids = user_teams.mapped(
+                            "crm_team_member_ids.user_id"
+                        ).ids
+                        domain_team = Domain([("user_id", "in", team_user_ids)])
+                        extra_domain = extra_domain | domain_team
```
carlos-lopez-tecnativa, line 53: I have these users that belong to the same team.
Marc Demo has the group: `User: Team Documents Only`
Joe has the group: `User: Own Documents Only`
image

The sales team is configured as shown above.
image


A customer is created with Joe as the salesman.
image

But Marc Demo cannot see the contact. I expected Marc Demo to see it.
What am I missing in my configuration?
image

sales_team_security/readme/DESCRIPTION.md
```diff
@@ -24,3 +24,7 @@ the permission scheme. Someone with this permission will see:
 For keeping consistent accesses, followers of the main and
 shipping/invoice contacts are synced according the salesman of the
 children contacts
+
+Additionally, users with the "User: Team documents" group can see contacts
```
carlos-lopez-tecnativa: This group name does not exist. Are you referring to `User: Team Documents Only`?

carlos-lopez-tecnativa (changes requested):

sales_team_security/readme/DESCRIPTION.md
lef-adhoc: Locally, the same case works for me, but not in Runboat. I'll keep checking...

sales_team_security/models/ir_rule.py
lef-adhoc: @carlos-lopez-tecnativa In local development, it works for me, but not in Runboat. I can't understand why. Could you try it?

sales_team_security/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,30 @@
...
+
+- Contacts without salesman assigned.
+- Contacts with them as salesman.
+- Contacts with them as follower.
+
+For keeping consistent accesses, followers of the main and
+shipping/invoice contacts are synced according the salesman of the
+children contacts
+
+Additionally, users with the "User: Team documents Only" group can see contacts
```
alexey-pelykh, line 28: The group name in the XML definition is **"User: Team Documents Only"** (all words capitalized, with "Only"). This line says `"User: Team documents"` (lowercase "d", missing "Only"). Two lines below (line 30) it says `"User: Team documents Only"` which is closer but still has a lowercase "d".

Please align both occurrences with the canonical group name from `security/sales_team_security.xml`:
```
User: Team Documents Only
```

This was already flagged by @carlos-lopez-tecnativa -- still not addressed in the latest push.

sales_team_security/readme/CONFIGURE.md
```diff
@@ -0,0 +1,3 @@
+1.  Go to *Configuration \> Users & Companies \> Users*.
+2.  Open or create a user.
+3.  On the section "Sale", select "User: Team documents".
```
alexey-pelykh, line 3: Same issue: `"User: Team documents"` should read **"User: Team Documents Only"** to match the group name defined in the XML.

sales_team_security/models/ir_rule.py
```diff
@@ -0,0 +1,56 @@
...
+                        ("id", "=", user.partner_id.id),
+                    ]
+                )
+                domain_user = Domain([("user_id", "in", [user.id, False])])
+                extra_domain = domain_followers | domain_user
+
+                if user.has_group(group_team_manager):
+                    user_teams = user.crm_team_member_ids.mapped("crm_team_id")
+                    if user_teams:
+                        team_user_ids = user_teams.mapped(
```
alexey-pelykh, line 49: Regarding the functional concern raised by @carlos-lopez-tecnativa: the domain logic itself looks correct to me. `user.crm_team_member_ids` resolves via the `res.users` model and doesn't go through partner record rules, so there shouldn't be a recursive access issue.

One thing worth checking on Runboat: `crm_team_member_ids` is a One2many on `res.users` that depends on `crm.team.member` records existing. If the test user on Runboat was assigned to a team via the team form (which sets `member_ids` on the team) but the `crm.team.member` record wasn't properly created for the user, `user.crm_team_member_ids` would return an empty recordset and the domain would never include teammates.

Also note that `_compute_domain` is `ormcache`'d -- after changing group assignments or team membership in a running instance, the rule cache needs to be cleared (this happens automatically on module install/update, but manual group changes via UI rely on `ir.rule` cache invalidation signals). Could explain the discrepancy between a fresh install (local) and Runboat (possibly stale cache).

alexey-pelykh (commented):
Reviewed the 18.0 -> 19.0 migration plus the new team-member contact visibility feature.

Migration mechanics look correct:
- `expression.OR`/`expression.AND` replaced with `Domain` operators (`|`, `&`) per Odoo 19.0 API.
- `category_id` replaced with `privilege_id` in the security group definition -- correct for the 19.0 privilege system.
- Version bump to `19.0.1.0.0`.
- Tests pass in CI (Odoo + OCB), pre-commit green.

The new team-manager domain logic in `_compute_domain` is sound: it correctly gates on `group_sale_team_manager` inside the existing `group_my_records and not group_all_records` guard, and the `Domain` OR composition is idiomatic.

There is one documentation inconsistency and one open functional concern (the Runboat issue raised by @carlos-lopez-tecnativa). See inline comments.

Overall the code is clean and well-tested. The documentation nit should be addressed before merge.

sales_team_security/models/ir_rule.py
rousseldenis: @lef-adhoc @carlos-lopez-tecnativa Could you synchronize in order to go forward on this ? Thanks

carlos-lopez-tecnativa: Tested locally, and it works.

I'm not sure why it isn't working properly on Runboat, but I remember seeing a similar issue there before. Anyway, let's go!

sales_team_security/tests/common.py
```diff
@@ -0,0 +1,126 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.team = cls.env["crm.team"].create({"name": "Test channel"})
+        cls.team2 = cls.env["crm.team"].create({"name": "Test channel 2"})
+        cls.user = cls.env["res.users"].create(
+            {
+                "login": "sales_team_security",
+                "name": "Test sales_team_security user",
+                "group_ids": [(4, cls.env.ref("sales_team.group_sale_salesman").id)],
```
carlos-lopez-tecnativa:
```suggestion
                "group_ids": [Command.link(cls.env.ref("sales_team.group_sale_salesman").id)],
```

sales_team_security/tests/common.py
```diff
@@ -0,0 +1,126 @@
...
+        self._check_permission(self.user2, False, True)
+        self._check_permission(False, self.team, True)
+        self._check_permission(False, self.team2, True)
+        self._check_permission(self.user, self.team, True)
+        self._check_permission(self.user, self.team2, True)
+        self._check_permission(self.user2, self.team2, True)
+        self._check_permission(self.user2, self.team, True)
+        # Regular internal user
+        if extra_checks and is_res_partner:
+            self.user.group_ids = [(6, 0, [self.env.ref("base.group_user").id])]
```
carlos-lopez-tecnativa:
```suggestion
            self.user.group_ids = [Command.set([self.env.ref("base.group_user").id])]
```

carlos-lopez-tecnativa (approved):
LGTM. If you could also replace the old ORM command syntax with the new `Command.link()`, `Command.set()`, etc., that would be great.

https://github.com/odoo/odoo/blob/19.0/odoo/orm/commands.py

BhaveshHeliconia (approved):
Tested locally LGTM!

rousseldenis:
@lef-adhoc Could you take into account @carlos-lopez-tecnativa comments ? Thanks

lef-adhoc:
@rousseldenis Done

rousseldenis (approved):

PR #3963, opened 2025-10-22
[19.0][MIG] sales_team_security_crm: Migration to 19.0

lef-adhoc:
Depends on:

- [ ] <https://github.com/OCA/sale-workflow/pull/3960>


sales_team_security_crm/tests/test_sales_team_security_crm.py
```diff
@@ -0,0 +1,21 @@
...
+            {
+                "name": "Test lead",
+                "partner_id": cls.partner.id,
+                "user_id": cls.user.id,
+                "team_id": cls.team.id,
+            }
+        )
+
+    def test_crm_lead_permissions(self):
+        self._check_whole_permission_set(extra_checks=False)
```
alexey-pelykh, line 21: The 17.0 version had an additional `test_crm_lead_permissions_multi_team` that created a third team and verified cross-team visibility. Since `membership_multi` is now always enabled in 19.0, the `set_param` call is correctly removed, but the test scenario itself (user in team1 + team3, checking that a record in team3 is visible) is still a valid assertion. Consider restoring it:

```python
def test_crm_lead_permissions_multi_team(self):
    team3 = self.env["crm.team"].create({"name": "Test channel 3"})
    self.env["crm.team.member"].create(
        {"user_id": self.user.id, "crm_team_id": team3.id}
    )
    self._check_permission(False, team3, True)
```

sales_team_security_crm/i18n/sales_team_security_crm.pot
```diff
@@ -0,0 +1,13 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 15.0\n"
```
alexey-pelykh, line 6: Nit: `Project-Id-Version` is `Odoo Server 15.0` -- stale from a previous migration. Not blocking since this gets overwritten by OCA tooling, but a fresh `pot` export would clean it up.

alexey-pelykh (commented):
Clean migration, the module structure is correct and CI tests pass (the unreleased-deps failure is expected while #3960 is pending). A few observations:

**i18n/sales_team_security_crm.pot** -- `Project-Id-Version` says `Odoo Server 15.0`. This gets regenerated by `oca-gen-addon-readme` / `setup/` tooling so it's cosmetic, but worth noting it wasn't regenerated for 19.0.

**Dropped multi-team test** -- In 17.0 there was a `test_crm_lead_permissions_multi_team` test that created a third team and validated cross-team visibility. The `membership_multi` config param is indeed gone in 19.0 (multi-team is now always enabled), so removing the `set_param` call is correct. However, the test case itself still exercised a useful code path (user belongs to team1 + team3, record assigned to team3 -- verifiable without the param). Consider keeping it.

**sales_team_team_rule** -- The rule `[('id', 'in', user.crm_team_ids.ids)]` on `crm.team` restricts team managers to only see their own teams. This is carried over unchanged from 17.0 and is intentional, but note it does **not** include `False` in the list, so unassigned team situations won't match. This is consistent with the design intent (team managers see their teams, not all teams).

Overall this is a straightforward migration. The security rules, manifest, tests, and `auto_install` glue pattern are all correct for 19.0.

CarlosRoca13:
Please, can you check red CI?

PR #3964, opened 2025-10-23
[18.0][MIG] sale_stock_warehouse_partner: Migration to 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT58152
could you please review?


sale_stock_warehouse_partner/views/res_partner_views.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--Copyright 2025 Tecnativa - Eduardo Ezerouali
+License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)-->
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
```
christian-ramos-tecnativa, line 7: The field `property_stock_supplier` is in the view `stock.view_partner_stock_form` so this change make the extension more robust
```suggestion
        <field name="inherit_id" ref="stock.view_partner_stock_form" />
```

sale_stock_warehouse_partner/models/sale_order.py
```diff
@@ -0,0 +1,28 @@
...
+
+    warehouse_id = fields.Many2one(
+        "stock.warehouse",
+        string="Warehouse",
+        required=True,
+        readonly=False,
+        compute="_compute_warehouse_id",
+        store=True,
+        check_company=True,
+    )
```
christian-ramos-tecnativa, line 18: This field already exists in the `sale.order` with the same attributes but the precompute. Is this needed?

christian-ramos-tecnativa (changes requested):
Can you check the CI? I think it might be related to the changes in the `company_dependent` fields in this version

sale_stock_warehouse_partner/views/res_partner_views.xml
eduezerouali-tecnativa: Thanks for the advice

sale_stock_warehouse_partner/models/sale_order.py
eduezerouali-tecnativa: Actually still working on this, as there a more things that change since 15.0. You are right no need for the field just only override de compute and adapt it. :)

eduezerouali-tecnativa:
@pedrobaeza I did discover that there is already a module that covers this possibility https://github.com/OCA/sale-workflow/tree/18.0/sale_stock_partner_warehouse. Anyway I will explore if there is any chance of improvement the exiting one. Also I have one big concern about module name, should I rename the module I am migrating from as this one was before the other to maintain consistency?

pedrobaeza:
Seeing the goal of the module, `sale_stock_partner_warehouse` is a better name, as the name goes from more general to more specific, so it should be read as "warehouse at partner level". Current name can be interpreted as "putting a partner in the warehouse" instead.

PR #3965, opened 2025-10-23
[19.0][MIG] sale_triple_discount: Migration to 19.0

lef-adhoc:
Based on this pr: https://github.com/OCA/sale-workflow/pull/3889

Depends on:

- [x] <https://github.com/OCA/account-invoicing/pull/2144>


ferran-S73 (approved):
Code review

lef-adhoc:
@sbejaoui Hi, could you check this PR?

Reyes4711-S73 (approved):
LGTM

Alexgars73:
@lef-adhoc could you change the class text-right to text-end? text-right is obsolete and has no effect. Thanks!

WesleyOliveira98 (approved):

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,177 @@
...
+        self.ensure_one()
+        if self.discounting_type == "additive":
+            return self._additive_discount()
+        elif self.discounting_type == "multiplicative":
+            return self._multiplicative_discount()
+        else:
+            raise ValidationError(
+                self.env._(
+                    "Sale order line %(name)s has unknown discounting type "
+                    "%(disc_type)s",
```
alexey-pelykh, line 87: The `.pot` and `.po` files contain `%(dic_type)s` but this code passes `disc_type=` as the keyword argument. The translation template is out of sync with the source. Regenerate the `.pot` to fix; otherwise translated error messages will raise `KeyError` at runtime.

sale_triple_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,177 @@
...
+    @api.depends("discount")
+    def _compute_discounts(self):
+        # We intentionally call super to avoid triggering the full compute logic,
+        # which would recalculate 'discount' from the three discount fields
+        # At this stage, those fields may not be updated yet, and calling
+        # _compute_discount() directly would overwrite the current value
+        # calling super ensures the cache is coherent before applying the inverse
+        super()._compute_discount()
+        self._inverse_discount()
+        return True
```
alexey-pelykh, line 134: Compute methods in Odoo 19.0 should not return a value. Remove the `return True`.

alexey-pelykh (commented):
## Migration Review

Good migration overall. The core discount logic, ORM patterns, view XML, and tests are solid. A few issues to address:

### Blocking

1. **Translation template mismatch** -- The `.pot` file (and all `.po` files) reference `%(dic_type)s` but the Python source uses `%(disc_type)s`. These are different placeholder names. The `.pot` needs regeneration to match the actual source string. Translations referencing the wrong placeholder will produce a `KeyError` at runtime.

2. **`_compute_discounts` returns `True`** -- Odoo 19.0 compute methods should not return a value. Returning `True` is a legacy pattern. Remove the `return True` statement.

### Non-blocking

3. **CI is red** -- All three test workflows are failing. The "Detect unreleased dependencies" failure is expected given the dependency on `OCA/account-invoicing#2144`. However, the actual test runs also fail, which should be investigated once the dependency PR is merged.

4. **`_additive_discount` clamps negatives to 0** -- The additive method returns 0 for negative sums, but the multiplicative method allows negative results (surcharges). This behavioral inconsistency could confuse users if additive mode is ever fully enabled. Since additive is currently blocked via `_discounting_type_additive_not_allowed`, this is informational only.

5. **`create` override fragility** -- The logic in `create` that conditionally calls `_compute_discounts` based on comparing `discount` and `discount1` values with `vals` is complex and could be fragile across Odoo upgrades. Consider adding a comment explaining the exact scenario this handles (e.g., when `discount` comes from a pricelist vs when it is manually set).

BhaveshHeliconia (commented):
`Additive` discount type missing due to open issue (https://github.com/OCA/sale-workflow/issues/3649), but LGTM for `Multiplicative` discount type

tao-thewarat (approved):
LGTM😎

nicolascol:
Hi @alexey-pelykh 
Could you merge please?

PR #3968, opened 2025-10-24
[19.0][MIG] sale_tier_validation: Migration to 19.0

niboo-vli:


bosd:
@niboo-vli Thanks for the migration. Can you please add a `test-requirements.txt` ?

```
git+https://github.com/em230418/tier-validation@19.0-mig-base_tier_validation#subdirectory=base_tier_validation
```

So it depends on: https://github.com/OCA/tier-validation/pull/1

OpenCode:
I think you can create a PR to migrate the module directly in https://github.com/OCA/tier-validation

sale_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2020 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```
is-ife, line 2: AGPL-3.0 or LGPL-3.0?

bosd:
> I think you can create a PR to migrate the module directly in https://github.com/OCA/tier-validation

Looks like there already is one: https://github.com/OCA/tier-validation/pull/9

sale_tier_validation/tests/test_tier_validation.py
alexey-pelykh, line 2: License should be `AGPL-3.0` to match the module manifest (`"license": "AGPL-3"`). Currently says `LGPL-3.0`.

sale_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,83 @@
...
+                            "product_id": cls.product.id,
+                            "product_uom_qty": 1,
+                            "product_uom": cls.product.uom_id.id,
+                            "price_unit": cls.product.list_price,
+                        },
+                    )
+                ],
+                "pricelist_id": cls.customer.property_product_pricelist.id,
+            }
+        )
```
alexey-pelykh, line 56: In Odoo 17+, `property_product_pricelist` on `res.partner` was refactored. Verify this still resolves correctly in 19.0. If `pricelist_id` is no longer required on `sale.order`, you can drop this line entirely -- other 19.0 modules (e.g. `sale_exception`) create the pricelist explicitly rather than relying on the partner property.

alexey-pelykh (changes requested):
Thanks for porting this module.

The main blocker right now is the **unreleased dependency**: `base_tier_validation` has not been merged to 19.0 yet on `OCA/server-ux` (see [server-ux#1234](https://github.com/OCA/server-ux/pull/1234) -- tests pass but pre-commit still fails there). This explains both CI test failures (`test with Odoo` and `test with OCB`).

Once that dependency lands, there are a few things to address in this PR:

1. **License mismatch in test file** (already flagged by @is-ife): `tests/test_tier_validation.py` header says `LGPL-3.0` but the manifest declares `AGPL-3`. Should be `AGPL-3.0` to match.

2. **`pricelist_id` in test setup**: The test creates a sale order with `"pricelist_id": cls.customer.property_product_pricelist.id`. In Odoo 17+, pricelist handling on `res.partner` changed. Verify that `property_product_pricelist` still resolves correctly in 19.0 and that the `pricelist_id` field is still required on `sale.order`. If not, this line can be dropped.

3. **Translation function**: The `self.env._()` pattern in `reports/sale_report.py` is correct for 19.0 -- good catch on updating that from the 18.0 `_()` import.

Code-wise the module looks solid -- clean model inheritance, proper tier validation hookup via `_state_from`/`_state_to`, correct notification subtypes, and adequate test coverage. No issues with the XML views or data files.

Please rebase once `base_tier_validation` is available on 19.0 and fix the license header.

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4197 to commit history before migration commit.

rousseldenis:
@niboo-vli Could you attend comments ?

bosd:
### Heads-up: filter domain breaks registry load on Odoo 19

Running `-i sale_tier_validation` on a 19.0 instance crashes at view load with:

```
odoo.tools.convert.ParseError: while parsing sale_tier_validation/views/sale_order_view.xml:5
Error while validating view near:

  <search string="Search Sales Order">
  …

Unknown field "sale.order.validated" in domain of <filter name="tier_validated"> ([('validated', '=', True)])
```

#### Cause

`base_tier_validation` 19.0 still exposes `validated` as a non-stored compute Boolean with `search=_search_validated`:

```python
# TODO: Delete in v19 in favor of validation_status field
validated = fields.Boolean(
    compute="_compute_validated_rejected", search="_search_validated"
)
```

Odoo 19's view validator no longer accepts that pattern in a search-view domain when the field comes in via an `AbstractModel` `_inherit`. The view dies before the install can complete, so the whole registry init fails.

#### Suggested fix

Swap the filter to the stored `validation_status` Selection field (already defined in `base_tier_validation`, value `'validated'` is the same end-state):

```xml
<filter
    name="tier_validated"
    string="Validated"
    domain="[('validation_status', '=', 'validated')]"
    help="SOs validated and ready to be confirmed"
/>
```

This matches the `base_tier_validation` TODO note that `validated` is being phased out in favor of `validation_status` in v19. One-line view change, no model changes needed; tested on a real 19.0 install (where the unpatched filter reproduces the crash above and the patched one loads cleanly).

BhaveshHeliconia:
@niboo-vli this PR should be closed because the module was moved to this repository: [OCA/tier-validation](https://github.com/OCA/tier-validation?utm_source=chatgpt.com)

PR #3971, opened 2025-10-27
19.0 [MIG] sale order line input

Wodran14:


sale_order_line_input/views/sale_order_line_view.xml
```diff
@@ -0,0 +1,240 @@
...
+                    }"
+                    options="{
+                        'no_open': True,
+                    }"
+                    optional="show"
+                    domain="[('sale_ok', '=', True)]"
+                    widget="sol_product_many2one"
+                    column_invisible="True"
+                    placeholder="Type to find a product..."
+                />
```
alexey-pelykh, line 90: The `uom` context key in `product_uom_qty` is still set to `product_uom_id`. In the Odoo 19.0 core sale order form, this context key is `product_uom_id` as well, so this is correct. Just noting for clarity.

sale_order_line_input/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+            line.company_id = line.force_company_id
+
+    @api.onchange("order_partner_id")
+    def _onchange_order_partner_id(self):
+        """Create order to correct compute of taxes"""
+        if not self.order_partner_id or self.order_id:
+            return
+        SaleOrder = self.env["sale.order"]
+        new_so = SaleOrder.new({"partner_id": self.order_partner_id})
+        for onchange_method in new_so._onchange_methods["partner_id"]:
```
alexey-pelykh, line 70: This onchange creates a real `sale.order` record via `SaleOrder.create(order_vals)` as a side effect. This pattern is carried from earlier versions and works, but it means an incomplete form interaction leaves orphaned sale orders in the database. Consider guarding or deferring creation in a future improvement.

alexey-pelykh (approved):
Straightforward 18.0 → 19.0 migration. CI is green (pre-commit, tests with Odoo and OCB all pass). The Python model is unchanged, and the view/test changes correctly handle the Odoo 19.0 field renames.

**What was done:**
- Manifest version bump to 19.0.1.0.0
- `product_uom` → `product_uom_id` in view XML, search filter group-by, and tests
- `tax_id` → `tax_ids` in view XML
- `product_packaging_id` / `product_packaging_qty` fields dropped from tree view (removed in Odoo 19.0 sale module)
- `uom` context key updated to `product_uom_id` in product field contexts

**Notes:**
- The `_onchange_order_partner_id` method still uses the legacy `_onchange_methods` pattern with `_convert_to_write`. This works in 19.0 but is increasingly dated — worth refactoring in a follow-up to use computed fields or `Command` API. Not blocking for a migration PR.
- Codecov fails are expected (module-level coverage threshold, not a regression).

rousseldenis (approved):

PR #3973, opened 2025-10-28
[19.0][MIG] sale_order_line_sequence: Migration to 19.0

BhaveshHeliconia:


Quan-nhm (approved):
LGTM, code review and functional test.

sale_order_line_sequence/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+    @api.depends("move_id.invoice_line_ids")
+    def _compute_related_so_sequence(self):
+        for rec in self:
+            if len(rec.move_id.line_ids.sale_line_ids.order_id) > 1:
+                rec.related_so_sequence = (
+                    f"{rec.sale_line_ids.order_id.name}/"
+                    f"{rec.sale_line_ids.visible_sequence}"
+                )
+            else:
+                rec.related_so_sequence = str(rec.sale_line_ids.visible_sequence)
```
alexey-pelykh, line 32: Regression from 18.0. When `rec.sale_line_ids` contains multiple records, `rec.sale_line_ids.order_id.name` on a multi-record recordset will concatenate names (or raise, depending on Odoo version) rather than producing the expected `SO001/1, SO002/2` format.

Also missing the `len(rec.sale_line_ids) > 1` check that was in 18.0.

Please restore the 18.0 implementation:
```python
@api.depends("move_id.invoice_line_ids")
def _compute_related_so_sequence(self):
    for rec in self:
        if (
            len(rec.move_id.line_ids.sale_line_ids.order_id) > 1
            or len(rec.sale_line_ids) > 1
        ):
            sequences = sorted(
                set(
                    (sl.visible_sequence, sl.order_id.name)
                    for sl in rec.sale_line_ids
                )
            )
            rec.related_so_sequence = ", ".join(
                f"{seq[1]}/{seq[0]}" for seq in sequences
            )
        else:
            rec.related_so_sequence = str(
                rec.sale_line_ids.visible_sequence
            )
```

sale_order_line_sequence/tests/test_sale_order_line_sequence.py
```diff
@@ -0,0 +1,116 @@
...
+            {
+                "name": "E-commerce Box",
+                "default_code": "E-COM12",
+                "weight": 0.01,
+            }
+        )
+
+    def _create_sale_order(self):
+        sale_type = self.env["sale.order.type"].search([], limit=1)
+        vals = {
```
alexey-pelykh, line 36: `sale.order.type` comes from `sale_order_type`, which is not in this module's `depends`. This will break on databases without that module.

The 18.0 tests did not reference this model. Please remove the `sale_type` lookup (line 37) and the `type_id` key (line 40).

sale_order_line_sequence/tests/test_sale_order_line_sequence.py
```diff
@@ -0,0 +1,116 @@
...
+        so.order_line.qty_delivered = 5
+        self.invoice = so._create_invoices()
+        self.assertEqual(
+            str(so.order_line[0].visible_sequence),
+            self.invoice.line_ids[0].related_so_sequence,
+        )
+        self.assertEqual(
+            str(so.order_line[1].visible_sequence),
+            self.invoice.line_ids[1].related_so_sequence,
+        )
```
alexey-pelykh, line 114: The `test_combined_invoice_sequence` test from 18.0 was dropped. That test validates the multi-SO / merged-invoice-line scenario -- exactly the case broken by the `_compute_related_so_sequence` regression. Please re-add it, adapting for 19.0 conventions (`Command.create`, `product_uom_id`, `BaseCommon`).

alexey-pelykh (changes requested):
Thanks for the migration work. I found two issues that need to be addressed before this can be merged.

### 1. Regression in `_compute_related_so_sequence`

The 18.0 version correctly handles invoice lines linked to **multiple sale order lines** (e.g. merged invoice lines across SOs). It iterates over `rec.sale_line_ids`, builds sorted `(visible_sequence, order_name)` tuples, and joins them.

The 19.0 version simplified this and accesses `rec.sale_line_ids.order_id.name` / `rec.sale_line_ids.visible_sequence` directly in the f-string. When `sale_line_ids` is a multi-record recordset, this will produce incorrect results (concatenated strings like `SO001,SO002/1,2` instead of `SO001/1, SO002/2`). The condition also dropped the `len(rec.sale_line_ids) > 1` check.

The corresponding `test_combined_invoice_sequence` test that validated this scenario in 18.0 was also dropped.

### 2. Test references undeclared `sale.order.type`

`_create_sale_order` uses `self.env["sale.order.type"].search([], limit=1)` and passes `type_id` to the SO vals. This model comes from `sale_order_type`, which is **not** in this module's `depends`. The 18.0 tests did not use this. This will raise `KeyError` on a database without `sale_order_type` installed.

CI passed because `sale_order_type` likely happens to be present in the test environment, but the test itself is not portable.

BhaveshHeliconia:
@alexey-pelykh Thanks for the detailed review and for pointing these issues out. I've addressed the points you raised. Please have another look when you get a chance.

alexey-pelykh (approved):
All three items from my previous review are resolved:

1. `_compute_related_so_sequence` regression: fixed — multi-SO/merged-invoice logic restored from 18.0.
2. `sale.order.type` test dependency: removed.
3. `test_combined_invoice_sequence`: re-added and passing.

Code and tests look correct. CI green.

rousseldenis (approved):

PR #3974, opened 2025-10-28
[19.0][MIG] sale_order_price_recalculation: Migration to 19.0

BhaveshHeliconia:


sale_order_price_recalculation/views/sale_order_view.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="name">sale.order.form - Add recompute buttons</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <field name="state" position="before">
+                <button
+                    name="action_update_prices"
+                    class="ml-2 btn btn-default"
+                    string="Recalculate prices"
+                    Help="Apply the current pricelist"
```
alexey-pelykh: Pre-existing nit: `Help` (capital H) is not recognized by the Odoo view engine — it should be lowercase `help`. The tooltip is silently discarded. Same on line 20. Not introduced by this PR, but worth fixing while migrating.

alexey-pelykh (approved):
Clean migration to 19.0. The key adaptations are correct:

- Test class properly migrated from `TransactionCase`/`setUp` to `BaseCommon`/`setUpClass` with `@classmethod`
- `product_uom` field renamed to `product_uom_id` matching the Odoo 19 rename in `sale.order.line`
- `group.users = [(4, ...)]` updated to `group.user_ids = [Command.link(...)]`
- Tuple-style commands replaced with `Command` API
- `cls.partner` correctly inherited from `BaseCommon` instead of being created manually

CI is green. Model, view, and manifest look fine. `action_update_prices` is a standard Odoo method exposed via button; `action_update_names` is a thin wrapper calling `_compute_name()`.

One pre-existing nit (not blocking): the `Help` attribute on both buttons in `sale_order_view.xml` should be lowercase `help` (lines 13, 20). XML attributes are case-sensitive, so the capitalized `Help` is silently ignored and the tooltip never shows. This has been carried since older versions, so consider fixing separately.

sale_order_price_recalculation/views/sale_order_view.xml
BhaveshHeliconia: Thanks for pointing this out. You're right — the Help attribute should be lowercase help.

rousseldenis (approved):
LGTM

PR #3976, opened 2025-10-29
[18.0][FIX] product_customerinfo_sale: pass partner context from sale to get correct price

JordiMForgeFlow:
Partner context was otherwise not passed when prices are calculated in the SO. The module was not even testing that the prices were passed correctly.

### DIFF
--- a/product_customerinfo_sale/tests/test_product_customerinfo_sale.py
+++ b/product_customerinfo_sale/tests/test_product_customerinfo_sale.py
@@ -106,6 +106,11 @@ class TestProductCustomerInfoSale(BaseCommon):
             self.customerinfo.min_qty,
             "Error: Min qty was not passed to the sale order line",
         )
+        self.assertEqual(
+            line.price_unit,
+            100.0,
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
:+1: Thanks for covering it in a test case

GuillemCForgeFlow (approved):

AaronHForgeFlow (approved):
code review :+1:

product_customerinfo_sale/tests/test_product_customerinfo_sale.py
```diff
@@ -106,6 +106,11 @@ def test_product_customerinfo(self):
             self.customerinfo.min_qty,
             "Error: Min qty was not passed to the sale order line",
         )
+        self.assertEqual(
+            line.price_unit,
+            self.customerinfo.price,
```
pedrobaeza: Hardcode the expected price, not a "variable".

pedrobaeza (approved):
/ocabot merge patch

PR #3977, opened 2025-10-29
[18.0][ADD] sale_down_payment_deduct_partial

LauraCForgeFlow:


AaronHForgeFlow (approved):
Functional review :+1:

AaronHForgeFlow (changes requested):
There is an issue with the tax computation.

Added module sale_create_invoices_hook to fix it.

sale_create_invoices_hook/hooks.py
```diff
@@ -0,0 +1,250 @@
+# Copyright 2025 ForgeFlow, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```
alexey-pelykh, line 2: License header says LGPL-3.0 but `__manifest__.py` declares AGPL-3. These must match.

sale_create_invoices_hook/hooks.py
```diff
@@ -0,0 +1,250 @@
...
+                    "self": move,
+                    "origin": move.line_ids.sale_line_ids.order_id,
+                },
+                subtype_xmlid="mail.mt_note",
+            )
+        return moves
+
+    if not hasattr(SaleOrder, "_create_invoices_original"):
+        SaleOrder._create_invoices_original = _create_invoices_new
+    SaleOrder._create_invoices = _create_invoices_new
```
alexey-pelykh, line 250: This guard stores the *new* patched function as `_create_invoices_original`, which is misleading. If the intent is to keep a reference to the original for restoration, store `SaleOrder._create_invoices` *before* overwriting it. Also, `_create_invoices_original` is never referenced anywhere -- consider removing the dead code.

sale_down_payment_deduct_partial/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,131 @@
...
+                    line = dp_lines[0]
+                    taxes_res = line.tax_id.compute_all(
+                        1.0,
+                        currency=line.order_id.currency_id,
+                        quantity=1.0,
+                        product=line.product_id,
+                        partner=line.order_id.partner_shipping_id,
+                    )
+                    tax_ratio = (
+                        taxes_res["total_excluded"] / taxes_res["total_included"]
```
alexey-pelykh, line 119: This assumes all down payment lines share the same tax rate (takes `dp_lines[0]` and extracts ratio). If DP lines have different taxes, the tax-excluded conversion will be wrong. Consider computing per-line or validating that all DP lines share the same tax.

sale_down_payment_deduct_partial/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,131 @@
...
+                            "available down payment deduction."
+                        )
+                    )
+
+    @api.depends("sale_order_ids")
+    def _compute_total_deduction_amount(self):
+        # Separated into two methods to avoid recomputing when the other one is changed
+        # Although it's the same logic, one field is readonly and the other is editable
+        for wizard in self:
+            dp_lines = wizard.sale_order_ids.mapped("order_line").filtered(
```
alexey-pelykh, line 53: The down payment remaining calculation is duplicated across `_compute_total_deduction_amount`, `_compute_deduction_amount`, and `_create_invoices`. Extract into a shared helper to keep it DRY and avoid bugs from divergent copies.

sale_down_payment_deduct_partial/tests/test_sale_down_payment_deduct_partial.py
```diff
@@ -0,0 +1,225 @@
...
+        self.assertEqual(wizard.deduction_amount, 50)
+        inv = wizard._create_invoices(self.sale_order)
+        inv.action_post()
+        self.assertEqual(inv.move_type, "out_invoice")
+        self.assertEqual(inv.amount_untaxed, 1000 - 50)
+        dp_lines = self.sale_order.order_line.filtered(
+            lambda lin: lin.is_downpayment and not lin.display_type
+        )
+        dp_lines[0].qty_to_invoice = -0.5
+        dp_lines[1].qty_to_invoice = -1.0
```
alexey-pelykh, line 163: These are *assignments* (`dp_lines[0].qty_to_invoice = -0.5`), not assertions. They mutate state rather than verifying it. Should be `self.assertAlmostEqual(dp_lines[0].qty_to_invoice, -0.5)` or `assertEqual`. Same issue at lines 184-185 and 205-206.

alexey-pelykh (changes requested):
Thanks for this contribution. The idea of partial down payment deduction is useful, and the overall structure is sound. However there are several issues that should be addressed.

## Major concerns

### 1. `sale_create_invoices_hook` uses `post_load` to monkeypatch `_create_invoices`

This is a very invasive approach. The entire `_create_invoices` method (~250 lines) is copy-pasted from Odoo core and patched with a single hook call (`self.down_payment_final_rounding(final)`). This creates a massive maintenance burden:
- Any upstream bugfix or improvement to `_create_invoices` in Odoo 18.0 point releases will be silently lost.
- The flake8 noqa directives (C901, E501, E741) confirm the method was copied verbatim without cleanup.

Have you considered a lighter alternative? For instance, overriding `_create_account_invoices` or post-processing the move lines after `super()._create_invoices()` to reverse the rounding correction when context flag is set? That would avoid the full method copy.

### 2. License mismatch in `sale_create_invoices_hook`

`__manifest__.py` declares `AGPL-3`, but `hooks.py` header says `LGPL-3.0`. These must be consistent.

### 3. Duplicated computation logic

The down payment remaining amount calculation (iterating `dp_lines`, calling `tax_id.compute_all`) is repeated in three places:
- `_compute_total_deduction_amount`
- `_compute_deduction_amount`
- Inside `_create_invoices` override in the wizard

Extract this into a shared helper method on `sale.order` or on the wizard.

### 4. Tax ratio assumption in wizard `_create_invoices`

The tax-included-to-excluded conversion uses the tax rate from `dp_lines[0]` and assumes all down payment lines share the same tax rate. This is fragile -- if different DP lines have different taxes, the deduction will be incorrect. The per-line approach in `_adjust_downpayment_lines` is correct; the conversion before calling it should also be per-line.

## Minor concerns

### 5. Test assertions use assignment instead of assertion

In `test_partial_down_payment_deduction`, lines like:
```python
dp_lines[0].qty_to_invoice = -0.5
dp_lines[1].qty_to_invoice = -1.0
```
These are *assignments*, not assertions. They should be `assertEqual` calls to actually verify the result. As written, they always pass and verify nothing.

### 6. Missing `ir.model.access.csv` consideration

Both modules add new models/inherit existing ones. The wizard inherits `sale.advance.payment.inv` which should already have access rules, but verify no new model requires explicit security rules.

### 7. `_compute_total_deduction_amount` uses `total_included` but `_adjust_downpayment_lines` uses `total_excluded`

The wizard shows `total_deduction_amount` as tax-included, then converts to tax-excluded before calling `_adjust_downpayment_lines`. This is internally consistent but could confuse users if taxes are complex. Consider documenting which amounts are tax-included vs excluded in field help strings.

### 8. `_create_invoices_original` guard is misleading

In `hooks.py` line 371-372:
```python
if not hasattr(SaleOrder, "_create_invoices_original"):
    SaleOrder._create_invoices_original = _create_invoices_new
```
This stores the *new* function as `_create_invoices_original`, which is counterintuitive. If the intent is to keep a reference to the original method for potential restoration, store the *original* before overwriting. As written, `_create_invoices_original` is never used anyway.

rousseldenis:
@LauraCForgeFlow Could you take care of this ? It seems there are mixed modules here.

LauraCForgeFlow:
Hi everyone @rousseldenis @alexey-pelykh,

Since we are not currently using this module in our projects (it was an initial proof of concept), and given its complexity, we unfortunately don't have the availability right now to make the necessary adjustments to finish this. 

I'm going to close this PR for now. However, if anyone from the OCA wants to pick it up from where we left off, you are welcome to do so! If we find the time to resume this in the future, we will reopen this PR and continue working from here. 

Thank you, and apologies for not being able to finish this at the moment.

PR #3983, opened 2025-11-03
[12.0] FIX sale_fixed_discount : Wrong tax amount displayed on report

eLBati:
from https://github.com/OCA/sale-workflow/pull/1443

### DIFF
--- a/sale_fixed_discount/models/sale_order.py
+++ b/sale_fixed_discount/models/sale_order.py
@@ -4,6 +4,56 @@
 from odoo import api, fields, models, _
 import odoo.addons.decimal_precision as dp
 from odoo.exceptions import ValidationError
+from functools import partial
+from odoo.tools.misc import formatLang
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends('order_line.price_total')
+    def _amount_by_group(self):
+        """Override to consider fixed discount in tax calculations."""
+        super(SaleOrder, self)._amount_by_group()
+
+        for order in self:
+            # Check if any line has a fixed discount
+            if not any(line.discount_fixed for line in order.order_line):
+                continue
+
+            # Recompute considering fixed discounts
+            currency = order.currency_id or order.company_id.currency_id
+            fmt = partial(formatLang, self.with_context(lang=order.partner_id.lang).env, currency_obj=currency)
+            res = {}
+
+            for line in order.order_line:
+                # Calculate price_reduce considering both percentage and fixed discounts
+                price_reduce = line.price_unit * (1.0 - line.discount / 100.0)
+                # Subtract fixed discount if present
+                if line.discount_fixed:
+                    price_reduce -= line.discount_fixed
+
+                taxes = line.tax_id.compute_all(
+                    price_reduce,
+                    quantity=line.product_uom_qty,
+                    product=line.product_id,
+                    partner=order.partner_shipping_id
+                )['taxes']
+
+                for tax in line.tax_id:
+                    group = tax.tax_group_id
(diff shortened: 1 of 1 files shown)


sale_fixed_discount/models/sale_order.py
```diff
@@ -4,6 +4,56 @@
 from odoo import api, fields, models, _
 import odoo.addons.decimal_precision as dp
 from odoo.exceptions import ValidationError
+from functools import partial
```
alexey-pelykh, line 7: Import ordering: `from functools import partial` is a stdlib import and should come before the `odoo` imports (PEP8, isort). This is likely causing the pre-commit CI failure.

Suggested order:
```python
from functools import partial

from odoo import api, fields, models, _
import odoo.addons.decimal_precision as dp
from odoo.tools.misc import formatLang
```

sale_fixed_discount/models/sale_order.py
```diff
@@ -4,6 +4,56 @@
...
+from odoo.tools.misc import formatLang
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends('order_line.price_total')
+    def _amount_by_group(self):
+        """Override to consider fixed discount in tax calculations."""
+        super(SaleOrder, self)._amount_by_group()
```
alexey-pelykh, line 17: The `super()` call here computes `amount_by_group` for all orders, but then the loop below fully recomputes and overwrites it for orders that have fixed discount lines. The super computation is wasted for those orders.

Two options:
1. Remove the `super()` call and handle all orders here (but then orders without fixed discounts need the standard computation -- you could call super only for those).
2. Keep it as-is for simplicity/safety but add a comment explaining the intentional recomputation.

alexey-pelykh (changes requested):
## Review: [12.0] FIX sale_fixed_discount : Wrong tax amount displayed on report

Thanks for working on this fix. The approach of overriding `_amount_by_group` to account for the `discount_fixed` field in tax group calculations is the right direction. However, there are several issues that should be addressed before this can be merged.

### Issues Found

| # | Severity | Finding |
|---|----------|----------|
| 1 | Moderate | Pre-commit CI is failing (likely import ordering) |
| 2 | Moderate | No version bump in `__manifest__.py` (`12.0.1.0.0` -> `12.0.1.0.1` expected for a bugfix) |
| 3 | Moderate | No tests added to cover this fix |
| 4 | Minor | Unnecessary `super()` call -- the method calls `super()._amount_by_group()` but then fully recomputes and overwrites `order.amount_by_group` for any order with fixed discount lines, making the super call wasted computation |
| 5 | Minor | Import ordering: `from functools import partial` (stdlib) should come before `from odoo import ...` (third-party/framework) per PEP8 and isort conventions |

### Suggestions

1. **Fix import order**: Move `from functools import partial` above the odoo imports to satisfy pre-commit/isort.
2. **Bump version** to `12.0.1.0.1` in `__manifest__.py`.
3. **Add a test case** that creates a sale order with a fixed discount and verifies that `amount_by_group` reflects the correct tax amounts.
4. **Consider skipping super()** for orders with fixed discount lines, or restructuring to avoid the redundant computation. Alternatively, keep the super call for clarity but add a comment explaining why.

The core logic of computing `price_reduce` with the fixed discount subtracted and passing it to `compute_all` is correct and consistent with how the module handles fixed discounts elsewhere.

rousseldenis:
@eLBati What's the status of this ? Could you add a test case ?

rousseldenis:
@eLBati Do we keep this ? This seems still valid as there is : https://github.com/OCA/sale-workflow/pull/4510

PR #3984, opened 2025-11-03
[18.0][ADD] sale_order_lot_selection_price: New module

victoralmau:
New module

Please @christian-ramos-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT57112


sale_order_lot_selection_price/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_product_price_context(self):
+        ctx = super()._get_product_price_context()
+        ctx.update(from_sol=True, sol_lot=self.lot_id)
```
pedrobaeza: This is assuming that you always select lot, so I think it's not valid.

victoralmau: Lot (record) is actually being sent by context, which will arrive at https://github.com/Tecnativa/sale-workflow/blob/efe6a86aaab10cac696a16779f3623545e9ed080/sale_order_lot_selection_price/models/product_product.py#L12 and will only enter the condition if there is a defined lot. This use case (without lot) is tested at https://github.com/Tecnativa/sale-workflow/blob/efe6a86aaab10cac696a16779f3623545e9ed080/sale_order_lot_selection_price/tests/test_sale_order_lot_selection_price.py#L95

In any case, if you prefer, i can add a condition here so that it only adds things to the context if there is a defined lot.

pedrobaeza: OK, understood. And why 2 contexts? Isn't enough with `sol_lot`?

pedrobaeza (approved):

christian-ramos-tecnativa (approved):
LGTM

PR #3985, opened 2025-11-03
[19.0][MIG] sale_order_lot_selection: Migration to 19.0

lef-adhoc:
Depends on:

- [ ] <https://github.com/OCA/stock-logistics-workflow/pull/2151>


manuelcalerosolis (commented):
Functional review ok!

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4011 to commit history before migration commit.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,41 @@
...
+    def _prepare_procurement_values(self):
+        vals = super()._prepare_procurement_values()
+        if self.lot_id:
+            vals["restrict_lot_id"] = self.lot_id.id
+        return vals
+
+    @api.depends("product_id")
+    def _compute_product_tracking(self):
+        for sol in self:
+            sol.product_tracking = sol.product_id.tracking or sol.product_tracking
```
alexey-pelykh, line 37: Nit: the `or sol.product_tracking` fallback is only reached when `product_id` is empty (since `'none'` is truthy). In that case the view already hides the lot field via `invisible="not product_id ..."`, so the preserved stale value has no visible effect. Could simplify to just:
```python
sol.product_tracking = sol.product_id.tracking
```
Not blocking, just a readability point.

sale_order_lot_selection/tests/test_sale_order_lot_selection.py
```diff
@@ -0,0 +1,354 @@
...
+        for move in picking_in.move_ids:
+            self.assertEqual(move.state, "assigned", "Wrong state of move line.")
+        lot10 = False
+        lot11 = False
+        lot12 = False
+        for move in picking_in.move_ids:
+            if move.product_id == self.prd_cable:
+                lot10 = self.lot_model.create(
+                    {
+                        "name": "0000010",
```
alexey-pelykh, line 209: Minor: `"type": "consu"` is redundant in 19.0 -- products created with `is_storable=True` already have `type='consu'`. This write effectively only sets `tracking`. Consider removing the `type` key to avoid confusion during future migrations.

sale_order_lot_selection/views/sale_order_views.xml
```diff
@@ -0,0 +1,34 @@
...
+                position="after"
+            >
+                <field
+                    name="lot_id"
+                    domain="[('product_id','=', product_id)]"
+                    invisible="not product_id or product_tracking=='none'"
+                    context="{'default_product_id': product_id}"
+                    groups="stock.group_production_lot"
+                    optional="show"
+                />
```
alexey-pelykh, line 19: Consider adding `<field name="product_tracking" column_invisible="True"/>` explicitly in the list view. While Odoo 17+ auto-fetches fields used in `invisible` expressions, declaring it makes the dependency clear and avoids surprises if that auto-fetch behavior ever changes.

alexey-pelykh (commented):
Solid migration overall. The core lot selection logic, procurement integration, and view xpaths are correct for 19.0. A few observations:

**Functional scope change** -- The 16.0 version had `allow_to_change_lot_on_confirmed_so` (company setting + config view + `write()` override + `lot_id_readonly` computed field). All of this is dropped in the migration. If this is intentional simplification, it would be worth mentioning in the PR description or a migration note, since users upgrading from 16.0 will lose that capability.

**Minor code observations:**

1. `_compute_product_tracking` -- The fallback `sol.product_id.tracking or sol.product_tracking` works correctly for the visibility use-case but could be simplified to just `sol.product_id.tracking` since the compute fires on `product_id` change and the only consumer is the `invisible` expression. The `or sol.product_tracking` branch only preserves a stale value when `product_id` is cleared, which has no practical benefit since the field is hidden when `not product_id` anyway.

2. In the test `test_02_sale_order_lot_selection`, `self.product_46.write({"tracking": "lot", "type": "consu"})` -- the `"type": "consu"` is a no-op in 19.0 since these products are already `consu` (created with `is_storable=True`). Not a bug, but dead code carried forward from older versions.

3. The `product_tracking` field is used in the view `invisible` expression but not declared in the view XML. Odoo 17+ auto-fetches fields in expressions, so this works, but adding `<field name="product_tracking" column_invisible="True"/>` would make the dependency explicit.

CI: "Detect unreleased dependencies" failure is expected due to the pending `stock_restrict_lot` migration (OCA/stock-logistics-workflow#2151).

Tests pass on both Odoo and OCB.

odooNextev:
@lef-adhoc I would like to try it but the runboat doesn't work, could you please rebase?

odooNextev (approved):
tested locally, it works. lgtm

victoralmau (changes requested):
My comment https://github.com/OCA/sale-workflow/pull/3985#issuecomment-3811942772 is still pending.

lef-adhoc:
@victoralmau Done!

victoralmau (approved):

BhaveshHeliconia (approved):
Functional review LGTM!

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4338 to commit history before migration commit.

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4336 to commit history before migration commit.

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4412 to commit history before migration commit.

lef-adhoc:
@victoralmau I'm going to close the PR

PR #3986, opened 2025-11-04
[18.0] sale_order_carrier_auto_assign: refactor to improve consistency

simahawk:
When playing onchanges programmatically, we might have _origin as NewId instance.
Prevent having an empty default_order_id in the ctx when this happens.

Refactor carrier computation
    
* use computed field instead of onchange
* simplify logic

This change introduce consistency because no matter how you set the
carrier, we'll have the same logic.


sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -1,25 +1,38 @@
...
-            self._set_delivery_carrier(
-                set_delivery_line=False,
-                preserve_order_carrier=False,
-            )
+    carrier_id = fields.Many2one(
+        compute="_compute_carrier_id", store=True, readonly=False
+    )
+
+    @api.depends("partner_id", "partner_shipping_id", "order_line.product_id", "state")
+    @api.depends_context("override_carrier", "carrier_from_create")
```
rousseldenis: `depends_context` should not be used with stored fields

simahawk: ~any particular reason?~ that makes sense in general but here I want to ensure it gets recomputed if I inject the ctx key.. I'll do some tests as soon as I can.

mmequignon (approved):
Only code review, LG

simahawk:
FTR I had to discard the computed field approach because in too many cases (out of this module scope) the carrier was computed when not needed/expected. However, the carrier is not a visible field anymore on the SO form so there's no need for a computation nor an onchange IMO.

simahawk:
need to check tests

TDu (approved):

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,39 @@
...
         )
 
+    def _auto_set_carrier_on_create(self):
+        for rec in self:
+            if rec._is_auto_set_carrier_on_create():
+                rec._set_delivery_carrier()
+
     @api.model_create_multi
     def create(self, vals_list):
-        ctx_carrier_on_create = self.env.context.get("carrier_on_create")
```
jbaudoux, line 31: You need to keep those context variables to prevent the carrier to be set when first line is created. It must be computed at order level once all lines are created. There are modules adding more lines in cascade

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,39 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
-    @api.onchange("partner_id", "partner_shipping_id")
```
jbaudoux, line 10: By what is it replaced for the salesman editing the SO?
I would keep this

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,39 @@
...
-        self._set_carrier_on_create()
+        res = super().write(vals)
+        for rec in self:
+            if rec._is_auto_set_carrier_on_write(vals):
+                rec._set_delivery_carrier(preserve_order_carrier=False)
         return res
 
+    def _is_auto_set_carrier_on_write(self, vals):
+        return self.state in ("draft", "sent") and bool(
+            not vals.get("carrier_id")
```
jbaudoux, line 40: Writing lines may set a wrong carrier if the new partner is not yet written and it won't be updated here
Keep the context but improve the call to self._set_carrier_on_create() to add your preserve flag

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -71,9 +60,13 @@ def _set_delivery_carrier(
...
         for order in self:
-            if not order.order_line:
+            if (
+                not order.order_line
+                or isinstance(order.id, models.NewId)
+                and not order._origin
+            ):
                 continue
-            if order.delivery_set:
+            if order.delivery_set and preserve_order_carrier:
```
jbaudoux, line 71: In case of preserve, you still need to update the cost. This is used by the delivery_refresh module

jbaudoux (changes requested):
Too much breaking changes. Your refactor recreates the issues that were solved for the allo project. Can you please limit your changes to only the issue you faced?

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: no, is not needed. The field is not visible in the form anymore. In any case, this is the source of the original problem.

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: the ctx is there now

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: the ctx key is back

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: I understood you had issue when the order was not created by the form. So that onchange was not triggered. But when it's done by the form, can't we keep this onchange? I don't get why it's an issue

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: no, it's not on the call to create

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: no it's not on the call to write

jbaudoux (changes requested):

alexey-pelykh (commented):
### Summary

The PR refactors the carrier auto-assignment logic, replaces the `@api.onchange` on partner changes with a `write`-level hook, and improves `_set_delivery_carrier` to handle `NewId` edge cases. Test coverage is meaningfully expanded. CI is green.

However, the module maintainer (@jbaudoux) has open `CHANGES_REQUESTED` reviews that appear unaddressed in the current head. The core concerns -- removal of the onchange, removal of the `carrier_on_create` context guard during super calls, and the `delivery_set` / `preserve_order_carrier` logic change -- remain in the diff. Those are the blocking items.

Below are additional technical observations on top of the existing review thread.

---

### Key findings

**1. Removal of `carrier_on_create` context guard in `create` / `write` super calls**

The base branch wraps `super().create(vals_list)` and `super().write(vals)` with `self.with_context(carrier_on_create=True)` to prevent cascading carrier recomputation when downstream modules (e.g. modules adding extra lines in `create`) trigger their own writes. This PR removes that guard entirely in both `SaleOrder` and `SaleOrderLine`. This is the concern @jbaudoux raised and it has not been addressed in the latest commits.

**2. `_is_auto_set_carrier_on_write` may fire prematurely**

The new `write` override calls `_set_delivery_carrier(preserve_order_carrier=False)` when `partner_id` or `partner_shipping_id` is in `vals`. But `super().write(vals)` writes all vals atomically, so if a caller writes `{"partner_shipping_id": X, "carrier_id": Y}` in one call, the carrier is first set by `super().write` to Y, then immediately overwritten by `_set_delivery_carrier`. The `vals.get("carrier_id")` guard only prevents firing when carrier_id is explicitly in vals, but it does not prevent overwriting a carrier that was just set in the same write.

**3. Operator precedence in `_set_delivery_carrier` guard**

The condition:
```python
if (
    not order.order_line
    or isinstance(order.id, models.NewId)
    and not order._origin
):
```
is correct by Python precedence (`and` binds tighter than `or`), but the formatting across lines makes it look like a flat `or`/`and` chain. Adding explicit parentheses around the `and` clause would prevent future misreads.

**4. `delivery_set` logic change**

The base code skips carrier assignment when `order.delivery_set` is truthy (unconditionally). The PR changes this to `if order.delivery_set and preserve_order_carrier`. This means when `preserve_order_carrier=False` (the new write path), an order with an existing delivery line will now go through the wizard again. This could generate duplicate delivery lines or unwanted rate recalculations depending on how `action_open_delivery_wizard` behaves. This is the concern raised by @jbaudoux about the `delivery_refresh` module interaction.

**5. `delivery_wiz.order_id` assignment pattern**

The new code does:
```python
delivery_wiz = delivery_wiz_model.new({})
delivery_wiz.order_id = order
if not delivery_wiz.order_id and order._origin:
    delivery_wiz.order_id = order._origin
```
The fallback `if not delivery_wiz.order_id` relies on the assignment being silently dropped for unsaved records, which is fragile. The original code handled this through `default_order_id` in the context. Consider keeping the context-based approach and only patching the NewId edge case.

**6. `ensure_one()` removed from `_is_auto_set_carrier_on_create` and `_is_auto_set_carrier_on_confirm`**

These methods access `self.state`, `self.company_id`, etc. without `ensure_one()`. While the callers iterate with `for rec in self`, removing the guard means a future caller passing a multi-record recordset would get silently wrong results (accessing only the first record's fields) instead of an explicit error.

### Verdict

The maintainer's open change requests are the primary blocker. The technical points above are secondary but should also be addressed. I would not merge until the discussion with @jbaudoux is resolved.

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,41 @@
...
-        if self.company_id.carrier_on_create:
-            self._set_delivery_carrier(
-                set_delivery_line=False,
-                preserve_order_carrier=False,
-            )
-
     def _is_auto_set_carrier_on_create(self):
-        self.ensure_one()
         return (
-            self.state in ("draft", "sent")
```
alexey-pelykh, line 24: The `carrier_on_create` context guard that previously wrapped `super().create()` has been removed. The base branch sets `carrier_on_create=True` during super to prevent cascading modules from triggering carrier recomputation for each line added during create. Without this, modules that add extra lines in `create` overrides (e.g. promotions, bundles) will trigger `_auto_set_carrier_on_create` prematurely for each intermediate state.

This is the concern @jbaudoux raised -- can you either restore the guard or explain why it is no longer needed with the new flow?

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,41 @@
...
+            not self.carrier_id
+            and self.state in ("draft", "sent")
             and self.company_id.carrier_on_create
             and not self.is_all_service
```
alexey-pelykh, line 15: The `write` override no longer wraps `super().write(vals)` with `carrier_on_create=True` context. In the base branch, this prevents the carrier from being recomputed during intermediate writes (e.g. when order lines are updated as side effect of the write).

Also, consider the case where `vals` contains both `partner_shipping_id` and `carrier_id` -- `_is_auto_set_carrier_on_write` checks `not vals.get("carrier_id")`, but `_set_delivery_carrier(preserve_order_carrier=False)` will override whatever carrier was just written by `super().write(vals)`. The explicit carrier in vals should take precedence.

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,41 @@
...
-        ctx_carrier_on_create = self.env.context.get("carrier_on_create")
-        orders = super(SaleOrder, self.with_context(carrier_on_create=True)).create(
-            vals_list
-        )
-        orders = orders.with_context(carrier_on_create=ctx_carrier_on_create)
-        orders._set_carrier_on_create()
-        return orders
-
-    def _set_carrier_on_create(self):
+    def _auto_set_carrier_on_create(self):
```
alexey-pelykh, line 18: Parentheses would clarify intent:
```python
if (
    not order.order_line
    or (isinstance(order.id, models.NewId) and not order._origin)
):
```
Python precedence makes `and` bind tighter than `or`, so this is functionally correct, but the multi-line formatting makes it look like a flat chain.

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,41 @@
...
         if self.env.context.get("carrier_on_create"):
             return
-        for order in self:
-            if not order.carrier_id and order._is_auto_set_carrier_on_create():
```
alexey-pelykh, line 43: Previously the code skipped unconditionally when `delivery_set` was truthy. Now it only skips when `preserve_order_carrier` is also true. This means the write-triggered path (`preserve_order_carrier=False`) will re-enter the wizard even when a delivery line already exists. Depending on `action_open_delivery_wizard` behavior, this could create duplicate delivery lines or trigger unwanted rate recalculations.

This interacts with `delivery_refresh` as @jbaudoux noted -- when preserving the carrier, the delivery cost still needs updating.

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,41 @@
...
-                order.with_context(carrier_on_create=True)._set_delivery_carrier()
+        for rec in self:
+            if rec._is_auto_set_carrier_on_create():
+                rec.with_context(carrier_on_create=True)._set_delivery_carrier()
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
alexey-pelykh, line 26: This pattern is fragile. Assigning `order` (potentially a `NewId` record) to `delivery_wiz.order_id` may silently fail for unsaved records, and the fallback `if not delivery_wiz.order_id` relies on that silent failure.

The original approach of patching `default_order_id` in the wizard context was more explicit. Consider keeping that pattern and only adding a guard for the `NewId` edge case that motivated this PR.

sale_order_carrier_auto_assign/models/sale_order_line.py
```diff
@@ -8,23 +9,16 @@ class SaleOrderLine(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        # When product lines are added, set the carrier
-        ctx_carrier_on_create = self.env.context.get("carrier_on_create")
-        order_lines = super(
```
alexey-pelykh, line 13: The `carrier_on_create` context guard has been removed from the `SaleOrderLine.create` super call as well. Same concern as in `SaleOrder.create` -- without this guard, any module that creates additional order lines during `super().create()` will trigger carrier recomputation for each intermediate line.

rousseldenis:
@simahawk What's the status of this ?

simahawk:
I need to get back to this but I don't have time :sob: 
I'll try to find some time in April.

simahawk:
Finally found the time to get back on this.
@Shide as you've worked on https://github.com/OCA/sale-workflow/pull/4129 could you help w/ a review here?
Also a manual test on `delivery_auto_refresh` would be nice if possible.
As of now I've only ensured that tests pass there and I opened a PR to test on CI https://github.com/OCA/delivery-carrier/pull/1201

simahawk:
I've found a conflict with `delivery_multi_destination` that comes from the fact that `self.company_id.carrier_auto_assign` is not checked (and never was) on writes.
To me this flag should turn on/off completely the features provided by this module.
I added that to `_is_auto_set_carrier_on_write` condition.

https://github.com/OCA/sale-workflow/pull/3986/commits/c3e64b0cf03690d1b8cdf24a41a571d59a9d1039

CC @jbaudoux

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,45 @@
...
+        recs._auto_set_carrier_on_create()
+        return recs
 
     def write(self, vals):
-        # When product lines are added, set the carrier
-        res = super(SaleOrder, self.with_context(carrier_on_create=True)).write(vals)
-        self._set_carrier_on_create()
+        res = super().write(vals)
+        for rec in self:
+            if rec._is_auto_set_carrier_on_write(vals):
```
jbaudoux, line 34: You are after super, no reason to look at vals anymore.

You should check here is_carrier_auto_set_on_create. The purpose is to set a carrier if enabled on create in case all conditions were not met at create.

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,45 @@
...
+                rec._set_delivery_carrier(preserve_order_carrier=False)
         return res
 
+    def _is_auto_set_carrier_on_write(self, vals):
+        return (
+            self.company_id.carrier_auto_assign
+            and self.state in ("draft", "sent")
+            and bool(
```
jbaudoux, line 42: Those conditions should be part of the check ..._on_create

jbaudoux (changes requested):
This module has 2 purposes.
Set the carrier as soon as all conditions are met (on create)
Set a carrier at SO confirmation

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,45 @@
...
+    def create(self, vals_list):
+        recs = super().create(vals_list)
+        recs._auto_set_carrier_on_create()
+        return recs
 
     def write(self, vals):
-        # When product lines are added, set the carrier
-        res = super(SaleOrder, self.with_context(carrier_on_create=True)).write(vals)
-        self._set_carrier_on_create()
+        res = super().write(vals)
```
jbaudoux, line 32: The context was there to prevent to reenter this inside the write. You need to keep it

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -7,50 +7,45 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        recs = super().create(vals_list)
+        recs._auto_set_carrier_on_create()
+        return recs
 
     def write(self, vals):
-        # When product lines are added, set the carrier
-        res = super(SaleOrder, self.with_context(carrier_on_create=True)).write(vals)
-        self._set_carrier_on_create()
```
jbaudoux, line 49: I don't get why you change this. Purpose is to set the carrier on write if carrier is enabled on create and conditions were not met

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: I agree because it's not part of the view

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: because you are calling it w/ `carrien_on_create` by default even if you are not in a create call

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: The original write must happen anyway, afterwards you set the carrier if - still - needed.
The case on create is already managed by the create. This is write. I don't get why you are mixing them 🤔

sale_order_carrier_auto_assign/models/sale_order.py
simahawk: why??? There were no conditions on create before and on create you don't care to check if you are changing something...

jbaudoux: Ah you want to detect the partner is modified... For the rest, the conditions are the applicability and the same as on create

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: Read it "carrier on setup" that applies as soon as you have a partner and a non service line. It's done on create and write and with the same conditions

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: You need to manage the original "carrier on setup" that applies as soon as you have a partner and a non service line. It's done on create and write and with the same conditions.
And your new feature that updates the carrier when the partner is changed.

rousseldenis:
@simahawk

PR #3987, opened 2025-11-04
[18.0][IMP] sale_order_lot_selection: Be able to select only available lots

victoralmau:
Be able to select only available lots

Example use case:
- Create 5 lots (Lot 1, Lot 2, Lot 3, Lot 4, Lot 5) for a product
- Define stock only in lots: Lot 1, Lot 2, and Lot 5
- Only lots with stock should be selectable in the sales order line.

Please @pedrobaeza and @christian-ramos-tecnativa can you review it?

@Tecnativa TT57112

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -1,4 +1,5 @@
 from odoo import api, fields, models
+from odoo.tools import float_compare
 
 
 class SaleOrderLine(models.Model):
@@ -13,9 +14,11 @@ class SaleOrderLine(models.Model):
         selection=_selection_product_tracking,
         compute="_compute_product_tracking",
     )
+    domain_lot_id = fields.Binary(compute="_compute_domain_lot_id")
     lot_id = fields.Many2one(
         "stock.lot",
         "Lot",
+        domain="domain_lot_id",
         copy=False,
         compute="_compute_lot_id",
         store=True,
@@ -33,6 +36,35 @@ class SaleOrderLine(models.Model):
     def _compute_product_tracking(self):
         for sol in self:
             sol.product_tracking = sol.product_id.tracking or sol.product_tracking
+
+    @api.depends("product_id", "product_uom_qty", "warehouse_id")
+    def _compute_domain_lot_id(self):
+        dp = self.env["decimal.precision"].precision_get("Product Unit of Measure")
+        for sol in self:
+            domain = []
+            if sol.product_id and sol.product_tracking != "none":
+                # Only available lots should be displayed. In pickings, the
+                # corresponding stock.quant record is selected directly, so we
+                # use the same filters that are used.
+                quants = self.env["stock.quant"].search(
+                    [
+                        ("product_id", "=", sol.product_id.id),
+                        ("quantity", ">=", sol.product_uom_qty),
+                        ("lot_id", "!=", False),
+                        ("location_id.usage", "=", "internal"),
+                        ("location_id.warehouse_id", "=", sol.warehouse_id.id),
+                    ]
+                )
+                available_quants = quants.filtered(
+                    lambda x, qty=sol.product_uom_qty: float_compare(
+                        x.available_quantity,
+                        qty,
+                        precision_digits=dp,
+                    )
+                    >= 0
+                )
+                domain = [("id", "in", available_quants.mapped("lot_id").ids)]
+            sol.domain_lot_id = domain
 
     @api.depends("product_id")
     def _compute_lot_id(self):
--- a/sale_order_lot_selection/views/sale_order_views.xml
+++ b/sale_order_lot_selection/views/sale_order_views.xml
@@ -10,7 +10,6 @@
             >
                 <field
                     name="lot_id"
-                    domain="[('product_id','=', product_id)]"
                     invisible="not product_id or product_tracking=='none'"
                     context="{'default_product_id': product_id}"
                     groups="stock.group_production_lot"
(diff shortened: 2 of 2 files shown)


sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -34,6 +36,24 @@ def _compute_product_tracking(self):
...
 
+    @api.depends("product_id")
+    def _compute_domain_lot_id(self):
+        for sol in self:
+            domain = []
+            if sol.product_id:
+                # Only available lots should be displayed. In pickings, the
+                # corresponding stock.quant record is selected directly, so we
+                # use the same filters that are used.
+                quants = self.env["stock.quant"].search(
```
rousseldenis, line 49: @victoralmau Isn't there a mean to avoid the searches and to compute a domain instead ?

rousseldenis: IMHO, you can use in domain `quant_ids.quantity` or `product_qty` that has a search method.

pedrobaeza: @victoralmau if this is code from stock module, can you link to it?

victoralmau: The appropriate filter is to use the `product_qty` field for lots.

In `stock`, lots are not specifically selected; quant is selected, which is why the same cannot be used.

pedrobaeza: What do you mean with "batches"?

victoralmau: Sorry, lots

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -34,6 +36,17 @@ def _compute_product_tracking(self):
...
             sol.product_tracking = sol.product_id.tracking or sol.product_tracking
 
+    @api.depends("product_id")
+    def _compute_domain_lot_id(self):
+        for sol in self:
+            domain = []
+            if sol.product_id:
+                domain = [
+                    ("product_id", "=", sol.product_id.id),
+                    ("product_qty", ">", 0),
```
pedrobaeza: IMO, it should be only product available, as you may have it reserved for other pickings.

victoralmau: That is not possible using only lot fields; it would be necessary to filter quants first and then filter the corresponding lots. In any case, I think the domain added by this PR is better than nothing, because currently it is possible to select any lot.

pedrobaeza: Yeah, but being a computed, you can search for quants, and then do `.mapped("lot_id")` and return a domain `[("id", "in", ...)]` one.

victoralmau: Okay, changed, that was my initial approach.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -34,6 +37,39 @@ def _compute_product_tracking(self):
...
+            domain = []
+            if sol.product_id:
+                # Only available lots should be displayed. In pickings, the
+                # corresponding stock.quant record is selected directly, so we
+                # use the same filters that are used.
+                quants_to_delete = self.env["stock.quant"]
+                quants = self.env["stock.quant"].search(
+                    [
+                        ("product_id", "=", sol.product_id.id),
+                        ("quantity", ">=", sol.product_uom_qty),
```
pedrobaeza, line 52: It should be `available_quantity` (or the corresponding field).

victoralmau: The `available_quantity` field cannot be used because it is compute, not store.: https://github.com/odoo/odoo/blob/0bf53e6b2ab8c4c1705ebbb776102d493ac7445c/addons/stock/models/stock_quant.py#L88

pedrobaeza: Then you should do a `filtered` over this result to get the proper results.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -34,6 +37,35 @@ def _compute_product_tracking(self):
         for sol in self:
             sol.product_tracking = sol.product_id.tracking or sol.product_tracking
 
+    @api.depends("product_id", "product_uom_qty", "warehouse_id")
+    def _compute_domain_lot_id(self):
+        dp = self.env["decimal.precision"].precision_get("Product Unit of Measure")
+        for sol in self:
+            domain = []
+            if sol.product_id:
```
pedrobaeza: I don't remember the exact name/values of the field, but for the idea:

```suggestion
            if sol.product_id and sol.product_id.tracking in ("lot", "serial"):
```

victoralmau: Condition added, although the `lot_id` field is invisible in this case: https://github.com/OCA/sale-workflow/blob/294b53f3c1a60e50c1050fac68c4de1790ac5a7b/sale_order_lot_selection/views/sale_order_views.xml#L14

pedrobaeza: Yeah, but I would say that the computed is triggered... we can check it just for future references.

pedrobaeza (approved):

sale_order_lot_selection/models/sale_order_line.py
carlos-lopez-tecnativa: The lot has a `search` function. I think the solution could be to simply add a domain `("product_qty", ">", 0)`. Have you tried this?
https://github.com/odoo/odoo/blob/2a6f08bf26e34222a788396a2e4b6ea815f4104f/addons/stock/models/stock_lot.py#L214

Another option is to use _read_group, which is more efficient, similar to:
https://github.com/odoo/odoo/blob/2a6f08bf26e34222a788396a2e4b6ea815f4104f/addons/stock/models/stock_lot.py#L224C46-L224C57

pilarvargas-tecnativa (approved):

PR #3990, opened 2025-11-05
[17.0][FIX]sale_product_multi_add: （ensure accurate price calculation with pricelist）

maxhe2014:
[17.0][fix]sale_product_multi_add: （ensure accurate price calculation with pricelist）

- What was fixed:

1. The final price is now correctly computed using the sales order's assigned pricelist, instead of relying solely on the base list price.
2. Pricing calculations now properly consider the product quantity and customer information.

- Impact:

1. Ensures automatic application of customer-specific pricing strategies, quantity discounts, and promotional policies.
2. Chinese translations for the module have been optimized for better clarity.


rousseldenis:
@maxhe2014 Thanks for this.

It should be great to enhance your PR title if you want reviewer come (like `[17.0][IMP] sale_import_products: (...)`)

Moreover, you should not use other language than english in code comments (translation ok). Thanks

maxhe2014:
@rousseldenis

 Thank you for your guidance.

maxhe2014:
@rousseldenis   目前测试己通过，下一步我还需要做什么操作？The test has passed. What should I do next?

maxhe2014:
@rousseldenis Thank you for your guidance! I have:
1. Modified the PR title format as you requested
2. Changed all code comments from Chinese to English
3. Ensured all tests have passed (all 7 checks are green)

Could you please let me know if this PR now meets the merging criteria? Do I need to make any other changes?

sale_product_multi_add/wizards/sale_import_products.py
```diff
@@ -35,15 +35,34 @@ def create_items(self):
...
+        # Use Odoo's product price calculation method
+        # First get the product's base price
+        price = item.product_id.list_price
+
+        # Get applicable pricelist and calculate price
+        # (considering quantity and customer)
+
+        if sale.pricelist_id:
+            price = sale.pricelist_id._get_product_price(
+                item.product_id, item.quantity, sale.partner_id
```
alexey-pelykh: Bug: `_get_product_price(product, quantity, partner)` is wrong for Odoo 17. The method signature is `_get_product_price(self, product, *args, **kwargs)` and it passes through to `_compute_price_rule(products, quantity, currency=None, uom=None, date=False, ...)`. Here, `sale.partner_id` is passed as the `currency` positional argument, which will cause incorrect behavior or an error.

Use:
```python
price = sale.pricelist_id._get_product_price(
    item.product_id, item.quantity
)
```

Alternatively, consider removing the manual price assignment entirely and letting the `sale.order.line.new()` onchange/compute handle it -- that is how Odoo natively applies pricelist prices.

sale_product_multi_add/wizards/sale_import_products.py
```diff
@@ -35,15 +35,34 @@ def create_items(self):
...
+            )
+
+        # Set correctly calculated price
+        sale_line.price_unit = price
+
```
alexey-pelykh: Redundant: `product_uom` is already set in the `new()` dict above (line 41). This second assignment overwrites whatever computed fields may have adjusted.

sale_product_multi_add/tests/test_coverage_improvement.py
```diff
@@ -0,0 +1,144 @@
+# Copyright 2024 Your Name
```
alexey-pelykh: `# Copyright 2024 Your Name` -- please update to your actual name and the correct year. Same applies to `test_enhanced_coverage.py`.

Also, there is substantial overlap between this file and `test_enhanced_coverage.py`. Consider consolidating into one test file.

sale_product_multi_add/tests/test_sale_import_products.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # Set quantity
+        item = wizard.items[0]
+        item.quantity = 2.0
+
+        # Test _get_line_values method
+        line_values = wizard._get_line_values(so, item)
+
+        # Check that the returned values are correct
+        self.assertIsInstance(line_values, dict)
```
alexey-pelykh, line 133: This test creates a pricelist with no items/rules, so `_get_product_price` will just return the list price. The test does not actually verify that pricelist pricing works. Add a pricelist item (e.g., a fixed or percentage discount) and assert the computed `price_unit` differs from `list_price`.

sale_product_multi_add/i18n/zh_CN.po
```diff
@@ -7,7 +7,7 @@ msgstr ""
 "Project-Id-Version: Odoo Server 14.0\n"
 "Report-Msgid-Bugs-To: \n"
 "PO-Revision-Date: 2021-03-11 10:45+0000\n"
-"Last-Translator: 黎伟杰 <674416404@qq.com>\n"
```
alexey-pelykh, line 10: Typo: `qq.ccom` should be `qq.com`.

alexey-pelykh (changes requested):
Thanks for working on pricelist support in the multi-add wizard. The core idea is right -- the original code hardcoding `list_price` ignores pricelists entirely. However, there are several issues that need to be addressed before this can be merged.

### Bugs

**Wrong `_get_product_price` call signature.** In Odoo 17, `pricelist._get_product_price()` passes `*args, **kwargs` through to `_compute_price_rule(products, quantity, currency=None, uom=None, date=False, ...)`. The current call `_get_product_price(product, quantity, partner)` passes the `partner` record as the `currency` positional parameter, which will either raise an error or silently compute the wrong price. The `partner` parameter is not part of this API.

Correct call:
```python
price = sale.pricelist_id._get_product_price(
    item.product_id, item.quantity
)
```

**Redundant `product_uom` assignment.** The UoM is already set in the `new()` dict on line 41. Setting it again after pricelist computation (`sale_line.product_uom = item.product_id.uom_id`) is redundant and overwrites whatever `product_id_change` or compute methods may have set.

### Design concern

The `_get_line_values` method is decorated `@api.model`, so it has no `self` recordset context. When a `sale.order.line` is created via `.new(...)`, Odoo triggers `product_id_change` (or computed fields) which already computes `price_unit` based on the order's pricelist. So manually calling `_get_product_price` and overwriting `price_unit` afterward may conflict with or duplicate Odoo's native onchange logic. Consider whether it is sufficient to just remove the hardcoded `list_price` and let the ORM compute the price via the standard onchange mechanism -- that would be a simpler and more maintainable fix.

### Test files

- Both new test files use `# Copyright 2024 Your Name` -- this needs to be updated to the actual author name and year.
- `test_enhanced_coverage.py` and `test_coverage_improvement.py` have significant overlap (both test `create_items`, `select_products`, `_get_line_values`). These should be consolidated into a single test file.
- The pricelist test (`test_get_line_values_with_pricelist`) creates a pricelist with no items and no discount rules, so it does not actually verify that pricelist pricing differs from `list_price`. Add a pricelist item (e.g., a 10% discount) and assert the computed price reflects it.

### Translation file

- `Last-Translator` has a typo: `qq.ccom` should be `qq.com`.
- Missing newline at end of file (the diff shows `\ No newline at end of file`).

### Style

- The comments in the wizard ("Create sale order line record", "Use Odoo's product price calculation method", "Set correctly calculated price", "Ensure correct product unit of measure") describe *what* the code does, which is already apparent from the code itself. OCA convention favors minimal comments that explain *why*, not *what*.

alexey-pelykh (changes requested):
Thanks for the rework — good progress. The duplicate test files are cleaned up, the pricelist test now has actual discount rules, and relying on `.new()` onchange for pricing is the cleaner approach.

However, CI is red (pre-commit, test with OCB, test with Odoo all failing):

1. **Missing newline at end of file** in `test_sale_import_products.py` — this will trip pre-commit.
2. **`setUp` should be `setUpClass`** — the test class uses `def setUp(self)`. OCA convention is `@classmethod` + `setUpClass(cls)` to avoid re-creating fixtures per test method.
3. **"What" comments** — e.g. `# Create a new sale order line with basic information` describes what the code does, which is apparent. Comments should explain *why*.

Please fix CI failures first — happy to re-review once green.

rousseldenis:
@maxhe2014 Could you squash commits that can be together ? You should use only English in commits messages too.

Moreover, commits messages should look like `[IMP] sale_product_multi_add: (...)`.

Thanks

rousseldenis:
@maxhe2014 What's the status of this ?

PR #3991, opened 2025-11-05
[16.0][IMP] sale_discount_display_amount: modify report

Mohajiro:
- Add strikethrough pricing for discounted items
- Show total discount and undiscounted amounts
- Improve discount visibility in order totals


ivs-cetmix (changes requested):
Looks good, the only thing I would suggest is that the old prices should be red instead of grey.  Same as it's done for the optional products. This adds additional accent on the discount.

ivs-cetmix (approved):
Code and functional review LGTM

kaynnan (approved):

ivs-cetmix:
Hey @OCA/crm-sales-marketing-maintainers , here is a tiny PR which makes a lot of useful UX improvements though. Your review is highly appreciated! 😄

sale_discount_display_amount/report/sale_report_template.xml
```diff
@@ -1,16 +1,107 @@
...
+            </tr>
+
+            <!-- Custom TOTAL line -->
+            <tr>
+                <td><strong>Total</strong></td>
+                <td class="text-end">
+                    <t t-if="doc.discount_total">
+                        <!-- Struck-through total before discount -->
+                        <div
+                            style="text-decoration: line-through; color: #c9302c; font-size: 0.9em;"
```
rousseldenis: @Mohajiro @ivs-cetmix I'm not sure about this.

This should be done through a class instead (and a css definition too).

Mohajiro: @rousseldenis thank you for your review. I was fixed it, can you check it please?

alexey-pelykh (commented):
Solid improvement to discount visibility on reports. The inline-style issue flagged by @rousseldenis has been properly addressed with CSS classes. A few things to consider before merge:

**Blocking concern:**

1. `hide_original_total_row` inherits `account.document_tax_totals` **unconditionally**. This template is shared across sale orders, invoices, and potentially other documents. Hiding the original Total row globally will affect invoice reports and any other document using `account.document_tax_totals`. This should be scoped -- either guard the `t-if` with a condition like `t-if="not doc.discount_total"` instead of `False`, or use a different approach that only applies in the sale order context.

**Non-blocking suggestions:**

2. The string `<strong>Total</strong>` is hardcoded in English. For OCA modules this should be translatable. Consider using `t-esc` with a translatable source, or at minimum ensure it picks up translations (QWeb auto-translates static text in templates, so this may be acceptable, but worth confirming).

3. When a discount exists, the actual total is not wrapped in `<strong>` (just a plain `<div class="discount-total-actual">`), while the no-discount case wraps it in `<strong>`. This creates a visual weight difference. Consider making the discounted total bold as well.

4. The `discount-total-actual` CSS class is referenced in the XML but has no corresponding rule in `report.css`. If it is not needed, remove it; if styling is planned, add the rule.

5. Minor indentation inconsistency around the `o_struck_price` div in the Custom TOTAL section (extra leading whitespace on lines 37-42 in the template).

sale_discount_display_amount/report/sale_report_template.xml
```diff
@@ -1,16 +1,101 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
+    <!-- Hiding original Total row from account module -->
+    <template id="hide_original_total_row" inherit_id="account.document_tax_totals">
+        <xpath expr="//tr[contains(@class, 'o_total')]" position="attributes">
+            <attribute name="t-if">False</attribute>
```
alexey-pelykh, line 6: This `hide_original_total_row` template unconditionally hides the Total row in `account.document_tax_totals`, which is used by invoices, purchase orders, and other reports -- not just sale orders.

This will break the Total display on invoice PDFs for any company that installs this module.

Consider either:
- Adding a condition so it only applies when the document actually has discount data, e.g. `t-if="doc._name == 'sale.order' and not doc.discount_total"` (though this mixes concerns), or
- Moving the Total hiding + custom Total into the `sale.report_saleorder_document` inheritance only, avoiding any modification to the shared `account.document_tax_totals` template.

sale_discount_display_amount/report/sale_report_template.xml
```diff
@@ -1,16 +1,101 @@
...
+                    <t t-if="doc.discount_total">
+                        <!-- Struck-through total before discount -->
+                                <div class="o_struck_price">
+                                    <span
+                                t-field="doc.price_total_no_discount"
+                                t-options="{'widget': 'monetary', 'display_currency': doc.pricelist_id.currency_id}"
+                            />
+                                </div>
+                        <!-- Actual total -->
+                        <div class="discount-total-actual">
```
alexey-pelykh, line 35: Indentation is inconsistent here. The `<div>` and inner `<span>` have extra leading whitespace compared to the surrounding elements. Minor, but worth cleaning up for readability.

sale_discount_display_amount/report/sale_report_template.xml
```diff
@@ -1,16 +1,101 @@
...
+                            <span
+                                t-field="doc.amount_total"
+                                t-options="{'widget': 'monetary', 'display_currency': doc.pricelist_id.currency_id}"
+                            />
+                        </div>
+                    </t>
+                    <t t-else="">
+                        <strong>
+                            <span
+                                t-field="doc.amount_total"
```
alexey-pelykh, line 45: When no discount: `<strong><span t-field="doc.amount_total" .../></strong>`.
When discount exists: `<div class="discount-total-actual"><span t-field="doc.amount_total" .../></div>` -- no `<strong>`.

This makes the total visually lighter when a discount is applied, which seems unintentional. Consider wrapping the discounted total in `<strong>` as well for consistent emphasis.

sale_discount_display_amount/static/src/css/report.css
```diff
@@ -0,0 +1,11 @@
+.o_struck_price {
+    text-decoration: line-through;
+    color: #888;
+    font-size: 0.9em;
```
alexey-pelykh, line 4: `.o_struck_price` uses `color: #888` (grey) for the header-level strikethrough, while `.o_struck_price_line` below uses `color: #c9302c` (red) for line items. The earlier review by @ivs-cetmix suggested red to match optional products. Was the grey intentional for the header total? If so, a brief comment explaining the design choice would help future maintainers.

sale_discount_display_amount/report/sale_report_template.xml
```diff
@@ -1,16 +1,101 @@
...
+        </xpath>
+
+        <!-- Line totals with strikethrough for discounted items - using attributes instead of replace -->
+        <xpath expr="//td[@name='td_subtotal']" position="attributes">
+            <attribute name="t-if">False</attribute>
+        </xpath>
+
+        <!-- Add our custom subtotal display -->
+        <xpath expr="//td[@name='td_subtotal']" position="after">
+            <td
```
alexey-pelykh, line 73: The original `td_subtotal` renders for all lines including downpayments. This replacement cell has `t-if="not line.is_downpayment"`, which means downpayment lines will have a missing `<td>` in the row, potentially breaking the table layout.

Consider either keeping a fallback `<td>` for downpayment lines, or verifying that the original hidden `td_subtotal` still renders for downpayments (it won't, since it is set to `t-if="False"` above).

rousseldenis (approved):

Issue #3975, opened 2025-10-28
[16.0][sale_advance_payment] Incorrect amount_residual in sales orders with refund invoices


kikopeiro:
## Module

sale_advance_payment

## Describe the bug

The amount_residual field in sales orders shows an incorrect value when there are refund invoices (credit notes) associated with the order. This happens because the current calculation doesn't properly handle the sign of refund invoices.

## To Reproduce
**Affected versions**: 16.0

1. Create a sales order
2. Confirm the order and create the customer invoice
3. Post the invoice
4. Ceate a refund invoice (credit note)
5. Post the refund invoice
6. Check the amount_residual field in the sales order

**Expected behavior**

The amount_residual should correctly reflect the actual pending amount, properly accounting for both the original invoice and the refund invoice with their corresponding signs.

**Additional context**

The issue appears to be in the computation logic at: https://github.com/OCA/sale-workflow/blob/16.0/sale_advance_payment/models/sale.py#L57

The calculation should use amount_residual_signed instead of amount_residual to properly handle the sign of refund invoices.

Example runboat scenario:

<img width="1141" height="834" alt="Image" src="https://github.com/user-attachments/assets/e7422d1b-c4ac-40be-a0ce-384b843900d7" />

<img width="1602" height="182" alt="Image" src="https://github.com/user-attachments/assets/6b929919-4eaf-4ca3-9bb1-ab7062d0ff6b" />

Related PR: https://github.com/OCA/sale-workflow/pull/3863

kikopeiro:
CC @rousseldenis @JordiMForgeFlow @LoisRForgeFlow

PR #3992, opened 2025-11-05
[16.0][FIX] sale_advance_payment: Correct sale residual_amount value when s…

aritzolea:
Correct sale residual_amount value when sale has refund invoices.

Related issue: https://github.com/OCA/sale-workflow/issues/3975

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -83,8 +83,15 @@ class SaleOrder(models.Model):
                     advance_amount += line_amount
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
-            for inv in order.invoice_ids.filtered(lambda x: x.state != "cancel"):
-                paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual
+            invoices = order.invoice_ids.filtered(lambda x: x.state != "cancel")
+            for inv in invoices:
+                if inv.move_type == "entry" or inv.is_outbound():
+                    sign = -1
+                else:
+                    sign = 1
+                paid_amount = (
+                    inv.amount_total_in_currency_signed - sign * inv.amount_residual
+                )
                 if inv.currency_id != order.currency_id:
                     paid_amount = inv.currency_id._convert(
                         paid_amount,
--- a/sale_advance_payment/tests/test_sale_advance_payment.py
+++ b/sale_advance_payment/tests/test_sale_advance_payment.py
@@ -560,3 +560,48 @@ class TestSaleAdvancePayment(common.TransactionCase):
             }
         )._create_payments()
         self.assertEqual(self.sale_order_1.amount_residual, 3600)
+
+    def test_06_sale_advance_payment_invoice_refund(self):
+        self.assertEqual(self.sale_order_1.amount_residual, 3600)
+        # Confirm Sale Order
+        self.sale_order_1.action_confirm()
+
+        # Create Invoice
+        invoice = self.sale_order_1._create_invoices()
+        invoice.action_post()
+        self.assertEqual(self.sale_order_1.amount_residual, 3600)
+
+        refund = (
+            self.env["account.move.reversal"]
+            .with_context(
+                active_model="account.move",
+                active_ids=invoice.ids,
+            )
+            .create(
+                {
+                    "reason": "no reason",
+                    "refund_method": "refund",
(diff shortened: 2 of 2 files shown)


almumu (approved):

aritzolea:
@rousseldenis @JordiMForgeFlow @LoisRForgeFlow could you review this please?

sale_advance_payment/models/sale.py
```diff
@@ -83,8 +83,15 @@ def _compute_advance_payment(self):
...
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
-            for inv in order.invoice_ids.filtered(lambda x: x.state != "cancel"):
-                paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual
+            invoices = order.invoice_ids.filtered(lambda x: x.state != "cancel")
+            for inv in invoices:
+                if inv.move_type == "entry" or inv.is_outbound():
+                    sign = -1
+                else:
+                    sign = 1
```
alexey-pelykh, line 91: Good catch on the sign handling. This mirrors how Odoo core computes signed residuals. One minor readability note: you could also use `inv.amount_residual_signed` directly (which is already sign-aware), which would simplify this to:
```python
paid_amount = inv.amount_total_in_currency_signed - inv.amount_residual_signed
```
However, `amount_residual_signed` is a company-currency field, so the current approach with explicit sign logic is actually safer when multi-currency is involved. Keep as-is.

alexey-pelykh (commented):
The sign correction logic for `amount_residual` when refund invoices are present is sound. In Odoo 16, `amount_residual` is always positive (absolute) while `amount_total_in_currency_signed` is negative for credit notes. Without the sign adjustment, the `paid_amount` formula produces incorrect values for refunds, which is exactly the bug reported in #3975.

The approach using `is_outbound()` and `move_type == "entry"` to determine the sign is consistent with how Odoo core handles signed amounts (e.g. in `_compute_amount` on `account.move`). The test covers the key scenario: partial invoice + partial refund + partial payment.

One thing missing: the module version in `__manifest__.py` should be bumped (e.g. `16.0.1.1.2`) per OCA migration/versioning guidelines for bug fix PRs.

Overall, clean fix with good test coverage. LGTM after the version bump.

sale_advance_payment/tests/test_sale_advance_payment.py
```diff
@@ -560,3 +560,48 @@ def test_05_residual_amount_credit_note(self):
...
+        self.env["account.payment.register"].with_context(
+            active_model="account.move", active_ids=invoice.ids
+        ).create(
+            {
+                "payment_date": fields.Date.today(),
+                "amount": 600.0,
+            }
+        ).action_create_payments()
+        self.assertEqual(invoice.payment_state, "partial")
+        self.assertEqual(self.sale_order_1.amount_residual, 3000)
```
alexey-pelykh, line 607: Good scenario: you verify that after a partial refund (2 of 3 lines removed) and a 600 partial payment on the original invoice, the residual correctly stays at 3000.

Minor: `test_05` above uses `_create_payments()` while this test uses `action_create_payments()`. Both work, but for consistency within the test file, consider using the same method (`_create_payments()`).

sale_advance_payment/models/sale.py
aritzolea: The difference between `amount_residual` and `amount_residual_signed` fields is not only the sign.

`amount_residual` is in invoice currency.
`amount_residual_signed` is in company currency

alexey-pelykh (commented):
@aritzolea Good point — you are right that `amount_residual` is in invoice currency while `amount_residual_signed` is in company currency. The distinction is about currency, not just sign. Thanks for the clarification.

The fix itself is still sound — the sign correction logic is correct for the residual calculation. My comment about the version bump still applies though.

aritzolea:
> @aritzolea Good point — you are right that `amount_residual` is in invoice currency while `amount_residual_signed` is in company currency. The distinction is about currency, not just sign. Thanks for the clarification.
> 
> The fix itself is still sound — the sign correction logic is correct for the residual calculation. My comment about the version bump still applies though.

Module version is bumpled on PR merge with `ocabot merge` command. It's not bumped manually.

rousseldenis (approved):
LGTM

rousseldenis:
@aritzolea Don't take into accounts some of @alexey-pelykh comments. AI review.

PR #3993, opened 2025-11-05
[18.0][FIX] sell_only_by_packaging: min_sellable_qty invisible condition to use consu

rlimaeco:
After this change on the commit the `min_sellable_qty` on product variants list view (e.g.) will display values correctly


henrybackman (approved):

simahawk (commented):
LG but commit message should be rewritten to drop the version

simahawk:
fixed

simahawk (approved):

PR #3994, opened 2025-11-06
[16.0][ADD] sale_invoice_plan_report: add new module

kaynnan:
cc @marcelsavegnago @WesleyOliveira98 @CristianoMafraJunior


marcelsavegnago (approved):

marcelsavegnago:
ping @kittiu

CristianoMafraJunior (approved):

WesleyOliveira98 (approved):

kaynnan:
ping @kittiu

marcelsavegnago (approved):
LGTM

sale_invoice_plan_report/views/report_saleorder.xml
```diff
@@ -0,0 +1,62 @@
...
+                                            t-field="plan_line.plan_date"
+                                            t-options='{"widget": "date"}'
+                                        />
+                                    </td>
+                                    <td name="td_percentage" class="text-end">
+                                        <span t-field="plan_line.percent" />% </td>
+                                    <td name="td_amount" class="text-end">
+                                        <span
+                                            t-field="plan_line.amount"
+                                            t-options='{"widget": "monetary", "display_currency": doc.pricelist_id.currency_id}'
```
alexey-pelykh: Prefer `doc.currency_id` over `doc.pricelist_id.currency_id` -- it's the field the core sale report uses and avoids an extra hop through the pricelist record.

sale_invoice_plan_report/views/report_saleorder.xml
```diff
@@ -0,0 +1,62 @@
...
+                                        <span t-field="plan_line.installment" />
+                                    </td>
+                                    <td name="td_date" class="text-start">
+                                        <span
+                                            t-field="plan_line.plan_date"
+                                            t-options='{"widget": "date"}'
+                                        />
+                                    </td>
+                                    <td name="td_percentage" class="text-end">
+                                        <span t-field="plan_line.percent" />% </td>
```
alexey-pelykh: Minor: `t-field="plan_line.percent"` followed by a raw `%` can produce inconsistent spacing across locales. You could use `t-esc="'%.2f%%' % plan_line.percent"` for tighter control, or format it with `format_amount` / a dedicated widget.

alexey-pelykh (commented):
Thanks for this contribution, @kaynnan. Clean module structure and the QWeb template is well done overall. A few observations below.

**No tests**

The module ships without any tests. Even for a report-only module, it would be good to have at least one basic test that creates a sale order with an invoice plan and calls the report rendering to confirm it doesn't crash (a smoke test). Something like calling `self.env.ref('sale.action_report_saleorder').with_context(discard_logo_check=True)._render_qweb_html(sale_order.ids)` would be enough. This also ensures the xpath expression stays valid if the base template changes in a patch release.

**Currency via `pricelist_id.currency_id`**

The amount cell uses `doc.pricelist_id.currency_id` for the monetary widget. In Odoo 16, `sale.order` exposes `currency_id` directly (computed from the pricelist). Using `doc.currency_id` would be more robust -- it follows the same pattern the core sale report uses and avoids any edge case where the pricelist record might have changed after confirmation.

**Percentage display**

The `percent` field is a Float displayed with `t-field` and then a literal `%` appended after the span. This works but may produce awkward spacing or formatting depending on locale settings (e.g., `33.33 %` with varying decimal separators). Consider using `t-esc="'%.2f%%' % plan_line.percent"` or a `t-field` with `widget` options to keep the formatting consistent.

None of these are blocking -- the module works as intended and CI is green. Good job keeping it simple.

sale_invoice_plan_report/views/report_saleorder.xml
alexey-pelykh: The `percent` field is a `Float`, so this could render with unwanted decimal noise (e.g. `33.330000000000005%`). Consider:
```xml
<span t-esc="'%.2f' % plan_line.percent"/> %
```
or using `t-options='{"precision": 2}'` to cap the decimal places.

sale_invoice_plan_report/views/report_saleorder.xml
```diff
@@ -0,0 +1,62 @@
...
+                                    class="text-end"
+                                >Percentage</th>
+                                <th name="th_amount" class="text-end">Amount</th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            <t t-foreach="doc.invoice_plan_ids" t-as="plan_line">
+                                <tr>
+                                    <td name="td_installment" class="text-start">
+                                        <span t-field="plan_line.installment" />
```
alexey-pelykh, line 35: The `installment` field is an `Integer` (sequence number 1, 2, 3...), not a description. The column header "Installment" and README wording ("Installment names/descriptions") might mislead users into expecting a textual label. Consider renaming the header to `#` or `No.`, or showing the `invoice_type` alongside to differentiate advance vs installment rows.

alexey-pelykh (commented):
Thanks for this addition -- the invoice plan section is a nice complement to the standard sale report.

A few items I noticed while reviewing:

**Currency reference** -- The template uses `doc.pricelist_id.currency_id` for the monetary widget, which matches the convention in the standard Odoo 16 `report_saleorder_document`. That said, `doc.currency_id` (a stored related field pointing to the same value) would be slightly more concise and resilient to edge cases where the pricelist record might not be loaded. Low priority, just a thought.

**Installment column semantics** -- The `installment` field on `sale.invoice.plan` is an `Integer` (the installment sequence number: 1, 2, 3...). The column header says "Installment" and the README describes it as "Installment names/descriptions", which could set the wrong expectation. Consider either renaming the header to something like "#" or "No.", or rendering the `invoice_type` field alongside so the reader can distinguish advance vs installment rows.

**Percentage display** -- `<span t-field="plan_line.percent" />%` renders the raw float value followed by a literal `%`. For a plan_line with `percent = 33.33`, this would show `33.33%` which is fine, but if the float has trailing decimals (e.g. `33.330000000000005`) the display could look odd. Using `t-options='{"precision": 2}'` or formatting with `t-esc="'%.2f' % plan_line.percent"` and appending `%` would give cleaner output.

**No tests** -- The module ships without any test. A minimal test that installs the module (or renders the report on a sale order with an invoice plan) would help CI catch regressions. Even a simple `tests/test_report.py` that calls `self.env.ref('sale.action_report_saleorder').render_qweb_html(sale_order.ids)` would add value.

**Missing `__init__.py` content** -- The file exists but is empty (0 bytes). That's technically fine for a module with no Python code, but confirm this is intentional.

Overall, the approach is clean: a single inherited QWeb template with a well-targeted xpath, correct conditional guard (`use_invoice_plan and invoice_plan_ids`), and proper OCA scaffolding. The issues above are minor and don't block merging for an Alpha module, but the percentage formatting and test coverage would be worth addressing.

rousseldenis:
@kaynnan Thanks for this. I'm wondering if it should not be moved to sale-reporting repository

PR #3997, opened 2025-11-07
[19.0][MIG] sale_order_archive

theangryangel:
Adaptions required for 19.0:
* `toggle_active` removed under https://github.com/odoo/odoo/pull/183691, adapted to work with `action_archive` instead
* Removed dependency on demo data as per https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0


GarethGloSystems (approved):
Functionally tested on Runboat.

Approved based on the fact the module is functionally equivalent to V18.

My only pedantic feedback is that the default workflow allows end users to

- Archive a locked order in the 'Sales Order' state
- Unlock the order

This allows the order to enter a state which is explicitly forbidden in `action_archive`.

However, this is how the module works in at least V17 and V18 as well.

alexey-pelykh (approved):
Thanks for the migration, Karl.

Reviewed the `toggle_active` → `action_archive` adaptation, model logic, tests, views, and manifest.

The migration is clean and minimal:

- `action_archive()` override correctly replaces the removed `toggle_active()`, and the existing filter lambda (`not so.locked and so.state != "cancel" and so.active`) is semantically correct for the archive-only method since it already guarded only the archive direction.
- Tests properly updated to call `action_archive()` instead of `toggle_active()`, and demo data refs replaced with `create()` per the 19.0 migration guidelines.
- Views are unchanged (no method references in XML), manifest version bumped correctly.

All CI green. LGTM.

BhaveshHeliconia (approved):
Functional review LGTM!

rousseldenis (approved):
LGTM

PR #3998, opened 2025-11-08
[18.0][IMP] sale_product_identification

edescalona:
@Binhexteam

A new field is added to the IDs so that more dynamic validations can be performed using a Python expression.


edescalona:
Hello @arielbarreiros96 @epieters @rrebollo, if you could do a review of this PR, thank you.

epieters (approved):
Looks good.

rrebollo (approved):
### Code Review

LGTM!  
However, I suggest adding some tests to assess the impact of the new feature on the existing flow.

edescalona:
Hi @rousseldenis can you review please?

sale_product_identification/models/sale_order.py
arielbarreiros96, line 76:
```suggestion
        if products_opt_identification_ids:
            self.order_line.mapped(
                "product_template_id"
            )._eval_expression_identification(self)
```

Why not making the validation also for non mandatory ids?

sale_product_identification/models/product_template.py
```diff
@@ -40,3 +40,24 @@ def _check_product_tmpl_category_ids(self):
...
+                            "identification category %(category_identification)s "
+                            "are not met or the result is "
+                            "not being evaluated in the variable "
+                            "`result`, please verify."
+                        )
+                        % {
+                            "product": product_templ.name,
+                            "category_identification": prod_category.category_id.name,
+                        }
+                    )
```
arielbarreiros96:
```suggestion
        messages = []
        for product in self:
            failed_categories = self.env["product.template.id_category"]
            for prod_category in product.product_tmpl_category_ids:
                if not prod_category._eval_value(record or product):
                    failed_categories |= prod_category
            if failed_categories:
                categories = "\n".join(
                    _("\u2003\u2022\u2009%(category)s%(details)s")
                    % {
                        "category": category.category_id.name,
                        "details": f" ({category.message})"
                        if category.message
                        else "",
                    }
                    for category in failed_categories
                )
                messages.append(
                    _("\n%(product)s\n%(categories)s")
                    % {"product": product.name, "categories": categories}
                )
        if messages:
            raise ValidationError(
                _(
                    "The following identifications could not be validated, "
                    "please review the configured conditions:%(details)s"
                )
                % {"details": "".join(messages)}
            )
        return True
```
I would suggest to add a more user friendly message when validation fails, taking advantage from the messages already set on validation lines

Image

sale_product_identification/models/product_template_id_category.py
```diff
@@ -18,3 +32,91 @@ class ProductTemplateIdcategory(models.Model):
...
+    def _default_eval_context(self):
+        return self.env._("""# Available variables:
+            # env: environment on which the action is triggered
+            # order: Current record of the sale.order
+            # time, datetime, dateutil, timezone: useful Python libraries
+            # float_compare: utility function to compare
+            #                floats based on specific precision
+            # result: Variable in which the result of
+            #         the expression must be stored
+        """)
```
arielbarreiros96:
```suggestion
        return self.env._("""# Extra validation executed during identification checks.
            # Leave this field empty or commented to skip the extra validation.
            # Available variables:
            # env: environment on which the action is triggered
            # order: Current record of the sale.order
            # time, datetime, dateutil, timezone: useful Python libraries
            # float_compare: utility function to compare
            #                floats based on specific precision
            # result (bool): Set to True to approve the validation, keep False to fail it.
            # Example:
            # result = order.amount_total < 500.0  # approve only when amount is below 500
        """)
```

The ace widget shows no question mark (?) next to field name (actually field name is not visible), even in dev mode, making quite difficult for the end user to figure out the purpose of this field at first glance. I suggest to add some guidance as comments

arielbarreiros96 (changes requested):

arielbarreiros96 (approved):

sale_product_identification/tests/test_sale_product_identification.py
```diff
@@ -76,3 +86,118 @@ def test_validate_identification_id_number(self):
...
+                            "product_id": product.id,
+                            "product_uom_qty": 5,
+                        }
+                    )
+                ],
+            }
+        )
+        with self.assertRaises(ValidationError):
+            self.order.action_confirm()
+            self.assertEqual(self.order.state, "draft")
```
alexey-pelykh, line 133: This assertion will never execute. When `action_confirm()` raises on the previous line, control jumps to the `assertRaises` handler and skips the rest of the `with` block.

Move it outside:
```python
    self.order.action_confirm()
self.assertEqual(self.order.state, "draft")
```

sale_product_identification/models/product_template_id_category.py
```diff
@@ -18,3 +32,98 @@ class ProductTemplateIdcategory(models.Model):
...
+    def _onchange_value(self):
+        for prod_category in self:
+            prod_category._normalize_value()
+            if prod_category.value:
+                prod_category._test_python_expr()
+
+    def _eval_value(self, record):
+        self.ensure_one()
+        if self.value:
+            self._test_python_expr()
```
alexey-pelykh, line 103: `_test_python_expr()` is already called in `_onchange_value` at save time. Re-running it on every order confirmation adds overhead. Consider replacing this with an `@api.constrains('value')` so syntax is validated once at write time, and drop this call from `_eval_value`.

sale_product_identification/models/product_template_id_category.py
```diff
@@ -18,3 +32,98 @@ class ProductTemplateIdcategory(models.Model):
...
+                )
+                safe_eval(
+                    self.value,
+                    context,
+                    mode="exec",
+                    nocopy=True,
+                )
+                return context.get("result", False)
+            except Exception as ex:
+                raise ValidationError(message_error + str(ex)) from ex
```
alexey-pelykh, line 128: Catching bare `Exception` and passing `str(ex)` to the user can leak internal details (SQL, ORM tracebacks, etc.). Consider truncating to just the first meaningful line, e.g.:
```python
except Exception as ex:
    raise ValidationError(
        message_error + str(ex).split('\n')[0]
    ) from ex
```

sale_product_identification/i18n/sale_product_identification.pot
```diff
@@ -15,62 +17,112 @@ msgstr ""
...
+" # result = order.amount_total < 500.0 # aprobar solo cuando\n"
+" # # el importe es inferior a 500\n"
+" "
 
 #. module: sale_product_identification
 #. odoo-python
 #: code:addons/sale_product_identification/models/sale_order.py:0
 msgid "%(order)s%(message)s"
-msgstr ""
+msgstr "%(order)s%(message)s"
```
alexey-pelykh, line 69: The `.pot` template should not contain translated `msgstr` values -- those belong in language-specific `.po` files only. All `msgstr` entries in `.pot` must remain empty. Please regenerate with the standard Odoo i18n export.

alexey-pelykh (changes requested):
Thanks for the work on expression-based validation -- useful addition to the module.

A few items to address:

edescalona:
ping @alexey-pelykh

rrebollo (approved):
Code Review. LGTM!

arielbarreiros96 (approved):
Overall LGTM, though it seems to have a conflict on the es.po files, it's only a matter of adjusting a couple of terms

alexey-pelykh (approved):
Thanks for the follow-ups, @edescalona. All four items from my earlier review are addressed in the current diff:

- `self.assertEqual(self.order.state, "draft")` is dedented outside the `with self.assertRaises(...)` block, so it actually runs now.
- Syntax validation moved to `@api.constrains("value")` (`_check_value` → `_test_python_expr`) and is no longer re-run on every confirmation inside `_eval_value`.
- The eval error path surfaces only the first line: `str(ex).split("\n")[0]`.
- The `.pot` entries all carry empty `msgstr ""` now (translations live in `es.po`).

Good expansion of the test coverage around the expression feature too. LGTM.

rousseldenis:
@edescalona Could you fix conflict ?

PR #4001, opened 2025-11-10
[19.0][REF] sale_order_type: replace analytic_account_id by analytic distribution

lef-adhoc:
The analytic_account_id field was defined in the sale.order.type model but was not being used anywhere in the module's logic. This field has no practical functionality in the current implementation.

Additionally, this field was removed from the standard Odoo sale.order model in version 18.0

### DIFF
--- a/sale_order_type/migrations/19.0.1.3.0/post-migration.py
+++ b/sale_order_type/migrations/19.0.1.3.0/post-migration.py
@@ -1,0 +1,14 @@
+def migrate(cr, version):
+    cr.execute(
+        "SELECT EXISTS (SELECT 1 FROM information_schema.columns "
+        "WHERE table_name = 'sale_order_type' AND column_name = 'analytic_account_id')"
+    )
+    if not cr.fetchone()[0]:
+        return
+    cr.execute(
+        """
+        UPDATE sale_order_type
+        SET analytic_distribution = jsonb_build_object(analytic_account_id::text, 100.0)
+        WHERE analytic_account_id IS NOT NULL
+        """
+    )
--- a/sale_order_type/__manifest__.py
+++ b/sale_order_type/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     "name": "Sale Order Type",
-    "version": "19.0.1.2.0",
+    "version": "19.0.1.3.0",
     "category": "Sales Management",
     "author": "Grupo Vermon,"
     "AvanzOSC,"
--- a/sale_order_type/models/sale.py
+++ b/sale_order_type/models/sale.py
@@ -205,3 +205,10 @@ class SaleOrderLine(models.Model):
             if order_type.route_ids:
                 line.route_ids = order_type.route_ids
         return res
+
+    @api.depends("order_id.type_id")
+    def _compute_analytic_distribution(self):
+        res = super()._compute_analytic_distribution()
+        for line in self.filtered("order_id.type_id.analytic_distribution"):
+            line.analytic_distribution = line.order_id.type_id.analytic_distribution
+        return res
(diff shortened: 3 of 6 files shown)


lef-adhoc:
@pedrobaeza Could you check this?

pedrobaeza:
Well, I think the goal is to have a default analytic mapping, but as you say, right now it hasn't been covered.

matiasperalta1 (approved):
LGTM

alexey-pelykh (approved):
Clean removal of dead code. I verified that `analytic_account_id` is defined on `sale.order.type` but never consumed anywhere in the module's logic -- no compute, no `_prepare_invoice`, no tests, no onchange references it. The field was purely decorative (visible in the form but never wired to anything). The PR description correctly notes this field was also dropped from standard `sale.order` in Odoo 18.0.

CI is fully green. LGTM.

rousseldenis:
In fact, it has been removed partially in 18.0 migration following : https://github.com/OCA/sale-workflow/pull/3352#discussion_r1814369380

IMHO, it should be replaced by a default analytic distribution.

lef-adhoc:
@rousseldenis @pedrobaeza What do you think of the change? (Please ignore the version change in the manifest – this is because we’ve included the PR and need to update the module; I’ll delete it before it gets merged.)

pedrobaeza (commented):
You should include a migration script to convert the analytic account into the analytic distribution.

lef-adhoc:
@pedrobaeza I added the migration script there – how does it look now?

sale_order_type/migrations/19.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+
+    cr.execute(
+        """
+        SELECT id, analytic_account_id
+        FROM sale_order_type
+        WHERE analytic_account_id IS NOT NULL
+        """
+    )
+
+    for record_id, analytic_account_id in cr.fetchall():
```
pedrobaeza: Don't iterate on this, but do it in one shot with a single SQL. You can build a JSON in SQL.

lef-adhoc: @pedrobaeza Done!

sale_order_type/migrations/19.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
```
pedrobaeza: As you are not using anything of `openupgradelib` here, you can remove it and us raw `cr`.

sale_order_type/migrations/19.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+def migrate(env, version):
+    cr = env.cr
+
```
pedrobaeza: Better to not introduce empty lines inside a method.

lef-adhoc:
@pedrobaeza Done! Removed [openupgradelib](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and used raw [cr](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) with a plain SQL check, and removed the empty lines inside the method.

sale_order_type/migrations/19.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
+def migrate(cr, version):
+    cr.execute(
```
pedrobaeza, line 2: You can use here `tools.column_exists` in Odoo for commodity, but let's merge anyway,as this is correct.

pedrobaeza (approved):
/ocabot merge nobump

PR #4004, opened 2025-11-11
[18.0][ADD] sale_partner_shipping_default_carrier

AungKoKoLin1997:
This module assigns the carrier in the sales order based on its shipping partner.

@qrtl QT5938


yostashiro (commented):
The module isn't quite useful as is for our client without an editable carrier field. Let me check with them to see if they could go with the standard functionality instead.

AungKoKoLin1997:
Closing this PR since this module is not quite useful and we can just use the standard functionality for proposing the delivery method.

PR #4007, opened 2025-11-11
[18.0][IMP-FIX] sale_order_product_recommendation: Test the complete flow of a sales order with action_confirm

pilarvargas-tecnativa:
@Tecnativa

If the flow is not completed with the confirm action, it may result in a misleading test, and furthermore, common tests are not inheritable for the child modules of this functionality.

@pedrobaeza @victoralmau please review


pedrobaeza (approved):
Fast-tracking as obvious and only touching tests:

/ocabot merge patch

PR #3677, opened 2025-04-14
[18.0][MIG] sale_order_product_recommendation: MIgration to 18.0

CarlosRoca13:
This PR superseds https://github.com/OCA/sale-workflow/pull/3419 keeping the authorship of the migration commit.

cc @Tecnativa TT54286

ping @sergio-teruel @david-banon-tecnativa

PR #4010, opened 2025-11-12
[18.0][MIG] sale_order_product_recommendation_secondary_unit: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT54288

@carlos-lopez-tecnativa @victoralmau please review


pedrobaeza:
/ocabot migration sale_order_product_recommendation_secondary_unit

victoralmau (approved):
Code review OK

sale_order_product_recommendation_secondary_unit/readme/CONFIGURE.md
```diff
@@ -0,0 +1,10 @@
+To configure product secondary units, you need:
+
+- To have the "Display secondary units" permission.
```
carlos-lopez-tecnativa: This group doesn’t exist. I understand that the secondary unit uses the standard UoM group, so please update the README accordingly.

sale_order_product_recommendation_secondary_unit/wizards/sale_order_recommendation_view.xml
```diff
@@ -0,0 +1,73 @@
...
+                    invisible="not secondary_uom_id"
+                    widget="numeric_step"
+                    optional="show"
+                />
+                <field
+                    name="secondary_uom_id"
+                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"
+                    readonly="product_uom_readonly"
+                    optional="show"
+                />
```
carlos-lopez-tecnativa, line 31: If the README says that it’s necessary to configure the groups, then please add the group to the view.
Currently, I can see these fields without having the Units of Measure group assigned.

sale_order_product_recommendation_secondary_unit/wizards/sale_order_recommendation_view.xml
```diff
@@ -0,0 +1,73 @@
...
+                    name="secondary_uom_qty"
+                    invisible="not secondary_uom_id"
+                    widget="numeric_step"
+                    optional="show"
+                />
+                <field
+                    name="secondary_uom_id"
+                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"
+                    readonly="product_uom_readonly"
+                    optional="show"
```
carlos-lopez-tecnativa, line 30: I think it’s necessary to prevent creating records from this side.
When a new record is created here, it isn’t created correctly, the product or other default values aren’t set.
So, it’s better to prevent this. What do you think?

```suggestion
                    name="secondary_uom_id"
                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"
                    readonly="product_uom_readonly"
                    options="{'no_create': True}"
                    optional="show"
```

carlos-lopez-tecnativa (changes requested):
In the Kanban view, I got this error, and I think the migration of this module https://github.com/OCA/sale-workflow/pull/3677 to v18 is incomplete:
https://github.com/OCA/sale-workflow/blob/be211dcacc8dbd39192706fd4a02d95d969dd7a6/sale_order_product_recommendation/wizards/sale_order_recommendation_view.xml#L92

image

Additionally, I noticed that in the code it uses `self.env.user.company_id`, but that’s incorrect, it retrieves the user’s default company instead of the current one.
It should be changed to `self.env.company` to properly use the current company context.
https://github.com/OCA/sale-workflow/blob/be211dcacc8dbd39192706fd4a02d95d969dd7a6/sale_order_product_recommendation/wizards/sale_order_recommendation.py#L201

Could you please take a look?

pilarvargas-tecnativa:
> https://github.com/OCA/sale-workflow/blob/be211dcacc8dbd39192706fd4a02d95d969dd7a6/sale_order_product_recommendation/wizards/sale_order_recommendation.py#L201
> 
> Could you please take a look?

Add in new commit

sale_order_product_recommendation_secondary_unit/wizards/sale_order_recommendation.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        if (
+            float_compare(
+                self.secondary_uom_qty,
+                qty,
+                precision_rounding=self.secondary_uom_id.uom_id.rounding,
+            )
+            != 0
+        ):
+            self.secondary_uom_qty = qty
```
carlos-lopez-tecnativa, line 79: I think the logic for computing quantities could be simplified and made more consistent if we inherit from this mixin:
https://github.com/OCA/product-attribute/blob/18.0/product_secondary_unit/models/product_secondary_unit_mixin.py

For example, the `Dependency Type` field is not taken into account in the current implementation, but it is handled correctly in other modules that inherit from the mixin, which leads to different computed values.

I understand this might be out of scope for the migration itself, but if possible, refactoring the code to reuse the mixin would be a great improvement.
What do you think?
image

carlos-lopez-tecnativa (commented):
Please let me know what you think about this refactor; otherwise, I can approve these changes, but reconsider.

sale_order_product_recommendation_secondary_unit/wizards/sale_order_recommendation.py
pilarvargas-tecnativa: I would have to analyze it, but if you have already done so and believe that the change does not require much time, you can intervene in this same PR by adding that refactoring.

alexey-pelykh (approved):
Thanks for the migration and for addressing the `self.env.company` fix, @pilarvargas-tecnativa.

I went through both the base-module fix and the new secondary-unit glue module. Overall it looks solid -- CI is green, tests cover the key flows, and the latest push addresses the issues @carlos-lopez-tecnativa raised (UoM group gates, `no_create` option, company context). A few observations:

**The `self.env.company` fix (base module)**

Good catch -- `self.env.user.company_id` returns the user's *default* company rather than the company from the current context (which matters in multi-company setups). The two-line change is correct and self-contained.

**New module structure**

Manifest, dependencies, `auto_install`, views, tests, translations, readme -- everything expected for an OCA glue module is present.

**On the mixin suggestion from @carlos-lopez-tecnativa**

I agree this is a good candidate for a follow-up. The current onchange-based conversion duplicates logic that `product.secondary.unit.mixin` already handles (including `dependency_type` and cross-UoM factor). Inheriting from the mixin would reduce code and keep behaviour consistent with `sale_order_secondary_unit` and the purchase equivalent. That said, it is a refactor that goes beyond the scope of a migration PR, so treating it as a separate task seems reasonable.

**Minor note (non-blocking)**

The `product_uom_readonly` related field on `SaleOrderRecommendationLine` traverses `sale_line_id`, which is empty for *new* recommendation lines (products not yet on the SO). In that case it will resolve to `False`, which is the correct default for new lines, so this is fine in practice -- just noting it for context.

rousseldenis:
@pilarvargas-tecnativa What's the status of this ?

sale_order_product_recommendation_secondary_unit/wizards/sale_order_recommendation.py
CarlosRoca13: I have made the changes to use the mixin

CarlosRoca13 (approved):

carlos-lopez-tecnativa (approved):

pilarvargas-tecnativa:
Hello @rousseldenis 

I think it's ready to be merged

eduezerouali-tecnativa (approved):

eduezerouali-tecnativa:
can we merge this?

CarlosRoca13:
ping @pedrobaeza

eduezerouali-tecnativa:
ping @pedrobaeza can we mege this?

PR #4011, opened 2025-11-12
[18.0][IMP] sale_order_lot_selection: Define the appropriate lot data in stock moves

victoralmau:
Define the appropriate lot data in stock moves

Example use case:
- Product with Lot A and Lot B.
- SO with Lot A linked.
- The SO is confirmed and the OUT is generated with Lot A linked in the stock move
- The SO is modified to define Lot B.
- The OUT stock move is changed to Lot B.

Please @pedrobaeza and @christian-ramos-tecnativa can you review it?

@Tecnativa TT57112

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -1,4 +1,5 @@
 from odoo import api, fields, models
+from odoo.exceptions import ValidationError
 from odoo.tools import float_compare
 
 
@@ -21,6 +22,7 @@ class SaleOrderLine(models.Model):
         domain="domain_lot_id",
         copy=False,
         compute="_compute_lot_id",
+        inverse="_inverse_lot_id",
         store=True,
         readonly=False,
         precompute=True,
@@ -71,3 +73,19 @@ class SaleOrderLine(models.Model):
         for sol in self:
             if sol.product_id != sol.lot_id.product_id:
                 sol.lot_id = False
+
+    def _inverse_lot_id(self):
+        for item in self.filtered(
+            lambda x: x.product_id and x.product_tracking != "none" and x.move_ids
+        ):
+            moves = item.move_ids.filtered(lambda x: x.restrict_lot_id)
+            if any(move.state == "done" for move in moves):
+                raise ValidationError(
+                    self.env._(
+                        "You can't modify the Lot/Serial number "
+                        "because some stock move has already been done."
+                    )
+                )
+            pending_moves = moves.filtered(lambda x: x.state != "cancel")
+            if pending_moves:
+                pending_moves._set_restrict_lot_id_from_sol(item.lot_id)
--- a/sale_order_lot_selection/models/__init__.py
+++ b/sale_order_lot_selection/models/__init__.py
@@ -1,1 +1,2 @@
 from . import sale_order_line
+from . import stock_move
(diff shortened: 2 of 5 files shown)


pedrobaeza (commented):
If not possible to modify, this should return an exception that rollback the change for not having incoherent information.

Another doubt is to implement as `inverse`, as it's executed before the save, isn't it?

victoralmau:
> If not possible to modify, this should return an exception that rollback the change for not having incoherent information.
> 
So, do I add an error in the inverse if there is already a move done/cancelled? Perhaps another “_simpler_” possibility would be to make the `lot_id` field of the sales order line read-only if something has already been delivered.

> Another doubt is to implement as `inverse`, as it's executed before the save, isn't it?
The inverse is executed when saved, therefore, IMO is correct.

pedrobaeza:
And you should restrict it to the selected warehouse as well. The question is to restrict the selection to the minimum possible values for not confusing user selecting incorrect lots.

victoralmau:
> And you should restrict it to the selected warehouse as well. The question is to restrict the selection to the minimum possible values for not confusing user selecting incorrect lots.

Okay, I'll apply that restriction in `https://github.com/OCA/sale-workflow/pull/3987` instead. 
This would only be to propagate the change to stock moves.

pedrobaeza:
Sorry, I have mixed PRs... Yes, please raise an exception for impossible cases.

victoralmau:
> Sorry, I have mixed PRs... Yes, please raise an exception for impossible cases.

But don't you think it's better to make the field read-only once something has been delivered and avoid the problem?

pedrobaeza:
Ok, if you think you can do it without affecting performance, go ahead. Start filtering out non trackable products.

pedrobaeza (commented):
I'm thinking that the exception should still be for any bad code trying to assign it.

victoralmau:
> I'm thinking that the exception should still be for any bad code trying to assign it.

The exception has already been added. Some important comments:
- Although the restriction has been added, if it did not exist it would not be a problem because https://github.com/OCA/stock-logistics-workflow/blob/17f871dc175b6a0bded574aec9f1af634036533e/stock_restrict_lot/models/stock_move.py#L151 would be displayed.
- Currently, when changing the lot of a stock move, the linked `stock.move.line` are **NOT** deleted, i.e., the `_do_unreserve()` method is not automatically executed. Do you think this is something that should be added to `stock_restrict_lot` in https://github.com/OCA/stock-logistics-workflow/blob/17f871dc175b6a0bded574aec9f1af634036533e/stock_restrict_lot/models/stock_move.py#L158

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -39,3 +41,19 @@ def _compute_lot_id(self):
...
+            if any(move.state == "done" for move in moves):
+                raise ValidationError(
+                    self.env._(
+                        "You can't modify the Lot/Serial number "
+                        "because some stock move has already been done."
+                    )
+                )
+            pending_moves = moves.filtered(lambda x: x.state != "cancel")
+            if pending_moves:
+                pending_moves.restrict_lot_id = item.lot_id
```
pedrobaeza: I don't like the field `restrict_lot_id`. As I'm thinking this module, it's just a guidance to auto-select the lot when assigning, not a restriction. Not sure if that's the mission and the field name is very bad. In any case, let's fix the unreservation thing if needed.

pedrobaeza (approved):
Code review

victoralmau:
Ping @carlos-lopez-tecnativa

carlos-lopez-tecnativa (approved):

victoralmau:
This is ready to merge, right?

PR #4016, opened 2025-11-17
[18.0][FIX] sale_order_product_recommendation: improve the tests to be compatible with other modules tests

AungKoKoLin1997:
Before this PR, if there is an implementation that updates the `price_unit` of a `sale.order.line` after the line is created, the test cases in `sale_order_product_recommendation` fail. The reason is that the recomputation of `price_unit` is not skipped for a manually set `price_unit `(it should be skipped), so the price is recalculated.

With the current creation flow, creating an order line with a manual `price_unit` triggers the method at
https://github.com/odoo/odoo/blob/ec212f38e50520edf612522717952855c8c09a80/addons/sale/models/sale_order_line.py#L1274-L1278

which makes `technical_price_unit` and `price_unit` identical. As a result, the condition that should ignore recomputation for manual prices (see
https://github.com/odoo/odoo/blob/ec212f38e50520edf612522717952855c8c09a80/addons/sale/models/sale_order_line.py#L563-L570
)
is not met, and the recomputation is incorrectly applied.

This PR fixes that behavior.

Reference for incompatible module - https://github.com/OCA/sale-workflow/pull/4002

@qrtl QT5924

### DIFF
--- a/sale_order_product_recommendation/tests/test_recommendation_common.py
+++ b/sale_order_product_recommendation/tests/test_recommendation_common.py
@@ -73,7 +73,6 @@ class RecommendationCase(BaseCommon):
         cls.order1 = cls.env["sale.order"].create(
             {
                 "partner_id": cls.partner.id,
-                "locked": "True",
                 "date_order": "2021-05-05",
                 "order_line": [
                     Command.create(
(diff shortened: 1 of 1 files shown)


yostashiro (approved):
Code review. LGTM.

yostashiro:
@AungKoKoLin1997 Can you please update the commit message to keep sufficient information (please make it less verbose than your opening comment)? Please also change the tag to FIX.

kanda999 (approved):
Code check

yajo (approved):
/ocabot merge patch

PR #4018, opened 2025-11-17
[14.0][FIX] sale_order_line_price_history: Handle partner/product con…

luciaavanzosc:
…text in new sale order lines before save.

### DIFF
--- a/sale_order_line_price_history/static/src/js/sale_line_price_history_widget.js
+++ b/sale_order_line_price_history/static/src/js/sale_line_price_history_widget.js
@@ -34,9 +34,20 @@ odoo.define("sale_order_line_price_history.price_history_widget", function (requ
             if (this.data.id) {
                 additional_context = {active_id: this.data.id};
             } else {
+                var partner_id = false;
+                var product_id = false;
+
+                if (this.getParent().state.data.partner_id) {
+                    partner_id =
+                        this.getParent().state.data.partner_id.res_id ||
+                        this.getParent().state.data.partner_id[0];
+                }
+                if (this.data.product_id) {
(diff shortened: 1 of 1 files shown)


TadeuszKarpinski (approved):

alexey-pelykh (approved):
Good fix. The original code directly accessed `.res_id` on `this.data.order_partner_id` and `this.data.product_id` without null guards, which crashes when the price history button is clicked on a new (unsaved) sale order line where these fields may not yet be populated.

The fix correctly:
1. Defaults `partner_id` and `product_id` to `false` (safe Odoo convention for empty Many2one values).
2. Reads the partner from the parent form state (`this.getParent().state.data.partner_id`) instead of the line's related field (`this.data.order_partner_id`), since the related field may not yet be computed on a NewId record.
3. Adds null checks before accessing both fields.
4. Uses the `res_id || [0]` fallback pattern to handle both saved records (where `res_id` is available) and unsaved records (where the value may be in tuple format `[id, name]`).

CI is all green (pre-commit, tests with Odoo and OCB, codecov, runboat). Already has one approval from an OCA member.

APPROVE.

rousseldenis (approved):

rousseldenis:
@luciaavanzosc Could you fix pre-commit ?

luciaavanzosc:
@rousseldenis Done

.pre-commit-config.yaml
rousseldenis, line 144: Please don't modify this in this PR.

rousseldenis:
@luciaavanzosc Could you remove pre-commit change ?

PR #4020, opened 2025-11-17
[FIX][OU][17.0]: sale_order_invoice_amount : bad table name referenced in post migration

flotho:


flotho:
hum...
my bad, the table doesn't exist anymore. So we should deal it another way.

sale_order_invoice_amount/migrations/17.0.1.0.0/post-migration.py
```diff
@@ -37,46 +37,4 @@ def migrate(cr, version):
...
-                cr,
-                f"""
-                UPDATE ir_exports_line iel
-                SET name = '{new_field}'
-                FROM ir_exports ie
-                WHERE iel.name = '{old_field}' AND
-                ie.id = iel.export_id AND ie.resource = '{model}'
-                """,
-            )
+    openupgrade.rename_fields(env=cr, field_spec=_field_renames, no_deep=True)
```
alexey-pelykh, line 40: `env=cr` is wrong: `rename_fields` expects an Odoo `Environment` object (calls `env.cr` internally), but `cr` here is a raw cursor. This will raise `AttributeError`.

You already have `env = api.Environment(cr, SUPERUSER_ID, {})` on line 32. Use that:
```python
openupgrade.rename_fields(env, _field_renames, no_deep=True)
```

Also note `rename_fields` calls `rename_columns` internally -- since pre-migration already handled the column rename, verify this doesn't conflict.

alexey-pelykh (changes requested):
Thanks for identifying the root cause -- the `_ir_translation` table reference is indeed invalid in Odoo 17.0 (translations moved to jsonb columns since 16.0).

However, the proposed fix introduces a new bug: **`env=cr` is incorrect.** In the `migrate(cr, version)` signature, `cr` is a raw database cursor, but `openupgrade.rename_fields()` expects an Odoo `Environment` object (it internally does `cr = env.cr`). Passing the cursor as `env` will raise `AttributeError: 'Cursor' object has no attribute 'cr'`. This is confirmed by the failing CI.

You already have a proper `env` constructed earlier in the function (`env = api.Environment(cr, SUPERUSER_ID, {})`). Use that:
```python
openupgrade.rename_fields(env, _field_renames, no_deep=True)
```

Also note that `rename_fields` internally calls `rename_columns`, which will try to ALTER TABLE RENAME COLUMN. Since the pre-migration already handles column data transfer and drop, verify this doesn't conflict.

Alternatively, the simplest fix is to just delete the `_ir_translation` UPDATE block from the manual queries, since that's the only statement referencing a nonexistent table.

sale_order_invoice_amount/migrations/17.0.1.0.0/post-migration.py
rousseldenis: That's because env can be cr in `rename_fields()`

rousseldenis:
@flotho What's the status of this ?

flotho:
Hi @rousseldenis , I'm currently not working on this for the moment.
Feel free to superseeds

PR #4021, opened 2025-11-18
[18.0][MIG] sale_stock_reservation_issue_on_qty_at_date_widget: Migration to 18

Andrii9090-tecnativa:
Straightforward migration to v18
https://www.loom.com/share/b88a4ba36e694b13936f8ca3926dd8c8
@moduon
MT-11834


chienandalu (approved):
Nice! 👍

Shide (approved):
Code review

Shide:
/ocabot migration sale_stock_reservation_issue_on_qty_at_date_widget

Gelojr (commented):
Hello,

During functional testing of the module we found a potentially inconsistent behavior in the Qty at Date widget when a reservation conflict exists. Below is the exact reproduction scenario.

Steps to reproduce

1) Initial Setup
• Create a storable product with 20 units on hand.
• Ensure there are no pending moves or previous reservations.

2) Sales Order A (Customer 1)
• Create a Sales Order for Customer 1.
• Add the product with quantity 10.
• Set Scheduled Date = tomorrow.

Image

• Confirm the order.
• Ensure the picking is reserved.

3) Sales Order B (Customer 2) — Case 1 WITHOUT Scheduled Date
• Create a Sales Order for Customer 2.
• Add the same product with quantity 15.
• Do not set Scheduled Date.

Image

• Save the quotation.
• Check the Qty at Date widget.
→ Observed result: the icon appears YELLOW (reservation conflict). This seems correct.

Image

4) Sales Order B (Customer 2) — Case 2 WITH Scheduled Date
• In the same Sales Order, now set Scheduled Date = tomorrow (same date as Sales Order A).

Image

• Save the quotation.
• Check the widget again.
→ Observed result: the icon now appears RED, showing a standard stock coverage issue.

Image

Possible Issue observed
• Without Scheduled Date → YELLOW (reservation conflict).
• With Scheduled Date = tomorrow → RED, even though the reservation conflict is the same and total stock (20 units) is enough.

Question
Is this behavior expected?
Should the widget remain YELLOW in both cases (since the reservation conflict is identical), or is the RED alert intentionally taking priority when a Scheduled Date is provided?

Thank you for reviewing this point.

Andrii9090-tecnativa:
Thank you @Gelojr!

It works the same in v16.
This happens because when we do not set the Delivery Date, Odoo set by defaults today and, according to this, Odoo has a virtual available quantity to cover this order, but when we set the Delivery Date for tomorrow, it recalculates the virtual available quantity and indicate that it is out of stock.
image

This module works correctly and serves to notify us in the event that free quantity today is unless than quantity to deliver.

Gelojr (approved):
Congratulations on the work done @Andrii9090 .
The following functional tests were executed for this module.

- Test 1: Verified that the widget shows the default color when there is enough Available quantity and no reservation conflict; result OK.
- Test 2: Checked that without Scheduled Date the widget correctly shows yellow for reservation conflict, and with Scheduled Date it switches to red due to Odoo’s standard forecast logic; result OK.
- Test 3: Confirmed that when there is a reservation conflict and no Scheduled Date is set, the widget and the Available value in the popover turn yellow, and that setting Scheduled Date switches to red according to standard behavior; result OK.
- Test 4: Verified that the widget shows yellow for reservation conflict only in draft/sent when no Scheduled Date is set, and after confirming (with Scheduled Date set) it follows Odoo’s standard behavior; result OK.
- Test 5: Verified that in cases of real stock shortage without reservations the widget stays red according to Odoo’s native logic and never turns yellow; result OK.
- Test 6: Confirmed that only the line with reservation conflict shows yellow when no Scheduled Date is set, and that setting Scheduled Date switches the line to red following standard logic; result OK.

rafaelbn:
/ocabot migration sale_stock_reservation_issue_on_qty_at_date_widget

PR #4025, opened 2025-11-20
[16.0][ADD] sale_order_payment_terms_from_invoice_address

nicolas-delbovier-acsone:
Compute the "payment terms" on  sale orders based on the "invoice partner" instead of "Customer".

## Business case

Take the example of a business that relies on a primary logistics model where we fulfill orders placed through an **external e-commerce platform** or third-party reseller.

In this scenario:
1.  **Customer (`partner_id`):** The final recipient of the goods (the end-user).
2.  **Invoice Address (`partner_invoice_id`):** The external e-commerce platform/reseller responsible for payment.

The default Odoo behavior calculates the `payment_term_id` on the Sale Order based on the **Customer (`partner_id`)**. This is incorrect because our financial relationship is strictly with the reseller (the **Invoice Address**).

image

image


rousseldenis (approved):
Code review

alexey-pelykh (approved):
Thanks for this clean contribution, Nicolas. The business case is well-articulated and the implementation is solid.

I went through the module in detail:

- **ORM / compute logic** (`sale_order.py`): The `_compute_payment_term_id` override correctly calls `super()` first, then conditionally applies the invoice-partner payment term. Adding `@api.depends("partner_invoice_id")` properly extends the base dependency. The `with_company` usage for property field resolution is correct.

- **Config parameter**: The `ir.config_parameter` / `res.config.settings` wiring follows the standard pattern. Since Odoo deletes the param key when a boolean setting is unchecked, the truthiness check on `get_param` works correctly in both directions.

- **Settings view**: Clean xpath placement after `proforma_configuration`, standard OCA setting box layout.

- **Tests**: Good coverage of both the enabled and disabled paths.

- **Manifest**: Dependencies, versioning, and license are all correct.

One very minor observation (not blocking at all): the `get_param` call inside the compute method is executed on every recomputation of the field, though in practice it is cached within the Odoo environment so the overhead is negligible. This is a common and accepted pattern for feature toggles.

Looks good to merge.

PR #4029, opened 2025-11-24
[16.0][FIX] sale_order_lot_selection: Added a check, if the lot id has actually changed, before propagating it to stock moves. This prevents unwanted propagation of lot_id from sale order line to stock move.

baguenth:
This pull request is a follow up to https://github.com/OCA/sale-workflow/pull/3934. 

I have discovered some issues with this module after this pull request https://github.com/OCA/sale-workflow/pull/3057 was integrated.

My scenario is:

Sale order line doesn't have a lot id
Related stock moves get a restrict_lot_id independently from the sale order line
Sale order is changed and saved, lot_id was not changed but is in the vals for write function (it's still false)
Since the sale order line propagates the lot_id, no matter if it is filled or not, we try to set restrict_lot_id = False on stock moves which have a restrict_lot_id and are may already done. This is the moment, where the validation in stock_restrict_lot kicks in.

https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_restrict_lot/models/stock_move.py#L147

I am not exactly sure if my fix is what we want and if it is complete.

I think in general we should question:

Should we propagate empty lot_id to move ids
Should we propagate lot_id to move ids, regardless of the move ids state

We run the proposed fix productive in our products without issues.

FYI @rousseldenis @pedrobaeza

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -39,12 +39,41 @@ class SaleOrderLine(models.Model):
             )
 
     def write(self, vals):
+        """
+        Override of write to manage propagation of lot changes to stock moves.
+
+        Behavior
+        - If vals does not include lot_id, this method behaves exactly as the
+          standard write and returns.
+        - If lot_id is present, we compare the original lot per line with the
+          new lot after super().write(vals):
+          - If the lot did not actually change, the call is a no-op regarding
+            stock moves (nothing is propagated and no error is raised).
+          - If the order is in draft/quotation (not in sale/done), we propagate
+            the change to related stock moves by writing move.restrict_lot_id
+            to the new value (which can be a lot id or False).
+          - If the order is in sale/done:
+            - When the company setting allow_to_change_lot_on_confirmed_so is
+              True, we propagate the change to the related stock moves.
+            - When the company setting is False, we prevent changing the lot
+              and raise a UserError.
+        """
+        # Capture original lot ids to detect actual changes per record
+        original_lots = {rec.id: rec.lot_id.id for rec in self}
         res = super().write(vals)
+        if "lot_id" not in vals:
+            return res
         allow_to_change_lot = self.env.company.allow_to_change_lot_on_confirmed_so
-        if "lot_id" in vals and (
-            allow_to_change_lot or self.order_id.state not in ["sale", "done"]
-        ):
-            self.move_ids.write({"restrict_lot_id": vals.get("lot_id")})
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Waiting for @rousseldenis' approval.

jojicnikola (approved):

sebinthomasst (approved):

baguenth:
Hey @rousseldenis, 

any chance you can take a look?

alexey-pelykh (approved):
Thanks for the fix, @baguenth. The scenario you describe (lot_id present in vals but unchanged, causing unwanted propagation to done moves) is a real and nasty edge case. The approach is solid.

What I like:

- Capturing `original_lots` before `super().write()` and comparing per-line afterward is the right pattern for detecting actual changes vs. ORM noise.
- Switching from batch `self.move_ids.write(...)` to per-line `line.move_ids.write(...)` is a correctness improvement on its own -- the old code could write a single lot to moves belonging to different lines with different lots.
- The docstring clearly explains the branching logic, which makes the method much easier to reason about.

Minor observations (non-blocking):

1. **No test for the core scenario** -- the existing tests cover lot propagation and the `allow_to_change_lot` setting, but none exercise the specific case this fixes (lot_id in vals but unchanged, with moves that have an independently-set restrict_lot_id). A test that writes a SO line with `lot_id=False` while related moves already have a restrict_lot_id would lock this fix in against regressions.

2. **`self.env.company` outside the loop** -- `allow_to_change_lot` is read once from `self.env.company`, but if lines span multiple companies (multi-company recordset), the check should use `line.company_id` or `line.order_id.company_id` inside the loop (same as `_compute_lot_id_readonly` does). This is pre-existing, not introduced here, so not blocking.

All CI green, logic is correct -- LGTM.

baguenth:
@pedrobaeza How can we proceed here, since i am not getting any feedback from @rousseldenis?

baguenth:
Hi @pedrobaeza, hi @rousseldenis, 

just checking in on this PR again, which is ready to merge 🙂

It’s been open for a while, and I’d really like to get it over the finish line.
If there’s anything missing or concerns I can address, just let me know!

Thanks a lot!

baguenth:
Hi @pedrobaeza, hi @rousseldenis, 

another ping and request to integrate the pr.

FYI: We are using this in production for months now

Thanks!

pedrobaeza:
Can you add a test exercising the problem?

baguenth:
> Can you add a test exercising the problem?

Added the tests

pedrobaeza (approved):
/ocabot merge patch

PR #4033, opened 2025-11-26
[17.0][MIG] sale_partner_pricelist

syci:
Migration of sale_partner_pricelist from 16.0


sale_partner_pricelist/models/res_partner.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    @api.constrains("property_product_pricelist", "allowed_pricelist_ids")
+    def _check_allowed_pricelist(self):
+        for partner in self:
+            if (
+                self.env.company.use_partner_pricelist
+                and partner.commercial_partner_id.allowed_pricelist_ids
+                and partner.commercial_partner_id.property_product_pricelist
+                and partner.commercial_partner_id.property_product_pricelist
```
TadeuszKarpinski: Duplicate? Can you delete it?

sale_partner_pricelist/tests/test_sale_pricelist.py
```diff
@@ -0,0 +1,51 @@
...
+            "Please select one of the allowed pricelists.",
+        ):
+            self.partner.property_product_pricelist = self.pricelist_2
+
+    def test_02_sale_order_constraint(self):
+        with self.assertRaisesRegex(
+            ValidationError, "The selected Pricelist is not allowed for this Partner."
+        ):
+            order = self.env["sale.order"].create(
+                {
```
TadeuszKarpinski, line 46: Why you changed `Form` to `env["sale.order"].create` ?

rousseldenis: @TadeuszKarpinski Now, the use of Form() with sale orders should not be required as all onchanges that compute values have vanished from core.

sale_partner_pricelist/models/res_partner.py
rousseldenis: @syci

sale_partner_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+                record.partner_allowed_pricelist_ids = self.env[
+                    "product.pricelist"
+                ].search([])
+
+    @api.constrains("pricelist_id")
+    def _check_allowed_pricelist(self):
+        for record in self:
+            if (
+                self.company_id.use_partner_pricelist
+                and record.pricelist_id not in record.partner_allowed_pricelist_ids
```
alexey-pelykh, line 38: `self.company_id` should be `record.company_id` here -- in a multi-record constraint, `self` is the full recordset, which will raise `ValueError: Expected singleton` if multiple orders with different companies are constrained together.

sale_partner_pricelist/tests/test_sale_pricelist.py
```diff
@@ -0,0 +1,51 @@
...
+                    "partner_id": self.partner.id,
+                    "pricelist_id": self.pricelist_2.id,
+                }
+            )
+            order.save()
```
alexey-pelykh: This `.save()` call is a leftover from the `Form()` pattern used in 16.0. Since the test now uses `create()`, `.save()` is not an ORM method and would raise `AttributeError`. It passes only because the `ValidationError` fires during `create()` first.

Please remove this line -- the `assertRaisesRegex` context manager already catches the error from `create()`.

alexey-pelykh (changes requested):
Thanks for the migration, @syci.

Overall the module structure is clean and CI is green, which is great. I found a couple of issues that should be addressed before merge:

### Bug in `sale_order.py` constraint

In `_check_allowed_pricelist`, line 38 uses `self.company_id` instead of `record.company_id`:

```python
if (
    self.company_id.use_partner_pricelist  # <-- should be record.company_id
    and record.pricelist_id not in record.partner_allowed_pricelist_ids
):
```

Since `self` is the recordset being constrained (potentially multiple records), `self.company_id` will raise a `ValueError` ("expected singleton") if the constraint fires on a batch of sale orders belonging to different companies. Should be `record.company_id` to match the loop variable.

(I see this was carried over from 16.0 where it was also `self.company_id` -- good opportunity to fix it here.)

### Broken test: `order.save()` after `create()`

In `test_02_sale_order_constraint`, the test now uses `self.env["sale.order"].create(...)` (correct adaptation from `Form()` for 17.0), but `.save()` on line 51 is still there. `save()` is a `Form` method, not an ORM method -- this will fail with an `AttributeError`.

The `create()` call itself already triggers the constrains, so the `assertRaisesRegex` should wrap just the `create()` and the `.save()` line should be removed:

```python
def test_02_sale_order_constraint(self):
    with self.assertRaisesRegex(
        ValidationError, "The selected Pricelist is not allowed for this Partner."
    ):
        self.env["sale.order"].create(
            {
                "partner_id": self.partner.id,
                "pricelist_id": self.pricelist_2.id,
            }
        )
```

Since CI is passing, it seems like the `ValidationError` is raised during `create()` before `.save()` is reached, so the test passes by accident. Still, the dead `.save()` call should be cleaned up.

### Minor: `.pot` file version header

The `.pot` file still says `Project-Id-Version: Odoo Server 16.0` -- regenerating it via `oca-gen-addon-readme` / the standard export should update this to 17.0.

sale_partner_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+            else:
+                record.partner_allowed_pricelist_ids = self.env[
+                    "product.pricelist"
+                ].search([])
+
+    @api.constrains("pricelist_id")
+    def _check_allowed_pricelist(self):
+        for record in self:
+            if (
+                self.company_id.use_partner_pricelist
```
alexey-pelykh: `self.company_id` here operates on the full recordset, not the current loop record. Should be `record.company_id` to correctly evaluate the constraint per-record.

This bug was also present in 16.0 (`self.company_id` instead of `rec.company_id`), but it should be fixed as part of the migration.

```suggestion
                record.company_id.use_partner_pricelist
```

sale_partner_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+                record.partner_id
+                and record.partner_id.commercial_partner_id.allowed_pricelist_ids
+            ):
+                record.partner_allowed_pricelist_ids = (
+                    record.partner_id.commercial_partner_id.allowed_pricelist_ids
+                )
+            else:
+                record.partner_allowed_pricelist_ids = self.env[
+                    "product.pricelist"
+                ].search([])
```
alexey-pelykh, line 31: When the partner has no `allowed_pricelist_ids`, this does `search([])` which returns **all** pricelists across all companies. In a multi-company environment this would show pricelists from other companies in the domain filter.

Consider filtering by the current company:
```python
record.partner_allowed_pricelist_ids = self.env[
    "product.pricelist"
].search([("company_id", "in", (self.env.company.id, False))])
```

sale_partner_pricelist/tests/test_sale_pricelist.py
alexey-pelykh: `order.save()` doesn't exist on a standard ORM recordset -- `create()` already persists the record. The 16.0 version used `Form(...).save()` which makes sense because `Form` simulates the UI and `save()` commits the form.

Since the constraint is `@api.constrains('pricelist_id')`, it fires during `create()`. The `assertRaisesRegex` context should just wrap the `create()` call without the extra `.save()`. This line can be removed.

Alternatively, if you want to test through the UI simulation (like 16.0 did), you could keep using `Form` -- which is still available in 17.0 for testing. @rousseldenis clarified that `Form()` is no longer required for sale orders in 17.0, so the direct `create()` approach is fine, but then drop the `.save()` line.

sale_partner_pricelist/models/res_config_settings.py
```diff
@@ -0,0 +1,18 @@
...
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    use_partner_pricelist = fields.Boolean(
+        related="company_id.use_partner_pricelist",
+        readonly=False,
+    )
+
+    @api.onchange("use_partner_pricelist")
+    def _onchange_use_partner_pricelist(self):
```
alexey-pelykh: This `@api.onchange` was added during the migration (not present in 16.0). In 17.0, `@api.onchange` on `res.config.settings` is generally discouraged in favor of compute/inverse or explicit constraints.

That said, the logic itself (auto-enabling `group_product_pricelist` when `use_partner_pricelist` is activated) is reasonable. Just flagging it as something to keep an eye on -- if it causes issues with the settings save flow, consider moving it to a compute with `store=True` or handling it in `set_values()`.

alexey-pelykh (changes requested):
Thanks for the migration, @syci. The module structure and views look solid overall. I spotted a few issues that should be addressed before merge.

**Summary:**
- Bug in `sale_order.py` constraint: `self.company_id` should be `record.company_id`
- Test migration is incomplete: `order.save()` after `create()` is not valid ORM usage
- The `_compute_partner_allowed_pricelist_ids` fallback loads all pricelists without company filtering
- Minor: the `@api.onchange` in `res_config_settings.py` was added during migration but this pattern is discouraged in 17.0

rousseldenis (approved):

rousseldenis:
@TadeuszKarpinski Could you review ?

alexey-pelykh (commented):
Re-reviewed at `540ae52`. Both of my blocking points are fixed:

- Singleton bug resolved — `models/sale_order.py:36-38` now uses `record.company_id` inside the loop.
- Broken test cleaned up — `tests/test_sale_pricelist.py:41-50` wraps `create()` in `assertRaisesRegex` with no stray `.save()`.
- (Bonus) the migration-added `@api.onchange` was moved into a `set_values()` override — `models/res_config_settings.py:15-19`.

One minor, non-blocking point remains: the fallback in `_compute_partner_allowed_pricelist_ids` still does `search([])` (`models/sale_order.py:29-31`). Multi-company record rules already scope `product.pricelist`, so this is fine as the documented "show all pricelists" fallback — an explicit `[("company_id","in",(company.id,False))]` would just make the intent clearer. Fine as a follow-up.

Clearing my change request. Note `codecov/project` is red (coverage delta, not a test failure) — deferring to @rousseldenis's approval and the author on whether that's acceptable for the migration.

BhaveshHeliconia (approved):

syci:
> Thanks for the migration, @syci. The module structure and views look solid overall. I spotted a few issues that should be addressed before merge.
> 
> **Summary:**
> 
>     * Bug in `sale_order.py` constraint: `self.company_id` should be `record.company_id`
> 
>     * Test migration is incomplete: `order.save()` after `create()` is not valid ORM usage
> 
>     * The `_compute_partner_allowed_pricelist_ids` fallback loads all pricelists without company filtering
> 
>     * Minor: the `@api.onchange` in `res_config_settings.py` was added during migration but this pattern is discouraged in 17.0

Hello, the @BinhexTeam team will try to review your proposals, thank you.

PR #4039, opened 2025-11-26
[16.0][ADD] sale_partner_sale_contact[_on_project]: Add sale contact to orders, invoices and projects

maisim:
## New modules: Sale Contact Management

This PR introduces two complementary modules for managing sale contacts across the sales workflow:

### sale_partner_sale_contact

* Adds a sale_contact_partner_id field to sales orders and invoices
* Allows selecting a specific contact person from the customer's organization
* Automatically propagates the contact from order to invoice
* Optional display on PDF reports (configurable in settings)
* Auto-switches customer to parent company when a contact person is selected

### sale_partner_sale_contact_on_project

* Extends the contact management to projects
* Automatically copies the sale contact when creating projects from sales orders
* Maintains contact information throughout the project lifecycle

## Use case

In B2B scenarios, you often deal with specific contacts within a customer organization (procurement managers, project leads, etc.). While the invoice must go to the company, you need to track who you're actually working with for follow-ups, communication, and relationship management.

## Design choices

* Uses child contact hierarchy (existing Odoo pattern)
* Auto-switch behavior prevents accidental invoicing to individuals while preserving contact info
* Optional restriction modules available via settings for stricter enforcement
* No modification of core partner/invoice logic - purely additive

## Benefits

* Better contact tracking across sales → invoice → project workflow
* Cleaner reports with relevant contact information
* Improve global consistncy and avoid to get stuffs like Analytic Account linked to a person
* Respects commercial entity relationships while maintaining operational contacts
* All features are backward compatible and don't affect existing workflows. :crossed_fingers: 

## Related PRs
You can optionally use this module to limit partner_id selection to  companies

* account_invoice_partner_id_company_only: Restrict partner to companies  https://github.com/OCA/account-invoicing/pull/2180
* project_partner_id_company_only: Restrict partner to company https://github.com/OCA/project/pull/1611
* sale_partner_id_company_only: Restrict partner to companies https://github.com/OCA/sale-workflow/pull/4042


rrebollo:
@maisim are you sure we can achieve similar features with built-in?

maisim:
> @maisim are you sure we can achieve similar features with built-in?

I looked, I didn't found anything

rousseldenis:
@maisim As @rrebollo said, you can use the contact in standard.

Here, I've just add an invoice address if different (tested in runbot):

image

alexey-pelykh (changes requested):
Thanks for this contribution, Simon. Useful B2B use case. I went through both modules in detail and have feedback on a few areas.

### Functional overlap question

I see @rousseldenis and @rrebollo already raised this -- standard Odoo 16 does have invoice/delivery/contact addresses on partners, and the SO form shows invoice and delivery address fields. However, I understand this module targets a *different* concept: a named commercial contact person (procurement manager, project lead) that travels across SO → invoice → project, distinct from the billing/shipping addresses. If that understanding is correct, it would help to explain that distinction explicitly in the module description to avoid future confusion.

### Code-level findings

**1. Duplicated onchange logic across 3 models**

The `_onchange_partner_id_clear_sale_contact` method is copy-pasted identically in `sale.order`, `account.move`, and `project.project`. This is a maintenance burden -- any bug fix or behavior change would need to be applied in three places.

Consider extracting this into a mixin model (e.g. `sale.contact.mixin`) that all three inherit. The field definition and onchange would live in one place.

**2. Domain uses `child_of` but onchange checks `parent_id`**

The field domain uses `('id', 'child_of', partner_id)` which includes *all* descendants (children, grandchildren, etc.), but the onchange validation only checks `self.sale_contact_partner_id.parent_id != self.partner_id` -- this would incorrectly clear a valid grandchild contact. The validation should match the domain, e.g. using `commercial_partner_id` or walking up the parent chain.

**3. Auto-switch onchange may conflict with standard `onchange_partner_id`**

In `sale.order`, the standard `onchange_partner_id` already fires when `partner_id` changes and updates fiscal position, pricelist, payment terms, etc. Your onchange modifies `partner_id` *during* an onchange on `partner_id`, which can cause infinite recursion or unpredictable sequencing depending on the Odoo version. This is fragile -- the auto-switch from contact → company should probably be handled differently, perhaps via an `@api.onchange('partner_id')` with a guard flag, or by overriding the standard `onchange_partner_id` instead of adding a second one.

Same concern applies to `account.move` where the standard `_onchange_partner_id` is already complex.

**4. Report templates use different HTML elements**

The sale order report uses a `<p>` tag while the invoice report uses a `<div>` tag for the same content block. Should be consistent.

**5. `module_*` fields in `res.config.settings` reference non-existent modules**

The settings form includes checkboxes for `module_sale_partner_id_company_only`, `module_account_invoice_partner_id_company_only`, and `module_project_partner_id_company_only`. These reference modules from other PRs that aren't merged yet (and `project_partner_id_company_only` is in a completely different repo). Including auto-install triggers for modules that don't exist in this repo's dependency tree is risky -- if a user toggles the checkbox, Odoo will try to install a module that may not be available, resulting in an error. These should either be removed from this PR or the related modules should be merged first.

**6. Missing `copy=False` consideration**

When duplicating a sale order or invoice, `sale_contact_partner_id` will be copied by default. This might be desired, but it's worth being explicit about the choice -- add `copy=True` (to document intent) or `copy=False` if the contact should not carry over to duplicates.

**7. `account.move` view uses `attrs` for readonly instead of `states`**

The invoice view uses `attrs="{'readonly': [('state', '!=', 'draft')]}"` but the sale order view does not have any readonly condition. Both should be consistent. For the SO view, you'd typically want readonly in locked/done states too.

### Test coverage

The tests are solid for the happy path. A couple of gaps:
- No test for the auto-switch behavior (selecting a contact person as partner → auto-switch to company)
- No test for the `_prepare_invoice` propagation when `sale_contact_partner_id` is False (the `if` guard -- the test `test_06` covers this indirectly but doesn't assert the dict key is absent)
- The `codecov/patch` check is failing, so some lines aren't covered

### OCA conventions

- Missing `i18n/*.pot` export via `oca-gen-addon-readme` (the .pot files look hand-crafted rather than exported)
- The copyright header says "Odoo Community Association (OCA)" but the author is "OpenStudio SAS" -- for OCA contributions, copyright typically belongs to the contributing company. The copyright line in source files should match the `author` in manifest.

Overall a useful feature, but the duplicated onchange logic and the domain/validation mismatch should be addressed before merge.

JordiBForgeFlow:
@maisim Can you change the module to LGPL? Being a quite low level feature can be benefitial to both CE and EE users.

maisim:
> @maisim Can you change the module to LGPL? Being a quite low level feature can be benefitial to both CE and EE users.

No worries, I'll do it right away

maisim:
Hi @alexey-pelykh,
Thank you so much for your feedback; I've started making the changes accordingly. I'll commit each point individually, which will make reviewing easier. If necessary, I'll then merge them into a single commit per module, as is currently the case.

sale_partner_sale_contact/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = ["sale.order", "sale.contact.mixin"]
+    _name = "sale.order"
+
+    sale_contact_partner_id = fields.Many2one(
+        # Copied on duplication: the sale contact is part of the commercial
+        # context of the order and should carry over when duplicating.
+        copy=True,
```
MiquelRForgeFlow: By default, this already has copy=True.

sale_partner_sale_contact_on_project/models/project_project.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class ProjectProject(models.Model):
+    _inherit = ["project.project", "sale.contact.mixin"]
+    _name = "project.project"
+
+    sale_contact_partner_id = fields.Many2one(
+        # Copied on duplication: when cloning a project for the same customer
+        # the sale contact should carry over, just like the partner itself.
+        copy=True,
```
MiquelRForgeFlow: same as previous comment

sale_partner_sale_contact_on_project/models/project_project.py
```diff
@@ -0,0 +1,32 @@
...
+        help="Contact person for this project from the sale order. "
```
MiquelRForgeFlow: By the way, the point of the mixing is avoid declaring this field everywhere, but just to customize this help, you are redeclaring it again anyway, so the mixin loses part of its purpose.

sale_partner_sale_contact/models/sale_order.py
maisim: Hi @MiquelRForgeFlow,
 It's a request from @alexey-pelykh be be explicit on the behavior

sale_partner_sale_contact_on_project/models/project_project.py
maisim: Same as above

sale_partner_sale_contact/models/sale_order.py
MiquelRForgeFlow: Well, you can put it in the mixin instead :S

sale_partner_sale_contact_on_project/models/project_project.py
MiquelRForgeFlow: The help in the mixin could be: `Contact person. Only child contacts of the customer can be selected.`, this way you can avoid customizing for each model.

maisim: So it is better to just remove the field here?
I readd it here because I had a db error, the orm try to recreate the field, but I think I change the way I call the mixin between, let me do some tests to double check

sale_partner_sale_contact/models/sale_order.py
maisim: Oh yes sorry

maisim:
> 5. module_* fields in res.config.settings reference non-existent modules

Any change to get concerned module merged instead of remove the refs?

sale_partner_sale_contact_on_project/models/project_project.py
maisim: Seems to be ok, let me do some functional tests from the ui

maisim: The last changes broke the "auto switch" feature from the UI.
I first tried to re-add @api.onchange decorator but it seems to brake the MRO.
I had to use a new method name (instead of override _onchange_partner_id) , then remove the super()
Both onchange are now called on partner_id change.
It works.
Do you think this is the good way to do that?

MiquelRForgeFlow:
Repo was greenified in https://github.com/OCA/sale-workflow/pull/4230 so please rebase. Also, don't add last commit, thus modules are not dependencies. Also, I think it would be better to squash commits, there are too many.

maisim:
> Repo was greenified in #4230 so please rebase. Also, don't add last commit, thus modules are not dependencies. 

It was to ease functionnal test of modules installation from the settings. I will remove it

> Also, I think it would be better to squash commits, there are too many.

I did it it seprate commits to get one per asked modification. I will squash them in one commit per addon

MiquelRForgeFlow (approved):

sale_partner_sale_contact/models/res_config_settings.py
```diff
@@ -0,0 +1,29 @@
...
+    # Module installation checkboxes
+    module_sale_order_partner_company_only = fields.Boolean(
+        string="Restrict customer selection to company only in quotation and order",
+    )
+    module_account_invoice_partner_company_only = fields.Boolean(
+        string="Restrict customer selection to company only on invoice",
+    )
+    module_project_partner_company_only = fields.Boolean(
+        string="Restrict customer selection to company only in projects",
+    )
```
MiquelRForgeFlow: I think this have to be removed, this should be added when merging the corresponding PRs.

MiquelRForgeFlow (changes requested):

sale_partner_sale_contact/models/res_config_settings.py
maisim: There, it's done.

I took a quick look and don't see any other references to the additional modules in the commits.

I think it's not the right approach anyway. It's more of a configuration issue and doesn't really make sense, as @rrebollo kindly pointed out to me. https://github.com/OCA/account-invoicing/pull/2180

Either we do without it here and let everyone override their views to limit the display to companies, or I'll look into integrating it later as a configuration option in this module.
What do you think?

MiquelRForgeFlow (approved):

maisim:
Hi @MiquelRForgeFlow @alexey-pelykh ,

any chance to get this merged ?  :smiley: 

Thank you for your time. Let me know if I can help in any other way.

MiquelRForgeFlow:
@rousseldenis

alexey-pelykh (approved):
Thanks for the substantial rework, @maisim. Re-reviewed the current diff — the key items are addressed:

- The duplicated onchange logic is now a single `sale.contact.mixin` AbstractModel that `sale.order`, `account.move`, and `project.project` inherit. Clean.
- The clear-logic compares `commercial_partner_id` (matching the `child_of` domain) instead of `parent_id`, so valid grandchild contacts are no longer wrongly cleared — and the comment explains exactly why.
- The `module_*` settings fields that referenced not-yet-available modules are removed.
- `copy` is explicit and documented: `copy=True` on the mixin, overridden to `copy=False` on `account.move` (propagated instead via `_prepare_invoice`).
- Copyright headers now match the contributor.

The auto-switch termination is sound — once `partner_id` becomes the company, `commercial_partner_id == partner_id` stops the switch, so no recursion. LGTM.

rousseldenis:
@maisim @MiquelRForgeFlow I was concerned here above about the standard feature.

Could you explain the limitation ?

maisim:
> @maisim @MiquelRForgeFlow I was concerned here above about the standard feature.
> 
> Could you explain the limitation ?
@rousseldenis 
https://github.com/OCA/sale-workflow/pull/4039#pullrequestreview-3872531643

rousseldenis:
> I see @rousseldenis and @rrebollo already raised this -- standard Odoo 16 does have invoice/delivery/contact addresses on partners, and the SO form shows invoice and delivery address fields. However, I understand this module targets a _different_ concept: a named commercial contact person (procurement manager, project lead) that travels across SO → invoice → project, distinct from the billing/shipping addresses. If that understanding is correct, it would help to explain that distinction explicitly in the module description to avoid future confusion.

@maisim Yes I saw this, but I did not find an answer to my concern as:

- Why not using the `partner_id` field to represent your sale contact you want to introduce here? That field should be used for your purpose as you can select a contact in the customer hierarchy. That is in fact the sale contact.

maisim:
> > I see @rousseldenis and @rrebollo already raised this -- standard Odoo 16 does have invoice/delivery/contact addresses on partners, and the SO form shows invoice and delivery address fields. However, I understand this module targets a _different_ concept: a named commercial contact person (procurement manager, project lead) that travels across SO → invoice → project, distinct from the billing/shipping addresses. If that understanding is correct, it would help to explain that distinction explicitly in the module description to avoid future confusion.
> 
> @maisim Yes I saw this, but I did not find an answer to my concern as:
> 
>     * Why not using the `partner_id` field to represent your sale contact you want to introduce here? That field should be used for your purpose as you can select a contact in the customer hierarchy. That is in fact the sale contact.

In standard Odoo `partner_id` **can** hold a child contact, and in that sense it already carries "a contact". The reason this module introduces a dedicated `sale_contact_partner_id` instead of relying on `partner_id` is that the two fields answer different questions and are used differently downstream:

- **`partner_id` is the commercial/legal counterpart** of the document. It drives the fiscal position, pricelist, payment terms, receivable account, and the invoice/shipping addresses. In our B2B use case we deliberately want this to be the **company**, so that invoicing, analytic accounting and reporting stay attached to the legal entity — not to an individual (that's the "avoid analytic account linked to a person" point).
- **`sale_contact_partner_id` is the operational contact person** (procurement manager, project lead…) that must stay **stable and identical across the sale order, its invoices and the generated project**. `partner_id` cannot guarantee this: each document computes its own `partner_id`, and using it for the contact would force us to put an individual there — which is exactly what we want to avoid.

In short: `partner_id` = *who we contract/invoice with* (the company); `sale_contact_partner_id` = *who we actually deal with day-to-day* (the person), propagated consistently across the workflow.

The `auto-switch` behaviour reflects this: if a user picks a person in `partner_id`, we promote it to its parent company and move the person into `sale_contact_partner_id`, so the commercial entity stays correct while the contact is preserved.

I will make the module DESCRIPTION more explicit on this exact distinction. Let me know if this addresses your concern, if I am totally wrong, or if you'd prefer a different approach.

@MiquelRForgeFlow @JordiBForgeFlow @alexey-pelykh Do you already use the module? Any feedback?

rousseldenis:
Ok, I understand. In fact, this is the flow combined with this: https://github.com/OCA/sale-workflow/pull/4042

maisim:
> Ok, I understand. In fact, this is the flow combined with this: #4042

#4042 allows only company-type contacts to be displayed in selection fields (it hides person-type contacts).
It is complementary but not mandatory; the effect is purely visual. It can be installed on its own, without `sale_partner_sale_contact`.

Even without it, `sale_partner_sale_contact` will replace a person-type contact with their company and set set person as contact

In short, for now, they are complementary but not necessarily dependent on each other.

sale_partner_sale_contact/models/sale_contact_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+        # normally propagated from the originating sale order via
+        # _prepare_invoice(); carrying it over on a manual duplicate risks
+        # stale data (person has left, role changed, etc.).
+        copy=True,
+        help="Contact person for this record. "
+        "Only child contacts of the partner can be selected.",
+    )
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id_sale_contact_auto_switch(self):
```
rousseldenis: @maisim Prefer converting `sale_contact_partner_id` to a computed, stored, readonly=False field instead. Onchanges should be avoided

sale_partner_sale_contact/models/sale_contact_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+            and self.partner_id.commercial_partner_id != self.partner_id
+        ):
+            contact = self.partner_id
+            self.partner_id = contact.commercial_partner_id
+            self.sale_contact_partner_id = contact
+            return True
+        return False
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id_clear_sale_contact(self):
```
rousseldenis: Same as above. And merge code of both method as this is not convenient to have several places where the field is computed.

rousseldenis (changes requested):

FrancoMaxime (commented):
I agree with @rousseldenis comments

sale_partner_sale_contact/models/sale_contact_mixin.py
maisim: It seems not possible to keep the live change on the ui without onchange. (Change is visible only after saving)
Nevertheless, I move the field to computed field to manage the behavior outside of the ui (api, ...)

sale_partner_sale_contact/models/sale_contact_mixin.py
maisim: The field is now computed (store=True, readonly=False, precompute=True). The clearing rule lives in the compute, and the customer switch in a single helper called by create() and write()

rousseldenis (approved):

MiquelRForgeFlow:
Could we proceed with the merge?

PR #4040, opened 2025-11-26
feature - previous records with type_id and sale_type_id do not conta…

nmesa-binovo:
When the module is installed, sales orders and invoices (normal and corrective) do not contain a value, and since it is a mandatory field, they must contain some value in the old records.


sale_order_type/models/account_move.py
```diff
@@ -27,7 +27,9 @@ def _compute_sale_type_id(self):
             "active_model", False
         ) in ["sale.order", "sale.advance.payment.inv"]:
             return
-        sale_type = self.env["sale.order.type"].browse()
+        sale_type = self.env["sale.order.type"].search(
+            [("company_id", "in", [self.env.company.id, False])], limit=1
+        )
```
alexey-pelykh, line 32: This changes the default `sale_type` from an empty recordset to an actual record. This value is used in two places downstream:

1. **Line 33** (`self.sale_type_id = sale_type`) -- bulk-assigns ALL moves in `self` before the per-record loop. Previously this cleared them to `False`; now every record (including vendor bills, journal entries, etc.) gets a real sale order type.
2. **The `not in ["out_invoice", "out_refund"]` branch** -- assigns `sale_type` to non-customer-invoice moves, which should keep `sale_type_id` empty.

The original `browse()` (empty recordset) was intentional: it resets all records first, then the per-record loop applies the correct type only to customer invoices/refunds with a partner.

The issue described in the PR (old records missing a value on module install) is better solved with a `post_init_hook` that backfills `sale_type_id` on existing `out_invoice`/`out_refund` records where it is `False`, rather than changing the compute default for all move types.

sale_order_type/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 from . import models
 from . import reports
-from . import wizards
+from . import wizards
```
alexey-pelykh, line 5: The trailing newline was removed. This is the cause of the `pre-commit` CI failure. Please restore it.

sale_order_type/models/sale.py
```diff
@@ -70,7 +70,9 @@ def _compute_sale_type_id(self):
             )
             # Default user sale type value
             if not sale_type:
-                sale_type = record.default_get(["type_id"]).get("type_id", False)
+                type_id = record.default_get(["type_id"]).get("type_id", False)
+                if type_id:
```
alexey-pelykh, line 74: Good fix. `default_get` returns `{"type_id": <int>}`, not a recordset, so the original code was assigning an integer to `sale_type`. While Odoo tolerates integer assignment to Many2one in some contexts, explicitly browsing into a recordset is correct and avoids potential issues with downstream truthiness checks (e.g. `if not sale_type:` on integer `0` vs empty recordset).

alexey-pelykh (changes requested):
Thanks for the contribution. The fix in `sale.py` is correct and needed -- `default_get` returns an integer ID, not a recordset, so browsing it explicitly before assignment is the right approach.

However, the change in `account_move.py` introduces a regression, and the `__init__.py` edit removes the trailing newline (pre-commit failure). Details inline.

rousseldenis:
@nmesa-binovo IMHO, this seems weird.

In fact, I think the compute method is wrong as `depends_context` should not be used with stored fields as that can lead to wrong values.

rousseldenis:
Closing as invalid and user seems to have no activity

PR #4041, opened 2025-11-27
[IMP] sale_order_line_product_attribute_values: Change filter string

u0f:
With this change, the filter is less confusing for non-technical users, and maintaining consistency with the group_by.

image

MT-12788 @moduon @Shide @Gelojr


Shide (approved):

Andrii9090-tecnativa (approved):

Gelojr (approved):
Good job. @u0f  The following tests were executed.

- Test 1: Verified that the search field for product attribute values displays the updated label “Attribute Values” correctly in the English interface and appears only once.
- Test 2: Confirmed that in the Spanish interface the field still appears correctly with the expected translated label and without duplicates or visual inconsistencies.
- Test 3: Checked that filtering by attribute values using the updated “Attribute Values” field works as expected and returns only the sale order lines matching the selected attribute value.

rafaelbn (approved):
👍🏼 

/ocabot merge patch

PR #4042, opened 2025-11-27
[16.0][ADD] sale_order_partner_company_only: Restrict partner to companies

maisim:
Restrict partner to companies in quotations, orders and invoices


sale_partner_id_company_only/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Sale Partner ID Company Only",
+    "version": "16.0.1.0.0",
+    "category": "Sales",
+    "license": "AGPL-3",
+    "summary": "Restrict partner selection to companies only in sales and invoices",
+    "author": "OpenStudio SAS, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": [
+        "sale",
+        "account",
```
legalsylvain: I know it's more work, but I think that you should split that feature in two modules. one depending on sale, in sale-workflow and another depending on account, in account-invoicing.

legalsylvain (commented):
LGTM, except a dependency question.

thanks !

sale_partner_id_company_only/__manifest__.py
maisim: Yes I can create a module for the sale part (quotations, orders) and one for the accounting (invoices) part.

maisim: - Made change to restrict this module scope to sales (quotations and orders)
- Created a new module for the invoicing part : https://github.com/OCA/account-invoicing/pull/2180

sale_order_partner_company_only/readme/DESCRIPTION.rst
rrebollo, line 1: The addon's technical name should be `sale_partner_company_only`. I know is more work, but better now than never. Also it fits better with the other addon you mention here: `sale_partner_sale_contact`.

sale_order_partner_company_only/views/sale_order_views.xml
```diff
@@ -0,0 +1,15 @@
...
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form.partner.company.only</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']" position="attributes">
+                <attribute
+                    name="domain"
+                >[('is_company', '=', True), ('type', '!=', 'private'), ('company_id', 'in', [False, company_id])]</attribute>
+            </xpath>
```
rrebollo, line 13: **Regarding attribute interactions with other addons:**  
If another module modifies this same attribute, there could be conflicts. I strongly recommend depending on the `base_view_inheritance_extension` addon from the OCA/server-tools repository. This provides a cleaner and more maintainable way to handle view inheritance.

For example, you could conditionally add domains using join operators (`AND` by default if omitted):

```xml
<attribute name="$attribute" operation="domain_add"
           condition="$field_condition" join_operator="OR">
    $domain_to_add
</attribute>
```

This approach is more explicit and avoids overriding other modules' changes. You can find more details and examples here:  
https://github.com/OCA/server-tools/tree/16.0/base_view_inheritance_extension

rrebollo (changes requested):
Thank for your contribution. Please, consider my suggestions.

sale_order_partner_company_only/readme/DESCRIPTION.rst
maisim: Hi, Thank you very much for your review, it is really appreciated
We chose this name to be precise and avoid confusion. (explicit is better than implicit)
There more than one related field linked to partners on this model, with this name, when you read the name of the module, you directly know what it will do. 
It's sometime complicated to choose a module name, but if this one isn't perfect, I don't think it's "off-topic" enough to justify spending time changing it.

sale_order_partner_company_only/views/sale_order_views.xml
maisim: Wow! I've dreamt about it without even knowing it existed, I'm going to check it out. Thanks!

sale_order_partner_company_only/readme/DESCRIPTION.rst
rrebollo: My take: I understand your technical reasoning, but OCA naming conventions generally avoid reflecting implementation details (like `_id` suffixes) in the field name itself. If every addon did this, we'd see many `_id` fields, but that's not the current practice here.

Let's wait for other maintainers' opinions on this.

rrebollo: I'm just thinking following your line of thought and mine: `sale_order_partner_company_only`

sale_partner_id_company_only/views/sale_order_views.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form.partner.company.only</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']" position="attributes">
+                <attribute
+                    name="domain"
+                >[('is_company', '=', True), ('type', '!=', 'private'), ('company_id', 'in', [False, company_id])]</attribute>
```
alexey-pelykh: This replaces the full domain, overriding the base Odoo conditions `[("type", "!=", "private"), ("company_id", "in", (False, company_id))]`. If another installed module also sets a domain on `partner_id`, whichever loads last wins.

Using `base_view_inheritance_extension` with `domain_add` as @rrebollo suggested would make this additive:

```xml
<xpath expr="//field[@name='partner_id']" position="attributes">
    <attribute name="domain" operation="domain_add">
        [('is_company', '=', True)]
    </attribute>
</xpath>
```

This way the base domain conditions are preserved and the module only appends its own filter.

sale_order_partner_company_only/tests/test_sale_order_partner_company_only.py
```diff
@@ -0,0 +1,70 @@
...
+                        0,
+                        {
+                            "product_id": self.product.id,
+                            "product_uom_qty": 1,
+                        },
+                    )
+                ],
+            }
+        )
+        self.assertEqual(sale_order.partner_id, self.company_partner)
```
alexey-pelykh, line 78: Consider adding a negative test to verify the domain actually filters out individual contacts. For example:

```python
def test_03_individual_not_in_domain(self):
    """Test that individual partners are excluded by the domain."""
    individual = self.env["res.partner"].create({
        "name": "Test Individual",
        "is_company": False,
    })
    domain = [('is_company', '=', True), ('type', '!=', 'private')]
    partners = self.env["res.partner"].search(domain)
    self.assertNotIn(individual, partners)
```

This gives confidence the filtering actually works, not just that the domain attribute is present in the XML.

sale_partner_id_company_only/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2025 Odoo Community Association (OCA)
```
alexey-pelykh: Copyright should reflect the actual author (`OpenStudio SAS`) rather than OCA. OCA is typically listed as copyright holder only for code written directly by OCA contributors without a company affiliation, or after copyright transfer. Same applies to `tests/__init__.py`.

alexey-pelykh (commented):
Thanks for the contribution, Simon. Clean module structure, CI is green, and the split from account-invoicing was a good move.

I agree with @rrebollo's suggestion to use `base_view_inheritance_extension` with `domain_add` rather than overwriting the full domain attribute. The current approach replaces the base Odoo domain `[("type", "!=", "private"), ("company_id", "in", (False, company_id))]` entirely, which means any other module that modifies the `partner_id` domain will conflict depending on load order. With `domain_add`, you only inject the `("is_company", "=", True)` tuple and the base conditions remain untouched.

A few additional items to consider:

**1. Missing negative test** -- `test_02` confirms a company partner works, but there is no test asserting that the domain actually excludes individual contacts. Something like creating a contact with `is_company=False`, calling `name_search` with the view domain, and asserting the contact is not in the results would strengthen confidence.

**2. Missing newline at EOF** in `readme/DESCRIPTION.rst` (minor, `oca-gen-addon-readme` may complain on regeneration).

**3. Copyright headers** -- `__init__.py` and `tests/__init__.py` use `Copyright 2025 Odoo Community Association (OCA)` but the actual author/company is OpenStudio SAS per the manifest. The copyright should reflect the entity that wrote the code.

On the naming discussion: I see merit in both perspectives. The `_id` suffix is unusual in OCA module naming, but the argument about disambiguation from other partner fields on the model is reasonable. I lean toward `sale_order_partner_company_only` as @rrebollo suggested -- it removes the implementation-detail `_id` suffix while keeping the specificity the author wants.

Overall this is heading in the right direction. The `domain_add` change and a negative test would get this ready to merge.

sale_order_partner_company_only/views/sale_order_views.xml
alexey-pelykh, line 13: This overwrites the full domain, dropping the base Odoo conditions `[("type", "!=", "private"), ("company_id", "in", (False, company_id))]`. With `base_view_inheritance_extension` and `domain_add` you can inject only `('is_company', '=', True)` additively:

```xml
<xpath expr="//field[@name='partner_id']" position="attributes">
    <attribute name="domain" operation="domain_add">
        [('is_company', '=', True)]
    </attribute>
</xpath>
```

This avoids conflicts with any other module that modifies the same field's domain.

sale_order_partner_company_only/tests/test_sale_order_partner_company_only.py
alexey-pelykh, line 78: Please add a negative test that creates an `is_company=False` partner, applies the view domain in a `search()` or `name_search()`, and asserts the individual is not in the results. Without it we only know the happy path works, not that the restriction actually filters.

sale_partner_id_company_only/__init__.py
alexey-pelykh: Copyright should be `OpenStudio SAS` (the actual author per the manifest), not `Odoo Community Association (OCA)`. Same applies to `tests/__init__.py`.

alexey-pelykh (changes requested):
Thanks for this module, Simon -- the split from account-invoicing was the right call, and CI is fully green. A few things still need addressing before this can be merged:

**1. Domain overwrite (blocking)** -- The view currently replaces the full `partner_id` domain, dropping the base Odoo conditions (`type != private`, `company_id` filter). If another module also sets a domain on the same field, whichever loads last wins. Please switch to `base_view_inheritance_extension` with `domain_add` as @rrebollo suggested -- that way you only inject `('is_company', '=', True)` and the base conditions stay intact.

**2. Missing negative test** -- `test_02` proves a company partner can be assigned, but nothing verifies that individual contacts are actually *excluded* by the domain. A short `test_03` that creates an `is_company=False` partner, searches with the view domain, and asserts it is absent would close that gap.

**3. Copyright headers** -- `__init__.py` and `tests/__init__.py` attribute copyright to OCA, but the actual author is OpenStudio SAS per the manifest. These should match.

**4. Missing newline at EOF** in `readme/DESCRIPTION.rst` (minor, but `oca-gen-addon-readme` may complain on regeneration).

On the naming discussion: I lean toward @rrebollo's compromise of `sale_order_partner_company_only` -- it drops the `_id` implementation detail while keeping the specificity you wanted. Happy to hear other maintainers weigh in.

Items 1 and 2 are the blockers; 3 and 4 are quick fixes. Looking forward to the next iteration.

maisim:
@alexey-pelykh thank you very much for your review. I will do the changes in the week

PR #4045, opened 2025-11-28
[18.0][FIX] sale_order_lot_selection: Avoid the error when trying to select a lot if the user does not have Inventory permissions

victoralmau:
Avoid the error when trying to select a lot if the user does not have Inventory permissions

**Before**
antes

@Tecnativa TT57124


carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
How this have worked till now when you are a pure salesman!?

/ocabot merge patch

PR #4046, opened 2025-11-29
[FIX] sale_company_currency: Fix currency conversion logic (division instead of multiplication)

### LINKED ISSUE #4026: Company Currency in Sale Orders sale_company_currency
## Module
Company Currency in Sale Orders
sale_company_currency

## Describe the bug
When you display in list view SO, Total Currency (Company) show the amount multipliying Foreing Currency by Currency rate, for i.e.: SO in foreing Currency is 100.000 Bs, Rate Currency (Bs. / USD = 250Bs / 1USD), Company currency amount should show = 100.000 / 250 = 400 USD, instead currently show 250.000.000 USD.

## To Reproduce
**Affected versions**:

1. Create a SO with foreing currency and you should see the error in list view.
2.
3.

**Expected behavior**
SO in foreing Currency is 100.000 Bs, Rate Currency (Bs. / USD = 250Bs / 1USD), Company currency amount should show = 100.000 / 250 = 400 USD, instead currently show 250.000.000 USD. 

Image

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

fsmw:
## Summary
- Fix critical financial bug #4026 where foreign currency amounts were incorrectly multiplied instead of divided
- Update test to validate correct division logic  
- Add specific test case to ensure multiplication vs division logic is correct

## Bug Description
The bug caused incorrect currency conversion in the `sale_company_currency` module. When converting from foreign currency to company currency, the calculation was using multiplication instead of division, leading to values that were orders of magnitude incorrect.

**Example of the bug:**
- Order amount: 100,000 Bs (foreign currency)
- Exchange rate: 250 Bs / 1 USD
- **Expected result**: 100,000 ÷ 250 = 400 USD
- **Bug result**: 100,000 × 250 = 25,000,000 USD

## Changes Made
1. **Fixed conversion logic** in `sale_order.py:32`: Changed `*` to `/`
2. **Updated existing test** to use correct division logic
3. **Added new test** to specifically validate division vs multiplication logic

## Testing
- All syntax checks pass
- Code follows OCA conventions
- Tests validate the correct mathematical operation

Closes #4026

### DIFF
--- a/sale_company_currency/models/sale_order.py
+++ b/sale_company_currency/models/sale_order.py
@@ -29,5 +29,5 @@ class SaleOrder(models.Model):
             if order.currency_id.id == order.company_id.currency_id.id:
                 to_amount = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
+                to_amount = order.amount_total / order.currency_rate
             order.amount_total_curr = to_amount
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -168,8 +168,17 @@ class TestSaleOrder(TransactionCase):
         )
 
         # Set different currency rates for both orders
-        # First, change the currency of the original order to force conversion
-        self.sale_order.currency_id = self.currency_eur
+        # First, change the currency of both orders to force conversion
+        # Check what currency the company uses
+        if self.company.currency_id == self.currency_usd:
+            # Company uses USD, so use EUR for both orders
+            self.sale_order.currency_id = self.currency_eur
+            sale_order_2.currency_id = self.currency_eur
+        else:
+            # Company uses something else, use USD for both orders
+            self.sale_order.currency_id = self.currency_usd
+            sale_order_2.currency_id = self.currency_usd
+
         self.sale_order.currency_rate = 0.85
         sale_order_2.currency_rate = 1.20
 
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -168,6 +168,8 @@ class TestSaleOrder(TransactionCase):
         )
 
         # Set different currency rates for both orders
+        # First, change the currency of the original order to force conversion
+        self.sale_order.currency_id = self.currency_eur
         self.sale_order.currency_rate = 0.85
         sale_order_2.currency_rate = 1.20
 
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -101,3 +101,110 @@ class TestSaleOrder(TransactionCase):
             places=2,
             msg="Multiplication should give different result than correct division.",
         )
+
+    def test_04_amount_total_curr_rate_one(self):
+        """Test conversion when currency_rate = 1 (should behave like same currency)."""
+        self.sale_order.currency_id = self.currency_eur
+        self.sale_order.currency_rate = 1.0
+
+        self.sale_order._compute_amount_company()
+
+        self.assertEqual(
+            self.sale_order.amount_total_curr,
+            self.sale_order.amount_total,
+            "When currency_rate = 1, amount_total_curr should equal amount_total",
+        )
+
+    def test_05_amount_total_curr_extreme_rates(self):
+        """Test conversion with very small and very large currency rates."""
+        self.sale_order.currency_id = self.currency_eur
+
+        # Test with very small rate (strong target currency)
+        self.sale_order.currency_rate = 0.01  # 1 EUR = 0.01 USD
+        self.sale_order._compute_amount_company()
+        expected_small = self.sale_order.amount_total / 0.01
+        self.assertAlmostEqual(
+            self.sale_order.amount_total_curr,
+            expected_small,
+            places=2,
+            msg="Conversion with very small rate should work correctly",
+        )
+
+        # Test with very large rate (weak target currency)
+        self.sale_order.currency_rate = 1000.0  # 1 EUR = 1000 USD
+        self.sale_order._compute_amount_company()
+        expected_large = self.sale_order.amount_total / 1000.0
+        self.assertAlmostEqual(
+            self.sale_order.amount_total_curr,
+            expected_large,
+            places=2,
+            msg="Conversion with very large rate should work correctly",
+        )
+
+    def test_06_amount_total_curr_multiple_orders(self):
+        """Test computation with multiple sale orders in one batch."""
+        # Create a second sale order
+        sale_order_2 = self.env["sale.order"].create(
+            {
+                "partner_id": self.partner.id,
+                "company_id": self.company.id,
+                "currency_id": self.currency_eur.id,
+                "order_line": [
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": self.product1.id,
+                            "product_uom_qty": 3,
+                            "price_unit": 75.0,
+                        },
+                    )
+                ],
+            }
+        )
+
+        # Set different currency rates for both orders
+        self.sale_order.currency_rate = 0.85
+        sale_order_2.currency_rate = 1.20
+
+        # Compute for both orders in batch
+        orders = self.sale_order + sale_order_2
+        orders._compute_amount_company()
+
+        # Verify both orders computed correctly
+        expected_1 = self.sale_order.amount_total / 0.85
+        expected_2 = sale_order_2.amount_total / 1.20
+
+        self.assertAlmostEqual(
+            self.sale_order.amount_total_curr,
+            expected_1,
+            places=2,
+            msg="First order should compute correctly in batch",
+        )
+        self.assertAlmostEqual(
+            sale_order_2.amount_total_curr,
+            expected_2,
+            places=2,
+            msg="Second order should compute correctly in batch",
+        )
+
+    def test_07_amount_total_curr_edge_cases(self):
+        """Test edge cases like zero amount_total and extreme values."""
+        self.sale_order.currency_id = self.currency_eur
+
+        # Test with zero amount_total
+        self.sale_order.order_line[0].price_unit = 0.0
+        self.sale_order.order_line[1].price_unit = 0.0
+        self.sale_order.currency_rate = 0.85
+
+        self.sale_order._compute_amount_company()
+
+        self.assertEqual(
+            self.sale_order.amount_total_curr,
+            0.0,
+            "Zero amount_total should result in zero amount_total_curr",
+        )
+
+        # Reset to normal values
+        self.sale_order.order_line[0].price_unit = 50.0
+        self.sale_order.order_line[1].price_unit = 100.0
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -62,13 +62,42 @@ class TestSaleOrder(TransactionCase):
         self.sale_order._compute_amount_company()
         amount_total_curr_rounded = round(self.sale_order.amount_total_curr, 2)
         amount_total_converted_rounded = round(
-            self.sale_order.amount_total * self.sale_order.currency_rate, 2
+            self.sale_order.amount_total / self.sale_order.currency_rate, 2
         )
         self.assertEqual(
             amount_total_curr_rounded,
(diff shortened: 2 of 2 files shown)


Suppli360sw:
@fsmw Saludos Fernando, entiendo que aun no ha sido aprobado este PR por OCA? Aun esta en revisión?

Suppli360sw:
Who can help me to know if this bug was fixed? This PR was merge to main? I would like to test it in my deploy. @fsmw

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +62,42 @@ def test_02_amount_total_curr_different_currency(self):
...
+                "using the currency rate (division, not multiplication)."
             ),
         )
+
+    def test_03_amount_total_curr_conversion_logic(self):
+        """Test specific conversion logic to ensure division is used, not multiplication."""
+        # Set up a specific scenario to test the conversion logic
+        self.sale_order.currency_id = self.currency_eur
+        # Set a specific currency rate for testing
+        self.sale_order.currency_rate = 0.85  # 1 EUR = 0.85 USD
```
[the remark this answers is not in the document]
twalter-c2c: To me, the comment is clear as is, no need to change it. I agree that usually EUR is worth more than USD, so the currency rate could be, let's say, 1.17 instead of 0.85. But it does not affect the logic, which is correct now.

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +62,42 @@ def test_02_amount_total_curr_different_currency(self):
...
+        )
+        
+        # Verify that multiplication would give a different (wrong) result
+        wrong_amount = self.sale_order.amount_total * 0.85
+        self.assertNotAlmostEqual(
+            self.sale_order.amount_total_curr,
+            wrong_amount,
+            places=2,
+            msg="Multiplication should give a different result than the correct division."
+        )
```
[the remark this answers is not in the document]
twalter-c2c: While this may be relevant for real-life scenarios, I do not think it matters in the case of a controlled test. Looks good, IMO. Especially that the `currency_rate` of 1.0 is verified below.

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
...
+                        },
+                    )
+                ],
+            }
+        )
+
+        # Set different currency rates for both orders
+        # First, change the currency of both orders to force conversion
+        # Check what currency the company uses
+        if self.company.currency_id == self.currency_usd:
```
twalter-c2c, line 173: I am not sure why we need to check the company currency here. We should know the currency (it is USD if I am not mistaken), no? Also, in the `test_02_amount_total_curr_different_currency`, the EUR is assigned without checking the company currency. You can always verify if the company currency is USD in the test to double-check. Or do I miss something?

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
...
+                "using the currency rate (division, not multiplication)."
             ),
         )
+
+    def test_03_amount_total_curr_conversion_logic(self):
+        """Test specific conversion logic to ensure division is used."""
+        # Set up a specific scenario to test the conversion logic
+        self.sale_order.currency_id = self.currency_eur
+        # Set a specific currency rate for testing
+        self.sale_order.currency_rate = 0.85  # 1 EUR = 0.85 USD
```
twalter-c2c, line 85: To be honest, I would prefer not to set the `currency_rate` on the SO directly. It is a computed stored field. It would be cleaner to update the rate directly on the currency. In fact, you probably do not even need to update it, but rather just check the rate. It should already be set to a proper value.

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
...
+            # Company uses USD, so use EUR for both orders
+            self.sale_order.currency_id = self.currency_eur
+            sale_order_2.currency_id = self.currency_eur
+        else:
+            # Company uses something else, use USD for both orders
+            self.sale_order.currency_id = self.currency_usd
+            sale_order_2.currency_id = self.currency_usd
+
+        self.sale_order.currency_rate = 0.85
+        sale_order_2.currency_rate = 1.20
```
twalter-c2c, line 183: Re: my previous comment to not update the `currency_rate` directly. I mentioned that there is probably no need to update it. For this test case, you would need to update the rate (but on the currency, as I suggested) and make sure that the proper rate is taken into account based on `date_order`.

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
...
+        )
+        self.assertAlmostEqual(
+            sale_order_2.amount_total_curr,
+            expected_2,
+            places=2,
+            msg="Second order should compute correctly in batch",
+        )
+
+    def test_07_amount_total_curr_edge_cases(self):
+        """Test edge cases like zero amount_total and extreme values."""
```
twalter-c2c, line 208: What are the extreme values here?

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
...
+        self.assertEqual(
+            self.sale_order.amount_total_curr,
+            0.0,
+            "Zero amount_total should result in zero amount_total_curr",
+        )
+
+        # Reset to normal values
+        self.sale_order.order_line[0].price_unit = 50.0
+        if len(self.sale_order.order_line) > 1:
+            self.sale_order.order_line[1].price_unit = 100.0
```
twalter-c2c, line 227: What is the purpose of this part?

twalter-c2c (commented):
Thank you for this fix. While the fix itself is fine, I would suggest refactoring tests. Please check my comments. Additionally, `test_05_amount_total_curr_extreme_rates` does not seem to bring any value if I am not mistaken. It is not harmful in any way, either, so I did not suggest removing it, but please check if there is a point in keeping it.

TDu:
This fix has already been done for version 16.0 https://github.com/OCA/sale-workflow/pull/3240 
Why not just doing a forward port ?

Although this PR adds lots of unit tests but the  code coverage decreases.
Are they useful ?

And please squash some of the commits !
Like this one https://github.com/OCA/sale-workflow/pull/4046/changes/5ceceaba7417b2e76afcdc28e70eb78af02faaf9 is clearly related to the previous changes.
And the one for the unit test.

:robot:  ?

sale_company_currency/models/sale_order.py
```diff
@@ -29,5 +29,5 @@ def _compute_amount_company(self):
             if order.currency_id.id == order.company_id.currency_id.id:
                 to_amount = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
+                to_amount = order.amount_total / order.currency_rate
```
mmequignon, line 32: To avoid errors, use res.currency._convert.
Provide `date=order.date_order or None`.
same as we use float utils for comparing float, etc…

mmequignon (changes requested):
I agree with @TDu , we should forward port the 16.0 pr.

alexey-pelykh (commented):
Thanks for tackling this, Fernando. The one-line fix in `sale_order.py` is correct and well-motivated -- `amount_total / currency_rate` aligns with how Odoo core itself uses the field (see `_compute_partner_credit_warning` in `sale.order`).

**Important note on the 16.0 forward-port suggestion:** PR #3240 on 16.0 actually changed `/` to `*` (the opposite direction from this PR). Forward-porting #3240 to 18.0 would re-introduce the multiplication bug. The direction in this PR (`*` -> `/`) is the correct one.

That said, I agree with the existing feedback that the tests need rework before this can merge:

1. **Squash commits** -- 7 commits including 5 "fix test" iterations should be squashed into 1-2 clean commits (fix + test improvements).

2. **Don't write directly to `currency_rate`** -- as @twalter-c2c noted, it's a computed stored field. Set rates via `res.currency.rate` records so the compute chain is exercised end-to-end.

3. **Trim test bloat** -- `test_03` is redundant with `test_02`. `test_05` (extreme rates) and `test_07` (edge cases) add little value and contribute to the coverage decrease flagged by codecov. The tail of `test_07` that resets prices without asserting anything is dead code. I'd suggest keeping `test_02` (conversion logic), `test_04` (rate=1), and `test_06` (batch), and dropping the rest.

4. **Call out the setUpClass fix** -- the original test data had a broken order line tuple (two dicts in one `(0, 0, ...)` command). This PR quietly fixes it, which changes `amount_total`. Worth noting in the commit message.

sale_company_currency/models/sale_order.py
alexey-pelykh, line 32: The fix is correct. In Odoo core, `currency_rate` is computed via `_get_conversion_rate(from_currency=company_currency, to_currency=order_currency)`, so dividing is the right operation to convert back from order currency to company currency. Odoo's own `sale.order` does exactly `amount_total / order.currency_rate` in `_compute_partner_credit_warning`.

Note: the 16.0 PR #3240 that @TDu referenced actually went in the **opposite** direction (changing `/` to `*`), which introduced this same bug on 16.0. Forward-porting #3240 to 18.0 would be incorrect here -- the current PR's direction is the right one.

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -34,12 +34,16 @@ def setUpClass(cls):
                             "product_uom_qty": 2,
                             "price_unit": 50.0,
                         },
+                    ),
```
alexey-pelykh, line 37: Good catch fixing the `setUpClass` -- the original code had two dicts inside a single `(0, 0, ...)` tuple, which meant the second product was silently ignored. Worth calling this out explicitly in the commit message since it changes the test data (now `amount_total` includes both order lines).

sale_company_currency/tests/test_sale_company_currency.py
alexey-pelykh, line 85: Agree with @twalter-c2c: directly writing to `currency_rate` (a computed stored field) is fragile and bypasses the actual compute chain. Better to create `res.currency.rate` records for the test date so the rate flows through `_compute_currency_rate` naturally. This also validates the full dependency chain (`currency_rate` -> `_compute_amount_company`).

sale_company_currency/tests/test_sale_company_currency.py
```diff
@@ -62,13 +66,162 @@ def test_02_amount_total_curr_different_currency(self):
         self.sale_order._compute_amount_company()
         amount_total_curr_rounded = round(self.sale_order.amount_total_curr, 2)
         amount_total_converted_rounded = round(
-            self.sale_order.amount_total * self.sale_order.currency_rate, 2
+            self.sale_order.amount_total / self.sale_order.currency_rate, 2
```
alexey-pelykh, line 69: `test_02` already validates that the computed result equals `amount_total / currency_rate`. With that in place, `test_03` is largely redundant (it re-checks division with a hardcoded rate and then asserts multiplication gives a different result, which is a tautology for any rate != 1). Consider dropping `test_03` and keeping the existing `test_02` as the conversion logic test.

sale_company_currency/tests/test_sale_company_currency.py
alexey-pelykh, line 208: The `test_07` tail that resets prices but never asserts anything after the reset is dead code -- it doesn't verify any behavior. Either add assertions or remove the reset block.

fsmw:
Closing in favor of clean single-commit PR #NEW with all reviewer feedback addressed:

- **Squashed** from 7 commits to 1 clean commit
- **Forward-port clarification**: PR #3240 on 16.0 changed  to , which introduced the bug. This fix ( to ) is the correct direction per @alexey-pelykh
- ** records** instead of writing directly to computed  field
- **Trimmed tests**: kept tests 02 (different currency), 03 (rate=1), 04 (batch); dropped redundant/extreme-rate/edge-case tests
- **Fixed setUpClass** — original had both order line dicts in a single (0,0,...) tuple, silently truncating one line

PR #3962, opened 2025-10-22
[19.0][MIG] sales_team_security_sale: Migration to 19.0

lef-adhoc:
Depends on:

- [ ] <https://github.com/OCA/sale-workflow/pull/3960>

PR #4049, opened 2025-12-01
[DUMMY][MIG] sales_team_security_sale: Migration to 19.0

lef-adhoc:
The original PR is: https://github.com/OCA/sale-workflow/pull/3962


sales_team_security_sale/i18n/sales_team_security_sale.pot
```diff
@@ -0,0 +1,13 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 15.0\n"
```
alexey-pelykh, line 6: Stale version reference: `Odoo Server 15.0` should be `Odoo Server 19.0`. The `.pot` file should be regenerated for the target branch.

sales_team_security_sale/security/security.xml
```diff
@@ -0,0 +1,37 @@
...
+            name="groups"
+            eval="[(4, ref('sales_team_security.group_sale_team_manager'))]"
+        />
+    </record>
+    <record id="sale_order_line_team_rule" model="ir.rule">
+        <field name="name">Team Order Lines</field>
+        <field name="model_id" ref="sale.model_sale_order_line" />
+        <field
+            name="domain_force"
+        >['|', ('salesman_id', 'in', [user.id, False]), ('order_id.team_id', 'in', user.crm_team_ids.ids + [False])]</field>
```
alexey-pelykh, line 31: Please verify that `salesman_id` still exists on `sale.order.line` in Odoo 19.0. In recent Odoo versions, this field may have been renamed or removed. If so, the domain would need to use the updated field name (e.g., `order_id.user_id`).

alexey-pelykh (commented):
## OCA PR Review: `sales_team_security_sale` Migration to 19.0

This PR is marked as **[DUMMY]** and is a draft. The original migration PR is #3962 (still open). Posting observations as a comment for reference.

---

### Observations

#### 1. `.pot` file references Odoo Server 15.0

The translation template at `i18n/sales_team_security_sale.pot` contains:
```
"Project-Id-Version: Odoo Server 15.0\n"
```
This should be updated to `Odoo Server 19.0` to match the target branch. Likely carried over from the previous version without regeneration.

#### 2. `salesman_id` field on `sale.order.line` (verify for 19.0 compatibility)

In `security/security.xml`, the `sale_order_line_team_rule` uses:
```xml
('salesman_id', 'in', [user.id, False])
```
In recent Odoo versions, the `salesman_id` field on `sale.order.line` may have been renamed or removed. Please verify this field still exists in Odoo 19.0. If it was replaced (e.g., by `order_id.user_id`), the domain needs updating.

#### 3. Test coverage is minimal but appropriate for a glue module

The test creates a sale order and delegates to `_check_whole_permission_set(extra_checks=False)` from the base `sales_team_security` test common class. This is a reasonable approach for a glue module that only adds `ir.rule` records.

#### 4. Module structure looks correct

- `auto_install: True` is appropriate for a glue module between `sale` and `sales_team_security`.
- Version `19.0.1.0.0` follows OCA conventions.
- License headers are present.
- `noupdate="1"` on security rules is correct (rules should not be overwritten on module update).
- Empty `__init__.py` is expected since this is a data-only module (no Python models).

#### 5. Security rules domain logic

The three rules follow the standard OCA pattern:
```python
['|', ('user_id', 'in', [user.id, False]), ('team_id', 'in', user.crm_team_ids.ids + [False])]
```
This allows users to see records they own OR records belonging to their teams (including unassigned records with `False`). The logic is correct.

---

### Summary

Since this is a **DUMMY** PR, no action is needed here. The substantive review should be conducted on the original PR #3962. The two items worth checking on the real PR are:
1. Regenerate the `.pot` file for 19.0
2. Verify `salesman_id` still exists on `sale.order.line` in Odoo 19.0

PR #4056, opened 2025-12-03
[MIG] sale_substate: Migration to 19.0

np-oerp:
- sale_substate module migrated form v18 to v19.
This module is depends on base_substate module. For that, The reference PR (https://github.com/OCA/server-ux/pull/1198).

CC: @flotho @absal-smile @eugen-don @gaelTorrecillas


np-oerp:
Hi @flotho @gaelTorrecillas, 

This module (sale_substate) has been successfully migrated from Odoo v18 to v19.
Ready for code review and merge into 19.0.
Could you please review the changes when you have a moment?

Thanks in advance!

rousseldenis:
@np-oerp In order to review functionaly with runboat and in order to get green tests, you can follow this : https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

np-oerp:
Hi @rousseldenis 

I've added the necessary test-requirements.txt (referencing the base_substate PR) to the repository root. This resolve the Runboat test. Could you please check and confirm if this is correct?

Thank you.

sale_substate/tests/test_sale_substate.py
```diff
@@ -0,0 +1,75 @@
...
+                "name": "Test base substate to basic sale",
+                "partner_id": partner.id,
+                "order_line": [
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": self.product_1.id,
+                            "product_uom_qty": 2,
+                            "product_uom": self.product_1.uom_id.id,
```
rrebollo:
```suggestion
                            "product_uom_id": self.product_1.uom_id.id,
```
It looks product_uom has been renamed in 19.0

rrebollo (changes requested):
I think this should solve some of your red tests.

rrebollo:
Also your last two commits can be squashed.

np-oerp:
Hi @rrebollo 

Thank you for the feedback. I’ve applied the suggested changes and squashed the commits accordingly.
Please review and let me know if any additional input is needed from my side.

rrebollo (approved):
Code Review. LGTM!

np-oerp:
Hi @rrebollo 

Just following up on this PR. The code review has been completed—could you please let me know if any next steps are needed from my side, or if it can be merged?

Thank you.

rrebollo:
@np-oerp you need at least another approval or someone with merging capabilities to get this done. One way to get reviewers is to review other's PRs, maybe on this same repository, and request them to return the favor.

ioans73 (approved):
Functional review 👍🏻

Reyes4711-S73 (approved):
LGTM

rousseldenis:
@np-oerp Could you now remove the test-requirements change ? Thanks

np-oerp:
Hi @rousseldenis 

I have removed the test-requirements. 
Please review and let me know if any next steps are needed from my side.

Thank you.

sale_substate/__manifest__.py
```diff
@@ -3,18 +3,16 @@
...
+    "author": "Akretion, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "license": "AGPL-3",
     "depends": ["base_substate", "sale_management"],
     "data": [
         "views/sale_views.xml",
         "data/sale_substate_mail_template_data.xml",
         "data/sale_substate_data.xml",
-    ],
-    "demo": [
```
rousseldenis: @np-oerp Why changing this ?

np-oerp: Hi @rousseldenis 

When I run the pre-commit checks, I encounter an error stating that the record base_substate_under_nego is not found, which causes the checks to fail.

To address this, I moved the file reference from demo to data in the manifest.

Could you please confirm whether we should keep the same file structure as in v18 ?

rousseldenis: Moving demo data to normal data is a nonsense.

But, from v19, Odoo does not run tests with demo data (which are used to ... demo).

You should ensure tests data are created in tests and in tests only

sale_substate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "category": "Tools",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["base_substate", "sale_management"],
+    "data": [
+        "views/sale_views.xml",
+        "data/sale_substate_mail_template_data.xml",
+        "data/sale_substate_data.xml",
+        "data/sale_substate_demo.xml",
```
alexey-pelykh, line 18: This file was under `"demo"` in 18.0 and should stay there. Sample substate records ("In negotiation", "Won", "Lost", etc.) are illustrative examples, not required configuration for production databases.

Restore the 18.0 structure:
```python
    "data": [
        "views/sale_views.xml",
        "data/sale_substate_mail_template_data.xml",
        "data/sale_substate_data.xml",
    ],
    "demo": [
        "data/sale_substate_demo.xml",
    ],
```

sale_substate/tests/test_sale_substate.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2019 Akretion Mourad EL HADJ MIMOUNE
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import ValidationError
+from odoo.tests import tagged
+from odoo.tests.common import TransactionCase
+
+
+@tagged("-at_install", "post_install", "with_demo")
```
alexey-pelykh: Since Odoo 19.0 no longer loads demo data during tests by default, `@tagged("with_demo")` won't help you here. Instead of relying on `self.env.ref()` for demo XML IDs, create the substate records directly in `setUp()` (similar to how you already create `product_1` and the partner).

For example:
```python
@tagged("-at_install", "post_install")
class TestBaseSubstate(TransactionCase):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        target_draft = cls.env.ref("sale_substate.target_state_value_draft")
        target_sale = cls.env.ref("sale_substate.target_state_value_sale")
        cls.substate_under_nego = cls.env["base.substate"].create({
            "name": "In negotiation",
            "sequence": 1,
            "target_state_value_id": target_draft.id,
        })
        # ... create other substates needed by the test
```
Also consider switching from `setUp` to `setUpClass` for better performance, since the test data is shared across test methods.

alexey-pelykh (changes requested):
Thanks for the migration work, @np-oerp.

One issue still needs to be resolved before this can be merged -- the demo-to-data promotion that @rousseldenis flagged.

**Summary of the problem:**

In 18.0 the manifest correctly had `sale_substate_demo.xml` under `"demo"`. In this PR it has been moved into `"data"`, which means every production database will get sample substate records ("In negotiation", "Won", "Lost", "Waiting for legal documents", etc.) loaded as regular data. These are illustrative examples, not required configuration -- they should remain demo data.

The reason for the move was that the test class uses `@tagged("with_demo")` and fetches the records with `self.env.ref()`, and Odoo 19.0 no longer loads demo data during test runs by default. But the correct fix (as @rousseldenis noted) is to create the test data inside `setUp()` rather than promoting demo data to production data.

**What needs to change:**

1. Restore `sale_substate_demo.xml` to the `"demo"` key in `__manifest__.py`.
2. Remove the `@tagged("with_demo")` decorator (or replace with `@tagged("-at_install", "post_install")`).
3. In `setUp()`, create the substate records that the test needs instead of relying on `self.env.ref()` for demo XML IDs. You already create a partner and product inline -- do the same for the substates.

Everything else (model, views, mail templates, data XML, translations, readme) looks fine for a straightforward migration.

np-oerp:
Hi @rousseldenis 

I have updated the module with the required changes for test functions.
Please review and let me know if you need anything from my side.

Thank you.

rousseldenis (approved):

PR #4058, opened 2025-12-05
[18.0][IMP] sale_order_lot_selection: Add lot_id field to kanban view

victoralmau:
Add `lot_id` field to kanban view

**Before**
antes

**After**
despues

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT57124

### DIFF
--- a/sale_order_lot_selection/views/sale_order_views.xml
+++ b/sale_order_lot_selection/views/sale_order_views.xml
@@ -27,6 +27,23 @@
                     groups="stock.group_production_lot"
                 />
             </xpath>
+            <xpath
+                expr="//field[@name='order_line']/kanban//div[hasclass('row')]"
+                position="after"
+            >
+                <div
+                    class="text-muted"
+                    t-if="record.lot_id.value"
+                    groups="stock.group_production_lot"
+                >
+                    Lot:
+                    <field
+                    name="lot_id"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

sale_order_lot_selection/views/sale_order_views.xml
```diff
@@ -27,6 +27,20 @@
...
+            >
+                <div class="text-muted">
+                    Lot:
+                    <field
+                    name="lot_id"
+                    invisible="not product_id or product_tracking=='none'"
+                    context="{'default_product_id': product_id}"
+                    groups="stock.group_production_lot"
+                />
+                </div>
```
TadeuszKarpinski: What happens when lot_id is missing or the user doesn't have the stock.group_production_lot group? Does it render an empty "Lot:"?
It might be safer to move both groups and invisible to the `<div>` element.

Also, could you replace the hardcoded "Lot:" with a proper `<label for='lot_id'/>`?

victoralmau: Thanks, groups has been changed and applied to div.

The label you suggest does not work.

TadeuszKarpinski (approved):

PR #4061, opened 2025-12-05
[18.0][FIX] sale_elaboration: route inheritance

chienandalu:
- Get along with other modules adding automatic route mechanisms, like sale_order_type or stock_customer_deposit
- Now when we unset the elaborations, the elaboration routes are removed.

Pending:

- [ ] Add test cases for routes

cc @moduon MT-11838

### DIFF
--- a/sale_elaboration/models/sale_order.py
+++ b/sale_elaboration/models/sale_order.py
@@ -70,8 +70,17 @@ class SaleOrderLine(models.Model):
 
     @api.depends("elaboration_ids")
     def _compute_route_id(self):
-        for line in self:
+        # Reset routes: elaboration might be unset.
+        self.filtered(
+            lambda x: x.route_id not in x.elaboration_ids.route_ids
+        ).route_id = False
+        # Other modules might use this compute. Let's respect inheritance
+        res = None
+        if hasattr(super(), "_compute_route_id"):
+            res = super()._compute_route_id()
+        for line in self.filtered("elaboration_ids"):
             line.route_id = line.get_elaboration_stock_route()
+        return res
 
     @api.depends("elaboration_ids", "order_id.pricelist_id")
     def _compute_elaboration_price_unit(self):


rousseldenis:
IMHO a glue module should be more appropriate

chienandalu:
> IMHO a glue module should be more appropriate

Not exactly. There are several modules that use that declare that method for their own purpuses. The thing is to let each one do their thing. We want't avoid that the last one to come squash the declaration.

Additionally, glue modules might be needed, but the main thing is to try to stablish a chain of inheritance. See that `sale_order_type` already does it for this same method, so if we want them to coexist, this is necessary:

https://github.com/OCA/sale-workflow/blob/2b32421b8d3db119c9f6adc11e46cb8507ef799e/sale_order_type/models/sale.py#L192-L203

yajo (approved):

rrebollo (commented):
Code Review. LGTM! Didactic.
📋 Strong Recommendation
I would strongly recommend implementing the tests mentioned in your PR description before merging.

sale_elaboration/models/sale_order.py
```diff
@@ -76,10 +76,19 @@ def get_elaboration_stock_route(self):
 
     @api.depends("elaboration_ids")
     def _compute_route_id(self):
-        for line in self:
+        # Reset routes: elaboration might be unset.
+        self.filtered(
+            lambda x: x.route_id not in x.elaboration_ids.route_ids
```
alexey-pelykh, line 75: This will also reset `route_id` on lines with no elaborations but a route set by another module (e.g. `sale_order_type`), because `route_id not in empty_recordset` is always `True` for a truthy route. Consider scoping to lines that actually have elaborations:
```python
self.filtered(
    lambda x: x.elaboration_ids and x.route_id not in x.elaboration_ids.route_ids
).route_id = False
```

sale_elaboration/models/sale_order.py
```diff
@@ -76,10 +76,19 @@ def get_elaboration_stock_route(self):
...
+        ).route_id = False
+        # Other modules might use this compute. Let's respect inheritance
+        res = None
```
alexey-pelykh, line 78: `hasattr(super(), "_compute_route_id")` is unusual in Odoo. If no other module adds this compute, `super()` resolves to the same class in the MRO chain, so it would find *this* method. Calling `super()._compute_route_id()` unconditionally is the standard Odoo pattern and avoids the fragile introspection.

sale_elaboration/models/sale_order.py
```diff
@@ -76,10 +76,19 @@ def get_elaboration_stock_route(self):
...
+        if hasattr(super(), "_compute_route_id"):
+            res = super()._compute_route_id()
+        for line in self.filtered("elaboration_ids"):
             route_id = line.get_elaboration_stock_route()
             if route_id:
                 line.route_id = route_id
+        return res
```
alexey-pelykh, line 83: Compute methods in Odoo don't use return values -- the framework ignores whatever is returned. This `return res` is dead code. I'd remove it for clarity.

alexey-pelykh (commented):
Thanks for tackling the route inheritance cooperation, David. The core idea is sound -- calling super and resetting stale routes is the right direction for playing nicely with other modules.

A few things I noticed:

**Route reset may be too broad** -- The filter `x.route_id not in x.elaboration_ids.route_ids` will also match lines that have *no* elaborations but *do* have a route set by another mechanism (since empty `route_ids` means any truthy `route_id` passes the `not in` check). If a line has a route from `sale_order_type` but no elaborations, this resets it to `False` before `super()` gets a chance to restore it. It works only if super cooperatively re-sets all routes on every recompute. Safer would be to scope the reset to lines that *previously had* an elaboration route, e.g.:

```python
self.filtered(
    lambda x: x.elaboration_ids and x.route_id not in x.elaboration_ids.route_ids
).route_id = False
```

or even track which routes were set by this module.

**`hasattr(super(), "_compute_route_id")` is fragile** -- In Odoo's inheritance chain, `super()` always resolves through the MRO. If no other module adds `_compute_route_id`, the method on `super()` is *this same method* (since `sale.order.line` inherits it from the `_inherit`). `hasattr` could also mask unexpected exceptions. A more conventional Odoo pattern would be to just call `super()._compute_route_id()` unconditionally -- if no other module overrides it, it simply calls itself (which is a no-op since we'd be in the same method). Or if the intent is specifically to detect *other* overrides, that would need a different approach.

**Return value from compute method** -- Compute methods in Odoo are not expected to return a value (the framework ignores it). Storing and returning `res` from `super()` is harmless but adds dead code. Unless there's a specific downstream consumer, I'd drop the `return res`.

**Missing tests (codecov failing)** -- The PR description already flags this and I see the codecov checks are red. Agree with @rrebollo that tests for route assignment/removal would solidify the fix, especially given the edge cases around cooperative inheritance.

carlosdauden:
Picking up on @alexey-pelykh's review about the route reset being too broad.

His suggested guard (`x.elaboration_ids and x.route_id not in x.elaboration_ids.route_ids`) avoids resetting a route that came from another mechanism, but it swaps in the opposite regression: once the elaboration is fully removed (`elaboration_ids` becomes empty), that condition never matches, so a previously elaboration-set route is never cleared again - which is exactly the second goal stated in this PR's description.

Since what we actually need is whether the *previous* value of `route_id` came from the *previous* elaborations (not the current, possibly-empty one), `_origin` gives us that without any extra field:

```python
self.filtered(
    lambda x: x.route_id in x._origin.elaboration_ids.route_ids
    and x.route_id not in x.elaboration_ids.route_ids
).route_id = False
```

This only clears a route that this line's own elaboration actually used to offer and no longer does, and never touches a route coming from `sale_order_type` or anywhere else in the chain.

On `hasattr(super(), "_compute_route_id")`: I'd keep it rather than call `super()._compute_route_id()` unconditionally. If nothing else in the MRO defines the method, `super().foo()` raises `AttributeError` - it doesn't fall back to calling itself - so the guard is what makes this safe when `sale_elaboration` is installed without another module extending the same compute.

Agree on dropping the dead `return res`, and on adding test coverage for both the reset-on-unset case and the cooperation with an externally-set route (e.g. from `sale_order_type`) before merging.

rafaelbn:
@chienandalu please review https://github.com/OCA/sale-workflow/pull/4061#issuecomment-5633684153

PR #4064, opened 2025-12-10
[18.0][ADD] sale_order_line_base_price_and_has_discount

rousseldenis:
image


rrebollo:
@rousseldenis given the picture in PR description, is the design/inspiration for this feature influenced by or related to the EE?

sale_order_line_base_price_and_has_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,30 @@
...
+            line.has_discount_price = bool(pricelist_price < base_price)
+        (self - lines_with_product).has_discount_price = False
+
+    @api.depends("product_id")
+    def _compute_base_price(self):
+        lines_with_product = self.filtered("product_id")
+        for line in lines_with_product:
+            base_price = line._get_pricelist_price_before_discount()
+            line.base_price = base_price
+        (self - lines_with_product).base_price = False
```
rrebollo: Did you consider compute both fields in a single method? I'm not sure it would be a good idea due to the differences in the re computation triggers but ... I'm just asking

rrebollo (approved):
Code Review. LGTM!

sale_order_line_base_price_and_has_discount/models/sale_order_line.py
rousseldenis: Yes, I will. Thanks.

I will also store them as they should remain even if pricelist prices change with time.

sale_order_line_base_price_and_has_discount/models/product_pricelist_item.py
```diff
@@ -0,0 +1,29 @@
...
+    def _get_compute_price_methods_for_base_price_pricelist(self):
+        return ["percentage", "formula"]
+
+    def _compute_price_before_pricelist(self, *args, **kwargs):
+        """Compute the base price of the lowest pricelist rule,
+        taking into account percentage and formula items
+        """
+        pricelist_item = self
+        methods = self._get_compute_price_methods_for_base_price_pricelist()
+        while pricelist_item.base == "pricelist":
```
alexey-pelykh, line 19: This `while` loop walks chained pricelists but has no guard against circular references (A -> B -> A). If a user misconfigures two pricelists referencing each other, this will loop forever.

Consider something like:
```python
seen = set()
while pricelist_item.base == "pricelist":
    if pricelist_item.id in seen:
        break
    seen.add(pricelist_item.id)
    ...
```

sale_order_line_base_price_and_has_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,81 @@
...
+        (self - lines_with_product).base_price_discount = 0.0
+
+    @api.depends("price_unit", "base_price")
+    def _compute_has_discount_price(self):
+        """
+        Computing the boolean field based on :
+
+            - base price
+            - price unit
+        """
```
alexey-pelykh, line 43: Since both code paths ultimately read `self.pricelist_item_id` (either directly via `_get_pricelist_price_before_discount` or via the new `_compute_price_before_pricelist`), I think `pricelist_item_id` should be in the `@api.depends` here. Without it, `base_price` (which is stored) won't recompute when the applicable pricelist rule changes (e.g. due to a quantity break crossing a threshold).

sale_order_line_base_price_and_has_discount/views/sale_order.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/list/field[@name='price_unit']"
+                position="before"
+            >
+                <field
+                    name="base_price"
+                    class="text-decoration-line-through"
+                    invisible="not has_discount_price"
```
alexey-pelykh, line 16: nit: `has_discount_price` is used in `invisible` but not declared as a field in this list view. Might be worth adding `<field name="has_discount_price" column_invisible="True"/>` for explicitness and forward-compatibility.

sale_order_line_base_price_and_has_discount/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,9 @@
+This module allows to display the original price before pricelist application.
+This can be completed by the 'sale_report_crossed_out_original_price' module to
+display crossed prices on sale order report.
+
+In some cases, prices are based on several pricelists based on formulas and 
+we want to display the base price in that case too. This beahaviour is
```
alexey-pelykh, line 6: Typo: "beahaviour" -> "behaviour"

sale_order_line_base_price_and_has_discount/tests/test_sale_line_price.py
```diff
@@ -0,0 +1,103 @@
...
+                ],
+            }
+        )
+        self.assertEqual(10.0, order.order_line.base_price)
+        self.assertEqual(
+            9.0,
+            order.order_line.price_unit,
+        )
+        self.assertTrue(order.order_line.has_discount_price)
+        self.assertEqual(order.order_line.base_price_discount, 10.0)
```
alexey-pelykh, line 79: nit: "dubble" -> "double"

alexey-pelykh (commented):
Thanks for this module, Denis. The base price / has-discount concept is genuinely useful for B2B quotations.

I have a few observations on the current state of the code (reviewed at HEAD `1ed8b61`).

---

### Potential infinite loop in `_compute_price_before_pricelist`

`product_pricelist_item.py` walks chained pricelists with `while pricelist_item.base == "pricelist"`. If two pricelists reference each other (A -> B -> A), this loops forever. Odoo core's `_compute_price_before_discount` has the same kind of traversal but it guards against circular references. Consider adding a `seen` set or a max-depth counter to bail out safely.

### `_compute_base_price` depends on `price_unit` but not on `pricelist_item_id`

The compute depends on `("price_unit", "product_id")`, and the body branches on `company_id.display_base_price_method`. The "discount" path calls `_get_pricelist_price_before_discount()` which internally reads `self.pricelist_item_id`. If `pricelist_item_id` changes (e.g. quantity break), `base_price` won't recompute because it is not in `@api.depends`. Since `base_price` is stored, stale values would persist. I'd suggest adding `pricelist_item_id` to the depends list.

### `has_discount_price` field is not declared in the tree view

In `sale_order.xml`, `has_discount_price` is referenced in `invisible="not has_discount_price"` but is never added as a (hidden) field to the list. Odoo 18 list views require fields used in domain/invisible expressions to be present in the view. This may already work due to the compute engine making it available, but it would be safer and more explicit to add `<field name="has_discount_price" column_invisible="True"/>` to avoid issues with future Odoo versions or edge cases.

### Missing `security/ir.model.access.csv`

No new models are introduced (only inheritance), so technically not required. Just confirming this is intentional -- looks fine.

### Minor: typo in description

`readme/DESCRIPTION.md` has "beahaviour" -- should be "behaviour".

### Minor: test comment typo

Line 79 of `test_sale_line_price.py`: `# Check the dubble discount` -- should be "double".

### Test coverage

Tests cover the two main paths (discount-only vs discount+formula) well. It could be worth adding a case for a product with no pricelist rule (to verify `base_price` and `has_discount_price` are both falsy), but that's optional.

---

Overall the module is well-structured and follows OCA conventions. The `pre_init_hook` for the stored field, the configurable company setting, and the pricelist chain traversal are all solid design choices. The two items I'd flag for attention before merge are the **circular pricelist guard** and the **missing `pricelist_item_id` dependency**.

PR #4065, opened 2025-12-10
[18.0][MIG] sale_pricelist_global_rule: Migration to 18.0

DucTruongKomit:
sale_pricelist_global_rule: Migration to 18.0


rousseldenis:
@DucTruongKomit Thanks for this.

Could you improve your commit message removing duplicate lines:

image

sale_pricelist_global_rule/tests/test_pricelist_ancestor.py
```diff
@@ -99,25 +99,25 @@ def _mk_item(
             "PI0",
             product=cls.prod_a,
         )
-        # PI1: Ancestor Cat C -> 20% (min 5)
+        # PI1: Ancestor Cat C -> 20% (min 3)
```
rousseldenis: @DucTruongKomit Why changing the use case ??

DucTruongKomit: > @DucTruongKomit Why changing the use case ??

Perhaps I typed wrong, I will revert this.

DucTruongKomit:
> @DucTruongKomit Thanks for this.
> 
> Could you improve your commit message removing duplicate lines:
> 

Should be ok now.

sale_pricelist_global_rule/tests/test_pricelist_ancestor.py
DucTruongKomit: should be ok now.

sale_pricelist_global_rule/models/product_pricelist.py
```diff
@@ -5,19 +5,23 @@
 class ProductPricelistItem(models.Model):
     _inherit = "product.pricelist.item"
 
-    applied_on = fields.Selection(
-        selection_add=[
+    display_applied_on = fields.Selection(
+        selection_add=[("3_global", "Global")], ondelete={"3_global": "set default"}
```
cuongnmtm: Is it safe to convert "Global" to the default "Product"?

Should we use cascade?

sale_pricelist_global_rule/models/product_pricelist.py
```diff
@@ -0,0 +1,290 @@
...
+                for k, v in by_categ_raw.items()
+            }
+            # Sum total quantity across the whole ancestor branch
+            total_qty = sum(by_categ_id.get(cid, 0.0) for cid in child_categories.ids)
+            # Check minimum quantity threshold over the branch total
+            if self.min_quantity and total_qty < self.min_quantity:
+                is_applicable = False
+
+        if not is_applicable:
+            return False
```
alexey-pelykh, line 245: **Bug: unreachable code after `return False`**

Lines 246–257 (the comment block and the `_has_better_percentage_rule` call) will never execute because `return False` on line 245 exits the method unconditionally when `is_applicable` is `False`.

It looks like the intent was:
```python
if not is_applicable:
    return False
# Ensure current price item has best discount
if self._has_better_percentage_rule(product, qty_in_product_uom):
    is_applicable = False
return is_applicable
```

But the current indentation/placement means the best-percentage logic is dead code. Since the tests do pass with correct discount selection, the logic might be working via a different path (Odoo's own rule ordering), but this block is definitely not contributing. Please either fix the control flow so `_has_better_percentage_rule` is actually evaluated, or remove the dead block if it's not needed.

sale_pricelist_global_rule/static/description/index.html
```diff
@@ -0,0 +1,531 @@
...
+<<<<<<< HEAD
+<p><a class="reference external image-reference" href="https://odoo-community.org/page/development-status"><img alt="Beta" src="https://img.shields.io/badge/maturity-Beta-yellow.png" /></a> <a class="reference external image-reference" href="http://www.gnu.org/licenses/agpl-3.0-standalone.html"><img alt="License: AGPL-3" src="https://img.shields.io/badge/licence-AGPL--3-blue.png" /></a> <a class="reference external image-reference" href="https://github.com/OCA/sale-workflow/tree/17.0/sale_pricelist_global_rule"><img alt="OCA/sale-workflow" src="https://img.shields.io/badge/github-OCA%2Fsale--workflow-lightgray.png?logo=github" /></a> <a class="reference external image-reference" href="https://translation.odoo-community.org/projects/sale-workflow-17-0/sale-workflow-17-0-sale_pricelist_global_rule"><img alt="Translate me on Weblate" src="https://img.shields.io/badge/weblate-Translate%20me-F47D42.png" /></a> <a class="reference external image-reference" href="https://runboat.odoo-community.org/builds?repo=OCA/sale-workflow&amp;target_branch=17.0"><img alt="Try me on Runboat" src="https://img.shields.io/badge/runboat-Try%20me-875A7B.png" /></a></p>
```
alexey-pelykh, line 373: **Merge conflict markers left in the generated HTML**

This file contains unresolved git conflict markers (`<<<<<<< HEAD`, `=======`, `>>>>>>> ee935fa27`). This is a generated file, so it needs to be regenerated cleanly with `oca-gen-addon-readme`. Please re-run the tool and ensure no conflict remnants remain.

sale_pricelist_global_rule/i18n/sale_pricelist_global_rule.pot
```diff
@@ -0,0 +1,104 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* sale_pricelist_global_rule
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 17.0\n"
```
alexey-pelykh, line 7: **Stale `.pot` file – missing ancestor category strings**

The `.pot` still references `Odoo Server 17.0` and is missing all strings introduced by the ancestor category feature:
- `"Ancestor product category: %s"`
- `"Please specify the product ancestor category for which this global rule should be applied"`
- The `3_3_global_product_ancestor_category` selection label
- The `ancestor_product_category_id` field description

Same applies to `it.po`. Please regenerate the `.pot` file (e.g. via `./odoo-bin --modules=sale_pricelist_global_rule -d <db> --i18n-export=...`) and update translations accordingly.

sale_pricelist_global_rule/models/product_pricelist.py
```diff
@@ -0,0 +1,290 @@
...
+                is_applicable = False
+        elif self.global_applied_on == "3_3_global_product_ancestor_category":
+            ancestor_categ = self.ancestor_product_category_id
+            if not ancestor_categ:
+                return False
+
+            Category = self.env["product.category"]
+
+            # ancestor + all descendants (fast, uses parent_path internally)
+            child_categories = Category.search_fetch(
```
alexey-pelykh, line 220: **Performance consideration: `search_fetch` inside `_is_applicable_for`**

This `search_fetch` call runs a DB query for every `(rule, product)` pair in the ancestor category path. For a sale order with many lines and multiple ancestor rules, this could result in a significant number of queries.

Consider caching the child category set per ancestor (e.g. in the context dict alongside `pricelist_global_cummulative_quantity`) so it's computed once per ancestor per recompute cycle, rather than once per rule+product evaluation.

sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -0,0 +1,73 @@
...
+            key_categ = line.product_id.categ_id
+            qty_data["by_template"].setdefault(key_template, 0.0)
+            qty_data["by_template"][key_template] += qty_in_product_uom
+            qty_data["by_categ"].setdefault(key_categ, 0.0)
+            qty_data["by_categ"][key_categ] += qty_in_product_uom
+        return qty_data
+
+    @api.depends("pricelist_id")
+    def _compute_has_pricelist_global(self):
+        for sale in self:
```
alexey-pelykh, line 40: Minor: `_get_cummulative_quantity` – the standard English spelling is "cumulative" (single 'm'). Same for the context key `pricelist_global_cummulative_quantity`. Not blocking, but worth fixing for consistency if you're touching the code anyway (this was inherited from the 16.0 version, so up to you).

alexey-pelykh (changes requested):
Thanks for the migration work @DucTruongKomit.

I found a few issues that should be addressed before merge. The overall structure is solid and tests are comprehensive, but there's a bug in `_is_applicable_for` and the generated files need regenerating.

See inline comments below.

sale_pricelist_global_rule/static/description/index.html
rousseldenis: @DucTruongKomit

sale_pricelist_global_rule/tests/test_pricelist_ancestor.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+        cls.cat_c = env["product.category"].create(
+            {"name": "Cat C", "parent_id": cls.cat_a.id}
+        )
+        cls.cat_d = env["product.category"].create(
+            {"name": "Cat D", "parent_id": cls.cat_c.id}
+        )
+
+        # --- Products (list_price=100 for easy math)
+        def _mk(name, categ):
```
rousseldenis, line 32: These kind of inline functions are not wanted. Moreover they should be named with a human readable name.

hieulucky111:
@DucTruongKomit 
We are facing this error when open the Sale Order with the Section / Note line
```
File "/opt/odoo/odoo/oca-sale-workflow/sale_pricelist_global_rule/models/sale_order.py", line 50, in _compute_has_pricelist_global
    suitable_rule = pricelist._get_product_rule(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/odoo/odoo/addons/product/models/product_pricelist.py", line 159, in _get_product_rule
    return self._compute_price_rule(product, *args, compute_price=False, **kwargs)[product.id][1]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: False
```

rousseldenis:
@DucTruongKomit

PR #4066, opened 2025-12-11
[19.0][IMP] sale_require_po_doc-default: Set default 'Customer Requires PO' at company level

AaronHForgeFlow:
For some companies this requirement is for all customers, so this setting allow you to do that without changing current behavior

### DIFF
--- a/sale_require_po_doc/models/res_partner.py
+++ b/sale_require_po_doc/models/res_partner.py
@@ -10,9 +10,13 @@ class ResPartner(models.Model):
     customer_need_po = fields.Boolean(
         string="Customer Requires PO",
         tracking=True,
+        default=lambda self: self.env.company.customer_need_po_default
+        if self.env.context.get("default_customer_rank", 0) > 0
+        else False,
         help="A PO number will be required on the"
         " Sales Order field for Customer Reference",
     )
+
     sale_doc = fields.Text(
         string="Sales Require Documentation",
         tracking=True,
--- a/sale_require_po_doc/models/res_partner.py
+++ b/sale_require_po_doc/models/res_partner.py
@@ -10,6 +10,7 @@ class ResPartner(models.Model):
     customer_need_po = fields.Boolean(
         string="Customer Requires PO",
         tracking=True,
+        default=lambda self: self.env.company.customer_need_po_default,
         help="A PO number will be required on the"
         " Sales Order field for Customer Reference",
     )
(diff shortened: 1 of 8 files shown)


rousseldenis (approved):

rrebollo (commented):
The implementation looks solid from a technical perspective. Code structure, logic, and style are well-executed.

Consideration for improvement: Would you be open to adding a basic test case to validate the new approach?

rrebollo:
If you happen to implement the test ping me. I will give you my approve. Also, would you be so kind to review my work (#4072) back?

AaronHForgeFlow:
Thank you @rrebollo I added a small test. 

Your PR https://github.com/OCA/sale-workflow/pull/4072/ is already approved but I can review any other PR

rrebollo (approved):
Code Review. LGTM!

alexey-pelykh (commented):
Thanks for this addition -- having a company-level default for "Customer Requires PO" is a practical improvement.

Overall the implementation follows the standard `res.company` / `res.config.settings` / field default pattern correctly. Tests cover both the enabled and disabled cases, CI is green. A couple of things to address:

**1. Incorrect help text in the settings view**

In `views/res_config_settings.xml`, the `<setting>` help text reads:

> Check this if you want the sale order type required on sale order forms.

This doesn't match the feature. It should describe the PO requirement default, something like:

> If checked, new customers will have 'Customer Requires PO' enabled by default.

**2. Missing license headers on new Python files**

`models/res_company.py` and `models/res_config_settings.py` are missing the AGPL-3 copyright header that the rest of the module uses. Every other `.py` file starts with:

```python
# Copyright (C) 2021 Open Source Integrators
```

Please add appropriate headers (with your own copyright line or ForgeFlow's).

Both are small fixes. The rest looks good to me.

sale_require_po_doc/models/res_partner.py
```diff
@@ -10,6 +10,7 @@ class ResPartner(models.Model):
     customer_need_po = fields.Boolean(
         string="Customer Requires PO",
         tracking=True,
+        default=lambda self: self.env.company.customer_need_po_default,
```
BhaveshHeliconia: We should set default value only for customers, not for all contacts.

AaronHForgeFlow: good point. I considered it.

AaronHForgeFlow:
@alexey-pelykh I applied changes upon your comments, can you update your review?

BhaveshHeliconia (approved):

PR #4078, opened 2025-12-18
[18.0][ADD] sales_team_department

Saran440:
This module adds a feature to automatically add members from a department to a sales team.


rousseldenis (approved):
LGTM

sales_team_department/models/crm_team.py
```diff
@@ -0,0 +1,21 @@
...
+        comodel_name="hr.department",
+        string="Departments",
+    )
+
+    @api.depends("crm_team_member_ids.active", "department_ids")
+    def _compute_member_ids(self):
+        res = super()._compute_member_ids()
+        for team in self:
+            if team.department_ids:
+                team.member_ids = team.department_ids.mapped("member_ids.user_id")
```
alexey-pelykh, line 20: This line **replaces** `member_ids` with only the department users instead of merging. If the team also has individually-added members they'll be wiped out.

Consider:
```python
team.member_ids |= team.department_ids.mapped("member_ids.user_id")
```

sales_team_department/models/crm_team.py
```diff
@@ -0,0 +1,21 @@
...
+
+class CrmTeam(models.Model):
+    _inherit = "crm.team"
+
+    department_ids = fields.Many2many(
+        comodel_name="hr.department",
+        string="Departments",
+    )
+
+    @api.depends("crm_team_member_ids.active", "department_ids")
```
alexey-pelykh, line 15: The depends is missing the deeper relation -- when an employee joins/leaves a department, this won't recompute. Something like:
```python
@api.depends("crm_team_member_ids.active", "department_ids", "department_ids.member_ids.user_id")
```
would ensure the field refreshes when department membership changes.

sales_team_department/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+
+{
+    "name": "Sales Team Department",
+    "summary": "Add member from Department to Sales Team",
+    "version": "18.0.1.0.0",
+    "category": "Tools",
+    "author": "Ecosoft, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "AGPL-3",
+    "depends": ["sales_team", "hr"],
```
alexey-pelykh, line 12: Minor: there's no `security/ir.model.access.csv` shipped. While the module doesn't introduce a new model, OCA linters usually expect the file and it's good practice to have one confirming access rules are intentionally inherited.

alexey-pelykh (changes requested):
Thanks for this contribution, @Saran440 -- the idea of linking departments to sales teams is useful.

I do have a few concerns about the current implementation that I think need to be addressed before merging.

**`_compute_member_ids` replaces instead of extending**

When `department_ids` is set the override does `team.member_ids = team.department_ids.mapped("member_ids.user_id")`, which **replaces** whatever the `super()` call (and any manually-added members) just set. If a team has departments *and* individually-added members, the individual ones are silently dropped. The assignment should use `|=` (union) to merge department users into the existing set, not overwrite it.

**Missing `department_ids.member_ids` in `@api.depends`**

The compute depends on `department_ids` but not on the employees that belong to those departments. If someone joins or leaves a department, the computed field won't recompute because the ORM doesn't know about the deeper relation. Adding `department_ids.member_ids` (or even `department_ids.member_ids.user_id`) to the depends would fix this.

**No tests**

The codecov checks are failing because there is no test coverage at all. Even a minimal test that creates a department with employees, links it to a crm.team, and asserts the resulting `member_ids` would cover the happy path and protect against regressions.

**No `ir.model.access` / security rules**

The module adds a `Many2many` on `crm.team` backed by a new relation table, but there is no `security/ir.model.access.csv`. Depending on how Odoo handles inherited models this may or may not cause an access-rights warning, but it's good practice to ship one (even if the model itself doesn't need new ACLs, the linter usually expects the file).

**Minor: `_compute_member_ids` return value**

The base `_compute_member_ids` doesn't return anything meaningful (compute methods in Odoo set values on records, they don't return). Capturing and returning `res` is harmless but unnecessary -- you can just call `super()` without assigning.

Would love to see these addressed. Happy to re-review once updated.

PR #3339, opened 2024-10-01
[16.0][MIG] sale_mrp_link: Migration to 16.0 (from 12.0)

legalsylvain:

PR #4080, opened 2025-12-22
[18.0][MIG] sale_mrp_link: Migration to 18.0

ypapouin:
Migration from 16.0 PR https://github.com/OCA/sale-workflow/pull/3339

We still use this module even though it is somewhat redundant with `mrp_sale_info`.
It would probably be interesting to move the sale order view from `sale_mrp_link` to `mrp_sale_info`.


sale_mrp_link/models/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for values in vals_list:
+            if "origin" in values:
+                # Checking first if this comes from a 'sale.order'
+                sale_id = self.env["sale.order"].browse(values.get("sale_order_id"))
+                if not sale_id:
+                    sale_id = self.env["sale.order"].search(
+                        [("name", "=", values["origin"])], limit=1
```
rousseldenis, line 23: I'm not fond of such links on strings...

alexey-pelykh (approved):
Thanks for the migration, looks good overall.

The `create` override is properly updated to `@api.model_create_multi`, views use the 18.0-style `invisible` expression, manifest and dependencies are correct, and tests cover both single-MO and multi-MO scenarios. CI is green across the board.

One small thing worth noting: in `mrp_production.py`, the chained MO search that resolves `sale_order_id` from a parent production doesn't use `limit=1`. If multiple MOs somehow share the same name (edge case, but possible with sequences reset or manual edits), `production_id.sale_order_id.id` would raise on the singleton check. This is inherited from the 12.0 source though, so not a blocker for this migration -- just something to be aware of for a future improvement.

The roadmap links still point to `10.0` branches, but again that's pre-existing.

hbrunn (commented):
should the link just be done via the procurement group?

hbrunn (commented):
actually, I think now this is obsoleted by sale_mrp

flotho:
Hi @ypapouin 

As @hbrunn wrote, I also think that this PR should be closed as Odoo imporved the links between objects.
If you still need such a module, please provide clear improvements and dependency to the core.
Regards

ypapouin:
Alright, not a problem.

PR #4082, opened 2025-12-22
[IMP] sale_order_type: fr translation

dreispt:


HaraldPanten (changes requested):
@dreispt Why don't you use Weblate? 🤔

dreispt:
@HaraldPanten Sometimes it's much faster to edit a text file, especially when you editor can suggest edits.

pedrobaeza:
@dreispt you can upload .po files to Weblate.

HaraldPanten:
But you're just updating 4 or 5 terms, and you can import files, as well in Weblate.

rousseldenis:
@dreispt Do you plan to use weblate ?

alexey-pelykh (changes requested):
Thanks for filling in these missing French translations, Daniel -- they look accurate.

However, I agree with the other reviewers: these should go through [Weblate](https://translation.odoo-community.org/) rather than direct `.po` file edits. Committing translations directly to the repo bypasses Weblate's sync and can cause conflicts when Weblate next pushes its snapshot.

You can upload a `.po` file to Weblate if you prefer working with a text editor -- that way the translations land through the proper channel.

Please submit these via Weblate and close this PR.

dreispt:
Yes, I will do so. Thank you Alexey.

PR #4083, opened 2025-12-22
[FIX] sale_order_type: Smart Button click on existing SOs

dreispt:
After installing sale_order_type, on existing confirmed orders
it is not possible to use the smart buttons to navigate to the
related deliveries and invoices.

### DIFF
--- a/sale_order_type/views/sale_order_view.xml
+++ b/sale_order_type/views/sale_order_view.xml
@@ -7,7 +7,7 @@
             <field name="currency_id" position="after">
                 <field
                     name="type_id"
-                    required="1"
+                    required="state in ['draft', 'sent']"
                     readonly="state in ['sale', 'cancel'] or locked"
                 />
             </field>


rvalyi (approved):

jbaudoux (approved):
Why should this field be required at all?

Why all SO needs to be typed once this module is installed? e-commerce and edi sales won't be typed. So why is this field not just optional acting like an helper if the user want to.

rousseldenis:
@dreispt @jbaudoux Indeed, I asked myself that question...

alexey-pelykh (approved):
Thanks for the fix, Daniel.

The change is correct and well-scoped. Making `type_id` required only in draft/sent states avoids the validation error on existing confirmed orders that have no type assigned (pre-installation scenario). The existing Python compute methods already guard with `self.filtered("type_id")`, so empty values on confirmed SOs won't cause downstream issues.

I note @jbaudoux's question about whether the field should be required at all -- that's a valid design discussion for a follow-up, but this PR correctly fixes the immediate regression without changing the module's intended behavior for new orders.

rousseldenis (approved):

PR #4086, opened 2025-12-23
[18.0][MIG] sale_order_safe_commitment_date: Migration to 18.0

chienandalu:
- Now product lead times are integers, so we better deal with whole days instead.
- Added a sales cutt-off calendar so we can avoid salesmen scheduling deliveries that won't reach on time.

https://www.loom.com/share/3f582ee0d01b4130ad0d61db275b254f

cc @moduon MT-12188


rousseldenis:
@chienandalu Could you check tests ?

alexey-pelykh (approved):
Nice module! The cut-off calendar approach using `resource.calendar` is smart — it leverages existing Odoo infrastructure rather than reinventing scheduling logic.

The `_inverse_date_for_commitment` setting time to 00:00 of the day in user TZ and converting to UTC is the correct pattern for date-to-datetime fields. Tests with `freezegun` properly validate the cut-off scenarios.

Code review LGTM

tonygalmiche (approved):
Tested on a local Odoo 18.0 instance:
- Sales cut-off calendar correctly pushes the expected delivery date based on working days/hours.
- The warning banner and popup correctly trigger when the requested delivery date is earlier than the computed minimum.
- The delivery date is correctly auto-corrected to the minimum expected date on order confirmation.

Works as expected.

BhaveshHeliconia (approved):
LGTM!

PR #4087, opened 2026-01-02
[FIX] sale_invoice_blocking: return text in _nothing_to_invoice_error_message

dhx:
Return error message directly instead of a UserError to prevent an error when this module is used together with the enterprise subscription module which tries to append its own message:

result = _call_kw_multi(method, model, args, kwargs)
File "/opt/odoo/src/odoo/odoo/api.py", line 469, in _call_kw_multi
File "/opt/odoo/src/enterprise/sale_subscription/models/sale_order.py", line 731, in action_invoice_subscription
raise UserError(self._nothing_to_invoice_error_message())
File "/opt/odoo/src/enterprise/sale_subscription/models/sale_order.py", line 1534, in _nothing_to_invoice_error_message
error_message += _(
TypeError: unsupported operand type(s) for +=: 'UserError' and 'str'

### DIFF
--- a/sale_invoice_blocking/models/sale_order.py
+++ b/sale_invoice_blocking/models/sale_order.py
@@ -1,7 +1,6 @@
 # Copyright 2021 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from odoo import _, api, fields, models
-from odoo.exceptions import UserError
 
 
 class SaleOrder(models.Model):
@@ -33,4 +32,4 @@ class SaleOrder(models.Model):
         error = super()._nothing_to_invoice_error_message()
         error += _("\n- You may have an invoice blocking reason on the sale order")
 
-        return UserError(error)
+        return error


sale_invoice_blocking/models/sale_order.py
```diff
@@ -33,4 +33,4 @@ def _nothing_to_invoice_error_message(self):
         error = super()._nothing_to_invoice_error_message()
         error += _("\n- You may have an invoice blocking reason on the sale order")
 
-        return UserError(error)
```
rousseldenis, line 36: @damdam-s The fix seems legit as the error is raised later. Do you know why you did this ?

alexey-pelykh (commented):
Thanks for this fix -- the traceback in the description makes the problem very clear.

The change itself is correct: upstream `_create_invoices` already does `raise UserError(self._nothing_to_invoice_error_message())`, so this method's contract is to return a plain string, not a `UserError` instance. Returning a `UserError` breaks any downstream code that tries to concatenate onto the result (like enterprise `sale_subscription` does).

One thing to address before this can be merged: removing `return UserError(error)` makes the `from odoo.exceptions import UserError` import on line 4 unused, which is why the pre-commit / autoflake check is failing. Please drop that import line and the CI should go green.

dhx:
@alexey-pelykh thanks a lot for the review! Rebased and removed the unneeded import as suggested.
@rousseldenis does this also look good to you?

alexey-pelykh (approved):
Rebase and import cleanup look good -- CI is green. Thanks for the quick turnaround!

PR #3755, opened 2025-06-12
[FIX] sale_stock_delivery_address: prevent overriding procurement group key without dest_address_id

RicardCForgeFlow:
PR created from the discussion at https://github.com/OCA/sale-workflow/pull/3462#discussion_r2140582884.
Avoid overriding procurement group key when `dest_address_id` is not set. This ensures correct priority handling and avoids interfering with other modules' grouping logic.
image

PR #4090, opened 2026-01-05
[18.0][IMP] sale_stock_delivery_address: Specific test for 'get_procurement_group_key'

RicardCForgeFlow:
Test added to validate the FIX: https://github.com/OCA/sale-workflow/pull/3755


StefanRijnhart (approved):
Thanks for the follow-up 😄

StefanRijnhart (changes requested):
Wait, the test fails 😮 mouth:

RicardCForgeFlow:
> Wait, the test fails 😮 mouth:

The problem was that with the `sale_sourced_by_line` module, that sets the priority to 10

We can do `parent_key = super(type(self.line_1), self.line_1)._get_procurement_group_key()`

To ensure that when no `dest_address_id` the key does not change: `self.assertEqual(key, parent_key)`

StefanRijnhart (approved):
Alright!

rousseldenis (approved):

PR #4091, opened 2026-01-06
[18.0][IMP] sale_order_split_strategy: Allow not posting messages

grindtildeath:
In case there are no lines to split through the strategy, we can now avoid posting the message that the order cannot be split according to its strategy.

### DIFF
--- a/sale_order_split_strategy/models/sale_order.py
+++ b/sale_order_split_strategy/models/sale_order.py
@@ -1,9 +1,12 @@
 # Copyright 2024 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import logging
 from collections import defaultdict
 
 from odoo import fields, models
 from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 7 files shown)


henrybackman (approved):

sale_order_split_strategy/models/sale_order.py
```diff
@@ -52,7 +52,7 @@ def _handle_no_lines_to_split(self, silent_errors=False):
         )
         if not silent_errors:
             raise UserError(msg)
-        else:
+        elif not self.env.context.get("_no_lines_to_split_no_msg"):
```
simahawk: What about adding a boolean on the strategy instead?

grindtildeath: I see a couple issues with this:
- If we add a Boolean about the message posting, we still have an error management through code where `silent_errors` is used. Then whatever the value set on the strategy would not apply if silent_errors is True in the code execution.
- If we define a Selection field to handle the multiple possibilities (ie raise error, post message, do nothing), we cannot set it at strategy level since documents with different split strategies could be split through the same action.

IMO the best would then be to have a company setting with the Selection field. What do you think?

TDu: Not sure to understand why this selection field could not be set on the strategy.
But whatever the solution chosen would be good to include some notes in the readme about it.

TDu (approved):
That new company field could be used in the unit tests

sale_order_split_strategy/models/sale_order.py
grindtildeath: Action to split is meant to be called on multiple records. So if different records would have different strategies with different error handling (eg raise error and do nothing), we would raise the error and those with do nothing would still not be split...

simahawk (approved):

simahawk:
@grindtildeath pls rebase and squash

PR #4092, opened 2026-01-07
[16.0][FIX] sale_discount_display_amount: Dedicated compute method for discounts

qgroulard:
Because when the discount total is changed on a SOL, the discount total of every lines of the SO is recomputed.
If we use the same method as the one that computes the price total (`_compute_amount()`), it triggers the re-computation of fields such as 'discount' and 'price_unit', which is unexpected.

See in this video how price and discount of other lines are recomputed when we edit a line (before this patch):

[Capture vidéo du 07-01-26 14:50:06.webm](https://github.com/user-attachments/assets/1c1f93ae-a2c7-4b86-ba88-0d9fb8cc5bbd)

### DIFF
--- a/sale_discount_display_amount/hooks.py
+++ b/sale_discount_display_amount/hooks.py
@@ -48,4 +48,4 @@ def post_init_hook(cr, registry):
     order_ids = cr.fetchall()
 
     orders = env["sale.order"].search([("id", "in", order_ids)])
-    orders.mapped("order_line")._update_discount_display_fields()
+    orders.mapped("order_line")._compute_discount_amount()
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
Seems legit

alexey-pelykh (approved):
Thanks for this fix -- the root cause analysis is spot on. Piggybacking on `_compute_amount()` triggered recomputation of `discount`, `price_unit` etc. on sibling lines, which is clearly undesirable.

The dedicated `_compute_discount_amount()` with its own `@api.depends` is the right approach. It keeps the same dependency set (`product_uom_qty`, `discount`, `price_unit`, `tax_id`) and only writes the two display fields it owns. Clean separation.

One minor nit: the version in `__manifest__.py` should be bumped to `16.0.1.1.1` for this bugfix, per OCA versioning guidelines.

lmignon:
This branch has been merged as a patch while it changes the API by renaming some methods and removing others..... for the record. Another one fixing the same issue was submitted prior to this one.... [#13706](https://github.com/OCA/sale-workflow/pull/3706)

PR #4097, opened 2026-01-08
[19.0][ADD] sale_minimum_amount

AaronHForgeFlow:
Same concept as https://github.com/OCA/purchase-workflow/tree/16.0/purchase_minimum_amount

Depends on:

- [x] https://github.com/OCA/sale-workflow/pull/4098


rousseldenis (approved):
LGTM

alexey-pelykh (commented):
Thanks for porting this from the purchase side, Aaron -- clean module overall.

A few things I noticed while going through the code:

The `@api.constrains` on `SaleOrder._check_minimum_amount` only lists `partner_id`, but the block/unblock logic also depends on `amount_untaxed`. Right now the line-level constraint in `sale_order_line.py` bridges that gap by calling the order method when qty/price change, which works. Still, explicitly adding `amount_untaxed` (or the fields that compose it) to the order-level decorator would make the intent clearer and protect against someone later calling `_check_minimum_amount` from a different path that bypasses the line constraint.

The dynamic `@api.constrains(lambda self: self._check_minimum_amount_fields())` on the line model is the same pattern used in `purchase_minimum_amount`, so I trust it works, but it is worth double-checking on 19.0 that the lambda resolves at decoration time the way we expect -- Odoo's metaclass has changed a few things across versions.

`_check_minimum_amount` returning `True` is a bit unusual for a constrains method (which normally either raises or returns nothing). Not a blocker, just slightly misleading for readers.

On test coverage: the two existing tests cover the main happy paths well. It would be nice to also cover the edge cases where `minimum_so_amount` is 0 (should never block) and where the partner on an existing SO is changed to one with a different minimum. Again, not a blocker.

CI failures are expected since #4098 hasn't landed yet -- no concern there.

alexey-pelykh (commented):
Following up on my earlier comments since there haven't been new commits -- the points below are still open.

The main one worth addressing before merge: the `blocked_so` search filter in `sale_order_view.xml` replaces the domain from the base `sale_order_approval_block` module entirely. If the base filter already has a working domain for `approval_block_id != False`, this override is either redundant or, if the base uses a different expression, it silently changes that filter's behavior for all users of the dependency module. Worth double-checking whether this override is actually needed or whether the base filter already covers it.

The `@api.constrains("partner_id")` observation and the test coverage suggestions from my previous review still stand but are minor enough that they shouldn't block this.

AaronHForgeFlow:
Thank you for your comments @alexey-pelykh I actually removed the sale line triggers as I tested it and it worked fine with the api.contrains in the sales order. I think it makes more sense.

Also removed the filter inheritance, it was not needed.

rousseldenis:
@AaronHForgeFlow Could you rebase ?

sale_minimum_amount/views/sale_order_view.xml
```diff
@@ -0,0 +1,21 @@
...
+            <field name="date_order" position="after">
+                <field name="minimum_so_amount" />
+            </field>
+            <xpath
+                expr="//button[@name='button_release_approval_block']"
+                position="attributes"
+            >
+                <attribute
+                    name="invisible"
+                >(not approval_block_id or state not in ['draft', 'sent'])</attribute>
```
alexey-pelykh: This `invisible` value — `(not approval_block_id or state not in ['draft', 'sent'])` — is character-for-character identical (modulo the wrapping parentheses) to the button's own definition in the base module (`sale_order_approval_block/views/sale_order_view.xml`: `invisible="not approval_block_id or state not in ['draft', 'sent']"`). So this whole `position="attributes"` xpath is a no-op and can be removed. If the intent was to actually change the condition (e.g. add a minimum-amount term), the new expression would need to differ from the base.

sale_minimum_amount/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        approval_block_exception = self.env.ref(
+            "sale_order_approval_block.so_excep_approval_block",
+            raise_if_not_found=False,
+        )
+        for rec in self:
+            under_min = rec.amount_untaxed < rec.minimum_so_amount
+            force_release = rec.env.context.get(
+                "force_so_approval_block_release", False
+            )
+            if under_min and not rec.approval_block_id and not force_release:
```
alexey-pelykh, line 27: Question on the release flow: base `button_release_approval_block` just does `approval_block_id = False` (no `force_so_approval_block_release` in context). So if a manager releases an order still under the minimum and it's then edited again (amount still < minimum), this branch re-enters (`under_min and not approval_block_id and not force_release`) and re-applies the block. The tests always confirm immediately after release, so this path isn't exercised. Is re-blocking on further edit intended, or should a manual release persist until the amount clears the minimum? Not a blocker — just confirming intent.

alexey-pelykh (commented):
Thanks for the rebase and for acting on the earlier round, Aaron — removing the sale-line triggers and the filter override both landed cleanly, and the order-level `@api.constrains("partner_id", "amount_untaxed")` driving the block is nicely covered by `test_so_amount_block_above_minimum`. CI is green.

Two small things on the current head, plus one non-blocker:

1. The `button_release_approval_block` attributes override in `sale_order_view.xml` re-declares `invisible` with the exact condition the base module already sets — so the xpath is a no-op and can be dropped (inline). Same redundant-override shape I flagged for the filter last round.
2. A question on the release interaction — inline.

Non-blocking (carried from earlier): edge-case tests for `minimum_so_amount == 0` and swapping the SO partner to a different minimum would round out coverage. Not required for this PR.

sale_minimum_amount/models/sale_order.py
AaronHForgeFlow: Yes, that is intended. Test added.

sale_minimum_amount/views/sale_order_view.xml
AaronHForgeFlow: Yes, it can be removed and I did it.

AaronHForgeFlow:
@alexey-pelykh thanks for the review. I added a test case for the edge case and apply the suggestion.

rousseldenis:
@AaronHForgeFlow Could you attract reviewers ?

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
@sbidoul

sbidoul:
name reserverd

PR #4098, opened 2026-01-08
[19.0][ADD] sale_order_approval_block

AaronHForgeFlow:
Same concept as https://github.com/OCA/purchase-workflow/blob/18.0/purchase_order_approval_block


rousseldenis:
@AaronHForgeFlow So, we agree this is not the same as : https://github.com/OCA/sale-workflow/tree/17.0/sale_tier_validation ?

AaronHForgeFlow:
@rousseldenis This module focuses more on the blocking of wrong orders and not on the validation process. It will be used mainly:

- To prevent mistakes when entering the quantities and price units
- To prevent sale users to confirm sales that are not profitable enough for some reason.

 The release of the blocking will not be the most common use case but the correction of the sales order.

sale_order_approval_block/tests/test_sale_order_approval_block.py
```diff
@@ -0,0 +1,96 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderApprovalBlock(TransactionCase):
+    def setUp(self):
```
rousseldenis: Use class method instead

sale_order_approval_block/models/sale_order.py
```diff
@@ -0,0 +1,63 @@
...
+            if vals.get("approval_block_id"):
+                so.message_post(
+                    body=self.env._(
+                        'Order "%(order)s" blocked with reason "%(reason)s"',
+                        order=so.name,
+                        reason=so.approval_block_id.name,
+                    )
+                )
+        return records
+
```
alexey-pelykh, line 33: `sale.order` in Odoo 19.0 has no `button_approve` -- that method exists on `purchase.order`. This override will never be called. The exception wizard already handles releasing the block on ignore, so this method can likely just be removed. If you do want auto-release on confirm, you'd need to hook into `action_confirm` instead.

sale_order_approval_block/views/sale_order_view.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="sale_order_form_block_reason" model="ir.ui.view">
+        <field name="name">sale.order.form.block.reason</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <button name="action_cancel" position="attributes">
+                <attribute name="invisible">
```
alexey-pelykh, line 9: This replaces the full `invisible` domain of `action_cancel`, dropping both the `'draft'` state and the `or not id or locked` conditions from core. After this change users can't cancel draft orders anymore and the locked-order protection is lost. Consider using `invisible="state not in ['draft', 'sent', 'sale'] or not id or locked or (approval_block_id and state == 'draft')"` or whatever the intended blocking condition is, rather than a full replacement.

sale_order_approval_block/tests/test_sale_order_approval_block.py
alexey-pelykh: As @rousseldenis noted, please use `@classmethod` + `setUpClass(cls)` instead of `setUp(self)` -- that's the standard OCA testing convention and avoids re-creating all fixtures for every test method.

alexey-pelykh (changes requested):
Thanks for porting this from purchase-workflow -- useful pattern for blocking orders that shouldn't be confirmed yet.

A few things I'd like to see addressed:

**`button_approve` override is dead code** -- `sale.order` in Odoo 19.0 doesn't define `button_approve` (that's a `purchase.order` method). The confirmation flow goes through `action_confirm`, which `sale_exception` already overrides. So the `button_approve` method here will never be called and the auto-release-on-approve logic won't work. This probably needs to hook into `action_confirm` or the exception wizard flow instead (the wizard already handles it, so maybe just remove `button_approve` entirely?).

**Cancel button visibility override is too aggressive** -- the `action_cancel` attribute change replaces the original `invisible` domain with `state not in ['sale', 'sent']`, which drops the `'draft'` state and also loses the `or not id or locked` conditions from core. This means users can no longer cancel draft orders and the locked-order guard is gone. If the intent is only to hide cancel while blocked, this needs a more targeted condition that preserves the original logic.

**Existing review feedback** -- @rousseldenis noted to use `setUpClass` instead of `setUp` in the tests. That's still unaddressed and is the standard OCA convention.

Some smaller observations:

- The `create` method posts a message when `approval_block_id` is in `vals`, but unlike the purchase-workflow equivalent, it doesn't check `"approval_block_id" in vals` first -- it uses `vals.get("approval_block_id")` which is fine functionally but slightly inconsistent with the `write` method's pattern in the same file. Minor, not blocking.

- Missing copyright headers on all Python files. OCA guidelines expect AGPL-3 headers.

- The `tests/__init__.py` imports are in reverse dependency order (the dependent class is imported before the base class). Doesn't break anything since Python handles it, but worth reordering for clarity.

alexey-pelykh (commented):
Friendly follow-up -- the code hasn't changed since my last review, so the previous CHANGES_REQUESTED items are still outstanding. The main ones worth tackling first:

1. Remove `button_approve` (dead code on `sale.order` in 19.0).
2. Fix the `action_cancel` invisible domain so it preserves the original draft/locked conditions instead of replacing them wholesale.
3. Switch `setUp` to `setUpClass` per OCA convention (flagged earlier by @rousseldenis too).

Happy to re-review once those are addressed. No rush -- just wanted to keep this on the radar.

AaronHForgeFlow:
Thank you for your comments I agree with them and fixed them accordingly.

alexey-pelykh (commented):
Thanks for the updates — the three items from my previous review are all resolved:

- `button_approve` dead code: removed.
- `action_cancel` invisible domain: now preserves the core `not id or locked` conditions.
- `setUpClass`: done.

One remaining minor: Python files are still missing AGPL-3 copyright headers (OCA guideline). Not blocking.

DavidJForgeFlow (approved):

AaronHForgeFlow:
@alexey-pelykh I added the headers. Ok to go?

MiquelRForgeFlow (approved):

AaronHForgeFlow:
@rousseldenis Is this ok to be merged?

BhaveshHeliconia (approved):
LGTM!

alexey-pelykh (approved):
Thanks @AaronHForgeFlow — the AGPL-3 headers are in place now, which was the last open item from my previous review. The three earlier points are all resolved: the dead-code `button_approve` override is removed, the `action_cancel` invisible domain now preserves the core `not id or locked` guard, and the tests use `setUpClass`. CI is green.

Good to go on my side. LGTM.

rousseldenis (approved):

AaronHForgeFlow:
Can we retry the merge? or is it something to be manually handled in PyPI?

rousseldenis:
@sbidoul

sbidoul:
name reserved

PR #4108, opened 2026-01-15
[18.0][ADD] sale_order_partner_restrict: new module

juancarlosonate-tecnativa:
This module automatically blocks and cancels sale orders from unwanted customers at creation time. This is essential when orders come from external integrations without manual review.

Configure rules to block orders based on partner field values (email, phone, VAT, ZIP, country, etc.). When matched, orders are auto-canceled with a message in the chatter.

@tecnativa TT59844
@sergio-teruel @carlosdauden @eduezerouali-tecnativa


eduezerouali-tecnativa (approved):
LGTM, I tested on runboat and works fine 

Image

lfreeke (approved):
👍 Functional test

sale_order_partner_restrict/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+from markupsafe import Markup
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _check_partner_block_rules(self, partner):
+        rules = self.env["sale.order.block.rule"].search([("active", "=", True)])
```
sergio-teruel, line 13: The mothod search returns active record by default so you do not need this domain .. simply use .search([])

sale_order_partner_restrict/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from markupsafe import Markup
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
```
sergio-teruel, line 11: @api.model ??

sale_order_partner_restrict/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2026 Juan Carlos Oñate - Tecnativa <juancarlos.onate@tecnativa.com>
```
sergio-teruel, line 1: Change copy right like this
Copyright 2026 Tecnativa - Juan Carlos Oñate

sergio-teruel (changes requested):

sale_order_partner_restrict/models/sale_order.py
alexey-pelykh, line 13: Echoing sergio-teruel's point: `search()` already filters on `active=True` by default, so the explicit domain `[("active", "=", True)]` is unnecessary. Plain `self.env["sale.order.block.rule"].search([])` is cleaner.

sale_order_partner_restrict/models/sale_order.py
alexey-pelykh, line 11: This method only uses `self.env`, not any field from the current recordset. It should be decorated with `@api.model` so it can also be called without a record context.

sale_order_partner_restrict/models/sale_order.py
alexey-pelykh, line 1: OCA convention for Tecnativa copyright lines is `Copyright <year> Tecnativa - <author name>`. Same applies to `sale_order_block_rule.py` and the test file.

sale_order_partner_restrict/views/sale_order_block_rule_views.xml
```diff
@@ -0,0 +1,92 @@
...
+        <field name="help" type="html">
+            <p class="o_view_nocontent_smiling_face">
+                Create your first partner block rule
+            </p>
+            <p>
+                Configure rules to automatically block sale orders based on partner field values.
+            </p>
+        </field>
+    </record>
+    <menuitem
```
alexey-pelykh, line 78: This menuitem has no `name` so Odoo uses the XML ID as display text. Either assign a proper translatable name or remove this record if it was intended as a visual separator (menuitems don't support that pattern).

sale_order_partner_restrict/models/sale_order_block_rule.py
```diff
@@ -0,0 +1,95 @@
...
+            ("street", "Street"),
+            ("city", "City"),
+            ("country_id", "Country"),
+            ("state_id", "State"),
+        ],
+        required=True,
+        help="Partner field to validate",
+    )
+    blocked_values = fields.Text(
+        help="Blocked values (comma-separated list)",
```
alexey-pelykh, line 30: Consider adding an `@api.constrains` to validate that `zip_range_from <= zip_range_to` (as integers). If a user accidentally reverses them, the range check silently never matches -- hard to debug.

sale_order_partner_restrict/models/sale_order_block_rule.py
```diff
@@ -0,0 +1,95 @@
...
+        "Useful for blocking ZIP codes by country (e.g., Spanish ZIPs 35001-35999)",
+    )
+    zip_range_from = fields.Char(
+        string="ZIP From",
+        help="Starting ZIP code for range blocking",
+    )
+    zip_range_to = fields.Char(
+        string="ZIP To",
+        help="Ending ZIP code for range blocking",
+    )
```
alexey-pelykh, line 45: For `country_id` and `state_id` the check resolves to `.name` and does a string comparison against the comma-separated text. The name is locale-dependent ("Spain" vs "España"), making this fragile in multilingual databases. A Many2many to `res.country` / `res.country.state` would be more robust for these relational fields.

alexey-pelykh (commented):
Thanks for this module, it addresses a real need for shops with external integrations.

I see sergio-teruel already flagged three items that still look unaddressed on the current HEAD. I have a few additional observations beyond those.

The separator menuitem `menu_sale_order_separator` has no `name` attribute, so Odoo falls back to the XML ID as the display string (visible in the .pot as `msgid "menu_sale_order_separator"`). Either give it a translatable label or remove it entirely -- a bare menuitem does not render as a visual separator in the webclient.

For the relational selection values `country_id` and `state_id`, the blocked-values check resolves them to `.name` and then does a string comparison against the comma-separated text field. This is fragile because the name is locale-dependent ("Spain" vs "España"). A more robust approach would be to use Many2many fields to `res.country` / `res.country.state` when the partner_field is one of these relational types, or at minimum document that the comparison uses the English name.

There is no validation that `zip_range_from <= zip_range_to`. If a user accidentally reverses them the range check silently never matches, which is hard to debug. A `@api.constrains` would catch that at save time.

The `create` override posts the chatter message and then calls `_action_cancel()`. Depending on installed modules, `_action_cancel` may itself post a cancellation message. Worth checking whether this produces a duplicate chatter entry -- if so, consider passing a context flag to suppress the standard cancel message or reordering the calls.

PR #4111, opened 2026-01-16
[19.0][MIG] product_customerinfo_sale

JasminSForgeFlow:
Standard Migration

@ForgeFlow

Depends on:

- [x] https://github.com/OCA/product-attribute/pull/2167


AaronHForgeFlow (approved):
Functional test :+1:

product_customerinfo_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    product_customer_code = fields.Char(
+        compute="_compute_product_customer_code",
+    )
+
+    @api.depends("product_id")
+    def _compute_product_customer_code(self):
```
alexey-pelykh, line 19: This depends on `product_id` only, but the body calls `_select_customerinfo(partner=line.order_partner_id)`. If the SO partner changes after lines exist, the customer code won't refresh. Consider:
```python
@api.depends("product_id", "order_partner_id")
```

product_customerinfo_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+                partner=item.order_partner_id
+            )
+            if customerinfo.product_code:
+                # Avoid to put the standard internal reference
+                item = item.with_context(display_default_code=False)
+            super(SaleOrderLine, item)._compute_name()
+            if customerinfo.product_code:
+                item.name = f"[{customerinfo.product_code}] {item.name}"
+        return
+
```
alexey-pelykh, line 46: Nit: the bare `return` at the end of the method is unnecessary (Python returns `None` implicitly). Feel free to drop it.

product_customerinfo_sale/i18n/product_supplierinfo_for_customer_sale.pot
```diff
@@ -0,0 +1,30 @@
+# Translation of Odoo Server.
```
alexey-pelykh: This `.pot` file still references the old module name `product_supplierinfo_for_customer_sale`. Looks like a leftover from a previous version -- safe to remove.

alexey-pelykh (commented):
Thanks for the migration, looks solid overall -- tests are thorough and the code is clean.

One functional concern: `_compute_product_customer_code` depends only on `product_id`, but the computation uses `order_partner_id`. If the partner on the SO changes after a line already exists, the customer code won't recompute. Adding `order_partner_id` to the `@api.depends` would fix this.

Also a minor cleanup note: `i18n/product_supplierinfo_for_customer_sale.pot` is a leftover from the old module name and can be removed.

alexey-pelykh (commented):
Thanks for migrating this to 19.0. The onchange rename (`_onchange_product_id_warning` -> `_onchange_product_id`) and the test adaptations (own product fixtures instead of demo data refs, `group_ids` rename) are the right changes for the new version.

One point that I flagged in my inline comments earlier and still stands: `_compute_product_customer_code` depends only on `product_id`, but the body reads `order_partner_id` to select the customerinfo. Since `order_partner_id` is a stored related field, adding it to `@api.depends` would make the code recompute correctly when the SO partner changes. This is technically a pre-existing issue from 18.0, but worth fixing while we're here.

The stale `i18n/product_supplierinfo_for_customer_sale.pot` is a leftover from the old module name and can be dropped.

Otherwise the manifest, security, views, and tests all look correct for 19.0.

JasminSForgeFlow:
> Thanks for migrating this to 19.0. The onchange rename (`_onchange_product_id_warning` -> `_onchange_product_id`) and the test adaptations (own product fixtures instead of demo data refs, `group_ids` rename) are the right changes for the new version.
> 
> One point that I flagged in my inline comments earlier and still stands: `_compute_product_customer_code` depends only on `product_id`, but the body reads `order_partner_id` to select the customerinfo. Since `order_partner_id` is a stored related field, adding it to `@api.depends` would make the code recompute correctly when the SO partner changes. This is technically a pre-existing issue from 18.0, but worth fixing while we're here.
> 
> The stale `i18n/product_supplierinfo_for_customer_sale.pot` is a leftover from the old module name and can be dropped.
> 
> Otherwise the manifest, security, views, and tests all look correct for 19.0.

Updated, Thanks

product_customerinfo_sale/models/sale_order_line.py
JasminSForgeFlow: This will cause missing-return warning in pre-commit

BhaveshHeliconia (approved):
LGTM!

BhaveshHeliconia:
Please squash the commits as explained here: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

BhaveshHeliconia:
@JasminSForgeFlow the dependency has been merged. You can drop the last commit and rebase the PR.

rousseldenis (approved):

PR #4112, opened 2026-01-17
[17.0][ADD] sale_partner_address_filter: Select Only Customer Contacts

dreispt:


rousseldenis:
@dreispt Could you check https://github.com/OCA/sale-workflow/tree/18.0/sale_partner_address_restrict ?

dreispt:
@rousseldenis I did, and it doesn't work for my case.
There is an explanation in the README.

rousseldenis:
> @rousseldenis I did, and it doesn't work for my case. There is an explanation in the README.

@dreispt What is complex in configuration in sale_partner_address_restrict ?

dreispt:
It assumes the user will setup exceptions for the partners that allow any address selection.
All the other are blocked by default.
That add a setup burden on users.

This implementation moves that decision to the sales order entry time.
There is a switch to enable any address selection.
It works at the UX level, and no hard restrictions are implemented.

sale_partner_address_filter/views/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2025 Open Source Integrators
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
```
alexey-pelykh, line 2: This file should be removed. The `views/` directory contains only XML data files -- `__init__.py` is not needed here and is not standard for OCA modules.

sale_partner_address_filter/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+    def _compute_address_domain(self):
+        """Compute domain for address fields"""
+        for order in self:
+            comm_partner = order.partner_id.commercial_partner_id
+            address_domain = (
+                [("commercial_partner_id", "=", comm_partner.id)]
+                if order.only_customer_addresses
+                else []
+            )
+            # Return domain as string for XML compatibility
```
alexey-pelykh, line 27: Using `str()` to serialize a domain list into a `Text` field is fragile. `sale_partner_address_restrict` (already in this repo) uses `fields.Binary` for the same purpose. Consider `fields.Binary` or `fields.Json`.

sale_partner_address_filter/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2025 Open Source Integrators
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Sale Partner Address Filter",
+    "version": "17.0.1.0.0",
```
alexey-pelykh, line 6: Missing `"summary"` key. OCA convention expects a short one-line summary.

sale_partner_address_filter/i18n/sale_partner_address_filter.pot
```diff
@@ -0,0 +1,41 @@
...
+msgstr ""
+
+#. module: sale_partner_address_filter
+#: model:ir.model,name:sale_partner_address_filter.model_sale_order
+msgid "Sales Order"
+msgstr "Bon de commande"
+
+#. module: sale_partner_address_filter
+#: model:ir.model.fields,field_description:sale_partner_address_filter.field_sale_order__only_customer_addresses
+msgid "Select Only Customer Contacts"
```
alexey-pelykh, line 30: `.pot` template should have empty `msgstr` values. This has French translations filled in. Regenerate a clean `.pot`.

alexey-pelykh (changes requested):
Thanks for the contribution, @dreispt. A few significant concerns:

**Overlap with existing module.** This repo already ships `sale_partner_address_restrict` on 17.0 which does the same thing: filters `partner_shipping_id` and `partner_invoice_id` by `commercial_partner_id`. The existing module has a company-level toggle, server-side `@api.constrains` validation, and tests. Could you clarify the gap that justifies a separate module rather than contributing to the existing one?

**No tests.** OCA guidelines require tests for new modules.

**`views/__init__.py` should be removed.** Views directory contains only XML files -- `__init__.py` is unnecessary.

**`.pot` has translations filled in.** The `.pot` template has French translations in `msgstr` values. Those belong only in `fr.po`. Regenerate a clean `.pot`.

**`str()` domain serialization is fragile.** Using `fields.Text` with `str()` to serialize a domain list is unusual. `sale_partner_address_restrict` uses `fields.Binary` for the same purpose. Consider `fields.Binary` or `fields.Json` for a cleaner approach.

**Mixed concerns.** The `sale_order_invoice_amount` fix is bundled in this PR -- should be a separate PR.

**Missing `summary` in manifest and `_description` on model class.**

PR #4114, opened 2026-01-20
[18.0][IMP] sale_order_secondary_unit: pass secondary_uom_id to account move line

AungKoKoLin1997:
This PR propagates `secondary_uom_id` from sale order lines to account move lines when the field exists on account.move.line. This is useful when `account_move_secondary_unit` is installed, and avoids the need for a glue module.

Related: https://github.com/OCA/account-invoicing/pull/2230

@qrtl QT6210

### DIFF
--- a/sale_order_secondary_unit/models/sale_order.py
+++ b/sale_order_secondary_unit/models/sale_order.py
@@ -62,3 +62,12 @@ class SaleOrderLine(models.Model):
                     line.secondary_uom_unit_price = 0
             else:
                 line.secondary_uom_unit_price = 0
+
+    def _prepare_invoice_line(self, **optional_values):
+        # Set secondary UoM values only when account_move_secondary_unit is installed
+        # (i.e., the fields exist on account.move.line).
+        res = super()._prepare_invoice_line(**optional_values)
+        aml_fields = self.env["account.move.line"]._fields
+        if "secondary_uom_id" in aml_fields and self.secondary_uom_id:
+            res["secondary_uom_id"] = self.secondary_uom_id.id
+        return res


yostashiro (approved):
Code review. LGTM. 👍

nobuQuartile (approved):
Code review: LGTM

rousseldenis (approved):
LGTM. A test case is welcome

sale_order_secondary_unit/models/sale_order.py
```diff
@@ -62,3 +62,12 @@ def _compute_secondary_uom_unit_price(self):
...
                 line.secondary_uom_unit_price = 0
+
+    def _prepare_invoice_line(self, **optional_values):
+        # Set secondary UoM values only when account_move_secondary_unit is installed
+        # (i.e., the fields exist on account.move.line).
+        res = super()._prepare_invoice_line(**optional_values)
+        aml_fields = self.env["account.move.line"]._fields
+        if "secondary_uom_id" in aml_fields and self.secondary_uom_id:
+            res["secondary_uom_id"] = self.secondary_uom_id.id
+        return res
```
alexey-pelykh, line 73: The manifest version is still `18.0.1.0.1` (same as base branch). Since this is an IMP, it should be bumped -- `18.0.1.1.0` per OCA versioning convention.

alexey-pelykh, line 73: As @rousseldenis mentioned, a test case would be valuable here. Even without `account_move_secondary_unit` installed, you can at least test that `_prepare_invoice_line` still returns a proper dict and doesn't break the standard invoice creation flow (i.e., verify the `if` branch is simply skipped when the field doesn't exist). That alone would fix the codecov failure and serve as a basic regression guard.

alexey-pelykh (commented):
Thanks for the contribution! The approach of probing `_fields` to conditionally set the secondary UoM is clean and avoids a hard dependency on `account_move_secondary_unit` -- nice.

Two things worth addressing before merge:

sale_order_secondary_unit/models/sale_order.py
AungKoKoLin1997: I believe the version bump should be handled by OCA Bot during the merge, and manually bumping the version could cause conflicts if multiple PRs are merged for the same module.

alexey-pelykh: Fair point — you're right that OCA Bot handles the version bump at merge time. Manual bumps would just create conflicts. Noted, thanks for the correction.

PR #4115, opened 2026-01-20
[ADD][16.0] sale_price_compliance

Shide:
This PR introduces a new Price Compliance mechanism to help sales users adhere to pricing policies. 
It provides visual feedback via a widget and flexible configuration options to define discount/price thresholds.

## Key Features
* **Price Compliance Widget**
    * Added a visual widget on the Sales Order Line. This allows users to easily view the applicable price tiers (Thresholds) directly on the **form**, **tree** or **kanban** view.

* **Flexible Configuration Hierarchy**
    * The thresholds are highly configurable and follow a specific fallback order to ensure granular control. You can define them at:
        1.  **Product** (Specific)
        2.  **Product Category** (Group)
        3.  **Company** (Global default)

* **Pricelist Compatibility**
    * The system fully respects the price computation from **Pricelists** and **Currency**. The compliance check is performed against the final calculated price, ensuring that specific pricelist rules are not overridden.

https://www.loom.com/share/b6d97d85617747b1a708a539af4a9d94

## PR Series
- **sale_price_compliance**: https://github.com/OCA/sale-workflow/pull/4115
- sale_report_delivered_price_compliance: https://github.com/OCA/sale-reporting/pull/352
- sale_commission_product_criteria_price_compliance: https://github.com/OCA/commission/pull/654

---

MT-12373 MT-13322 @moduon @rafaelbn @fcvalgar @chienandalu @yajo @EmilioPascual @Gelojr please review if you want 😄


sale_price_compliance/models/sale_order.py
```diff
@@ -0,0 +1,40 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _action_confirm(self):
+        res = super()._action_confirm()
+        for record in self:
+            error_message = record._check_compliant_pricing()
+            if error_message:
+                raise exceptions.UserError(error_message)
```
yajo: suggestion: IMHO it would make sense that the method called `_check_...` raises the exception by itself:


```suggestion
            record._check_compliant_pricing()
```

sale_price_compliance/models/sale_order.py
```diff
@@ -0,0 +1,40 @@
...
+                        self.env.user.name,
+                    )
+                )
+            return
+        if non_compliant_lines:
+            return _(
+                "The order contains lines with non-compliant prices.\n"
+                "Please review the prices before confirming the order or "
+                "contact your Sales manager for further assistance."
+            )
```
yajo: suggestion: raise here.


```suggestion
            return exceptions.UserError(_(
                "The order contains lines with non-compliant prices.\n"
                "Please review the prices before confirming the order or "
                "contact your sales manager for further assistance."
            ))
```

sale_price_compliance/models/sale_order_line.py
```diff
@@ -0,0 +1,224 @@
...
+        readonly=True,
+        help="Indicates the Tier of Price Compliance based on the unit price and "
+        "applied discount compared to defined thresholds.",
+    )
+    price_compliance_data = fields.Json(
+        compute="_compute_price_compliance_tier",
+        store=True,
+        readonly=True,
+        help="Holds additional data related to price compliance calculations.",
+    )
```
yajo, line 25: issue(style): please move fields to the top of the class, as per the guidelines.

yajo (changes requested):
Code review, only minor issues.

yajo (approved):
Code review.

rousseldenis:
@Shide @yajo @rafaelbn We agree this has no link with margins (maybe to have simpler rules ? ) ?

sale_price_compliance/static/src/widgets/price_compliance_level_widget.xml
```diff
@@ -0,0 +1,119 @@
...
+                t-att-tabindex="props.record.data.price_compliance_tier ? '0' : '-1'"
+                t-on-click="showPopup"
+                class="cursor-pointer"
+            >
+                <t t-out="calcData.price_compliance_tier_display" />
+            </a>
+        </div>
+    </t>
+
+    <t t-name="sale_price_compliance.PriceComplianceTierPopover" class="hola1" owl="1">
```
chienandalu: Testing?
```suggestion
    <t t-name="sale_price_compliance.PriceComplianceTierPopover" owl="1">
```

Shide:
> @Shide @yajo @rafaelbn We agree this has no link with margins (maybe to have simpler rules ? ) ?

I didn't fully understand your question, but I think you are asking if it is possible to base the price tiers on the margin rather than the product price.

We decided to take this approach because cost price variability (when using FIFO) can be confusing for the Commercial User, and also because our client requested it this way (and we agreed with the decision).

The idea behind this module is to give Commercial User some flexibility to set prices and earn commissions based on that. We never considered the scenario where the discount drops below the cost price, because the Company must always make a profit regardless of the applied discount.

Example: A product has a price of €100 and a cost of €50. We would never create a 50% tier (adjusted to the cost price, even if variable) because the company wouldn't make money on that transaction. We would apply tiers of 5%, 15%, 25% (for example) so that in the worst-case scenario, the company still gains 25%.

If we were to provide a configuration option to define the base for the tiers (Product Base Price vs. Margin), it wouldn't be as simple as just enabling a checkbox. The tiers would change completely because they would be compared against a much lower value (the margin).

Similarly, we would still need to calculate the unit prices that the Commercial User choose based on the widget, so the computational load would remain practically the same (though I might save on unit conversion).

I hope this answers your question. I'm open to further questions.

Andrii9090-tecnativa (approved):
Code & function review

Image

Shide:
Upgraded. Now you  can customize Text and Icon via system parameter.

Also Pricelist tier is checked only if Final Price is below Non Compliant. Price between last tier (T3 for example) and Pricelist, will be considered Pricelist (since you are improving the margin for the company).

Shide:
Now labels on the views shows the custom text if System Parameter is used

sale_price_compliance/models/price_compliance_threshold_config_mixin.py
```diff
@@ -0,0 +1,113 @@
...
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        """Replaces price_compliance_threshold_t<N> labels based on tier text
+        definitions from sale.order.line, which can be customized via system
+        parameters."""
+        result = super().get_view(view_id=view_id, view_type=view_type, **options)
+        # Get tier text selection from sale.order.line (only t1, t2, t3)
+        tiers_selection = self.env[
+            "sale.order.line"
+        ]._get_price_compliance_selection_tiers_text()[:3]
```
chienandalu, line 113: You could return early if no config_parameter is set as in that case you don't need override anything.

In the other hand, shouldn't these methods be in the abstract as well?

sale_price_compliance/static/src/widgets/price_compliance_level_widget.xml
Shide: lol 🙄

Shide:
@chienandalu separated selection field into a mixin and inherited the mixin in the other modules of the PR series

sale_price_compliance/models/price_compliance_threshold_config_mixin.py
```diff
@@ -0,0 +1,123 @@
...
+                OR (
+                    price_compliance_threshold_t1 >= 0.0 AND
+                    price_compliance_threshold_t2 >= 0.0 AND
+                    price_compliance_threshold_t3 >= 0.0
+                )
+            )""",
+            "Price Compliance Thresholds Percentage should be positive.",
+        ),
+        (
+            "check_price_compliance_le_1",
```
alexey-pelykh, line 46: Nit: the SQL constraints enforce positivity, max 1.0, and no gaps, but they don't enforce monotonic ordering (`t1 <= t2 <= t3`). A user could set `t1=30%, t2=10%, t3=20%` which would lead to overlapping ranges in the widget. Not blocking, but worth considering as a future improvement or documenting the expected behavior.

sale_price_compliance/models/price_compliance_threshold_tier_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        return self._get_price_compliance_selection_tiers_icon_color()
+
+    @api.model
+    def _get_tier_selections(self, param, default_selection):
+        """Get tier selections using config parameters"""
+        icp_st = self.env["ir.config_parameter"].sudo().get_param(param)
+        try:
+            st_dict = safe_eval(icp_st)
+        except Exception:
```
alexey-pelykh, line 38: Minor: OCA convention prefers lazy logger formatting over f-strings:
```suggestion
            _logger.warning("Wrong parameter value for %s", param)
```

sale_price_compliance/models/price_compliance_threshold_config_mixin.py
```diff
@@ -0,0 +1,123 @@
...
+            """CHECK (
```
alexey-pelykh, line 47: Tiny typo: "thresolds" -> "thresholds" in the docstring.

alexey-pelykh (approved):
Thanks for this well-designed module, Eduardo. The mixin-based architecture with Product > Category > Company fallback is clean, and the SQL constraints are thorough. I went through the full diff and have a few observations that are all non-blocking.

**Overall**

The code is solid. Good use of `pre_init_hook` to avoid expensive recomputation on install, proper `ensure_one()` guards, and the tier customization via system parameters is a nice touch. Tests cover the main flows well including constraint validation and parameter customization.

A few minor notes:

1. **Threshold ordering is not enforced** -- the SQL constraints check positivity, max 1.0, and no gaps, but nothing prevents `t1=0.30, t2=0.10, t3=0.20` (i.e. t2 < t1). The tier computation in `_compute_price_compliance_tier` would still work because it iterates the tiers sequentially, but it might confuse users configuring the thresholds since the discount ranges would overlap. Might be worth a constraint like `t1 <= t2 <= t3` or at minimum a note in the docs. Not blocking since the computation handles it correctly.

2. **`_get_price_compliance_thresholds` uses `if not threshold`** -- this means a threshold of exactly `0.0` is treated as "not set", which is consistent with the gap constraint logic. Just flagging it as intentional since `float(0.0)` is falsy in Python.

3. **Minor typo in docstring** -- `_get_price_compliance_thresholds` says "thresolds" instead of "thresholds" (line in config mixin).

4. **`f-string` in logger warning** -- in `_get_tier_selections`, the line `_logger.warning(f"Wrong parameter value for {param}")` uses an f-string. OCA guidelines generally prefer `_logger.warning("Wrong parameter value for %s", param)` with lazy formatting to avoid string formatting when the log level is disabled.

Nice work on the widget and the popover UX.

alexey-pelykh (approved):
Nice module, thanks @Shide for the contribution!

The threshold fallback chain (product -> template -> category -> company) is well designed, and the SQL constraints for gap/overlap prevention are solid. The `pre_init_hook` for pre-creating columns is a good touch for install performance. OWL widget looks clean too.

A couple of minor things I noticed:

- **Copyright year in `__manifest__.py`**: It says 2025 while all other files use 2026. Probably just an oversight from when the module was started.

- **f-string in logger** (`price_compliance_threshold_tier_mixin.py`): The `_logger.warning(f"Wrong parameter value for {param}")` should use lazy formatting (`_logger.warning("Wrong parameter value for %s", param)`) per Python logging best practices.

- **No test for the confirmation blocking flow**: The `_action_confirm` / `_check_compliant_pricing` path is core to the module's value -- blocking non-compliant orders for regular users and the manager bypass with `message_post`. Would be great to have a test covering both the `UserError` raise and the manager bypass scenario. Could be a follow-up.

None of these are blockers given the existing approvals and the overall quality of the module. Good work!

rafaelbn (approved):
👍🏼 tested in staging 

/ocabot merge nobump

PR #4120, opened 2026-01-21
[18.0][ADD] sale_order_line_split_by_packaging

mathieudelva:


rousseldenis (approved):
LGTM

sale_order_line_split_by_packaging/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
+from odoo import _, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    show_split_button = fields.Boolean(compute="_compute_show_split_button")
```
alexey-pelykh, line 7: Missing `@api.depends('product_packaging_id', 'product_uom_qty')` -- without it the compute will not react to field changes.

sale_order_line_split_by_packaging/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def _compute_show_split_button(self):
+        for record in self:
+            record.show_split_button = False
```
alexey-pelykh, line 11: Float modulo (`%`) is unreliable with Odoo quantities. Consider using `float_is_zero(record.product_uom_qty % record.product_packaging_id.qty, precision_rounding=record.product_uom.rounding)` instead, or `float_compare`.

sale_order_line_split_by_packaging/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+            if record.product_packaging_id:
+                if not record.product_uom_qty % record.product_packaging_id.qty == 0:
+                    record.show_split_button = True
```
alexey-pelykh, line 14: This action dict is missing a `context` key. The wizard's `default_get` does `self.env.context["active_id"]` which will `KeyError`. Add:
```python
"context": {"active_id": self.id},
```

sale_order_line_split_by_packaging/wizard/split_order_line_wizard.py
```diff
@@ -0,0 +1,110 @@
...
+        for pkg in packaging_ids:
+            qty = int(remaining_qty / pkg.qty) * pkg.qty
+            if qty > 0:
+                lines.append(self._prepare_line(order_line, qty, pkg, items_ids))
+                remaining_qty -= qty
+
+        if remaining_qty:
+            lines.append(
+                self._prepare_line(order_line, remaining_qty, False, items_ids)
+            )
```
alexey-pelykh, line 64: `items_id` (despite its name) could be a multi-record recordset if multiple pricelist rules match the same product+packaging combo. `.fixed_price` on a multi-record set will raise. Safeguard with `items_id[:1].fixed_price` or similar.

sale_order_line_split_by_packaging/wizard/split_order_line_wizard.py
```diff
@@ -0,0 +1,110 @@
...
+            }
+        )
+        line_ids = self.line_ids - line
+
+        for line in line_ids:
+            original_line.copy(
+                {
+                    "order_id": order_id.id,
+                    "product_uom_qty": line.product_uom_qty,
+                    "product_packaging_id": line.packaging_id.id,
```
alexey-pelykh, line 91: The packagings here are iterated in recordset order (by id). The split result will be non-deterministic if there are multiple packagings with different `qty`. Sorting by `qty` descending (`packaging_ids.sorted('qty', reverse=True)`) would give the most intuitive greedy split.

sale_order_line_split_by_packaging/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "summary": "Sale Order Line Split by Packaging",
+    "version": "18.0.1.0.0",
+    "category": "sale",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": " Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["mathieudelva"],
+    "license": "AGPL-3",
+    "depends": [
+        "sale_pricelist_packaging",
+    ],
```
alexey-pelykh, line 16: Leading space in author: `" Akretion, ..."` -- should be `"Akretion, Odoo Community Association (OCA)"`.

sale_order_line_split_by_packaging/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_split_order_line_wizard,split.order.line.wizard,model_split_order_line_wizard,base.group_user,1,1,1,1
```
alexey-pelykh, line 2: `base.group_user` is quite broad for a sales-specific wizard. Consider using `sales_team.group_sale_salesman` instead.

alexey-pelykh (changes requested):
Thanks for this module, Mathieu -- the split-by-packaging idea is quite useful.

A few things I noticed that should be addressed before merging:

The `show_split_button` computed field is missing an `@api.depends` decorator, so it will never recompute reactively when the user changes the packaging or the quantity on the line. It needs `@api.depends('product_packaging_id', 'product_uom_qty')`.

Also on that compute: using Python's `%` operator on float quantities is unreliable due to floating-point precision. You should use `float_compare` or `float_is_zero` from `odoo.tools` instead of `not record.product_uom_qty % record.product_packaging_id.qty == 0`.

The action returned by `split_order_line_wizard()` doesn't pass `active_id` through context to the wizard. The `default_get` relies on `self.env.context["active_id"]` but the action dict has no `context` key -- this will `KeyError` when the wizard opens. You need something like `"context": {"active_id": self.id}` in the returned action.

In `_prepare_line`, `items_id` could contain multiple pricelist items (e.g. multiple rules for the same product+packaging). Accessing `.fixed_price` on a multi-record recordset will raise. You should handle the case or use `[:1]`.

The iteration over other packagings in `default_get` doesn't sort by `qty` descending, so the split result can vary depending on recordset ordering. Sorting `packaging_ids` by qty (largest first) would make the behavior deterministic and give the user the most intuitive split.

Smaller items:
- Leading space in `__manifest__.py` author field: `" Akretion, ..."` should be `"Akretion, ..."`
- Test copyright header says 2024 while the manifest says 2026
- ACL uses `base.group_user` -- consider restricting to `sales_team.group_sale_salesman` since this is a sales-specific wizard

rousseldenis:
@mathieudelva someone else to review ?

PR #4122, opened 2026-01-21
[18.0][FIX] sale_order_general_discount: do not set discount to 0 if general discount is bypassed

JordiMForgeFlow:
With the current logic, whenever we are not applying the generic discount we are forcing always a 0 discount.

Bypassing the general discount should not imply having a 0 discount (maybe we want to set the discount manually for a particular line, or maybe the discount is applied via a configured pricelist rule). Therefore, we should rely on what super() has defined on the discount.

### DIFF
--- a/sale_order_general_discount/models/sale_order_line.py
+++ b/sale_order_general_discount/models/sale_order_line.py
@@ -24,6 +24,4 @@ class SaleOrderLine(models.Model):
                 line.order_id.general_discount or line.order_id._origin.general_discount
             ):
                 line.discount = line.order_id.general_discount
-            else:
-                line.discount = 0
         return res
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
Makes sense :+1:

AaronHForgeFlow (approved):
Makes sense :+1:

sebalix (approved):

sebalix:
Maybe Tecnativa as author would like to be noticed about this change in behavior? Ping @sergio-teruel

sergio-teruel (approved):
LGTM!!

pedrobaeza (approved):
/ocabot merge patch

JordiMForgeFlow:
@pedrobaeza not sure if it was a temporary issue, could you retry the merge?

PR #4125, opened 2026-01-26
[19.0][FW][FIX] sale_order_general_discount: do not set discount to 0 if general discount is bypassed

JordiMForgeFlow:
FW of https://github.com/OCA/sale-workflow/pull/4122

CC @pedrobaeza


pedrobaeza (approved):
/ocabot merge patch

PR #4129, opened 2026-01-27
[FIX][18.0] sale_order_carrier_auto_assign: Changing partner changes carrier

Shide:
Prior to this fix, when changing the partner on a **saved** sale order doesn't change the carrier. You must change manually even if you set `carrier_auto_assign` setting. Clicking on Update Shipping Cost doesn't shows the correct carrier.

Now, if you have activated `carrier_on_create` setting, when changing the partner on a saved sale, also changes the carrier.

Works well with [delivery_auto_refresh](https://github.com/OCA/delivery-carrier/tree/18.0/delivery_auto_refresh) module

https://www.loom.com/share/1828d8b94de045388ea3a60f9080b3b6

MT-13352 @moduon @chienandalu @Gelojr @fcvalgar @jbaudoux @yajo please review if you want 😄

### DIFF
--- a/sale_order_carrier_auto_assign/models/sale_order.py
+++ b/sale_order_carrier_auto_assign/models/sale_order.py
@@ -73,16 +73,20 @@ class SaleOrder(models.Model):
         for order in self:
             if not order.order_line:
                 continue
-            if order.delivery_set:
+            # Preserve the carrier only if explicitly set on the attribute
+            if preserve_order_carrier and order.delivery_set:
                 continue
             delivery_wiz_action = order.action_open_delivery_wizard()
             delivery_wiz_context = delivery_wiz_action.get("context", {})
             if not delivery_wiz_context.get("default_carrier_id"):
                 continue
+            # Use the real order because can be an onchange
+            if isinstance(delivery_wiz_context.get("default_order_id"), models.NewId):
+                delivery_wiz_context["default_order_id"] = order._origin.id or order.id
             delivery_wiz_model = self.env[
                 delivery_wiz_action.get("res_model")
             ].with_context(**delivery_wiz_context)
-            if self._origin:
+            if order._origin:
                 delivery_wiz = delivery_wiz_model.create({})
             else:
                 delivery_wiz = delivery_wiz_model.new({})
--- a/sale_order_carrier_auto_assign/views/sale_order_views.xml
+++ b/sale_order_carrier_auto_assign/views/sale_order_views.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_order_form_with_carrier" model="ir.ui.view">
+        <field name="name">delivery.sale.order.form.view.with_carrier</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="delivery.view_order_form_with_carrier" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='delivery_set']" position="after">
+                <!-- carrier_id must exist in view to be able to be wrote it in onchange -->
+                <field name="carrier_id" invisible="1" />
+            </xpath>
+        </field>
+    </record>
+</odoo>
--- a/sale_order_carrier_auto_assign/__manifest__.py
+++ b/sale_order_carrier_auto_assign/__manifest__.py
@@ -10,7 +10,10 @@
     "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
     "maintainers": ["jbaudoux"],
     "license": "AGPL-3",
-    "data": ["views/res_config_settings_views.xml"],
+    "data": [
+        "views/res_config_settings_views.xml",
+        "views/sale_order_views.xml",
+    ],
     "application": False,
     "installable": True,
     "depends": ["delivery"],
--- a/sale_order_carrier_auto_assign/readme/CONFIGURE.md
+++ b/sale_order_carrier_auto_assign/readme/CONFIGURE.md
@@ -1,4 +1,4 @@
 #. Go to *Settings > Sales > Shipping*.
 
-* Enable on creation: only the carrier will be set.
+* Enable on creation: only the carrier will be set when changing partner.
 * Enable on confirmation: the carrier and delivery line will be set.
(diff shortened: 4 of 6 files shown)


jbaudoux (changes requested):
The purpose was to only set the carrier, not compute the rate.
For the rate, you can install this module https://github.com/OCA/delivery-carrier/tree/18.0/delivery_auto_refresh
You shouldn't replicate here what that other module is already doing

Shide:
@jbaudoux restored the "no auto-refresh line"
are just 3 lines of code, but ok... we could get a real feature without installing another module and I think it's a mistake leave an incorrect delivery line with another carrier.

The rest of the PR is the same.
Please check the video to see the differences.

chienandalu (approved):
Nice :)

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -73,12 +73,16 @@ def _set_delivery_carrier(
         for order in self:
             if not order.order_line:
                 continue
-            if order.delivery_set:
+            # Preserve the carrier only if explicitly set on the attribute
+            if preserve_order_carrier and order.delivery_set:
```
jbaudoux, line 77: I'm quite sure you cannot do that as it will prevent to recompute the rate when the so content is modified. Please run the test of the refresh module to ensure you don't break anything

Also, don't embed auto generated files changes

jbaudoux (changes requested):

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -73,12 +73,16 @@ def _set_delivery_carrier(
...
                 continue
             delivery_wiz_action = order.action_open_delivery_wizard()
             delivery_wiz_context = delivery_wiz_action.get("context", {})
             if not delivery_wiz_context.get("default_carrier_id"):
                 continue
+            # Use the real order because can be an onchange
+            if isinstance(delivery_wiz_context.get("default_order_id"), models.NewId):
+                delivery_wiz_context["default_order_id"] = order.id.origin or order.id
```
chienandalu: Wouldn't this be something like this?
```suggestion
                delivery_wiz_context["default_order_id"] = order._origin.id
```

Shide: Sure, I can simplify this

sale_order_carrier_auto_assign/models/sale_order.py
jbaudoux: Use `export SKIP=oca-gen-addon-readme` for fix/imp commits
Auto-generated files are only for new modules and migrated modules.
Otherwise, your commit can't be fwd or backported to another version without conflicting on those generated files.

jbaudoux: I see https://github.com/OCA/delivery-carrier/blob/18.0/delivery_auto_refresh/models/sale_order.py is not relying anymore on `_set_delivery_carrier to update the rate in v18 compared to previous versions. Then it looks fine :)

jbaudoux: In previous version, there wasn't this test and the preserve carrier is managed a bit below in the code
https://github.com/OCA/sale-workflow/pull/4129/changes#diff-c0db59164377871a4b3083527e320de3d150ad86441abdf24da743f7edc55f27R94

See also here https://github.com/OCA/delivery-carrier/blob/a2ef4559d7dd704d1f6f5b062d14eae2baee58bb/delivery_auto_refresh/models/sale_order.py#L48

I would favor to drop that test so that the rate is updated when this method is called

sale_order_carrier_auto_assign/models/sale_order.py
```diff
@@ -73,12 +73,16 @@ def _set_delivery_carrier(
...
                 continue
-            if order.delivery_set:
+            # Preserve the carrier only if explicitly set on the attribute
+            if preserve_order_carrier and order.delivery_set:
                 continue
             delivery_wiz_action = order.action_open_delivery_wizard()
             delivery_wiz_context = delivery_wiz_action.get("context", {})
             if not delivery_wiz_context.get("default_carrier_id"):
                 continue
+            # Use the real order because can be an onchange
```
jbaudoux, line 83: Maybe add above:
```
if isinstance(order.id, models.NewId) and not order.id.origin:
  continue
```
or hereunder in your test instead of the fallback to order.id

Shide: Call to the `action_open_delivery_wizard` must be done with the new state (onchanged) version of the sale to get the new partner, but if `_get_delivery_rate` is called in any scenario, a real order should be used. This is why I change at this point the `default_order_id` with the id of the real order.

partner_id is preserved from context, but order_id is changed before the creation of the wizard.

Even if it's an onchange, code need to execute at least until `delivery_wiz.carrier_id = order.carrier_id` happens, so cannot apply the continue above

sale_order_carrier_auto_assign/models/sale_order.py
Shide: Playing with the onchange and auto-rate updates gives me an error when the line is recreated. I think the best option will be to remove completely the onchange and change it to a computed method.
All the problems with the onchanges would be dropped. Maybe in a later PR can address that.

jbaudoux: There was this attempt but it wasn't finalized https://github.com/OCA/sale-workflow/pull/3986

jbaudoux (approved):
Thanks

chienandalu (approved):

Shide:
@jbaudoux would you like to merge this PR? Or do you need more reviews?

jbaudoux:
> @jbaudoux would you like to merge this PR? Or do you need more reviews?

I'm not sales PSC, I can't merge

pedrobaeza (approved):
/ocabot merge patch

sale_order_carrier_auto_assign/views/sale_order_views.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_order_form_with_carrier" model="ir.ui.view">
+        <field name="name">delivery.sale.order.form.view.with_carrier</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="delivery.view_order_form_with_carrier" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='delivery_set']" position="after">
+                <!-- carrier_id must exist in view to be able to be wrote it in onchange -->
```
simahawk, line 9: Late to the party as I'm late to finalize #3986 😓 

This is just an hack, but I guess you know it :) . We cannot rely on the onchange for such functional behavior.
In fact, the onchange is useless and that's what I've changed in the other PR.

PR #4131, opened 2026-01-28
[FIX] sale_order_mass_action: Don't use SavepointCase anymore

rousseldenis:


rousseldenis:
Merging this to avoid warning logs

/ocabot merge patch

alexey-pelykh (approved):
Thanks for this cleanup! The switch from `SavepointCase` to `BaseCommon` is the right call for 16.0 — it silences the deprecation warnings and `BaseCommon` already takes care of `tracking_disable`, so removing that manual context override is correct. All CI green, LGTM.

PR #4135, opened 2026-01-30
[18.0][FIX] sale_advance_payment: Add check for paid orders in portal

ellbristow:
Prevent 'Accept & Pay' button for fully paid orders.

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -97,3 +97,9 @@ class SaleOrder(models.Model):
             order.payment_line_ids = mls
             order.amount_residual = amount_residual
             order.advance_payment_status = payment_state
+
+    def _has_to_be_paid(self):
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the portal
+        # This prevents that, as you do't want customers payign for orders that are already paid
+        res = super()._has_to_be_paid()
+        return res and self.amount_residual > 0
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -99,7 +99,8 @@ class SaleOrder(models.Model):
             order.advance_payment_status = payment_state
 
     def _has_to_be_paid(self):
-        # If an order 'has to be paid' it will show the "Accept & Pay" button in the portal
-        # This prevents that, as you do't want customers payign for orders that are already paid
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the
+        # portal. This prevents that, as you do't want customers payign for orders that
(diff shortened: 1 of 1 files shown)


flotho (approved):
code review LGTM
Thanks

sale_advance_payment/models/sale.py
```diff
@@ -97,3 +97,10 @@ def _compute_advance_payment(self):
             order.payment_line_ids = mls
             order.amount_residual = amount_residual
             order.advance_payment_status = payment_state
+
+    def _has_to_be_paid(self):
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the
+        # portal. This prevents that, as you do't want customers payign for orders that
+        # are already paid
+        res = super()._has_to_be_paid()
+        return res and self.amount_residual > 0
```
alexey-pelykh, line 106: This raw float comparison could be fragile. Due to floating-point arithmetic, `amount_residual` might end up as a tiny non-zero value (e.g. `1e-10`) even when the order is logically fully paid.

Consider using `float_compare` with the currency precision, consistent with how `_compute_advance_payment` handles it:

```python
from odoo.tools import float_compare

def _has_to_be_paid(self):
    res = super()._has_to_be_paid()
    return res and float_compare(
        self.amount_residual, 0.0,
        precision_rounding=self.currency_id.rounding
    ) > 0
```

sale_advance_payment/models/sale.py
```diff
@@ -97,3 +97,10 @@ def _compute_advance_payment(self):
             order.payment_line_ids = mls
             order.amount_residual = amount_residual
             order.advance_payment_status = payment_state
+
+    def _has_to_be_paid(self):
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the
```
alexey-pelykh, line 102: Small typos: "do't" → "don't" and "payign" → "paying".

alexey-pelykh (commented):
Thanks for this, good catch on the portal button showing for fully paid orders.

One thing I'd flag: the `amount_residual > 0` comparison uses a raw float check. Elsewhere in this same module (in `_compute_advance_payment`), amounts are compared using `float_compare` with `precision_rounding=order.currency_id.rounding` — and for good reason, since floating-point arithmetic can leave tiny residual values that aren't meaningfully non-zero. I'd suggest using `float_compare` here too for consistency and safety.

Also, minor typos in the comment: "do't" → "don't", "payign" → "paying".

A small test covering this behavior would be a nice addition, but I understand if you want to keep the scope focused.

rousseldenis (approved):

PR #4136, opened 2026-01-30
[19.0][FIX] sale_advance_payment: Add check for paid orders in portal

ellbristow:
Prevent 'Accept & Pay' button for fully paid orders.

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -105,3 +105,10 @@ class SaleOrder(models.Model):
             order.payment_line_ids = mls
             order.amount_residual = amount_residual
             order.advance_payment_status = payment_state
+
+    def _has_to_be_paid(self):
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the
+        # portal. This prevents that, as you do't want customers payign for orders that
+        # are already paid
+        res = super()._has_to_be_paid()
+        return res and self.amount_residual > 0


alexey-pelykh (approved):
Thanks for the fix! The `_has_to_be_paid` override is the right approach — checking `amount_residual > 0` prevents the "Accept & Pay" button from appearing on already-paid orders. Clean and minimal.

LGTM.

sale_advance_payment/models/sale.py
```diff
@@ -105,3 +105,10 @@ def _compute_advance_payment(self):
             order.payment_line_ids = mls
             order.amount_residual = amount_residual
             order.advance_payment_status = payment_state
+
+    def _has_to_be_paid(self):
+        # If an order 'has to be paid' it will show the "Accept & Pay" button in the
+        # portal. This prevents that, as you do't want customers payign for orders that
+        # are already paid
+        res = super()._has_to_be_paid()
+        return res and self.amount_residual > 0
```
rousseldenis, line 114: I'm not sure this should be in this module.

rousseldenis: No, it's ok

rousseldenis:
@ellbristow Could you squash your commits ? Thanks

sale_advance_payment/models/sale.py
ellbristow: Without it, portal users are prompted to pay for an order/invoice that is already fully paid

ellbristow:
> @ellbristow Could you squash your commits ? Thanks

Sorry, done

PR #4137, opened 2026-02-02
[18.0][FW][IMP] sale_automatic_workflow: add option for default payment journal

JordiMForgeFlow:
fw of https://github.com/OCA/sale-workflow/pull/3689 and also ensure module works properly when payments don't generate journal entries and using enterprise.

CC @kevinkhao


gbrito (approved):

sale_automatic_workflow/views/sale_workflow_process_views.xml
```diff
@@ -277,6 +277,19 @@
...
+                        <div
+                            class="row"
+                            groups="account.group_account_invoice"
+                            domain="[('type', '=', 'sale')]"
+                        >
+                            <div class="col-sm-12">
+                                <label
+                                    for="property_payment_journal_id"
+                                    class="col-lg-4 o_light_label"
+                                />
```
alexey-pelykh, line 285: This `domain` attribute sits on a `<div>`, which means the Odoo web client silently ignores it. Additionally, the filter `[('type', '=', 'sale')]` is inherited from the sales journal block above, but for a *payment* journal the correct filter would be `[('type', 'in', ('bank', 'cash'))]`.

Consider moving the domain to the `<field>` tag directly:
```xml
<field name="property_payment_journal_id" nolabel="1" domain="[('type', 'in', ('bank', 'cash'))]" />
```

(The same issue applies to `property_journal_id` above, but that's pre-existing.)

alexey-pelykh (commented):
Thanks for the forward-port and the enterprise compatibility fix.

The code changes look solid overall — using `_get_valid_payment_account_types()` instead of hardcoded account types is the right approach for 18.0, and the domain filter tightening from `!=paid` to `=not_paid` is a good catch.

One thing worth looking at in the view: the `domain` attribute on the wrapping `<div>` for the new payment journal field filters for `type=sale`, but a payment journal should really be filtered to `bank` or `cash` type journals. The same issue exists on the existing sales journal `<div>` above — but in that case `type=sale` is actually correct. More importantly, `domain` on a `<div>` is silently ignored by the web client, so neither field is actually filtered in the UI. It might be better to move the domain to the `<field>` element itself with the correct type filter.

rousseldenis (approved):

sale_automatic_workflow/views/sale_workflow_process_views.xml
rousseldenis: @JordiMForgeFlow Indeed, domain seems misplaced

JordiMForgeFlow:
@OCA/crm-sales-marketing-maintainers @rousseldenis could this one be merged?

rvalyi:
As per reviews
/ocabot merge minor

PR #4140, opened 2026-02-03
[17.0][ADD] portal_sale_confirm_require_login: New module

pilarvargas-tecnativa:
@Tecnativa TT60285

@carlos-lopez-tecnativa @victoralmau please review


portal_sale_confirm_require_login/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "sale.order"
+
+    def _get_mandatory_partner_fields(self, country_id=False):
+        required = ["phone", "vat", "street", "city", "country_id"]
+        if country_id:
+            country = self.env["res.country"].browse(country_id)
+            if country.state_required:
+                required += ["state_id"]
+            if country.zip_required:
+                required += ["zip"]
```
carlos-lopez-tecnativa: I see that you considered the country configuration, and that is OK, but I am not sure whether Odoo should handle this case or whether the field should be required in the portal view. I attached a video where I do not set the `ZIP code`(or `VAT`) and repeat the steps again, but I expected some message or the field to be marked in red (as required) to indicate that this field is mandatory, like other required fields. As a user, I can be confused because I do not know which field I am missing to fill in.

https://github.com/user-attachments/assets/8546505e-5d52-4234-8727-4fe90c4ddd5e

carlos-lopez-tecnativa (changes requested):

portal_sale_confirm_require_login/models/sale_order.py
pilarvargas-tecnativa: Thanks for the feedback, you are right.

Although Odoo can mark ZIP code as required depending on the country configuration, it is not strictly necessary to confirm a quotation from a billing point of view, and enforcing it here without proper UI feedback can be confusing for the user.

The main requirement we need to cover in this flow is the VAT, which is already enforced. Therefore, we will remove the ZIP requirement from this validation and keep the mandatory fields limited to the essential partner information.

If in the future there is a real need to enforce additional fields (such as ZIP) with proper user feedback in the portal, this can be addressed as a separate improvement.

alexey-pelykh (commented):
Good idea — enforcing login before portal quotation access closes a real security gap, and the mandatory partner field check is a nice addition.

The module would benefit from HttpCase tests covering the main flows (public user redirect to login, redirect to signup, partner field validation). CI codecov is failing which confirms no test coverage yet.

portal_sale_confirm_require_login/controllers/portal.py
```diff
@@ -0,0 +1,112 @@
...
+        # If disabled, public users can access the portal sale document page.
+        # Login is still enforced on protected actions.
+        login_required = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "portal_sale_confirm_require_login.portal_sale_access_login_required",
+            )
+        )
+        login_required = login_required == "True"
```
carlos-lopez-tecnativa, line 63: ????

portal_sale_confirm_require_login/controllers/portal.py
```diff
@@ -0,0 +1,112 @@
...
+            return request.redirect("/my")
+        partner = order_sudo.partner_id
+        redirect_url = f"/my/orders/{order_id}?access_token={access_token}"
+        portal_user = self._get_portal_user(partner)
+        if portal_user:
+            qs = url_encode(
+                {"login": portal_user.login or "", "redirect": redirect_url}
+            )
+            return request.redirect(f"/web/login?{qs}")
+        return request.redirect(self._get_signup_url(partner, redirect_url))
```
carlos-lopez-tecnativa, line 120: Sorry, but if I understand correctly, this is a **high security risk**. I create a sales order for any user (for example, demo, after setting an email), send this sales order by email, and copy the link to access the sales order. When I am not logged in, it redirects me to signup and **allows me to change the password without asking for the current password**.

I think this should redirect to the login page instead of the password change page.

Image

carlos-lopez-tecnativa (approved):

portal_sale_confirm_require_login/controllers/portal.py
pilarvargas-tecnativa: get_param() always returns strings, so the comparison against "True" is needed to convert the config parameter into a boolean value.

portal_sale_confirm_require_login/controllers/portal.py
pilarvargas-tecnativa: Good catch, thanks! I changed the logic to check for any active user linked to the partner instead of only portal users. If a user already exists, it now redirects to the login page and only falls back to signup when the partner has no active users associated.

portal_sale_confirm_require_login/controllers/portal.py
```diff
@@ -0,0 +1,111 @@
...
+
+from odoo.addons.sale.controllers.portal import CustomerPortal
+
+
+class CustomerPortal(CustomerPortal):
+    def _get_portal_user(self, partner):
+        """Return an active portal user for partner, or an empty recordset."""
+        return partner.user_ids.filtered(
+            lambda u: u.active and u.has_group("base.group_portal")
+        )[:1]
```
carlos-lopez-tecnativa: This function is not used anymore, so I think it can be removed.

carlos-lopez-tecnativa (approved):
LGTM, just a minor technical change.

portal_sale_confirm_require_login/controllers/portal.py
pilarvargas-tecnativa: Done, thanks!

victoralmau (approved):

eduezerouali-tecnativa (approved):
LGTM, just that little comment.

portal_sale_confirm_require_login/tests/test_portal_sale_confirm_require_login.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+    def test_existing_user_redirected_to_login(self):
+        self.env["res.users"].with_context(no_reset_password=True).create(
+            {
+                "name": self.partner.name,
+                "login": self.partner.email,
+                "email": self.partner.email,
+                "partner_id": self.partner.id,
+                "groups_id": [(6, 0, [self.env.ref("base.group_portal").id])],
```
eduezerouali-tecnativa: Can you use Command please.

pilarvargas-tecnativa: thanks

pilarvargas-tecnativa:
ping @pedrobaeza

pilarvargas-tecnativa:
This module has been running in a production environment for quite some time. Could it be merged?

PR #4143, opened 2026-02-04
[ADD] sale_product_multiple_qty

yankinmax:
This module adds a **Sales Multiple** unit of measure on products.

When a Sales Multiple is set, sales order line quantities are automatically
rounded **UP** to the nearest multiple of the selected unit of measure.
This is useful for products that must be sold in fixed pack sizes
(boxes, bundles, pallets, etc.).

The rounding is performed by converting the entered quantity to the Sales
Multiple UoM, rounding the number of packs **UP** to the nearest integer,
and converting the result back to the order line UoM.

Being said, if sales multiple UoM is divisible by the order line UoM, the result will be an integer.
Otherwise, we expect rounding issues.

For example (compatible UoMs: 100 is divisible by 5):
- order line UoM: *Pack of 5* (5 units)
- sales multiple UoM: *Pack of 100* (100 units)
- ordering 15 packs of 5 units (75 units) is rounded to 20 packs (100 units);
- ordering 55 packs of 5 units (275 units) is rounded to 60 packs (300 units).

For example (fractional result: 100 is not divisible by 6):
- order line UoM: *Pack of 6* (6 units)
- sales multiple UoM: *Pack of 100* (100 units)
- ordering 13 packs of 6 units (78 units) is rounded to 16.67 packs (100.02 units);

If the Sales Multiple UoM is not divisible by the order line UoM, the rounded
quantity may be fractional. In such cases, the module displays a warning suggesting the nearest valid integer and
prevents saving invalid quantities. It is the user's responsibility to
configure compatible units of measure.


sale_product_multiple_qty/models/product_product.py
```diff
@@ -0,0 +1,65 @@
...
+
+        The packaging reference unit (``relative_uom_id``) is the product base UoM.
+        """
+        uom = self.env["uom.uom"]
+        for product in self:
+            if not product.uom_id:
+                product.allowed_sale_multiple_uom_ids = False
+                continue
+            product.allowed_sale_multiple_uom_ids = uom.search(
+                Domain("relative_uom_id", "=", product.uom_id.id),
```
ivantodorovich: I'm not so sure about this computation.

`relative_uom_id` is optional, and it seems to act as a `parent_id` field to form a hierarchy of uoms.

So, can you test this use case?

Uoms:
- Unit(s)
- Box of 10
- Box of 100
- Box of 500

Product `uom_id` = Box of 10

Then, are you allowed to select the multiple Box of 100?
Cause if my concern is correct, then that's not possible.. and it should.

> [!TIP]
> In fact, can you add a unit tests with the case I just mentioned, or similar?

sale_product_multiple_qty/models/product_product.py
```diff
@@ -0,0 +1,65 @@
...
+        for product in self:
+            multiple_uom = product.sale_multiple_uom_id
+            if not multiple_uom:
+                continue
+            multiple_factor = multiple_uom.relative_factor
+            for uom in product.uom_ids:
+                if (
+                    # We can safely divide by multiple_factor
+                    # as it cannot be 0 (constraint on uom.uom)
+                    uom.relative_factor % multiple_factor != 0
```
ivantodorovich: This is only correct if the `relative_uom_id` are the same.
The `relative_factor` makes only sense when paired with a `relative_uom_id`.

You'd be closer with the `factor`, instead.. that's supposed to be the factor relative to the root uom (the ancestor uom with relative_uom_id = False). But that's not totally true.. and it would still fail.

Take for example this use case:

Packaging uom:
- Box of 250 (factor = 250)

Multiple uom:
- Box of 500 (factor = 500)


With the existing code, it would fail because `250 % 500 != 0`..

But this should be allowed, and what should really happen is that if you try to order 1 Box of 250, it should automatically be converted to 2 Boxes of 250 (to round up to 500 units)


> [!TIP]
> In fact, can you add a unit tests with the case I just mentioned, on the SO line?

----

I'm not totally sure this constraint makes sense anymore.
What is the use case we were trying to avoid @gaetan-vaujour ?

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,91 @@
...
+from odoo.tools import float_compare
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _round_sale_qty_to_multiple(self, qty_to_order: float) -> float:
+        """Round qty in product UoM to a multiple of the sales multiple UoM.
+
+        :param qty_to_order: quantity expressed in product.uom_id (base UoM).
```
ivantodorovich: That's not true. The quantity is expressed in `product_uom_id` (the sale order line UOM)

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,91 @@
...
+        :return: rounded quantity expressed in product.uom_id.
```
ivantodorovich: Also not true. It's expressed in the sale order uom

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,91 @@
...
+
+            new_qty = line._round_sale_qty_to_multiple(qty_to_order)
+            if (
+                float_compare(
+                    new_qty,
+                    qty_to_order,
+                    precision_rounding=line.product_uom_id.rounding,
+                )
+                != 0
+            ):
```
ivantodorovich: 💡 Tip: simpler

```suggestion
            if line.product_uom_id.compare(new_qty, qty_to_order) != 0:
```

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,91 @@
...
+                continue
+
+            qty_to_order = line.product_uom_qty or 0.0
+            if qty_to_order <= 0:
+                continue
+            # Get the nearest valid multiple quantity
+            valid_multiple_qty = line._round_sale_qty_to_multiple(qty_to_order)
+            # Entered quantity should be equal to the valid multiple quantity
+            if (
+                float_compare(
```
ivantodorovich: same tip as before

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,91 @@
...
+                )
+                != 0
+            ):
+                multiple = line.product_id.sale_multiple_uom_id.name
+                raise ValidationError(
+                    self.env._(
+                        "The qty {qty:.2f} is not valid for product '{product}'.\n"
+                        "It should be a multiple of {multiple} "
+                        "(valid qty: {valid_qty:.2f})."
+                    ).format(
```
ivantodorovich: Use the translation method built-in string interpolation instead

sale_product_multiple_qty/models/product_product.py
```diff
@@ -0,0 +1,65 @@
...
+                ):
+                    raise ValidationError(
+                        self.env._(
+                            "The UoM '{uom}' cannot be used for the product '{pt}' "
+                            "because its conversion ratio ({uom_factor:.2f}) "
+                            "is not a multiple of the sales multiple UoM "
+                            "'{multiple}' conversion ratio ({multiple_factor:.2f})."
+                        ).format(
```
ivantodorovich: Use the translation method built-in string interpolation instead

ivantodorovich (changes requested):

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: ok, I agree it's a sale order line UoM, but `product_uom_id` is computed mainly form `product.uom_id`. That's what I meant

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: This is a leftover during the updates yes, thanks

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: Indeed it's better, thanks

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: I've rewritten it several times to satisfy OCA linting limitations :smiley:

sale_product_multiple_qty/models/product_product.py
yankinmax: I see what you mean: the domain is perfectly fine if we always select "Units" as a `product.uom_id`, but if we, for example, take a "Box of 10" and create a new "Box of 100" with a `relative_uom_id` as "Box of 10", aka 10 boxes of 10.
How should be covered such case then?

sale_product_multiple_qty/models/sale_order_line.py
ivantodorovich: No, the `product_uom_id` is the UOM selected in the sale.order.line.
It defaults to `product.uom_id`, but it can be different if the user selects it

sale_product_multiple_qty/models/sale_order_line.py
ivantodorovich: either `self.env._("Hey %(name)s", name="Maksym")` or `self.env._("Hey %s", "Maksym")`
Both satisfy the OCA and it's the correct way to interpolate translatable terms in recent versions of Odoo.

yankinmax: Ok, but I thought `.format()` is applicable option too

ivantodorovich: format would fail if there are missing placeholders in the translated term.

e.g.:
```
>>> "{}, {}".format("asd")
  File "<python-input-6>", line 1, in <module>
    "{}, {}".format("asd")
    ~~~~~~~~~~~~~~~^^^^^^^
IndexError: Replacement index 1 out of range for positional args tuple
```

Using the built-in interpolation, Odoo can recover gracefully:
https://github.com/odoo/odoo/blob/1309b18ec0b4bac47a403979b206222be15870a3/odoo/tools/translate.py#L439-L441

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+            "The nearest valid qty is %(valid_qty).2f.",
+            entered_qty=entered_qty,
+            rounded_qty=rounded_qty,
+            product=self.product_id.display_name,
+            multiple=multiple_uom.display_name,
+            valid_qty=valid_qty,
+        )
+
+    @api.onchange("product_id", "product_uom_qty", "product_uom_id")
+    def _onchange_product_uom_qty(self):
```
ivantodorovich: This is a very generic method name for an onchange on a standard field.
If all modules did this, we would get conflicts



```suggestion
    def _onchange_product_uom_qty_round_multiple(self):
```

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+    def _nearest_valid_integer_qty(self, entered_qty: float) -> float:
+        """Return the nearest valid integer qty bigger than entered (UP).
+
+        The idea is simply to try integer quantities bigger than the entered
+        quantity until we find one that, when rounded to the sales multiple UoM,
+        gives an integer result.
+        """
+        self.ensure_one()
+        qty_to_check = int(entered_qty) + 1
+        while True:
```
ivantodorovich: This is too dangerous.

Are you sure this loop will always succeed in finding an integer multiple?
At what cost? how many iterations can it take?

And, this is only to show a warning message.. ?

My recommendation: get rid of it

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+    ) -> str:
+        """Prepare the warning message for incompatible UoMs.
+
+        :param entered_qty: quantity entered by the user.
+        :param rounded_qty: rounded quantity to the sales multiple UoM.
+        :return: warning message string.
+        """
+        self.ensure_one()
+        multiple_uom = self.product_id.sale_multiple_uom_id
+        valid_qty = self._nearest_valid_integer_qty(entered_qty)
```
ivantodorovich: See my commend above

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+    def _onchange_product_uom_qty(self):
+        """Round product_uom_qty to a multiple of the sales multiple UoM.
+
+        If sales multiple UoM is set on the product, this onchange rounds
+        the order line quantity to the nearest multiple of that UoM using
+        the "_round_sale_qty_to_multiple" method.
+
+        If the rounded quantity is a floating point number, it means that the
+        selected UoM is not compatible with the sales multiple UoM.
+        See an example in the "_round_sale_qty_to_multiple" method.
```
ivantodorovich: This statement is wrong IMO.

Makes sense for discrete uom like the common "Unit(s)", but not dimensional uoms like "Kg(s)", "Meter(s)", etc.. where a floating point number is perfectly acceptable.

Suggestion below: use `_has_common_reference`

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+            multiple_uom = line.product_id.sale_multiple_uom_id
+            if not multiple_uom:
+                continue
+
+            qty_to_order = line.product_uom_qty or 0.0
+            if qty_to_order <= 0:
+                continue
+
+            rounded_qty = line._round_sale_qty_to_multiple(qty_to_order)
+            if not rounded_qty.is_integer():
```
ivantodorovich:
```suggestion
            if line.product_uom_id._has_common_reference(uom_unit) and not rounded_qty.is_integer():
```

some lines above, before the loop starts:

```py
uom_unit = self.env.ref("uom.product_uom_unit")
```

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,156 @@
...
+                msg = line._prepare_incompatible_uom_warning_msg(
+                    entered_qty=qty_to_order,
+                    rounded_qty=rounded_qty,
+                )
+                raise ValidationError(msg)
```
ivantodorovich: This is wrong.

it should be doing something like this:


```suggestion
            if line.product_uom_id.compare(rounded_qty, qty_to_order) != 0:
                raise ValidationError(self.env._("some message about qty not being a multiple"))
```

ivantodorovich (changes requested):

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: Makes sense, thx

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: The idea was to prepare the cleanest message to the user, helping him to get the nearest valid integer if rounding issues happen.
Not necesary of course

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -148,9 +94,16 @@ def _check_rounded_sale_multiple_qty(self):
...
             rounded_qty = line._round_sale_qty_to_multiple(qty_to_order)
-            if not rounded_qty.is_integer():
-                msg = line._prepare_incompatible_uom_warning_msg(
+            qty_diff = line.product_uom_id.compare(rounded_qty, qty_to_order)
+            if qty_diff != 0:
+                msg = self.env._(
+                    "The entered qty %(entered_qty).2f is rounded to "
+                    "%(rounded_qty).2f which is not valid "
+                    "for product '%(product)s'.\n"
+                    "It should be a multiple of %(multiple)s.\n",
```
ivantodorovich: Here the message should say:

"The entered quantity %(entered_qty)s is not a multiple of %(multiple)s"

Why:
- The rounded quantity is only use for checking whether it's a multiple or not. Whatever triggered the error is not aware of it.

Moreover:
- Do not use `.2f` in the message. That would hardcode the precision, when in fact it's configurable. Since the `product_uom_qty` already has `digits='Product Unit'`, the value would already be rounded so normally it's not needed. But you could use `float_repr` with `precision_digits=self.env['decimal.precision'].precision_get('Product Unit')` to be sure there are no floating point representation errors

ivantodorovich (approved):

alexey-pelykh (commented):
Thanks for the clean implementation and for incorporating all the review feedback.

The module is well-structured and the rounding logic is solid. CI passes (tests + pre-commit), and ivantodorovich already approved. Just a couple of minor observations on the current state:

**Dead code in tests**

`_get_invalid_qty_warning_msg` (and the `formatLang` import) in the test file appear to be leftover from a previous iteration that included validation/warning logic. Since the constraint was removed in favor of silent onchange rounding, this helper is never called. Worth cleaning up to avoid confusion.

**Template compute/inverse test coverage**

The `product_template.py` compute/inverse logic for syncing `sale_multiple_uom_id` between template and single-variant product is not exercised by the current tests. A small test that sets the field on a template and verifies it propagates to the variant (and vice versa) would strengthen coverage and likely address the codecov gaps.

Neither of these is blocking -- the core functionality is correct and well-tested.

sale_product_multiple_qty/tests/test_sale_product_multiple_qty.py
```diff
@@ -0,0 +1,137 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import Command
+from odoo.tests import Form, TransactionCase
+from odoo.tools.misc import formatLang
```
alexey-pelykh: This import is unused -- `_get_invalid_qty_warning_msg` (which references `formatLang`) is never called by any test. Looks like a leftover from the validation/warning iteration. Safe to remove along with the helper method.

sale_product_multiple_qty/tests/test_sale_product_multiple_qty.py
```diff
@@ -0,0 +1,137 @@
...
+            Command.link(cls.uom_pack_5.id),
+            Command.link(cls.uom_pack_100.id),
+        ]
+        cls.product_screw.sale_multiple_uom_id = cls.uom_pack_100
+        # Set allowed UoMs and sales multiple UoM for tables
+        cls.product_table.uom_ids = [
+            Command.link(cls.uom_pack_10.id),
+            Command.link(cls.uom_pack_100.id),
+        ]
+        cls.product_table.sale_multiple_uom_id = cls.uom_pack_100
```
alexey-pelykh, line 48: This helper method is never called by any test -- it appears to be a remnant from when the module raised a warning/validation error. Consider removing to keep the test file clean.

yankinmax:
Hello @alexey-pelykh 
Thanks for your complete review on this PR.
1. The idea was to add a constraint on the SOL to ensure `product_uom_qty` is a multiple of the sales multiple UoM. But, the problem is that constraint check will be only called only when the onchange rounding is already happend. Thus we don't want to round up twice numbers already rounded. It becomes non stable validation and on floating point numbers can result in unexpected results. We decided to remove it from this PR. 
If you have any idea how to correctly handle such validation we're open to suggestions.
Here is the PR where we try an idea to round up with "HALF-EVEN" to check if rounding is already applied, before actually proceed to round "UP" => https://github.com/OCA/sale-workflow/pull/4151
2. Regarding the test to check `sale_multiple_uom_id` on the template and product.
This idea is inspired from the standard `website_sale` flow of `product.template::base_unit_id` which is covered by the tests:
https://github.com/odoo/odoo/blob/d038b2c23b9f7c74d381e25276d00155bf89331e/addons/website_sale/models/product_template.py#L220-L229
If I have some extra time in future I indeed can add a check for this, but it's not really needed I think.
3. I've cleaned the leftovers.

sale_product_multiple_qty/models/product_product.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    sale_multiple_uom_id = fields.Many2one(
+        comodel_name="uom.uom",
+        string="Sales Multiple",
+        help="When set, sale order quantities are rounded up to an "
+        "multiple number of this unit.",
```
vvrossem, line 13:
```suggestion
        help="When set, sale order quantities are rounded up to a "
        "multiple number of this unit.",
```

sale_product_multiple_qty/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+    _inherit = "product.template"
+
+    sale_multiple_uom_id = fields.Many2one(
+        comodel_name="uom.uom",
+        string="Sales Multiple",
+        compute="_compute_sale_multiple_uom_id",
+        inverse="_inverse_sale_multiple_uom_id",
+        store=True,
+        help="When set, sale order quantities are rounded up to an "
+        "multiple number of this unit.",
```
vvrossem, line 16:
```suggestion
        help="When set, sale order quantities are rounded up to a "
        "multiple number of this unit.",
```

sale_product_multiple_qty/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,21 @@
...
+Multiple UoM, rounding the number of packs **UP**, and converting the result back to the order line UoM.
+
+For example, with a Sales Multiple of *Pack of 100*:
+- ordering 15 packs of 5 units (75 units) is rounded to 20 packs (100 units);
+- ordering 55 packs of 5 units (275 units) is rounded to 60 packs (300 units).
+
+If the Sales Multiple UoM is not divisible by the order line UoM, the rounded
+quantity may be fractional. It is the user's responsibility to
+configure compatible units of measure.
+
```
vvrossem, line 21:
```suggestion
```

vvrossem (approved):
LGTM :+1:

cyrilmanuel (approved):

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        self.ensure_one()
+        multiple_uom = self.product_id.sale_multiple_uom_id
+        packs = self.product_uom_id._compute_quantity(qty_to_order, multiple_uom)
+        packs = fields.Float.round(packs, precision_digits=0, rounding_method="UP")
+        qty_rounded = multiple_uom._compute_quantity(packs, self.product_uom_id)
+        return qty_rounded
+
+    @api.onchange("product_id", "product_uom_qty", "product_uom_id")
+    def _onchange_product_uom_qty_round_multiple(self):
```
rousseldenis, line 133: IMHO, as `product_uom_qty` is a computed field and onchanges should be avoided, this should be changed to `compute_product_uom_qty()`

rousseldenis (changes requested):

rousseldenis:
@yankinmax

sale_product_multiple_qty/models/sale_order_line.py
yankinmax: @ivantodorovich what do you think?

ivantodorovich: AFAIR we used onchange to not have this UX rounding play out for programatic assignation, and keep it only for forms.

In case of programatic assignation, no autofix is done and the constraint plays out -- which is better to detect possible errors

yankinmax: Our goal is to propose the rounded value to a user. That's it.
AFAIR, the input value still can be used and there is no real impact anywhere except the forms.
Moreover, we dropped an idea to have any constraints here to allow rounding with decimals, because only the UoM's that share "Unit" as a common reference and have an integer as a sales multiple.

We probably might add a tiny improvement for the rounding, so the UoM's that share "Unit" as a common reference are rounded to the nearest multiple integer.

So, this example:

order line UoM: Pack of 6 (6 units)
sales multiple UoM: Pack of 100 (100 units)
ordering 13 packs of 6 units (78 units) is rounded to 16.67 packs (100.02 units);

Will have the rounded value 17.

cyrilmanuel:
hi @rousseldenis is it good for you ?

cyrilmanuel (approved):

divad1196:
How does it compare to Odoo's native [product packaging](https://www.odoo.com/documentation/19.0/applications/inventory_and_mrp/inventory/product_management/configure/packaging.html) ? (see [Sales unit of measure (packagings)](https://www.odoo.com/documentation/19.0/applications/inventory_and_mrp/inventory/product_management/configure/uom.html?highlight=sale%20uom#sales-unit-of-measure-packagings)

sale_product_multiple_qty/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+    @api.depends("product_variant_ids", "product_variant_ids.sale_multiple_uom_id")
+    def _compute_sale_multiple_uom_id(self):
+        self.sale_multiple_uom_id = self.env["uom.uom"]
+        for template in self.filtered(
```
divad1196: Working but we could write it the same way as the `_inverse_` function. (i.e. filtering and assignation in loop). Should be a bit more readable.

Also, would it suit to use `product_variant_id` instead of `product_variant_ids` ?

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,152 @@
...
+
+            - sales multiple UoM: 1 kg
+              order line UoM: 400 g
+              step quantity: 2.5
+        """
+        self.ensure_one()
+        multiple_uom = self.product_id.sale_multiple_uom_id
+        step_qty = multiple_uom._compute_quantity(1.0, self.product_uom_id)
+        if self._sale_multiple_uom_has_unit_reference():
+            step_qty = math.ceil(step_qty)
```
divad1196: We should avoid using `math.ceil` as it's easy to have "leftovers/artifcats".
`math.ceil(0.1 + 0.2 - 0.3) == 1.0`

This code is coming from another module:
```python
converted_length = int(float_round(converted_length, precision_rounding=1.0, rounding_method='UP')
```

sale_product_multiple_qty/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+            lambda template: len(template.product_variant_ids) == 1
```
divad1196: Would recommend to use the same approach as in `_inverse` method, i.e. loop + if/else. Better for readability, no double assign.

Could using `product_variant_id` (without trailing `s`) instead be acceptable ?

sale_product_multiple_qty/models/sale_order_line.py
```diff
@@ -0,0 +1,152 @@
...
+            if multiple_uom.compare(rounded_multiple, multiple) == 0:
+                return qty_to_order
+
+        packs = self.product_uom_id._compute_quantity(qty_to_order, multiple_uom)
+        packs = fields.Float.round(packs, precision_digits=0, rounding_method="UP")
+        qty_rounded = multiple_uom._compute_quantity(packs, self.product_uom_id)
+        # For UoMs sharing ``Units`` as common reference,
+        # we ceil the result to keep an integer count of items or packs.
+        if self._sale_multiple_uom_has_unit_reference():
+            qty_rounded = math.ceil(qty_rounded)
```
divad1196: Same remark as before about `math.ceil`, especially since you did it above with `fields.Float.round`.

Why not do it here as well ?

divad1196 (commented):
- `math.ceil` to avoid
- question about design choice to only consider product.template with a single variant

yankinmax:
Hello @divad1196 
I've addressed the PR suggestions.
Thank you for the deep review and useful suggestions.

yankinmax:
> How does it compare to Odoo's native [product packaging](https://www.odoo.com/documentation/19.0/applications/inventory_and_mrp/inventory/product_management/configure/packaging.html) ? (see [Sales unit of measure (packagings)](https://www.odoo.com/documentation/19.0/applications/inventory_and_mrp/inventory/product_management/configure/uom.html?highlight=sale%20uom#sales-unit-of-measure-packagings)

Native product packaging helps identify/select packaging, but this module targets a different UX: it rounds the sale order line quantity itself to a product-defined sales multiple.

So for example, if the line UoM is “Pack of 5” and the product sales multiple is “Pack of 100”, entering 15 packs is proposed as 20 packs. The logic is based on UoM conversion and is onchange/UI-only; it does not enforce or rewrite programmatic quantities.

divad1196:
@yankinmax Thank you for the changes and answer.

To clarify my question, I understood the general idea of editing the lines, but it looked like a way to avoid using packagings. I know some customers can find them "too complex".

UoM of kind "Group/Pack of N", like in the example, is already a way to avoid that. It can work, but it can also become a debt. You development is IMO more future-proof than using "pack UoM".

From your explanation, I understand that one advantage of this method is to be less rigid than packaging.
This makes sense.

divad1196 (approved):

yankinmax:
Hello @rousseldenis 
Can you pls take a look again and possibly trigger merge?

cyrilmanuel:
hi @rousseldenis  could you update your review please and merge this one if its all good for you :)

ivantodorovich:
Hello @rousseldenis 👋🏻 are you still blocking this PR or is it ok to you now?
Best

ivantodorovich:
Hello @rousseldenis, are you still blocking this PR? If not can I ask you to update the review? 🙏🏻

PR #4145, opened 2026-02-05
[17.0][FIX] sale_invoice_policy: invoice policy

Drsmartinkus:
Fix sale order invoice policy to work regardless of required setting.

Before this fix, the sale order invoice_policy field only affected invoicing calculations when invoice_policy_required was set to True. When invoice_policy_required was False, the invoice policy was ignored, and the system fell back to the default product level invoice policies.

This was problematic because users could set an invoice policy on sale orders, but it would not work. Now the sale order invoice policy works correctly even when the field is not marked as required, allowing optional but functional invoice policies.

### DIFF
--- a/sale_invoice_policy/models/sale_order_line.py
+++ b/sale_invoice_policy/models/sale_order_line.py
@@ -40,6 +40,7 @@ class SaleOrderLine(models.Model):
         other_lines = self.filtered(
             lambda line: line.product_id.type == "service"
             or line.order_id.invoice_policy == "product"
+            or not line.order_id.invoice_policy
         )
         super(SaleOrderLine, other_lines)._compute_qty_to_invoice()
         for line in self - other_lines:
--- a/sale_invoice_policy/models/sale_order_line.py
+++ b/sale_invoice_policy/models/sale_order_line.py
@@ -18,7 +18,6 @@ class SaleOrderLine(models.Model):
         other_lines = self.filtered(
             lambda line: line.product_id.type == "service"
             or not line.order_id.invoice_policy
-            or not line.order_id.invoice_policy_required
         )
         super(SaleOrderLine, other_lines)._compute_qty_to_invoice()
         for line in self - other_lines:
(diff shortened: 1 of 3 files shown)


sebalix (approved):

alexey-pelykh (approved):
Good fix. The `invoice_policy_required` guard was preventing the policy from being applied when the setting was disabled — exactly the scenario the new test covers. Also removing `readonly=True` from the field makes sense since users should be able to override per-order.

LGTM.

rousseldenis:
@Drsmartinkus Maybe could you review this ? https://github.com/OCA/sale-workflow/pull/3376

Drsmartinkus:
> @Drsmartinkus Maybe could you review this ? #3376

Seems like this one is merged already, but looks good to me. Looks like this guy is tackling the same problem but in abit more complex different way :)

rousseldenis:
@Drsmartinkus Could you squash your commits ? Thanks

rousseldenis:
> > @Drsmartinkus Maybe could you review this ? #3376
> 
> Seems like this one is merged already, but looks good to me. Looks like this guy is tackling the same problem but in abit more complex different way :)

Okay, didn't read this answer. I indeed did those improvements. That PR should be ported in fact. I will do it

rousseldenis:
@Drsmartinkus Could you rebase ?

Drsmartinkus:
@rousseldenis rebased

rousseldenis:
@Drsmartinkus Could you check tests ?

Drsmartinkus:
@rousseldenis I think we can close this PR as this is kind of redundant now since this was introduced 
https://github.com/OCA/sale-workflow/commit/62a1b8ba7074609e8a35f9de5f344acae9ff21c7

as before this commit there were no option to opt out and now I can see from this commit they added this option `product` which is a default behaviour. Since from that commit `invoice_policy_required` is removed these changes are no longer needed as bug was around this field invoice_policy_required.

sale_invoice_policy/models/sale_order_line.py
```diff
@@ -40,6 +40,7 @@ def _compute_qty_to_invoice(self):
         other_lines = self.filtered(
             lambda line: line.product_id.type == "service"
             or line.order_id.invoice_policy == "product"
+            or not line.order_id.invoice_policy
```
rousseldenis, line 43: https://github.com/OCA/sale-workflow/blob/17.0/sale_invoice_policy/models/sale_order.py#L19

The field is required and has always a default value, so this is not needed.

Closing

PR #4153, opened 2026-02-12
[17.0][ADD] sale_order_price_precision: set specific decimal precison for sales prices

dreispt:


rousseldenis:
@dreispt Wouldn't this weird towards other flows (invoicing,... ) that have not this change ?

dreispt:
I think you're right, and this should also be set on the invoice side.
It should be enough to stop there, I can think of other sales related unit price fields.

The underlying pain/motivation is that purchases and costing needs 5 or 6 digits, but sales only needs two.

dreispt:
@rousseldenis Added also on account.move.line
I don't think we're missing any other relevant unit price fields.

alexey-pelykh (changes requested):
Thanks for the contribution! The concept is straightforward and useful.

However, a new module needs tests — at minimum covering that the decimal precision is applied correctly on sale order lines and invoice lines. Also, `README.rst` and `static/description/index.rst` are empty — run `oca-gen-addon-readme` to generate them from the fragments.

PR #4154, opened 2026-02-13
[19.0][MIG] sale_restricted_qty

therbin200:
Migrated from v16, but taken v18 PR as a base (#3959) as it is ready to be merge and contains some minor improvements


Reyes4711-S73 (approved):
LGTM

alexey-pelykh (approved):
Clean 19.0 migration with proper migration scripts handling the column renames and flag conversions. LGTM

Reyes4711-S73:
@OCA/crm-sales-marketing-maintainers Please, can you review/merge this PR?

mmequignon:
You aggregated all the changed from 17.0 to 18.0 in the migration commit, right ?
I don't quite like that, because by doing so, we're losing the git history.
Also, there's a few remarks in the 18.0 pr, especially the one regarding the use of `float_is_zero`.
I would personally wait that 18.0 is merged, which will avoid doing backports.

alexey-pelykh:
@mmequignon Fair points. The 18.0 PR (#3959) should indeed land first — it contains the baseline that 19.0 should be migrated from, and merging 19.0 before 18.0 would lose the intermediate history.

@Reyes4711-S73 I would suggest waiting for #3959 to be merged, then rebasing this PR on top of it. That way the 19.0 migration commit only contains the actual 18.0→19.0 delta, preserving full git history.

alexey-pelykh (commented):
Thanks for the migration work.

After looking at this more carefully alongside the 18.0 PR (#3959), I'm withdrawing my earlier approval. As @mmequignon noted, aggregating all the 17.0-to-18.0 changes into the migration commit loses the intermediate git history. The 18.0 PR is still open and has its own pending feedback (notably the `float_is_zero` remark on the modulo check in `check_restricted_qty` and `_compute_restricted_qty_constraints` -- raw `qty % multiple_of_qty != 0` can misfire on float precision).

I'd recommend waiting for #3959 to land on 18.0 first, then rebasing this PR so the 19.0 migration commit only contains the actual 18.0-to-19.0 delta. That way:
- git history is preserved across versions,
- any fixes applied during 18.0 review (like `float_is_zero`) carry forward automatically, and
- the 19.0 migration path is cleaner.

The module structure, inheritance model, migration scripts, views, and tests all look solid otherwise -- no issues beyond the sequencing concern and the inherited `float_is_zero` gap.

rousseldenis:
@therbin200 I have some concerns with this as there are a lot of changes in the migration commit like @mmequignon said.

Moreover, it seems (personal feeling, no personal offense) that some added tests have been done through AI.

Naming tests with 'coverage' is not correct as they should reflect functional or technical features.

bosd:
@therbin200  Can you rebase and include the latest code from the V18, which is now merged?

Might also want to include open pr #4496

bosd (changes requested):
please base upon the v18 merged cde

PR #4155, opened 2026-02-16
[18.0][ADD] sale_order_forecast_availability

AaronHForgeFlow:
This module flag orders as Forecast available according to the lines:

1.- If one of the lines has red forecast then a warning appears on top:
image

2.- It adds a filter on sales with availability issues according to the forecast:
image


sale_order_forecast_availability/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    sale_forecast_available = fields.Boolean(
+        string="Forecast Available",
+        compute="_compute_sale_forecast_available",
+        store=True,
+        help="True if this line has forecast availability issues",
```
alexey-pelykh: The help string says "True if this line has forecast availability issues" but the field is `True` when all lines have **no** issues (i.e. everything is available). The help should read something like "True if all order lines have sufficient forecast availability" to match the actual semantics.

alexey-pelykh (approved):
Nice module, covers a real need. Hooks for the boolean column are a good call for performance on large datasets. Tests cover the main scenarios well.

One small nit on the help text (inline) — otherwise LGTM.

sale_order_forecast_availability/models/sale_order.py
AaronHForgeFlow: That's accurate. I corrected it. Thank you!

alexey-pelykh (commented):
@AaronHForgeFlow Thanks for the correction!

PR #4156, opened 2026-02-16
[15.0][IMP] sale_planner_calendar: Unsubscribe old salesman from partner

CarlosRoca13:
By checking the options "Assign New Salesperson To Partner" and "Unsuscribe Old Salesperson" the old salesperson of the partner will be unsubscribed from the mail followers.

cc @Tecnativa TT60922


sale_planner_calendar/wizard/sale_planner_calendar_reassign.py
```diff
@@ -117,9 +117,15 @@ def apply(self):
...
                 continue
             if self.assign_new_salesperson_to_partner:
+                old_salesperson = line.partner_id.user_id.partner_id
                 line.partner_id.with_context(
                     skip_sale_planner_check=True
                 ).user_id = line.new_user_id
+                if self.unsuscribe_old_salesperson and not self.new_end:
+                    # When the salesperson is assigned to the partner and
+                    # the option to unsubscribe is selected, the old one
+                    # should be removed from the partner followers.
```
alexey-pelykh, line 127: This should be covered by a test. Consider adding a case to `test_reassign_wizard_apply` (or a new test method) where `unsuscribe_old_salesperson=True` and `new_end` is empty, then assert `old_salesperson.partner_id not in line.partner_id.message_partner_ids` after `apply()`.

sale_planner_calendar/wizard/sale_planner_calendar_reassign.py
```diff
@@ -117,9 +117,15 @@ def apply(self):
             if not line.new_user_id:
                 continue
             if self.assign_new_salesperson_to_partner:
+                old_salesperson = line.partner_id.user_id.partner_id
                 line.partner_id.with_context(
                     skip_sale_planner_check=True
                 ).user_id = line.new_user_id
+                if self.unsuscribe_old_salesperson and not self.new_end:
+                    # When the salesperson is assigned to the partner and
```
alexey-pelykh, line 125: Minor: if the partner currently has no salesperson (`line.partner_id.user_id` is empty), `old_salesperson` will be an empty recordset. The `message_unsubscribe` call won't fail, but you could add `and old_salesperson` to the condition on line 127 for extra safety and clarity.

alexey-pelykh (commented):
Thanks for this improvement!

The logic looks correct -- capturing `old_salesperson` before the assignment and calling `message_unsubscribe` only when `not self.new_end` (immediate reassignment, not deferred) makes sense and aligns with the existing cron-based unsubscribe path for the deferred case.

A couple of observations:

1. **Missing test coverage**: The codecov checks are failing because the new `message_unsubscribe` branch is not exercised by any test. I'd suggest adding a test case (or extending `test_reassign_wizard_apply`) that sets both `assign_new_salesperson_to_partner = True` and `unsuscribe_old_salesperson = True` with `new_end` empty, then asserts that the old salesperson's partner is no longer in `partner.message_partner_ids`. This is important to prevent regressions on this new behavior.

2. **Edge case when partner has no salesperson**: If `line.partner_id.user_id` is empty (no current salesperson on the partner), `old_salesperson` will be an empty recordset. `message_unsubscribe(partner_ids=[])` is harmless, so no crash -- but you may want to guard against it for clarity, e.g.:

   ```python
   if self.unsuscribe_old_salesperson and not self.new_end and old_salesperson:
   ```

These are not blockers but would strengthen the contribution.

sergio-teruel (approved):

BhaveshHeliconia (approved):
LGTM!

carlosdauden (approved):

sergio-teruel:
@CarlosRoca13 Check if you need fw-port to 18.0

PR #4161, opened 2026-02-18
[18.0][MIG] sale_custom_rounding: Migration to 18.0

Anxo82:
@HaraldPanten 

- Replaced the stored compute on tax_calculation_rounding_method with an @api.onchange.
In Odoo 18 the field is defined in core as a company-related field, so the stored compute approach no longer worked reliably (it could be computed too early and stored with a wrong value).

- Removed the override of _compute_tax_totals_json.
In Odoo 18 the tax rounding is handled by the new tax engine and no longer depends on the round context flag, so the custom context-based logic was no longer applicable.

- Simplified the override of _compute_amount on sale.order.line. The method now simply extends the @api.depends to trigger recomputation when the rounding method changes and delegates the computation entirely to the core logic.

- [T-9321]


Jaimermaccione (approved):
@Anxo82 Module functionally tested and reviewed in a local environment.

#### Tests Performed

- Configured a tax rounding method at **contact level** → correctly inherited in the Sales Order.
- Configured a different tax rounding method at **invoicing address level** → invoicing address setting correctly overrides the main contact.
- Generated invoices from multiple Sales Orders with:
  - Different rounding methods
  - Same customer with different rounding configurations

#### Results

As described in the README, the tax rounding method is used as a grouping key during invoice creation. The system correctly generates one invoice per tax rounding method.

#### Conclusion

The README accurately reflects the module behavior.  
Functionality works as expected.

**LGTM.** Thank you!

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@HaraldPanten  Could you please review it ?
Thanks

HaraldPanten (approved):
/ocabot merge nobump

PR #4162, opened 2026-02-18
[17.0][FIX] sale_pricelist_global_rule: Exclude sections and notes from computes

pedrobaeza:
Adapted to 17 from ca5009ef337518c110e2fa5a48f3cd69a8418b2c

@Tecnativa TT61081

### DIFF
--- a/sale_pricelist_global_rule/models/sale_order.py
+++ b/sale_pricelist_global_rule/models/sale_order.py
@@ -19,7 +19,7 @@ class SaleOrder(models.Model):
             "by_template": {},
             "by_categ": {},
         }
-        for line in self.order_line.filtered("product_id"):
+        for line in self.order_line.filtered(lambda x: not x.display_type):
             qty_in_product_uom = line.product_uom_qty
             # Final unit price is computed
             # according to `qty` in the default `uom_id`.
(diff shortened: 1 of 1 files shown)


sale_pricelist_global_rule/models/sale_order.py
```diff
@@ -67,7 +67,8 @@ def button_compute_pricelist_global_rule(self):
         self.order_line.write({"discount": 0.0})
         qty_data = self._get_cummulative_quantity()
         sale_order = self.with_context(pricelist_global_cummulative_quantity=qty_data)
-        sale_order.order_line._compute_pricelist_item_id()
-        sale_order.order_line._compute_price_unit()
-        sale_order.order_line._compute_discount()
+        lines = sale_order.filtered(lambda x: not x.display_type)
```
carlos-lopez-tecnativa:
```suggestion
        lines = sale_order.order_line.filtered(lambda x: not x.display_type)
```

carlos-lopez-tecnativa (changes requested):

sale_pricelist_global_rule/models/sale_order.py
pedrobaeza: Yes, I was about to change it.

carlos-lopez-tecnativa (approved):

PR #4166, opened 2026-02-20
[19.0][MIG] portal_sale_personal_data_only: Migration to version 19.0

cristina-hidalgo-tecnativa:
- Changed `groups_id` → `group_ids` in user creation (Odoo 19 standard field rename)
- Added `.sudo()` in sale order creation within tests as portal users don't have create permissions for sale orders
- Updated security rules to grant access both to the direct customer (`partner_id`) and to followers (`message_partner_ids`), preserving the subscriber-based access behavior across versions
- Updated tests accordingly: `test_access_sale_order_followers` now verifies that followers do have access to the sale order

@Tecnativa TT58680
@christian-ramos-tecnativa @pedrobaeza


portal_sale_personal_data_only/tests/test_partner_access.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertTrue(self.order_b in orders_b)
+        self.assertTrue(self.order_c not in orders_b)
+        orders_c = self.env["sale.order"].with_user(self.user_c).search([])
+        self.assertTrue(self.order_a not in orders_c)
+        self.assertTrue(self.order_b not in orders_c)
+        self.assertTrue(self.order_c in orders_c)
+
+    def test_access_sale_order_followers(self):
+        if self.partner_a not in self.order_a.message_partner_ids:
+            self.order_a.message_subscribe(partner_ids=[self.partner_a.id])
```
christian-ramos-tecnativa: I think this shouldn't be needed, can you try it?

christian-ramos-tecnativa (changes requested):

portal_sale_personal_data_only/tests/test_partner_access.py
cristina-hidalgo-tecnativa: Reviewed and all good now

christian-ramos-tecnativa (approved):

portal_sale_personal_data_only/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,8 @@
+- Harald Panten \<<harald.panten@sygel.es>\>
+- [Tecnativa](https://www.tecnativa.com):
+  - David Vidal
+  - Víctor Martínez
+  - Stefan Ungureanu
```
pedrobaeza, line 5: Put you here.

alexey-pelykh (approved):
Thanks for the migration, @cristina-hidalgo-tecnativa!

Looks good. The 19.0 adaptations are clean and correct:

- `groups_id` -> `group_ids` rename: standard Odoo 19 change
- `.sudo()` on `Form(self.env["sale.order"].sudo())`: needed since portal users don't have sale order create rights
- Explicit `message_subscribe()`: properly handles the Odoo 19 change where partners are no longer auto-subscribed as followers on SO creation

Hooks, security rules, and manifest are all in order. Tests pass on both Odoo and OCB. LGTM.

portal_sale_personal_data_only/tests/test_partner_access.py
```diff
@@ -49,11 +49,13 @@ def _create_partner(self, user):
...
 
     def _create_sale_order(self, partner):
-        sale_form = Form(self.env["sale.order"])
+        sale_form = Form(self.env["sale.order"].sudo())
         sale_form.partner_id = partner
         with sale_form.order_line.new() as line_form:
             line_form.product_id = self.product
         sale = sale_form.save()
+        if partner not in sale.message_partner_ids:
+            sale.message_subscribe(partner_ids=[partner.id])
```
chienandalu: The thing is that Odoo has changed the behavior of automatic subscriptions in the sales orders: https://github.com/moduon/odoo/commit/d6fac81d5d387083680316cb97e60af05623f2aa#diff-ee4fbc318d85f8bedb5c477972e698a4d6a2640c0deacb30b59a248d2e93e866

And so the orignal ir.rule has change consecuently:

```diff
-        <field name="domain_force">[('message_partner_ids','child_of',[user.commercial_partner_id.id])]</field>
+        <field name="domain_force">[('partner_id','child_of',[user.commercial_partner_id.id])]</field>
```

I'd say that the module rules should adapt to that new behavior

chienandalu (changes requested):

portal_sale_personal_data_only/tests/test_partner_access.py
pedrobaeza: I think this module should keep the old behavior of subscribers as a fallback when you are not the direct customer, and also for keeping the behavior across versions.

chienandalu: Yes, might make sense for that scenario. What about a mix o both so we don't have to rely on subscribe the partner?

pedrobaeza: Yeah, we can put as authorized both the subscribers, and the specific contact of the sales order. To discuss if also adding the invoicing contact.

cristina-hidalgo-tecnativa: Modified.

portal_sale_personal_data_only/readme/CONTRIBUTORS.md
cristina-hidalgo-tecnativa: Added.

portal_sale_personal_data_only/security/security.xml
```diff
@@ -22,7 +28,7 @@
         >Portal display personal invoices linked to personal sale orders</field>
         <field name="model_id" ref="account.model_account_move" />
         <field name="domain_force">
-            [('invoice_line_ids.sale_line_ids.order_id.message_partner_ids', 'child_of', [user.partner_id.id])]
+            [('invoice_line_ids.sale_line_ids.order_id.partner_id', 'child_of', [user.partner_id.id])]
```
chienandalu: Why not considering both subscriptions and partner as in the sales orders as well?

cristina-hidalgo-tecnativa: Added now.

portal_sale_personal_data_only/security/security.xml
```diff
@@ -3,17 +3,23 @@
     <record id="portal_sale_order_user_rule" model="ir.rule">
         <field name="name">Portal Only Personal Quotations/Sales</field>
         <field name="model_id" ref="sale.model_sale_order" />
-        <field
-            name="domain_force"
-        >[('message_partner_ids','child_of', [user.partner_id.id])]</field>
+        <field name="domain_force">
+            ['|',
+                ('partner_id', 'child_of', [user.partner_id.id]),
+                ('message_partner_ids', 'child_of', [user.commercial_partner_id.id])
```
chienandalu: (for all cases) This is the rule that makes sense. Otherwise child A will see child B's stuff and that's what we want to avoid
```suggestion
                ('message_partner_ids', 'child_of', [user.partner_id.id])

Comment (cristina-hidalgo-tecnativa) [reply portal_sale_personal_data_only/security/security.xml]: Modified in all cases now.

File: portal_sale_personal_data_only/tests/test_partner_access.py [hunk 6]
```diff
@@ -49,7 +49,7 @@ def _create_partner(self, user):
         )
 
     def _create_sale_order(self, partner):
-        sale_form = Form(self.env["sale.order"])
+        sale_form = Form(self.env["sale.order"].sudo())
```
chienandalu, line 52: Is the sudo needed?

chienandalu (approved):
Nice. Thanks!

Small doubt:

PR #4168, opened 2026-02-23
[18.0][IMP] sale_order_revision: Change where the mock has to be loaded

victoralmau:
Change where the mock has to be loaded

With the changes introduced in odoo/odoo@de056cc#diff-b37c7fd5520c97a29ddc59495779e7be61a66e15e85603928e27b3affb9dc31f, an error was occurring because the change validation on the model was being performed before tearDownClass was executed. Therefore, it has been modified so that the mock changes are applied in each test and cleaned up after each test. This way, the error will no longer appear and the test will run normally.

@Tecnativa


pedrobaeza (approved):
Fast-tracking, as this is affecting only tests and solution already researched in multiple repositories.

/ocabot merge patch

PR #4171, opened 2026-02-23
[REF] sale_order_line_price_history: extract extensible method

aisopuro:


alexey-pelykh (approved):
Thanks for the refactor. Extracting `_determine_domain()` is a good call — makes the domain logic overridable without duplicating the whole `_onchange_partner_id`. No behavior change, just better extensibility.

Code review LGTM.

Shide (approved):
Code review. Helpful for inheritance

PR #4173, opened 2026-02-23
[17.0][IMP] product_supplierinfo_for_customer_sale: Short label

pedrobaeza:
As odoo resizes columns according their content, including the header labels, this column is taking a lot of space without the real need, so let's short it for better sales order display.

@Tecnativa TT61142


victoralmau (approved):

christian-ramos-tecnativa (approved):

alexey-pelykh (approved):
Looks good. Shortening the label to "Cust. Code" makes sense for tree/list views where column space is limited.

LGTM.

PR #4176, opened 2026-02-25
[ADD] partner_last_sale_info: customer last sale date info

dreispt:


partner_last_sale_info/tests/test_res_partner.py
```diff
@@ -0,0 +1,97 @@
...
+        sale_order2.action_confirm()
+
+        # Check that last sale order date is updated to the newer one
+        self.customer._compute_last_sale_order_date()
+        self.assertEqual(self.customer.last_sale_order_id, sale_order2.id)
+
+        # Cancel the latest order and check it's not counted
+        sale_order2.action_cancel()
+        self.customer._compute_last_sale_order_date()
+        self.assertEqual(self.customer.last_sale_order_id, sale_order1.id)
```
alexey-pelykh, line 92: Bug: `last_sale_order_id` is a `Many2one` (returns recordset), but `sale_order1.id` is an integer. This will always fail.

Should be:
```python
self.assertEqual(self.customer.last_sale_order_id, sale_order1)
```

Same issue on lines 118, 123, 128.

partner_last_sale_info/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        compute="_compute_last_sale_order_date",
+        store=True,
+        readonly=False,  # Allows data imports
+        help="Date of the last sale order for this customer",
+    )
+    last_sale_order_id = fields.Many2one(
+        "sale.order",
+        compute="_compute_last_sale_order_date",
+        store=True,
+        readonly=False,
```
alexey-pelykh, line 20: N+1 pattern: `search()` per partner. For batch recomputation, consider a single SQL query with `DISTINCT ON (partner_id)` to fetch all partners' last orders at once.

partner_last_sale_info/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        help="Last sale order for this customer",
+    )
+
+    @api.depends("sale_order_ids")
+    def _compute_last_sale_order_date(self):
+        for partner in self:
+            last_order = self.env["sale.order"].search(
+                [("partner_id", "=", partner.id), ("state", "=", "sale")],
```
alexey-pelykh, line 28: Missing `else` branch: when no confirmed SO exists, `last_sale_order_date` and `last_sale_order_id` are never reset. Stored computed fields must assign a value in all code paths.

Add:
```python
else:
    partner.last_sale_order_date = False
    partner.last_sale_order_id = False
```

partner_last_sale_info/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+    _inherit = "res.partner"
+
+    last_sale_order_date = fields.Datetime(
+        compute="_compute_last_sale_order_date",
+        store=True,
+        readonly=False,  # Allows data imports
+        help="Date of the last sale order for this customer",
+    )
+    last_sale_order_id = fields.Many2one(
+        "sale.order",
```
alexey-pelykh, line 17: Depending on `sale_order_ids` triggers recomputation on every new quotation (draft), but the compute only considers `state = 'sale'`. This causes unnecessary recomputations.

partner_last_sale_info/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "category": "Sales",
+    "author": "Open Source Integrators,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-workflow",
+    "depends": ["sale"],
+    "data": [
+        "views/res_partner.xml",
+    ],
+    # Uncomment in large databases to skip automatically populating the data
```
alexey-pelykh, line 17: `pre_init_hook` is commented out -- dead code. Either enable it or remove it.

alexey-pelykh (changes requested):
Thanks for the contribution, @dreispt. Surfacing the last sale order date/reference on the partner form is useful and generic enough for OCA. However, the CI tests are failing and there are several issues to address.

**Test bug (CI failure).** Line 92 compares a `Many2one` recordset with an integer: `self.assertEqual(self.customer.last_sale_order_id, sale_order1.id)`. Should be `self.assertEqual(self.customer.last_sale_order_id, sale_order1)`. Same issue on lines 118, 123, 128.

**N+1 query pattern.** The compute does `search(..., limit=1)` inside a loop over partners. For batch recomputation (module install), this will be very slow. Consider a single SQL query with `DISTINCT ON (partner_id)` or `read_group`.

**Missing `else` branch.** When no confirmed SO exists, the stored computed fields are never reset, leaving stale values (e.g., when last confirmed SO is cancelled).

**`@api.depends("sale_order_ids")` fires too often.** This triggers on every new quotation (draft SO), but the compute only considers `state = 'sale'`. Consider triggering only on state change.

**`pre_init_hook` commented out.** The hook code exists but does nothing. On a large database, module install triggers stored compute for all partners.

**Tests use `setUp` instead of `setUpClass`.** For Odoo 17.0, `setUpClass` with `@classmethod` is standard.

Happy to re-review once the test assertion bug and compute performance are addressed.

dreispt:
@alexey-pelykh Thanks for reviewing. Yes, tests need attention, I should work on it.

partner_last_sale_info/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        compute="_compute_last_sale_order_date",
+        store=True,
+        readonly=False,
+        help="Last sale order for this customer",
+    )
+
+    @api.depends("sale_order_ids")
+    def _compute_last_sale_order_date(self):
+        for partner in self:
+            last_order = self.env["sale.order"].search(
```
rousseldenis, line 27: This is not efficient as will possibly generate a lot of queries following recordset length

rousseldenis (changes requested):

PR #4177, opened 2026-02-26
[18.0][IMP] sale_exception: avoid useless writes

TDu:
Although the ORM will properly handle multiple write (as mentioned in the comment)
Without the filtering the write method will still be called many time without need.


alexey-pelykh (approved):
Nice optimization. Filtering lines before the M2M write to skip no-ops avoids unnecessary ORM operations, especially on orders with many lines. LGTM.

henrybackman (approved):

simahawk (approved):

simahawk:
@TDu pls amend the commit msg to respect guidelines, tnx!

PR #4179, opened 2026-02-27
[18.0][ADD] sale_delivery_address_picking_update

JordiMForgeFlow:


AaronHForgeFlow (approved):
Code review LGTM!

alexey-pelykh (approved):
Thanks! Clean module. The approach of intercepting `write` on `sale.order` to propagate the address change to pending pickings, plus suppressing the standard activity via context flag on `activity_schedule`, is well thought out.

Good that done/cancelled pickings are excluded. Code review and tests LGTM.

sale_delivery_address_picking_update/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+from odoo import _, api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def write(self, values):
+        # The ctx update_delivery_shipping_partner already updates the
+        # address on the related deliveries in sale_stock module
+        if "partner_shipping_id" not in values or self._context.get(
```
rousseldenis:
```suggestion
        if "partner_shipping_id" not in values or self.env.context.get(
```

sale_delivery_address_picking_update/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+        if pickings:
+            pickings.partner_id = new_partner
+        return res
+
+    @api.onchange("partner_shipping_id")
+    def _onchange_partner_shipping_id(self):
+        # Replace the standard "do not forget to update" warning with an
+        # informational message that the update will happen automatically.
+        res = super()._onchange_partner_shipping_id() or {}
+        if res.get("warning"):
```
rousseldenis, line 34: @JordiMForgeFlow I will be cautious with this as this erases possible other overrides of the method

JordiMForgeFlow: @rousseldenis the most we can do is to detect if the warning message corresponds to the odoo standard warning message and in those cases update it. Nonetheless, if there are situations where the override is problematic I think the right way to handle it is via glue modules.

JordiMForgeFlow:
@OCA/crm-sales-marketing-maintainers @rousseldenis this one should be ready to merge :)

simahawk:
@JordiMForgeFlow wouldn't be safer to enable it by picking type or by partner?

JordiMForgeFlow:
@simahawk I am not sure if it would make sense to have to activate it only in some cases. For me it seems a behavior that either you always want or don't (then you simply don't use the module).

If you think it would be best to add a configuration per picking type I can do it, but I think it would be also something anyone could add if they actually find themselves in the situation. WDYT?

PR #4180, opened 2026-02-27
[18.0][FIX] sale_advance_payment: Wrong residual amount calculation on multi-currency Sale Orders after invoicing

### LINKED ISSUE #4160: [18.0][sale_advance_payment] Wrong residual amount calculation on multi-currency Sale Orders after invoicing
module: sale_advance_payment
version: 18.0

**Steps to reproduce**

1. Install the module sale_advance_payment.

2. Enable multi-currency and a currency different from the Company Currency (e.g., USD if company is EUR).

3. Create a Sale Order in the foreign currency (USD).

4. Register an advance payment using the "Pay Sale Advance" button. The amount_residual is calculated correctly at this stage.

5. Confirm the Sale Order and create a regular Invoice for the total amount.

6. Validate (Post) the Invoice.

**Current behavior**

Before issuing the invoice, the calculation of the pending amount (amount_residual) is correct, even in foreign currency, as the system properly handles the advance payment conversion.

Once the invoice is posted, the amount_residual on the Sale Order is recalculated incorrectly. The system subtracts the "signed" amount of the invoice (which is in the Company Currency) from the Sale Order total (which is in Foreign Currency). This results in a mismatching residual amount because it mixes two different currencies without conversion.

**Expected behavior**

The amount_residual should remain consistent. The paid amount from related invoices must be converted to the Sale Order currency before subtraction, or the calculation should use the document currency fields instead of the company-signed fields.

The issue is located in models/sale.py within the _compute_advance_payment method:

for inv in order.invoice_ids:
    invoice_paid_amount += (
        inv.amount_total_signed - inv.amount_residual_signed
    )

dtec-landoo:
**Description of the issue/feature:**

Fixes an issue with the calculation of amount_residual on Sale Orders when using multi-currency.

Currently, when a Sale Order is created in a foreign currency and its corresponding invoice is posted, the _compute_advance_payment method calculates the paid amount using amount_total_signed and amount_residual_signed. Since _signed fields are always in the company currency, subtracting this from the SO's amount_total (which is in the foreign currency) results in a wrong amount_residual.

This PR fixes the issue by:

* Using the document currency fields (amount_total and amount_residual) instead of the _signed fields.
* Converting the invoice paid amount to the Sale Order currency if they differ.
* Filtering related invoices to only consider posted ones.

**Steps to reproduce:**

1. Install sale_advance_payment.
2. Enable multi-currency and activate a currency different from your Company Currency (e.g., USD if the company is in EUR). Set an exchange rate.
3. Create a Sale Order in the foreign currency (e.g., USD).
4. Register an advance payment using the "Pay sale advanced" button.
5. Check the amount_residual on the SO.
6. Confirm the Sale Order, create a regular Invoice for the total amount, and Post it.

**Current behavior:**
Once the invoice is posted, the amount_residual on the Sale Order is recalculated incorrectly because it mixes company currency amounts with foreign currency amounts.

**Expected behavior:**
The amount_residual remains consistent and correct. The paid amount from related invoices is properly converted to the Sale Order currency before subtraction.

Fixes #4160

### DIFF
--- a/sale_advance_payment/models/sale.py
+++ b/sale_advance_payment/models/sale.py
@@ -80,10 +80,19 @@ class SaleOrder(models.Model):
                     advance_amount += line_amount
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
-            for inv in order.invoice_ids:
-                invoice_paid_amount += (
-                    inv.amount_total_signed - inv.amount_residual_signed
-                )
+            for inv in order.invoice_ids.filtered(lambda m: m.state == "posted"):
+                paid = inv.amount_total - inv.amount_residual
+                if inv.move_type == "out_refund":
+                    paid = -paid
+                if inv.currency_id != order.currency_id:
+                    paid = inv.currency_id._convert(
+                        paid,
+                        order.currency_id,
+                        order.company_id,
+                        inv.invoice_date or inv.date or fields.Date.today(),
+                    )
+                invoice_paid_amount += paid
+
             amount_residual = order.amount_total - advance_amount - invoice_paid_amount
             payment_state = "not_paid"
             if mls:
--- a/sale_advance_payment/tests/test_sale_advance_payment.py
+++ b/sale_advance_payment/tests/test_sale_advance_payment.py
@@ -454,3 +454,83 @@ class TestSaleAdvancePayment(BaseCommon):
             }
         )._create_payments()
         self.assertEqual(self.sale_order_1.amount_residual, 3600)
+
+    def test_06_multicurrency_residual_amount(self):
+        sale_order_usd = self.env["sale.order"].create(
+            {
+                "partner_id": self.res_partner_1.id,
+                "currency_id": self.currency_usd.id,
+                "order_line": [
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": self.product_1.id,
+                            "product_uom_qty": 1.0,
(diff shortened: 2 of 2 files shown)


dtec-landoo:
@dreispt @Nikul-Chaudhary

dreispt (commented):
Looks good, but I'm not comfortable approving without a test case.

dtec-landoo:
@dreispt Thanks for the review! Regarding the elif suggestion, I think we should keep it as an if. If we use elif, a refund (out_refund) that is ALSO in a foreign currency would skip the currency conversion block, leading to a wrong residual amount again. By keeping both as if, we ensure the amount is both negated and converted if needed.

dreispt:
Sure, the `elif` is wrong, I realized it and deleted the comment.

dtec-landoo:
@dreispt Hi! I've added the test case and the patch coverage is now at 100%. CI is fully green. Let me know if everything looks good to merge now! Thanks for the feedback.

dreispt (approved):

alexey-pelykh (approved):
Solid fix. Using amount_total - amount_residual without the _signed variants and applying sign based on move_type is cleaner and handles multi-currency correctly. The currency conversion with _convert is the right approach. Good test coverage too. LGTM.

rousseldenis (approved):

PR #4184, opened 2026-03-03
[18.0][MIG] sale_no_portal_button : Migration to 18.0

metaminux:
Migration from PR #1813 

CC @Kev-Roche @mourad-ehm


alexey-pelykh (commented):
Thanks for the migration @metaminux.

I looked through the code and everything checks out for 18.0 compatibility:

- Method signature of `_notify_get_recipients_groups` matches the 18.0 `mail.thread` / `sale.order` API
- `has_button_access` is still the correct key used in the group data dict
- State values (`draft`, `sent`, `sale`, `cancel`) are correct for 18.0
- Manifest, license, and structure are all in order

One observation (not blocking, inherited from the original module): the description in `DESCRIPTION.md` says the button is hidden "only ... for contacts that have a user associated with them", but the code unconditionally sets `has_button_access = False` for the `portal_customer` group without checking whether the partner has an associated user. Minor doc/behavior mismatch -- could be addressed separately.

No tests, but given the module's simplicity (single override, no data, no views), that's reasonable for the initial migration.

Looks good to me overall.

metaminux:
Thanks @alexey-pelykh for the review.

The DESCRIPTION.md file says the button is only DISPLAYED for users.
In fact, it removes access button for "portal_customer" group, so natively, only "users" group keep access to the button, but you're right, other modules could allow access to contacts not being users...

I'll keep it like that for the moment as a standard migration, and we could discuss if we need to change the condition or change the phrasing in DESCRIPTION.md later.

rousseldenis (approved):

mourad-ehm (approved):

MarwanBHL (approved):

metaminux:
@OCA/crm-sales-marketing-maintainers Can someone merge this approved migration PR ?

sale_no_portal_button/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
rousseldenis:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

rousseldenis (approved):

metaminux:
Hello @rousseldenis 

I applied your suggestion about https on links for AGPL and fixing another one in __manifest__.py
I also rebased the branch

Can you try to merge again ? I already encountered this "429 Too many requests" error on pipy... don't know if we could do anything but try it again later.

Kev-Roche (approved):

alexey-pelykh (approved):
Re-reviewed against 18.0 and the override still checks out: `_notify_get_recipients_groups(self, message, model_description, msg_vals=None)` matches the current 18.0 `mail.thread` signature, the 3-tuple group unpacking and the `has_button_access` key are correct, and the `state not in ("draft", "cancel")` guard is right. Manifest, license, and structure are all in order. Thanks for the migration, @metaminux. LGTM.

BhaveshHeliconia (approved):
LGTM!

PR #3223, opened 2024-07-05
[14.0][ADD] sale_order_line_date_next_reception

mathieudelva:

PR #4187, opened 2026-03-03
[18.0][MIG] sale_order_line_date_next_reception

metaminux:
Standard migration from PR #3223

CC @Kev-Roche @mourad-ehm


metaminux:
I renamed the field `date_next_reception` to `next_reception_date` to be more consistent with `product_next_reception_date`.

I didn't rename the module itself though...

sale_order_line_date_next_reception/test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo14-addon-product_next_reception_date @ git+https://github.com/OCA/product-attribute/pull/1712/head#subdirectory=setup/product_next_reception_date
```
mourad-ehm: The test-requirement must reference the product_next_reception_date module, which has already been migrated to v18 !

metaminux: Thanks @mourad-ehm 

In fact this `test-requirements.txt`was totally useless, to be effective, it should be located in the root folder and as you noticed, the `product_next_delivery_date` is already migrated to 18.0, so it is installed during the first stages of tests.

Good catch though ;)

Thanks again for the review

metaminux:
Hello @mourad-ehm and @Kev-Roche 

Can I have a review on this PR please ?

mourad-ehm:
Field renaming requires a migration script to preserve data. If you do not add migration scripts, could you document the field renaming information in the README to facilitate the creation of such a script.

metaminux:
> Field renaming requires a migration script to preserve data. If you do not add migration scripts, could you document the field renaming information in the README to facilitate the creation of such a script.

Hello @mourad-ehm 

This field is a not stored computed field so we don't need any migration script, right ?

metaminux:
Hello @mourad-ehm 

Can you approve this PR, please ?

mourad-ehm (approved):

Kev-Roche (approved):

sale_order_line_date_next_reception/models/sale_order_line.py
```diff
@@ -0,0 +1,20 @@
...
+
+    next_reception_date = fields.Date(compute="_compute_next_reception_date")
+
+    def _compute_next_reception_date(self):
+        for line in self:
+            line.next_reception_date = False
+            qty_available = line.product_id.with_context(
+                warehouse=line.order_id.warehouse_id.id
+            ).qty_available
+            if qty_available <= 0 and line.state not in ["done", "cancel"]:
```
rousseldenis: You should use `float_compare()`

rousseldenis (changes requested):

sale_order_line_date_next_reception/models/sale_order_line.py
metaminux: Done ✅

rousseldenis (approved):

PR #4012, opened 2025-11-13
[18.0][IMP] sale_commercial_partner: add options to filter the shipping address and invoice address by commercial partner of customer

AungKoKoLin1997:
This PR adds the options to filter the shipping address and invoice address by commercial partner of customer.

@qrtl QT5936

PR #4193, opened 2026-03-05
[18.0][MIG] sale_partner_shipping_invoice_domain : Migration to 18.0

metaminux:
Standard migration from 14.0

CC @Kev-Roche @mourad-ehm


mourad-ehm:
There'si no test in this module ?

metaminux:
Hello @mourad-ehm 

No, there was no test in 14.0 and it seems it isn't the only module without test in the repo...

Don't know if it's mandatory for migration, but I'll try to add some later.

sale_partner_shipping_invoice_domain/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+    def _compute_partner_shipping_domain(self):
+        for rec in self:
+            domain = [
+                ("company_id", "in", [False, rec.partner_id.company_id.id]),
+            ]
+            if rec.partner_id:
+                domain = expression.AND(
+                    [domain, self._get_shipping_partner_domain(rec)]
+                )
+            rec.partner_shipping_id_domain = json.dumps(domain)
```
rousseldenis, line 33: FYI, you can use Binary field to store the domain (see in account Odoo module for examples). Then, you don't need to `json.dumps()`

metaminux: Hello @rousseldenis 

Thanks for your advice.

Looking for examples of Binary fields for domain, I saw that `sale_commercial_partner` (which is in the dependencies of this module) is using it... for the exact same fields !!!

This is a new feature merged yesterday in #4012 but it can be activated or not in configuration settings.

As @mourad-ehm pointed earlier, this one as no test when `sale_commercial_partner` has one for both fields

So, I'd like some advice here from people involved in both modules...

@Kev-Roche @kevinkhao @AungKoKoLin1997 @yostashiro @atchuthan @alexis-via Do we just remove this module and keep sale_commercial_partner only ? Do one of you have a use case where this one should be kept as is ?

metaminux: Any update on this module migration ? I'd like to move forward...

kevinkhao: @metaminux on my side agreed this seems redundant now with sale_commercial_partner.

AungKoKoLin1997: @metaminux I think we don't need this glue module `sale_partner_shipping_invoice_domain` anymore since that feature is included in my PR and we can enable or disable by configuration.

metaminux:
Thanks @kevinkhao and @AungKoKoLin1997,

I just opened a PR on OpenUpgrade to declare the merge and close this one.

PR #4198, opened 2026-03-09
[19.0][MIG]  sale_automatic_workflow_payment_mode: Migration to 19.0

sergijPf:
as source 16.0 version module has been used for migration.

Few changes were done:
- improved if-statement check - make sure invoice has payment_mode set inside _register_payment_invoice() method. Without it, 'sale_automatic_workflow' tests will fail; 
- remove 'picking_ids' in tests as functionality has been moved to different place;
- added 'pre_compute=True' for sale_order's 'workflow_process_id' field. 
'wokrflow_process_id' is set as computed field here and originally, it is added as dependency to  '_compute_team_id()', which is pre-computed, as result it leads to issue that precomputed field cannot depend on non-precomputed, the relevant warning message is logged in console.


sale_automatic_workflow_payment_mode/readme/CREDITS.md
```diff
@@ -1,2 +0,0 @@
-The migration of this module from 13.0 to 14.0 was financially supported
```
rousseldenis: IMHO, this should not be removed (nothing to do with migration scripts)

sergijPf: revert it back now, no other changes

rousseldenis (approved):

BhaveshHeliconia (approved):
LGTM!

john-herholz-dt:
@rousseldenis could you kindly merge this PR since it is approved for some weeks now and blocking us. Thanks a lot!

PR #4210, opened 2026-03-12
[18.0][MIG] sale_order_line_cancel

divad1196:
- `action_done` ands `state: "done"` do not exist anymore. They are replaced by `action_lock` 
  We can therefore remove `action_done` from the test setups as `action_lock` must prevent any kind of change.
- The views were also broken


mmequignon (approved):
LGTM

sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+            ):
+                vals["product_uom_qty"] = line.qty_delivered
+            line.write(vals)
+
+    def cancel_remaining_qty(self):
+        lines = self.filtered(lambda line: line.can_cancel_remaining_qty)
+        lines._update_qty_canceled()
+        for line in lines:
+            line.order_id.message_post(
+                body=_(
```
simahawk:
```suggestion
                body=self.env._(
```
for all gettext calls

sale_order_line_cancel/tests/common.py
```diff
@@ -0,0 +1,64 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import Command
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderLineCancelBase(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 12:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (changes requested):
Thanks for your work. Pls squash the last commit on the migration one once you are done.

sale_order_line_cancel/tests/common.py
divad1196: Thank you, tracking is indeed not necessary.
I applied the changes and merge my changes into the migration commit.

sale_order_line_cancel/models/sale_order_line.py
divad1196: Done. Good to be aligned with Odoo's changes

twalter-c2c (approved):

sale_order_line_cancel/views/sale_order_line.xml
```diff
@@ -9,13 +9,38 @@
...
+                    />
+                    <button
+                        name="%(sale_order_line_cancel.action_sale_order_line_cancel)d"
+                        type="action"
+                        icon="fa-ban"
+                        title="Cancel remaining qty"
+                        invisible="not can_cancel_remaining_qty"
+                    />
+                </div>
+                <field name="can_cancel_remaining_qty" invisible="1" />
```
TDu, line 42: Since 18 the fields used in expression are automatically added as invisible in the view. So could be removed.

TDu (approved):
Thanks for this LG.
Small non blocking comment.

sale_order_line_cancel/tests/common.py
rousseldenis: Or you can use `odoo.addons.base.tests.common.BaseCommon`

mmequignon: > Or you can use `odoo.addons.base.tests.common.BaseCommon`

Well I think this test class is doing too many things and is slowing down tests.
@simahawk do you confirm ?

simahawk (approved):

sale_order_line_cancel/tests/common.py
rousseldenis: > > Or you can use `odoo.addons.base.tests.common.BaseCommon`
> 
> Well I think this test class is doing too many things and is slowing down tests. @simahawk do you confirm ?

IMHO not: https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/tests/common.py#L31

The only record created is a partner (you do the same here under) and some variables are added (which are convenient).

So, tests are not slowed down. Fake news :smile:

divad1196: > > Or you can use `odoo.addons.base.tests.common.BaseCommon`
> [ ... ]
> So, tests are not slowed down. Fake news 😄

@rousseldenis  That's an interesting insight.  

The goal here was to migrate with as little changes as possible from the original version. This makes review, back/forward-ports easier.

There are a lot of possible improvments. A few others were mentioned already and I can see even more of them. I just think they should be part of a different PR.

PR #4211, opened 2026-03-12
[18.0][MIG] sale_order_line_cancel_sale_stock

divad1196:
Depends on
- #4210 

Only had to adapt the mock data for the tests.


sale_order_line_cancel_sale_stock/models/sale_order_line.py
```diff
@@ -11,8 +11,8 @@ class SaleOrderLine(models.Model):
 
     def _get_moves_to_cancel(self):
         lines = self.filtered(
-            lambda l: l.qty_delivered_method == "stock_move"
-            and l.can_cancel_remaining_qty
+            lambda line: line.qty_delivered_method == "stock_move"
+            and line.can_cancel_remaining_qty
         )
         return lines.move_ids.filtered(lambda m: m.state not in ("done", "cancel"))
```
mmequignon, line 17: Can it be done with a single search ?
```suggestion
return self.env["stock.move"].search(
    [
        ("sale_line_id", "in", self.ids),
        ("sale_line_id.qty_delivered_method", "=", "stock_move"),
        ("sale_line_id.can_cancel_qty", "=", True),
        ("state", "not in", ("done", "cancel)),
    ]
)

**Review by mmequignon — COMMENTED** [review]
LG overall. Maybe a tiny perf improvement

**Comment by mmequignon** [comment]
Also, instead of having all the commits of sale_order_line_cancel in your PR, you can add a dep to another PR in the `test-requirements.txt` file.
Here's the doc https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

File: sale_order_line_cancel_sale_stock/models/sale_order_line.py (hunk 1 above)
Comment (divad1196) [reply sale_order_line_cancel_sale_stock/models/sale_order_line.py:17]: @mmequignon The version is the same as it was in 16.0, I limited the migration to the minimum changes necessary.

It can be done in a a single query, yes. But I wonder why it wasn't done this way initially.
One benefit I can see for the current version is that, because of Odoo's cache, the first version can avoid IO. It might be lighter/faster on small recordsets.

**Comment by divad1196** [comment]
> Also, instead of having all the commits of sale_order_line_cancel in your PR, you can add a dep to another PR in the `test-requirements.txt` file. Here's the doc https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

Thank you, I follow the instructions and submitted a new PR !

**Review by twalter-c2c — APPROVED** [review]
LGTM. I do not feel confident to suggest a way to proceed with the performance improvements. I agree that the migration itself should be limited to minimum. But if suggestion from @mmequignon leads to improved performance on larger recordsets, you may want to consider including that in a separate commit (or in a separate PR).

**Review by TDu — APPROVED** [review]

**Comment by simahawk** [comment]
dependency is getting merged

**Comment by divad1196** [comment]
@TDu  @twalter-c2c @simahawk The dependency is merged and this PR is approved.

Could you merge it?
Thank you in advance !

**Review by BhaveshHeliconia — APPROVED** [review]
LGTM!

File: sale_order_line_cancel_sale_stock/tests/common.py [hunk 2]
```diff
@@ -18,7 +18,12 @@ def setUpClass(cls):
     @classmethod
     def _prepare_product_vals(cls):
         vals = super()._prepare_product_vals()
-        vals["type"] = "product"
+        vals.update(
+            {
+                # "type": "consu",
```
jbaudoux: remove commented line as it's default value

sale_order_line_cancel_sale_stock/models/sale_order_line.py
divad1196: @mmequignon @twalter-c2c @simahawk Based on the full discussion,

Can we agree to resolve this comment since:
- Performance improvment should be measured
- The PR should focus on the migration and minimal changes necessary

If this improvment is wanted, we can open a dedicated PR for that.

jbaudoux (approved):

PR #4212, opened 2026-03-12
[19.0][MIG] sale_invoice_blocking: Migration to 19.0

yankinmax:


sale_invoice_blocking/tests/test_sale_invoice_blocking.py
```diff
@@ -0,0 +1,101 @@
...
+from odoo import Command
+from odoo.exceptions import UserError
+from odoo.tests.common import TransactionCase, mute_logger
+
+
+class TestSaleInvoiceBlocking(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis: @yankinmax FYI, you can also use `odoo.addons.base.tests.common.BaseCommon` to avoid this line.

rousseldenis (approved):
LGTM

sale_invoice_blocking/tests/test_sale_invoice_blocking.py
yankinmax: Hello @rousseldenis 
I completely agree with you about `BaseCommon` class.
But, I would prefer not to load what I don't need to simulate the tests.
That one line reduces some calls and loggings, just as simple as it can be.
It's much easier to create some demo data and see it directly in test without relying on what we can use from BaseCommon.

rousseldenis: > But, I would prefer not to load what I don't need to simulate the tests.

Just a partner ?

ivantodorovich (approved):

sale_invoice_blocking/tests/test_sale_invoice_blocking.py
yankinmax: Yep, not too much :smile:
Honestly, this way seems clearer to maintain and see all we load for the concrete test.
Isn't it blocking?

rousseldenis: Nope. I just want to be sure there is no rumor with its usage and the real impact :smiley:

yankinmax:
@rousseldenis can we pls go for a merge?

BhaveshHeliconia (approved):
Functional review LGTM!

rousseldenis (approved):
LGTM

PR #718, opened 2018-10-16
[10.0][FIX] issue #717

tafaRU:
Fixes #717

Issue #717, opened 2018-10-16
[10.0] base branch fails sale_mrp_link tests


SimoRubi:
First occurrence of the fail: https://travis-ci.org/OCA/sale-workflow/jobs/436183007#L1789

tafaRU:
@SimoRubi, thanks for reporting the issue.
The fail is due to the installation of `sale_procurement_group_by_line`as explained in https://github.com/OCA/sale-workflow/pull/661#discussion_r214031110

It is necessary to also exclude `sale_delivery_split_date`. I'll give a try in the next days.

rafaelbn:
Solved in https://github.com/OCA/sale-workflow/pull/718

tafaRU:
@rafaelbn, wouldn't be better to close the issue once the linked PR will be merged?

PR #4213, opened 2026-03-13
[MIG] sale_procurement_group_by_commitment_date: Migration to 18.0

tgaona-sbg:


tgaona-sbg:
Hey @tafaRU @SimoRubi @rafaelbn, I think this issue is linked to something you have already seen a few years ago : https://github.com/OCA/sale-workflow/issues/717#issuecomment-430231334

-> On my local instance tests are going well (as the `sale_delivery_split_date` is in fact uninstalled).

-> Should we apply separate tests as it was done in this related fix : https://github.com/OCA/sale-workflow/pull/718/changes

Here what says the README.rst : 

Known issues / Roadmap
======================

This module is not compatible with sale_delivery_split_date. A glue
module will be needed to install sale_delivery_split_date and
sale_procurement_group_by_commitment_date. This is an already known
issue:
https://github.com/OCA/sale-workflow/issues/717#issuecomment-430231334


Thanks in advance for your help !

rousseldenis:
@tgaona-sbg Could you check tests ?

Maybe a glue module, yes, but IMHO, this behaviour change should be configurable (activable when needed).

tgaona-sbg:
Hello @rousseldenis, thanks for the answer !

So tests on my local instance (without `sale_delivery_split_date` installed) are actually going fine : 
image

However, as the README says, this module is not compatible with `sale_delivery_split_date` -> which is actually inside the same repo OCA/sale-workflow, and installed and tested during the checks phase.

The previous found solution by OCA members was to set the tests running distinctly for each incompatible module with an exclusion of their reciprocal incompatible counterpart.

image

Considering the glue module, I didn't found any reference yet to it (not sure it actually exists as it seems it was only a suggestion at the time).

Any clue on this will be welcome !

Thank you again

PR #3580, opened 2025-02-11
[14.0] [FIX] Change partner on line when partner it's changed on order

pablo-lp:

PR #4215, opened 2026-03-16
[16.0][FIX] sale_order_line_price_history: fix order_partner_id sync on partner change

felipemotter:
## Summary

The `order_partner_id` stored related field on `sale.order.line` stops syncing when the order's partner is changed. This causes stale partner data on order lines, breaking any logic that depends on `order_partner_id` (e.g., price history lookups for the wrong customer).

## Steps to reproduce

1. Create a sale order with Partner A and add a product line
2. Save the order
3. Change the partner to Partner B
4. Save the order
5. Check `order_partner_id` on the line — it still shows Partner A

## Root cause

The module redefines `order_partner_id` with `depends=["product_id"]` to ensure the value is available on NewId records in the view. However, this overrides the implicit depends from the `related` attribute (`order_id.partner_id`), so the ORM no longer triggers a recompute when the partner changes on the parent order.

## Solution

Explicitly declare both `related` and `depends` so the field keeps its related behavior while also triggering on `product_id` changes:

```python
order_partner_id = fields.Many2one(
    related="order_id.partner_id",
    depends=["product_id", "order_id.partner_id"],
)
```

A regression test is included that fails without the fix and passes with it.

### DIFF
--- a/sale_order_line_price_history/models/sale_order_line.py
+++ b/sale_order_line_price_history/models/sale_order_line.py
@@ -8,4 +8,7 @@ class SaleOrderLine(models.Model):
 
     # In core this a related field. We need to trigger its value on view, so we can
     # have it even when we're in a NewId
-    order_partner_id = fields.Many2one(depends=["product_id"])
+    order_partner_id = fields.Many2one(
+        related="order_id.partner_id",
+        depends=["product_id", "order_id.partner_id"],
+    )
--- a/sale_order_line_price_history/tests/test_sale_order_line_price_history.py
+++ b/sale_order_line_price_history/tests/test_sale_order_line_price_history.py
@@ -1,7 +1,7 @@
 # Copyright 2018 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import Form, TransactionCase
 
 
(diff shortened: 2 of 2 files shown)


felipemotter:
This fixes the same issue reported and fixed in earlier versions:
- v13.0: #2367
- v14.0: #3580

antoniospneto (approved):

sale_order_line_price_history/models/sale_order_line.py
```diff
@@ -8,4 +8,7 @@ class SaleOrderLine(models.Model):
 
     # In core this a related field. We need to trigger its value on view, so we can
     # have it even when we're in a NewId
-    order_partner_id = fields.Many2one(depends=["product_id"])
+    order_partner_id = fields.Many2one(
```
rousseldenis, line 11: Seems legit. I'm just wondering why this is in this module (as it can be useful elsewhere)

felipemotter: Thanks for the review\! The field `order_partner_id` is a `related` field in core Odoo and works fine there. This module overrides it (removing the `related`) to handle NewId records in the price history wizard, which is what introduces the bug. So I believe the fix belongs here, but I'm open to suggestions if you think there's a better approach.

Shide (approved):

BhaveshHeliconia (approved):

PR #4218, opened 2026-03-17
[18.0][MIG] sale_price_compliance

Shide:
Straightforward migration

List view. Custom Icons and Texts
image

Kanban view
image

Form view
image

Sale Order message if confirmed with non compliant prices
image

MT-13560 @moduon @chienandalu @fcvalgar @Gelojr @rafaelbn please review if you want 😄


sale_price_compliance/tests/test_sale_price_compliance_constraints.py
```diff
@@ -12,7 +12,7 @@ class TestPriceComplianceConstraints(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        # Create a clean company for testing
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis: FYI, you can use `odoo.addons.base.tests.common.BaseCommon` to avoid this line

PR #4221, opened 2026-03-17
[16.0][MIG] sale_triple_discount: migration script for 16.0

dreispt:
This seems to be missing ... or am I missing something?


rousseldenis:
@dreispt Yes :smile: 

https://github.com/OCA/sale-workflow/pull/3907


This is a long history one and should be merged to align with further versions.

Could you review it ? thanks

lmignon (changes requested):
This code doesn't apply to the 16.0 branch. The new implementation is provided by  #3907

rousseldenis:
@dreispt Could we close this, using #3907 ?

PR #4223, opened 2026-03-19
[18.0][IMP] sale_invoice_plan: case 1 installment and advance

ROBBYHOOD9:
**Detail**

This PR is create for allow sale invoice plans can create with 1 installment and advance.
image


rousseldenis:
@ROBBYHOOD9 Could you add a test case to avoid regressions ? Thanks

ROBBYHOOD9:
@rousseldenis Sorry for the late update. I have already updated the test case for the 1-installment with advance scenario.

Saran440 (approved):
Code Review 👍

PR #4228, opened 2026-03-24
[18.0][IMP] sale_elaboration: auto notes parameter

chienandalu:
For some workflow cases it's annoying to have the elaboration notes auto filled with elaboration title.

This commit adds a parameter to disable that behavior so everyone is happy with their own preference.

cc @moduon MT-14286

please review @Shide @yajo


rousseldenis (approved):

sale_elaboration/readme/CONFIGURE.md
```diff
@@ -14,6 +14,11 @@ To configure this module you need to:
 You can define elaboration profiles to limit the elaborations that can be
 selected for each product.
 
+If you want to prevent the elaboration note from being filled automatically
+from the selected elaborations, set the system parameter
+`sale_elaboration.disallow_auto_notes` to a truthy value in
```
Shide: If parameter must be true, use `from odoo.tools.misc import str2bool`, otherwise change this to say "any" value

Shide (changes requested):

sale_elaboration/readme/CONFIGURE.md
chienandalu: It's not really necessary as I implemented it (I just check the existance of the key). I'd rather keep it simple. In the other hand although not necesary, it's more explicit if the user sets the value to something meaningful.

Shide (approved):

sale_elaboration/models/product_elaboration_mixin.py
```diff
@@ -24,6 +24,12 @@ def onchange_elaboration_ids(self):
         """Use onchange method instead of compute because if any other data of any line
         force a recompute of all lines and the elaboration custom notes are reset.
         """
+        if (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("sale_elaboration.disallow_auto_notes")
+        ):
```
yajo: suggestion: use `odoo.tools.misc.str2bool` here and avoid double negatives:


```suggestion
        if not str2bool(
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("sale_elaboration.auto_notes", True)
        ):
```

Benefit:
- Before, setting the parameter to "false", "0", or "no" would actually enable it. Now it disables it.
- Before, disabling the parameter enabled the feature. Now enabling enables, and disabling disables.

sale_elaboration/readme/CONFIGURE.md
yajo:
```suggestion
`sale_elaboration.disallow_auto_notes` to `false` in
```

yajo (changes requested):
A small suggestion to improve predictability and usability.

Shide (approved):

yajo (approved):
/ocabot merge minor

PR #4229, opened 2026-03-24
[18.0][FIX] sale_invoice_frequency: move back to sales tab

chienandalu:
In the migration to v18 the invoice frequency field was moved to the invoicing tab in the partner form. This seemd pretty convient as it's an invoicing setting, but it's breaking the users workflow that need to set different frequency for different addresses.

cc @moduon MT-14294

please review @Shide @Gelojr @EmilioPascual

### DIFF
--- a/sale_invoice_frequency/views/res_partner_view.xml
+++ b/sale_invoice_frequency/views/res_partner_view.xml
@@ -7,11 +7,12 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="account.view_partner_property_form" />
         <field name="arch" type="xml">
-            <xpath
-                expr="//group[@id='invoice_send_settings']/field[@name='invoice_sending_method']"
-                position="before"
-            >
-                <field name="invoice_frequency_id" options="{'no_create': True}" />
+            <xpath expr="//group[@name='sale']" position="inside">
+                <field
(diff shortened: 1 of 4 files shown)


sale_invoice_frequency/views/res_partner_view.xml
```diff
@@ -7,11 +7,12 @@
...
-            <xpath
-                expr="//group[@id='invoice_send_settings']/field[@name='invoice_sending_method']"
-                position="before"
-            >
-                <field name="invoice_frequency_id" options="{'no_create': True}" />
+            <xpath expr="//group[@name='sale']" position="inside">
+                <field
+                    name="invoice_frequency_id"
+                    options="{'no_create': True}"
+                    groups="account.group_account_invoice"
```
rousseldenis, line 14: The question could be: isn't this too limitative ?

rousseldenis: No, ok, it was there in invoicing group

chienandalu: Yes, I'm just reverting it to how it was in the previous version

rousseldenis (approved):

Shide (approved):

PR #4235, opened 2026-03-26
18.0 fix procurement double run

sprintitviki:
When confirming a sales order containing consumable products, the overridden `_action_launch_stock_rule` method processed the procurements but then passed the entire recordset to `super()`. This caused the core Odoo logic to run the procurements a second time, resulting in duplicate stock moves (picking lines) for those products.

This commit resolves the issue by tracking the sale order lines processed by the custom logic and filtering them out of the recordset. `super()` is now only called with the remaining, unprocessed lines (such as storable products) to ensure each line is only procured once.

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -31,15 +31,16 @@ class SaleOrderLine(models.Model):
         """
         if self._context.get("skip_procurement"):
             return True
-            
-        precision = self.env["decimal.precision"].precision_get("Product Unit of Measure")
+
+        precision = self.env["decimal.precision"].precision_get(
+            "Product Unit of Measure"
+        )
         procurements = []
         groups = {}
         if not previous_product_uom_qty:
             previous_product_uom_qty = {}
-            
-        # 1. Create an empty recordset to track which lines we process here
-        processed_lines = self.env['sale.order.line'] 
+
+        processed_lines = self.env["sale.order.line"]
 
         for line in self:
             line = line.with_company(line.company_id)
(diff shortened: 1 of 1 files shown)


rousseldenis:
@sprintitviki Thanks for this. In order to get reviews, please pay attention to your PR title.

It should contain at least the module name like `[18.0][FIX] sale_procurement_group_by_line: (...)`

sale_procurement_group_by_line/model/sale.py
```diff
@@ -31,13 +31,16 @@ def _action_launch_stock_rule(self, previous_product_uom_qty=False):
...
-        )
+            
+        precision = self.env["decimal.precision"].precision_get("Product Unit of Measure")
         procurements = []
         groups = {}
         if not previous_product_uom_qty:
             previous_product_uom_qty = {}
+            
+        # 1. Create an empty recordset to track which lines we process here
+        processed_lines = self.env['sale.order.line']
```
rousseldenis, line 42: `processed_lines = self.env['sale.order.line'].browse()` is better

sale_procurement_group_by_line/model/sale.py
```diff
@@ -46,17 +49,13 @@ def _action_launch_stock_rule(self, previous_product_uom_qty=False):
...
-            if (
-                float_compare(qty, line.product_uom_qty, precision_digits=precision)
-                == 0
-            ):
+            if float_compare(qty, line.product_uom_qty, precision_digits=precision) == 0:
                 continue
 
             group_id = line._get_procurement_group()
 
-            # Group the sales order lines with same procurement group
```
rousseldenis, line 58: If you remove this comment, you should change it with a better one if this does not fit

rousseldenis:
@sprintitviki Could you add a test case that highlight the problem and avoid regression ? Thanks

rousseldenis:
This seems correct, could you fix pre-commit too and squash your commits if needed?

rousseldenis:
@sprintitviki

PR #4233, opened 2026-03-26
[18.0][IMP] sale_order_line_cancel post-migration improvments

divad1196:
A column is currently displayed with no value due to the `invisible="1"` used.  
It should have been `column_invisible="1"` since V17

Also removed state `done` from the possible states of a sale.order.line since the state does not exist anymore

PR #4237, opened 2026-03-27
[18.0][IMP] Making columns optional&hidden by default

divad1196:
Made the fields `Qty Canceled` and `Qty To Deliver` hidden by default.    
Not all users need to see these fields.

### DIFF
--- a/sale_order_line_cancel/README.rst
+++ b/sale_order_line_cancel/README.rst
@@ -96,6 +96,15 @@ OCA, or the Odoo Community Association, is a nonprofit organization whose
 mission is to support the collaborative development of Odoo features and
 promote its widespread use.
 
+
+.. |maintainer-jbaudoux| image:: https://github.com/jbaudoux.png?size=40px
+    :target: https://github.com/jbaudoux
+    :alt: jbaudoux
+
+Current `maintainer <https://odoo-community.org/page/maintainer-role>`__:
+
+|maintainer-jbaudoux| 
+
 This module is part of the `OCA/sale-workflow <https://github.com/OCA/sale-workflow/tree/18.0/sale_order_line_cancel>`_ project on GitHub.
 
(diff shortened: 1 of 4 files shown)


rousseldenis:
@divad1196 Thanks for this. You should respect commit message format like `[IMP] sale_order_line_cancel: (...)`

christian-ramos-tecnativa:
IMO it’s better to set it as `optional="show"`, since this is the main purpose of the module. Otherwise, it might seem that nothing changes after installing it.

divad1196:
@rousseldenis changed, thank you for pointing it out.

@christian-ramos-tecnativa The proposal comes from @jbaudoux
It will depend on the users, but in my experience:
- Users don't like unexpected changes.
- Not everyone will need them and not everyday

I am personnaly more in favor of having it hidden by default, but if @jbaudoux, @rousseldenis agree I have nothing against changing it.

jbaudoux:
> IMO it’s better to set it as `optional="show"`, since this is the main purpose of the module. Otherwise, it might seem that nothing changes after installing it.

The main purpose is provide an action on the SO line to cancel what is remaining. That icon is visible. The other colums are not usefull for a salesman placing an order.

As you are on it, could you add me as maintainer of the module as I'm the initial author and maintaining it.

jbaudoux (approved):

christian-ramos-tecnativa:
Could you remove the column `Can Cancel Remaining Qty`?
image

jbaudoux:
> Could you remove the column `Can Cancel Remaining Qty`?

He fixed it here https://github.com/OCA/sale-workflow/pull/4233/changes/0b95548e46f1b32747c71697a991fae1e149f972
* https://github.com/OCA/sale-workflow/pull/4233

christian-ramos-tecnativa (approved):

pedrobaeza:
Please resolve the conflict

divad1196:
> Please resolve the conflict

@pedrobaeza  Solved

PR #4239, opened 2026-03-31
[18.0][ADD] sale_order_finish_service

etobella:
Odoo way of handling sale orders has only 4 states (draft, sent, sale,
cancel).

Also, on all services sale orders (linked to timesheets), it only checks
if the sale order is in state sale. In projects, it works fine, but it
can lead to errors on users, because they are not allowed to "close" an
existing sale order.

The concept of this module is to add a new flag that will detected
"finished sale orders" and will not allow to select them anymore.

Usage
=====

Access a sale order. When it is in state sale order, a Finish/Unfinish
button will appear in top.


luisDIXMIT (commented):
When a sale is marked as finished,

https://github.com/user-attachments/assets/cef3a0ec-aefe-4266-aac0-e61d457a9972

 it is not searchable on timesheets but by default is setted. I show you in a video.

etobella:
@luisDIXMIT I see the problem, however, that might happen also without this module if we cancel a Sale Order. it will still be used :thinking: 

If we want to fix this, we probably need to change Odoo core to fix it :thinking:

luisDIXMIT (approved):
Since my comments are not related to the module, I tested it on Runboat and the behavior is correct. LGTM!

ValentinVinagre (approved):
Lgtm 👍

PR #4240, opened 2026-03-31
[19.0][FIX] sale_invoice_frequency: move back to sales tab

chienandalu:
fw of:

- https://github.com/OCA/sale-workflow/pull/4229

In the migration to v18 the invoice frequency field was moved to the invoicing tab in the partner form. This seemd pretty convient as it's an invoicing setting, but it's breaking the users workflow that need to set different frequency for different addresses.

cc @Moduon MT-14294

please review @yajo  @Gelojr


rousseldenis (approved):

EmilioPascual (approved):
LGTM

Shide (approved):

PR #4241, opened 2026-03-31
[19.0][FIX] sale_global_discount: discount amounts on orders without …

adrip-s73:
When modifying quantities on a confirmed order that has no global discounts, the fields amount_untaxed_before_global_discounts and amount_total_before_global_discounts were not being updated.

The _compute_amounts method used an early continue when global_discount_ids was empty, leaving those compute fields unassigned and retaining stale data in the cache/database.

Explicitly assign the standard untaxed and total amounts to the "before discount" fields, and set the global discount amount to 0.0 before continuing the loop.

image


rousseldenis (approved):
Seems legit

ioans73 (approved):
LGTM

PR #4242, opened 2026-03-31
[18.0][ADD] sale_order_line_price_unit_digits: new module

chienandalu:
From this change https://github.com/odoo/odoo/pull/243987 , `price_unit` in sale order lines has a minimum display digits instead of a rounding in screen as it used to be. This is for compatibility issues with Peppol.

For the cases this isn't needed, the new behavior is quite anoying, as prices computed from the pricelists often throw quite a bunch of decimals (e.g.: 10.000023), which is ok for accuracy sake but is terrible for ux.

cc @Moduon MT-14324

please review @Shide @EmilioPascual


sale_order_line_price_unit_digits/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+{
+    "name": "Sale order line price unit digits",
+    "summary": "Legacy digits visualization for price unit",
+    "version": "18.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Sales/Sales",
+    "website": "https://github.com/OCA/sale-workflow",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["chienandalu", "rafaelbn"],
+    "license": "LGPL-3",
```
cgarcia-solvos, line 12: Code review: LGTM.

Minor comment: license should be "AGPL-3" instead of "LGPL-3" to follow OCA standards
```suggestion
    "license": "AGPL-3",
```

cgarcia-solvos (changes requested):

sale_order_line_price_unit_digits/__manifest__.py
chienandalu: There's no issue with it (ref: https://odoo-community.org/resources/faq)

image

cgarcia-solvos: You are right, thanks for the clarification. LGTM!

cgarcia-solvos (approved):

pedrobaeza:
For specific things like pricelists, you can set the type as formula, and put in `Round off to` 0.01, but it's a general problem, and lots of us have complained about it, but no reaction from Odoo.

chienandalu:
Yeah, I know, and Odoo is proposing that workaround as well (https://github.com/odoo/odoo/pull/243987#issuecomment-3834348907)... the thing is, with thousands of price list items, that workaround is a pain to maintain for users and constantly annoys users....

EmilioPascual (approved):
LGTM. Code review

chienandalu:
@rousseldenis there's an errata in the merge command

mariancuadranetkia:
@chienandalu can you rebase, please? I like review your module. 
About failed merge, it's due a failed test that checks decimal precission, isn't it? Which should be the best way to resolve this situation? How could we help to resolve it?

PR #4243, opened 2026-03-31
[FIX] sale_exception: add a register_hook to prevent other action_con…

lef-adhoc:
…firm being executed if there is an exception rule

review: #3767


rousseldenis:
@lef-adhoc Could you check https://github.com/OCA/sale-workflow/pull/4255 ?

As this seems better way of managing that.

PR #4245, opened 2026-04-01
[18.0][ADD] portal_sale_list_extra_info

cgarcia-solvos:
Adds additional fields 'client_order_ref' and 'amount_untaxed' to the portal view.
It also makes it easier to add other fields in the future.

portal_sale_order_list_fields_image


dalonsod:
@cgarcia-solvos this actually covers `my/orders`, but still not `my/quotes`. Could you add it?

cgarcia-solvos:
@dalonsod Done!

dalonsod:
amount untaxed column should be better placed here for quotations IMO:

imagen

cgarcia-solvos:
@dalonsod Done!

dalonsod (approved):
LGTM 👍

cgarcia-solvos:
@yostashiro  @AungKoKoLin1997 could you review? Thanks!!

portal_sale_order_list_fields/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,53 @@
+<odoo>
+    <template
+        id="custom_portal_my_orders"
+        inherit_id="sale.portal_my_orders"
+        name="My Sale Orders (in portal_sale_order_list_fields)"
+    >
+        <xpath expr="//thead/tr[hasclass('active')]/th[1]" position="after">
```
yostashiro, line 7: I think it'd be safer to anchor on classes or another available attribute instead of using the index. The same applies to other xpath expr occurrences.

portal_sale_order_list_fields/static/description/icon.png
yostashiro, line 1: You can remove this icon so that a new one will be generated upon merge.

portal_sale_order_list_fields/tests/test_portal_my_orders.py
```diff
@@ -0,0 +1,95 @@
+# © 2026 Solvos Consultoría Informática (<https://www.solvos.es>)
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo.tests.common import HttpCase, tagged
+
+
+@tagged("post_install", "-at_install", "pruebas")
```
yostashiro:
```suggestion
@tagged("post_install", "-at_install")
```

portal_sale_order_list_fields/tests/test_portal_my_orders.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.portal_user = cls.env["res.users"].create(
+            {
+                "name": "Portal User Test",
+                "login": "portal_user_test",
+                "password": "portal_password",
+                "groups_id": [(6, 0, [cls.env.ref("base.group_portal").id])],
```
yostashiro: Non-blocking.
```suggestion
                "groups_id": [Command.set([cls.env.ref("base.group_portal").id])],
```

yostashiro (commented):
Couold the module name just be `portal_sale_order_list`?

portal_sale_order_list_fields/views/sale_portal_templates.xml
cgarcia-solvos: I agree the index is fragile. I tried anchoring to `th[text()='Total']` but the OCA pre-commit blocks it. The original `<th>` tags lack name or id attributes. They are literally `<th class="text-end">Total</th>` tags. And anchoring to CSS classes like `hasclass('text-end')` is just as fragile as an index. There is genuinely no better alternative.

portal_sale_order_list/tests/test_portal_my_orders.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+        cls.sale_order = cls.env["sale.order"].create(
+            {
+                "partner_id": cls.portal_partner.id,
+                "client_order_ref": "REF-CLIENT-999",
+                "order_line": [
+                    (
+                        0,
+                        0,
```
yostashiro:
```suggestion
                    Command.create(
```

portal_sale_order_list/tests/test_portal_my_orders.py
```diff
@@ -0,0 +1,96 @@
...
+        cls.sale_order.action_confirm()
+
+        cls.quotation = cls.env["sale.order"].create(
+            {
+                "partner_id": cls.portal_partner.id,
+                "client_order_ref": "REF-QUOTE-888",
+                "order_line": [
+                    (
+                        0,
+                        0,
```
yostashiro:
```suggestion
                    Command.create(
```

yostashiro (approved):
LGTM. 👍

AungKoKoLin1997 (approved):
Code and functional review: LG 👍 

Image
Image

cgarcia-solvos:
@pedrobaeza Is this ready to be merged? Thanks!

pedrobaeza (commented):
The name of the module is not the best one IMO, as the main module is `sale`, not `portal`, and the feature is not clear also, so I suggest `sale_portal_list_extra_info`.

dalonsod:
> The name of the module is not the best one IMO, as the main module is `sale`, not `portal`, and the feature is not clear also, so I suggest `sale_portal_list_extra_info`.

Agree with addon name suffix, but we've selected `portal_sale_` as preffix because there are similar addons like `portal_sale_order_search` that have such nomenclature. What do you think @pedrobaeza ?

pedrobaeza:
I see, my own fault in that moment that I didn't think on this when developing the other modules...

OK, let's go with `portal_sale_list_extra_info`.

portal_sale_list_extra_info/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# © 2026 Solvos Consultoría Informática (<https://www.solvos.es>)
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Portal Sale List Extra Info",
+    "summary": """
+        Adds additional fields 'client_order_ref' and 'amount_untaxed'
```
pedrobaeza: You shouldn't use triple quota, as this also saves the extra spaces, line feeds, etc.

pedrobaeza (approved):
/ocabot merge nobump

PR #4247, opened 2026-04-01
[FIX]Sale_order_secondary_unit: erase the default secondary unit of sale when the uom is deactivated

mav-adhoc:
If i have a default secondary unit for sales and i deactivate this UoM, it doesn't erase the UoM from the default secondary unit for sale, not only allowing the user to still use that UoM in a sale order, but also selecting this secondary UoM as default in the sale order.
image
image

### DIFF
--- a/sale_order_secondary_unit/models/product.py
+++ b/sale_order_secondary_unit/models/product.py
@@ -3,6 +3,18 @@
 from odoo import fields, models
 
 
+class ProductSecondaryUnit(models.Model):
+    _inherit = "product.secondary.unit"
+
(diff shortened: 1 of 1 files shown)


florenciafrigieri2 (approved):
Functional review ok. When performing a sale, then the secondary unit is empty and if you want to select one, then the actives are the only availables.

Evidence: https://drive.google.com/file/d/1O6LGz_yXAE3d_TXCWlbBKEWgRsExsnDo/view

jcadhoc (approved):

PR #4255, opened 2026-04-06
[18.0][IMP] sale_exception: Use hook to popup exception after rollback

grindtildeath:
With OCA/server-tools#3590 changing the way exceptions are detected, the error raising doesn't allow to call `_popup_exception` as we used to, but we can use the new hook to have the popup displayed smoothly.

Depends on:
 - [ ] https://github.com/OCA/server-tools/pull/3590

### DIFF
--- a/sale_exception/models/sale_order.py
+++ b/sale_exception/models/sale_order.py
@@ -47,11 +47,11 @@ class SaleOrder(models.Model):
         if orders:
             orders._check_exception()
 
+    def _must_popup_exception(self):
+        return self.env.company.sale_exception_show_popup
+
     def action_confirm(self):
-        if self.detect_exceptions():
-            if not self.env.company.sale_exception_show_popup:
-                return
-            return self._popup_exceptions()
+        self.detect_exceptions()
         return super().action_confirm()
 
     def action_draft(self):
--- a/sale_exception/models/sale_order_line.py
+++ b/sale_exception/models/sale_order_line.py
@@ -4,7 +4,9 @@
 import html
 
 from odoo import api, fields, models
+from odoo.api import Environment
 from odoo.fields import Command
+from odoo.tools import config
 
 
 class SaleOrderLine(models.Model):
@@ -53,15 +55,31 @@ class SaleOrderLine(models.Model):
 
     def _detect_exceptions(self, rule):
         records = super()._detect_exceptions(rule)
-        # Thanks to the new flush of odoo 13.0, queries will be optimized
-        # together at the end even if we update the exception_ids many times.
-        # On previous versions, this could be unoptimized.
-        lines_to_remove_exception = (self - records).filtered(
-            lambda line: rule.id in line.exception_ids.ids
+        test_mode = config["test_enable"] and not self.env.context.get(
+            "test_base_exception"
         )
-        lines_to_remove_exception.exception_ids = [Command.unlink(rule.id)]
-        lines_to_add_exception = records.filtered(
-            lambda line: rule.id not in line.exception_ids.ids
-        )
-        lines_to_add_exception.exception_ids = [Command.link(rule.id)]
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with self.env.registry.cursor() as new_cr:
+            new_env = (
+                Environment(new_cr, self.env.uid, self.env.context)
+                if not test_mode
+                else self.env
+            )
+            lines_to_remove_exception = (self - records).filtered(
+                lambda line: rule.id in line.exception_ids.ids
+            )
+            lines_to_remove_exception.with_env(new_env).exception_ids = [
+                Command.unlink(rule.id)
+            ]
+            lines_to_add_exception = records.filtered(
+                lambda line: rule.id not in line.exception_ids.ids
+            )
+            lines_to_add_exception.with_env(new_env).exception_ids = [
+                Command.link(rule.id)
+            ]
         return records.mapped("order_id")
+
+    def _detect_exception_get_exc_class_values(self):
+        res = super()._detect_exception_get_exc_class_values()
+        return dict(res, target_model="sale.order")
--- a/sale_exception/tests/common.py
+++ b/sale_exception/tests/common.py
@@ -1,0 +1,26 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+try:
+    from decorator import decoratorx as decorator
+except ImportError:
+    from decorator import decorator
+
+from contextlib import contextmanager
+from unittest.mock import patch
+
+
+@contextmanager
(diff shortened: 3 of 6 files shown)


wouitmil (approved):

SilvioC2C (approved):

vvrossem (approved):

sale_exception/models/sale_order.py
```diff
@@ -47,11 +47,12 @@ def sale_check_exception(self):
...
 
+    def _must_popup_exception(self):
+        return self.env.company.sale_exception_show_popup
+
     def action_confirm(self):
         if self.detect_exceptions():
-            if not self.env.company.sale_exception_show_popup:
-                return
-            return self._popup_exceptions()
+            return
```
divad1196: I had the same question for the purchase modules, but shouldn't we call `action_popup_exceptions` there ?

sale_exception/models/sale_order_line.py
```diff
@@ -53,15 +55,31 @@ def _reverse_field(self):
 
     def _detect_exceptions(self, rule):
         records = super()._detect_exceptions(rule)
-        # Thanks to the new flush of odoo 13.0, queries will be optimized
-        # together at the end even if we update the exception_ids many times.
-        # On previous versions, this could be unoptimized.
-        lines_to_remove_exception = (self - records).filtered(
-            lambda line: rule.id in line.exception_ids.ids
+        test_mode = config["test_enable"] and not self.env.context.get(
```
divad1196, line 58: Should that be an utility function in https://github.com/OCA/server-tools/pull/3590 ?
If we change what "test_mode" means, we currently would need to find and edit all modules with this snippet

sale_exception/models/sale_order_line.py
```diff
@@ -53,15 +55,31 @@ def _reverse_field(self):
...
-        lines_to_remove_exception.exception_ids = [Command.unlink(rule.id)]
-        lines_to_add_exception = records.filtered(
-            lambda line: rule.id not in line.exception_ids.ids
-        )
-        lines_to_add_exception.exception_ids = [Command.link(rule.id)]
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with self.env.registry.cursor() as new_cr:
+            new_env = (
+                Environment(new_cr, self.env.uid, self.env.context)
```
divad1196, line 65: Same idea for the utility here. In fact, we only really need an utility to return the correct env since `test_mode` is only used for that.

```python
with base_exception_test_env(self.env) as new_env:
```
?

sale_exception/models/sale_order_line.py
```diff
@@ -53,15 +55,31 @@ def _reverse_field(self):
...
+                if not test_mode
+                else self.env
+            )
+            lines_to_remove_exception = (self - records).filtered(
+                lambda line: rule.id in line.exception_ids.ids
```
divad1196, line 70: `rule in line.exception_ids` should work without `.id/.ids`

sale_exception/models/sale_order_line.py
```diff
@@ -53,15 +55,31 @@ def _reverse_field(self):
...
+            )
+            lines_to_remove_exception.with_env(new_env).exception_ids = [
```
divad1196, line 72: Same code as https://github.com/OCA/server-tools/pull/3590/changes#diff-21c3e74ef9aeefd2e85dac72b05522fcb846b00d0602a87b311aa595cbcea29aR104

Should that also not be some kind of utility? The pattern is the same even if the code is differen (immediat `write` vs assignation)

sale_exception/tests/common.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+try:
+    from decorator import decoratorx as decorator
```
divad1196, line 4: I think decoratorx is legacy and [shouldn't be used](https://github.com/micheles/decorator/blob/master/docs/documentation.md):
> Rather than using decoratorx, you should fix your introspection routines to use inspect.Signature without fiddling with the __code__ object.

That's the only mention I could find.
In general, I think we can just use `functools.wraps` instead of `decorator`, even if that implies an explicit closure ( `Explicit is better than implicit.` 😄 )

sale_exception_product_sale_manufactured_for/tests/test_sale_exception.py
```diff
@@ -2,13 +2,17 @@
...
+    swallow_base_exception_error,
+)
 
 
 class TestSaleException(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-
+        cls.env.context = dict(cls.env.context, test_base_exception=True)
```
divad1196, line 15: Can we always use the same approach here?
cf `cls.env = cls.env(context=...)`

divad1196 (approved):
Changes work I think, just ensure for the `action_confirm` return statement.
The rest are just recommendations

sale_exception/tests/common.py
grindtildeath: I used this since it's used in https://github.com/odoo/odoo/blob/18.0/odoo/tests/common.py#L75-L85

I guess it's fine to use it in that case 🤷

sale_exception/models/sale_order.py
grindtildeath: As written in https://github.com/OCA/purchase-workflow/pull/3012#discussion_r3324593992 we can remove the `if` altogether

sale_exception/models/sale_order_line.py
grindtildeath: That's right but it was actually the legacy code that was indented

sale_exception/models/sale_order_line.py
grindtildeath: It could... but again, this is the legacy code that is adapted to the new exception mechanism, I don't want to rewrite everything and prefer to keep the diff that only adds the use of new environment

sale_exception/models/sale_order_line.py
grindtildeath: Technically you're right. But I don't think it brings that much added value compared to the actual code

sale_exception/models/sale_order_line.py
grindtildeath: Answered below

sale_exception/models/sale_order_line.py
divad1196: The added value is the maintainability coming from the encapsulation.

If tomorrow something needs to be changed, you will need to track all places where this code as been used and create multiple PRs.
Also, cursor are a sensitive matter, the more we have it, the more likely we are to have issues.
Having it in a single place reduce the risk of issues and makes it easier to troubleshoot

BhaveshHeliconia (approved):

simahawk:
Depending PR merged. Rebased.

PR #4251, opened 2026-04-03
[18.0][FIX] sale_partner_primeship tests with freeze time

sergiocorato:
#4192

Issue #4192, opened 2026-03-04
[18.0][BUG] sale_partner_primeship : tests are failing from 11pm to midnight


metaminux:
Not sure yet about this, but I encountered some tests failures in `sale_partner_primeship` when updating an unrelated PR for `sale_invoice_delivery_state`

## Module
sale_partner_primeship

## Describe the bug
This is the logs I had when running tests for #4183 (test job : https://github.com/OCA/sale-workflow/actions/runs/22646567189/job/65635911957?pr=4183)

```
2026-03-03 23:01:26,944 438 INFO odoo odoo.modules.loading: Loading module sale_partner_primeship (174/267) 
2026-03-03 23:01:27,395 438 INFO odoo odoo.modules.registry: module sale_partner_primeship: creating or updating database tables 
2026-03-03 23:01:27,415 438 INFO odoo odoo.models: Prepare computation of res.partner.active_primeship 
2026-03-03 23:01:27,791 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/views/product_template_views.xml 
2026-03-03 23:01:27,823 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/views/sale_primeship_views.xml 
2026-03-03 23:01:27,839 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/views/res_partner_views.xml 
2026-03-03 23:01:27,866 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/security/ir.model.access.csv 
2026-03-03 23:01:27,877 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/security/sale_partner_primeship.xml 
2026-03-03 23:01:27,882 438 INFO odoo odoo.modules.loading: loading sale_partner_primeship/data/ir_cron.xml 
2026-03-03 23:01:28,013 438 INFO odoo odoo.addons.sale_partner_primeship.tests.test_res_partner_primeship: Starting TestResPartnerPrimeship.test_check_expired_primeships ... 
2026-03-03 23:01:28,030 438 INFO odoo odoo.addons.sale_partner_primeship.tests.test_res_partner_primeship: ====================================================================== 
2026-03-03 23:01:28,030 438 ERROR odoo odoo.addons.sale_partner_primeship.tests.test_res_partner_primeship: FAIL: TestResPartnerPrimeship.test_check_expired_primeships
Traceback (most recent call last):
[...]

PR #4257, opened 2026-04-08
[FIX] sale_partner_primeship : freeze time in tests to avoid errors d…

metaminux:
…ue to DST

Following #4251 (thanks to @sergiocorato)

Freeze time in all other tests as the issue #4192 is still there for 2 tests
See https://github.com/OCA/sale-workflow/actions/runs/24109913027/job/70341911143


nayatec:
of course, thks @metaminux !

PR #4260, opened 2026-04-08
[19.0][ADD] sale_order_line_section

santostelmo:
This module adds a stored Section field on sale order lines.

In standard Odoo, the relation between a line and its section is only implicit:
a line belongs to the nearest previous section according to the order sequence.
This is inconvenient for domains, grouping, reporting, or custom business logic
because there is no direct field to use.

With this module, each non-section sale order line is linked to its nearest
previous section line.

Typical use cases include:

- grouping sale order lines by section in tree views, exports, or reports;
- filtering all lines that belong to a given section without recomputing the
  relation from the sequence;


sale_order_line_section/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+        comodel_name="sale.order.line",
+        string="Section",
+        compute="_compute_section_id",
+        compute_sudo=True,
+        precompute=True,
+        store=True,
+        index=True,
+    )
+
+    @api.depends("order_id.order_line.sequence", "order_id.order_line.display_type")
```
imlopes:
```suggestion
    @api.depends("order_id.order_line.sequence", "order_id.order_line.display_type", "sequence")
```

imlopes (approved):

sale_order_line_section/models/sale_order_line.py
santostelmo: Thank you @imlopes updated dependency

cyrilmanuel (approved):

cyrilmanuel:
hi @pedrobaeza could you merge this pr please :)

Ricardoalso (commented):
You probably need a migration script to compute `section_id` for all existing records

pedrobaeza:
More than a migration script, it should be a pre-init hook for avoiding a big delay installing the module in a populated DB with lots of `sale.order.line`.

santostelmo:
@Ricardoalso @pedrobaeza pre-init hook added to create column and populate stored field values via sql

sale_order_line_section/hooks.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def _column_exists(cr, table_name, column_name):
```
Ricardoalso: You can use `from odoo.tools.sql import column_exists`

santostelmo: you're right. I updated the code to use `odoo.tools.sql`

Ricardoalso (approved):

sale_order_line_section/models/sale_order_line.py
```diff
@@ -0,0 +1,42 @@
...
+    _inherit = "sale.order.line"
+
+    section_id = fields.Many2one(
+        comodel_name="sale.order.line",
+        string="Section",
+        compute="_compute_section_id",
+        compute_sudo=True,
+        precompute=True,
+        store=True,
+        index=True,
```
simahawk, line 20: is the index really needed? I don't see any search for it.

sale_order_line_section/tests/test_sale_order_line_section.py
```diff
@@ -0,0 +1,101 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrderLineSection(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create({"name": "Section Partner"})
```
simahawk, line 12: disable tracking pls

sale_order_line_section/tests/test_sale_order_line_section.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_section_id_recomputed_when_section_sequence_changes(self):
+        order = self._create_order()
+        s1 = order.order_line.filtered(lambda line: line.name == "S1")
+        s2 = order.order_line.filtered(lambda line: line.name == "S2")
+        l1 = order.order_line.filtered(lambda line: line.name == "L1")
+        l2 = order.order_line.filtered(lambda line: line.name == "L2")
+        l3 = order.order_line.filtered(lambda line: line.name == "L3")
+
+        s2.write({"sequence": 25})
+        order.order_line._compute_section_id()
```
simahawk: if you call compute directly it means the dependency is not correct

simahawk (changes requested):
I'm not sure to understand what is the purpose of this module 🤔 
Is it to have a different report? To view lines grouped in a search view?
The description might benefit from real use case examples.

sale_order_line_section/tests/test_sale_order_line_section.py
santostelmo: no need to call `_compute_section_id`anymore

sale_order_line_section/models/sale_order_line.py
santostelmo: is intentional. The field is stored specifically to support direct domains and section-based analysis without reconstructing the relation from sequence. Since `section_id` is expected to be used in searches and filters, indexing it is justified.

santostelmo:
> I'm not sure to understand what is the purpose of this module 🤔 Is it to have a different report? To view lines grouped in a search view? The description might benefit from real use case examples.

I improved the description to better explain the use case of this module

sale_order_line_section/models/sale_order_line.py
simahawk: you can copy paste this msg as a comment on the field ;)

sale_order_line_section/models/sale_order_line.py
```diff
@@ -0,0 +1,47 @@
...
+                if order_line == line:
+                    break
+                if order_line.display_type == "line_section":
+                    current_section = order_line
+            line.section_id = current_section
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        lines.mapped("order_id.order_line")._compute_section_id()
```
simahawk: is this needed because of `precompute`? 🤔

santostelmo: yes it was needed because of `precompute`. But dropped the precompute

santostelmo:
@simahawk can you please have a look. Still any blocking point ?

sale_order_line_section/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,15 @@
+This module adds a stored Section field on sale order lines.
```
simahawk, line 1: :heart:

simahawk (approved):

PR #4261, opened 2026-04-08
[19.0][ADD] sale_section_stock_move

santostelmo:
This module propagates sale order section information to stock
operations.

It adds a Sale Section on stock moves from the related sale order line
section, and exposes the section name on stock moves and stock move
lines.

The section information is available in stock move and stock move line
list views, search filters, group by options, and in the Operations tab
of stock pickings.

Depends of https://github.com/OCA/sale-workflow/pull/4260


cyrilmanuel (approved):

PaoloYam (approved):

sale_section_stock_move/models/stock_move.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    sale_section_name = fields.Text(
+        string="Section Name",
+        related="sale_section_id.name",
+        store=True,
+        readonly=True,
+    )
+
+
+class StockMoveLine(models.Model):
```
rousseldenis: Even if tolerated, better to have one file per model.

santostelmo: Thanks @rousseldenis  I updated the code

sale_section_stock_move/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    sale_section_id = fields.Many2one(
```
simahawk, line 10: is this relevant for ALL moves and move lines? 🤔 
A change on so line section might have a big perf impact. no?

santostelmo: is this relevant for ALL moves and move lines? 

The fields are added on stock.move and stock.move.line, so the columns exist for all records, but they’ll only be filled (non-null) for moves/move lines that come from a sale order line (sale_line_id) and where that sale line has a section (section_id). For other moves (inventory, manufacturing, purchase, etc.) it stays empty.

A change on so line section might have a big perf impact. no?

It can, yes.
Because sale_section_id / sale_section_name are stored related fields, if you change the section on a sale order line, Odoo may need to recompute + write the stored value on all dependent records:
- all stock.move linked to that sale line (sale_line_id)
- and then all stock.move.line linked to those moves
So if someone bulk-updates sections on confirmed orders with lots of moves/move lines, that can trigger a large cascade of writes (and be noticeable on big DBs). 
If section changes are rare, the impact is usually limited.

Another option:
Make sale_section_* non-stored computed/related fields.
- Pros: changing the section won’t trigger mass writes on moves/move lines.
- Cons: you generally lose efficient search/group-by on those fields (or it becomes slower because it computes on the fly).

BhaveshHeliconia (approved):
LGTM!

sale_section_stock_move/models/stock_move.py
rousseldenis: @santostelmo Maybe something in the README about performances. But, seems ok for me.

rousseldenis (approved):

divad1196 (approved):

divad1196:
Hello, this looks great !

This has been open for more than a month, could it be merged ?

divad1196:
@santostelmo hello, what is the status of the PR ?

sale_section_stock_move/models/stock_move.py
simahawk: Hmm I still tend to say that this can have a big undesired impact. It's stored, indexed and it's also propagated to move lines.
What is the main purpose of this field? Search and group? Reports?

simahawk (changes requested):

santostelmo:
@simahawk The main use cases for these fields are:
- Search and filtering — Users need to filter stock moves/move lines by sale section in the Operations tab and list views
- Grouping — Group operations by section for better organization and visibility
- Reporting — Include section information in stock move reports

You're correct that storing indexed fields can have implications, especially when sections are changed on confirmed orders. However:
- In practice, section changes on confirmed orders are relatively rare in most implementations — sections are typically set during order creation/drafting
- The current design enables efficient queries — For the search/filter/group-by use cases above, we need stored fields for performance (on-the-fly computation would be slower)
- The cascade writes are minimal in typical usage — Only moves linked to that specific order line are affected, not ALL moves

PR #4263, opened 2026-04-08
[18.0][ADD] sale_stock_put_to_order: put-to-order zone configuration and target resolution

StephaneMangin:
Configure locations as put-to-order zones to direct incoming goods to bins that already hold products from the same order.

Features of this module:
 - Recursive is_pto boolean on stock.location (parent → children)
 - Sale-order-aware product scope for bin selection
 - Optional auto-select mode via _apply_putaway_strategy
 - Storage-category validation on candidate bins
 - Main extension entrypoints: _is_pto_location_valid, _get_pto_source_products, _prepare_pto_bin_group_vals

Supervised coding using Copilot AI with Claude Opus 4.6. Here are the different phases:
 - customer need analysis
 - reusable modules analysis (including integration efforts regarding initial budget)
 - module creation and human testing
 - Tests generation
 - Optimisations (module architecture and extensibility, speed tests, etc)

Easily integrable with the enterprise module stock_barcode to help pickers select the proper bin as quickly as possible (during a pre-pack action for instance)


sale_stock_put_to_order/models/stock_location.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    @api.depends("location_id.is_pto")
+    def _compute_is_pto(self):
+        for location in self:
+            if location.location_id:
+                location.is_pto = location.location_id.is_pto
+
+    def _inverse_is_pto(self):
+        return
```
yvaucher, line 36: To remove ?

sale_stock_put_to_order/models/stock_location.py
```diff
@@ -0,0 +1,56 @@
...
+        inverse="_inverse_is_pto",
+        store=True,
+        recursive=True,
+        readonly=False,
+        help=(
+            "Technical field indicating whether this location"
+            " belongs to a put-to-order area."
+        ),
+    )
+    parent_is_pto = fields.Boolean(
```
yvaucher, line 21: Shouldn't this be a related field on `location_id.is_pto` ?

sale_stock_put_to_order/models/stock_quant.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def _search_pto(self, locations, products):
+        """Return quants with positive stock for *products* in *locations*."""
+        return self.search(
+            [
+                ("location_id", "in", locations.ids),
+                ("product_id", "in", products.ids),
+                ("quantity", ">", 0),
+            ]
+        )
```
petrus-v, line 18: Just a kindly reminder, I've not read the whole PR !

as far I remember, at least in previous versions, quant are not meant to be unique to avoid some concurrency issues. So if you really needs products with available quantities you should consider group by.

 Also, I'm wonder if you consider physical stock or available stock ?

sale_stock_put_to_order/models/stock_location.py
StephaneMangin: A related="location_id.is_pto" reads from the parent and writes back to the parent if not readonly (which would be unusable for this usecase).
Ans so it's the opposite of what we want — when you tick is_pto on a root location, all children should inherit from it automatically.

sale_stock_put_to_order/models/stock_quant.py
StephaneMangin: This is the physical quantity indeed. I think this is sufficient for this behavior because we just want to know where certain products are to select the correct location for the put-to-order action.

```
    quantity = fields.Float(
        'Quantity',
        help='Quantity of products in this quant, in the default unit of measure of the product',
        readonly=True, digits='Product Unit of Measure')

    reserved_quantity = fields.Float(
        'Reserved Quantity',
        default=0.0,
        help='Quantity of reserved products in this quant, in the default unit of measure of the product',
        readonly=True, required=True, digits='Product Unit of Measure')

    available_quantity = fields.Float(
        'Available Quantity',
        help="On hand quantity which hasn't been reserved on a transfer, in the default unit of measure of the product",
        compute='_compute_available_quantity', digits='Product Unit of Measure')
```

Hope i understood well you concern

sale_stock_put_to_order/models/stock_location.py
StephaneMangin: The key is in `fields.py` at line 448:

```python
attrs['readonly'] = attrs.get('readonly', not attrs.get('inverse'))
```
When a field has compute, Odoo defaults readonly to not inverse:
 - No inverse → readonly = True → writes are silently discarded in write() and create()
 - With inverse → readonly = False → writes go through, then the inverse method is called
 
That's why the empty inverse (return) exists on is_pto: without it, readonly would default to True and Odoo would silently ignore any manual write on the field — making it impossible to set the root location as PTO.

yvaucher (approved):

StephaneMangin:
I've made some changes after Usecases review. It seems it missed a relation to the procurement group if the sale order line relation is not set on moves. Added in the Sale link retrieval.

Thus, the location proposition was a bit trival in its approach, too trivial. What's has changed : filter quants for products position + move analysis to get the reel locations related to the move -> sale order record.

StephaneMangin:
Overseeded by https://github.com/OCA/sale-workflow/pull/4266

PR #4266, opened 2026-04-09
[18.0][ADD] sale_stock_put_to_order: put-to-order zone configuration and target resolution

StephaneMangin:
Configure locations as put-to-order zones to direct incoming goods to bins that already hold products from the same order.

Features of this module:
 - Recursive is_pto boolean on stock.location (parent → children)
 - Sale-order-aware product scope for bin selection
 - Optional auto-select mode via _apply_putaway_strategy
 - Storage-category validation on candidate bins
 - Main extension entrypoints: _is_pto_location_valid, _get_pto_source_products, _prepare_pto_bin_group_vals

Supervised coding using Copilot AI with Claude Opus 4.6. Here are the different phases:
 - customer need analysis
 - reusable modules analysis (including integration efforts regarding initial budget)
 - module creation and human testing
 - Tests generation
 - Optimisations (module architecture and extensibility, speed tests, etc)

Easily integrable with the enterprise module stock_barcode to help pickers select the proper bin as quickly as possible (during a pre-pack action for instance)


sale_stock_put_to_order/models/stock_move_line.py
```diff
@@ -0,0 +1,46 @@
...
+            .sudo()
+            .get_param("sale_stock_put_to_order.auto_select_location")
+            == "True"
+        )
+        if not auto_select:
+            return super()._apply_putaway_strategy()
+
+        pto_handled = self.browse()
+        resolved_cache = {}
+        for sml in self:
```
rousseldenis, line 22: @StephaneMangin Shouldn't we check if picking.location_dest_id is not a pto before doing anything else ?

rousseldenis: Ok in `_get_pto_root_location()`

sale_stock_put_to_order/models/stock_picking.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_pto_root_location(self):
+        """Return the configured put-to-order root location."""
+        self.ensure_one()
+        dest = self.picking_type_id.default_location_dest_id
+        return dest if dest and dest.is_pto else self.env["stock.location"]
```
rousseldenis, line 16:
```suggestion
        return dest if dest and dest.is_pto else self.env["stock.location"].browse()
```

is better

sale_stock_put_to_order/models/stock_picking.py
```diff
@@ -0,0 +1,115 @@
...
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_pto_root_location(self):
+        """Return the configured put-to-order root location."""
+        self.ensure_one()
+        dest = self.picking_type_id.default_location_dest_id
```
rousseldenis, line 15: IMHO, this is not correct. You should use the picking `location_dest_id` instead as this is the real one used to compute the putaways in core (as this is the same on stock moves).

sale_stock_put_to_order/models/stock_move_line.py
```diff
@@ -0,0 +1,46 @@
...
+            picking = sml.picking_id
+            if not picking:
+                continue
+            if picking.id not in resolved_cache:
+                root = picking._get_pto_root_location()
+                if root:
+                    dest = next(
+                        (
+                            loc
+                            for loc, _ in (picking._find_pto_dest_location_and_quants())
```
jbaudoux, line 32: From DESCRIPTION, I understand you are on an incoming or replenishment here.
This methods collects the products with `_get_pto_source_products` which looks at `sale_orders = self.move_ids.sale_line_id.order_id`.
How can a incoming transfer be linked to a sale ? I'm lost...

jbaudoux (commented):
As it's about computing putaway, IMO this module should be in https://github.com/OCA/stock-logistics-putaway repo

sale_stock_put_to_order/models/stock_picking.py
```diff
@@ -0,0 +1,115 @@
...
+    def _get_pto_source_products(self):
+        """Return products relevant to PTO resolution.
+
+        When a sale order is linked (via sale_line_id or procurement group),
+        returns the full SO product list (including lines not yet delivered)
+        so that bin selection considers the complete order context.
+        Falls back to move products otherwise.
+        """
+        self.ensure_one()
+        sale_orders = self.move_ids.mapped("sale_line_id.order_id")
```
rousseldenis, line 27: @StephaneMangin As I see, this is the only dependency on sale modules. 

IMHO, this has to be moved to https://github.com/OCA/stock-logistics-putaway and extract this in another module.

PR #4269, opened 2026-04-12
[16.0][FIX] sale_automatic_workflow_payment_mode: fix payment method propagation

### LINKED ISSUE #3286: [16.0] sale_automatic_workflow_payment_mode: propagating wrong payment method to payments.
## Module
sale_automatic_workflow_payment_mode

## Describe the bug
Wrong payment method is propagated to automatic workflow created payment despite the payment method set on the Payment Modes

## To Reproduce
**Affected versions**: 16.0

1. Create a new payment method named electronic and add it in the last line of the Bank Journal as in the below images.
image

2. Create a new Payment Mode with the payment method we created in the step 1. Example below.
image

3. Created an Sales Order with that Payment Mode and choose an Automatic Workflow to process further. Example below.
image

4. You can see that the final payment created using the workflow with wrong payment method.
image

**Expected behavior**
The payment method should be from the Payment Mode that we setup on the sales order.

**Additional context**
The payment method (payment.method) on the payments is now a new object (payment.method.line).
This breaks the propagation, as a result it is fetching the first method on the inbound payment of a journal
We need to change our app accordingly.
Proposed a fix https://github.com/OCA/bank-payment/pull/1196

muslim-foda:
This fix updates the automatic workflow payment creation to use the payment method line matching the payment mode journal and payment method, instead of propagating the old payment method field directly.
This ensures the correct payment method is used on generated payments.

Closes #3286

### DIFF
--- a/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
+++ b/sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
@@ -15,18 +15,26 @@ class AutomaticWorkflowJob(models.Model):
         vals = super()._prepare_dict_account_payment(invoice)
         if invoice.payment_mode_id:
             payment_mode = invoice.payment_mode_id
+            journal = payment_mode.fixed_journal_id
+            method_line = journal.inbound_payment_method_line_ids.filtered(
+                lambda l: l.payment_method_id == payment_mode.payment_method_id
+            )[:1]
             vals["payment_type"] = payment_mode.payment_type
-            vals["payment_method_id"] = payment_mode.payment_method_id.id
-            vals["journal_id"] = payment_mode.fixed_journal_id.id
-        return vals
+            vals["journal_id"] = journal.id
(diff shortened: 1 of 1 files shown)


rousseldenis:
@muslim-foda Thanks for this. Maybe a test for that change as this is quite critical flow.

sale_automatic_workflow_payment_mode/models/automatic_workflow_job.py
```diff
@@ -15,18 +15,26 @@ def _prepare_dict_account_payment(self, invoice):
...
-        return super()._register_payment_invoice(invoice)
+            vals["journal_id"] = journal.id
+            vals.pop("payment_method_id", None)
+        if method_line:
+            vals["payment_method_line_id"] = method_line.id
+
+            return vals
+
+
+def _register_payment_invoice(self, invoice):
```
rousseldenis, line 31: The method seems to not belong the class (indent problem)

PR #4271, opened 2026-04-13
[19.0] [FIX] sale_delivery_state: Update sale_demo.xml

rven:


rven:
@pedrobaeza Shouldn't there also be a test pipeline to check if the demo data still works in the modules for v19?
Now I often notice that these are not correctly migrated, because the demo data is not installed by default anymore.

pedrobaeza:
Yeah, indeed this is something missing now in 19.0, as an extra CI step, but no actions have been done yet. If you are able to do it, please ping me. cc @OCA/board

pedrobaeza (approved):

sbidoul:
> Shouldn't there also be a test pipeline to check if the demo data still works in the modules for v19

I created https://github.com/OCA/oca-addons-repo-template/issues/343 to track this.

BhaveshHeliconia (approved):
LGTM!

PR #4273, opened 2026-04-14
[18.0][FIX] sale_order_type: preserve manual payment term when changing inv…

lef-adhoc:
…oice journal

When a sale type has no payment_term_id and the user sets a manual term on the sale order, changing the invoice journal triggered a recompute chain (journal_id → company_id → sale_type_id) that caused super() to silently reset invoice_payment_term_id to the partner's default.

Skip the super() call for records where neither the partner nor the sale type changed and the sale type carries no term, so the chosen term is kept. Adds a regression test covering this scenario.

### DIFF
--- a/sale_order_type/models/account_move.py
+++ b/sale_order_type/models/account_move.py
@@ -46,9 +46,15 @@ class AccountMove(models.Model):
 
     @api.depends("sale_type_id")
     def _compute_invoice_payment_term_id(self):
+        previous = {move: move.invoice_payment_term_id for move in self}
         res = super()._compute_invoice_payment_term_id()
-        for move in self.filtered("sale_type_id.payment_term_id"):
-            move.invoice_payment_term_id = move.sale_type_id.payment_term_id
+        for move in self:
+            if move.sale_type_id.payment_term_id:
+                move.invoice_payment_term_id = move.sale_type_id.payment_term_id
+            else:
+                previous_term = previous.get(move)
+                if previous_term:
+                    move.invoice_payment_term_id = previous_term
         return res
 
     @api.depends("sale_type_id")
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -191,6 +191,33 @@ class TestSaleOrderType(BaseCommon):
         self.assertEqual(invoice.invoice_payment_term_id, sale_type.payment_term_id)
         self.assertEqual(invoice.journal_id, sale_type.journal_id)
 
+    def test_invoice_change_journal_keeps_manual_payment_term(self):
+        partner = self.env["res.partner"].create(
+            {
+                "name": "Invoice Term Test Partner",
+                "property_payment_term_id": self.immediate_payment.id,
+            }
+        )
+        manual_payment_term = self.env.ref("account.account_payment_term_30days")
+        sale_type_without_payment_term = self.sale_type.copy(
+            {"name": "Sale Type Without Payment Term", "payment_term_id": False}
+        )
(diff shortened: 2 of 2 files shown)


flotho (approved):
code Review
LGTM

lef-adhoc:
Hi @pedrobaeza, what do you think about this?

pedrobaeza (changes requested):
Reading the previous code, I didn't see the problem you depicted, so I have gone to a runboat to make sure, and confirm that there's no such problem:

Image

pedrobaeza:
EDIT: Seeing that it's on moves, not orders. Trying again.

pedrobaeza:
Still no luck...

Peek 16-04-2026 08-22

lef-adhoc:
@pedrobaeza Here's a video showing it running on Runboat
https://drive.google.com/file/d/14XV3Vg_VUSRmaDPEtHt-MHmaSL4x6xvQ/view?usp=sharing

pedrobaeza:
But that's not because of `sale_order_type`, but the standard that restores the SO payment term.

lef-adhoc:
@pedrobaeza If you delete the _compute_invoice_payment_term_id method from the sale_order_type, you’ll see that the payment term isn’t updated when you update the journal. You can test this in Odoo’s Runbot.

sale_order_type/models/account_move.py
```diff
@@ -46,9 +46,15 @@ def _compute_sale_type_id(self):
...
         res = super()._compute_invoice_payment_term_id()
-        for move in self.filtered("sale_type_id.payment_term_id"):
-            move.invoice_payment_term_id = move.sale_type_id.payment_term_id
+        for move in self:
+            if move.sale_type_id.payment_term_id:
+                move.invoice_payment_term_id = move.sale_type_id.payment_term_id
+            else:
+                previous_term = previous.get(move)
+                if previous_term:
+                    move.invoice_payment_term_id = previous_term
```
pedrobaeza, line 57: For me, the code should be:

```suggestion
        for move in self:
            if not move.invoice_payment_term_id and move.sale_type_id.payment_term_id:
                move.invoice_payment_term_id = move.sale_type_id.payment_term_id
```

lef-adhoc: I think this wouldn't solve the issue that motivated the PR.

In this case, after [super()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [invoice_payment_term_id](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is already filled with the partner payment term, so your condition would not apply and the manual value would still be lost.

It would also change the current behavior of the module, because sale_order_type would stop overriding the computed payment term and would only act as a fallback when the field is empty.

nicolascol (approved):
LGTM
https://drive.google.com/file/d/1wDO7Syi-YXBKWLedPjgACvHcyWrq94Up/view
Could you watch the video?
@pedrobaeza

rousseldenis:
> hanging the invoice journal triggered a recompute chain (journal_id → company_id → sale_type_id)

@lef-adhoc @pedrobaeza 

IMHO, the problem is in sale type recomputation when changing the company (from the journal). If a sale type is set on the sale order and user changes the journal, in sale type computation, we just need to check if the previous filled in sale type belong to the company. If not, we recompute it.

But IMHO too, there are problems in compute method as there is a `depends_context` that is unwanted on stored fields. That can lead to unwanted values.

@lef-adhoc  I can check as I use this

PR #4277, opened 2026-04-16
[19.0][MIG] sale_elaboration: Migration to 19.0

BhaveshHeliconia:

PR #4278, opened 2026-04-16
[19.0][MIG] product_customerinfo_elaboration: Migration to 19.0

BhaveshHeliconia:
depends on :- 
- https://github.com/OCA/sale-workflow/pull/4111
- https://github.com/OCA/sale-workflow/pull/4277


BhaveshHeliconia:
@CarlosRoca13 @sergio-teruel Could you please review this? Thanks!

sergio-teruel (approved):

CarlosRoca13 (approved):
Diff review 👍

product_customerinfo_elaboration/models/sale_order_line.py
```diff
@@ -0,0 +1,50 @@
...
+        )[:1]
+        if not customerinfo:
+            customerinfo = self.product_id.customer_ids.filtered(
+                lambda pc: pc.elaboration_ids
+                and pc.partner_id == self.order_id.partner_id.commercial_partner_id
+            )[:1]
+        return customerinfo
+
+    @api.onchange("elaboration_ids")
+    def onchange_elaboration_ids(self):
```
rousseldenis, line 38: IMHO, onchange methods should be private.

BhaveshHeliconia: Thanks for the review!
We actually have to keep this public because it overrides `onchange_elaboration_ids` from `product.elaboration.mixin` (in `sale_elaboration`), which is public. 

If we rename this to a private method (`_onchange_elaboration_ids`), Odoo will treat them as two separate methods and run both of them. When they run together, they conflict and overwrite the elaboration note, which breaks the module's behavior.

rousseldenis (approved):

PR #4281, opened 2026-04-16
[18.0][IMP] sale_order_lot_selection: Filter pending moves only

victoralmau:
Filter pending moves only

Example use case:
- 2 steps delivery
- Move a > Picking step 1 (lot a) > done
- Move b > Picking step 2 (lot a) > backorder
- Move b > Backorder > Set lot_id=lot b to sale order line

Exclude pending move return in certain cases.

Example use case:
- 2 steps delivery
- Move a > Picking step 1 (lot a) > done
- Move b > Picking step 2 (lot a) > pending
- Move c > Return move a (lot a) > pending

If we change the lot in the sales order line to lot b, that lot should only be changed in move b

@Tecnativa TT61067


victoralmau:
Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

PR #4291, opened 2026-04-17
Sale Cancel Reason v19 app

Gravitai-IPanchal:


rousseldenis:
@Gravitai-IPanchal Thanks for this.

But, for a migration, you should follow those instructions: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

Thanks

PR #4294, opened 2026-04-20
[19.0][MIG] sale_partner_address_restrict: Migration to 19.0

BhaveshHeliconia:


rousseldenis (approved):

tonygalmiche (approved):
Tested on a local Odoo 19.0 instance:

- Sale Partner Address Restriction: enabling the company setting correctly restricts the delivery/invoice address selectors to children of the selected partner
- Confirmed the domain via UI and the underlying save constraint via shell (forcing an address from a different partner correctly raises a ValidationError for both partner_invoice_id and partner_shipping_id)

Bug found: "Dropship Address" (is_dropship_address on res.partner) does not update the displayed name. `_get_complete_name()` is overridden correctly (returns the right value when called directly), but the stored `complete_name` field's `@api.depends` isn't extended with `is_dropship_address`, so toggling the checkbox has no visible effect until another field that IS in the dependency list changes and triggers a recompute. Suggest adding `@api.depends("is_dropship_address")` (or extending the existing depends list) on the `complete_name` compute in this module.

sebalix (approved):

sebalix:
Merging despite the small bug spotted by @tonygalmiche . Maybe you could open a PR once it's merged?

/ocabot merge nobump

PR #4302, opened 2026-04-22
[16.0][FIX] sale_automatic_workflow: skip invoice creation when refund exists

muslim-foda:
### Issue
When a posted invoice is reversed by a credit note, the automatic workflow
incorrectly creates a new invoice.

### Cause
The workflow does not check if a refund (credit note) already exists for
a posted invoice before creating a new one.

### Solution
Skip invoice creation when:
- A posted invoice already exists, or
- A posted refund linked to that invoice exists.

### Additional changes
- Updated tests to reflect the new behavior.

### Result
Prevents duplicate invoices after refunds and ensures consistent workflow behavior.


rvalyi (dismissed):
I confirm the issue and the fix. It is a pretty critical issue breaking the Odoo refund feature!

muslim-foda:
@rvalyi Im waiting for your feedback

PR #4303, opened 2026-04-22
[16.0][FIX] sale_stock_picking_blocking: preserve delivery_block_id on create/write

ntsirintanis:
This PR fixes an issue where the delivery_block_id set manually on a new Sales Order could be lost when saving.

In detail:
The field delivery_block_id is implemented as a stored computed field depending on partner_id and payment_term_id. During record creation, the compute method was triggered and could overwrite the value provided manually by the user, often resetting it to False or to a default.
A similar issue could occur when changing the partner or payment term on an existing Sales Order, causing an already set value to be lost.

Solution:
The field remains computed, but the logic is adjusted so that:

*existing values are never overwritten by the compute method
*defaults are only applied when the field is empty
*create() and write() ensure consistent behavior

This ensures that manually set values are always preserved, while default values are still applied as needed.

previous attempt: https://github.com/OCA/sale-workflow/pull/3906

### DIFF
--- a/sale_stock_picking_blocking/__manifest__.py
+++ b/sale_stock_picking_blocking/__manifest__.py
@@ -1,12 +1,13 @@
 # Copyright 2019 ForgeFlow S.L.
 #   (http://www.forgeflow.com)
+# Copyright 2026 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Sale Stock Picking Blocking",
     "summary": "Allow you to block the creation of deliveries from a sale order.",
-    "version": "16.0.1.3.1",
-    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "version": "16.0.1.4.1",
+    "author": "ForgeFlow, Therp BV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-workflow",
     "category": "Sales",
     "depends": ["sale_stock"],
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -1,5 +1,6 @@
 # Copyright 2019 ForgeFlow S.L.
 #   (http://www.forgeflow.com)
+# Copyright 2026 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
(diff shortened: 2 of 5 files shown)


ntsirintanis:
@rousseldenis please provide your precious insight :)

ntsirintanis:
@thomaspaulb @NL66278

sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -23,17 +20,64 @@ def action_done(self):
...
+    def _onchange_delivery_block_id(self):
+        """Set a default delivery block reason from partner or payment term,
+        but do not overwrite a value already set manually by the user"""
         for so in self:
-            if so.partner_id.default_delivery_block:
-                so.delivery_block_id = so.partner_id.default_delivery_block
-            else:
-                so.delivery_block_id = (
-                    so.payment_term_id.default_delivery_block_reason_id or False
+            if so.delivery_block_id:
```
NL66278, line 32: Is it not enough to keep the computed field computed, but to add these two lines that a value once set will not be overridden when partner of payment term changes?

ntsirintanis: Not really. Since delivery_block_id is the computed field itself, checking if so.delivery_block_id would prevent recomputation for both manual and automatic values. Beyond the fact that manual vs auto would be indistinguishable, it would also break expected behavior when partner or payment term changes.

The actual issue is that the field is both computed and editable. This commit makes it a regular field which defaults to something explicit, making sure that manual values are preserved when are input, while default behavior remains as is.

NL66278: I have to disagree with you on this one. Both in the onchange version and in the depends version an existing value for delivery_block_id would be left alone. If partner_id or payment_term_id changes, a value will be set for the delivery_block_id if not set already. But as onchange is not triggered on programmatic write and create you have to override those functions.  This also impacts the write function as you have to loop over the whole recordset again, and replace the batched write with single writes for those cases that have delivery_block_id not set already.

ntsirintanis: your approach works too, in my head  it’s more about choosing between cleaner semantics or lighter implementation. The cost here is adding more code in create() and write(). This is fine in my book, as ultimately the computed field is not a cleanly computed field. Rather, it is an editable field with fallback and defaults. The business rule here is really about preserving manual input first, default otherwise.

That said, it can be reworked towards a  "preserve existing value in compute + minimally  patch create/write” approach,  so we respect semantics and lighter implementation.

thomaspaulb: Hmm. Been staring at this for a while.

I think functionally, what would be most intuitive behaviour of this functionality is, the moment people choose a customer, or a payment term, and the sale order is still in Draft, the block reason is changed to whatever is the default for that customer or payment term. Unless a manual one is chosen, then it should pick that manual one. The field should never be emptied when it's filled, but, when it's filled manually and the payment term or partner is changed, I do believe it should switch to the new partner's default, if it's filled. Similarly for the payment term.

Evaluating the solutions against that kind of functional behaviour:

1. Editable computed field that does not overwrite pre-existing values (NL66278 proposal): codewise light, however if a partner is chosen and then another partner, the value will stick to the default of the first partner, which might be counterintuitive.
2. Onchange that only fills if a default is chosen - codewise light, seems functionally a good fit
3. Current solution - codewise not that light, functionally i think also a good fit (compute function runs, but manual writes or creates that include a value always trump the computation)

So unless I'm mistaken in my interpretation of how 3 works, I think I'm good with 3, however 2 seems an easier solution with the same functionality. I think ntsirintantis argument against that would be a regression in current functionality that also triggers the computation when partner or payment term is being written to programmatically; but I think there are lots of onchange-fields that don't get triggered programmatically, which is why normally you have to do something like a play_onchanges to get all default values in. I think that's acceptable)

sale_stock_picking_blocking/__manifest__.py
```diff
@@ -1,12 +1,13 @@
 # Copyright 2019 ForgeFlow S.L.
 #   (http://www.forgeflow.com)
+# Copyright 2026 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Sale Stock Picking Blocking",
     "summary": "Allow you to block the creation of deliveries from a sale order.",
-    "version": "16.0.1.3.1",
```
NL66278, line 8: No need to change version, that will be handled by the ocabot merge agent.

NL66278:
@thomaspaulb I think you could change your  remarks to an approval. @ntsirintanis I would still prefer the more light approach, because what would happen if for every writable computed field we would need to add coding to create and write functions. But I leave it to other people using this module or responsible for this repository to add their reviews to see what they think.

thomaspaulb (approved):

PR #4314, opened 2026-04-28
[18.0][FIX] sale_manual_delivery: Trying to improve performance during installation

divad1196:
Slight performance improvment of pre_init_hook

Same logic is kept but
- Unused joins with products are removed
- Unified duplicate join on stock_move table
- Used `WITH` statement for the subquery to make it more readable.

I replaced a `LEFT JOIN` with an `INNER JOIN` to skip records that don't need updated values.

Sadly, on big table with lot of records, the bottlneck is the write, not so much the read.
Still looking for the best solution to this issue.

### DIFF
--- a/sale_manual_delivery/hook.py
+++ b/sale_manual_delivery/hook.py
@@ -1,5 +1,6 @@
 def pre_init_hook(env):
     cr = env.cr
+
     cr.execute(
         """
         ALTER TABLE sale_order_line ADD COLUMN IF NOT EXISTS qty_procured numeric;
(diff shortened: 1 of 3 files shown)


sale_manual_delivery/tests/test_pre_init_hook.py
```diff
@@ -0,0 +1,200 @@
...
+        self.env.cr.execute(
+            "UPDATE sale_order_line "
+            "SET qty_procured = %s, qty_to_procure = %s "
+            "WHERE id = %s",
+            (qty_procured, qty_to_procure, line.id),
+        )
+        self.env.invalidate_all()
+
+    def test_pre_init_hook_fully_delivered(self):
+        """Tout livré: standard order, picking validated for the full qty.
```
jbaudoux:
```suggestion
        """Fully delivered: standard order, picking validated for the full qty.
```

sale_manual_delivery/tests/test_pre_init_hook.py
```diff
@@ -0,0 +1,200 @@
...
+
+        self._set_qty(line, None, None)
+        pre_init_hook(self.env)
+
+        qty_procured, qty_to_procure = self._read_qty(line)
+        self.assertEqual(qty_procured, 10.0)
+        self.assertEqual(qty_to_procure, 0.0)
+
+    def test_pre_init_hook_partially_delivered(self):
+        """Partiellement livré: stock move qty manually reduced below the
```
jbaudoux:
```suggestion
        """Partially delivered: stock move qty manually reduced below the
```

sale_manual_delivery/tests/test_pre_init_hook.py
```diff
@@ -0,0 +1,200 @@
...
+
+        self._set_qty(line, None, None)
+        pre_init_hook(self.env)
+
+        qty_procured, qty_to_procure = self._read_qty(line)
+        self.assertEqual(qty_procured, 4.0)
+        self.assertEqual(qty_to_procure, 6.0)
+
+    def test_pre_init_hook_nothing_delivered(self):
+        """Rien livré: a service line has no stock_move at all.
```
jbaudoux:
```suggestion
        """Nothing delivered: a service line has no stock_move at all.
```

jbaudoux (commented):
Can you check pre-commit and drop changes in `‎sale_planner_calendar`?

divad1196:
`sale_planner_calendar` was failling the pre-commit hook on my machine.
All good now, removed it

jbaudoux (approved):

sale_manual_delivery/hook.py
```diff
@@ -13,53 +20,69 @@ def pre_init_hook(env):
...
+                        ((sm.product_uom_qty / sm_product_uom.factor)
+                        * sol_product_uom.factor),
+                        SCALE(sol_product_uom.rounding)
+                    ) * -1
+                ELSE 0
+            END
+        ) AS qty_procured
+    FROM
+    sale_order_line AS sol
+    INNER JOIN stock_move AS sm ON (
```
TDu, line 67: By using the `INNER JOIN` are we not having a risk of excluding the lines without moves ? And the new columns added on sale order line will not be initialized with their correct values, but NULL ?

divad1196: @TDu You are right and that's the goal actually. If a line does not have moves for the moment, they don't need these values. When moves get created, the compute method is triggered and will set the value.

That's how we managed to reduce the time during installation.

TDu: Ok, good to know it will not be an issue.

TDu (approved):

rousseldenis (approved):

PR #4317, opened 2026-04-28
[19.0][ADD] sale_order_line_price_source: track unit price origin

Ricardoalso:
This module tracks from where each sale order line unit price comes from.

It extends ``sale.order.line`` with:
 - ``price_source`` (``pricelist``, ``product``, ``manual``)
 - ``price_source_pricelist_item_id`` (the applied ``product.pricelist.item`` when source is ``pricelist``)


PaoloYam (approved):

sale_order_line_price_source/models/sale_order_line.py
```diff
@@ -0,0 +1,99 @@
...
+            ("pricelist", "Pricelist Rule"),
+            ("product", "Product Sales Price"),
+            ("manual", "Manual Override"),
+        ],
+        default="product",
+        help="Indicates the source of the unit price: Pricelist rule, "
+        "product base price, or manual override.",
+    )
+
+    price_source_pricelist_item_id = fields.Many2one(
```
santostelmo, line 26: If the price list item is deleted, what happens to `price_source` ? You keep it as pricelist ?

santostelmo (approved):

ivantodorovich (approved):
Thanks!

sale_order_line_price_source/models/sale_order_line.py
Ricardoalso: Indeed, the `price_source` would be kept as the pricelist. However, this mimics the core behavior associated with the pricelist and the `action_update_prices` method. The new values from the pricelists are only applied when you click the "Update Prices" button. The same applies here: as soon as the `_reset_price_unit` method is called, the price_source will be updated (calling `_sync_price_source_from_pricing`).

I hope this answers your question :pray:

cyrilmanuel (approved):

cyrilmanuel:
hi @pedrobaeza  could you merge this one please ?

PR #4318, opened 2026-04-29
[18.0][FIX] sale_company_currency: Fix currency conversion (division not multiplication)

### LINKED ISSUE #4026: Company Currency in Sale Orders sale_company_currency

fsmw:
Change `order.amount_total * order.currency_rate` to `order.amount_total / order.currency_rate` in the currency conversion logic.

The conversion rate returned by `_get_conversion_rate(from_company, to_order)` is the amount of company currency per unit of order currency. To convert an order amount to the company currency, division is required, not multiplication. This aligns with how Odoo core uses `currency_rate`, e.g. in `_compute_partner_credit_warning` on `sale.order`.

Also fix the broken `setUpClass` where both order line dictionaries were placed inside a single `(0, 0, ...)` command tuple, which silently truncated to one line and broke `amount_total`. The new tests use `res.currency.rate` records instead of writing directly to the computed `currency_rate` field, so the full compute chain is exercised end-to-end.

Closes #4026

**Note on forward-port suggestion**: PR #3240 on 16.0 changed `/` to `*`, which introduced the bug. This fix (`*` to `/`) is the correct direction — forward-porting #3240 would re-introduce the multiplication bug.

### DIFF
--- a/sale_company_currency/models/sale_order.py
+++ b/sale_company_currency/models/sale_order.py
@@ -29,5 +29,5 @@ class SaleOrder(models.Model):
             if order.currency_id.id == order.company_id.currency_id.id:
                 to_amount = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
+                to_amount = order.amount_total / order.currency_rate
             order.amount_total_curr = to_amount
--- a/sale_company_currency/tests/test_sale_company_currency.py
+++ b/sale_company_currency/tests/test_sale_company_currency.py
@@ -139,11 +139,15 @@ class TestSaleOrder(TransactionCase):
         """Test rate change triggers recomputation of amount_total_curr."""
         self.sale_order.currency_id = self.currency_eur
         self._set_currency_rate(self.currency_eur, 0.5)
+        # Invalidate currency_rate to force Odoo to look up the new rate record
+        self.sale_order.invalidate_recordset(["currency_rate"])
         self.sale_order.flush_recordset(["amount_total_curr"])
         first_total_curr = self.sale_order.amount_total_curr
 
         # Change the rate and check if amount_total_curr is updated automatically
         self._set_currency_rate(self.currency_eur, 0.25)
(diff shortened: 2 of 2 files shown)


sale_company_currency/models/sale_order.py
```diff
@@ -29,5 +29,5 @@ def _compute_amount_company(self):
             if order.currency_id.id == order.company_id.currency_id.id:
                 to_amount = order.amount_total
             else:
-                to_amount = order.amount_total * order.currency_rate
+                to_amount = order.amount_total / order.currency_rate
```
grindtildeath, line 32:
```suggestion
                to_amount = order.company_currency_id.round(order.amount_total / order.currency_rate)
```

grindtildeath (commented):
Looks good, but we need to round after dividing to avoid floating point issues...

twalter-c2c (commented):
I agree with @grindtildeath regarding the rounding. Also, `test_05_automatic_recompute_on_rate_change` is failing (see 🔴 CI). Could you please take a look @fsmw?

PR #4320, opened 2026-04-30
[18.0][FIX] sale_global_discount: skip tax recompute when no discount

### LINKED ISSUE #4200: [18.0][BUG] sale_global_discount : Amount 0.01 or -0.01 without having global discount and subtraction from the taxable base.
When no global discount is assigned to orders, calculate 0.01 or -0.01 in the total global discount, thus modifying the taxable base.

fsmw:
When no global discount is assigned, `_compute_amounts` still re-computed all line taxes via `compute_all()`, which introduced floating-point rounding differences that caused $0.01 or -$0.01 ghost amounts in `amount_global_discount`, incorrectly modifying the taxable base.

**Fix**: When `discounts` is empty, skip the entire re-computation loop and preserve the correct standard Odoo amounts. Set `amount_global_discount = 0` explicitly.

Closes #4200

### DIFF
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -100,9 +100,17 @@ class SaleOrder(models.Model):
         res = super()._compute_amounts()
         for order in self:
             order._check_global_discounts_sanity()
+            discounts = order.global_discount_ids.mapped("discount")
+            if not discounts:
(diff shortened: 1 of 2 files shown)


fsmw:
This has been green for a while and fixes an active bug (#4200). Could a maintainer review when possible?

PR #4323, opened 2026-04-30
[19.0][FIX] sale_order_type: preserve manual payment term when changing invoice journal

lef-adhoc:
It's the same as #4273 

When a sale type has no payment_term_id and the user sets a manual term on the sale order, changing the invoice journal triggered a recompute chain (journal_id → company_id → sale_type_id) that caused super() to silently reset invoice_payment_term_id to the partner's default.

Skip the super() call for records where neither the partner nor the sale type changed and the sale type carries no term, so the chosen term is kept. Adds a regression test covering this scenario.


matiasperalta1 (approved):
LGTM

PR #4325, opened 2026-05-05
[ADD][18.0] sale_order_line_packaging_required

Shide:
Allows to make packaging required on Sale Order Form if product has packaging available.

https://www.loom.com/share/6cc81ade9ca44818bbc565126147f5c5

MT-14564 @moduon @chienandalu @fcvalgar @Gelojr please review if you want 😄


fcvalgar (changes requested):
Great work on the module.

I have functionally reviewed the sale_order_line_packaging_required module and performed the following tests:

Test 1: OK - Module installation, checking that it installs correctly and no dependency errors appear.
Test 2: OK - Product with packaging configured, verifying that when it is manually added to a sales order line, the Packaging field becomes required.
Test 3: OK - Product with packaging selected, checking that the line is saved correctly when the corresponding packaging is set.
Test 4: OK - Product without packaging configured, verifying that the order can be saved without setting packaging when the product has no available packaging.
Test 5: OK - Changing from a product without packaging to a product with packaging, checking that the packaging requirement is updated when the product is changed on the line.
Test 6: OK - Changing from a product with packaging to a product without packaging, verifying that the line can be saved without packaging when the new product has no available packaging.
Test 7: OK - Confirming a quotation with packaging set, checking that the standard sales flow is not blocked when the line is correctly completed.
Test 8: OK - Blocking save with an incomplete line, verifying that a manually added line with a product that has available packaging cannot be saved without setting the Packaging field.
Test 9: OK - Multiple products in the same order, checking that the requirement is applied line by line and only affects products with available packaging.
Test 10: OK - Removing the packaging from the product, verifying that in new lines the field is no longer required when the product no longer has available packaging.
Test 11: OK - Standard salesperson user, checking that the behavior also applies to a non-admin sales user.
Test 12: OK - Product without packaging with a standard salesperson user, verifying that they can create lines without packaging when the product has no available packaging.
**Test 13: Not OK - Adding a product from the catalog, checking that when a product with available packaging is selected from the catalog, the order can still be confirmed even though the line has no Packaging field set.**

I am not approving the PR yet because the behavior is not consistent when the product is added from the catalog. In that flow, the order can be confirmed even though the product has available packaging and the line has no packaging selected.

It would be necessary to review the line creation flow from the catalog so that it applies the same validation as when the line is added manually. Once the changes are uploaded, I can validate it again.

Image

Shide:
Closed in favor of https://github.com/OCA/sale-workflow/pull/4327

PR #4330, opened 2026-05-06
[18.0][ADD] sale_order_show_currency_rate: New Module

cristina-hidalgo-tecnativa:
Adds an informative exchange rate box to the sale order PDF report when the order currency differs from the company's base currency, showing the rate and order date. Uses the existing `currency_rate` field from core.

@Tecnativa TT62266
@pedrobaeza


sale_order_show_currency_rate/report/sale_order_report.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="report_saleorder_document_currency_rate"
+        inherit_id="sale.report_saleorder_document"
+    >
+        <xpath expr="//div[@name='so_total_summary']" position="after">
+            <t
+                t-if="doc.currency_id != doc.company_id.currency_id and doc.currency_rate"
```
pedrobaeza: For our use case, we need to show the equivalent in EUR for the quotation in RON, but the quotation is going to be in RON, which is the same currency than the company's one, so maybe you can call a method here that in this method do this check, but we override it in custom for being always true for the company chosen.

cristina-hidalgo-tecnativa: Done. The inline condition has been extracted to a `_show_currency_rate()` method on `sale.order`, so it can be overridden in custom modules.

sale_order_show_currency_rate/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _show_currency_rate(self):
```
pedrobaeza:
```suggestion
    def _show_currency_rate_in_report(self):
```

pedrobaeza (approved):

sale_order_show_currency_rate/models/sale_order.py
cristina-hidalgo-tecnativa: Renamed.

eduezerouali-tecnativa (approved):
Code LGTM + tested on runboat 

Image

PR #4332, opened 2026-05-11
[18.0][FIX] sale_procurement_group_by_line: avoid duplicate procurement

LauraCForgeFlow:
Delivering SO line to an internal location creates duplicate stock moves and duplicate MOs.

This module processes procurements for each SO line and then calls `super()`. Core Odoo relies on `_get_qty_procurement()` to skip lines that are already processed. However, that core method only counts moves going to "outgoing" locations.

With this fix, we track the SO lines processed by this module and filter them before calling `super()`. This ensures no line is processed twice, regardless of its destination type.

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -36,6 +36,7 @@ class SaleOrderLine(models.Model):
         )
         procurements = []
         groups = {}
+        procured_line_ids = set()
         if not previous_product_uom_qty:
             previous_product_uom_qty = {}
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
👍 Code review

MiquelRForgeFlow (approved):

JoanSForgeFlow (approved):

LauraCForgeFlow:
Hi @rousseldenis @pedrobaeza, can you check this fix? Thanks in advance! :)

PR #4336, opened 2026-05-14
[18.0][IMP] sale_order_lot_selection: Add quantity serial number validation when confirming the order

victoralmau:
Add quantity serial number validation when confirming the order

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT61756


victoralmau:
What do you think of this approach, @pedrobaeza ?

sale_order_lot_selection/models/res_company.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    sale_order_confirm_lot_qty_check = fields.Boolean(
```
pedrobaeza: Don't add this check. If you are selecting a serial number, it should be available as there can't be more than one quantity, and if not, don't reserve such serial number.

sale_order_lot_selection/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+            )[product.id]["free_qty"]
+            if (
+                float_compare(
+                    free_qty,
+                    line.product_uom_qty,
+                    precision_rounding=line.product_uom.rounding,
+                )
+                < 0
+            ):
+                return _("%(lot_name)s doesn't have enough quantity.") % {
```
pedrobaeza: The message should be specific for serial numbers, not lots.

pedrobaeza (changes requested):
This is a safeguard for avoiding the SO confirmation in case the same serial number is selected twice, but we should provide a way to not letting selecting a serial number if already in other SO. What you can do is to modify the available lot compute, and if the product is by serial number, remove from the selection those included in draft/sent orders.

What do you think?

victoralmau:
> This is a safeguard for avoiding the SO confirmation in case the same serial number is selected twice, but we should provide a way to not letting selecting a serial number if already in other SO. What you can do is to modify the available lot compute, and if the product is by serial number, remove from the selection those included in draft/sent orders.
> 
> What do you think?

I don't agree with this: does using a lot (selecting it) in a sales order line mean it can no longer be selected in any other order? What happens if that order is never confirmed? Do you have to cancel the order so it can be selected in another order? What happens if you don't want to cancel the entire order? 
I think that approach could be problematic; in any case, that approach could be implemented via `stock_reserve_sale` (or similar), automatically generating a reserve for that lot when a lot is selected in a sales order line, and release the reserve when the sales order is confirmed or canceled.

In any case, the change added in this PR is an additional and optional (configurable) feature that prevents orders from being confirmed with lots that do not have sufficient quantity.

sale_order_lot_selection/models/res_company.py
victoralmau: We had previously discussed internally adding this option to make this check optional and configurable, but that's fine—I'll remove it.

sale_order_lot_selection/models/sale_order.py
victoralmau: I'm sorry if this sounds like a silly question, but what do you mean by “serial numbers”? Lots and serial numbers are the same thing—they're `stock.lot` records. The message displays the corresponding record (`stock.lot`); which “serial number” do you want to see displayed?

victoralmau:
Another important point: 

I think it might be helpful to display a warning in the header of the sales order (similar to what happens, for example, with partners in some cases) if there is a lot that does not have sufficient stock; this message would appear only on pending orders (not confirmed or canceled) so that the user is aware of it before an error occurs when attempting to confirm the sales order.

pedrobaeza:
Yes, that's exactly what we want to achieve, Víctor. While you don't cancel the quotation or remove the reservation, you can't use it in other quotations. That's what a reservation means. And it works exactly the same as a good being reserved in a delivery order.

And I insist, this shouldn't be configurable, as by the nature of the serial number makes that you can't confirm a sales order if such serial is not available.

sale_order_lot_selection/models/sale_order.py
pedrobaeza: A serial number is a specialize type of lot that only admits quantity=1 in its quants. That's why I'm insisting all the time that you should make it non optional and limited to serial numbers.

pedrobaeza:
> Another important point:
> 
> I think it might be helpful to display a warning in the header of the sales order (similar to what happens, for example, with partners in some cases) if there is a lot that does not have sufficient stock; this message would appear only on pending orders (not confirmed or canceled) so that the user is aware of it before an error occurs when attempting to confirm the sales order.

What we want to get is the reservation for serial numbers, which should behave as I commented. For other usages like lots that aren't serial numbers, that improvements can be done, but are out of scope of our development.

victoralmau:
> Yes, that's exactly what we want to achieve, Víctor. While you don't cancel the quotation or remove the reservation, you can't use it in other quotations. That's what a reservation means. And it works exactly the same as a good being reserved in a delivery order.
> 
> And I insist, this shouldn't be configurable, as by the nature of the serial number makes that you can't confirm a sales order if such serial is not available.

Okay, no problem. I'll completely change my approach to achieve this (hopefully no one else will modify the batch in the sales order lines) without using `stock_reserve`.

In addition to what you mentioned, should we keep the error message when confirming the order?

sale_order_lot_selection/models/sale_order.py
victoralmau: Okay, but that limitation already exists https://github.com/OCA/sale-workflow/pull/4336/changes#diff-d00f9cc7b2420de7e3822754e17384af0bc26ac56d8ee94946bdc7cce669df1eR15

pedrobaeza:
> In addition to what you mentioned, should we keep the error message when confirming the order?

Yes, it's a safeguard for blocking things in case the serial numbers were assigned by code, bypassing the domain at the view level.

And remember, all of this is just for serial numbers.

sale_order_lot_selection/models/sale_order.py
pedrobaeza: Yeah, but behind a configuration option, and giving a message in a generic way. You should change `%(lot_name)s doesn't have enough quantity.` by `The serial number %(serial)s is not available`.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -72,7 +72,22 @@ def _compute_domain_lot_id(self):
...
                 )
-                domain = [("id", "in", available_quants.mapped("lot_id").ids)]
+                available_lots = available_quants.mapped("lot_id")
+                if sol.product_tracking == "serial":
+                    available_lots -= (
+                        self.env["sale.order.line"]
+                        .sudo()
+                        .search(
+                            [
+                                ("state", "not in", ("sale", "cancel")),
```
pedrobaeza: Better to put "in" for avoiding problems with custom states.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #4338, opened 2026-05-15
[18.0][IMP] sale_order_lot_selection: Add a method for the stock.quant domain used in `_compute_domain_lot_id()`

victoralmau:
Add a method for the `stock.quant` domain used in `_compute_domain_lot_id()`

@Tecnativa TT61756

### DIFF
--- a/sale_order_lot_selection/models/sale_order_line.py
+++ b/sale_order_lot_selection/models/sale_order_line.py
@@ -39,6 +39,19 @@ class SaleOrderLine(models.Model):
         for sol in self:
             sol.product_tracking = sol.product_id.tracking or sol.product_tracking
 
+    def _domain_lot_id_quant_domain(self):
+        """This method defines the domain that will be used to obtain the appropriate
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think for our use case, we need to extract this domain, as the idea is to put into `sale_order_lot_selection` the logic about serial numbers, and I don't know if we can reflect the logic through a domain in the quants instead of substracting from the available quants, those serials used in draft/sent orders.

victoralmau:
This change is **NOT** related to https://github.com/OCA/sale-workflow/pull/4336; there may be cases where you need to modify this domain in other modules (for example, by using all the stock locations from all the warehouses of that company); this allows for that, thus avoiding the need to completely rewrite the `_compute_domain_lot_id()` method in other modules or customizations.

pedrobaeza (approved):
OK

pilarvargas-tecnativa (approved):

PR #4339, opened 2026-05-16
[19.0][FIX] sale_delivery_state: avoid duplicate delivery_status column in list view

### LINKED ISSUE #4290: [19.0][BUG] sale_delivery_state: double column of delivery status in Sale order list
## Module
sale_delivery_state

## Describe the bug
in Sale order list, enable the delivery status column, it will show 2 column

## To Reproduce
19.0. not checked other version.

1. 
2.
3.

**Expected behavior**
only 1 column show.

**Additional context**
check on the runboat.

Image

fsmw:
## Issue

Closes #4290

In Odoo 19.0, sale_stock adds delivery_status to the sale.order list view. sale_delivery_state was adding a second column before invoice_status, resulting in two delivery_status columns visible in the tree view.

## Fix
Change the view inheritance from adding a new field to replacing the existing delivery_status field with the OCA version (which uses different state values: pending vs not_delivered, etc.), and lower priority to 99 so it applies after sale_stock.

## Changes
- Changed xpath from adding field at invoice_status position to replacing the existing delivery_status field
- Added priority=99 to ensure application order
- Renamed view for clarity

## Compatibility
- Backward compatible: doesn't change the field itself, only removes the duplicate column behavior
- OCA functional tests pass

### DIFF
--- a/sale_delivery_state/__manifest__.py
+++ b/sale_delivery_state/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Sale delivery State",
     "summary": "Show the delivery state on the sale order",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
     "category": "Product",
     "website": "https://github.com/OCA/sale-workflow",
     "author": "Akretion, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


fsmw:
This [FIX] is all green (including Codecov) and ready for review. Could a maintainer take a look?

BhaveshHeliconia (approved):
LGTM!

PR #4301, opened 2026-04-22
[19.0][MIG] sale_stock_delivery_state: Migration to 19.0

BhaveshHeliconia:

PR #4346, opened 2026-05-18
[19.0][FIX] sale_delivery_state: remove duplicate delivery_status column in tree view

### LINKED ISSUE #4290: [19.0][BUG] sale_delivery_state: double column of delivery status in Sale order list

fsmw:
## Description

In Odoo 19, `sale_stock` already adds the `delivery_status` field to the sale order tree view with badge widget and color decorations. This OCA module was adding it again, resulting in a duplicate column.

This PR removes the redundant tree view inheritance since core already provides this functionality.

## Changes

- Removed `view_order_tree_inherit_delivery_status` view record from `views/sale_order_views.xml`
- Added comment explaining why the view was removed

## Before/After

**Before:** Two `delivery_status` columns shown in sale order list view
**After:** Single column from Odoo core, with same badge widget and decorations

## Related

- Fixes #4290

### DIFF
--- a/sale_delivery_state/views/sale_order_views.xml
+++ b/sale_delivery_state/views/sale_order_views.xml
@@ -29,23 +29,8 @@
         </field>
     </record>
 
-    <record id="view_order_tree_inherit_delivery_status" model="ir.ui.view">
-        <field name="name">sale.order.tree</field>
-        <field name="inherit_id" ref="sale.view_order_tree" />
-        <field name="model">sale.order</field>
-        <field name="arch" type="xml">
-            <field name="invoice_status" position="before">
-                <field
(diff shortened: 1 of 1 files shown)


yostashiro:
@fsmw Can you please review this PR instead: https://github.com/OCA/sale-workflow/pull/4301? Since sale_delivery_state doesn’t depend on sale_stock, I don’t think the proposed change would be accepted.

PR #4349, opened 2026-05-19
[18.0][IMP] sale_order_split_strategy: force split, default from SO

HekkiMelody:
- Add a new field "force split" which automatically splits the SO on order confirmation
- Add new fields for "force split" and "split strategy" on contact, to be used as default on the SO.


sale_order_split_strategy/models/sale_order.py
```diff
@@ -173,3 +190,10 @@ def _postprocess_split_from(self):
 
     def _postprocess_split_to(self, origin_order):
         self.ensure_one()
+
+    def action_confirm(self):
+        orders_to_confirm = self
+        for order in self:
+            if order.force_split:
+                orders_to_confirm += order.action_split()
```
grindtildeath, line 198: The issue as I see it is to have split orders getting confirmed without previous review or that could block a bulk confirmation.

IMO this is an implementation choice. Should split orders be confirmed upon split, or should orders to be split not be confirmed together with order that should not be split, is a question that will probably have different answers based on specific needs.

That's why I would prefer not to have that in this module, and keep the module's scope focused on the ability to split an order.

sale_order_split_strategy/models/sale_order.py
```diff
@@ -12,11 +12,28 @@
...
     )
 
+    @api.depends("partner_id")
+    def _compute_split_strategy_id(self):
+        for order in self:
+            partner = order.partner_id
+            if not partner:
+                continue
+            order.force_split = partner.so_force_split
+            order.split_strategy_id = partner.so_split_strategy_id
```
grindtildeath, line 35: When we implemented this module, the split strategy was actually based on the delivery carrier, so I don't like having it tied to the partner. (since different goods had to be shipped using different carriers)

I would say that again this is IMO an implementation choice that should not be part of this module.

grindtildeath (changes requested):
First things first, thank you for your contribution.

However, please have a look at my answers. Although these changes could make sense for different implementations, I would prefer not to have it in this module, that IMO should only provide the split ability.

Here are my suggestions to avoid changing the scope of this module while still making the OCA benefit from you contribution:
- Have a sale_order_split_strategy_partner module that would define fields on res.partner and change the SO fields to computed one.
- Have a sale_order_split_on_confirm module that provides automatic split on confirmation based on the strategy. (I'd still suggest to add a configuration at company level to define if split orders have to be confirmed automatically or not upon split)

What do you think?

PR #4355, opened 2026-05-21
[18.0][IMP] sale_order_show_currency_rate: Show reference currency rate when order currency matches company currency

cristina-hidalgo-tecnativa:
The existing module only showed the informative exchange rate box on sale order PDFs when the order currency differed from the company currency. 

This improvement adds support for the case where both currencies are the same by allowing a reference currency to be configured per company (e.g. EUR for a RON company) via `Settings > Sales`. 

The box will only appear when there is a rate available for the reference currency at the order date, avoiding misleading information if rates have not been loaded yet.

@Tecnativa TT62266
@pedrobaeza @christian-ramos-tecnativa


sale_order_show_currency_rate/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    currency_rate_report_currency_id = fields.Many2one(
```
pedrobaeza: Change the name and the help for indicating that is only for sale. It can be `sale_order_currency_report_id`.

cristina-hidalgo-tecnativa: Done.

pedrobaeza (approved):

BhaveshHeliconia (approved):
LGTM!

PR #4363, opened 2026-05-26
[18.0][IMP] sale_elaboration: use state instead of parent.state

CarlosRoca13:
`parent.state` is only available when a view is rendered embedded inside a One2many (e.g. the order_line subform). When `sale.order.line` is opened as a standalone form (sale_order_line_view_form_readonly), there is no parent context, causing Owl to raise:

EvalError: Name 'parent' is not defined

`sale.order.line` already exposes `state = fields.Selection(related='order_id.state')`, so replacing `parent.state` with `state` is semantically equivalent and works in both embedded and standalone contexts.

cc @Tecnativa TT62707

### DIFF
--- a/sale_elaboration/views/sale_order_views.xml
+++ b/sale_elaboration/views/sale_order_views.xml
@@ -13,12 +13,12 @@
                     <field
                         name="elaboration_ids"
                         widget="many2many_tags"
-                        readonly="not product_updatable or parent.state in ('done', 'cancel')"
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):

eduezerouali-tecnativa (approved):

eduezerouali-tecnativa:
ping @pedrobaeza

PR #4364, opened 2026-05-26
[18.0][ADD] sale_multi_discount

LorenzoC0:
Depends on https://github.com/OCA/account-invoicing/pull/2364

This module adds a distribution-like field in the sale order line discount column, enabling infinite multiplicative discounts.

This module likely conflicts with some other modules that modify the behavior of the discount field, such as `sale_global_discount`. I could use some help figuring out which of these modules are incompatible, so I can add them as exclusions in the manifest. Thanks!


sale_multi_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,145 @@
...
+
+        ``sale.order._recompute_prices`` (invoked by *Update Prices*) calls
+        ``_compute_discount`` directly to overwrite the legacy ``discount``
+        field on each line. Since ``discount`` is now a read-only aggregate
+        of ``discount_distribution``, run the distribution recompute
+        instead — otherwise the pricelist would feed a single percentage
+        back through our ``write`` override and squash a freshly-edited
+        multi-discount rule into a single flattened value.
+        """
+        self._compute_discount_distribution()
```
eLBati, line 118: Why not calling super?

sale_multi_discount/models/sale_order_line.py
```diff
@@ -0,0 +1,145 @@
...
+        if self.discount_distribution:
+            res["discount_distribution"] = self.discount_distribution
+        return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if "discount" in vals and "discount_distribution" not in vals:
+                discount_val = vals.pop("discount")
+                if discount_val:
```
eLBati: if discount_val is 0, shouldn't we still empty discount_distribution?

sale_multi_discount/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "maintainers": ["LorenzoC0"],
+    "website": "https://github.com/OCA/sale-workflow",
+    "license": "LGPL-3",
+    "depends": [
+        "sale",
+        "account_multi_discount",
+    ],
+    "data": [
+        "views/sale_order_views.xml",
+        "report/sale_order_report.xml",
```
eLBati, line 20: We should probably extend sale.sale_portal_templates as well

sale_multi_discount/report/sale_order_report.xml
```diff
@@ -0,0 +1,29 @@
...
+            <t t-set="display_discount"
+               t-value="any(l.discount_distribution for l in lines_to_report)"/>
+        </xpath>
+
+        <!-- Replace discount cell with multi-discount display -->
+        <xpath expr="//td[contains(@t-if,'display_discount')]" position="replace">
+            <td t-if="display_discount" class="text-end text-nowrap">
+                <t t-if="line.discount_distribution and len(line.discount_distribution) > 1">
+                    <t t-foreach="line.discount_distribution" t-as="d" t-key="d_index">
+                        <span t-out="'%.2f' % d"/>%<t t-if="not d_last"> + </t>
```
eLBati: is it possible to use formatLang?

eLBati (changes requested):

sale_multi_discount/models/sale_order_line.py
eLBati: if `product_id` or `product_uom` or `product_uom_qty` change, the distribution should be emptied

sale_multi_discount/__manifest__.py
eLBati: we can also not do it for now

sale_multi_discount/report/sale_order_report.xml
LorenzoC0: I updated these lines, following the same way Odoo core treats decimals in the standard report. 

https://github.com/odoo/odoo/blob/18.0/addons/sale/report/ir_actions_report_templates.xml#L119

sale_multi_discount/models/sale_order_line.py
LorenzoC0: Done, added regression test `test_create_explicit_zero_discount_overrides_pricelist`

PR #4365, opened 2026-05-26
[18.0][ADD] sale_pricelist_multi_discount

LorenzoC0:
Depends on https://github.com/OCA/account-invoicing/pull/2364 and https://github.com/OCA/sale-workflow/pull/4364

This module adds a distribution-like field in the pricelist discount field, enabling infinite multiplicative discounts.

This module likely conflicts with some other modules that modify the behavior of the discount field in pricelists. I could use some help figuring out which of these modules are incompatible, so I can add them as exclusions in the manifest. Thanks!


sale_pricelist_multi_discount/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,9 @@
+This module configures multiple multiplicative discounts on pricelist
+rules and propagates them to sale order lines.
+
+It adds the `discount_distribution` JSON field on
+`product.pricelist.item` and exposes it through the Discount
+Distribution widget in the *Percentage* and *Formula* compute modes.
+When the rule has a non-empty distribution, the rule's `_compute_price`
+applies all discounts multiplicatively and the sale order line copies
+the same list into its own `discount_distribution` field.<>
```
eLBati: Remove the `<>`

sale_pricelist_multi_discount/models/product_pricelist_item.py
```diff
@@ -0,0 +1,74 @@
...
+    _inherit = "product.pricelist.item"
+
+    discount_distribution = fields.Json(
+        string="Discount Distribution",
+        default=list,
+        help="Ordered list of percentage discounts applied multiplicatively "
+             "by this pricelist rule.",
+    )
+
+    def _compute_price(self, product, quantity, uom, date, currency=None):
```
eLBati: instead of overriding _compute_price, is it possible to store the aggregated percentage into the native percent_price / price_discount whenever discount_distribution changes (compute/inverse). Then core's `_compute_price` works unmodified and the override disappears entirely.

Also, README should be changed about `percent_price`

eLBati (changes requested):
thanks

rousseldenis:
@LorenzoC0 What's the status of this ?

PR #4369, opened 2026-05-26
[19.0][ADD] sale_partner_display_ref

JasminSForgeFlow:
This module prefixes the customer reference (`res.partner.ref`) to the
partner's `display_name` whenever it is rendered inside a Sales view —
typically as `[C00123] Acme Corp` in the autocomplete dropdown of the
customer field on a sales quotation or order.

The decoration mechanism lives in the generic `partner_display_ref` module;
this module only injects the `partner_display_ref_field` context key (set to
`ref`) into the Sales views. Contacts, CRM, Invoicing and any other module that reads
`display_name` continues to see the plain partner name.

Depends on:

- [ ] https://github.com/OCA/partner-contact/pull/2384

@ForgeFlow


AaronHForgeFlow (commented):
Code LGTM. Perhaps you can consolidate tests into 1-2 tests. Having so many tests for so little code is too much I think.

sale_partner_display_ref/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @api.depends("ref")
+    @api.depends_context("formatted_display_name", "show_partner_ref")
+    def _compute_display_name(self):
+        super()._compute_display_name()
+        ctx = self.env.context
+        if not (ctx.get("formatted_display_name") and ctx.get("show_partner_ref")):
+            return
+        for partner in self:
+            if partner.ref:
```
ChristianSantamaria:
```suggestion
        for partner in self.filtered(lambda x: x.ref):
```

ChristianSantamaria (changes requested):
Hey, funtional and code review, 👍 👍 

I think this is a good module for improving contact visibility, but I believe this add-on could be more generic. For example, the base module could be "partner_display_ref", and another, dependent module, "sale_partner_display_ref", would add the necessary context to the sales views. This would allow for future application to other modules.

What do you think?

Anyway see minor change..

sale_partner_display_ref/models/res_partner.py
rousseldenis: @ChristianSantamaria You can use directly the field like `self.filtered("ref")`.

I was using `filtered()` like you suggested, BUT, `filtered()` do a loop on all records of the recordset. So, it is not advantageous in terms of performance.  @JasminSForgeFlow You can let the code as is.

rousseldenis (approved):

BhaveshHeliconia (approved):
Functional review LGTM!

MarinaAForgeFlow (approved):

PR #4371, opened 2026-05-27
[19.0][IMP] sale_order_mass_action: adds missing target states

glitchov:
The original module only permits to confirm in mass so this enhancement targets other states plus the locked flag.
The original idea comes from @rousseldenis in v14.0


sale_order_mass_action/tests/test_sale_order_mass_action.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertEqual("sent", self.sale.state)
+
+    def test_sale_lock(self):
+        # with an unlocked sale
+        # Launch the wizard on Sale Order
+        # Choose lock action
+        # Check if the sale order is locked
+        self.assertFalse(self.sale.locked)
+        self.wizard = self.wizard_obj.with_context(
+            active_model="sale.order", active_ids=[self.sale.id]
```
FrancoMaxime, line 79:
```suggestion
            active_model="sale.order", active_ids=self.sale.ids
```
Is it not better to do that?

FrancoMaxime (approved):
LGTM: code review (a minor comment in unit test)

rousseldenis:
@glitchov This is in fact a migration, don't forget to make PR title consistent.

PR #4372, opened 2026-05-28
[14.0][FIX] sale_procurement_group_by_line: Don't return super(), run_procu…

Kiplangatdan:
…rements executed twice

### DIFF
--- a/sale_procurement_group_by_line/model/sale.py
+++ b/sale_procurement_group_by_line/model/sale.py
@@ -25,6 +25,22 @@ class SaleOrderLine(models.Model):
         """
         return 8, self.order_id.id
 
+    def _get_qty_procurement(self, previous_product_uom_qty=False):
+        """Override to respect previous_product_uom_qty for lines already
+        processed by _action_launch_stock_rule.
+
+        Without this, modules like sale_mrp override _get_qty_procurement for
+        kit products and ignore previous_product_uom_qty, returning 0 for
+        nested kit components. This causes super()._action_launch_stock_rule
+        to run procurement a second time, doubling all stock moves.
+        """
+        self.ensure_one()
+        if previous_product_uom_qty and self.id in previous_product_uom_qty:
+            return previous_product_uom_qty[self.id]
+        return super()._get_qty_procurement(
+            previous_product_uom_qty=previous_product_uom_qty
+        )
+
     def _action_launch_stock_rule(self, previous_product_uom_qty=False):
         """
         Launch procurement group run method.
@@ -36,9 +52,6 @@ class SaleOrderLine(models.Model):
         groups = {}
         if not previous_product_uom_qty:
             previous_product_uom_qty = {}
-
-        processed_lines = self.env["sale.order.line"]
-
         for line in self:
             line = line.with_company(line.company_id)
             if line.state != "sale" or line.product_id.type not in ("consu", "product"):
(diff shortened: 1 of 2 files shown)


sale_procurement_group_by_line/model/sale.py
```diff
@@ -25,6 +25,22 @@ def _get_procurement_group_key(self):
...
+        processed by _action_launch_stock_rule.
+
+        Without this, modules like sale_mrp override _get_qty_procurement for
+        kit products and ignore previous_product_uom_qty, returning 0 for
+        nested kit components. This causes super()._action_launch_stock_rule
+        to run procurement a second time, doubling all stock moves.
+        """
+        self.ensure_one()
+        if previous_product_uom_qty and self.id in previous_product_uom_qty:
+            return previous_product_uom_qty[self.id]
```
thomaspaulb, line 39: This fixes the problem for nested kit products, however it triggers for any id that's already in `previous_product_uom_qty`. I would like to ensure that this does not cast too wide a net and returns the previous qty for cases where that shouldn't happen. Maybe we should also add an extra check if it's a kit product? `is_kit_product = self.env['mrp.bom']._bom_find(product=self.product_id, bom_type='phantom')`

sale_procurement_group_by_line/tests/test_sale_procurement_group_by_line.py
```diff
@@ -64,6 +65,28 @@ def _create_sale_order(self):
...
+    def _create_kit_product(self, name, components):
+        """Create a storable product with a phantom (kit) BoM."""
+        kit_product = self.product_model.create(
+            {
+                "name": name,
+                "categ_id": self.product_ctg.id,
+                "type": "product",
+            }
+        )
+        self.env["mrp.bom"].create(
```
thomaspaulb, line 77: This test wouldn't work here, because `sale_procurement_group_by_line` does not depend on `mrp`.

It also opens up the discussion, since this is more of a generic fix to the `sale_mrp` behaviour, where the fix should live. At least in some module that depends on `sale_mrp`.

Is it even limited to this module only or is it a generic Odoo issue? If generic, then either in OCB or in a separate fix module. And if it's still there in Odoo 18.0, then first with a PR to Odoo itself.

ntsirintanis: Agreed. Probably a good solution would be a glue module between this one and sale_mrp. What do you think?

ntsirintanis: Or a better one would be to fix this directly into sale_mrp. I'll check both

thomaspaulb: Still confused as to what actually triggers this. My best guess is that it's still what Dan pointed out in the beginning, the fact that this module [runs procurements](https://github.com/OCA/sale-workflow/blob/993a3f8bd570fe05c4e054423e40e339b4c52cfe/sale_procurement_group_by_line/model/sale.py#L101) and then relies on the `previous_product_uom_qty` parameter being passed and respected, which in the case of nested kit products, is not respected. So that's what your fix fixes, then indeed it should be going into a glue module of this module with `sale_mrp`.

Then again, if there are any other OCA modules or even core functionalities that rely on this `previous_product_uom_qty` methodology working well, then we'd rather need a standalong kit-fixing module or an Odoo/OCB patch to sale_mrp.

I'm leaning towards doing the first thing, clearly showing the issue with the test you already made, making it green.

Was looking around a bit and it seems a similar fix has been done in 19.0 core to address a different problem:

https://github.com/odoo/odoo/commit/d394c91ddc7189f2321cefd8c247ac22ba8a5425

ntsirintanis:
@Kiplangatdan let's close this one, in favor of https://github.com/OCA/sale-workflow/pull/4409

PR #4374, opened 2026-05-29
[IMP] sale_order_line_chained_move: Fill in related sale line in moves with SQL

rousseldenis:
As module install on an existing database can last, using an sql query is better.


BhaveshHeliconia (approved):
LGTM!

PR #4376, opened 2026-05-29
[19.0][MIG] sale_order_qty_change_no_recompute: Migration to 19.0 

apik-mgu:


pedrobaeza:
AFAIK, this module is not needed anymore, as Odoo core does it.

BhaveshHeliconia:
@pedrobaeza In odoo core, `prict_unit` is not calculated whenever quantity is changed, but `discount` is being recalculated, so for `discount` field we may need this module.

pedrobaeza:
That sounds very weird, and I think that can be considered a bug if happening. Anyway, the CI in this PR is red.

rousseldenis:
@apik-mgu

PR #3556, opened 2025-01-30
[16.0] [MIG] sale_order_rename: Migration to 16.0

patrickt-oforce:

PR #4378, opened 2026-06-01
[16.0] [MIG] sale_order_rename: Migration to 16.0

pcastelovigo:
Same as #3556 with the changes requested by @alexey-pelykh 

Functional test on my side


BhaveshHeliconia (approved):
Functional review LGTM!

alexey-pelykh (approved):
Thanks @pcastelovigo — reviewed as the follow-up to my #3556 review, and all three of my blocking points are now addressed:

1. `@api.constrains("name")` raises `ValidationError` instead of `UserError` — correct convention. ✔
2. The per-record `search_count` N+1 is gone, replaced by a single `read_group` over `(name, company_id)` with `lazy=False` collecting `__count > 1` keys. The logic is sound: the `name in ... and company_id in ...` domain is deliberately broad, but grouping by both fields re-separates the pairs, so no false positives. ✔
3. The noisy `_logger.error()` before the raise is removed. ✔

Test coverage is solid — `test_05_raise_exception_batch_create_with_same_name_in_same_company` exercises the exact batch/`read_group` path, and test_01–04 cover cross-company allow plus same-company create/rename raise. `attrs` in the view is the right syntax for 16.0.

Non-blocking FYI: the diff also renames two orders in `sale_pricelist_display_surcharge`'s test (`"Sale order"` → `"Sale order 1/2"`). That's necessary collateral — this module's cross-company uniqueness constraint would otherwise reject that module's two same-named orders under the repo-wide test install — but it's worth a one-line note in the PR description so the cross-module touch isn't a surprise.

LGTM.

Issue #4337, opened 2026-05-14
[16.0][sale_order_invoice_amount] Incorrect invoiced_amount value when SO and invoices share a foreign currency


Jaimermaccione:
## Module

sale_order_invoice_amount

## Describe the bug

An inconsistency has been detected in the `invoiced_amount` field when working with foreign currencies.

When a Sales Order is created in a currency different from the company currency (for example, EUR in a company configured with USD), the field **Invoiced Amount** displays an incorrect value.

Instead of showing the invoiced amount in the Sales Order currency, the system displays the amount in company currency without applying the proper conversion, interpreting it incorrectly as if it were expressed in the Sales Order currency.

### Example use case

The Sales Order `S00021` is created in EUR and contains two invoices also issued in EUR.

| Invoice | Amount |
|---|---|
| INV/2026/005 | 100.00 EUR |
| INV/2026/006 | 50.00 EUR |
| Real invoiced total | 150.00 EUR |
| Sales Order total | 150.00 EUR |

<img width="1162" height="677" alt="Image" src="https://github.com/user-attachments/assets/7ba5b4eb-d82f-4021-83a6-6798def69d7b" />

<img width="1162" height="677" alt="Image" src="https://github.com/user-attachments/assets/2cb8ff47-2610-4ab8-9e60-5505aa09918e" />

<img width="1151" height="617" alt="Image" src="https://github.com/user-attachments/assets/c13fd31e-1588-42c5-8d9b-af503cebcfbf" />

### Current behavior in Odoo

The value displayed in `invoiced_amount` does not match the sum of the invoices in EUR.

However, it matches exactly the total amount in USD shown in the invoice list view:

- 152.89 USD
- 76.45 USD
- Total: 229.34 USD

<img width="1838" height="275" alt="Image" src="https://github.com/user-attachments/assets/f6762347-7cdb-45c9-bf55-c92b1ca519ca" />

This indicates that the module is taking the company currency amount from the invoices and displaying it on the Sales Order as if it were expressed in the Sales Order currency.

The Sales Order is correctly marked as fully invoiced, but the displayed invoiced amount is incorrect due to a currency inconsistency.

## To Reproduce

**Affected versions**:
- 16.0

1. Configure a company with USD as company currency.
2. Create a Sales Order in EUR through a customer pricelist.
3. Confirm the Sales Order.
4. Create and post invoices also in EUR.
5. Review the `invoiced_amount` field on the Sales Order.

## Expected behavior

[...]

PR #4382, opened 2026-06-02
[16.0] [FIX] sale_order_invoice_amount: invoiced_amount value when SO and invoices share a foreign currency

Tisho99:
Fixes the use case explained in https://github.com/OCA/sale-workflow/issues/4337

   

Unit tests have been improved in a separate commit. The hardcoded invoiced amounts used in the asserts contained very specific decimal overflow errors that:
- Causes the tests be less robust.
- Are only present in this version (see [version 18.0](https://github.com/OCA/sale-workflow/blob/db78d3e0a47964e6fd732c8b1e43bf2b2db7e935/sale_order_invoice_amount/tests/test_sale_order_invoice_amount.py#L205-L210))

Regarding the FIX:
- The `rec.invoiced_amount += invoice.amount_total_signed` line only works correcly if both monetary fields have the same currency. rec.currency_id for rec.invoiced_amount, and invoice.company_currency_id for invoice.amount_total_signed. A condition to verify that has been added.

- The `rec.invoiced_amount += invoice.currency_id._convert(
                                invoice.amount_total_in_currency_signed,
                                rec.currency_id,
                                invoice.company_id,
                                invoice.invoice_date or fields.Date.today()
                            )` operation correctly converts all currencies to the sale currency, as the amount_total_in_currency_signed already has the total amount of the invoice converted to the currency of the invoice (In case the currency of the invoice and the company currency are different). Taking this into account, that operation has been moved to the else directive, applying it in the rest of the cases

A unit test of this use case has been added. This tests fails without this FIX

T-9647

### DIFF
--- a/sale_order_invoice_amount/tests/test_sale_order_invoice_amount.py
+++ b/sale_order_invoice_amount/tests/test_sale_order_invoice_amount.py
@@ -403,3 +403,53 @@ class TestSaleOrderInvoiceAmount(common.TransactionCase):
             0.0,
             "Uninvoiced Amount should be calculated.",
         )
+
+    def test_04_sale_order_invoiced_amount_different_currencies_company(self):
+        """Test when SO and invoices share a foreign currency"""
+        pricelist_eur = self.env["product.pricelist"].create(
+            {"name": "EUR Pricelist", "currency_id": self.currency_eur.id}
+        )
+        # Setting this pricelist sets the currency of the sale to eur
+        self.sale_order_1.pricelist_id = pricelist_eur
+        self.sale_order_1.action_confirm()
+        aml1 = self.order_line_1._prepare_invoice_line(
+            **{
+                "currency_id": self.currency_eur.id,
+            }
+        )
+        aml2 = self.order_line_2._prepare_invoice_line(
+            **{
+                "currency_id": self.currency_eur.id,
+            }
+        )
+        self.env["account.move"].create(
+            {
+                "move_type": "out_invoice",
+                "invoice_date": fields.Date.from_string("2024-01-01"),
+                "date": fields.Date.from_string("2024-01-01"),
+                "partner_id": self.res_partner_1.id,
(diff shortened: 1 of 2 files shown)


Jaimermaccione (approved):
@Tisho99 

Reviewed and tested on the PR runboat.

The **Invoiced Amount** field now correctly displays the invoiced amount in the Sales Order currency. In the tested scenario, the Sales Order shows the invoiced amount in the correct currency, matching both the Sales Order total and the sum of the related invoices.

The behavior is now consistent with the expected result.

**LGTM!**

sale_order_invoice_amount/tests/test_sale_order_invoice_amount.py
```diff
@@ -196,10 +199,11 @@ def test_02_sale_order_invoiced_amount_different_currencies_invoice(self):
             }
         )
         test_invoice.action_post()
-        self.assertEqual(
+        self.assertAlmostEqual(
             self.sale_order_1.invoiced_amount,
-            242.11,
-            "Invoiced Amount should be 242.11.",
+            242.0,
```
ValentinVinagre, line 204: The overflow is interesting; could you add a test that takes decimals into account and isn't all square? Ty

Tisho99: I do not recomend adding a test with the .11 decimals because invoiced lines of the order sums 242.00.

A decimal precision error ocurrs when creating the invoice because the prices are converted to other currency. However, hardcoding the raw value with the decimal error is not the correct solution. 

This test can fail in the following escenarios: 
- if the invoiced_amount is computed using a different valid field  (The case of this PR using amount_total_in_currency_signed)
- if the decimal precision changes
- if the decimal rounding method
- if the invoice is created in a different valid manner

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):
/ocabot merge patch

PR #4384, opened 2026-06-02
[19.0][ADD] sale_invoice_to_partner

JordiBForgeFlow:
This module lets you define, on a customer, a separate partner that is
in charge of receiving and paying its invoices.

Unlike a standard *Invoice Address* (a child contact of the same
company), the **Invoice To** partner can be a completely independent
customer. When a sales order is created for the customer, its *Invoice
Address* (``partner_invoice_id``) is set to the **Invoice To** partner,
so the invoices generated from the order are owed by that partner
instead of by the ordering customer.

Configuration
=============

1. Open a customer.
2. Go to the *Invoicing* tab, *Customer Invoices* section.
3. Set the **Invoice To** partner.

When a contact of a company has no **Invoice To** of its own, it
inherits the one defined on its commercial (parent company) partner.

Usage
=====

Create a sales order for the customer. The *Invoice Address* is
automatically populated with the **Invoice To** partner. It can still be
changed manually on the order if needed.


BhaveshHeliconia (approved):
Code and functional review LGTM!

rousseldenis (approved):

rvalyi:
Let's see if we are below the pypi limits today...
/ocabot merge nobump

PR #4385, opened 2026-06-04
[19.0][MIG] sale_order_lot_selection: Migration to 19.0

alan196:
**Description**
Migration of `sale_order_lot_selection` to 19.0.

The branch now contains the **full commit history of the 18.0 branch**, followed by:
- `[MIG]` Standard migration to 19.0 (`product_uom_id` rename, `_prepare_procurement_values()` signature, `self.env._` translations, view xpaths, test adaptations). All 18.0 features are preserved (serial availability check on confirmation, "exclude pending orders" setting, `domain_lot_id`, sale report `lot_id`, extension hooks).
- `[FIX]` Wrong state comparison in `_set_restrict_lot_id_from_sol` (the tuple was written as a single string, so moves were never reassigned). Backportable to 18.0.
- `[IMP]` New functionality extending the `qty_at_date_popover` OWL widget to select lots directly from the UI when checking availability (backportable):
  - **UI/UX**: Compact one-line layout per lot, displaying expiration dates when available.
  - **Sorting**: Automatic backend sorting using FEFO/FIFO logic for lots.
  - **Stability**: Pre-fills previous selections and prevents `RPC_ERROR` and duplicate lines on splitting.
  - **Validation**: The selected lot is dropped when it has no stock in the order warehouse, scoped per (warehouse, lot) pair so multi-warehouse batches are computed correctly.
  - **Customer visibility**: The selected lot and its expiration date (when `product_expiry` is installed) are shown on the sale order PDF report and on the customer portal order page.


HectorMerazChavez (approved):

rousseldenis:
@alan196 Thanks for this. Couldn't you put the enhancements in separate commits if they can be backported ?

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,147 @@
...
+                            sol.order_id.warehouse_id.view_location_id.id,
+                        ),
+                        ("quantity", ">", 0),
+                    ],
+                    limit=1,
+                )
+                if not quants:
+                    sol.lot_id = False
+
+    def action_split_lines_by_lot(self, lot_qty_dict):
```
BhaveshHeliconia, line 160: Use english comments.

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,147 @@
...
+        if not lot_qty_dict:
+            return self
+
+        # Evitar duplicaciones si se re-ejecuta el widget para el mismo producto.
+        # Eliminamos otras líneas del mismo producto en esta misma orden.
+        other_lines = self.order_id.order_line.filtered(
+            lambda ln: ln.product_id == self.product_id and ln.id != self.id
+        )
+        if other_lines:
+            other_lines.unlink()
```
BhaveshHeliconia: Why we delete sale order lines here?

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,147 @@
...
+            vals["restrict_lot_id"] = self.lot_id.id
+        return vals
+
+    @api.depends("product_id", "order_id.warehouse_id")
+    def _compute_lot_id(self):
+        for sol in self:
+            if sol.product_id != sol.lot_id.product_id:
+                sol.lot_id = False
+            elif sol.lot_id and sol.order_id.warehouse_id:
+                quants = self.env["stock.quant"].search(
```
BhaveshHeliconia: We can use single grouped read outside the loop for performance enhancements.

BhaveshHeliconia (changes requested):

sale_order_lot_selection/models/sale_order_line.py
alan196: Comments updated in english

sale_order_lot_selection/models/sale_order_line.py
alan196: Logic changed to update existing line, only delete if the quantity of the lot is 0

sale_order_lot_selection/models/sale_order_line.py
alan196: The logic was changed to address this comment

alan196:
> @alan196 Thanks for this. Couldn't you put the enhancements in separate commits if they can be backported ?

The enhancements now are in a separate commit

alan196:
@rousseldenis @BhaveshHeliconia @HectorMerazChavez could you help me to review 🙏

BhaveshHeliconia (approved):
LGTM!

victoralmau:
There is already an older migration PR (https://github.com/OCA/sale-workflow/pull/3985).

victoralmau:
This PR does not have the correct commit history; everything on the 18.0 branch is missing.

victoralmau (changes requested):
The full commit history for v18 is missing

ThomasBinsfeld (approved):

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,207 @@
...
+                [
+                    ("lot_id", "in", lot_ids),
+                    ("location_id", "child_of", wh.view_location_id.id),
+                    ("quantity", ">", 0),
+                ],
+                ["lot_id"],
+            )
+            valid_lots.update(q["lot_id"][0] for q in quants)
+
+        for sol in lines_to_check:
```
acsonefho, line 140: What about this flow:
If I edit (in one shot) 2 orders with 2 differents warehouse (lot with qty in WH1 but not in WH2).
For both, lot_id won't be dropped.

But if I validated them one by one, the SO with WH1 won't be dropped (ok) but it will be dropped for the second (ok).

So IMO there is a bug with the first scenario.

**To reproduce:**

```
# Copyright 2020 ACSONE SA/NV

from odoo import Command, api, models
from odoo.tests.common import TransactionCase


class TestComputeLotIdMultiWarehouse(TransactionCase):
    def test_compute_lot_id_multi_warehouse(self):
        """A lot has stock in WH1 but not in WH2.

        - Recomputed one by one: WH1 line keeps the lot, WH2 line drops it (OK).
        - Recomputed in one shot (ORM batches both lines together): the WH2 line
          wrongly keeps the lot because `valid_lots` is a global set shared
          across warehouses instead of being scoped per warehouse.
        """
        product = self.env["product.product"].create(
            {
                "name": "Tracked product",
                "type": "consu",
                "is_storable": True,
                "tracking": "lot",
            }
        )

        # WH1 = default warehouse of the company, WH2 = a second one
        wh1 = self.env["stock.warehouse"].search(
            [("company_id", "=", self.env.company.id)], limit=1
        )
        wh2 = self.env["stock.warehouse"].create({"name": "Warehouse 2", "code": "WH2"})

        # One single lot, with stock ONLY in WH1, nothing in WH2
        lot = self.env["stock.lot"].create(
            {"name": "LOT-001", "product_id": product.id}
        )
        self.env["stock.quant"]._update_available_quantity(
            product, wh1.lot_stock_id, 10.0, lot_id=lot
        )

        partner = self.env["res.partner"].create({"name": "Customer"})

        def _make_line(warehouse):
            order = self.env["sale.order"].create(
                {
                    "partner_id": partner.id,
                    "warehouse_id": warehouse.id,
                    "order_line": [
                        Command.create(
                            {
                                "product_id": product.id,
                                "product_uom_qty": 1.0,
                                "lot_id": lot.id,
                            }
                        )
                    ],
                }
            )
            line = order.order_line
            line.lot_id = lot  # ensure the lot is set before we start
            return line

        line_wh1 = _make_line(wh1)
        line_wh2 = _make_line(wh2)

        # --- One shot: both lines recomputed in a single batch (the bug) ---
        (line_wh1 + line_wh2)._compute_lot_id()

        self.assertEqual(
            line_wh1.lot_id,
            lot,
            "WH1 line has stock for the lot -> must be kept",
        )
        # >>> This assertion FAILS with the current code, proving the bug. <<<
        self.assertFalse(
            line_wh2.lot_id,
            "WH2 line has NO stock for the lot -> must be dropped, but the "
            "shared `valid_lots` set keeps it because the lot is valid in WH1.",
        )
```


**Probably part of the fix:**

```
    @api.depends("product_id", "order_id.warehouse_id")
    def _compute_lot_id(self):
        lines_to_check = self.filtered(
            lambda sol: sol.lot_id
            and sol.product_id == sol.lot_id.product_id
            and sol.order_id.warehouse_id
        )
        (self - lines_to_check).lot_id = False

        if not lines_to_check:
            return

        warehouses = lines_to_check.mapped("order_id.warehouse_id")
        valid_pairs = set()

        for wh in warehouses:
            wh_lines = lines_to_check.filtered(
                lambda line, w=wh: line.order_id.warehouse_id == w
            )
            lot_ids = wh_lines.mapped("lot_id").ids
            if not lot_ids:
                continue

            quants = self.env["stock.quant"].search_read(
                [
                    ("lot_id", "in", lot_ids),
                    ("location_id", "child_of", wh.view_location_id.id),
                    ("quantity", ">", 0),
                ],
                ["lot_id"],
            )
            valid_pairs.update((wh.id, q["lot_id"][0]) for q in quants)

        for sol in lines_to_check:
            if (sol.order_id.warehouse_id.id, sol.lot_id.id) not in valid_pairs:
                sol.lot_id = False
```

acsonefho (changes requested):

sale_order_lot_selection/models/sale_order_line.py
```diff
@@ -0,0 +1,207 @@
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    lot_id = fields.Many2one(
+        "stock.lot",
+        "Lot",
+        copy=False,
+        compute="_compute_lot_id",
+        store=True,
+        readonly=False,
```
ThomasBinsfeld, line 27: We need `precompute` to avoid side effect warnings with other modules like `sale_margin`.

```suggestion
        readonly=False,
        precompute=True,
```

ThomasBinsfeld (changes requested):

sale_order_lot_selection/models/sale_order_line.py
alan196: Fixed in the last push using the suggested (warehouse, lot) pair approach, and the test was added as `test_06_compute_lot_id_multi_warehouse`. Thanks for the detailed analysis!

sale_order_lot_selection/models/sale_order_line.py
alan196: Done. The branch was rebuilt on top of the full 18.0 history, so the field definition (including `precompute=True`) is preserved from 18.0.

alan196:
@victoralmau @acsonefho @ThomasBinsfeld The branch was rebuilt, all review comments are now addressed:

- **Full 18.0 history restored**: the branch now contains all 126 commits of the module from the 18.0 branch, so every 18.0 feature is preserved (serial availability check on confirmation, "exclude pending orders" setting, `domain_lot_id`, sale report `lot_id`, extension hooks) and authorship is kept.
- **`precompute=True`**: preserved from the 18.0 field definition.
- **Multi-warehouse `_compute_lot_id` bug**: fixed with the suggested (warehouse, lot) pair approach + regression test.
- The enhancements remain in a separate `[IMP]` commit so they can be backported, plus a small backportable `[FIX]` for a state comparison typo in `_set_restrict_lot_id_from_sol`.

Could you please review again? Thanks!

acsonefho (approved):

ThomasBinsfeld (approved):

mourad-ehm (approved):

victoralmau:
Please, cherry-pick https://github.com/OCA/sale-workflow/pull/4480 to commit history before migration commit.

The commit “_Add lot selection widget, report, and portal info_” adds a lot of things that would need to be analyzed in detail and that, in my opinion, should be in a separate PR after merging this one without that commit.

rousseldenis:
> Please, cherry-pick #4480 to commit history before migration commit.
> 
> The commit “_Add lot selection widget, report, and portal info_” adds a lot of things that would need to be analyzed in detail and that, in my opinion, should be in a separate PR after merging this one without that commit.

@alan196

PR #4389, opened 2026-06-08
[19.0] [MIG] sale_invoice_split_payment : Migration to 19.0

uncannycs:
[19.0] [MIG] sale_invoice_split_payment : Migration to 19.0


cgarcia-solvos (changes requested):
Code and functional review. LGTM. Only minor change:

sale_invoice_split_payment/tests/test_sale_order.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2025 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleOrer(TransactionCase):
```
cgarcia-solvos, line 7: tracking_disable should be applied here (or inherit from BaseCommon) to align with what the guide says:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#:~:text=Disable,way%29%2E


```suggestion
from odoo.addons.base.tests.common import BaseCommon


class TestSaleOrder(BaseCommon):
```

BhaveshHeliconia (commented):
Do we need this module? IMO this functionality can be achieved by just unchecking `Consolidated Billing` on `Create Invoice` wizard.

angelmoya:
> Do we need this module? IMO this functionality can be achieved by just unchecking `Consolidated Billing` on `Create Invoice` wizard.

Unchecking the box doesn't work for my workflow when dealing with many invoices and different payment methods. I need to consolidate invoices by payment type; for example, if I have 20 invoices across three payment types, I need them grouped into just three consolidated invoices.

I've tested this in runbot, and it works exactly as expected.

angelmoya (approved):
LGTM

BhaveshHeliconia:
> > Do we need this module? IMO this functionality can be achieved by just unchecking `Consolidated Billing` on `Create Invoice` wizard.
> 
> Unchecking the box doesn't work for my workflow when dealing with many invoices and different payment methods. I need to consolidate invoices by payment type; for example, if I have 20 invoices across three payment types, I need them grouped into just three consolidated invoices.
> 
> I've tested this in runbot, and it works exactly as expected.

@angelmoya Thanks for explaining use case.

BhaveshHeliconia (approved):
LGTM!

sale_invoice_split_payment/tests/test_sale_order.py
rousseldenis: @uncannycs

PR #4392, opened 2026-06-10
[19.0][IMP] sale_order_type: allow ordering default types

yankinmax:
Add sequence based ordering for sale order types so the fallback default type can be configured predictably.

### DIFF
--- a/sale_order_type/README.rst
+++ b/sale_order_type/README.rst
@@ -42,6 +42,9 @@ You can see sale types as lines of business.
 You are able to select a sales order type by partner so that when you
 add a partner to a sales order it will get the related info to it.
 
+When no partner or context default applies, the first available sale
+order type by sequence is used as the default.
+
 Additionally, it adds a warning message to notify users when there is a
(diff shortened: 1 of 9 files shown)


BhaveshHeliconia (approved):
LGTM!

sale_order_type/__manifest__.py
```diff
@@ -8,7 +8,7 @@
 
 {
     "name": "Sale Order Type",
-    "version": "19.0.1.5.0",
+    "version": "19.0.1.5.1",
```
ivantodorovich: This will be done by OCA bot upon merge.

ivantodorovich (approved):
Thanks! Pre-approving

sale_order_type/__manifest__.py
yankinmax: I was thinking about it and decided, better "safe than sorry" (says one classic :smile: )

yankinmax:
Thank you for the reviews guys

rousseldenis (approved):

PR #4394, opened 2026-06-10
[19.0][FIX] sale_stock_picking_blocking: allow removing block with auto_done

Ricardoalso:
When auto_done_setting is active, saving a sale order after removing delivery_block_id could still raise ValidationError.

The stored compute for delivery_block_id could reapply partner/payment-term defaults on save, making the constraint fail again.

### DIFF
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -17,10 +17,16 @@ class SaleOrder(models.Model):
         store=True,
     )
 
+    @property
+    def _user_has_auto_done_setting_group(self):
+        """Whether the current user has the "auto_done_setting" group."""
+        return self.env.user.has_group("sale.group_auto_done_setting")
+
     @api.constrains("delivery_block_id")
     def _check_not_auto_done(self):
-        auto_done = self.env.user.has_group("sale.group_auto_done_setting")
-        if auto_done and any(so.delivery_block_id for so in self):
+        if self._user_has_auto_done_setting_group and any(
+            so.delivery_block_id for so in self
+        ):
             raise ValidationError(
                 self.env._(
                     'You cannot block a sale order with "auto_done_setting" active.'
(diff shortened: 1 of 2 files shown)


sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -32,7 +38,11 @@ def _compute_delivery_block_id(self):
         """Add the 'Default Delivery Block Reason' if set in the partner
         or in the payment term."""
         for so in self:
-            if so.partner_id.default_delivery_block:
+            if so._user_has_auto_done_setting_group:
```
ivantodorovich, line 41: FTR, It'd be better to check this outside the loop, and assign all the records at once


But honestly I'm not so sure about this change. I may be missing something though..

The `sale.group_auto_done_setting` is a global setting .. so why even attempt to use this module with that setting enabled ? 

I guess the exception in `_check_not_auto_done` was there in the first place to catch the misconfiguration and let the user know about this.

With this new change, though, it will silently be ignored.

Are you sure this isn't just a misconfiguration you're trying to solve?

sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -17,10 +17,16 @@ class SaleOrder(models.Model):
         store=True,
     )
 
+    @property
+    def _user_has_auto_done_setting_group(self):
+        """Whether the current user has the "auto_done_setting" group."""
+        return self.env.user.has_group("sale.group_auto_done_setting")
+
```
ivantodorovich, line 24:
```suggestion
```

useless method IMHO.
there's no perf benefit, no lines of code avoided

sale_stock_picking_blocking/models/sale_order.py
Ricardoalso: Just checked it with a key user and indeed it was a security/groups misconfiguration. I will close this PR

Ricardoalso:
https://github.com/OCA/sale-workflow/pull/4394#discussion_r3388929754

PR #4398, opened 2026-06-11
[FIX] sale_product_set: Update wizard view to avoid horizontal scrollbar

wekeesler:
### Summary

* Removed group wrapping one2many field so that the tree view renders in the bounds of the wizard pop-up.
* Set display_type field to invisible because it is a technical field.

### Before

sale_product_set_wizard_before

### After

sale_product_set_wizard_after


BhaveshHeliconia (approved):
LGTM!

rousseldenis (approved):

PR #4399, opened 2026-06-12
[ADD] sale_order_service_description: new module

NICO-SOLUTIONS:
Adds a service description to Sales Orders


rousseldenis:
@NICO-SOLUTIONS Thanks for this. Could you maybe add a CONTEXT section in order to explain in which case this is useful. Thanks!

NICO-SOLUTIONS:
@rousseldenis 
Thanks for the suggestion! I've added a CONTEXT section.

There is no real business logic involved. It only provides a free-text field on the sales order to store a service description, which can be used for searching, filtering, and grouping independently of e.g. products or tags.

rousseldenis:
> @rousseldenis Thanks for the suggestion! I've added a CONTEXT section.
> 
> There is no real business logic involved. It only provides a free-text field on the sales order to store a service description, which can be used for searching, filtering, and grouping independently of e.g. products or tags.

In fact, what do you mean by 'service description' ?

NICO-SOLUTIONS:
@rousseldenis 
It's simply a free-text field for describing the services to be delivered as part of the sales order. It's just meant to store a free-text description of the agreed scope of work.

PR #4403, opened 2026-06-15
[18.0][FIX] sale_order_product_availability_inline: Using deprecated warehouse context key…

carlosdauden:
… instead of warehouse_id

@Tecnativa

ping @sergio-teruel @CarlosRoca13


carlosdauden:
Is needed this in 18.0?

PR #4405, opened 2026-06-16
[19.0][FIX] sale_stock_picking_blocking: Preserve delivery block on recompute

bobslee:
The delivery_block_id field is a stored computed field, but its compute unconditionally overwrote the stored value with the partner/payment-term default (or False when none is configured). Any recompute therefore wiped a manually-set blocking reason: changing the partner/payment term, or an upgrade/redeploy recomputing the stored field. Once cleared, deliveries were launched for orders that should have been blocked, which surfaced as "the block works on install but is gone a day later".

Guard the compute so it only seeds the default when no reason is set yet, preserving a manually-chosen reason on every recompute. Add precompute=True and readonly=False so an explicitly-entered value also survives the create-time pass (the first-save wipe of issue #3815, fixed for create by PR #3906). Complete the @api.depends so changing a partner/payment-term default propagates to orders that have no block yet.

Align the _action_launch_stock_rule override signature with core 19.0 and add regression tests covering the recompute and partner-change paths.

### DIFF
--- a/sale_stock_picking_blocking/models/sale_order.py
+++ b/sale_stock_picking_blocking/models/sale_order.py
@@ -15,6 +15,8 @@ class SaleOrder(models.Model):
         string="Delivery Block Reason",
         compute="_compute_delivery_block_id",
         store=True,
+        readonly=False,
+        precompute=True,
     )
 
     @api.constrains("delivery_block_id")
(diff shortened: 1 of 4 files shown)


sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -27,11 +29,23 @@ def _check_not_auto_done(self):
                 )
             )
 
-    @api.depends("partner_id", "payment_term_id")
+    @api.depends(
+        "partner_id",
+        "partner_id.default_delivery_block",
```
rousseldenis, line 34: @bobslee You cannot do that. As a change on partner side value will recompute on every sale.

sale_stock_picking_blocking/models/sale_order.py
```diff
@@ -27,11 +29,23 @@ def _check_not_auto_done(self):
...
+        "payment_term_id",
+        "payment_term_id.default_delivery_block_reason_id",
```
rousseldenis, line 36: Same here

PR #4188, opened 2026-03-03
[18.0][MIG] sale_order_line_display_stock_per_warehouse

metaminux:
Standard migration from PR #3451 

CC @Kev-Roche @mourad-ehm

PR #4406, opened 2026-06-18
[19.0][ADD] sale_stock_qty_by_warehouse

sbejaoui:
Display the quantity by warehouse for a product inside a sale order line

image

Thanks @benwillig for this nice module


sale_stock_qty_by_warehouse/models/sale_order_line.py
```diff
@@ -0,0 +1,72 @@
...
+        for rec in self:
+            product = rec.product_id
+            rec.display_qty_by_warehouse_widget = (
+                product.type == "consu" and product.is_storable
+            )
+
+    @api.depends("product_id")
+    def _compute_qty_by_warehouse_widget_data(self):
+        warehouses = self.env["stock.warehouse"].search([])  # pylint: disable=no-search-all
+        locations = warehouses.mapped("lot_stock_id")
```
rousseldenis, line 30: @sbejaoui Is it really correct as IMHO, you should use `view_location_id` field instead.

rousseldenis:
In fact, the same philosophy module exists: https://github.com/OCA/sale-workflow/pull/4188

In that one, I'd prefer:

- Using `warehouse` in context, instead of `lot_stock_id`
- The option on warehouse to display quantity or not

But I'd prefer your widget :-)


Maybe something to do to merge both concepts. @Kev-Roche

rousseldenis:
@sbejaoui @benwillig Could you review #4188 ?

benwillig:
>     * Using `warehouse` in context, instead of `lot_stock_id`

It depends on the use case, but it could be a setting I guess.

PR #4408, opened 2026-06-18
[19.0][FIX] sale_order_global_discount: avoid general discount on dow…

adrip-s73:
In Odoo 17, `sale.advance.payment.inv._prepare_base_downpayment_line_values()`
created down payment sale order lines with an explicit `discount: 0.0`. The
invoice line was then built through `sale.order.line._prepare_invoice_line()`,
which copied `self.discount`, but that value was already zero for down payment
lines.

In Odoo 19, `sale.advance.payment.inv._create_invoices()` was refactored to
compute down payments from tax base lines. It now calls
`account.tax._prepare_down_payment_lines()` and then
`sale.order._create_down_payment_lines_from_base_lines()`, which delegates to
`sale.order._prepare_down_payment_line_values_from_base_line()`. That method
builds the down payment sale order line from the computed base line values but
does not explicitly reset `discount` to zero.

This matters for `sale_order_general_discount` because the source sale order
lines have their `discount` computed by `sale.order.line._compute_discount()`
from `order_id.general_discount`. The Odoo 19 down payment computation starts
from those discounted order lines, and the generated invoice line still uses
`sale.order.line._prepare_invoice_line()`, which copies `discount: self.discount`
to the `account.move.line`.

This change restores the previous behavior by forcing down payment sale order
lines to keep `discount = 0.0`. Down payments already represent a computed
portion of the order amount, so applying the sale order general discount again
on the invoice line would discount the advance payment twice / display an
unintended discount.

### DIFF
--- a/sale_order_general_discount/models/sale_order_line.py
+++ b/sale_order_general_discount/models/sale_order_line.py
@@ -20,8 +20,17 @@ class SaleOrderLine(models.Model):
             # the case where a discount was set to a value != 0 and then
             # set again to 0 to remove the discount on all the lines at the same
             # time
-            if not line.product_id.bypass_general_discount and (
-                line.order_id.general_discount or line.order_id._origin.general_discount
+            if (
+                not line._is_downpayment_line()
+                and not line.product_id.bypass_general_discount
+                and (
+                    line.order_id.general_discount
+                    or line.order_id._origin.general_discount
+                )
             ):
                 line.discount = line.order_id.general_discount
         return res
+
+    def _is_downpayment_line(self):
+        self.ensure_one()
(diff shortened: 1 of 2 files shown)


Reyes4711-S73 (approved):
LGTM

Raul-S73 (approved):

BhaveshHeliconia (approved):
LGTM!

rousseldenis:
@adrip-s73 Thanks for this. Could you maybe add a test for that change?

adrip-s73:
@rousseldenis tests done!

rousseldenis (approved):

PR #4409, opened 2026-06-19
[14.0][ADD] sale_procurement_group_by_line_mrp

ntsirintanis:
### Problem
When both `sale_procurement_group_by_line` and `sale_mrp` are installed, confirming a sale order with a **nested kit product** (a kit whose BoM contains a component that is itself a kit) creates doubled stock moves. One set from `sale_procurement_group_by_line` and a second set from native `_action_launch_stock_rule`.

**Root cause:**
 `sale_mrp` overrides `_get_qty_procurement` for kit products to use `_compute_kit_quantities()` instead of counting moves directly. For nested kits, `_compute_kit_quantities` filters moves by `m.bom_line_id == bom_line`,  but the inner kit's component moves are tagged with the inner kit's BoM lines, not the outer kit's. So it returns 0, native code thinks nothing has been procured yet, and runs `procurement.group.run()` a second time.
The deeper root cause is in `_compute_kit_quantities` itself not handling nested kits recursively, but fixing that is more invasive and out of scope here.

### Solution
This glue module overrides `_get_qty_procurement` to check `previous_product_uom_qty` first for kit products. When `_action_launch_stock_rule` has already processed a line, `previous_product_uom_qty` is populated with the procured qty, returning it early signals to native code that procurement is complete, preventing the second run.

Continuation of: https://github.com/OCA/sale-workflow/pull/4372


sale_procurement_group_by_line_mrp/models/sale_order_line.py
```diff
@@ -0,0 +1,33 @@
...
+
+        sale_mrp overrides _get_qty_procurement for kit products and ignores
+        previous_product_uom_qty, using _compute_kit_quantities() instead.
+        For nested kits, the inner kit's component moves don't match the outer
+        kit's BoM, so it returns 0, causing super()._action_launch_stock_rule
+        to run procurement a second time and double all stock moves.
+
+        By checking previous_product_uom_qty first (only for kit products),
+        we signal that procurement was already handled upstream, without
+        affecting the default behavior for non-kit products.
```
thomaspaulb, line 20: Something like this or the description should also go into the readme, otherwise I think this is good!

PR #4423, opened 2026-06-24
[16.0][IMP] portal_sale_order_search: also search by parent company name

nobuQuartile:
**Summary**
When searching by "Customer" or "All" in the portal sale order list, the search was only matching on the direct contact name (partner_id.name)
Orders where the partner is a contact linked to a parent company could not be found by searching the company name
Added partner_id.parent_id.name to the search domain for both partner and all search types so that searching by company name also returns orders whose partner belongs to that company

@qrtl
QT6930

### DIFF
--- a/portal_sale_order_search/controllers/main.py
+++ b/portal_sale_order_search/controllers/main.py
@@ -15,6 +15,7 @@ class CustomerPortalSaleOrderSearch(portal.CustomerPortal):
             "name": {"label": _("Order Ref"), "input": "name"},
             "ref": {"label": _("Order Customer Ref"), "input": "ref"},
             "partner": {"label": _("Customer"), "input": "partner"},
+            "company": {"label": _("Customer Company"), "input": "company"},
         }
 
     def _get_search_order_domain(self, search):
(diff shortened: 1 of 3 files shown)


pilarvargas-tecnativa (changes requested):
I wonder if adding partner_id.parent_id.name under the “Customer” search option could be a bit confusing, as the visible customer of the order may still be the child contact. Would it make more sense to add a dedicated “Company” search input, while keeping it included in “All”?

AungKoKoLin1997:
> Would it make more sense to add a dedicated “Company” search input, while keeping it included in “All”?

@pilarvargas-tecnativa I added a new search "Customer Company" using `commercial_partner_id`. 
Can you please review?

pilarvargas-tecnativa (approved):
LGTM, thanks for the contribution!

yostashiro (approved):
Code review. LGTM.

smorita7749 (approved):
Code and functional review: LGTM!

rousseldenis (approved):

PR #4427, opened 2026-06-26
[18.0][ADD] sale_line_vendor_comment: New module

pilarvargas-tecnativa:
@Tecnativa TT57654

@CarlosRoca13 @juancarlosonate-tecnativa please review


juancarlosonate-tecnativa (changes requested):
It seems this module duplicates the functionality of sale_order_line_note:
https://github.com/OCA/sale-workflow/tree/18.0/sale_order_line_note

CarlosRoca13:
I'm not sure at all. With this module we want to fill the comment with the comment set on supplier.

What I think we need to do here is make the field compute, store, readonly=False to prefill it with the supplier comment added by `product_supplierinfo_comment`.

What do you think?

CarlosRoca13 (approved):
It's well likr it is right now

carlosdauden (approved):
The field added by the module
https://github.com/OCA/sale-workflow/tree/18.0/sale_order_line_note
is of type ‘text’ rather than ‘char’, and may also be intended for internal comments or for use with the client, whereas the field added by this module is for communication with the supplier.

juancarlosonate-tecnativa (approved):

PR #4430, opened 2026-06-29
[FIX] sale_global_discount: group lines by tax combination to avoid rounding errors

### LINKED ISSUE #4397: [19.0] [sale_global_discount] Rounding error on order total with multiple lines and compound taxes (e.g. IVA)
## Module
sale_global_discount

## Describe the bug
When a sale order has multiple lines with compound taxes (e.g. IVA 21% + Recargo de Equivalencia 5.2%), the total amount shown is a few cents higher than expected due to accumulated rounding errors.

The root cause is in `_compute_amounts` in `sale_order.py`: taxes are computed **line by line** and then summed, which causes each line's tax amount to be rounded individually before accumulation:

```python
amount_discounted_tax += sum(
    t.get("amount", 0.0) for t in discounted_tax.get("taxes", [])
)
```

Odoo core avoids this by grouping lines by tax combination before computing taxes, so rounding is applied once per group rather than once per line. With many lines and multiple taxes, the accumulated rounding error can reach several cents.

## To Reproduce

**Affected versions**: 19.0

1. Configure two taxes on products: IVA 21%
2. Create a sale order with 50+ lines, all with both taxes applied
3. Add a global discount to the order
4. Check the total amount shown in the order

**Expected behavior**
The total should match the result of computing taxes on the grouped base amount (as Odoo core does), not the sum of individually rounded per-line tax amounts.

Example: with a taxable base of 438.34 €, IVA 21% and 59 order lines, the displayed total is 553.22 € when the correct value is 553.18 €.

**Additional context**

The fix should group lines by tax combination before calling `compute_all`, mirroring Odoo core's approach in `_compute_amounts`. Instead of:

```python
for line in order.order_line:
    discounted_tax = line.tax_ids.with_context(force_price_include=False).compute_all(
        discounted_subtotal, ...
    )
    amount_discounted_tax += sum(
        t.get("amount", 0.0) for t in discounted_tax.get("taxes", [])
    )
```

The base amounts should be grouped by tax combination first, then `compute_all` called once per group, so rounding is only applied once per tax group rather than once per line.

SaraMohamedlorim:
When a sale order has multiple lines with compound taxes, computing taxes line by line causes accumulated rounding errors because each line's tax amount is rounded individually before accumulation.

The fix groups order lines by their tax combination before calling compute_all, mirroring Odoo core's approach. This ensures rounding is applied once per tax group rather than once per line, eliminating the accumulated rounding error.

Fixes #4397

### DIFF
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -130,14 +130,16 @@ class SaleOrder(models.Model):
                 tax_groups[tax_key]["base"] += discounted_subtotal
             # Compute taxes once per group to avoid rounding accumulation
             for group in tax_groups.values():
-                discounted_tax = group["tax_ids"].with_context(
-                    force_price_include=False
-                ).compute_all(
-                    group["base"],
-                    group["currency"],
-                    1.0,
-                    product=group["product"],
-                    partner=group["partner"],
+                discounted_tax = (
+                    group["tax_ids"]
+                    .with_context(force_price_include=False)
+                    .compute_all(
+                        group["base"],
+                        group["currency"],
+                        1.0,
+                        product=group["product"],
+                        partner=group["partner"],
+                    )
                 )
                 amount_discounted_tax += sum(
                     t.get("amount", 0.0) for t in discounted_tax.get("taxes", [])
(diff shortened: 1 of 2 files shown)


rousseldenis:
@SaraMohamedlorim Could you check tests ?

PR #4431, opened 2026-06-29
[18.0][MIG] sale_order_qty_change_no_recompute: Migration to 18.0

palomagrc93:
Migration to v18

cc https://github.com/APSL 48441

@shirashi3771 @lbarry-apsl @miquelalzanillas @peluko00 @BernatObrador @cubells @miquelpascual @ppyczko @isugac @javierobcn please review.


shirashi3771 (approved):
LGTM!

peluko00 (approved):
LGTM! Code review and tested in runboat

BhaveshHeliconia (approved):
LGTM!

ppyczko (approved):

sale_order_qty_change_no_recompute/tests/test_sale_order_qty_change.py
rousseldenis: @palomagrc93 To avoid this line, you can use `odoo.addons.base.tests.common.BaseCommon` instead

rousseldenis:
Could you also squash your migration commits ? Thanks

pedrobaeza (changes requested):
This module is theoretically not needed in this version, as now core doesn't recompute prices when changing quantity. Why are you migrating it?

palomagrc93:
Hi @pedrobaeza, thanks for pointing that out, You're right that price_unit is now protected natively in 18.0 via technical_price_unit. However, _compute_discount still recomputes when quantity changes with no equivalent protection, so the module still has value for use cases where a manually set discount should be preserved when quantity is updated.

pedrobaeza:
Then the README should be updated to reflect such changes in the scope.

palomagrc93:
> Then the README should be updated to reflect such changes in the scope.

Done! I've updated the README to reflect it.

pedrobaeza:
OK, thanks. Please squash this last commit into the migration one.

palomagrc93:
> OK, thanks. Please squash this last commit into the migration one.

Done!

pedrobaeza (approved):
/ocabot merge nobump

PR #4435, opened 2026-06-30
[18.0][FIX] sale_order_secondary_unit: keep packaging and secondary qty in sync

mav-adhoc:
## Problem
When a product is sold using both a packaging and a secondary unit of
measure, changing the number of packages (`product_packaging_qty`) on the
sale order line did not recompute `product_uom_qty` nor `secondary_uom_qty`.

The module overrode `_compute_product_uom_qty` to call `super()` (base
packaging → quantity logic) and then re-derive the quantity from the
secondary unit. Since `secondary_uom_qty` is itself computed from
`product_uom_qty`, this created a recursive dependency between both fields
and the packaging compute, and the packaging change was discarded.

## Fix
Move the `secondary unit → product_uom_qty` conversion out of the compute
and into an `inverse` on `secondary_uom_qty`:

- `product_uom_qty → secondary_uom_qty` stays an `@api.depends` compute, so
  it reacts to any quantity change (manual, packaging-driven, programmatic).
- `secondary_uom_qty → product_uom_qty` is handled by the inverse, which
  fires on `write()` too (unlike an `@api.onchange`) and does not create a
  recompute cycle.

## Relation to #3875
Alternative to #3875 (thanks @maq-adhoc), which fixed the same bug by
switching `_compute_product_uom_qty` to an `@api.onchange`. Keeping the
computed-field design works on programmatic writes (`write()`, imports, EDI,
API) — an `onchange` only fires in the form view — and preserves the base
`super()` packaging logic and the documented mixin pattern.

## Tests
- `test_packaging_qty_updates_quantities`: changing the packaging quantity
  propagates to both `product_uom_qty` and `secondary_uom_qty`, also on a
  programmatic `write()`.

### DIFF
--- a/sale_order_secondary_unit/models/sale_order.py
+++ b/sale_order_secondary_unit/models/sale_order.py
@@ -19,19 +19,13 @@ class SaleOrderLine(models.Model):
 
     product_uom_qty = fields.Float(copy=True)
 
-    @api.depends(
-        "display_type",
-        "product_id",
-        "product_packaging_qty",
+    @api.onchange(
         "secondary_uom_qty",
         "secondary_uom_id",
-        "product_uom_qty",
     )
-    def _compute_product_uom_qty(self):
-        res = super()._compute_product_uom_qty()
+    def _compute_target_field_qty(self):
         for line in self:
             line._compute_helper_target_field_qty()
-        return res
 
     @api.depends("product_id")
     def _compute_product_uom(self):
@@ -42,12 +36,22 @@ class SaleOrderLine(models.Model):
 
     @api.onchange("product_id")
     def _onchange_product_id_warning(self):
+        """
+        If default sales secondary unit set on product, put on secondary
+        quantity 1 for being the default quantity. We override this method,
+        that is the one that sets by default 1 on the other quantity with that
+        purpose.
+        """
         res = super()._onchange_product_id_warning()
         if self.product_id and not self.env.context.get("skip_secondary_uom_default"):
+            line_uom_qty = self.product_uom_qty
             self.secondary_uom_id = self.product_id.sale_secondary_uom_id
-            if self.product_uom_qty == 1.0:
-                self.secondary_uom_qty = 1.0
-                self._onchange_helper_product_uom_for_secondary()
+            if self.secondary_uom_id:
+                if line_uom_qty == 1.0:
+                    self.secondary_uom_qty = 1.0
+                    self._compute_helper_target_field_qty()
+                else:
+                    self.product_uom_qty = line_uom_qty
         return res
 
     @api.depends("secondary_uom_qty", "product_uom_qty", "price_unit")
--- a/sale_order_secondary_unit/models/sale_order.py
+++ b/sale_order_secondary_unit/models/sale_order.py
@@ -18,20 +18,19 @@ class SaleOrderLine(models.Model):
     )
 
     product_uom_qty = fields.Float(copy=True)
+    # Inverse (not a compute) to avoid a recursive dependency with the base
+    # packaging compute; it also covers write()/imports/EDI, unlike onchange.
+    secondary_uom_qty = fields.Float(inverse="_inverse_secondary_uom_qty")
 
-    @api.depends(
-        "display_type",
-        "product_id",
-        "product_packaging_qty",
-        "secondary_uom_qty",
-        "secondary_uom_id",
-        "product_uom_qty",
-    )
-    def _compute_product_uom_qty(self):
-        res = super()._compute_product_uom_qty()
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

sale_order_secondary_unit/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Sale Order Secondary Unit",
     "summary": "Sale product in a secondary unit",
-    "version": "18.0.1.0.2",
+    "version": "18.0.1.0.3",
```
MiquelRForgeFlow: No need, the mergebot will update the version automatically.

rousseldenis:
@mav-adhoc

mav-adhoc:
@MiquelRForgeFlow @rousseldenis Done!

rousseldenis (approved):

rousseldenis:
@sergio-teruel Nothing against this ?

sergio-teruel:
Hi.
But if you use an inverse method the changes does not shown before the record has been stored!

mav-adhoc:
@sergio-teruel Hi! makes sense that an inverse only runs on write, so the converted `product_uom_qty` won't refresh in the form until the record is saved. The reason for the inverse instead of te onchange is to keep the secondary  quantity conversion working on programmatic writes.

Maybe keeping both is a good alternative. The inverse for the programmatic writes and the onchange for the live changes.

My only doubt is that if by adding the onchange we could reintroduce the recompute round-trip in the form (packaging -> quantity -> secondary unit -> onchange -> quantity).

Please tell me what do you think about this and if you have another alternative. 

Thank you for your help!

maq-adhoc:
@mav-adhoc @sergio-teruel  Regarding the round-trip concern: from reading onchange() in addons/web/models/models.py (18.0), I think it can't loop forever — each field is added to a done set the first time it's processed and is never re-evaluated, so the dispatch loop looks hard-bounded by the number of fields in the view. I haven't tested this in a running form yet, so take it as a reading of the code, not a verified result.

What can happen is one extra, harmless pass: when product_packaging_qty changes, product_uom_qty/secondary_uom_qty get recomputed as a side effect, and since their values now differ from the initial snapshot, the onchange loop would fire secondary_uom_qty's onchange one round later even though the user never touched it — reconverting back into product_uom_qty. Since the conversion is the exact inverse of the compute, this should be a no-op value-wise (mod rounding), not a real cycle — but again, worth confirming in practice.

We could try keeping the inverse for programmatic writes and adding a plain onchange that reuses the same helper, for live feedback in the form:
```python

product_uom_qty = fields.Float(copy=True)
# inverse: covers write()/imports/EDI/API, which don't run onchange
secondary_uom_qty = fields.Float(inverse="_inverse_secon

def _inverse_secondary_uom_qty(self):
    for line in self:
        line._compute_helper_target_field_qty()

# onchange: same helper, so product_uom_qty also refresh
# form before the record is saved                                                                                                              @api.onchange("secondary_uom_qty")
def _onchange_secondary_uom_qty(self):                                                                                                             for line in self:
        line._compute_helper_target_field_qty()              
```                                                                                 
To make sure the extra pass doesn't introduce rounding drift, a Form-based test covering both edits in the same session would catch it without needing a browser:

```python
def test_packaging_then_secondary_onchange_no_regression
    packaging = self.env["product.packaging"].create(
        {"name": "box-10", "product_id": self.product.id
    )
    with Form(self.order) as order_form:
        with order_form.order_line.edit(0) as line:
            line.secondary_uom_id = self.secondary_unit
            line.product_packaging_id = packaging
            line.product_packaging_qty = 2
            self.assertEqual(line.product_uom_qty, 20.0)
            self.assertEqual(line.secondary_uom_qty, 40.
            line.secondary_uom_qty = 60.0
            self.assertEqual(line.product_uom_qty, 30.0)
```

mav-adhoc:
@sergio-teruel @maq-adhoc Thanks to both! I kept the inverse for programmatic writes (`write()`, imports, EDI, API) and added an `@api.onchange` on `secondary_uom_qty` that reuses the same helper, so `product_uom_qty` now refreshes in the form before saving too.

The onchange doesn't reintroduce the original `@api.depends` cycle (it's not a compute dependency), and the base helper
already calls `remove_to_compute` on `secondary_uom_qty` to avoid retriggering its recompute. 

Added a `Form`-based test that does both edits in the same session (set the packaging qty, then edit the secondary qty) and asserts the resulting quantities, which would catch any drift or lost live update.

I appreciate if you can review it again! Thanks!

mav-adhoc:
@sergio-teruel Hi Sergio! Can you review it again? Thank you!

PR #4436, opened 2026-07-01
[18.0][ADD] sale_order_secondary_unit_catalog

cgarcia-solvos:
Links secondary unit management in Sales Orders with product catalog


pedrobaeza:
Why this is not part of the main module, as it doesn't require any extra dependency?

cgarcia-solvos:
> Why this is not part of the main module, as it doesn't require any extra dependency?

The main module simply offers the possibility of working with secondary units on sale order lines when they are configured but the choice is left to the user. This addon, however, forces the use of the secondary unit from the catalog, so it is better to keep the functionality separate. This way, users who want to use secondary units occasionally can install just the base module without this addon.

pedrobaeza:
OK

cgarcia-solvos:
@AungKoKoLin1997 @uncannycs could you review? Thanks!!

PR #4440, opened 2026-07-02
[18.0][ADD] sale_order_secondary_unit_stock_catalog

cgarcia-solvos:
Also shows stock in secondary unit within Sales Orders Catalog


cgarcia-solvos:
@divad1196 @yankinmax could you review? Thanks!!

sale_order_secondary_unit_stock_catalog/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+                )
+                product.update(
+                    {
+                        "sale_secondary_unit_qty_available": ssu_qty_available,
+                        "has_sale_secondary_unit_qty_available": not float_is_zero(
+                            ssu_qty_available, precision_digits=dp
+                        ),
+                    }
+                )
+        (self - products_ssu).update(
```
divad1196, line 71: I would start by updating `self` completely:

```python
self.update({ ... })
if not if self.env.context.get("product_catalog_order_model", False) == "sale.order":
    return

...
```

Reasons:
- error-safe: values **MUST** be assigned in the compute. having it at the top with a default ensure it is assigned
- reduce nesting -> help with readability
- Removes the need to track `products_ssu`
- `self - products_ssu` can be heavier than just assigning a value for nothing.

sale_order_secondary_unit_stock_catalog/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+                lambda product: product.sale_secondary_uom_id
+                and product.sale_secondary_uom_id.factor > 0.0
+                and product.sale_secondary_uom_id.dependency_type != "independent"
+            )
+            for product in products_ssu:
+                ssu_qty_available = (
+                    product.qty_available / product.sale_secondary_uom_id.factor
+                )
+                product.update(
+                    {
```
divad1196, line 64: Would also move the dict computation logic in a `_get_sale_secondary_unit_qty_available_dict`. Odoo usually does that for readability and be modulable

divad1196 (approved):
I created a test product `Fabric` with UoM meter and secondary be centimeters.
- Added a secondary UoM
- Added a default secondary unit for sales
- Added `Second unit for inventory` as well or the Secondary Unit quantity would stay 0.

From a sale:
Image

Image

looks good.
I made a couple of remarks

PR #4419, opened 2026-06-23
[18.0][IMP] sale_automatic_workflow: Global setting to configure workflow copy behavior + auto_assign module

divad1196:
PR adds  global setting to change how workflows are copied on sales

PR #4445, opened 2026-07-03
[18.0][ADD] sale_automatic_workflow_auto_assign: Allow workflows to be automatically assigned based on predicats

divad1196:
Depends on:

- https://github.com/OCA/sale-workflow/pull/4419

It adds a new module that allows to auto-assign workflows on sales based on a domain matching


sale_automatic_workflow_auto_assign/models/res_config_settings.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
```
henrybackman: Are the years correct?

henrybackman (approved):

sale_automatic_workflow_auto_assign/models/res_config_settings.py
divad1196: The module already existed on our side in a private repo.
I ~~will~~ have updated them

santostelmo (approved):
PR LG.

some improvement (optional):
Add a test that covers multiple matching workflows and proves the lowest priority value wins. The README and implementation both state that matching workflows are ordered by priority, but the current tests only cover a single workflow match / no match cases. Since priority is a core part of the feature contract, this should be locked down with a regression test in `sale_automatic_workflow_auto_assign/tests/test_sale_workflow.py`

sale_automatic_workflow_auto_assign/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Camptocamp
```
simahawk, line 1: Add your name and the ones that contributed prior to this PR

sale_automatic_workflow_auto_assign/models/sale_workflow_process.py
```diff
@@ -0,0 +1,24 @@
...
+
+    auto_assign = fields.Boolean(
+        string="Auto assign",
+        help=(
+            "When enabled the workflow will be assigned at sales order creation, "
+            "if the assignation domain below applies."
+        ),
+    )
+    auto_assign_domain = fields.Text(string="Auto assign domain")
+    priority = fields.Integer(
```
simahawk: If the purpose is specific, I'd call it
```suggestion
    auto_assign_priority = fields.Integer(
```

simahawk (changes requested):
depending PR is getting merged

sale_automatic_workflow_auto_assign/models/sale_workflow_process.py
divad1196: @simahawk I did the change, but I also had to add a migration to it.

sale_automatic_workflow_auto_assign/readme/CONTRIBUTORS.md
divad1196: done

divad1196:
@santostelmo good idea. I added the test

@simahawk 
> depending PR is getting merged

I removed the dependency from test-requirements.

sale_automatic_workflow_auto_assign/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
```
rousseldenis:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
```

simahawk (approved):

PR #4449, opened 2026-07-07
[17.0][FIX] sale_fixed_discount: forward kwargs to super in _convert_to_tax_base_line_dict

MiquelRForgeFlow:
Lines without a fixed discount dropped the `**kwargs` (notably `handle_price_include=False`) that the down payment wizard passes, so fixed-amount down payments on price-included taxes produced an inconsistent tax breakdown.

Last commit is backport from 18.0

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -17,7 +17,8 @@ odoo_test_flavor: Both
 odoo_version: 17.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- sale_packaging_default
 repo_description: This project aim to deal with modules related to manage sale and
     their related workflow.
 repo_name: sale-workflow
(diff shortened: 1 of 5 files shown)


AaronHForgeFlow (approved):
code LGTM

JordiBForgeFlow (approved):

MiquelRForgeFlow:
@rousseldenis

rousseldenis:
@MiquelRForgeFlow Excluding sale_packaging_default in tests flows is hiding the problem IMHO.

But, ok, don't block current PR's with that

MiquelRForgeFlow:
> Excluding sale_packaging_default in tests flows is hiding the problem IMHO.

I know, but in v18 was excluded too. Do you want me to fix that module?

rousseldenis:
> > Excluding sale_packaging_default in tests flows is hiding the problem IMHO.
> 
> I know, but in v18 was excluded too. Do you want me to fix that module?

No, if you don't use it. I'm just wondering why the code in it is so complex

Issue #4450, opened 2026-07-07
[17.0] sale_packaging_default: Main branch is broken


rousseldenis:
@yajo Could you look at 17.0 main branch as tests are failing on sale_packaging_default module?

I'm quite concerned by the complexity of the code and hacks done there.

rousseldenis:
FYI, I tried some installs with last merged modules and iterating with Odoo commits history, it seems that all sale-workflow from branch 17.0 install is the problem (I didn't try to remove one by one - too much modules 😄 )

yajo:
Sorry we don't support v17 anymore.

You can mark the modules as rebel to avoid those kind of problems.

PR #4451, opened 2026-07-07
[17.0][FIX] sale_order_secondary_unit: avoid recursive qty recompute cycle (backport)

MiquelRForgeFlow:
The _compute_product_uom_qty override depended on product_uom_qty and secondary_uom_qty, creating a recursive cycle with the base packaging compute: changing product_packaging_qty was silently discarded (e.g. combined with sale_packaging_default). Move the secondary->product sync to an inverse method on secondary_uom_qty so it only runs on write.

Also includes backport of OCA/sale-workflow#4435 to 17.0.


rousseldenis:
@MiquelRForgeFlow I've opened: https://github.com/OCA/sale-workflow/issues/4450

PR #4452, opened 2026-07-08
[18.0][FIX] sale_order_invoicing_finished_task: Incorrect assignation in tests

pedrobaeza:
`self.manager` is a `res.users` record, so assigning it to a many2one field for `res.partner` is not correct.

This has worked by chance if the ID of the user exists in the partner table.

@Tecnativa


pedrobaeza:
Fast-tracking this as obvious patch.

/ocabot merge patch

PR #4456, opened 2026-07-09
[19.0][MIG] sale_semaphore : Migration to 19.0.

uncannycs:
[19.0][MIG] sale_semaphore : Migration to 19.0.


sale_semaphore/tests/test_sale_semaphore.py
```diff
@@ -219,11 +221,70 @@ def test_price_below_pricelist_blocks(self):
...
                 "price_unit": 70.0,
             }
         )
         self.assertTrue(line.price_below_semaphore)
         self.assertTrue(line._is_price_below_pricelist())
         with self.assertRaises(UserError):
             order.with_user(self.salesperson).action_confirm()
+
+    def test_additional_coverage(self):
+        pre_init_hook(self.env)
```
rousseldenis: This is a non sense as that hook has already been called during install.

sale_semaphore/tests/test_sale_semaphore.py
```diff
@@ -219,11 +221,70 @@ def test_price_below_pricelist_blocks(self):
...
+                "product_uom_id": self.product.uom_id.id,
                 "price_unit": 70.0,
             }
         )
         self.assertTrue(line.price_below_semaphore)
         self.assertTrue(line._is_price_below_pricelist())
         with self.assertRaises(UserError):
             order.with_user(self.salesperson).action_confirm()
+
+    def test_additional_coverage(self):
```
rousseldenis: Prefer using a name that means something (the aim of this test)

rousseldenis (changes requested):
LGTM.

At least change the test name

sale_semaphore/tests/test_sale_semaphore.py
uncannycs: @rousseldenis - We completely understand that the hook is already executed during the module's installation. However, We explicitly called pre_init_hook(self.env) here in the test case for the purpose of Code Coverage.

Since the standard test suite runner sometimes misses the initial installation hooks in its coverage report, manually triggering it here ensures that the hook's logic is fully executed and covered by the coverage metrics. If there is a more preferred OCA standard way to test and cover pre_init_hook without calling it directly here, please let me know and We will gladly update it!

Also, We have addressed your other point and renamed the test method to test_semaphore_edge_cases_and_reporting to better reflect its aim.

rousseldenis: @uncannycs IMHO, we should not be jesuistics on coverage. Some corner cases will never be tested and that's normal.

rousseldenis (approved):

PR #4459, opened 2026-07-10
[16.0][FIX] sale_order_line_cancel: Give back canceled qty when the ordered qty is raised again

mt-software-de:
cc @vuwnevska @rousseldenis @jbaudoux @lmignon

### DIFF
--- a/sale_order_line_cancel/models/sale_order_line.py
+++ b/sale_order_line_cancel/models/sale_order_line.py
@@ -44,6 +44,31 @@ class SaleOrderLine(models.Model):
             qty_remaining = max(0, qty_to_deliver - line.product_qty_canceled)
             line.product_qty_remains_to_deliver = qty_remaining
 
+    def write(self, values):
+        # When the ordered qty is raised again after some qty was canceled,
+        # give the canceled qty back so the line becomes deliverable again.
+        # This must happen before super() so the qty is already restored when
+        # the procurement (and any resulting move merge/cancel) runs.
+        if "product_uom_qty" in values and "product_qty_canceled" not in values:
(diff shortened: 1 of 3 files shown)


sale_order_line_cancel/models/sale_order_line.py
```diff
@@ -44,6 +44,31 @@ def _compute_product_qty_remains_to_deliver(self):
             qty_remaining = max(0, qty_to_deliver - line.product_qty_canceled)
             line.product_qty_remains_to_deliver = qty_remaining
 
+    def write(self, values):
+        # When the ordered qty is raised again after some qty was canceled,
+        # give the canceled qty back so the line becomes deliverable again.
+        # This must happen before super() so the qty is already restored when
+        # the procurement (and any resulting move merge/cancel) runs.
+        if "product_uom_qty" in values and "product_qty_canceled" not in values:
```
vuwnevska, line 52: should this logic run only if the order is in the state 'sale'?

mt-software-de: Normally this will only run if in state `sale`, because by default it is not allowed to write product_uom_qty in state `done` and in state `draft` the `product_qty_canceled` should be zero. 
I would rather not include the check against state, because it is then depending for state `done` if the super call raises an error. And in state draft it should already be ignored because of the 0 `product_qty_canceled`

lmignon (approved):
@mt-software-de Your fixes makes sense to me.

jbaudoux (approved):
Indeed. Looks good.

mt-software-de:
@OCA/crm-sales-marketing-maintainers can you merge this one?

rousseldenis (approved):

PR #4460, opened 2026-07-10
[18.0][FWP][FIX] sale_order_line_cancel: give back canceled qty when the ordered qty is raised again

mt-software-de:
FWP of https://github.com/OCA/sale-workflow/pull/4459


jbaudoux (approved):
Still draft?
LG. Also approved on v16.0

vuwnevska (approved):

PR #4426, opened 2026-06-25
[18.0][FIX] sale_automatic_workflow: use _get_eval_domain() to support datetime in filter

nobuQuartile:
@qrtl
QT6730

PR #4461, opened 2026-07-13
[19.0][IMP] sale_automatic_workflow: support dynamic dates in filter domains

AungKoKoLin1997:
Forward port of https://github.com/OCA/sale-workflow/pull/4426/change but I use different approach since `_get_eval_domain` in v19.0 doesn't use datetime and context_today anymore.

@qrtl QT5088


cgarcia-solvos (approved):
Code review. LGTM!

rousseldenis (approved):

PR #4463, opened 2026-07-13
[FIX] sale_quotation_number: Name change doesn't fail on dynamic prefixes

AlienAtSystem:
…ynamic prefixesDescription of the issue/feature this PR addresses:
sale_quotation_number does not work when the prefix for the sale.quotation sequence includes a dynamic placeholder. sale_quotation_number also fails to account for the sequence suffix.

Current behavior before PR:
When the prefix of the sale.quotation sequence includes a dynamic placeholder, the detection during `action_confirm()` that sale order name is not custom fails. This is because it just compares the first part of the name with the value of the `prefix` field of the sequence ([Link](https://github.com/OCA/sale-workflow/blob/6e079fbec359ba033476ff3c885ab08efa5e9a70/sale_quotation_number/models/sale_order.py#L51)). For a dynamic placeholder, this will never be true, because both the length and the content of the formatting string are of course not equivalent to the formatted string output of `next_by_code()`.
As a side consequence, the detection can also result in false positives when prefixes overlap, e.g. if the sale.quotation prefix were "S" and the sale.order prefix were "SO" - a confirmed, then reset sale order would get a new name.
There is absolutely no check against the suffix, including no check that the sale order name wasn't edited by appending something behind the sequence-generated part.

Desired behavior after PR is merged:
The detection for whether a sale order name fits the sale.quotation scheme compares the whole order name, not just the prefix part. It accounts for dynamic placeholders by generating a fitting regex string that matches the possible outputs very closely.
The tests have been updated and extended accordingly.

### DIFF
--- a/sale_quotation_number/__manifest__.py
+++ b/sale_quotation_number/__manifest__.py
@@ -16,6 +16,6 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["sale_management"],
+    "depends": ["sale_management", "sequence_regex"],
     "data": ["data/ir_sequence_data.xml", "views/sales_config.xml"],
 }
--- a/sale_quotation_number/models/sale_order.py
+++ b/sale_quotation_number/models/sale_order.py
@@ -44,13 +44,18 @@ class SaleOrder(models.Model):
         return super().copy(default)
 
     def action_confirm(self):
-        sequence = self.env["ir.sequence"].search(
-            [("code", "=", "sale.quotation")], limit=1
-        )
         for order in self:
-            if sequence and self.name[: len(sequence.prefix)] != sequence.prefix:
-                continue
             if order.state not in ("draft", "sent") or order.company_id.keep_name_so:
+                continue
+            sequence = self.env["ir.sequence"].search(
+                [
+                    ("code", "=", "sale.quotation"),
+                    ("company_id", "in", [order.company_id.id, False]),
+                ],
+                order="company_id",
+                limit=1,
+            )
+            if sequence and not sequence.name_fits_sequence(order.name):
                 continue
             if order.origin and order.origin != "":
                 quo = order.origin + ", " + order.name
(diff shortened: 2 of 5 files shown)


sale_quotation_number/tests/test_sequence.py
```diff
@@ -0,0 +1,43 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSequence(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis: This can be avoided using `odoo.addons.base.tests.common.BaseCommon`

sale_quotation_number/models/ir_sequence.py
```diff
@@ -0,0 +1,65 @@
...
+# © 2017 valentin vinagre  <valentin.vinagre@qubiq.es> (QubiQ)
+# © 2020 Manuel Regidor  <manuel.regidor@sygel.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import re
+
+from odoo import models
+
+
+class IrSequence(models.Model):
```
rousseldenis: Maybe all of this can be done in another module ?

AlienAtSystem: It could, yes. I've decided to move it to a new module, which means a new Pull Request over in another repository: https://github.com/OCA/server-tools/pull/3672

sale_quotation_number/tests/test_sequence.py
AlienAtSystem: Noted. It's actually irrelevant here, I copied it over from test_sale_order where it's been like this for a while - I fixed it now.

rousseldenis:
@AlienAtSystem FYI, oca_dependencies is no more used. Follow this in order to run tests and runboat: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

PR #4468, opened 2026-07-14
[18.0][ADD] sale_order_line_final_price: add new module

eduezerouali-tecnativa:
cc @Tecnativa TT63696
@pedrobaeza @carlos-lopez-tecnativa


sale_order_line_final_price/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,10 @@
+This module adds a *Final Price* field to sale order lines, holding the unit
```
rousseldenis, line 1: @eduezerouali-tecnativa Could you add context to help why and when this should be used ? Thanks

pedrobaeza:
Please simplify the name to `sale_order_line_final_price`, and add CONTEXT.md to explain why this is useful, for always maintaining the list price, and only touch discounts in an indirect mode.

Check also the CI.

sale_order_line_final_price/readme/DESCRIPTION.md
eduezerouali-tecnativa: I have added CONTEXT.md to explain why this is useful.

eduezerouali-tecnativa:
Name changed, CONTEXT.md added and took care of CI.

rousseldenis:
Ok, I understand the use case now. It seems to me there is already a module for that or at least a PR. I will check

sale_order_line_final_price/views/sale_order_views.xml
```diff
@@ -0,0 +1,69 @@
...
+            <xpath
+                expr="//field[@name='order_line']/list/field[@name='discount']"
+                position="after"
+            >
+                <field name="price_edit_allowed" column_invisible="1" />
+                <field
+                    name="price_final"
+                    readonly="qty_invoiced &gt; 0"
+                    optional="show"
+                />
```
carlos-lopez-tecnativa, line 39: I think this feature should only be visible when the discount feature is enabled. Otherwise, users will see two similar fields, even though discounts are not displayed.

What do you think about adding the `sale.group_discount_per_so_line` group?

carlos-lopez-tecnativa (commented):
Functional review: LGTM. Just one technical comment:

Please update the PR title to reflect the new module name.

juancarlosonate-tecnativa (approved):
LGTM, decimal precision behavior reviewed and working as documented, just rename the .pot file to the new module name and update the PR title accordingly.

sale_order_line_final_price/views/sale_order_views.xml
eduezerouali-tecnativa: True. Done!

eduezerouali-tecnativa:
Changes done

carlos-lopez-tecnativa (approved):
LGTM

sale_order_line_final_price/models/sale_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+        readonly=False,
+        precompute=True,
+        min_display_digits="Product Price",
+        help="Unit price once the discount is applied. Set here the price agreed "
+        "with the customer and the discount will be computed accordingly, keeping "
+        "the unit price given by the pricelist.",
+    )
+    price_edit_allowed = fields.Boolean(
+        compute="_compute_price_edit_allowed",
+        depends_context=("uid",),
```
pedrobaeza: Why putting this here instead on the compute method?

sale_order_line_final_price/models/sale_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+            if price_final is not None:
+                line.price_final = price_final
+        return lines
+
+    def _apply_price_final(self):
+        """Move the discount so that the line lands on the requested final price."""
+        self.ensure_one()
+        if self.display_type:
+            return
+        self.discount = self._get_price_final_discount()
```
pedrobaeza: In this case, the code this way is better readable IMO and more compact:

```suggestion
        if not self.display_type:
            self.discount = self._get_price_final_discount()
```

sale_order_line_final_price/models/sale_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+        The inverse method can't do it while creating: the amounts are precomputed
+        and protected against recomputation until the creation is over, so they
+        would keep the figures of a line without discount.
+        """
+        vals_list = [dict(vals) for vals in vals_list]
+        prices_final = [vals.pop("price_final", None) for vals in vals_list]
+        lines = super().create(vals_list)
+        for line, price_final in zip(lines, prices_final, strict=False):
+            if price_final is not None:
+                line.price_final = price_final
```
pedrobaeza: This drains the performance, as you need to do create + write. Please find another way.

sale_order_line_final_price/models/sale_order_line.py
eduezerouali-tecnativa: Changed!

sale_order_line_final_price/models/sale_order_line.py
eduezerouali-tecnativa: Done!

sale_order_line_final_price/models/sale_order_line.py
eduezerouali-tecnativa: Let me know if you like the new approach.

eduezerouali-tecnativa:
We also added the case when `price_unit` is 0. As no discount will lift a 0 price, now if `price_unit` is 0 editable from `price_final` in this particular case. Also added to the test.

pedrobaeza (approved):
/ocabot merge nobump

pedrobaeza (changes requested):
Oh, wait, this is still doing 2 SQL operations (create and write), no matter if you do it directly or through `modify`. Please include everything needed in the `vals` dictionary for doing in one shot.

eduezerouali-tecnativa:
With new approach we should avoid doing 2 SQL operations.

sale_order_line_final_price/models/sale_order_line.py
```diff
@@ -0,0 +1,123 @@
...
+        for line in self:
+            line._apply_price_final()
+        # What the line is really worth is what the unit price and the discount say,
+        # and a rounded discount doesn't always land on the requested figure. The
+        # write shields the field being inversed from its own recomputation, so it
+        # has to be asked for, or the line would keep an unreachable final price.
+        self.env.add_to_compute(self._fields["price_final"], self)
+
+    @api.onchange("price_final")
+    def _onchange_price_final(self):
```
pedrobaeza: You should avoid `onchange` at all cost, because they only works on screen, but not by code. Isn't there any other option? Maybe the other formula didn't work because you didn't put `force_save="1"` on the view?

eduezerouali-tecnativa: I know, but i have tried with `force_save` and none reflects the change except `onchange`. I can remove it  and relay on the save, but i think it will be a bit odd for the user. Maybe I am missing something.

eduezerouali-tecnativa: ping @pedrobaeza ??

pedrobaeza (approved):
/ocabot merge nobump

PR #4476, opened 2026-07-17
[19.0][IMP] sale_order_product_recommendation: use Product Price decimal precision for price_unit

joaoc-exo:
Forward-port of #3898 (merged in 17.0) to 19.0.

`price_unit` on the recommendation wizard line was a `Monetary` field, so the last sale price / pricelist price shown in the wizard was rounded to the currency's 2 decimal places, losing precision when products are priced with more decimals ("Product Price" decimal precision). Use a `Float` with `digits="Product Price"` instead, as `sale.order.line.price_unit` does.


yajo (approved):
/ocabot merge minor

PR #4477, opened 2026-07-17
[FIX] sale_invoice_policy: field invoice_policy in sale_order is now required but could be false

sergiocorato:
In a migration from 17.0, the field invoice_policy could be false, but in 18.0 is required and this block the constraint.
This PR add a post migration script to assign the value to the field.


rousseldenis (approved):
Indeed, thanks !

rousseldenis:
@sergiocorato Could you just squash your commits ?

sergiocorato:
> @sergiocorato Could you just squash your commits ?

Done, thanks!

BhaveshHeliconia (approved):
LGTM!

PR #4480, opened 2026-07-20
[18.0][FIX] sale_order_lot_selection: Wrong state comparison when reassigning moves

victoralmau:
Backport from 19.0: https://github.com/OCA/sale-workflow/pull/4385

Wrong state comparison when reassigning moves

The tuple ('confirmed', 'partially_available') was written as a single string, so moves were never reassigned after changing the lot on the sale order line.

@Tecnativa


pedrobaeza (approved):
Clear bugfix, so fast-tracking it:

/ocabot merge patch

PR #4483, opened 2026-07-21
[16.0][FIX] sale_blanket_order: display_type = false

CristianoMafraJunior:
**Reason**: The "Blanket Order Lines" screen was listing all lines from the blanket order, including section and note lines (used only to organize/comment the order, with no product attached). That's why blank rows (no product) showed up in the list, as seen in the screenshot.

**Change**: I added a filter (domain) to the action that opens this screen, excluding section/note lines (display_type), so it only shows actual product lines — which is the correct and expected behavior for this list.

cc @kaynnan @marcelsavegnago @WesleyOliveira98


kaynnan (approved):

marcelsavegnago (approved):
LGTM

rvalyi (approved):

PR #4488, opened 2026-07-22
[19.0] [MIG] sale_team_payment_term : Migration to 19.0

kalpeshgajera-uncannycs:
[19.0] [MIG] sale_team_payment_term : Migration to 19.0


sale_team_payment_term/models/sale_order.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    payment_term_id = fields.Many2one(
+        compute="_compute_payment_term_id",
+        inverse="_inverse_payment_term_id",
+        readonly=False,
```
IriaAlonso: `readonly` attribute is now defined on the views
```suggestion
```

IriaAlonso (commented):
LGTM code review and test on runboat
only see minor comment

bitigloo-matthias:
Thanks for the migration. Could you please confirm the intended recomputation semantics of payment_term_id?

Odoo 19 core uses @api.depends("partner_id"), while this PR uses @api.depends("partner_id.property_payment_term_id", "team_id.team_payment_term_id").

Since payment_term_id is stored, wouldn't changing the payment term on a partner or sales team potentially recompute and overwrite the payment term on existing sale orders, including confirmed orders or orders where the term was manually overridden?

Would @api.depends("partner_id", "team_id") be preferable to preserve the payment term as a snapshot on existing orders?

Or is the current recomputation behavior intentional?

_(I am NOT a developer - found out while testing and with AI ;-))_

PR #4489, opened 2026-07-22
[19.0] [MIG] web_widget_product_label section_and_note_full_label_sale : Migration to 19.0

kalpeshgajera-uncannycs:
[19.0] [MIG] web_widget_product_label section_and_note_full_label_sale : Migration to 19.0


web_widget_product_label_section_and_note_full_label_sale/static/src/sale_product_field/sale_product_field.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates>
+    <t t-inherit="sale.SaleProductField" t-inherit-mode="extension">
+        <xpath
+            expr="//t[@t-if='columnIsProductAndLabel.value and label']/textarea"
```
IriaAlonso: Testing on runboat, it doesnt find this `xpath`

IriaAlonso (changes requested):

web_widget_product_label_section_and_note_full_label_sale/static/src/sale_product_field/sale_product_field.xml
kalpeshgajera-uncannycs: Thanks for pointing it out. We've resolved the issue, and the XPath is now working correctly.

IriaAlonso:
Tested in runboat
Adding a product with variants, you can add a customization:
imagen
So, in the lines it apears of this form
imagen
If the value is empty, the label could be invisible and also fix the `Custom:` test

in addition, if you modify the text the name of the product appears again
imagen
can you review it @kalpeshgajera-uncannycs ?

PR #4492, opened 2026-07-23
[18.0][FIX] sale_elaboration: All moves has same sale_id

CarlosRoca13:
Before this changes, all moves shown in the screen has the same sale linked.

To test it you can go to Inventory > Reports > Elaborations Report and see that all lines has the same sale linked.

After the changes, they has the correct sale_id.

cc @Tecnativa TT63801


sergio-teruel (approved):

carlosdauden (approved):

pedrobaeza (commented):
Is this preserving the sale line when there are multiple chained moves?

CarlosRoca13:
I have tested it with a 3 step delivery, and it is linked in all moves

PR #4503, opened 2026-07-28
[19.0][ADD] sale_order_show_full_address: show invoice and delivery addresses

LauraVForgeFlow:
Shows the complete invoice address and delivery address on sales orders, the
same way Odoo already does for the customer, using the show_address context
key. View only, no new fields.

Both fields are only visible with the account.group_delivery_invoice_address
group.


AaronHForgeFlow (commented):
code LG. The module name could be changed. In general plurals are not desirable as module names. Also the name could be more explicative, because the address does really show but not the full address. Maybe sale_show_full_invoice_picking_address is too much, I open to suggestions 😬

RicardCForgeFlow:
Works as expected:
image
I agree with @AaronHForgeFlow , maybe just _sale_show_full_address_?

IriaAlonso:
LGTM code review ans test on runboat
And my suggestion for the name, `sale_order_show_full_address` what do you think? :)

LauraVForgeFlow:
Renamed to sale_order_show_full_address as @IriaAlonso suggested. Thanks for the reviews!

rousseldenis (approved):
LGTM

RicardCForgeFlow (approved):

IriaAlonso (approved):

rousseldenis:
@sbidoul

sbidoul:
name reserved

PR #4506, opened 2026-07-29
[19.0][ADD] sale_commitment_date_required_on_confirm

LauraVForgeFlow:
This module makes the delivery date (`commitment_date`) required to confirm a sales order, and only then: the quotation can be created, saved and edited with an empty delivery date, and the check is done in `_confirmation_error_message()`. Since the date becomes a condition to confirm the order, the field is moved from the *Other Info* tab to the order header.

It is enabled per company with the *Delivery Date Required* setting, disabled by default.


sale_commitment_date_required_on_confirm/views/sale_order_views.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 ForgeFlow S.L. (https://www.forgeflow.com)
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0) -->
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form.commitment.date.header</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <!-- Applied after sale_stock and delivery, which alter this block. -->
```
MiquelRForgeFlow: Do you mention `sale_stock` and `delivery`. Should you add these modules in the manifest depends?

MiquelRForgeFlow (commented):
I think you should add `[19.0]` in the PR title.

sale_commitment_date_required_on_confirm/tests/test_sale_commitment_date_required_on_confirm.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertEqual(self.order.state, "sale")
+
+    def test_confirm_without_commitment_date_when_not_required(self):
+        self.company.sale_commitment_date_required = False
+        self._add_line()
+        self.order.action_confirm()
+        self.assertEqual(self.order.state, "sale")
+
+    def test_commitment_date_is_in_the_order_header(self):
+        arch = etree.fromstring(
```
AaronHForgeFlow: I think testing the UI with unit tests is not needed. Better to reserve that for functional tests.

AaronHForgeFlow (commented):
Tried to test in runboat but I was able to confirm a SO without delivery date: http://oca-sale-workflow-19-0-pr4506-725f9b502f16.runboat.odoo-community.org/odoo/sales/27

Perhaps another OCA module is influencing the test?

sale_commitment_date_required_on_confirm/views/sale_order_views.xml
LauraVForgeFlow: Thanks for the review! 
I added "[19.0]" to the PR title, and about `sale_stock` and `delivery`, I shouldn't have mentioned them. I was trying to explain the view priority, but now I have reworded it so it does not name them.

sale_commitment_date_required_on_confirm/tests/test_sale_commitment_date_required_on_confirm.py
LauraVForgeFlow: To try it, the "Delivery Date Required" setting has to be enabled, in Sales > Configuration > Settings > Quotations & Orders. It is disabled by default, so maybe it was not enabled in the runboat build.

About the unit tests, I removed them.

Thank you!

AaronHForgeFlow (approved):
Functional review + soft code review

sale_commitment_date_required_on_confirm/tests/test_sale_commitment_date_required_on_confirm.py
AaronHForgeFlow: It seems to be working now :)

PR #4510, opened 2026-07-30
[14.0][FIX] sale_fixed_discount: Wrong tax amount displayed on report

hitrosol:
Fixes wrong tax amount shown on the sale order/quotation report: amount_by_group ignored the fixed discount and computed tax on the undiscounted price. Overrides sale.order._amount_by_group to use the discounted price, plus a covering test. Ref #3983.


rousseldenis:
@hitrosol The better would have been to cherry-pick the commit to preserve original author. Thanks

hitrosol:
Thanks for your comment @rousseldenis, hope this is ok.

nurinayati (changes requested):
Funtional review
1. Total without Discount, conversion to company currency & Discount subtotal are not showing correcly

Image

Expected result and how it works with % discount

Image

2. On the sales order , invoice print out should show the currency for fixed discount

Image

hitrosol:
Thanks for the functional review @nurinayati. Both points look out of scope for this PR — details below.

Total Without Discount / Discount subtotal / company currency total

Those three amounts are not provided by sale_fixed_discount:

- price_total_no_discount and discount_total come from sale_discount_display_amount
- amount_total_curr comes from sale_company_currency

I would suggest to create glue modules to handle this case.

In order to test for this specific PR, You can try on runboat using the baseonly database and install only related to this module

nurinayati (commented):
The tax amount result on customer invoice is incorrect if some lines not applying tax

Image

nurinayati:
> Thanks for the functional review @nurinayati. Both points look out of scope for this PR — details below.
> 
> Total Without Discount / Discount subtotal / company currency total
> 
> Those three amounts are not provided by sale_fixed_discount:
> 
> * price_total_no_discount and discount_total come from sale_discount_display_amount
> * amount_total_curr comes from sale_company_currency
> 
> I would suggest to create glue modules to handle this case.
> 
> In order to test for this specific PR, You can try on runboat using the baseonly database and install only related to this module

one cosmetic suggestion that would be great if its updated is to show currency on the print out. its ok and nice to have. 

@hitrosol please check my latest comment, the tax amount on the customer invoice is incorrect

hitrosol:
The inconsistent tax amount is caused by a bug in account_invoice_fixed_amount, not by the changes in this PR. I'd suggest merging this one first, then opening a separate follow-up PR against account_invoice_fixed_amount to fix the root cause.

On the currency display: this implementation keeps Discount (fixed amount) and Discount (%) as two separate columns, which differs from the newest version where both values share one column. In that layout the currency symbol is needed to disambiguate the two, but here the column itself already makes the unit unambiguous, so I'd prefer to leave the currency out and keep the line view compact.

nurinayati (approved):
Functional review done, LGTM.

I open an issue to fix problem the wrong tax amount on customer invoice. ;) https://github.com/OCA/account-invoicing/issues/2428

PR #4513, opened 2026-07-30
[16.0][FIX] sale_blanket_order: remaining_qty should not force UoM rounding

kaynnan:
@Escodoo 

[HT02034]

cc @marcelsavegnago @douglascstd @WesleyOliveira98 @CristianoMafraJunior 

`remaining_qty` converts `remaining_uom_qty` into the product's reference UoM (`product_id.uom_id`), but forces rounding to that UoM's own `rounding` attribute (0.01 by default) via `_compute_quantity(..., round=True)`. This rounding is completely independent of the "Product Unit of Measure" decimal precision that already governs `remaining_uom_qty` and every other quantity field on the line.

In practice, this silently truncates the remaining balance whenever the blanket order line's own UoM tracks finer fractions than the product's reference UoM allows — e.g. `original_uom_qty = 1.00`, an order for `0.633` is placed, the expected remainder is `0.367`, but `remaining_qty` ends up rounded to `0.36`/`0.37` instead, while `remaining_uom_qty` correctly stays at the finer precision. This breaks any workflow that needs to sequence further partial orders against the same blanket order line down to an exact remaining fraction (e.g. services billed by progressive measurement).

## What changed

- `sale_blanket_order/models/blanket_orders.py`: `_compute_quantities()` now converts `remaining_qty` with `round=False`, so it is not forced through the reference UoM's own rounding on top of whatever precision `remaining_uom_qty` already has.
- `sale_blanket_order/tests/test_blanket_orders.py`: added `test_05b_remaining_qty_precision_survives_uom_rounding`, which deliberately coarsens the product's reference UoM rounding and asserts `remaining_qty` stays equal to `remaining_uom_qty` instead of being independently rounded away.

## Test plan
- [x] New test fails against the pre-fix code (`0.5 != 0.37`) and passes with the fix.
- [x] Full `sale_blanket_order` test suite: 10/10 tests pass.
- [x] pre-commit (black, isort, flake8, pylint_odoo) clean.

### DIFF
--- a/sale_blanket_order/models/blanket_orders.py
+++ b/sale_blanket_order/models/blanket_orders.py
@@ -680,7 +680,7 @@ class BlanketOrderLine(models.Model):
             )
             line.remaining_uom_qty = line.original_uom_qty - line.ordered_uom_qty
             line.remaining_qty = line.product_uom._compute_quantity(
-                line.remaining_uom_qty, line.product_id.uom_id
+                line.remaining_uom_qty, line.product_id.uom_id, round=False
             )
 
     def _validate(self):
--- a/sale_blanket_order/tests/test_blanket_orders.py
+++ b/sale_blanket_order/tests/test_blanket_orders.py
@@ -350,6 +350,75 @@ class TestSaleBlanketOrders(common.TransactionCase):
         self.assertEqual(blanket_order.line_ids[0].remaining_qty, 12.0)
         self.assertEqual(sale_order.order_line[0].price_unit, 20.0)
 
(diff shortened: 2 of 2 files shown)


douglascstd (approved):
[Funcional Review] OK

**using 2 decimal precision**
Image

**set 4 decimal precision**
image

kaynnan:
ping @pedrobaeza

WesleyOliveira98 (approved):
LGTM

CristianoMafraJunior (approved):

PR #4522, opened 2026-08-10
[19.0][MIG] sale_pricelist_from_commitment_date

weinni2000:
simple migration


tonygalmiche (changes requested):
Tested on a local Odoo 19.0 instance:

- Pricelist rule correctly applied based on commitment_date instead of order date when set
- Delivery date header/mandatory settings work as configured
- Bug found: adding a line via the product configurator fails once commitment_date is set (`Invalid props for component 'ProductConfiguratorDialog': unknown key 'date'`) — caused by static/src/js/sale_product_field.esm.js setting both `props.soDate` and `props.date`, but Odoo 19's ProductConfiguratorDialog only accepts `soDate`. Fix: remove the `props.date = date;` line. Confirmed the fix resolves the issue locally.
- Behavioral note: changing commitment_date silently recomputes price_unit on all lines that weren't manually overridden (manually-set prices are correctly preserved), unlike changing the pricelist itself which requires an explicit "Update Prices" action on existing lines. Not destructive, but worth flagging as a UX inconsistency.

BhaveshHeliconia (approved):
LGTM!

BhaveshHeliconia:
@tonygalmiche The issue you reported has been resolved, could you please review again when you have a moment?

BhaveshHeliconia:
@weinni2000 could you please squash the administrative and bot commits?

PR #4523, opened 2026-08-11
[19.0][ADD] sale_order_line_smartbutton

imlopes:


sale_order_line_smartbutton/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+        Reuses the standard sale order line list/form views so users can
+        search/filter lines and edit prices, quantities, uom, etc. subject
+        to the same restrictions as the sale order form (see
+        ``sale.order.line`` readonly fields and its ``write`` override).
+        """
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "sale_order_line_smartbutton.sale_order_line_action"
+        )
+        action["domain"] = [("order_id", "=", self.id)]
```
ivantodorovich: Isn't this domain redundant with `search_default_order_id`?


```suggestion
```

Feels weird to remove the filter and still see the order lines instead of all lines:

https://github.com/user-attachments/assets/879b2852-e4ce-4c6e-b850-37723cfe4883

ivantodorovich (approved):
Thanks!

Ricardoalso (approved):
Technically LG

sale_order_line_smartbutton/models/sale_order.py
imlopes: You're right but to avoid confusion I think it could be better to remove the default `search` instead. What do you think ?

ivantodorovich: Either way seems good 👍🏻

sale_order_line_smartbutton/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
```
Ricardoalso:
```suggestion
        domain = Domain(action["domain"])
        domain += Domain("order_id", "=", self.id)
        action["domain"] = domain
```

The action sale_order_line_action defines domain: [('display_type', '=', False)] to exclude sections/notes, but that filter is discarded if not anded here

Runboat result
image

BhaveshHeliconia (approved):
LGTM!

ivantodorovich:
@imlopes the OCB ci is red, unrelated issue it seems.
Can you rebase? might be solved

PR #4531, opened 2026-08-18
[17.0][IMP] sale_order_line_sequence: Reduce column size through label + optional - #3856

victoralmau:
FWP from 16.0: https://github.com/OCA/sale-workflow/pull/3856

Extra changes:
- Change string to sale order pdf
- Change string to invoice form view
- Change string to invoice pdf

@Tecnativa

### DIFF
--- a/sale_order_line_sequence/views/account_move_view.xml
+++ b/sale_order_line_sequence/views/account_move_view.xml
@@ -13,6 +13,8 @@
                 <field
                     name="related_so_sequence"
                     column_invisible="parent.move_type != 'out_invoice'"
+                    string="#"
+                    optional="show"
                 />
             </xpath>
         </field>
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

sale_order_line_sequence/views/account_move_view.xml
```diff
@@ -13,6 +13,7 @@
                 <field
                     name="related_so_sequence"
                     column_invisible="parent.move_type != 'out_invoice'"
+                    string="#"
```
carlos-lopez-tecnativa, line 16: This field is optional in the sale order, but not in the account.move. I think it would be great to keep it optional here as well.

carlos-lopez-tecnativa (approved):
Just a non-blocking comment: in v16, I didn't take this into account. If you can update it, that would be great; otherwise, LGTM.

PR #4309, opened 2026-04-23
[18.0][IMP] sale_commercial_partner: add related field commercial_partner_invoice_id

alexis-via:
New related stored field commercial_partner_invoice_id (related="partner_invoice_id.commercial_partner_id"). Useful for domains that depend on the parent of the invoice partner.

PR #4534, opened 2026-08-18
[19.0][FWPORT] sale_commercial_partner: add related field commercial_partner_invoice_id

alexis-via:
This PR is a forward port from 18 -> 19 of https://github.com/OCA/sale-workflow/pull/4309


njeudy (approved):
LGTM, an work as needed

gdgellatly:
pls, @alexis-via if you are adding fields that require a db upgrade it is not a minor. Minor is for non breaking

PR #4537, opened 2026-08-20
[19.0][ADD] sale_order_priority_partner

PaoloYam:
Add sale_priority selection field on res.partner
use this field as default at sale.order creation


sale_order_priority_partner/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
Ricardoalso:
```suggestion
```

sale_order_priority_partner/tests/test_sale_order_priority_partner.py
```diff
@@ -0,0 +1,71 @@
...
+
+from odoo.fields import Command
+from odoo.tests import TransactionCase
+
+
+class TestSaleOrderPriorityPartner(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner_urgent = cls.env["res.partner"].create(
```
Ricardoalso, line 13:
```suggestion
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
        cls.partner_urgent = cls.env["res.partner"].create(
```

missing  `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

sale_order_priority_partner/models/sale_order.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # override to use partner_id.sale_priority as default priority if set
+        for vals in vals_list:
+            partner = self.env["res.partner"].browse(vals.get("partner_id"))
```
Ricardoalso:
```suggestion
            partner_id = vals.get("partner_id") or self.env.context.get(
                "default_partner_id"
            )
            partner = self.env["res.partner"].browse(partner_id)
```

 If a caller creates an order relying on default_partner_id in context (quick-create/wizard flows)

Ricardoalso (commented):
Small comments, 

LG otherwise

Ricardoalso (approved):

sale_order_priority_partner/models/sale_order.py
```diff
@@ -0,0 +1,20 @@
...
+    def create(self, vals_list):
+        # override to use partner_id.sale_priority as default priority if set
+        for vals in vals_list:
+            partner_id = vals.get("partner_id") or self.env.context.get(
+                "default_partner_id"
+            )
+            partner = self.env["res.partner"].browse(partner_id)
+            if partner.sale_priority:
+                vals["priority"] = partner.sale_priority
+        return super().create(vals_list)
```
BhaveshHeliconia, line 20:
```suggestion
        # use partner_id.sale_priority as default priority if not explicitly provided
        for vals in vals_list:
            if "priority" not in vals:
                partner_id = vals.get("partner_id") or self.env.context.get(
                    "default_partner_id"
                )
                if partner_id:
                    partner = self.env["res.partner"].browse(partner_id)
                    priority = (
                        partner.sale_priority
                        or partner.commercial_partner_id.sale_priority
                    )
                    if priority:
                        vals["priority"] = priority
        return super().create(vals_list)
```
Since sale_priority is a default, we should only apply it if priority wasn't explicitly passed in vals (if "priority" not in vals:). Otherwise, creating an order via API or custom code with an explicit priority gets overwritten. We can also fallback to commercial_partner_id for child contacts.

BhaveshHeliconia:
I think it would be better to add an onchange on partner_id so the priority updates immediately when a customer is selected in the UI. Right now, it only updates after saving, which may be a bit confusing for the user.

ivantodorovich:
Hello @PaoloYam , can you check @BhaveshHeliconia 's comments above? 🙏🏻

BhaveshHeliconia:
@PaoloYam squash the commits into a single commit.

Ricardoalso:
> @PaoloYam squash the commits into a single commit.

Should be ready now @BhaveshHeliconia  :smiley:

PR #4541, opened 2026-08-21
[19.0][MIG] portal_sale_confirm_require_login: Migration to version 19.0

pilarvargas-tecnativa:
@Tecnativa TT64000

@victoralmau @carlos-lopez-tecnativa please review


portal_sale_confirm_require_login/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Tecnativa S.L. - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Portal Sale Confirm Require Order",
```
carlos-lopez-tecnativa:
```suggestion
    "name": "Portal Sale Confirm Require login",
```

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor change to the module name.

Additionally, with the demo user, I get this message. I don't understand why this happens. Could you review this?

I remember there was something special about the demo user, but I don't remember exactly what it was. With another contact/user, it works as expected.

Image

pilarvargas-tecnativa:
> Additionally, with the demo user, I get this message. I don't understand why this happens. Could you review this?
> 
> I remember there was something special about the demo user, but I don't remember exactly what it was. With another contact/user, it works as expected.

In Odoo's DEMO data, there are two demo contacts for the same user, but that doesn't happen in real use. A user won't have two contacts, and we encountered the same problem in Odoo 17

portal_sale_confirm_require_login/__manifest__.py
pilarvargas-tecnativa: thanks

carlos-lopez-tecnativa (approved):
LGTM

victoralmau (approved):

PR #4544, opened 2026-08-25
[19.0][MIG] sale_sourced_by_line

JasminSForgeFlow:
### Scope reduced: most of the feature is now in core

Odoo 19 `sale_stock` already provides the bulk of what this module used to add:

| 18.0 module behaviour | 19.0 |
|---|---|
| `warehouse_id` on `sale.order.line` | core — `sale_stock/models/sale_order_line.py` (computed, stored) |
| Stock moves / pickings sourced from the **line's** warehouse | core — `_prepare_procurement_values` passes `self.warehouse_id` |
| One delivery order per warehouse | core — falls out of the stock rules, no grouping code needed |
| `_get_procurement_group_key()` / `_prepare_procurement_group_vals()` | **removed** — `procurement.group` no longer exists in 19.0 (replaced by `stock.reference`), so there is nothing to override |

What remains, and is what this module now does:

* make `sale.order.line.warehouse_id` **user-editable** (core declares it compute-only, derived from the line routes);
* keep a manually set line warehouse from being overwritten — `_compute_warehouse_id` only recomputes lines where the field is empty;
* expose the field on the sale order line list and form;
* relabel `sale.order.warehouse_id` as *Default Warehouse* with its help text.

The end-user behaviour is unchanged from 18.0.

### Changes

* `__manifest__.py`: version `19.0.1.0.0`; dependency `sale_procurement_group_by_line` → **`sale_stock`**. The former has no 19.0 branch and its purpose (one procurement group per line) is obsolete now that `procurement.group` is gone.
* `model/sale.py`:
  * dropped the three procurement-group related methods (see table above);
  * `sale.order.line.warehouse_id` redefined with `readonly=False`;
  * `_compute_warehouse_id` `@api.depends`: `route_id` → `route_ids`, and `product_packaging_id` dropped (`product.packaging` was merged into `uom.uom` in 19.0);
  * **dropped the legacy `readonly=True` on `sale.order.warehouse_id`.** Core declares that field with `precompute=True`, and `models.create()` discards values for `precompute + readonly` fields (`odoo/orm/models.py`), so keeping `readonly=True` would silently swallow `warehouse_id` passed at sale order creation. The field stays labelled *Default Warehouse*.
* `view/sale_view.xml`:
  * line form xpath anchor `route_id` → `route_ids`;
  * removed the `{'default_warehouse_id': warehouse_id}` context on `order_line`. It is redundant — core computes and stores the line warehouse from the order — and it would both override core's route-based warehouse resolution and clobber the `context` attribute core sets on that field.
* `tests/`: rewritten. No demo data on 19.0, so all fixtures are created in `setUpClass`; assertions moved from procurement groups to picking count and `move_ids.warehouse_id`. Added a test asserting a line-level warehouse survives a change of the order warehouse.

@ForgeFlow


AaronHForgeFlow:
Hi @JasminSForgeFlow we have to make this module to depend on sale_stock_reference_by_line and adapt the methods to work like it was with the previous dependency.

JasminSForgeFlow:
> Hi @JasminSForgeFlow we have to make this module to depend on sale_stock_reference_by_line and adapt the methods to work like it was with the previous dependency.

Done, Thanks

JasminSForgeFlow:
CI fixed here https://github.com/OCA/sale-workflow/pull/4567

AaronHForgeFlow (commented):
Code + Functional LG. Just please split sale.py, one file per model.

PR #4227, opened 2026-03-24
[17.0][FW] sale_invoice_policy: Port PR #3376

rousseldenis:

PR #4554, opened 2026-08-31
[17.0][FIX] sale_invoice_policy: drop unsupported states property

GuillemCForgeFlow:
Move readonly prop to the view, the supported way in odoo 17.0+

removing the warning introduced in https://github.com/OCA/sale-workflow/pull/4227

```
WARNING devel py.warnings: /opt/odoo/custom/src/odoo/odoo/fields.py:542: UserWarning: Since Odoo 17, property sale.order.invoice_policy.states is no longer supported.
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 542, in setup
    warnings.warn(f'Since Odoo 17, property {self}.states is no longer supported.')
```

cc @ForgeFlow

### DIFF
--- a/sale_invoice_policy/models/sale_order.py
+++ b/sale_invoice_policy/models/sale_order.py
@@ -17,7 +17,6 @@ class SaleOrder(models.Model):
         store=True,
         readonly=False,
         required=True,
-        states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
         precompute=True,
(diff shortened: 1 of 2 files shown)


GuillemCForgeFlow:
@rousseldenis perhaps we can merge this small fix directly? 😄

rousseldenis (approved):

PR #4556, opened 2026-08-31
[18][ADD]sale_order_type_filter_product

Anxo82:
@luis-ron @ValentinVinagre 

- This PR adds a new module to restrict products on sale orders based on the selected sale order type.
Restrictions can be configured at product or variant level. It also allows configuring whether incompatible lines should prevent the sale order type change or be automatically removed, with multi-company support.

- [T-10183]


luis-ron (approved):
Functional review: LGTM 👍🏻

PR #4557, opened 2026-08-31
[18.0][FIX] sale_global_discount: discount amounts with round_globally rounding method

LorenzoC0:
This patch disables the calculation of global discount amounts if none is selected and computes discount amounts consistently from per-line subtotals instead of mixing them with globally rounded order totals.

You can trigger a -0,01€ discount without any global discount selected (or a global discount set to 0%) with these steps:

1. Create a sales order. 
2. Set the unit price precision to 3 decimal places. 
3. In the sales order, create 4 lines: 
    - Product 1 - qty 3 - unit price 3.460; 
    - Product 2 - qty 1 - unit price 25,000; 
    - Product 3 - qty 6 - unit price 4,713; 
    - Product 4 - qty 6 - unit price 4,501. 
4. Do not set a global discount. 
5. Check the amount of the global discount applied.

### DIFF
--- a/sale_global_discount/models/sale_order.py
+++ b/sale_global_discount/models/sale_order.py
@@ -100,11 +100,26 @@ class SaleOrder(models.Model):
         res = super()._compute_amounts()
         for order in self:
             order._check_global_discounts_sanity()
-            amount_untaxed_before_global_discounts = order.amount_untaxed
-            amount_total_before_global_discounts = order.amount_total
             discounts = order.global_discount_ids.mapped("discount")
+            if not any(discounts):
+                order.update(
+                    {
+                        "amount_untaxed_before_global_discounts": (
+                            order.amount_untaxed
+                        ),
+                        "amount_total_before_global_discounts": order.amount_total,
+                        "amount_global_discount": 0.0,
+                    }
+                )
+                continue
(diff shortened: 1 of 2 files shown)


sale_global_discount/models/sale_order.py
```diff
@@ -100,11 +100,27 @@ def _compute_amounts(self):
...
+                continue
+            amount_total_before_global_discounts = order.amount_total
+            amount_untaxed_before_global_discounts = 0
             amount_discounted_untaxed = amount_discounted_tax = 0
             for line in order.order_line:
+                # Compare per-line rounded amounts with per-line rounded
+                # amounts.Mixing them with the core (possibly globally
+                # rounded) untaxed amount would leak rounding differences
+                # into the discount amount.
+                amount_untaxed_before_global_discounts += line.price_subtotal
```
eLBati, line 121: To reduce the risk of discrepancies, can't we calculate both based on the rows?

```
for line in order.order_line:
    amount_untaxed_before_global_discounts += line.price_subtotal
    amount_total_before_global_discounts += line.price_total
    ...
```

LorenzoC0: Nice catch, thank you.

eLBati (approved):
Thanks

PR #4562, opened 2026-09-03
[19.0][FIX] sale_pricelist_global_rule: cannot access local variable 'res'

adasatorres-tecnativa:
@Tecnativa TT64094
@pedrobaeza @pilarvargas-tecnativa 

When self is empty, an error was raised because res was defined inside the for loop. This should fix the issue.


pedrobaeza (approved):
Clear patch, so merging.

/ocabot merge patch

PR #4380, opened 2026-06-02
[18.0][IMP] sale_order_secondary_unit: Consider 2nd UoM in product catalog

CarlosRoca13:
cc @Tecnativa TT62824

ping @sergio-teruel @carlosdauden

PR #4564, opened 2026-09-07
[18.0][FIX] sale_stock_secondary_unit: Propagate a count correction to the pending move

carlosdauden:
Correcting the piece count on an already-confirmed line (without touching the primary quantity) had no effect on the still-pending delivery move. The write() override built a procure_secondary_uom_qty context and relaunched the stock rule (_action_launch_stock_rule), mirroring what this module did in 15.0 - but that call is a no-op in 18.0: _get_qty_procurement() derives "quantity already procured" from the actual existing moves and ignores its previous_product_uom_qty argument entirely, so with product_uom_qty unchanged it always finds qty == line.product_uom_qty and skips, never touching an existing move's secondary_uom_qty at all (confirmed empirically against a throwaway savepoint before writing the fix).

Replaced it with a direct write of the corrected quantity onto the pending (non-done, non-cancelled) move, when there is exactly one - which also recomputes the move's own product_uom_qty from the new count for "secondary_priority" units, since the primary quantity is still estimated from the secondary one. A line already partially delivered (several pending moves splitting it) isn't handled - would need a proportional distribution, left as a documented limitation.

Also fixed a live bug in the same method: the delta-precision calculation read self.secondary_uom_id.uom_id.rounding (self being the whole, possibly multi-line, write() recordset) instead of line.secondary_uom_id.uom_id.rounding - raises "Expected singleton" on a multi-line write spanning different secondary units.

Also made stock.move's own _prepare_procurement_values() (added this session in stock_secondary_unit) fall back correctly when no sale line is found for a move (e.g. a pure internal replenishment) - this module's override used to unconditionally overwrite secondary_uom_id/secondary_uom_qty to False/0.0 in that case, clobbering whatever the base module had already derived from the move itself.

[IMP] sale_stock_secondary_unit: Distribute a count correction across several pending moves

The previous fix only handled the common case of a single pending move still covering a line's full remaining demand - a line already partially processed (a backorder pick still pending alongside an earlier leg not yet processed) has several pending moves at once, and was left untouched, a documented limitation.

Distribute the corrected total proportionally to each pending move's current secondary_uom_qty share of the previous total, largest- remainder style: every move but the last gets its rounded share, the last absorbs whatever's left so the sum always stays exact. Falls back to an even split when there's nothing to base a proportion on (no count entered on any of them yet).

@Tecnativa

Depends on:
 - #4380 Included here to avoid conflicts with tests file


pedrobaeza:
@sbidoul do you know if there's a problem with runboat?

sbidoul:
I htink runboat is ok. I clicked `reset` and you can see the build log that is stuck at `loading sale_stock/data/sale_order_demo.xml`.

pedrobaeza:
Uhm, maybe a fault in Odoo upstream... thanks for checking.

carlos-lopez-tecnativa (changes requested):
Please check the commits. I think you force-pushed a merged branch together with other branches.

sale_stock_secondary_unit/tests/test_sale_stock_secondary_unit.py
```diff
@@ -59,6 +59,43 @@ def setUpClass(cls):
...
+                "list_price": 100.00,
+            }
+        )
+        cls.product_2.product_tmpl_id.write(
+            {
+                "secondary_uom_ids": [
+                    (
+                        0,
+                        0,
+                        {
```
carlos-lopez-tecnativa:
```suggestion
                   Command.create(
                        {
```

carlos-lopez-tecnativa (changes requested):
I think it would be great to add the field to this view in the Sales menu.

Image

carlosdauden:
Changes done

PR #4565, opened 2026-09-07
[18.0][FIX] sale_order_general_discount: Restore discount reset on removal

carlosdauden:
_compute_discount() lost its reset-to-zero branch between 17.0 and 18.0 - removing the order's general discount no longer cleared it back off of already-discounted lines, leaving them stuck at the last applied percentage.

Restructured rather than literally restoring the old else branch: the current test suite's own expectation (confirmed unrelated to this bug) is that a bypass_general_discount line is never touched by this compute at all, not reset to 0 either - skip those lines entirely, then apply the general discount or reset to 0 for the rest.

@Tecnativa

ping @sergio-teruel @CarlosRoca13 @carlos-lopez-tecnativa

### DIFF
--- a/sale_order_general_discount/models/sale_order_line.py
+++ b/sale_order_general_discount/models/sale_order_line.py
@@ -20,8 +20,10 @@ class SaleOrderLine(models.Model):
             # the case where a discount was set to a value != 0 and then
             # set again to 0 to remove the discount on all the lines at the same
             # time
-            if not line.product_id.bypass_general_discount and (
-                line.order_id.general_discount or line.order_id._origin.general_discount
-            ):
+            if line.product_id.bypass_general_discount:
+                continue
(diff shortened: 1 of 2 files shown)


carlos-lopez-tecnativa (commented):
LGTM. 
Just a question that came up while I was reviewing this module: why doesn't general_discount have the `sale.group_discount_per_so_line` group? Is this correct?

carlosdauden:
The ‘sales.group_discount_per_so_line’ group merely restricts the field at the view level, with no further technical implications; it is therefore up to the user to decide whether or not to display the detailed discount per line.

carlos-lopez-tecnativa (approved):

BhaveshHeliconia (approved):

sergio-teruel (approved):

carlosdauden:
Ready to merge

PR #4569, opened 2026-09-09
[18.0][FIX] sale_order_type: Apply the type analytic distribution to the order lines

carlosdauden:
The migration to 18.0 dropped sale.order.analytic_account_id and its compute because the core replaced the order level analytic account by the line level analytic_distribution, but nothing took over the analytic account configured on the type: the field stayed editable on sale.order.type and had no effect at all on the orders, unlike every other setting of the type (route, pricelist, payment term, salesperson, team, incoterm), which did keep its compute adapted to the new architecture.

Following the design already merged upstream for 19.0, the type inherits analytic.mixin and now holds a full analytic distribution rather than a single account, which is how 18.0 expresses analytic assignment. A post-migration script turns the previous analytic_account_id into a 100% distribution on that same account.

sale.order.line._compute_analytic_distribution is extended with the pattern the module uses for the rest of the type settings, so the distribution of the type is applied after the one resolved by account.analytic.distribution.model and prevails over it. Choosing a type is an explicit decision taken order by order, the same way its pricelist prevails over the partner one, and a company wide catch-all distribution model would otherwise neutralize the setting completely. Lines with a display_type are skipped, as the core compute does.

@Tecnativa

### DIFF
--- a/sale_order_type/migrations/18.0.1.5.0/post-migration.py
+++ b/sale_order_type/migrations/18.0.1.5.0/post-migration.py
@@ -1,0 +1,20 @@
+# Copyright 2026 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.tools.sql import column_exists
+
+
+def migrate(cr, version):
+    """Turn the old single analytic account into a full analytic distribution."""
+    if not column_exists(cr, "sale_order_type", "analytic_account_id"):
+        return
+    cr.execute(
+        """
+        UPDATE sale_order_type
+        SET analytic_distribution = jsonb_build_object(
+            analytic_account_id::text, 100.0
+        )
+        WHERE analytic_account_id IS NOT NULL
+        AND analytic_distribution IS NULL
+        """
+    )
--- a/sale_order_type/tests/test_sale_order_type.py
+++ b/sale_order_type/tests/test_sale_order_type.py
@@ -4,7 +4,7 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 from freezegun import freeze_time
 
-from odoo import fields
+from odoo import Command, fields
 from odoo.tests import Form
 
 from odoo.addons.base.tests.common import BaseCommon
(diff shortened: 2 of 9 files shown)


sale_order_type/migrations/18.0.1.5.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+        """
+        SELECT EXISTS (
+            SELECT 1 FROM information_schema.columns
+            WHERE table_name = 'sale_order_type'
+            AND column_name = 'analytic_account_id'
+        )
+        """
+    )
+    if not cr.fetchone()[0]:
+        return
```
carlos-lopez-tecnativa: You can use the [column_exists](https://github.com/odoo/odoo/blob/21eb3047af51b427180063b331201bd8e2b1176f/odoo/tools/sql.py#L325) function.

sale_order_type/tests/test_sale_order_type.py
```diff
@@ -239,6 +253,72 @@ def test_sale_order_flow_route(self):
...
+                        0,
+                        0,
+                        {
+                            "product_id": self.product.id,
+                            "name": self.product.name,
+                            "product_uom_qty": 2.0,
+                            "price_unit": self.product.lst_price,
+                        },
+                    )
+                ]
```
carlos-lopez-tecnativa, line 286:
```suggestion
                "order_line": [
                    Command.create(
                        {
                            "product_id": self.product.id,
                            "name": self.product.name,
                            "product_uom_qty": 2.0,
                            "price_unit": self.product.lst_price,
                        },
                    )
                ]
```

carlos-lopez-tecnativa (approved):
LGTM, just a non-blocking comment, but if you can improve this, that would be great.

sale_order_type/migrations/18.0.1.5.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+    cr.execute(
+        """
+        UPDATE sale_order_type
+        SET analytic_distribution = jsonb_build_object(
+            analytic_account_id::text, 100.0
+        )
+        WHERE analytic_account_id IS NOT NULL
+        AND analytic_distribution IS NULL
+        """
+    )
```
carlos-lopez-tecnativa, line 20: Now I think this change should be performed in V16 instead of V18.

Additionally, you can use the [fill_analytic_distribution](https://github.com/OCA/openupgradelib/blob/10ab8ff44e182b9c6420707578b6408feded200b/openupgradelib/openupgrade_160.py#L442) function:

```suggestion
      openupgrade_160.fill_analytic_distribution(
              env,
              table="sale_order_type",
              m2m_rel=False,
              m2m_column1=False,
              m2m_column2=False,
              column="analytic_distribution",
              analytic_account_column="analytic_account_id",
          )
```

carlosdauden: sale.order.analytic_account_id (the header-level Many2one) is still present in core on both 16.0 and 17.0 — it's only removed in 18.0, fully replaced by line-level analytic_distribution. sale_order_type's own models/sale.py still relies on that single-account header field on those versions.

carlosdauden:
Changes done

CarlosRoca13 (approved):

carlosdauden:
@sergio-teruel can you review?

marielejeune (approved):
LGTM (code review)

PR #4577, opened 2026-09-10
[17.0][MIG] sell_only_by_packaging: Migration to 17.0

Anxo82:
@HaraldPanten @Jaimermaccione 

- [T-10156]


Jaimermaccione (changes requested):
@Anxo82 Reviewed on local environment. In general the module works well, but I found some issues and edge cases during normal sales operations. Here is a summary of what is working and what it should be reviewed:

---
#### What works well
* **Packaging validation:** Once a packaging is set, Odoo blocks saving if the quantity is not complete.
* **Editing existing lines:** When a line already has a packaging, changing the quantity correctly rounds up to the next multiple.
* **Master data safety:** It prevents deactivating packaging levels that are in use by active products.
---
####  Points to review
#####  1. Creating a new line vs. editing an existing line
* **Current behavior:** On a product with "Sell only by packaging", if you enter a quantity that is not a multiple on a new line, it does not round up. The packaging field stays empty and Odoo shows a validation error when saving. The automatic rounding only works if the line was already saved with a correct multiple.
* **Expected behavior:** When selecting a product with "Sell only by packaging", Odoo should automatically set its default packaging and round the quantity to the nearest multiple directly when saving.
#####  2. Users without packaging permissions
* **Current behavior:** Sales users without the group **"Manage Product Packaging"** (`product.group_stock_packaging`) cannot see the packaging columns in the sales order. When they try to sell a restricted product, they get an error popup and cannot see or select the packaging.
* **Expected behavior:** The packaging columns should be visible for sales users on the order lines, or the packaging should be assigned automatically in the background so users are not blocked.
#####  3. Odoo Product Catalog
* **Current behavior:** When clicking "Add" in the Product Catalog view, it fails with a validation error because Odoo tries to add 1 unit. Also, the catalog does not show packaging info or allow increasing by pack quantities.

Image

Image

* **Expected behavior:** 
  * *Option A (development):* The "Add" button should insert 1 full package quantity, and the `+` / `-` buttons should increase by package multiples.
  * *Option B (documentation):* If catalog integration is too complex right now, we can add a note in the **"Known Issues / Roadmap"** section of the README advising users to add these products using the normal order lines.
#####  4. Products with variants
* **Current behavior:** It is possible to enable "Sell only by packaging" on the product template even if only one variant has a packaging configured. This leaves all other variants impossible to sell, because they inherit the rule but have no packagings.
* **Expected behavior:** Proposal: Keep the option on the template, but validate that **all active variants** have at least one salable packaging before allowing to save the template.
#####  5. External channels and recommendation tools
* **Note:** Tools that create sales lines directly—such as e-commerce connectors, Excel imports, or the `sale_order_product_recommendation` wizard—will probably fail with validation errors if the quantities are not exact multiples.
* **Suggestion:** If we ensure the default packaging and quantity rounding are always handled when creating lines, we will also avoid issues with these tools.
---
####  Suggested updates for the README
1. **Permissions (`Configuration`):** Add a note explaining that users need the **"Manage Product Packaging"** group to see and use packaging fields on sales orders.
2. **Catalog limitation (`Known Issues / Roadmap`):** Mention that adding products through the Odoo 17 Product Catalog is currently not supported (if we choose Option B).
3. **Usage example (`Usage`):** *Only if we do not fix Point 1 in code*, we should update the example in the README to explain that the user needs to select the packaging first (or enter an exact multiple) before the automatic rounding works on edits.
Let me know what you think and how you prefer to handle these points!

Thank you!

PR #4578, opened 2026-09-11
[19.0][MIG] sale_order_line_price_lock_by_pricelist: Migration to 19.0

kalpeshgajera-uncannycs:


sale_order_line_price_lock_by_pricelist/tests/test_sale_order_line_price_lock_by_pricelist.py
```diff
@@ -0,0 +1,129 @@
...
+        cls.user_sales_sales_manager = mail_new_test_user(
+            cls.env,
+            login="user_sales_sales_manager",
+            name="Paco Sales Manager",
+            email="crm_salesmanager@test.example.com",
+            company_id=cls.env.company.id,
+            notification_type="inbox",
+            groups="sales_team.group_sale_manager",
+        )
+        cls.customer = cls.env["res.partner"].create(
```
prayagrk, line 86: You could consider avoiding the creation of a res.partner record when using BaseCommon, as it is already provided by the base class.

sale_order_line_price_lock_by_pricelist/tests/test_sale_order_line_price_lock_by_pricelist.py
```diff
@@ -0,0 +1,129 @@
...
+                "list_price": 100.0,
+            }
+        )
+        cls.product_other = cls.env["product.template"].create(
+            {
+                "name": "Test product 2",
+                "list_price": 120.0,
+            }
+        )
+        cls.customer = cls.env["res.partner"].create(
```
prayagrk, line 25: This too.

Twaze17:
Thank you for the migration. 

Functional Test Review : 

Configured AS IS from the README but I could modify the fixed price from the SO with a pricelist when the lock is activated. 

Did I miss something ?

PR #4582, opened 2026-09-13
[18.0][FIX] sale_planner_calendar: New order from the planner opens the product catalog

carlosdauden:
action_open_sale_order(new_order=True) never actually created a sale order - it just returned the base "Quotations" action as-is, whose own default view_mode is list,kanban,form,... (list first) and carries no partner/salesperson context. So the button intended to start a brand new order from the calendar planner opened the plain quotations LIST instead of a blank, pre-filled order.

This used to be routed through a configurable
sale_planner_calendar.action_open_sale_order system parameter, meant to let an installation swap in a different starting action - originally the now-retired sale_order_product_picker's own picker view. Its 18.0 successor, product.catalog.mixin.action_add_from_catalog(), isn't a standalone act_window an admin could point a parameter at: it is an instance method that needs a real order to already exist. So that configurability had no second sane value left to offer - dropped entirely in favor of a hardcoded sale.action_quotations_with_onboarding for the existing-order path, matching the parameter's own previous default.

For new_order=True, the order is now actually created (same partner/commercial-partner and salesperson defaults the existing-order path already computed, extracted into
_get_planner_sale_order_partner()), linked to the event via sale_planner_calendar_event_id, and action_add_from_catalog() is returned instead - landing the user straight on the product catalog, the closest 18.0-native equivalent of the old picker experience. action_add_from_catalog() comes from product.catalog.mixin, already inherited by sale.order in core sale (a hard dependency of this module) - deliberately not depending on sale_product_catalog_extended, since that optional extension may not be installed alongside this module.

Add a setting to keep the partner's own salesperson.

@Tecnativa


carlos-lopez-tecnativa (commented):
I didn't change any configuration. I only installed the module, clicked the button, and got this error. Am I missing any configuration?
Image

carlosdauden:
Changes done

carlos-lopez-tecnativa (approved):

CarlosRoca13 (approved):
Thanks

pedrobaeza:
Please rebase and check the CI

carlosdauden:
Unrelated test error caused by sell_only_by_packaging

carlosdauden:
CI green :green_apple:

PR #4583, opened 2026-09-14
[18.0][FIX] sale_exception: regression test for MissingError on confirmed order

sergio-teruel:
Reproduces the production crash directly: adding a product line to an already confirmed sale order raised MissingError, because base_exception.detect_exceptions() re-checked the new (not yet committed) sale.order.line through a genuinely independent second DB connection that could not see it yet.

Odoo's TestCursor shares the same underlying connection as the main test transaction (always "sees its own writes"), so it can't reproduce a real cross-connection visibility bug on its own. This test opens a real second connection with db_connect(...).cursor() - the same technique core uses in its own test_db_cursor.py - and substitutes it via mock_base_exception_method_env, matching what a real (non-test) request does.

@Tecnativa TT64467

ping @carlosdauden @carlos-lopez-tecnativa

depends on:

- [x] https://github.com/OCA/server-tools/pull/3725


carlosdauden (approved):

carlos-lopez-tecnativa (commented):
Could you please temporarily add a reference to the other PR in the test-requirements.txt file so it can be tested easily in Runboat?

sergio-teruel:
@carlos-lopez-tecnativa Rebase done, the https://github.com/OCA/server-tools/pull/3725 PR was merged.

carlos-lopez-tecnativa (approved):

PR #4584, opened 2026-09-14
[FIX]sale_exception: avoid MissingError on new sale order lines

Raul-S73:
# Suggested commit title

`[FIX] sale_exception: avoid MissingError on new sale order lines`

### Changes proposed in this PR

**Avoid `MissingError` when detecting exceptions after creating a sale order line.**

When a confirmed sale order is edited and a new line is added, `sale_exception` can run its exception detection from `sale.order.write()` because `order_line` is one of the fields that triggers the check.

The exception detection introduced by commit f6954851c stores exception relations using a dedicated database cursor. A line created by the current transaction is visible from the main cursor, but it is not visible from that dedicated cursor until the transaction is committed. The line recordset was nevertheless passed to `lines.detect_exceptions()`. When the secondary environment tried to access the line, Odoo raised:

```text
MissingError: The record does not exist or has been deleted.
(Record: sale.order.line(<new_id>,), User: 1)
```

This change checks which order lines are visible from the dedicated cursor and only runs line-level exception detection on those persisted lines. Order-level exception detection remains unchanged, and existing persisted lines continue to be processed normally. The check also keeps the existing behavior of using the current cursor while installing modules or running tests.

### How to test the changes

1. Install `sale_exception` and keep an applicable exception rule enabled.
2. Create and confirm a sale order.
3. Add a new product line to the confirmed order.
4. Verify that adding the line no longer raises a `MissingError` for the new `sale.order.line` record.
5. If the order still violates a configured blocking rule, verify that the normal sale exception handling is shown instead of the technical error.

The original failure was reproduced on Odoo 18 with a confirmed order while
adding a line through the web client.

### Notes

- This fixes the transaction-visibility issue without uninstalling `sale_exception` or disabling exception validation globally.
- `python3 -m py_compile sale_exception/models/sale_order.py` passes.
- `git diff --check` passes.
- The full `pre-commit` run could not be completed in the local environment because its hook repositories could not be downloaded from GitHub.
  
### Error
  
image


Reyes4711-S73 (approved):
LGTM

arantxa-s73 (approved):
LGTM

BhaveshHeliconia (approved):

Reyes4711-S73:
@OCA/crm-sales-marketing-maintainers Please, can you review/merge this PR?

pedrobaeza:
@grindtildeath as you committed the new cursor commit, can you please check?

grindtildeath:
Hello @Raul-S73 

Looks legit, thank you.

FTR I'm also working on a similar issue where records being updated prior to the exceptions detection, and their update are not visible by the new cursor used to persist the exception, which in turn raise the exception when it should not. 

I'll try to consider your use case and see if we can have a better handling of newly created / modified records before the new cursor is used. However it would still be great if we can have a test for this.

grindtildeath (approved):

sergio-teruel (changes requested):
Thanks for tackling this. I think filtering the new lines out avoids the `MissingError` at the cost of not checking those lines at all, which silently bypasses line rules.                                                                                                                               
                                                                                                                                                                                                                                                                                                            
  To reproduce, on 18.0 with this PR: activate a `sale.order.line` rule (e.g. `sale_exception.excep_no_dumping`, sold below cost), confirm an order, then add a new line that matches the rule and save. The order is saved with no exception and no popup. Before this PR, the same save failed with       
  `MissingError`. The error message was confusing, but at least the save was blocked. Lines added after confirmation are often the ones these rules are meant to catch, so it's safer if they are still evaluated.                                                                                          
                                                                                                                                                                                                                                                                                                            
  The underlying problem is that the independent connection writing the exceptions cannot see records created by the ongoing transaction. We hit a related case where that connection also waits forever for rows locked by the ongoing transaction, which in production left every HTTP worker blocked.    
  Both are handled in:                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                            
  - https://github.com/OCA/server-tools/pull/3739: `_write_exceptions_independently()` writes through the independent connection with a lock timeout. If that times out or raises `MissingError`, it writes the exceptions in the ongoing transaction (so they are still detected and the popup is shown)   
  and writes them again after a rollback, skipping records that no longer exist.                                                                                                                                                                                                                            
  - https://github.com/OCA/sale-workflow/pull/4599: order lines use that helper, with regression tests that use genuinely separate connections (the test cursor shares a single connection, so neither problem can show up with it).                                                                        
                                                                                                                                                                                                                                                                                                            
  With those, adding a rule-matching line to a confirmed order raises the exception as expected, so this PR would no longer be needed. What do you think?

PR #4588, opened 2026-09-15
[18.0][FIX] sale_automatic_workflow: do not register zero amount payments

astirpe:
### Problem

`_register_payments` creates a payment using `invoice.amount_residual`.
If that amount is zero, the payment changes nothing about the invoice.
The invoice still matches the payment filter five minutes later, so the
cron creates another payment. And another, on every run.

Nothing shows up in the log, because `savepoint()` swallows exceptions.

This is easy to hit with invoices that total zero: a 100% discount, a
free first order, a replacement shipment. They have a zero residual from
the moment they are posted and can stay at `payment_state = 'not_paid'`.

### Fix

1. `_register_payment_invoice` returns an empty recordset when the
   residual amount is zero.
2. New `_do_register_payment` method, so payment registration re-checks
   the filter and returns "job bypassed" like the other four actions.
3. `('amount_residual', '!=', 0)` added to the default payment filter,
   with a migration script for existing databases. The script leaves a
   customised filter alone.

The same code is in 14.0 through 19.0.


sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -149,18 +149,41 @@ def _prepare_dict_account_payment(self, invoice):
...
         self = self.with_company(invoice.company_id)
+        # Never create a payment that settles nothing. Such a payment leaves
+        # the invoice exactly as it found it, so the invoice keeps matching
+        # the payment filter and a new payment is created on every run of the
+        # cron, forever.
+        if invoice.currency_id.is_zero(invoice.amount_residual):
+            _logger.info(
+                "Skipping payment registration for %s: residual amount is zero.",
+                invoice.display_name,
+            )
```
lmignon: Is-it really useful? 🤔

sale_automatic_workflow/models/automatic_workflow_job.py
```diff
@@ -149,18 +149,41 @@ def _prepare_dict_account_payment(self, invoice):
...
+    def _do_register_payment(self, invoice, domain_filter):
+        """Register a payment for an invoice, filter ensure no duplication"""
+        if not self.env["account.move"].search_count(
+            [("id", "=", invoice.id)] + domain_filter
+        ):
+            return f"{invoice.display_name} {invoice} job bypassed"
+        if invoice.currency_id.is_zero(invoice.amount_residual):
+            return f"{invoice.display_name} {invoice} job bypassed: nothing to pay"
+        self._register_payment_invoice(invoice)
+        return f"{invoice.display_name} {invoice} register payment successfully"
```
lmignon: The check over is_zero is redundant with the one done in _register_payment_invoice... why not

```suggestion
        payment = self._register_payment_invoice(invoice)
        if not payment:
            return f"{invoice.display_name} {invoice} job bypassed: nothing to pay"
        return f"{invoice.display_name} {invoice} register payment successfully"
```

lmignon (changes requested):
Thank you for the fix... some little comments

sale_automatic_workflow/models/automatic_workflow_job.py
astirpe: @lmignon 
Thank you for your review! I changed this to _logger.debug(), is it ok?

lmignon (approved):
LGTM (Code review only)

BhaveshHeliconia (approved):

rousseldenis (approved):

PR #4590, opened 2026-09-16
[19.0][IMP] sale_stock_picking_blocking

SilvioC2C:
- refactor test suite to improve heritability and customization
- fix block reason being recomputed when saving in the SO form view (even if the block was manually set by the user)
- fix block reason being recomputed on SOs that have been already confirmed

Depends on: https://github.com/OCA/sale-workflow/pull/4591


SilvioC2C:
Tests currently failing because, when using `odoo.tests.form.Form` with a sale order, Odoo builds the arch with a smart button with computed field `payment_transaction_count` (from module `sale_transaction_form_link`) that counts items in field `transaction_ids`. That field's access is restricted to a specific accounting group, which the test user doesn't have, but the button visibility is not dependent on that group:

- `with Form(sale)` mimics onchanges and defaults triggered when the form view is accessed
- it tries to compute `payment_transaction_count`, whose value is visible in the smart button (always displayed, since it has no `groups` attribute) 
- the field's computed method needs to read field `sale.order.transaction_ids`
- the field is accessible only to accounting users
- an `AccessError` is raised

I will propose a fix for that in another PR. 

EDIT: fixed :green_circle:
See: https://github.com/OCA/sale-workflow/pull/4591

ivantodorovich (approved):
Thanks!

ivantodorovich:
depended PR got merged 🥳

Ricardoalso (approved):

rousseldenis (approved):

